PR #1, opened 2016-10-29
StyleCop

dlemstra:
This adds StyleCop.Analyzers that will enable StyleCop during the build. The documentation for the library can be found here: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/DOCUMENTATION.md. I added stylecop.json file and you can find more information about how that file works here: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md.

### DIFF
--- a/src/ImageSharp/Common/Helpers/ImageMaths.cs
+++ b/src/ImageSharp/Common/Helpers/ImageMaths.cs
@@ -171,8 +171,8 @@ namespace ImageSharp
             const float Epsilon = .00001f;
             int width = bitmap.Width;
             int height = bitmap.Height;
-            Point topLeft = new Point();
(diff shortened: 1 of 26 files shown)


JimBobSquarePants:
Does this also list stylecop analysers as a dependency in the output Nuget package?

dlemstra:
I was wondering that also. I haven't tested that yet but I can take a look at it. And we should probably also create a .ruleset file that we should configure in the same way as the stylecop.json file (http://stackoverflow.com/a/38094474/2410453)

dlemstra:
Just pushed a patch to resolve the dependency.

PR #17, opened 2016-10-30
Jpeg

codegateau:
This might not be directly related to imaging , but it is worth to be viewed .

Try to use Fresnel for dielectrics and improve the accuracy of GGX .


JimBobSquarePants:
Hi @codegateau  thanks for your PR but I'm not sure what you are trying to achieve here?

I already have a [working jpeg implementation](https://github.com/JimBobSquarePants/ImageSharp/tree/5216b86b518eddfdef68bff3cf954b5e16b9a2b5/src/ImageSharp/Formats/Jpg) within the master branch. It's extremely rough around the edges being a port from the Golang implementation but I plan to improve upon that given time.

Could you explain further?

Cheers

James

PR #20, opened 2016-11-05
Ensure Color structure stores elements in RGBA order

mellinoe:
@JimBobSquarePants I figured I'd take a shot at fixing this per our discussion yesterday. Let me know what you think about this approach. I added an additional test that ensures the bytes are laid out as expected in memory.

### DIFF
--- a/src/ImageSharp/Colors/Color.cs
+++ b/src/ImageSharp/Colors/Color.cs
@@ -19,6 +19,11 @@ namespace ImageSharp
     /// </remarks>
     public partial struct Color : IPackedPixel<uint>, IEquatable<Color>
     {
+        private const int RedShift = 0;
+        private const int GreenShift = 8;
+        private const int BlueShift = 16;
+        private const int AlphaShift = 24;
+
         /// <summary>
         /// The maximum byte value.
         /// </summary>
@@ -63,6 +68,9 @@ namespace ImageSharp
             {
                 throw new ArgumentException("Hexadecimal string is not in the correct format.", nameof(hex));
             }
+
+            // Order parsed from hex string will be backwards, so reverse it.
+            packedValue = Pack(A, B, G, R);
         }
 
         /// <summary>
@@ -106,12 +114,12 @@ namespace ImageSharp
         {
             get
             {
-                return (byte)(this.packedValue >> 24);
+                return (byte)(this.packedValue >> RedShift);
             }
 
             set
             {
-                this.packedValue = this.packedValue & 0x00FFFFFF | (uint)value << 24;
+                this.packedValue = this.packedValue & 0xFFFFFF00 | (uint)value << RedShift;
             }
         }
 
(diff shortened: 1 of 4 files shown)


JimBobSquarePants:
Ah excellent! 👍   @mellinoe  This is a good start, I wasn't looking forward to doing this.

There's a couple of other files that will require updating also to finish up.

All the protected methods here
https://github.com/JimBobSquarePants/ImageSharp/blob/d4f8f8ceac73d35617108701981789da534954e7/src/ImageSharp/Image/PixelAccessor.cs#L232

And their overrides here
https://github.com/JimBobSquarePants/ImageSharp/blob/d4f8f8ceac73d35617108701981789da534954e7/src/ImageSharp/PixelAccessor.cs

Looking at the original implementations, I don't believe we need to check for endianness since we know the order in which the bytes are stored and will deliver them the correct way in the decoders/encoders.

dlemstra:
I think you are right that we can remove the endianness check since we don't care about the actual integer value. Feel free to remove them.

mellinoe:
Could you give me a bit more guidance as to what the `PixelAccessor` does and what needs to be changed in it? Right now it looks fairly correct to me, but I'm not sure what the intended semantics are.

mellinoe:
Since the current tests pass, maybe you could just show me a failing test case that I could add and then fix.

JimBobSquarePants:
@mellinoe Yeah of course, The methods identified in `PixelAccessor<TColor, TPacked>` and the optimised `PixelAccessor` classes allow it to read and write to the encoders in the the requested order. These methods will eventually be used to optimise the codecs (Bmp has already been done)

`ComponentOrder.XYZW` is the equivalent of `RGBA` in `Color` and `ComponentOrder.ZYXW` of `BGRA`.

There is a visual test [DecodeThenEncodeImageFromStreamShouldSucceed](https://github.com/JimBobSquarePants/ImageSharp/blob/d4f8f8ceac73d35617108701981789da534954e7/tests/ImageSharp.Tests/Formats/GeneralFormatTests.cs#L48) Which should be giving your garbled output with your PR in its current status. Fixing, the order we read/write in the identified methods should fix them.

Hope that helps!

mellinoe:
The test images look okay to me, as far as I can tell... Could you take a look for me and see if anything is wrong?

JimBobSquarePants:
Yeah, will do. 🙂

JimBobSquarePants:
Odd... I would have sworn we would have had to change those methods. 

@mellinoe  As long as everything is now working for you (no more red) then I'm happy to merge this. @dlemstra Can you just double check I've not missed something?

dlemstra:
I just pushed some unit tests (https://github.com/JimBobSquarePants/ImageSharp/commit/e82a5f6e3c0e019d542e0471222d39a7f7a13fac) that make sure the optimized version of PixelAccessor still works after your changes. Copying in your file breaks these tests but its pretty easy to resolve. You will need to fix the order here: https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/PixelAccessor.cs#L63 and here https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/PixelAccessor.cs#L80.

mellinoe:
Latest version should be correct, and tests work for me.

dlemstra:
Looks great, and thanks for the help. I'll let you handle the merge @JimBobSquarePants.

JimBobSquarePants:
Me also, Excellent stuff. Very much appreciated @mellinoe

PR #23, opened 2016-11-11
Don't allocate memory for png decode filters, reuse scanline buffers

boguscoder:
This PR is part of perf improvements
- Stop allocating memory for png decode filters, reuse scanline buffers
- Added png test images with different filters for local tests

P.S. This passed on test data before your last encoder commit, need to retest once you have fix...
P.S2 This revealed a bug in decoding 4bbp png pictures (the last one in this PR), decoder just crashes, needs to be fixed separately


JimBobSquarePants:
Thanks for this! Good to see and really appreciate it. 

Do you have any benchmarks for before and after?

I'll get a fix for the encoder asap. Will have to roll back to a working version and incrementally make changes to figure out what is wrong.

The 4bpp fix should be easy enough I'm sure.

boguscoder:
Oh yeah, I'll provide benchmarks here soon

boguscoder:
I am actually not happy with perf results, let me dig a bit deeper

boguscoder:
Ok, Im seeing really nano improvements here:

Before:

                    Method |      Median |    StdDev | Scaled | Scaled-SD |
-------------------------- |------------ |---------- |------- |---------- |
    None-filtered PNG file |  74.1752 us | 1.4026 us |   1.00 |      0.00 |
     Sub-filtered PNG file |  59.0522 us | 1.3691 us |   0.79 |      0.02 |
      Up-filtered PNG file |  68.8289 us | 2.6562 us |   0.93 |      0.04 |
 Average-filtered PNG file |  91.3585 us | 1.1904 us |   1.23 |      0.03 |
   Paeth-filtered PNG file | 108.2761 us | 6.1980 us |   1.44 |      0.09 |

After:

                    Method |      Median |     StdDev | Scaled | Scaled-SD |
-------------------------- |------------ |----------- |------- |---------- |
    None-filtered PNG file |  73.1292 us |  1.9591 us |   1.00 |      0.00 |
     Sub-filtered PNG file |  58.1899 us |  2.8178 us |   0.79 |      0.04 |
      Up-filtered PNG file |  65.5541 us |  3.1199 us |   0.90 |      0.05 |
 Average-filtered PNG file |  58.2450 us |  3.2539 us |   0.81 |      0.05 |
   Paeth-filtered PNG file | 106.6377 us | 17.3648 us |   1.33 |      0.23 |


Note that your changes added to 'after' case and Im not sure that adding 'fixed' approach to filters really made it faster, I think I got some intermediate test results that showed slow down

dlemstra:
There should be less allocations though because you are using the input array. Still think it should be good to merge your changes.

boguscoder:
Even if I perform tests on the very same file I get rather different results, Im not sure if its because its that fast and timing it is hard or some other reason, so I would not trust numbers above a lot, but in general my changes should make sense..

JimBobSquarePants:
@boguscoder The improvements were more about memory usage. I'll merge this and run some tests locally regarding the fixed keyword usage.  I'll have a look then at fixing the  4bbp png issue now that I have fixed the encoder.

JimBobSquarePants:
@boguscoder  Here's the current state of play with the decoder with your changes and a couple of additional tweaks.

```
Host Process Environment Information:
BenchmarkDotNet-Dev.Core=v0.9.9.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-6700HQ CPU 2.60GHz, ProcessorCount=8
Frequency=2531249 ticks, Resolution=395.0619 ns, Timer=TSC
CLR=MS.NET 4.0.30319.42000, Arch=64-bit RELEASE [RyuJIT]
GC=Concurrent Workstation
JitModules=clrjit-v4.6.1586.0

Type=DecodePng  Mode=Throughput
```
             Method |     Median |    StdDev | Scaled | Scaled-SD | Gen 0 | Gen 1 |  Gen 2 | Bytes Allocated/Op |
------------------- |----------- |---------- |------- |---------- |------ |------ |------- |------------------- |
 System.Drawing Png |  3.1796 ms | 0.1052 ms |   1.00 |      0.00 |     - |     - |  47.10 |          90,395.43 |
     ImageSharp Png | 11.7749 ms | 0.4999 ms |   3.72 |      0.19 | 69.00 |  1.00 | 178.00 |         503,921.20 |

That's half the memory we were using before so a nice improvement. 

The obvious memory hog right now is the allocation of memory when assigning chunk data. Any additional performance improvements I'm not sure about.

PR #30, opened 2016-12-01
fixes #28

### LINKED ISSUE #28: Png's are decoding wrong after 1.0.0-alpha-000061
png's are decoded incorrectly on version v1.0.0-alpha-000062 and higher (including 95)

last know good version is v1.0.0-alpha-000061
so i think something went wrong in this commit https://github.com/JimBobSquarePants/ImageSharp/commit/27522da82b93d19c530e848b55ef8e9ad4e46d1c

i think it's the same as #24 but this time it's in the **de**coder, i only have 1 specific image that gives me the issue. all of my other images decode&encode just fine (on version 95)

tested with output format: png,jpeg,bmp,gif (all give the same issue)
used the same simple console app as in https://github.com/JimBobSquarePants/ImageSharp/issues/24#issuecomment-260329863

please don't forget to sleep this time 😄 

original source file:
source-file

single-thread output:
single-thread

multi-thread output:
bad-file

KLuuKer:
fixes #28 

- do not loop trough the entire rented array but only the bytesPerScanline.

- zero out the previousScanline because it might not have zero values all the way trough the array

### DIFF
--- a/src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
+++ b/src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
@@ -19,14 +19,15 @@ namespace ImageSharp.Formats
         /// </summary>
         /// <param name="scanline">The scanline to decode</param>
         /// <param name="previousScanline">The previous scanline.</param>
+        /// <param name="bytesPerScanline">The number of bytes per scanline</param>
         /// <param name="bytesPerPixel">The bytes per pixel.</param>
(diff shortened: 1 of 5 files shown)


KLuuKer:
i added that array.clear because some random perf test i googled says it faster for array's longer then 100
and because jimbob told me to 😄

JimBobSquarePants:
Cheers for that. Merged 💯

PR #33, opened 2016-12-07
Added Argb IPackedPixel

rokleM:
Adds support for ARGB packed pixels.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*.cs]
+indent_style = space
+indent_size = 4
--- a/src/ImageSharp/Colors/PackedPixel/Argb.cs
+++ b/src/ImageSharp/Colors/PackedPixel/Argb.cs
@@ -11,16 +11,16 @@ namespace ImageSharp
     /// This struct is fully mutable. This is done (against the guidelines) for the sake of performance,
     /// as it avoids the need to create new values for modification operations.
     /// </remarks>
-    public class Argb : IPackedPixel<uint>, IEquatable<Argb>
+    public struct Argb : IPackedPixel<uint>, IEquatable<Argb>
     {
-        const int  BlueShift  = 0;
-        const uint BlueMask   = 0xFFFFFF00;
-        const int  GreenShift = 8;
-        const uint GreenMask  = 0xFFFF00FF;
-        const int  RedShift   = 16;
-        const uint RedMask    = 0xFF00FFFF;
-        const int  AlphaShift = 24;
-        const uint AlphaMask  = 0x00FFFFFF;
+        const int BlueShift = 0;
+        const uint BlueMask = 0xFFFFFF00;
+        const int GreenShift = 8;
+        const uint GreenMask = 0xFFFF00FF;
+        const int RedShift = 16;
+        const uint RedMask = 0xFF00FFFF;
+        const int AlphaShift = 24;
+        const uint AlphaMask = 0x00FFFFFF;
 
         /// <summary>
         /// The maximum byte value.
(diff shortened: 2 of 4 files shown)


.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*.cs]
```
JimBobSquarePants: This is a good idea.

tests/ImageSharp.Tests/Colors/PackedPixelTests.cs
```diff
@@ -57,6 +57,51 @@ public void Alpha8()
         }
 
         [Fact]
+        public void Argb()
```
JimBobSquarePants: Great!

src/ImageSharp/Colors/PackedPixel/Argb.cs
```diff
@@ -0,0 +1,309 @@
...
+        /// Packs the four floats into a <see cref="uint"/>.
+        /// </summary>
+        /// <param name="x">The x-component</param>
+        /// <param name="y">The y-component</param>
+        /// <param name="z">The z-component</param>
+        /// <param name="w">The w-component</param>
+        /// <returns>The <see cref="uint"/></returns>
+        static uint Pack(float x, float y, float z, float w)
+        {
+            var value = new Vector4(x, y, z, w);
```
JimBobSquarePants: I don't use `var` myself but I can clean this up after the merging the PR so don't worry about it.

JimBobSquarePants:
Excellent! Thanks for this 💯

src/ImageSharp/Colors/PackedPixel/Argb.cs
rokleM: Oh, sorry, didn't see enough of the rest of the code to not use it (I do use var)!

rokleM:
Thank you for making the library happen! 😄

src/ImageSharp/Colors/PackedPixel/Argb.cs
olivif: why not using `var`? 😄 The rule I use these days is: use `var` by default, unless the type cannot be easily inferred from reading the code. It makes the code more maintainable - for instance if I change from an IEnumerable to IList, etc, I don't need to update all the places, it kinda just works.

JimBobSquarePants: @olivif  What is the resultant type of `x` in this bit of code?

```
var x = (byte)1 + (byte)2;
```

I'm sure you know this already but it's an `int` due to implicit casting since bytes don't have an `+` operator. (This is the simplest example I could think of as an easy mistake to make using `var`)

I work with streams a lot and I've lost count of the amount of times that the wrong value has been passed to a `BinaryWriter` due to the use of `var`. That's always a nightmare to debug I know it's more common to use now but in certain critical codebases like the Roslyn compiler it's still banned. 

I honestly believe that it's having a negative effect on modern programming. Devs aren't learning their types nor the framework properly as a result and are being careless, using `var` mostly because they don't know what the return type of a method should be.

olivif: @JimBobSquarePants this is super interesting, I've actually never run into these type problems due to `var`, thanks for sharing 😄 . I understand the perils now, up until now it was just the maintainability argument for me, but now I see how it could go wrong. 

> I honestly believe that it's having a negative effect on modern programming. Devs aren't learning their types nor the framework properly as a result and are being careless, using var mostly because they don't know what the return type of a method should be.

I fully agree here. I was fortunate enough to write a lot of C++ before coming to C# but I realize that might not be the norm these days.

JimBobSquarePants: You're better than I then! I tried C++ once..... Nope, too hard! 😄

olivif: Haha! Well, it was great at the time and definitely learnt a lot. Would I go back to unmanaged without a really good reason? Probably not 😄

Issue #5, opened 2016-10-30
Brushes


JimBobSquarePants:
_From @JimBobSquarePants on December 1, 2015 12:48_

This issue wraps several brush types.

I think we should implement at least:
- [ ] Path brush
- [ ] Pattern brush (Not essential but nice to have)
- [ ] Elliptical brush (Not essential but nice to have)
- [ ] Gradient brush (Not essential but nice to have)

I'm envisioning something like :

```
PathBrush brush = new PathBrush
{
    Color = Color.Black,
    Thickness = 10, // the thickness to draw a line
    Path = vectors // A list of vectors to follow
}

Image.Draw(brush)
```

The method would draw a Bezier curve following the given path, color, and thickness.

_Copied from original issue: JimBobSquarePants/ImageProcessor#264_

JimBobSquarePants:
_From @dampee on December 21, 2015 12:57_

If I get through my color space, I'll see what I can do for January.  Any startingpoints, hints, ... ?

JimBobSquarePants:
That'd be great if you could!

Honestly.... Not got much of a clue. I think the source for Cairo might be a place to start looking around as I think a few libraries wrap around it. I'm trying to get my head around this just now.

https://github.com/mono/sysdrawing-coregraphics/blob/master/System.Drawing.Drawing2D/PathGradientBrush.cs

JimBobSquarePants:
_From @mweber26 on March 16, 2016 12:49_

I don't understand this issue.  What are the brushes for?  There doesn't appear to be any drawing code to use them.

JimBobSquarePants:
I've expanded on the issue. We'd obviously have to create bezier classes, a draw method, etc 

I found a good primer on Beziers here. http://pomax.github.io/bezierinfo/

JimBobSquarePants:
_From @cartman300 on May 25, 2016 0:34_

Is there any progress on this? I have quite a good idea how to implement these. Might give it a go.

JimBobSquarePants:
@cartman300 None whatsoever. 

I'm not the greatest mathemagician in the world so I struggle with this kind of stuff so If you know how to do it please give it a go. I would really appreciate it!

I'm essentially looking at first implementing path drawing. I think we should be able to break bezier and quadratic curves into neat methods and take advantage of the new vector types for things like Vector.Dot etc.  Being able to set line thickness would very useful also. 

Cheers!

JimBobSquarePants:
[...]

PR #37, opened 2016-12-14
Add drawing methods.

tocsoft:
Adds a 4 `IPath`s;
 - `Path` : a generic open path described by an arbitrary collection of `ILineSegments` which are closed.
 - `Polygon` : a generic closed path described by an arbitrary collection of `ILineSegments` which are closed.

Adds 2 `ILineSegment`s;
 - `LinearLineSegment` : a segment made up of control points representing straight lines.
 - `BezierLineSegment` : a path made up of control points describing a bezier path.

Adds 4 `IShape`s
 - `Polygon` : a generic closed path described by an arbitrary collection of `ILineSegments` which are closed.
 - `LinearPolygon` : a shape made up of control points representing straight lines.
 - `BezierPolygon` : a shape made up of control points describing a set of bezier curves which are closed by a straight line
 - `ComplexPolygon` : a polygon described by one or more `IShape`s as either outlines or holes.

All `IShape`s can be presented as a collection of `IPath`.

Adds 2 `IBrush`s
 - `SolidBrush` : described by a `TColor`
 - `Patternbrush` : described by a foreground and background `TColor` and a pattern.

Adds 1 `IPen`:
 - `Pen` : described by an `IBrush` a width/thickness to draw the line and an optional pattern.

Adds 3 Image Processors
 - `FillProcessor` : flood fills the image with an `IBrush`.
 - `FillShapeProcessor` : flood fills inside an `IShape`.
 - `DrawPathProcessor` : draws a collection of `IPath`s or and `IShape`s onto the image using an `IPen` to determine `TColor`.

Adds a multitude of Image Extensions for accessing the above processors in a manor similar to `System.Drawing`


tocsoft:
Fixes #5 - mostly.

We'll want to spin out the Elliptical brush, and Gradient brush into there own issues that should be quite easy to implement once this PR lands, I think they would would make nice first timer PRs.

tocsoft:
Now off to look at [SharpFont](https://github.com/MikePopoloski/SharpFont) 😁 and see If I can get that code ported in and working with the new drawing stuff.

JimBobSquarePants:
@tocsoft You're a coding machine!!!

I'll do a quick review of the code to make sure everything fits. I have a couple of questions about the partial usage and IDisposable so I'll add them in the review. 

I'd also like to ask your thoughts on the API itself.

How did you find it to work with and extend? Did it feel natural and straightforward or are there things you would change if you could?

How is the current performance of the drawing algorithms. Were there API limitations that you feel had a negative impact on performance?

prepare:
@JimBobSquarePants 
I want to introduce another font lib 
that is https://github.com/LayoutFarm/Typography

:)

antonfirsov:
Had a quick look at the implementation. Unfortunately there are a few points that will definitely effect the performance:
- `bool[,]`
[Multidimensional arrays are slow](http://stackoverflow.com/questions/468832/why-are-multi-dimensional-arrays-in-net-slower-than-normal-arrays) I suggest to flatten them
- `IEnumerable<T>` and Linq with value types
Even frequent .ToArray() calls are better, then using `IEnumerable<T> and Linq with "stream-like" data. I suggest using simple arrays where possible, even if you break encapsulation a little bit on API-s like `ILineSegment.AsSimpleLinearPath()` . Later we can use [Span](https://github.com/dotnet/corefxlab/blob/master/docs/specs/span.md)
- `List<T>`
When the length of the result is known, or for members like `BezierLineSegment.linePoints` it's better to use arrays! (And if AsSimpleLinearPath()`  returns an array, you can return the array directly with without torturing the GC with Enumerator instance creation.

src/ImageSharp/Drawing/Processors/DrawPathProcessor.cs
```diff
@@ -0,0 +1,160 @@
...
+                    this.ParallelOptions,
+                    y =>
+                    {
+                        int offsetY = y - polyStartY;
+
+                        for (int x = minX; x < maxX; x++)
+                        {
+                            int offsetX = x - startX;
+
+                            var dist = this.paths.Select(p => p.Distance(offsetX, offsetY)).OrderBy(p => p.DistanceFromPath).First();
```
JimBobSquarePants: This line is an absolute performance killer. For each pixel we are creating two arrays. One for Select, one for OrderBy. We need to find a way to avoid this or the memory usage will be incredible.

JimBobSquarePants:
@prepare I've been following your work. 👍 I'll open a specific issue for font rendering so we can chat.

@antonfirsov Yeah, I saw pretty much the same and identified a pretty big issue on a hot path too, see my comments above.

@tocsoft There's a few things here I'd like to change. 

MultiDimensional arrays. Either use a linear or jagged array - both are way faster.

IDisposable. There doesn't seem to be any requirement for it and I swear I saw static instances of IDisposable objects in the code somewhere. 

Point, PointF. They're not adding anything to the API and require enumerable allocations every time they are used. I say scrap em.

Linq. Linq is useful but a performance killer, especially on hot paths like in DrawPathProcessor We need to find alternatives to that.

JimBobSquarePants:
#38 for font support.

JimBobSquarePants:
Replacing that Linq statement in `DrawPathProcessor` for a simple loop. 

```
var dist = this.paths.Select(p => p.Distance(offsetX, offsetY)).OrderBy(p => p.DistanceFromPath).First();
```

```
var dist = Closest(offsetX, offsetY);
private PointInfo Closest(int x, int y)
{
    PointInfo result = default(PointInfo);
    float distance = float.MaxValue;

    for (int i = 0; i < this.paths.Length; i++)
    {
        var p = this.paths[i].Distance(x, y);
        if (p.DistanceFromPath < distance)
        {
            distance = p.DistanceFromPath;
            result = p;
        }
    }

    return result;
}
```

**Before**

Method |       Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev |      Gen 0 |     Gen 1 |    Gen 2 | Allocated |
---------------------------- |----------- |---------- |---------- |------- |-------------- |----------- |---------- |--------- |---------- |
'System.Drawing Draw Lines' |  3.5471 ms | 0.0681 ms | 0.2549 ms |   1.00 |          0.00 |  1246.8750 |  434.3750 | 434.3750 |   5.15 MB |
'ImageSharp Draw Lines' | 35.8006 ms | 0.2836 ms | 1.0983 ms |  10.14 |          0.70 | 41125.0000 | 1269.5313 | 437.5000 | 135.68 MB |

**After**

Method |       Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev |     Gen 0 |    Gen 1 |    Gen 2 | Allocated |
---------------------------- |----------- |---------- |---------- |------- |-------------- |---------- |--------- |--------- |---------- |
'System.Drawing Draw Lines' |  3.5000 ms | 0.0356 ms | 0.1380 ms |   1.00 |          0.00 | 1115.6250 | 303.1250 | 303.1250 |   5.15 MB |
'ImageSharp Draw Lines' | 10.4963 ms | 0.1033 ms | 0.4258 ms |   3.00 |          0.16 |  929.8986 | 868.2432 | 867.3986 |   7.66 MB |

JimBobSquarePants:
@tocsoft meant to ask as I didn't see. Is there antialiasing options? I didn't see any. I'd like to be able to optionally turn antialiasing off if need be.

src/ImageSharp/Drawing/Brushes/SolidBrush.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// <returns>
+        /// The brush applicator for this brush
+        /// </returns>
+        /// <remarks>
+        /// The <paramref name="region" /> when being applied to things like shapes would ussually be the
+        /// bounding box of the shape not necessarily the bounds of the whole image
+        /// </remarks>
+        public IBrushApplicator<TColor, TPacked> CreateApplicator(RectangleF region)
+        {
+            return new SolidBrushApplicator(this.color);
```
tocsoft: They are disposed 
https://github.com/JimBobSquarePants/ImageSharp/pull/37/files#diff-51290ecbe2b99dd904013148fc22e4f4R66
here
https://github.com/JimBobSquarePants/ImageSharp/pull/37/files#diff-755b67a7f6c80ae11495b9d27b4d5935R76
here 
https://github.com/JimBobSquarePants/ImageSharp/pull/37/files#diff-94dfa090254876c958b23d72f7c0d460R193
and 
https://github.com/JimBobSquarePants/ImageSharp/pull/37/files#diff-94dfa090254876c958b23d72f7c0d460R246

src/ImageSharp/Drawing/Fill.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// <param name="source">The source.</param>
+        /// <param name="color">The color.</param>
+        /// <param name="points">The points.</param>
+        /// <returns>The Image</returns>
+        public static Image<TColor, TPacked> FillPolygon<TColor, TPacked>(this Image<TColor, TPacked> source, TColor color, PointF[] points)
+           where TColor : struct, IPackedPixel<TPacked>
+           where TPacked : struct
+        {
+            // using Polygon directly instead of LinearPolygon as its will have less indirection
+            return source.Fill(new SolidBrush<TColor, TPacked>(color), new Polygon(new LinearLineSegment(points)));
```
tocsoft: Nothing in this method is `IDisposable`

There are only 2 Interfaces which are `IDisposable` thats `IBrushApplicator` which is generated during processing and is disposed of by the Processors or Pen applicators or `IPenApplicator` which again is disposed by the processors.

src/ImageSharp/Drawing/Paths/InternalPath.cs
```diff
@@ -0,0 +1,279 @@
...
+
+                j = i;
+            }
+
+            return oddNodes;
+        }
+
+        private Vector2[] Simplify(IEnumerable<ILineSegment> segments)
+        {
+            return segments.SelectMany(x => x.AsSimpleLinearPath()).ToArray();
```
tocsoft: yep.. couple of loops and a `List<Vector2>` should do it.

src/ImageSharp/Drawing/Paths/LinearLineSegment.cs
```diff
@@ -0,0 +1,83 @@
...
+    public class LinearLineSegment : ILineSegment
+    {
+        private Vector2[] points;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LinearLineSegment"/> class.
+        /// </summary>
+        /// <param name="points">The points.</param>
+        public LinearLineSegment(IEnumerable<PointF> points)
+            : this(points?.Select(x => x.ToVector2()).ToArray())
```
tocsoft: these only happen at Contructor time and thus a one time hit, I think I had envisioned Shapes and this Paths and LineSegments being relatively long lived items and you wouldn't new a new one up each time if you wanted to apply it to many things, as the overhead required on things like the ComplexPolygon would be very high if newing them up all the time.

global.json
```diff
@@ -1,3 +1,6 @@
 {
-  "projects": [ "src" ]
+  "projects": [ "src" ],
+    "sdk": {
+    "version": "1.0.0-preview2-003121"
```
tocsoft: I couldn't even compile without it when I first started looking at ImageSharp... I'll see if it's still needed.

tocsoft:
RE: antialiasing

I can now see why `System.Drawing` has a `Graphics` class for handling some of that stuff for you now I've been doing this work. By letting you wrap the calls down to the image with an object that can be configured with almost global(for that context) options without having to add drawing specific things to an `Image`.

What to people think?

Should I create a `Graphics<,>` class wrapping `IImageBase<,>` which will manage antialiasing and maybe in the future things like composition modes? or should I create a standard drawing configuration object that will need being passed into each call to `DrawX(..)`/`FillX(...)`? or should simple add a simple bool into all  `DrawX(..)`/`FillX(...)` to toggle on antialiasing or on or off?

tocsoft:
> bool[,]
>
> Multidimensional arrays are slow I suggest to flatten them into a 1-dimensional array.

I'll flatten them inside the `PatternBrush` so the constructor will still take in the multi dimensional array as thats by far the easiest cleanest way to define pattern, with a flat array in the constructor you would also need to provide a stride width and i feel thats not going to be intuitive for most users, or use a jaggerd array but that you can't easily enforce that all the rows are the same width, we let the compiler ensure the array is initialized properly then not the runtime.

> IEnumerable<T> and Linq with value types
> 
> Even frequent & redundant .ToArray() calls are better, than using IEnumerable<T> and Linq with "stream-like" data. I suggest using simple arrays where possible, even if you break the encapsulation rules a little bit on A
>
>PI-s like ILineSegment.AsSimpleLinearPath() . Later we can use Span

Yeah i'll move those over to `Vector2[]` with the idea that they will eventually become `ReadonlySpan<Vector2>` once we have access.

> List<T>
> 
> When the length of the result is known, or for members like BezierLineSegment.linePoints it's better to use arrays! (And if AsSimpleLinearPath() returns an array, you can return the array directly without torturing the GC with Enumerator instance creation.)

Totally, will fix that too.

global.json
JimBobSquarePants: Kewl. I have no idea what version I have installed but I think it's the latest.

src/ImageSharp/Drawing/Fill.cs
JimBobSquarePants: Ok coke... Do we need them to be IDisposable? I don't mind the pattern as long as it's truly necessary. They're a great way to handle array pooling for example.

src/ImageSharp/Drawing/Brushes/SolidBrush.cs
JimBobSquarePants: Github wouldn't let me follow those links for some reason but I'll take your word for it 😉

src/ImageSharp/Drawing/Paths/InternalPath.cs
JimBobSquarePants: Sweet.

src/ImageSharp/Drawing/Brushes/SolidBrush.cs
tocsoft: basically there are using statements in the 3 new XProcessor classes and in the Pen applicators in Pen.cs

src/ImageSharp/Drawing/Paths/LinearLineSegment.cs
tocsoft: no problem, i'll drop the usage of them

JimBobSquarePants:
You survived the first round of code bashing! 🎉 

I always feel like I'm doing something horrible doing code reviews. It's all too easy to come across as too picky or simply miss intent. 

I think with a little bit of work we can actually out-perform System.Drawing which is simply brilliant so it shows you're definitely well on the right track. 

One other thing I was thinking perf wise was passing params to methods. That allocates an array which we have no control over. I was think that if we passed arrays instead the end user could pool them if need be which would give them greater power over performance. What do you think?

Re: antialias, I would prefer more granular control if possible. I see why Graphics was created but I'm not a fan.

tocsoft:
> One other thing I was thinking perf wise was passing params to methods. That allocates an array which we have no control over. I was think that if we passed arrays instead the end user could pool them if need be which would give them greater power over performance. What do you think?

You can still pass in an array when using params, just save those people who don't care about the array to not have to deal with them but allows those that do care to care.

antonfirsov:
@tocsoft it's ok, if we can suppose that the brushes are cached most of the time and not createted on the fly.
However, all of your patterns seem to be 4x4 in size. Isn't it possible to cache the FG/BG/Etc colors into a Matrix4x4 and batch the `OnApply()` -> `IBrushApplicator.GetColor()` -> `IBrush.GetColor()` calls to work on multiple pixels per call?
The difference in execution time will be **HUGE**.
Or is this an optimization for another time? @JimBobSquarePants what do you think?

tocsoft:
Well did some benchmarks on the different ways I access the arrays used by patterns Pattern and it looks like jagged array wins.

<table>
<thead><tr><th>                               Method</th><th>Mean</th><th>StdDev</th><th>Scaled</th><th>Scaled-StdDev</th><th>Gen 0</th><th>Gen 1</th><th>Gen 2</th>
</tr>
</thead><tbody><tr><td>       'ImageSharp Fill with Pattern'</td><td>42.3406 ms</td><td>1.2281 ms</td><td>1.00</td><td>   0.00</td><td>437.5000</td><td>437.5000</td><td>437.5000</td>
</tr><tr><td>'ImageSharp Fill with Pattern Flat Array'</td><td>41.5260 ms</td><td>0.7275 ms</td><td>0.98</td><td>   0.03</td><td>470.8333</td><td>470.8333</td><td>470.8333</td>
</tr><tr><td>'ImageSharp Fill with Pattern Jagged Array'</td><td>40.6324 ms</td><td>0.6159 ms</td><td>0.96</td><td>   0.03</td><td>370.8333</td><td>370.8333</td><td>370.8333</td>
</tr></tbody></table>

antonfirsov:
I'm skeptical about the memory part. Jagged arrays do RowCount+1 allocations. Are you sure there are no other differences?
It's also very hard to pool jagged arrays.

JimBobSquarePants:
>You can still pass in an array when using params, just save those people who don't care about the array to not have to deal with them but allows those that do care to care.

@tocsoft Oh yeah! totally forgot that. Ignore me then.

There's not much in the speed difference in those array benchmarks but definitely a difference in allocations. Do you think the lack of length enforcement is a major issue? I use jagged arrays throughout the convolution classes.

@antonfirsov I think I'd have to see a gist to understand you properly. I think I get the principle though so if we can, we should do that now.

antonfirsov:
@JimBobSquarePants I will create a gist tonight (morning in Sydney)

tocsoft:
Added a simple `ImageBrush` so you can now fill shapes/paths with Colors sourced from an image, making use of the fact that `IBrushApplication`/`IPenApplicator` are disposable.

@antonfirsov not all patterns need be 4x4 i've just updated some so they are different shapes as they didn't need to be bigger.

@JimBobSquarePants they way you've used  jagged arrays in the convolution classes looks fine the issue I have with them here is they are exposed via a public api where end users are expected the provide them for making custom pattern brushes

src/ImageSharp/Drawing/Fill.cs
tocsoft: see `ImageBrush` that alone means they need to be disposable

src/ImageSharp/Drawing/Brushes/PatternBrush.cs
```diff
@@ -14,6 +14,32 @@ namespace ImageSharp.Drawing.Brushes
     /// <summary>
     /// Provides an implementaion of a pattern brush for painting patterns.
     /// </summary>
+    /// <remarks>
```
JimBobSquarePants: Excellent! Thanks for that.

src/ImageSharp/Drawing/Brushes/ImageBrush.cs
```diff
@@ -0,0 +1,108 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Numerics;
+    using System.Threading.Tasks;
+    using Processors;
+
+    /// <summary>
+    /// Provides an implementaion of a solid brush for painting with repeating images.
+    /// </summary>
+    public class ImageBrush : ImageBrush<Color, uint>
```
JimBobSquarePants: Eureeka! Ok, I totally get it now. Sorry for the confusion.

src/ImageSharp/Drawing/Processors/DrawPathProcessor.cs
```diff
@@ -112,15 +127,17 @@ protected override void OnApply(ImageBase<TColor, TPacked> source, Rectangle sou
...
-
+                        var currentPoint = default(Vector2);
                         for (int x = minX; x < maxX; x++)
                         {
                             int offsetX = x - startX;
+                            currentPoint.X = offsetX;
+                            currentPoint.Y = offsetY;
 
-                            var dist = Closest(offsetX, offsetY);
+                            var dist = Closest(currentPoint);
```
JimBobSquarePants: This and the other Vector2 stuff is great!

src/ImageSharp/Drawing/Brushes/ImageBrush.cs
tocsoft: no problem, I figured actually just adding something that needed it would be the best explanation  😄

JimBobSquarePants:
Just had a quick look at the updates before I go to bed. It's looking much better now. 

I'll run the benchmarks in the morning but I reckon they must be much closer than before. If you're happy then I'll merge the PR then and we can continue with any improvements after. 

Thanks again for your efforts. It's been brilliant!

tocsoft:
I think I'm happy, as you said If I find anything that needs changing then I can always just make another PR.

tocsoft:
@antonfirsov 
> However, all of your patterns seem to be 4x4 in size. Isn't it possible to cache the FG/BG/Etc colors into a Matrix4x4 ColorBuffer4x4 structure and batch the OnApply() -> IBrushApplicator.GetColor() -> IBrush.GetColor() calls to work on multiple pixels per call?
> The difference in execution time will be HUGE.

I did early on have a version of `IBrushApplicator` that could return multiple pixels at once but i figured that the added complexity wasn't worth it in the initial drop. But I can totally see it being an optimisation. 

How is see it being done by creating `int IBrushApplicator.GetColors(TColor[] outputBuffer, Vector2 start, Vector2 end)`. that would populate the colors wrapping at each line, and returning  the number of pixels found.

antonfirsov:
@tocsoft @JimBobSquarePants I have some really crazy optimization ideas coming into my mind, can't wait to share that gist :)
I also think, that for now this PR is good as is, optimization can be added later.

tocsoft:
looking forward to it 😁

antonfirsov:
Ok guys, I will blow your mind another time, today I'd be happy enough if I could just finish my own PR & go to bed :D

JimBobSquarePants:
Good luck! Everything is merged no. Good times!

PR #42, opened 2016-12-17
Custom xUnit utilities to to support covering more cases with less code

antonfirsov:
- Core class:
`TestImageFactory`
- Support for testing multiple pixel formats with a single theory using the [PixelTypes](https://github.com/antonfirsov/ImageSharp/blob/master/tests/ImageSharp.Tests/TestUtilities/PixelTypes.cs) flag enum, which triggers `<TColor, TPacked>` substitution
- Support for taking test images from multiple sources (File, blank, generated etc.) with a custom `DataAttribute` specializations (`[WithBlankImage]`, `[WithFile]`, `[WithFileCollection]`, `[WithMemberFactory]` .. etc)

- With this PR usage adopted to: [PixelAccessorTests](https://github.com/antonfirsov/ImageSharp/blob/master/tests/ImageSharp.Tests/Image/PixelAccessorTests.cs), `JpegTests`

Sample application:
```
public static IEnumerable<string> AllJpegFiles => TestImages.Jpeg.All;

[Theory]
[WithFileCollection(nameof(AllJpegFiles), PixelTypes.All)]
public void OpenJpeg_SaveBmp<TColor, TPacked>(TestImageFactory<TColor, TPacked> factory)
    where TColor : struct, IPackedPixel<TPacked> where TPacked : struct
{
    var image = factory.Create();

    factory.Utility.SaveTestOutputFile(image, "bmp");
}
```
This creates ~300 output files.

More examples in [TestImageFactoryTests](https://github.com/antonfirsov/ImageSharp/blob/master/tests/ImageSharp.Tests/TestUtilities/TestImageFactoryTests.cs).


olivif:
@antonfirsov have you noticed any impact on the duration of the test suite after this change? I saw some reflection so I'm wondering if that is slowing it down at all 😄

antonfirsov:
@olivif @JimBobSquarePants I can improve `TestImageFactory` to cache already loaded / created images, so in the end the total execution time of the **current** unit tests can be decreased.
I think the effect of reflection is negligible in the forest of gigapixels :)

olivif:
Haha loved the giga pixels comment! I agree, I think caching the already loaded images is a great start. Don't really want to highjack your PR though so feel free to make that change in another PR or write up a comment on one of the issues with some details so someone else can pick it up :smile:

Also, on the main focus of the PR, I think it's great! This will give us the ability to have more tests with less duplication.

antonfirsov:
I'm closing this in favor of #57!

PR #43, opened 2016-12-17
Adding missing documentation and fixing style issues.

olivif:
I noticed the project had 124 stylecop warnings, so I fixed a few (down to 70 after this PR). Would be great to fix all and set warnings as errors on the project to get build time protection on this.

### DIFF
--- a/src/ImageSharp/Colors/PackedPixel/IPackedPixel.cs
+++ b/src/ImageSharp/Colors/PackedPixel/IPackedPixel.cs
@@ -6,6 +6,7 @@
 namespace ImageSharp
 {
     using System;
+
     /// <summary>
     /// An interface that represents a packed pixel type.
     /// </summary>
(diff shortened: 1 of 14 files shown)


dlemstra:
Thanks for helping us out with fixing those warnings. I noticed that some of your params/returns don't start with a capital letter. I know it is a bit OCD but could you change that?

olivif:
thanks @dlemstra. fixed the uppercase issue now. I also looked into if this can be a stylecop rule, so we can automate it, unfortunately couldn't find anything 😞

dlemstra:
There is a StyleCop rule for this but it has not yet been implemented. I am waiting on my pull request (https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/2242) being accepted and then can we enable the rule in this project.

olivif:
@dlemstra awesome!

JimBobSquarePants:
Ah this is great! I'm trying to make the code as readable as possible so this really helps. 👍

olivif:
Thanks for merging! Yes, I have some more ideas on how to improve it, just wanted to start with a small-ish PR to kick things off :smile:

JimBobSquarePants:
That's great to hear! Thanks for helping out 😄

Issue #48, opened 2016-12-19
Fix StyleCop/compiler warnings 


olivif:
Hey everyone! I've been working on getting rid of our build warnings - stylecop and compiler. 

There are 3 categories of warnings left, and I need your help on deciding how to fix those 😄 

**1. SA1649 - File name must match first type name (4)**
This is happening for all classes where we have a generic version and a regular version which share the same name - Brushes, ImageBrush, etc. and we use the convention of ```{name}`{numberOfTypes}.cs``` for the file name. 
The options I see are:
* keep the convention and ignore the warning for this specific case
* get rid of the convention and rename the generic and non generic versions 

I would personally like a better distinction between the T and non T versions, and that would also help with the readability of the code. 

**2. SA1401 - Field must be private (4)**

This is happening in `JpegDecoderCore` only, where we have some internal fields, like below.

```cs
        /// <summary>
        /// The byte buffer.
        /// </summary>
        internal Bytes Bytes;
```
This is a maintainability [rule](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1401.md) - the best practice is to keep all fields to private, and if anything needs to be exposed it should be done via properties. 

Should we change these fields to properties? I'm not very familiar with this code so I'm hoping you will have more reasons for one option or the other. 

**3. CS0219 - The variable <> is assigned but its value is never used (4)**

This is in the test project in `ReferenceImplementations`. r0 and r4 variables are never used. The fix to remove them is trivial, but I was wondering if them not being used is actually intentional.

tocsoft:
I would personally just rename the generic files (and header) to the style StyleCop wants rather than fighting with it.
I.E. Just rename ``Brushes`2.cs`` to` Brushes{TColor,TPacked}.cs`.

JimBobSquarePants:
@olivif Nice work!

>1. SA1649 - File name must match first type name (4)

I'm up for renaming to the explicit generic version. I just used the current naming on a whim as I was refactoring the brushes. 

>2. SA1401 - Field must be private (4)

If it's not references anywhere outside the class make it a private field, otherwise a property.

>3. CS0219 - The variable <> is assigned but its value is never used (4)

If they're not used it's an oversight and the variables should never be assigned.

Of course run the tests to make sure nothing breaks 😸

olivif:
thanks @JimBobSquarePants! 😄 makes sense! will see what I can do.

PR #49, opened 2016-12-19
Adding missing documentation and fixing style issues - part 2

olivif:
Alright, down to 12 warnings 🎉!!! Added more summaries and docs. Ignored one rule where I think it actually makes more sense to keep the code as is. 

I need some input on the remaining 12 warnings and how to best fix them, I filed an issue so we can discuss the approach there - #48.

### DIFF
--- a/src/ImageSharp/Bootstrapper.cs
+++ b/src/ImageSharp/Bootstrapper.cs
@@ -19,15 +19,24 @@ namespace ImageSharp
     public class Bootstrapper
     {
         /// <summary>
-        /// A new instance Initializes a new instance of the <see cref="Bootstrapper"/> class.
-        /// with lazy initialization.
+        /// A singleton of the <see cref="Bootstrapper"/> class.
         /// </summary>
-        private static readonly Lazy<Bootstrapper> Lazy = new Lazy<Bootstrapper>(() => new Bootstrapper());
+        private static readonly Bootstrapper Instance = new Bootstrapper();
 
         /// <summary>
-        /// The default list of supported <see cref="IImageFormat"/>
+        /// The list of supported <see cref="IImageFormat"/>.
         /// </summary>
         private readonly List<IImageFormat> imageFormats;
+
+        /// <summary>
+        /// The parallel options for processing tasks in parallel.
+        /// </summary>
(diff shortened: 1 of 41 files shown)


dlemstra:
Can you rebase this pull request? I made some changes to Bootstrapper.cs that broke this.

olivif:
thanks @dlemstra, just rebased and fixed all the remaining warnings. 0 warnings 🎈

olivif:
OK I found the bug and fixed it, it was super interesting actually. 

Here's what I did, I changed the `Bytes Bytes` field in the `JpegDecoderCore` from an internal field to a private backing field with a get only property. However, because Bytes is a mutable struct, this line would no longer work as intended, since we were changing the Bytes struct on a copy really. 

`byte c = decoder.Bytes.ReadByteStuffedByte(decoder.InputStream, out errorCode);`

So the fix was to make sure we reflect the changes property on the internal bytes. 

```
// Grab the decode bytes, use them and then set them
// back on the decoder.
var decoderBytes = decoder.Bytes;
byte c = decoderBytes.ReadByteStuffedByte(decoder.InputStream, out errorCode);
decoder.Bytes = decoderBytes;
```

dlemstra:
Thanks for the fix 🙌. I pushed a patch to make sure that OpenCover will return the exit code from the test. It should now fail when it does not pass the tests.

olivif:
@dlemstra awesome!

JimBobSquarePants:
Whoop! 🎉

dlemstra:
Great success 👏 All the warnings are gone (after I fixed one that I just added 👎)

olivif:
Thanks for merging guys. Indeed, it was so rewarding to see zero warnings :smile:

We can consider enabling warnings as errors so we can't accidentally introduce some more before it gets out of control. Maybe if we can only do it on appveyor builds? It's a bit excessive and annoying to not be able to build locally because of stylecop.

dlemstra:
I just enabled warnings as errors for the release build. And AppVeyor does a release build so that should prevent us adding extra warnings.

JimBobSquarePants:
And the winner of the "First person to push errors" award goes to me!

PR #51, opened 2016-12-19
Adding unit tests for equality and hash code for IPackedPixel types

olivif:
I added unit tests for:
- Equals
- Not equals
- HashCode equals
- HashCode not equals 

Found some strange behaviour on HashCode when two colours of the same type and seemingly different initialization values (and Vector4 values) seem to have the same hash code. See `NotEqualityDataDifferentObjects`  and lines with `// same hashcode??` comment.

### DIFF
--- a/tests/ImageSharp.Tests/Colors/ColorEqualityTests.cs
+++ b/tests/ImageSharp.Tests/Colors/ColorEqualityTests.cs
@@ -36,9 +36,9 @@ namespace ImageSharp.Tests.Colors
                { new Short4(Vector4.One * 0x7FFF), new Short4(Vector4.One * 0x7FFF) },
            };
 
-        public static readonly TheoryData<object, object> NotEqualityData =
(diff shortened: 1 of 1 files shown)


dlemstra:
It is possible that there is HashCode collision, should not be a problem. I did some minor refactoring of your code and now all values are enabled again.

JimBobSquarePants:
It seems odd that `Vector2.One` and `Vector2.Zero` passed to a constructor would produce the same hashcode though since the hashcodes are based on the packed value. Definitely worth investigating.

dlemstra:
I did some debugging and it turns out that the ulong that is created has the same hashcode.
```
> ulong a = 4323521613979991040;
> ulong b = 0;
> Console.WriteLine(a.GetHashCode())
0
> Console.WriteLine(b.GetHashCode())
0
```

JimBobSquarePants:
Thanks for checking!

PR #52, opened 2016-12-19
Port Clipper.cs into ImageSharp style

### LINKED ISSUE #47: Port/prune code in Clipper.cs into ComplexPolygon.cs
We want to drop the hard dependency on the Clipper.cs source file so we need to port across the code and try to add any optimisations as we go (use `Vector2` instead of `IntPoint` etc) to prevent the need for copying the data structures around.

_food for thought:_ could `LinkedList<Vector2>`s be used instead of `List<IntPoint>` as a fist optimisation?

tocsoft:
All the classes from Clipper.cs have been moved into there own folder/namespace.
Added README.md to preserve the original Attribution.
I've ported the code to use Vector2 instead of its own point classes.
I've removed all the unused clipping options, and polygon styles that are not required and thus removed a lot of branching logic.

Fixes #47 and should improve code coverage just be the fact there is now less code in here 😁


tocsoft:
now that was a slog... don't ask me how it works yet, but it 'should' be easier to read now.

JimBobSquarePants:
Hi @tocsoft Nice work getting this ported! I've merged everything in and the tests all seem to pass. 

Do you have Resharper? I noticed it was warning me about a few unused methods in the Clipper files. Just wanted to check you are seeing the same.

tocsoft:
Nope I don't have a resharper licence. I was using codelens to find unreferenced methods, must have missed a few in the last pass after I pruned down the code a bit more.

JimBobSquarePants:
I can pull them out, it's no biggie. Can barely code without Resharper now. Actually got the license from doing open source.

PR #56, opened 2016-12-22
Added a brush for recoloring a portion/all of an image.

tocsoft:
This is an example of the `RecolorBrush` changing `Color.Yellow` to `Color.HotPink` with a `threashold` of 0.2 being applied to a `Rectangle` 1/2 height full width, centered.

image

The main reason I've added this is as an example for applying things that might traditionally thought of as filters as brushes. And for me to extend the brushes API(before its locked down) to enable this sort of brush.

Using this technique would allow for filters/brushes to be applied to any arbitrary shapes and not just the whole image.


JimBobSquarePants:
@tocsoft 

That's pretty damn cool!

The only thing concerns me about this PR is the introduction of the new `IReadOnlyPixelAccessor<TColor>`. I don't think it is required and we should pass the concrete `PixelAccessor<TColor>`. I'm fairly certain using the interface on a hotpath will slow things down and we are never going to require another implementation of the functionality anyway. (If we test, we test the real thing).

Using the interface also restricts us if we ever need to change the functionality, Changing an interface is major version stuff.

src/ImageSharp/Image/IReadonlyPixelAccessor{TColor}.cs
```diff
@@ -0,0 +1,50 @@
...
+    using System;
+    using System.Diagnostics;
+    using System.Runtime.CompilerServices;
+    using System.Runtime.InteropServices;
+
+    /// <summary>
+    /// Provides per-pixel readonly access to generic <see cref="Image{TColor}"/> pixels.
+    /// </summary>
+    /// <typeparam name="TColor">The pixel format.</typeparam>
+    public interface IReadonlyPixelAccessor<TColor>
```
JimBobSquarePants: I don't think we need this. `PixelAccessor<TColor>` isn't an extension point.

tocsoft: My logic with the interface was I explicitly didn't want the brush to have write access to the pixels, the only other way I can think to do it would be with a very thin struct wrapper around `PixelAccessor` with aggressive inlining hints.

JimBobSquarePants: Och... I wouldn't worry about it. Anyone writing a new brush should know not to do that.

tocsoft:
@JimBobSquarePants I've dropped the `IReadonlyPixelAccessor<TColor>` in favour of `PixelAccessor<TColor>`

PR #57, opened 2016-12-23
xUnit TestUtilities Round2

antonfirsov:
Updated version of #42

- Core class:
`TestImageProvider<TColor>`
- Support for testing multiple pixel formats with a single theory using the [PixelTypes](https://github.com/antonfirsov/ImageSharp/blob/test-utils-fixed/tests/ImageSharp.Tests/TestUtilities/PixelTypes.cs) flag enum, which triggers a `TColor` substitution
- Support for taking test images from multiple sources (File, blank, generated etc.) with a custom `DataAttribute` specializations (`[WithBlankImage]`, `[WithFile]`, `[WithFileCollection]`, `[WithMemberFactory]` .. etc)
- Added several new test cases to [PixelAccessorTests](https://github.com/antonfirsov/ImageSharp/blob/test-utils-fixed/tests/ImageSharp.Tests/Image/PixelAccessorTests.cs) which are using `TestImageProvider<TColor>`
- Updated  [JpegTests](https://github.com/antonfirsov/ImageSharp/blob/test-utils-fixed/tests/ImageSharp.Tests/Formats/Jpg/JpegTests.cs) to use `TestImageProvider<TColor>`

Sample application:
```
public static IEnumerable<string> AllJpegFiles => TestImages.Jpeg.All;

[Theory]
[WithFileCollection(nameof(AllJpegFiles), PixelTypes.Color | PixelTypes.ColorWithDefaultImageClass | PixelTypes.Argb)]
public void OpenJpeg_SaveBmp<TColor>(TestImageProvider<TColor> provider)
    where TColor : struct, IPackedPixel, IEquatable<TColor>
{
    var image = provider.GetImage();

    provider.Utility.SaveTestOutputFile(image, "bmp");
}
```

More examples in [TestImageProviderTests.cs](https://github.com/antonfirsov/ImageSharp/blob/test-utils-fixed/tests/ImageSharp.Tests/TestUtilities/Tests/TestImageProviderTests.cs).

Other:
- Added new constructor to `PixelArea<T>`

### DIFF
--- a/tests/ImageSharp.Tests/FileTestBase.cs
+++ b/tests/ImageSharp.Tests/FileTestBase.cs
@@ -46,6 +46,9 @@ namespace ImageSharp.Tests
             // TestFile.Create(TestImages.Gif.Giphy) // Perf: Enable for local testing only
         };
 
+        // TODO: Find a better place for this
+        internal const string TestOutputRoot = "TestOutput/";
+
         protected string CreateOutputDirectory(string path, params string[] pathParts)
         {
             var postFix = "";
@@ -54,7 +57,7 @@ namespace ImageSharp.Tests
                 postFix  = "/" + string.Join("/", pathParts);
             }
 
-            path = "TestOutput/" + path + postFix;
+            path = TestOutputRoot + path + postFix;
 
             if (!Directory.Exists(path))
             {
--- a/tests/ImageSharp.Tests/Formats/Jpg/JpegTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Jpg/JpegTests.cs
@@ -8,6 +8,8 @@ using Xunit.Abstractions;
 
 namespace ImageSharp.Tests.Formats.Jpg
 {
+    using ImageSharp.Tests.TestUtilities;
+
     public class JpegTests
     {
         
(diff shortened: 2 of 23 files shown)


tests/ImageSharp.Tests/TestUtilities/TestUtilityExtensions.cs
```diff
@@ -0,0 +1,126 @@
...
+                            {
+                                if (!ca.Equals(cb))
+                                {
+                                    return false;
+                                }
+                            }
+                            else
+                            {
+                                ca.ToBytes(bytesA, 0, ComponentOrder.XYZ);
+                                cb.ToBytes(bytesB, 0, ComponentOrder.XYZ);
```
tocsoft: where are `bytesA` and `bytesB` being compared?

antonfirsov: OMG :S Thanks for pointing this out! :+1:

tocsoft: no problem.

tests/ImageSharp.Tests/FileTestBase.cs
```diff
@@ -54,7 +57,7 @@ protected string CreateOutputDirectory(string path, params string[] pathParts)
                 postFix  = "/" + string.Join("/", pathParts);
```
olivif: Could use Path.Combine here to avoid all the / concatenation 😄

tests/ImageSharp.Tests/Formats/Jpg/JpegTests.cs
```diff
@@ -17,62 +17,45 @@ public class JpegTests
...
-        {
-            fileName = CreateTestOutputFile(fileName);
-            Output?.WriteLine("Opened for write: "+fileName);
-            return File.OpenWrite(fileName);
-        }
-
-        public static IEnumerable<object[]> AllJpegFiles
-            => TestImages.Jpeg.All.Select(file => new object[] {TestFile.Create(file)});
-
+        // TODO: Turned off PixelTypes.All to be CI-friendly, what should be the practice?
```
olivif: Maybe we never use All?

tests/ImageSharp.Tests/TestUtilities/FlagsHelper.cs
```diff
@@ -0,0 +1,242 @@
...
+            return vals;
+        }
+
+        public static T Parse(string value, bool ignoreCase = false, bool parseNumeric = true)
+        {
+            return (T)Enum.ToObject(typeof(T), Converter.ParseInternal(value, ignoreCase, parseNumeric));
+        }
+
+        /// <summary>
+        /// Converts enum value to string
```
olivif: Not sure I understand why this is needed. If we had the strong enum type we could just do enum.ToString no? 😄

tests/ImageSharp.Tests/Image/PixelAccessorTests.cs
```diff
@@ -14,6 +15,114 @@ namespace ImageSharp.Tests
...
+                {
+                    for (int x = region.Left; x < region.Right; x++)
+                    {
+                        pixels[x, y] = color;
+                    }
+                }
+            }
+        }
+
+        [Theory]
```
olivif: Any existing tests which we should convert to this model for consistency? Maybe make a note or create an issue or something so we don't lose track of it 😄

tests/ImageSharp.Tests/TestUtilities/FlagsHelper.cs
olivif: I see there is some more parsing below but didn't dig too much. Maybe you can briefly explain how it works? ⚙️

dlemstra: I am now busy merging and doing some minor refactoring. The FlagsHelper is gone and only the GetSortedValues is still used but I wonder if we really need it.

tests/ImageSharp.Tests/Formats/Jpg/JpegTests.cs
dlemstra: All is used in the `PixelAccessorTests` but we really don't want to use it here :)

tests/ImageSharp.Tests/Image/PixelAccessorTests.cs
antonfirsov: Many components of the processors could be tested this way. Unfortunately I'm not yet familiar with their code.

tests/ImageSharp.Tests/TestUtilities/FlagsHelper.cs
antonfirsov: http://devopsreactions.tumblr.com/post/109288736875/useful-code-snippet

tests/ImageSharp.Tests/Formats/Jpg/JpegTests.cs
antonfirsov: The BMP output  differs from the input Jpeg with with several pixel formats. I don't know if it's a pixel type  specific "feature", or a bug. Decoders/encoders should be covered better somehow, but it's a longer discussion.

JimBobSquarePants: @antonfirsov Output will be different. Different color models only store certain information.

JimBobSquarePants: See the output here when I'm doing crazy things

https://github.com/JimBobSquarePants/ImageSharp/issues/27

PR #58, opened 2016-12-25
Jpeg Encoder Optimizations

antonfirsov:
## Jpeg Encoder Optimizations
- Refactored `JpegEncoderCore` to use `Block8x8F`
- Ported floating-point FDCT implementation from [dct_simd](https://github.com/norishigefukushima/dct_simd/blob/master/dct/dct8x8_simd.cpp)
- Refactored `ToYCbCr` to use `PixelArea<TColor>`
- Optimized the unzigging implementation in `WriteBlock`. Everything is done in a single step in `Block8x8F.UnZigDivRound()` now
- Other memory-locality optimizations

## General Jpeg Codec Cleanup
- Removed the old, integer IDCT related classes (`Block`, `IDCT`, `FDCT`)
- Removed `BlockF`
- Moved several nested classes out of `JpegEncoderCore`
- Moved small Decoder/Encoder related classes/srtucts to separate folders
- Had ideas about the next steps in the Jpeg code cleanup. Left [TODO-comments](https://github.com/antonfirsov/ImageSharp/blob/jpeg-optimizations/src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs#L862) in the code

## PixelArea
- Added an option to pool the internal `byte[]` of `PixelArea`
- Added a `Dispose(bool)` method
- Pooling is currently controlled by an optional constructor argument. I'm open for better solutions. 

## TestUtilities & Tests
- `[WithFile("foo.bmp", PixelTypes.StandardImageClass)]` was giving `Image<Color>` instead of `Image`. Fixed this.
- Fixed `TestImageProvider<TColor>.FileProvider`-s cache
- Added `Benchmark_JpegDecoder`

## ImageSharp.Benchmarks
- Added `MultiImageBenchmarkBase`
- Added `EncodeJpegMultiple`

## Benchmark results with the optimized version:
```
BenchmarkDotNet=v0.10.1, OS=Microsoft Windows NT 6.1.7601 Service Pack 1
Processor=Intel(R) Core(TM) i7-4810MQ CPU 2.80GHz, ProcessorCount=8
Frequency=2728232 Hz, Resolution=366.5377 ns, Timer=TSC
  [Host]     : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1085.0
  DefaultJob : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1085.0
```

Method |            Mode |        Mean |    StdDev | Scaled | Scaled-StdDev |     Gen 0 | Allocated |
-------------------------------------- |---------------- |------------ |---------- |------- |-------------- |---------- |---------- |
'EncodeJpegMultiple - ImageSharp' |             All | 114.8708 ms | 0.2612 ms |   3.58 |          0.01 |         - |    1.8 MB |
'EncodeJpegMultiple - System.Drawing' |             All |  32.0703 ms | 0.0567 ms |   1.00 |          0.00 | 2416.6667 |  12.93 MB |
'EncodeJpegMultiple - ImageSharp' | LargeImagesOnly |  99.2035 ms | 0.1410 ms |   5.61 |          0.02 |         - |   1.65 MB |
'EncodeJpegMultiple - System.Drawing' | LargeImagesOnly |  17.6716 ms | 0.0444 ms |   1.00 |          0.00 | 2012.5000 |  11.58 MB |
'EncodeJpegMultiple - ImageSharp' | SmallImagesOnly |  14.9452 ms | 0.0552 ms |   1.04 |          0.00 |         - | 153.38 kB |
'EncodeJpegMultiple - System.Drawing' | SmallImagesOnly |  14.4095 ms | 0.0443 ms |   1.00 |          0.00 |         - |   1.35 MB |


src/ImageSharp/Formats/Jpg/Components/Block8x8F.cs
```diff
@@ -15,6 +15,11 @@ namespace ImageSharp.Formats
     /// </summary>
     internal partial struct Block8x8F
     {
+        // Most of the static methods of this struct are instance methods by actual semantics: they use Block8x8F* as their first parameter.
```
olivif: Why not make them instance methods? Sounds a lot nicer, since that is the intention. Static is also generally harder to test.

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
```diff
@@ -559,42 +547,42 @@ private void ProcessDefineHuffmanTablesMarker(int remaining)
             }
         }
 
-        private void ProcessDefineHuffmanTablesMarkerLoop(ref Huffman huffman, ref int remaining)
+        private void ProcessDefineHuffmanTablesMarkerLoop(ref HuffmanTree huffmanTree, ref int remaining)
```
olivif: While we're on cleanup, I don't think these refs are needed here. It will be passed by reference not value anyway, the only time you'd want to use ref is if you want to re-assign the object and thus change the reference.

src/ImageSharp/Formats/Jpg/Components/Block8x8F.cs
antonfirsov: Slow:
```
public unsafe float this[int idx]
{
    get
    {
        fixed (Block8x8F* p = &this)
        {
            float* fp = (float*)p;
            return fp[idx];
        }
    }
}
```

Fast:
```
public static unsafe float GetScalarAt(Block8x8F* blockPtr, int idx)
{
    float* fp = (float*)blockPtr;
    return fp[idx];
}
```

There's no need to fix the block if it's on the stack, or it's member of an already-fixed heap object. Unfortunately the only way to "express" it in C# is to have a static method that takes `Block8x8F*` :(

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
antonfirsov: Again: Perf.
`HuffmanTree` should be a class in a "normal" C# code. I made it to be a struct because I wanted to improve data locality by removing the "array of objects" memory structure in `JpegDecoderCore.huffmanTrees`.
`HuffmanTree` has a lots of members, so copying [it inside a nested loop](https://github.com/antonfirsov/ImageSharp/blob/jpeg-optimizations/src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs#L1720) would be very slow.

olivif: Ohh it is a struct, did not see that coming! That makes sense, don't mind me then :smile:

antonfirsov: But:
`ProcessDefineHuffmanTablesMarkerLoop()` really should be moved to `HuffmannTree`. I'm already working on a decoder code cleanup to make further optimizations possible. I will take care of this too. Thanks for noticing the anomaly! :)

JimBobSquarePants: @antonfirsov I think the confusion is there as you are passing the int params by ref also. The compiler is heavily optimised for passing small structs so it's uncommon to see them passed by reference.

JimBobSquarePants:
@antonfirsov 46  files is a lot of changes to review! I think next time it might be prudent to split the changes up a little. 

Do you have to before benchmarks so we can compare like for like? My current machine has thermal limiting which ruins benchmarking.

Do you have at idea what causes the slowdown on larger images? It's quite a dramatic change.

src/ImageSharp/PixelAccessor.cs
```diff
@@ -110,6 +110,26 @@ protected override void CopyToZYX(PixelArea<Color> area, int sourceY, int source
         }
 
         /// <inheritdoc />
+        protected override unsafe void CopyToXYZ(PixelArea<Color> area, int sourceY, int sourceX, int width, int height)
```
JimBobSquarePants: Ha! I actually wrote this myself the other day then never pushed it. Great minds.....

tests/ImageSharp.Benchmarks/Image/DecodeJpegMultiple.cs
```diff
@@ -1,106 +1,38 @@
...
     using System.IO;
-    using System.Linq;
-
     using BenchmarkDotNet.Attributes;
 
     using Image = ImageSharp.Image;
     using ImageSharpSize = ImageSharp.Size;
 
-    public class DecodeJpegMultiple
+    public class DecodeJpegMultiple : MultiImageBenchmarkBase
```
JimBobSquarePants: This is cool.

tests/ImageSharp.Tests/Formats/Jpg/ReferenceImplementationsTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            MutableSpan<float> temp = new MutableSpan<float>(64);
+
+            ReferenceImplementations.iDCT2D_llm(floatSrc, dest, temp);
+
+            for (int i = 0; i < 64; i++)
+            {
+                float expected = intData[i];
+                float actual = dest[i];
+
+                Assert.Equal(expected, actual, new ApproximateFloatComparer(1f));
```
JimBobSquarePants: Wouldn't it be easier to use the overload specifying the decimal places as we have elsewhere?

tests/ImageSharp.Tests/TestUtilities/Factories/ImageFactory.cs
```diff
@@ -10,5 +10,11 @@ public class ImageFactory : GenericFactory<Color>
         public override Image<Color> CreateImage(byte[] bytes) => new Image(bytes);
 
         public override Image<Color> CreateImage(int width, int height) => new Image(width, height);
+
+        public override Image<Color> CreateImage(Image<Color> other)
+        {
+            Image img = (Image)other;
```
JimBobSquarePants: `return new Image<Color>(other);`

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
```diff
@@ -534,122 +620,84 @@ private int WriteBlock(ref Block block, QuantIndex index, int prevDC)
...
-            this.buffer[3] = (byte)verticalResolution;
+            // Loop through and collect the tables as one array.
+            // This allows us to reduce the number of writes to the stream.
+            int dqtCount = (QuantizationTableCount * Block8x8F.ScalarCount) + QuantizationTableCount;
+            byte[] dqt = ArrayPool<byte>.Shared.Rent(dqtCount);
+            int offset = 0;
 
-            this.outputStream.Write(this.buffer, 0, 4);
-        }
+            JpegUtils.WriteDataToDqt(dqt, ref offset, QuantIndex.Luminance, ref this.luminanceQuantTable);
```
olivif: That's my rule as well. If anyone else has a need for it then it can move into a common place.

src/ImageSharp/Formats/Jpg/Components/DCT.cs
```diff
@@ -0,0 +1,340 @@
...
+            d.V7L = c0 - c3;
+        }
+
+        /// <summary>
+        /// Original:
+        /// <see>
+        ///     <cref>https://github.com/norishigefukushima/dct_simd/blob/master/dct/dct8x8_simd.cpp#L15</cref>
+        /// </see>
+        /// </summary>
+        /// <param name="s">Source</param>
```
olivif: Actually I haven't hit that rule at all. It should trigger on parameter docs, class docs. Maybe they've changed the spec for it. Ahhh, it's been disabled, see [here](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/KnownChanges.md).

src/ImageSharp/Image/PixelAccessor{TColor}.cs
```diff
@@ -186,8 +186,7 @@ public void CopyFrom(PixelArea<TColor> area, int targetY, int targetX = 0)
             int width = Math.Min(area.Width, this.Width - targetX);
             int height = Math.Min(area.Height, this.Height - targetY);
 
-            this.CheckDimensions(width, height);
-
+            // this.CheckDimensions(width, height); TODO: Why was width == 0 or height == 0 considered a problem? Copy implementations do not fail on this (just do nothing)!
```
dlemstra: The intention is that the developer should get a notification when they call the method with an invalid x and y offset. Calling this method with an x or y offset that does nothing is also considered invalid.  The caller should add a check for this instead of doing a non-copy inside this class.

src/ImageSharp/Image/PixelArea{TColor}.cs
```diff
@@ -146,34 +168,48 @@ public PixelArea(int width, int height, ComponentOrder componentOrder, int paddi
...
         /// Gets the data pointer.
         /// </summary>
         public byte* PixelBase { get; private set; }
 
         /// <summary>
-        /// Gets the component order.
+        /// Gets number of bytes in a row.
         /// </summary>
-        public ComponentOrder ComponentOrder { get; }
+        public int RowByteCount { get; }
```
dlemstra: Some of these are all ones of mine :) But thanks for reviewing them.

antonfirsov:
@JimBobSquarePants:
To have a proper before/after comparsion I have to reproduce `JpegEncoderMutliple` on your master branch. I will take the time for this around 11PM CET, but there will be no big differences compared to my inital benchmarks (posted on gitter):
Before:
https://github.com/antonfirsov/ImageSharp/blob/experimental/JpegEncoderResults01.md
After:
https://github.com/antonfirsov/ImageSharp/blob/experimental/JpegEncoderResults03.md

Sorry for the large amount_of_changes/PR ratio, but I got addicted to this project. There is actually even more stuff waiting to be pushed to your review queue :D I'm already working on a [decoder code cleanup](https://github.com/antonfirsov/ImageSharp/blob/jpeg-optimizations-experimental/src/ImageSharp/Formats/Jpg/Components/Decoder/JpegScanDecoder.cs), which might enable more optimizations.

Larger images:
It seems to me that `Scaled` entry scales with the number of my cores. To prove this, I have to reproduce the benchmarks with only 1 core assigned to the process. I will be very happy if this is the case, because I see a possibility to parallelize the CPU-intensive work (DCT, color conversion, zigzagging etc.). However it's not trivial, if I want to [avoid false sharing](http://www.drdobbs.com/parallel/eliminate-false-sharing/217500206).

tests/ImageSharp.Tests/TestUtilities/Factories/ImageFactory.cs
antonfirsov: `Image` != `Image<Color>`, because it creates `PixelAccessor` instead of `PixelAccessor<Color>`, which has different method overloads.
Distinguishing between them in test cases results in increased test coverage. There is `PixelTypes.StandardImageClass` VS `PixelTypes.Color` to do this. `ImageFactory : GenericFactory<Color>` serves the `StandardImageClass` case, so not returning `new Image<Color>(other)` *is the purpose of this overload*.

tests/ImageSharp.Tests/Formats/Jpg/ReferenceImplementationsTests.cs
antonfirsov: LOL. I'm working with xUnit since 1.5 years, and I haven't noticed this overload yet :D

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
antonfirsov: In most of the cases this is my rule too, but:
1. Style(Killer)Cop places this method at a random place in JpegEncoderCore. I have to scroll up & down, impossible to oversee the code :(
2. `JpegEncoderCore` has still too much code. If certain parts are members of static utility classes, it makes things easier for a future split-up/refactor.

tests/ImageSharp.Tests/TestUtilities/Factories/ImageFactory.cs
JimBobSquarePants: Ah of course, makes sense now. I forgot we do that.

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
```diff
@@ -834,83 +833,107 @@ private void WriteStartOfScan<TColor>(PixelAccessor<TColor> pixels)
             this.Emit(0x7f, 7);
         }
 
+#pragma warning disable SA1201 // MethodShouldNotFollowAStruct
+
         /// <summary>
-        /// Encodes the image with no subsampling.
+        ///     Poor man's stackalloc for Encode420.
+        ///     This struct belongs to Encode420. Much easeier to understand code if they are close to each other. Why should I
```
antonfirsov: Good news: It might be refactored out of the encoder, because the decoder also has some cases, when it could be better to have a Block8x8F X 4 value type, than a `new  Block8x8F[4]` array.

JimBobSquarePants:
@antonfirsov I just updated your links to use the timestamped versions so we can never lose them.

A 2X improvement is nice! The memory improvements are incredible also. Great stuff! 

I'll tell you what though, that C# is starting to look less and less so... I'm gonna have to upskill to keep up with you.

Don't worry about the size of the PR, it's easy to get addicted... 4 1/2 years I've found myself doing imaging for now. No idea how I got here! 

We'll definitely have to try to make future changes more incremental though I think, or at least write as many comments per code as we can. Some of the changes are really quite difficult to follow. 

Definitely keen to see what plans you have re parallelization. Not read that article before and will need to check some existing work now. 

As for the decoder... You're crazy but I love it!

antonfirsov:
> I'll tell you what though, that C# is starting to look less and less so...

Unfortunately C#-s nice abstractions are not [zero cost](https://www.google.hu/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=zero%20cost%20abstractions) and you must drop some of them, or do strange tricks with them if you need [data driven](https://en.wikipedia.org/wiki/Data-driven_programming) code taking care of [cache coherence](https://en.wikipedia.org/wiki/Cache_coherence). With Jpeg I went so far, that I ended up writing C code in C#. I'm not happy with this. 
The only thing that could help us here is to early-adapt `Span<T>`. It's already [moved to the corefx repo](https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Span.cs) from corefxlabs, and the API seems to be pretty stable. We can include it as git submodule. I think we need to introduce it in ImageSharp before it gets too late!

> We'll definitely have to try to make future changes more incremental though I think, or at least write as many comments per code as we can. Some of the changes are really quite difficult to follow.

I was waiting for #57 being accepted before opening this PR & I will open my next one as soon as this one will be closed (with lot's of changes again).
I see a pretty high chance this project could become my main focus of everyday-geekness-outside-work in the next few months. I'll be very happy if we could find a way to speed-up merging my work. If you need more docs & comments from my side, I'm happy with it. However functions with lacking comments often belong to the original Jpeg codebase where I fail to understand all the details.

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
```diff
@@ -484,29 +429,170 @@ private void EmitHuffRLE(HuffIndex index, int runLength, int value)
...
+        /// <typeparam name="TColor">
+        /// The pixel format.
+        /// </typeparam>
+        /// <param name="pixels">
+        /// The pixel accessor providing access to the image pixels.
+        /// </param>
+        private void Encode444<TColor>(PixelAccessor<TColor> pixels)
+            where TColor : struct, IPackedPixel, IEquatable<TColor>
+        {
+            // TODO: Need a JpegEncoderScanProcessor<TColor> struct to encapsulate all this mess:
```
antonfirsov: This work is in progress for the decoder, I will describe all the details in my forcoming PR in comments & .MD. 
The changes I want to implement in the encoder will follow the same logic.

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
```diff
@@ -534,122 +620,84 @@ private int WriteBlock(ref Block block, QuantIndex index, int prevDC)
...
-            this.WriteProfile(image.ExifProfile);
+            this.outputStream.Write(dqt, 0, dqtCount);
+            ArrayPool<byte>.Shared.Return(dqt);
         }
 
         /// <summary>
         /// Writes the EXIF profile.
         /// </summary>
-        /// <param name="exifProfile">The exif profile.</param>
+        /// <param name="exifProfile">
```
antonfirsov: StyleCop extension again. At least I managed to remove the extra spaces here :D

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
```diff
@@ -834,83 +833,107 @@ private void WriteStartOfScan<TColor>(PixelAccessor<TColor> pixels)
...
+            Block8x8F onStackChrominanceQuantTable = this.chrominanceQuantTable;
+
+            UnzigData unzig = UnzigData.Create();
 
             // ReSharper disable once InconsistentNaming
-            int prevDCY = 0, prevDCCb = 0, prevDCCr = 0;
+            float prevDCY = 0, prevDCCb = 0, prevDCCr = 0;
 
-            for (int y = 0; y < pixels.Height; y += 16)
+            using (var rgbBytes = new PixelArea<TColor>(8, 8, ComponentOrder.XYZ, true))
```
antonfirsov: Dat controversial var. With simple types I dont like it either. With complex types ... I will refactor it for consistency :)
Sometimes I introduce vars unintentionally with R#

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
```diff
@@ -2,295 +2,169 @@
...
 
     /// <summary>
-    /// Image encoder for writing an image to a stream as a jpeg.
+    ///     Image encoder for writing an image to a stream as a jpeg.
     /// </summary>
     internal unsafe class JpegEncoderCore
     {
         /// <summary>
-        /// The number of quantization tables.
+        ///     The number of quantization tables.
```
JimBobSquarePants: Yeah, easy enough to fix. Don't worry about it. 

This is by far the best plugin for VS to remove trailing whitespace btw, it does it automatically on save.
https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
JimBobSquarePants: Okily dokily..

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
JimBobSquarePants: They're better of in the class though than in a utility class. They swiftly become dumping grounds.

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
JimBobSquarePants: Good good. 😄

src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs
JimBobSquarePants: Cool!

src/ImageSharp/Formats/Jpg/Components/DCT.cs
JimBobSquarePants: Monsters! I'm gonna pitch in there.

src/ImageSharp/Image/PixelAccessor{TColor}.cs
JimBobSquarePants: Definitely re-enable it then.

JimBobSquarePants:
@antonfirsov 

>However functions with lacking comments often belong to the original Jpeg codebase where I fail to understand all the details.

Absolutely appreciate how difficult this must be for you. The original Golang implementation is a mess. I don't know how they manage the framework with such low standards.

>The only thing that could help us here is to early-adapt Span<T>. It's already moved to the corefx repo from corefxlabs, and the API seems to be pretty stable. We can include it as git submodule. I think we need to introduce it in ImageSharp before it gets too late!

Agreed. There's a couple of things missing (Fill, Clear) but they are coming soon. We should definitely be  careful while using it in it's current form though as things will be subject to change still I'm sure.

Let's go ahead and try adding `Span<T>` we know we will be using it. 

>I'll be very happy if we could find a way to speed-up merging my work.

Yes, and I've been giving it some thought. The contributions that you, @tocsoft and @olivif have provided so far have been extraordinary and the library simply wouldn't be anywhere as good as it is just now without your efforts. So... I'm extending an invite to the three of you. If you would like and you think you can provide the commitment over time I can officially add you all to the team. 

But....
lgppr40105 great-power-great-responsibility-spiderman-art-print

It's paramount that we stick to the rules as I set them (StyleCop sets them). It's tried and tested and the only way I can ensure consistency throughout the library. 

We'll still use PR's for larger changes but smaller incremental changes could be made much more easily. 

What do you think?

src/ImageSharp/Image/PixelAccessor{TColor}.cs
antonfirsov: @JimBobSquarePants @dlemstra 
```
int width = Math.Min(area.Width, this.Width - targetX);
int height = Math.Min(area.Height, this.Height - targetY);
```
This will be calculated 2X then. It seems wasteful to me & also seems to be a code smell. It's not the only thing I don't like about the design of the current `PixelArea` <--> `PixelAccessor` and other copying solutions. The switch statements on `ComponentOrder` are very inefficient perf killers.
It's not the time to change the design however, so now I will re-enable the check and duplicate code on the call site, but we should definitely come back to rethink this.

src/ImageSharp/Image/PixelArea{TColor}.cs
```diff
@@ -85,8 +90,9 @@ public PixelArea(int width, int height, byte[] bytes, ComponentOrder componentOr
         /// <param name="width">The width.</param>
         /// <param name="height">The height.</param>
         /// <param name="componentOrder">The component order.</param>
-        public PixelArea(int width, int height, ComponentOrder componentOrder)
-            : this(width, height, componentOrder, 0)
+        /// <param name="usePool">True if the buffer should be rented from ArrayPool</param>
+        public PixelArea(int width, int height, ComponentOrder componentOrder, bool usePool = false)
```
antonfirsov: https://github.com/antonfirsov/ImageSharp/blob/jpeg-optimizations/src/ImageSharp/Formats/Jpg/JpegEncoderCore.cs#L459

src/ImageSharp/Image/PixelAccessor{TColor}.cs
JimBobSquarePants: How about explicit method calls instead of a switch then? More boilerplate for us but should yield much better performance.

src/ImageSharp/Image/PixelArea{TColor}.cs
```diff
@@ -146,34 +168,48 @@ public PixelArea(int width, int height, ComponentOrder componentOrder, int paddi
...
         /// </summary>
         public int Width { get; }
 
         /// <summary>
-        /// Gets the height.
+        /// Gets the pool used to rent <see cref="Bytes"/>, when it's not coming from an external source
         /// </summary>
-        public int Height { get; }
+        // ReSharper disable once StaticMemberInGenericType
+        private static ArrayPool<byte> BytesPool => ArrayPool<byte>.Shared;
```
antonfirsov: I'm afraid, using `ArrayPool<T>.Shared` is not the best for this library. Sample issue: What if you exceed the maximum size with large images?
I used it here because I was "lazy" to find the best solution for `PixelArea˛`. In such cases defining an extra indirection level could be useful for a future refactor. I will leave a comment about this.

src/ImageSharp/Image/PixelArea{TColor}.cs
JimBobSquarePants: My bad! All cool then.

src/ImageSharp/Image/PixelAccessor{TColor}.cs
antonfirsov: Much better! This will add instant perf boost to non-`Color` pixel type code using `IPackedBytes.ToBytes()`. Lots of refactor work however, Maybe I can take care of this later.

src/ImageSharp/Image/PixelArea{TColor}.cs
```diff
@@ -29,6 +29,11 @@ public sealed unsafe class PixelArea<TColor> : IDisposable
         private IntPtr dataPointer;
 
         /// <summary>
+        /// True if <see cref="Bytes"/> was rented from <see cref="BytesPool"/> by the constructor
+        /// </summary>
+        private bool isBufferRented;
```
antonfirsov: Yeah, but I still don't like the idea of controlling a major aspect of a class with a bool parameter. It was a lazy solution from my side, I'd like to see something like `PixelArea<T>.CreateWithRentedBuffer()` instead. What do you think?

src/ImageSharp/Image/PixelAccessor{TColor}.cs
JimBobSquarePants: It'll add a performance boost to the `Color` version also. I'll do that once we've merged this PR so don't worry about it.

src/ImageSharp/Image/PixelArea{TColor}.cs
JimBobSquarePants: I'd rather something much more explicit so yeah, I'm cool with that.

antonfirsov:
Cool!
I have to get used to that StyleCop guy however, we are definitely not friends yet :D It's tried & tested mostly with projects having a different focus & it makes things extra difficult & non-straightforward with the C-ish `struct`-ful code we have to develop for perf critical parts. 

But I'd <3 to be taken by :oncoming_police_car:  &  baton beaten if this is the price for contributing to a great library! :D

JimBobSquarePants:
@antonfirsov Great stuff! Let's get the final changes you suggested to `PixelArea` done and a little cleanup and I'll merge this later today. Welcome aboard! 😄

src/ImageSharp/Formats/Jpg/Utils/MutableSpanExtensions.cs
```diff
@@ -0,0 +1,164 @@
...
+
+namespace ImageSharp.Formats.Jpg
+{
+    using System.Numerics;
+    using System.Runtime.CompilerServices;
+
+    /// <summary>
+    /// MutableSpan Extensions
+    /// </summary>
+    internal static class MutableSpanExtensions
```
antonfirsov: The methods returning `MutableSpan<T>` are not "spanny", need to refactor them. (Next time, if possible.)

src/ImageSharp/Image/PixelAccessor{TColor}.cs
antonfirsov: @dlemstra @JimBobSquarePants  regarding `CheckDimensions`:
[I made this hack](https://github.com/antonfirsov/ImageSharp/blob/jpeg-optimizations/src/ImageSharp/Formats/Jpg/Utils/JpegUtils.cs#L47) to conform to the current dimension-checking API definition of `CopyTo()`.
It's like if you defined `Array.Copy()` to throw exception with `Array.Copy(src, dest, 0)`. There are reasons .NET guys have not chosen this definition :)

src/ImageSharp/Image/PixelArea{TColor}.cs
antonfirsov: Really need to sleep, just made the field readonly for now :( Actually using a rented buffer could be the default behavior, if `Bytes` is not coming from the outside, but I wanted to keep the existing Gif & Png code safe. Confused now, we have to think this over later.

tests/ImageSharp.Tests/Formats/Jpg/ReferenceImplementations.cs
```diff
@@ -56,9 +362,11 @@ private static void iDCT1Dllm_32f(MutableSpan<float> y, MutableSpan<float> x)
             float a0, a1, a2, a3, b0, b1, b2, b3;
             float z0, z1, z2, z3, z4;
 
+            float r0 = 1.414214f;
```
antonfirsov: I intentionally wanted to keep it similar to the original C++ code. Even it's mistakes can serve as documentation if someone ever goes into DCT implementation details. This is not the core codebase, it's here for testing & documentation.

olivif:
> Yes, and I've been giving it some thought. The contributions that you, @tocsoft and @olivif have provided so far have been extraordinary and the library simply wouldn't be anywhere as good as it is just now without your efforts. So... I'm extending an invite to the three of you. If you would like and you think you can provide the commitment over time I can officially add you all to the team.

This was the first thing I read as I woke up, I have to say that's a pretty nice way to start the day 🍰. I'd love to join, I've had a really great time contributing to this project and I'm super excited about what we can do with it. 

> We'll still use PR's for larger changes but smaller incremental changes could be made much more easily.

Actually was thinking a bit about this. Do you think it would make sense to create PRs for everything? Even if you merge them yourself 2 seconds later, it gives me and the others the ability to go back to closed PRs and see what was added, maybe with a nice description next to it. And I can ask a question or two, it would be a great learning tool for the new people or anyone wanting to contribute really. And you also get the post code-review benefits. What do you think? 😄 

> I have to get used to that StyleCop guy however, we are definitely not friends yet

Takes a bit of time but now it's just second nature to me, so give it a while. I can definitely see what @JimBobSquarePants is saying, I've worked on some really large projects, and it is pretty nice to browse the code and know where things are and have them be structured, named, documented in the same way.

tocsoft:
I'm with @olivif  just woke up to this and feel rather faltered and would love to join the team... I feel I could really 'own' the drawing stuff then 😁.

Don't think being a core contributor will make a massive difference to my workflow while contributing though, as I'll still probably make PRs for everything I work on (I always feel it nice to have a 2nd pair of eyes on things), will just probably just change the repository that I make my branches in. 

Also will be nice to be able to help with reviewing PRs and helping out with issue tracking.

JimBobSquarePants:
Awesome! I've sent you both invites and added you to the readme. 

>Do you think it would make sense to create PRs for everything?

PR's are probably the best thing to do for the most part but sometimes i'll forget.

tocsoft:
well as you say you don't call yourself 'Grand High Eternal Dictator' for nothing 😉

PR #59, opened 2016-12-28
Optimizing YCbCr to use a backing vector. Refactoring namespaces for Color Spaces.

olivif:
Touch #3

### DIFF
--- a/src/ImageSharp/Colors/Spaces/YCbCr.cs
+++ b/src/ImageSharp/Colors/Spaces/YCbCr.cs
@@ -7,6 +7,7 @@ namespace ImageSharp.Colors.Spaces
 {
     using System;
     using System.ComponentModel;
+    using System.Numerics;
 
     /// <summary>
     /// Represents an YCbCr (luminance, blue chroma, red chroma) color conforming to the full range standard used in digital imaging systems.
@@ -20,6 +21,11 @@ namespace ImageSharp.Colors.Spaces
         public static readonly YCbCr Empty = default(YCbCr);
 
         /// <summary>
+        /// The backing vector for SIMD support.
+        /// </summary>
+        private Vector3 backingVector;
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="YCbCr"/> struct.
         /// </summary>
         /// <param name="y">The y luminance component.</param>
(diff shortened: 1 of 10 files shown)


src/ImageSharp/Colors/Spaces/YCbCr.cs
```diff
@@ -117,13 +121,7 @@ public YCbCr(byte y, byte cb, byte cr)
...
         public override int GetHashCode()
         {
-            unchecked
-            {
-                int hashCode = this.Y.GetHashCode();
-                hashCode = (hashCode * 397) ^ this.Cb.GetHashCode();
-                hashCode = (hashCode * 397) ^ this.Cr.GetHashCode();
-                return hashCode;
-            }
+            return GetHashCode(this);
```
tocsoft: can't this just be `return this.backingVector.GetHashCode()` directly and drop the `GetHashCode(...)` method? I don't think the indirection is needed.

olivif: I completely agree, I was just following the pattern the other color spaces use. I can fix all of them while I'm at it, I thought maybe there was a reason for it 😄

tocsoft: might be worth deferring to @JimBobSquarePants to see if there specific logic to it then.

olivif: I've made the changes locally, I'll wait for an ack from @JimBobSquarePants before pushing 😄

JimBobSquarePants: Resharper will actually warn if you don't do this since there is the potential for the hashcode to change during the lifetime of the object. I've not been crazy strict about it though, especially in the `IPackedPixel` implementations. Truth be told, I'm not sure it's something I should really be concerned about.

http://www.jetbrains.com/resharperplatform/help?Keyword=NonReadonlyMemberInGetHashCode

olivif: Ahh I see. Well, since the backing vector doesn't change, we can actually make it `readonly` no?

JimBobSquarePants: Sounds like a plan!

olivif:
All fixed, this is ready to be merged if there's nothing else 😄 I don't know why codecov is shouting at me, the coverage should be the same ..

src/ImageSharp/Colors/Spaces/YCbCr.cs
```diff
@@ -28,28 +34,26 @@ public struct YCbCr : IEquatable<YCbCr>
         public YCbCr(byte y, byte cb, byte cr)
             : this()
         {
-            this.Y = y;
-            this.Cb = cb;
-            this.Cr = cr;
+            this.backingVector = Vector3.Clamp(new Vector3(y, cb, cr), Vector3.Zero, new Vector3(255));
```
JimBobSquarePants: Make `new Vector3(255)` a private static member. It reduces the allocations.

olivif: good catch! Fixed this one, also made a note to fix it in other places 😄

tocsoft:
looks good to me, i'm going to use my new powers and merge this 😁

olivif:
thanks @tocsoft !

PR #60, opened 2016-12-28
Refactoring Epsilon const value to remove duplication.

olivif:
I noticed we had a bunch of `private const float Epsilon = 0.001F;` duplicated throughout the code, and I see no reason why that shouldn't be in a common place.

I put it in the `Common` namespace, not sure how you feel about that. I fixed ColorSpaces and Transforms to use it, but there are a few more which I can fix if you guys think this is worth doing 😄 

Also added a test to protect the value, if that is accidentally changed we could be in quite some trouble.

### DIFF
--- a/src/ImageSharp/Common/Constants.cs
+++ b/src/ImageSharp/Common/Constants.cs
@@ -3,7 +3,7 @@
 // Licensed under the Apache License, Version 2.0.
 // </copyright>
 
-namespace ImageSharp.Common
+namespace ImageSharp
 {
     /// <summary>
     /// Common constants used throughout the project
--- a/src/ImageSharp/Common/Constants.cs
+++ b/src/ImageSharp/Common/Constants.cs
@@ -8,7 +8,7 @@ namespace ImageSharp.Common
     /// <summary>
     /// Common constants used throughout the project
     /// </summary>
-    public static class Constants
+    internal static class Constants
     {
         /// <summary>
         /// The epsilon for comparing floating point numbers.
--- a/src/ImageSharp/Common/Constants.cs
+++ b/src/ImageSharp/Common/Constants.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="CommonConstants.cs" company="James Jackson-South">
+﻿// <copyright file="Constants.cs" company="James Jackson-South">
 // Copyright (c) James Jackson-South and contributors.
 // Licensed under the Apache License, Version 2.0.
 // </copyright>
@@ -8,7 +8,7 @@ namespace ImageSharp.Common
     /// <summary>
     /// Common constants used throughout the project
     /// </summary>
-    public static class CommonConstants
+    public static class Constants
     {
         /// <summary>
         /// The epsilon for comparing floating point numbers.
--- a/src/ImageSharp/Colors/ColorspaceTransforms.cs
+++ b/src/ImageSharp/Colors/ColorspaceTransforms.cs
@@ -8,7 +8,6 @@ namespace ImageSharp
     using System;
     using System.Numerics;
     using Colors.Spaces;
-    using Common;
 
     /// <summary>
     /// Packed vector type containing four 8-bit unsigned normalized values ranging from 0 to 255.
--- a/src/ImageSharp/Colors/ColorspaceTransforms.cs
+++ b/src/ImageSharp/Colors/ColorspaceTransforms.cs
@@ -8,6 +8,7 @@ namespace ImageSharp
     using System;
     using System.Numerics;
     using Colors.Spaces;
+    using Common;
 
     /// <summary>
     /// Packed vector type containing four 8-bit unsigned normalized values ranging from 0 to 255.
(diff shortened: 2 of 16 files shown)


src/ImageSharp/Common/Constants.cs
```diff
@@ -0,0 +1,18 @@
...
+// Copyright (c) James Jackson-South and contributors.
+// Licensed under the Apache License, Version 2.0.
+// </copyright>
+
+namespace ImageSharp.Common
+{
+    /// <summary>
+    /// Common constants used throughout the project
+    /// </summary>
+    public static class Constants
```
tocsoft: I would be inclined to make this `internal` can't see why we would want to expose as part of the public API.

olivif: Agreed and fixed 😄

JimBobSquarePants: I've a feeling I have a few different values for Epsilon littered throughout the codebase. They should probably all be the same.

olivif:
Moving comment here since it's hidden up there. 

@JimBobSquarePants 
> I've a feeling I have a few different values for Epsilon littered throughout the codebase. They should probably all be the same.

Yeah, they are everywhere. These are the instances I could find. Are we OK with changing all of them to `0.001f` ? I also need to look in the tests see if this will break something and potentially refactor there to use a test epsilon or the same one.

```
Find all "Epsilon =", Subfolders, Find Results 1, Entire Solution, ""
 Colors\Spaces\CieLab.cs(26):        private const float Epsilon = 0.001f;
 Colors\Spaces\CieXyz.cs(26):        private const float Epsilon = 0.001f;
 Colors\Spaces\Cmyk.cs(25):        private const float Epsilon = 0.001f;
 Colors\Spaces\Hsl.cs(25):        private const float Epsilon = 0.001F;
 Colors\Spaces\Hsv.cs(25):        private const float Epsilon = 0.001F;
 Colors\ColorspaceTransforms.cs(25):        private const float Epsilon = 0.001F;
 Colors\Vector4BlendTransforms.cs(20):        private const float Epsilon = 0.0001F;
 Common\Helpers\ImageMaths.cs(94):            const float Epsilon = .00001F;
 Common\Helpers\ImageMaths.cs(169):            const float Epsilon = .00001f;
 Common\Helpers\ImageMaths.cs(281):            const float Epsilon = .00001F;
 Drawing\Processors\DrawPathProcessor.cs(30):        private const float Epsilon = 0.001f;
 Drawing\Processors\FillShapeProcessor.cs(24):        private const float Epsilon = 0.001f;
 Filters\Processors\Effects\BackgroundColorProcessor.cs(22):        private const float Epsilon = 0.001f;
 Filters\Processors\Transforms\RotateProcessor.cs(73):            const float Epsilon = .0001F;
 Filters\Processors\Transforms\RotateProcessor.cs(94):            const float Epsilon = .0001F;
 Numerics\LongRational.cs(93):            double epsilon = bestPrecision ? double.Epsilon : .000001;
 Quantizers\Wu\WuQuantizer.cs(39):        private const float Epsilon = 1e-5F;
```

JimBobSquarePants:
@olivif Yeah that's fine. Go ahead.

antonfirsov:
I wonder: is it possible that some day `AlmostEquals()` and it's friends will need to take a variable epsilon from outside?

JimBobSquarePants:
@antonfirsov  TBH I'm not even sure `AlmostEqual` should be a thing. It's a hack. I would vote for removing it and doing comparisons like any other object containing floating point properties.

olivif:
@JimBobSquarePants I was thinking about that actually. Who actually uses equality for color spaces? Is it only for test?

antonfirsov:
Having such things like `AlmostEqual` is quite usual in graphical applications, becasue of [numerical instability](https://en.wikipedia.org/wiki/Numerical_stability)!

Example use case: in the assertion part of the test cases you can load a pre-defined reference image and compare it with an epsilon, because you really can't expect exact equality with floating point algorithms!

JimBobSquarePants:
@antonfirsov  Every day is a school day! Sorry @olivif let's keep it then.

antonfirsov:
But it must not be mixed with standard `Equals`!
[This code is wrong](https://github.com/JimBobSquarePants/ImageSharp/blob/760439185e24b5fa562cce06a67a42e9cf60ccf0/src/ImageSharp/Colors/Colorspaces/Hsl.cs#L187)

It can happen that `v1.GethHashCode() != v2.GetHashCode()` while `v1.Equals(v2) == true` !

olivif:
@antonfirsov yes, that is true. Equals should use 0 precision right?

antonfirsov:
Better & faster if it's a completely separate thing.

antonfirsov:
Just delegate it to `backingVector.Equals(other.backingVector)`

olivif:
@antonfirsov I will do this, but in another PR as this one is already getting quite big. 

Ready to merge this btw, unless anyone has any other thoughts 😄

olivif:
Now it's ready to be merged, unless anyone has any other thoughts 😄 I removed the `Common` namespace as @dlemstra suggested.

PR #62, opened 2016-12-28
Jpeg decoder refactor & code cleanup

antonfirsov:
### Jpeg decoder refactor & code cleanup
- ~~I opened this PR to start the discussion & the reviewing process about the stuff that's not xmldoc/StyleCop specific~~
- ~~Part of the documentation is missing, the PR is failing the StyleCop verifications, I will continously work fixing it~~
**UPDATE:** all verifications pass now

### Major changes
- The SOS decoder logic is moved to [JpegScanDecoder](https://github.com/antonfirsov/ImageSharp/blob/jpeg-optimizations/src/ImageSharp/Formats/Jpg/Components/Decoder/JpegScanDecoder.cs)
  - [**See the docs!!!**](https://github.com/antonfirsov/ImageSharp/blob/jpeg-optimizations/src/ImageSharp/Formats/Jpg/Components/Decoder/JpegScanDecoder.md)
- The `GrayscaleImage` class has been refactored into `JpegPixelArea` and used everywhere to represent pixel regions of jpeg channels

### Performance
It was not the goal this time, but there are some improvements!
**Before:**
```
                                Method |            Mode |        Mean |    StdDev | Scaled | Scaled-StdDev |     Gen 0 |     Gen 1 |     Gen 2 | Allocated |
-------------------------------------- |---------------- |------------ |---------- |------- |-------------- |---------- |---------- |---------- |---------- |
     'DecodeJpegMultiple - ImageSharp' |             All | 322.3783 ms | 0.2442 ms |   2.80 |          0.01 | 4575.0000 | 4304.1667 | 4212.5000 |     64 MB |
 'DecodeJpegMultiple - System.Drawing' |             All | 115.1111 ms | 0.5644 ms |   1.00 |          0.00 | 1137.5000 |         - |         - |    5.4 MB |
     'DecodeJpegMultiple - ImageSharp' | LargeImagesOnly | 244.6698 ms | 0.5438 ms |   3.34 |          0.02 | 2241.6667 | 2179.1667 | 2179.1667 |  37.46 MB |
 'DecodeJpegMultiple - System.Drawing' | LargeImagesOnly |  73.1888 ms | 0.3579 ms |   1.00 |          0.00 |  950.0000 |         - |         - |   4.91 MB |
     'DecodeJpegMultiple - ImageSharp' | SmallImagesOnly |  77.6312 ms | 0.2488 ms |   1.85 |          0.03 | 1600.0000 | 1383.3333 | 1308.3333 |  26.51 MB |
 'DecodeJpegMultiple - System.Drawing' | SmallImagesOnly |  41.9075 ms | 0.6030 ms |   1.00 |          0.00 |         - |         - |         - | 489.48 kB |
```

**After:**
```
                                Method |            Mode |        Mean |    StdDev | Scaled | Scaled-StdDev |     Gen 0 |     Gen 1 |     Gen 2 | Allocated |
-------------------------------------- |---------------- |------------ |---------- |------- |-------------- |---------- |---------- |---------- |---------- |
     'DecodeJpegMultiple - ImageSharp' |             All | 315.6664 ms | 1.1802 ms |   2.69 |          0.04 | 1391.6667 | 1329.1667 | 1329.1667 |  52.12 MB |
 'DecodeJpegMultiple - System.Drawing' |             All | 117.3956 ms | 1.8571 ms |   1.00 |          0.00 | 1104.1667 |         - |         - |    5.4 MB |
     'DecodeJpegMultiple - ImageSharp' | LargeImagesOnly | 229.6526 ms | 1.6377 ms |   3.07 |          0.05 |         - |         - |         - |  28.32 MB |
 'DecodeJpegMultiple - System.Drawing' | LargeImagesOnly |  74.9007 ms | 1.0952 ms |   1.00 |          0.00 |  950.0000 |         - |         - |   4.91 MB |
     'DecodeJpegMultiple - ImageSharp' | SmallImagesOnly |  79.3212 ms | 0.2285 ms |   1.90 |          0.02 |  729.1667 |  729.1667 |  729.1667 |  23.78 MB |
 'DecodeJpegMultiple - System.Drawing' | SmallImagesOnly |  41.8073 ms | 0.5584 ms |   1.00 |          0.00 |         - |         - |         - | 489.48 kB |
```


src/ImageSharp/Formats/Jpg/Components/Decoder/YCbCrImage.cs
```diff
@@ -2,41 +2,51 @@
...
+    using System;
+    using System.Buffers;
+
     /// <summary>
     /// Represents an image made up of three color components (luminance, blue chroma, red chroma)
     /// </summary>
-    internal class YCbCrImage
+    internal class YCbCrImage : IDisposable
     {
+        // Complex value type field + mutable + available to other classes = the field MUST NOT be private :P
```
olivif: You can make it a property, I think that's what it wants :)

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
```diff
@@ -18,237 +17,154 @@ namespace ImageSharp.Formats
...
         /// </summary>
-        private Bits bits;
+        private JpegPixelArea blackImage;
 
         /// <summary>
-        /// The image width
+        /// A grayscale image to decode to.
         /// </summary>
-        private int imageWidth;
+        private JpegPixelArea grayImage;
```
JimBobSquarePants: This is interesting. Some neat refactoring undoing the existing mess!

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
```diff
@@ -18,237 +17,154 @@ namespace ImageSharp.Formats
...
-        private byte[] blackPixels;
+        public JpegDecoderCore()
+        {
+            this.HuffmanTrees = HuffmanTree.CreateHuffmanTrees();
+            this.QuantizationTables = new Block8x8F[MaxTq + 1];
+            this.Temp = new byte[2 * Block8x8F.ScalarCount];
+            this.ComponentArray = new Component[MaxComponents];
+            this.ProgCoeffs = new Block8x8F[MaxComponents][];
+            this.Bits = default(Bits);
+            this.Bytes = Bytes.Create();
```
JimBobSquarePants: Lazy loading of the Bytes.Buffer?

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
```diff
@@ -18,237 +17,154 @@ namespace ImageSharp.Formats
...
         /// <summary>
-        /// The array of keyline pixels in a CMYK image
+        /// Initializes a new instance of the <see cref="JpegDecoderCore" /> class.
         /// </summary>
-        private byte[] blackPixels;
+        public JpegDecoderCore()
+        {
+            this.HuffmanTrees = HuffmanTree.CreateHuffmanTrees();
+            this.QuantizationTables = new Block8x8F[MaxTq + 1];
+            this.Temp = new byte[2 * Block8x8F.ScalarCount];
```
JimBobSquarePants: That's only 128 bytes. Should we just use the ArrayPool?

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
```diff
@@ -1337,902 +766,688 @@ private void ConvertFromYCbCr<TColor>(int width, int height, Image<TColor> image
...
-            int bx, by, blockCount = 0;
-
-            Block8x8F b = default(Block8x8F);
-            Block8x8F temp1 = default(Block8x8F);
-            Block8x8F temp2 = default(Block8x8F);
-
-            UnzigData unzig = UnzigData.Create();
-
-            int* unzigPtr = unzig.Data;
+            this.ReadFull(this.Temp, 0, 2);
```
JimBobSquarePants: The `remaining` parameter is only used for precondition checks. Really this line should be. `this.ReadFull(this.Temp, 0, remaining); `

src/ImageSharp/Formats/Jpg/Components/Decoder/JpegScanDecoder.cs
```diff
@@ -0,0 +1,755 @@
...
+{
+    using System;
+    using System.Runtime.CompilerServices;
+    using System.Runtime.InteropServices;
+
+    /// <summary>
+    /// Encapsulates the impementation of Jpeg SOS decoder.
+    /// See JpegScanDecoder.md!
+    /// </summary>
+    internal unsafe struct JpegScanDecoder
```
JimBobSquarePants: Nice work!!

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
```diff
@@ -2243,18 +1458,10 @@ internal class MissingFF00Exception : Exception
...
-        /// Thrown when the decoder encounters an EOF marker without a proceeding EOI (End Of Image) marker
-        /// </summary>
-        internal class EOFException : Exception
-        {
-        }
-
-        /// <summary>
         /// The short huffman data exception.
         /// </summary>
         private class ShortHuffmanDataException : Exception
```
JimBobSquarePants: This doesn't seem to be in use any more.

JimBobSquarePants:
@antonfirsov What's the biggest cause of the allocations? We seem to be using huge amounts of memory compared to System.Drawing.

antonfirsov:
@JimBobSquarePants 
> What's the biggest cause of the allocations? 
`ImageBase<T>.pixelBuffer` not beiing pooled. Nothing, but `Image : IDisposable` can save us. :disappointed:

antonfirsov:
@JimBobSquarePants @olivif The explanation for the public field stuff:
https://gist.github.com/antonfirsov/ea30e8c68bca20308a97c469929d51bf

Here is my rule!

Precondigions:
- Having a mutable S struct (perf critical code)
- Having a struct as member in an other type B
- The struct member must be accessed yet another type C

Implication:
- The struct **must not be** exposed through a property, need a public member instead!

The only clean & safe alternative to this rule is to not have mutable structs :P 
C#7 `ref` returns can can help on this.

JimBobSquarePants:
Ah you've got me there! No worries.

>ImageBase<T>.pixelBuffer not beiing pooled. Nothing, but Image : IDisposable can save us. 😞

Really? I'm not seeing this with the other formats. I'm assuming it is due to the benchmark running on multiple images?

antonfirsov:
I did benchmarking with dotTrace memory profiler a few weeks ago, everything was pointing towards ImageBase.pixelBuffer.

> I'm assuming it is due to the benchmark running on multiple images?

large images + multiple images = real life scenario. All the other benchmarks are liers! :D

Large images also bring in the LOH fragmentation issue :cry:

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
antonfirsov: I like explicit things better, when possible. Lazy loads are for the case, where you don't want to think about lifecycle, letting it "just happen" :smile:

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
antonfirsov: It really doesn't matter with the frequency `JpegDecoderCore` instances are created. Actually this should be an on-stack buffer in the future. Don't want to early-optimize this now.

src/ImageSharp/Formats/Jpg/JpegDecoderCore.cs
antonfirsov: Kill 'Em All!

JimBobSquarePants:
I'll merge this now then continue our conversation. 

>Large images also bring in the LOH fragmentation issue 😢
You're gonna get LOH framentation with image processing anyway you cut it since you need to pin the array for processing. 

`IDisposable` could be a possibility but it introduces a lot of complexity throughout the API. For example, creating and assigning new `TColor` arrays to images like in `Crop`. We'd have to rent the array from within the processor but return it when the image is disposed. That feels spaghetti to me.

antonfirsov:
But it's our internal code! We are hardcore crazy people, we can manage it! :laughing: 

In my opinion the only tradoff exist between: keeping the public API as simple as possible --VS--maximizing performance.

JimBobSquarePants:
That's the argument in a nutshell. `IDisposable` isn't difficult (or certainly shouldn't be) but I've seen it misused so many times it bothers me. (Why ImageProcessor exists essentially). 

I may have  a play around with a branch to see what I come up with.

antonfirsov:
- How about an optional `PooledImage<TColor> : IDisposable`?
- Or an optional `ArrayPool<TColor>` parameter?
- Or `ImageBase<TColor> : IDisposable` with a finalizer that helps out noob users 90% of the time?

antonfirsov:
Or an
```
interface IPooledImageFactory
{
    Image<TColor> Create<TColor>( .... ); // x N-constructors
    void Release(IImage<TColor> image);
}
```

JimBobSquarePants:
I think `IDisposable` would be the last complicated approach. I'll try to have a look at that today.

>ImageSharp arrays are not always pinned. Only when you Lock()them for a PixelAccessor.

Yeah but you should pin a large array if you are doing unsafe manip. I can't think of a better way to do what we are doing just now.

PR #63, opened 2016-12-29
Fixing equality. Fixing almost equality. Adding tests. 

olivif:
One note on the bug fix for almost equality, without an inclusive <= precision there was an odd case where precision is 0f, and the objects are equal, but not almost equal :smile:


src/ImageSharp/Colors/Spaces/CieLab.cs
```diff
@@ -161,17 +161,17 @@ public override bool Equals(object obj)
         /// <inheritdoc/>
         public bool Equals(CieLab other)
         {
-            return this.AlmostEquals(other, Constants.Epsilon);
+            return this.backingVector.Equals(other.backingVector);
```
dlemstra: Are you sure this change is correct? Maybe @JimBobSquarePants can comment if the `AlmostEquals` was intended behavior?

JimBobSquarePants: Yeah happy to change this. As @antonfirsov noted correctly [yesterday](https://github.com/JimBobSquarePants/ImageSharp/pull/60#issuecomment-269468011) `Equals` and `AlmostEquals` should really be separate. My original code was buggy.

JimBobSquarePants:
@olivif Good fix for `<=` Definitely a bug.

Issue #64, opened 2016-12-29
Split ImageSharp into multiple projects


tocsoft:
ImageSharp should be split into multiple projects so you only have to pay for what you need.

- **ImageSharp**
This would contain the Image classes, Colors, Primitives, Bootstrapper, IImageFormat interface, and other core functionality.
- **ImageSharp.Formats.Jpeg**
The jpeg decoder/encoder (registered via Bootstrapper)
- **ImageSharp.Formats.Png**
The png decoder/encoder (registered via Bootstrapper)
- **ImageSharp.Formats.Gif**
The gif decoder/encoder (registered via Bootstrapper)
- **ImageSharp.Formats.Bmp**
The bmp decoder/encoder (registered via Bootstrapper)
- **ImageSharp.Processing** 
Contains methods like Resize, Crop, Skew, Rotate - Anything that alters the dimensions of the image.
Contains methods like Gaussian Blur, Pixelate, Edge Detection - Anything that maintains the original image dimensions.
- **ImageSharp.Metadata**
EXIF, XMP and other classes for editing image metadata.
- **ImageSharp.Drawing**
Brushes and various drawing algorithms.
- **ImageSharp.Drawing.Text**
Various text drawing algorithms.

All packages will live in there own namespaces except the various formats that will share the `ImageSharp.Formats` namespace.

All extension methods targeting `Image<TColor>` will live in the global `ImageSharp` namespace no matter which project they belong to.

~All packages to be versioned and release together under a common version number thus if a change happens in `ImageSharp.Formats.Gif` then `ImageSharp.Drawing` version number will be bumped so that all packages from a single build will have a single common number.~ *This is likely to not be the case TBC*

origin discussion here: https://gitter.im/ImageSharp/General?at=584378420da034021b65da62

This is all open to discussion and this issue is here to permanently capture any decisions made around this. 

EDIT: Updated based on feedback.

dlemstra:
I wonder if we should always have them under the same version number. I think we only need to bump versions when something has changed in that area.

p.s. I wonder if we should split up the text drawing in a separate library but maybe we are taking a step to far then.

JimBobSquarePants:
>ImageSharp.Transforms
Contains methods like Resize, Crop, Skew, Rotate - Anything that alters the dimensions of the image.
ImageSharp.Effects
Contains methods like Gaussian Blur, Pixelate, Edge Detection - Anything that maintains the original image dimensions.

[...]

PR #65, opened 2016-12-29
Reducing allocations in color spaces. 

olivif:
Making some constants static to reduce the number of allocations.

### DIFF
--- a/src/ImageSharp/Colors/Spaces/Bgra32.cs
+++ b/src/ImageSharp/Colors/Spaces/Bgra32.cs
@@ -20,6 +20,16 @@ namespace ImageSharp.Colors.Spaces
         public static readonly Bgra32 Empty = default(Bgra32);
 
         /// <summary>
+        /// Min range used for clamping
+        /// </summary>
+        private static readonly Vector4 VectorMin = Vector4.Zero;
+
+        /// <summary>
+        /// Max range used for clamping
+        /// </summary>
+        private static readonly Vector4 VectorMax = new Vector4(255);
+
+        /// <summary>
         /// The backing vector for SIMD support.
         /// </summary>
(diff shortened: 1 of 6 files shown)


antonfirsov:
@olivif these constants seem to be quite common, I'm using them same ones in Jpeg across multiple classes. Don't we need a common place for them in `ImageSharp/Numerics` or `ImageSharp/Colors`? @dlemstra @JimBobSquarePants @tocsoft Your thoughts?

tocsoft:
Most of these constants seem to be to be very specific to the particular color space they are declared in, I would probably leave these where they are. 

I think the only ones that make me stop and question is the `VectorX.Zero` ones but at the same time I can see it makes the code more readable and consistent across color spaces so I would be inclined to let them pass.

olivif:
Yes, I considered moving them out but I had the same thought as @tocsoft, they are very specific to the class.

antonfirsov:
Ok, actually only `new Vector4(255)` seems to be super-common, maybe it doesn't worth to create a utility class only for this.

PS:
@olivif this change does not reduce number of allocations, because `Vector4` is a value type. The code looks definitely better, but I'm not sure about the effects on performance, they can be even negative! It could be tested by inlining the constants in [DCT.cs](https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/Formats/Jpg/Components/DCT.cs).

JimBobSquarePants:
What's the current opinion on this? I'll need to be merged before #64

antonfirsov:
I just wanted to clear some misunderstandings. The code is definitely better, I think we should merge it now!

olivif:
thanks for merging guys! you are right on the allocations @antonfirsov, forgot Vector3/4 are structs 😄

PR #68, opened 2016-12-30
Add ability to set custom Bootstrapper.

JimBobSquarePants:
This allows the setting of custom formats and parallel processing rules.

Image formats are no longer set by default. Registration has to take place on startup. (This facilitates splitting up the format libraries. See #64 )


tocsoft:
Should bootstrapper be renamed Configuration now it's being passed into Image?

JimBobSquarePants:
@tocsoft Probably... Can't get the tests to pass on AppVeyor just now, everything works locally but I can't figure out from the limited stacktrace what is going wrong.

tocsoft:
I'll take a look for you now to see if i can't figure it out the build issues... what I find odd is locally i'm only seeing 378 tests but on the build server i'm seeing 1309.. wonder why that is.

JimBobSquarePants:
Cheers! Not sure why those numbers are different either.

src/ImageSharp/Quantizers/Wu/WuQuantizer.cs
```diff
@@ -773,7 +773,7 @@ private QuantizedImage<TColor> GenerateResult(PixelAccessor<TColor> imagePixels,
             Parallel.For(
                 0,
                 height,
-                Bootstrapper.ParallelOptions,
+                Bootstrapper.Default.ParallelOptions,
```
tocsoft: the Bootstrapper need passing along with the `PixelAccessor` we should only rely on `Bootstrapper.Default` int he `Image` constructor everywhere else should be explicitly passed.

src/ImageSharp/Image/ImageBase{TColor}.cs
```diff
@@ -25,20 +25,28 @@ public abstract class ImageBase<TColor> : IImageBase<TColor>
         /// <summary>
         /// Initializes a new instance of the <see cref="ImageBase{TColor}"/> class.
         /// </summary>
-        protected ImageBase()
+        /// <param name="bootstrapper">
+        /// The bootstrapper providing initialization code which allows extending the library.
+        /// </param>
+        protected ImageBase(Bootstrapper bootstrapper = null)
```
tocsoft: do we want a version that doesn't force having a height/width set?

src/ImageSharp/Image.cs
```diff
@@ -18,7 +18,11 @@ public class Image : Image<Color>
         /// <summary>
         /// Initializes a new instance of the <see cref="Image"/> class.
         /// </summary>
-        public Image()
+        /// <param name="bootstrapper">
+        /// The bootstrapper providing initialization code which allows extending the library.
+        /// </param>
+        public Image(Bootstrapper bootstrapper = null)
```
tocsoft: do we want a version that doesn't force having a height/width set?

tocsoft:
😞 nope sorry can't figure out whats happening.

What doesn't help is the way we loop the source images inside the test instead of passing them in, if you knew exactly which source image was failing it might help.

My only thought is you somehow have a 1x1 image being used when compiled by appveyor, that would cause your `width /2` to return `0` but I can't see how that would be the case, or how your changes might cause that to happen. 

Maybe it was a bad checkout in appveyor and it corrupted an image, maybe try forcing a fresh build to see if that works?

src/ImageSharp/Bootstrapper.cs
```diff
@@ -16,54 +16,48 @@ namespace ImageSharp
...
     /// </summary>
-    public static class Bootstrapper
+    public class Bootstrapper
     {
         /// <summary>
-        /// The list of supported <see cref="IImageFormat"/>.
+        /// A lazily initialized bootstrapper default instance.
         /// </summary>
-        private static readonly List<IImageFormat> ImageFormatsList;
+        private static readonly Lazy<Bootstrapper> Lazy = new Lazy<Bootstrapper>(() => new Bootstrapper());
```
tocsoft: This shouldn't be lazy, `Bootstapper` isn't expensive to construct, and we are more likely than not going to want it than not..

JimBobSquarePants:
Fixed it, the initializer code was never getting called and the image class was not throwing an error when there were no formats. All fixed now and ready for further review.

src/ImageSharp/Quantizers/Wu/WuQuantizer.cs
JimBobSquarePants: Yup. Will have a look at that.

tocsoft: looks like all the `Quantizer`s will probably need tweeking to pass along the bootstrapper, the ` IQuantizer<TColor>.Quantize(..)` method looks like the best place to me, and then proxied on to the `QuantizedImage` constructor.

src/ImageSharp/Image/ImageBase{TColor}.cs
JimBobSquarePants: This is required for `Image<TColor>(Stream stream)`

src/ImageSharp/Image.cs
JimBobSquarePants: No, I've removed it now and the same in `ImageFrame`

src/ImageSharp/Image/ImageBase{TColor}.cs
tocsoft: ah ok, that's fine then.

tests/ImageSharp.Tests/BootstrapperTests.cs
```diff
@@ -52,15 +52,15 @@ public void AddImageFormatGuardNull()
...
 
             var format = new TestFormat();
             format.Decoder = null;
 
             exception = Assert.Throws<ArgumentNullException>(() =>
             {
-                Bootstrapper.AddImageFormat(format);
+                Bootstrapper.Default.AddImageFormat(format);
             });
             Assert.Contains("decoder", exception.Message);
```
antonfirsov: Is it really a robust solution to have this kind of "stringly typed" assertions? Instead of these long test cases wouldn't it be better to have multiple short AAA-style tests?

tests/ImageSharp.Tests/TestFile.cs
```diff
@@ -2,31 +2,56 @@
...
+        /// </summary>
         private readonly string file;
 
+        /// <summary>
+        /// Initializes static members of the <see cref="TestFile"/> class.
+        /// </summary>
+        static TestFile()
+        {
+            // Register the individual image formats.
+            // TODO: Is this the best place to do this?
```
antonfirsov: For bootstrapping the test framework, I think it's good enough for now.
However, we need to find a more straightforward way for registration when #64 is finished. I think most of the users will vote for a `.RegisterAllKnownFormats()` extension method. (No idea how could we make it)

tocsoft: you would have a `.RegisterXXXFormat()` in each format project, and we could create a combined project that has a dependency on all our core formats that has a single `.RegisterCoreFormats()` extension method.

tests/ImageSharp.Tests/BootstrapperTests.cs
JimBobSquarePants: I didn't write the original tests so can't really comment on them.

src/ImageSharp/Bootstrapper.cs
```diff
@@ -16,54 +16,48 @@ namespace ImageSharp
...
-            };
-            SetMaxHeaderSize();
-            ParallelOptions = new ParallelOptions { MaxDegreeOfParallelism = Environment.ProcessorCount };
-        }
+        private readonly List<IImageFormat> imageFormatsList = new List<IImageFormat>();
+
+        /// <summary>
+        /// Gets the default <see cref="Bootstrapper"/> instance.
+        /// </summary>
+        public static Bootstrapper Default { get; } = Lazy.Value;
```
tocsoft: you you really want/need the lazy you are removing the effect here, by immediately invoking the value... just drop the Lazy part and new it up directly here.

tests/ImageSharp.Tests/TestFile.cs
antonfirsov: Yeah, It's pretty straightforward to do this, but having an extra dll sitting on all of the other dll-s is raping the nice modularization system :D
I would rather use :bat:*Reflection*:bat: for this. (We are talking about an optional utility class!)

src/ImageSharp/Bootstrapper.cs
JimBobSquarePants: Lazy is threadsafe, static Bootstrapper property newing up isn't see http://csharpindepth.com/Articles/General/Singleton.aspx

tests/ImageSharp.Tests/TestFile.cs
JimBobSquarePants: Reflection requires looking through every dll in the bin folder, anything in App_Code and elsewhere. It really slows down startup, especially if someone has installed frameworks like DevExpress or Telerik. Have been there and done it with ImageProcessor. You do not want this.

dlemstra:
I don't understand why we would need to be able to inject the bootstrapper. I even remember someone saying people who want this should bugger off: https://github.com/JimBobSquarePants/ImageSharp/issues/64#issuecomment-269628881. I think this adds extra complexity to the library and it is a feature that will probably be never used.

src/ImageSharp/Bootstrapper.cs
JimBobSquarePants: I'm using Lazy for both thread safety and lazy initialization. Even a static property isn't initialized until the class is called.

src/ImageSharp/Bootstrapper.cs
tocsoft: ok...that fine then, all go to go then for me.

tests/ImageSharp.Tests/TestFile.cs
antonfirsov: ... and it can introduce unwanted asmloadexceptions. I don't want this, I want something smarter :) Will think about this!

antonfirsov:
@dlemstra I had sooo many issues with configuration systems (& other solutions) relying on global variables. Most modern frameworks tend to avoid this. Maybe you're right and for an image processing library it's YAGNI, but I think it's still worth to follow the best practices. If ImageSharp will be the NextGen(TM) image processing library of the .NET core world, I think some large and complex applications will make use of this feature.

The complexity is mostly about having an optional argument for Image constructors, it's not that hard to manage.

dlemstra:
I had a chat with @JimBobSquarePants about this and he came up with an example where one part of the process was only allowed to read only jpeg (e.g. upload) and another parts was allowed to write all of the formats. At first sight this looked like a feature that would not be used but I think there can actually be a real world usage for this. I was just a bit worried that we would need to pass it around all the time. But if we can limit the use of the bootstrapper to just the image it should be okay.

tocsoft:
once this is merged I'll throw up a PR for the project splitting, its kinda dependent on the Configuration changes otherwise we wouldn't have been able to split off the formats from the core as easily.

Issue #66, opened 2016-12-29
Common .NET 4.6 Sandbox project for performance profiling


antonfirsov:
Most of our discussions are happening below separate issue/PR topics now. With the current team activity it seems to be more efficient to me, so I decided to open an issue for this topic, hope it's fine.

I think if we want a performance-aware team, everyone should (must!) be able to profile code. We can't wait for better netcore tooling, we need it now! 
It's actually very easy with a classic .NET project referencing ImageSharp as a .NET 4.6 dll. With MSBUILD `<Link>` [it's possible](https://github.com/antonfirsov/ImageSharp/blob/jpeg-optimizations-experimental/tests/ImageSharp.Tests46/ImageSharp.Tests46.csproj) to construct such a project adding just ~5 files!

Why don't have it on the common master branch? :)

antonfirsov:
@JimBobSquarePants any initial, quick feedback? :)

tocsoft:
👍 go for it.

antonfirsov:
I would wait for the outcome of @JimBobSquarePants -s fight against VS2017.

JimBobSquarePants:
VS2017 won't be happening anytime soon. I just uninstalled it. Carry on....

PR #69, opened 2017-01-01
Split ImageSharp into multipe projects

### LINKED ISSUE #64: Split ImageSharp into multiple projects

tocsoft:
This is a large change PR as I've had to move a lot of stuff around and re-namespaced the processing stuff... this is probably going to require a lot of cleanup and people probably wont be too pleased where I've dumped stuff in terms of folder names etc  but I want to get the project split quickly so we don't block each other with our changes, and we can clean it back up now its spilt and building correctly.

The way it works, RE versioning/building, is I've added a new dotnet projected into the build folder all living in a single Program.cs file this is where all the heavy lifting magic happens, this processes the project files determines how may commits each project (and its dependencies) there have been since the projects version number was updated. We then use a combination of number of commits since version change+ the branch name, to update the version number for that project. Then we just build/package them all before reverting the project.json files back in case you want to run the build process locally.

New projects will be automatically discovered by the build process so all you will need to do is add it into the `src` folder and it will be built, versioned  and packaged.

In conclusion ImageSharp will now have the following Package/Projects;
- **ImageSharp**
- **ImageSharp.Drawing**
- **ImageSharp.Formats.Png**
- **ImageSharp.Formats.Bmp**
- **ImageSharp.Formats.Gif**
- **ImageSharp.Formats.Jpeg**
- **ImageSharp.Formats** - this is a helper project that just takes a dependency on the all the core formats and exposes a single helper method for registering all the formats into a `Configuration` instance. If we think this is overkill then we can drop it but I defiantly think most people will just want all the formats registered 90%of the time and will want a nice easy way for doing it.

Fixes #64


tocsoft:
I've pinged you all for review as this is quite a large change, and feel we all need to have eyes on it.

Also keep in mind I feel we don't need the PR perfect exactly, as we can do subsequent updates do bash it into a better state.

src/ImageSharp/Properties/AssemblyInfo.cs
```diff
@@ -37,4 +37,4 @@
 // Ensure the internals can be tested.
 [assembly: InternalsVisibleTo("ImageSharp.Benchmarks")]
 [assembly: InternalsVisibleTo("ImageSharp.Tests")]
-[assembly: InternalsVisibleTo("ImageSharp.Tests46")]
+[assembly: InternalsVisibleTo("ImageSharp.Tests46")]
```
antonfirsov: Ooops .. I forgot to remove this from my previous PR. I will change this to `ImageSharp.Sandbox46` with #66

Andy-Wilkinson:
Just a comment to flag that the Travis CI build is only building the main ImageSharp assembly. Should be able to use the nice versioning & build system @tocsoft has built since `dotnet` is cross-platform, but I guess it will need corresponding Linux shell scripts.

Probably one to fix in a subsequent pull request - better to get this one merged and working with Appveyor first.

appveyor.yml
```diff
@@ -17,31 +17,8 @@ install:
...
-    npm install
-
-    node appveyor-project-version-patch.js
-
-    cd..
-
 build_script:
-- cmd: >-
-    dotnet pack src\ImageSharp --configuration Release --output "artifacts\bin\ImageSharp"
+- cmd: build.cmd
```
dlemstra: I also like it better to have a build.cmd file that builds the projects No need to edit appveyor.yml and easier to test locally.

build/Program.cs
```diff
@@ -0,0 +1,331 @@
...
+                var maxLength = 20; // dotnet will fail to populate the package if the tag is > 20
+                maxLength -= rootSpecialVersion.Length; // this is a required tag
+                maxLength -= 7; // for the counter and dashes
+                
+                if(branch.Length > maxLength)
+                {
+                    branch = branch.Substring(0, maxLength);
+                }
+                
+                return $"{version}{rootSpecialVersion}{branch}-{this.CommitCountSinceVersionChange:00000}";
```
dlemstra: Not sure if we need this kind of fancy version number stuff. The build should probably only just add a build number to the current version.

src/ImageSharp.Drawing/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Resources;
+using System.Runtime.CompilerServices;
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("ImageSharp")]
+[assembly: AssemblyDescription("A cross-platform library for processing of image files; written in C#")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("James Jackson-South")]
```
antonfirsov: The metadata/copyright parts are pretty the same for all the AssemblyInfo.cs-s. Don't we need a solution level `AssemblyInfo.Common.cs` for these parts? Is it possible with project.json?

build/build.cmd
```diff
@@ -0,0 +1,18 @@
+@echo Off
+set buildRoot="%cd%"
+
+ECHO restoring packages
```
dlemstra: My OCD says we need a capital letter here 😛

build/build.xproj
```diff
@@ -0,0 +1,25 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
dlemstra: Do we need this xproj file? It will be deprecated in a couple months.

antonfirsov:
How should we manage common utility classes/methods that:
- Are practical to be exposet to all ImageSharp subprojects and used in common, but
- Are part of of implementation details, so it's better to hide them on ImageSharp public API-s?

Is it a good practice to utilize `[InternalsVisibleTo("...")]`  for this?

src/ImageSharp.Drawing/project.json
```diff
@@ -0,0 +1,82 @@
...
+    "projectUrl": "https://github.com/JimBobSquarePants/ImageSharp",
+    "licenseUrl": "http://www.apache.org/licenses/LICENSE-2.0",
+    "iconUrl": "https://raw.githubusercontent.com/JimBobSquarePants/ImageSharp/master/build/icons/imagesharp-logo-128.png",
+    "requireLicenseAcceptance": false,
+    "repository": {
+      "type": "git",
+      "url": "https://github.com/JimBobSquarePants/ImageSharp"
+    },
+    "tags": [
+      "Image Resize Crop Gif Jpg Jpeg Bitmap Png Core"
```
dlemstra: We should probably have a look at the tags for each project, we could do that later though. Same goes for the description.

src/ImageSharp.Drawing/project.json
```diff
@@ -0,0 +1,82 @@
...
+      "type": "build"
+    },
+    "System.Buffers": "4.0.0",
+    "System.Numerics.Vectors": "4.1.1",
+    "System.Runtime.CompilerServices.Unsafe": "4.0.0"
+  },
+  "frameworks": {
+    "netstandard1.1": {
+      "dependencies": {
+        "System.Collections": "4.0.11",
```
dlemstra: I also think we need to check if we still need all these dependencies. Maybe some of the projects no longer require all the dependencies?

src/ImageSharp.Formats.Gif/ImageExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+            where TColor : struct, IPackedPixel, IEquatable<TColor>
+                        => source.Save(stream, new GifEncoder { Quality = quality });
+
+        /// <summary>
+        /// To the frame.
+        /// </summary>
+        /// <typeparam name="TColor">The type of the color.</typeparam>
+        /// <param name="source">The source.</param>
+        /// <returns>The frame</returns>
+        internal static ImageFrame<TColor> ToFrame<TColor>(this ImageBase<TColor> source)
```
dlemstra: I wonder if we should leave these two in the primary package and maybe even make them public? I wrote these myself and that are really tiny so maybe we should refactor them out or add these methods to the public interface. Your thoughts @JimBobSquarePants ?

dlemstra:
@antonfirsov  Maybe we should make some of these public then? We should be careful with that though. Another solution could be including (link) them in each project where we need them.

src/ImageSharp.Formats/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// Extension methods for the <see cref="Configuration"/> type.
+    /// </summary>
+    public static partial class ConfigurationExtensions
+    {
+        /// <summary>
+        /// Adds the common formats, PNG, JPEG, GIF and BMP.
+        /// </summary>
+        /// <param name="config">The config.</param>
+        /// <returns>The config</returns>
+        public static Configuration AddCommonFormats(this Configuration config)
```
antonfirsov: It's very practical to have this method, but I don't like the idea of having a whole package for this, messing up the dependency graph.
I'd rather define a simple, convention based utility which is placed in `ImageSharp.dll`, and uses reflection without enumerating / loading all the dll-s, only looks for ImageSharp core modules.

src/ImageSharp/Common/Helpers/Guard.cs
```diff
@@ -14,7 +14,7 @@ namespace ImageSharp
     /// Provides methods to protect against invalid parameters.
     /// </summary>
     [DebuggerStepThrough]
-    internal static class Guard
+    public static class Guard
```
dlemstra: After seeing so many pieces of shared (duplicated) code and the comment from @antonfirsov and now making something public that should not be public I do think we should use `InternalsVisibleTo` for the Helpers and keep them all together inside the ImageSharp project.

tests/ImageSharp.Tests/project.json
```diff
@@ -20,9 +20,24 @@
     }
```
dlemstra: Happy that you did not split up the test project. Maybe others feel different but I think we should keep all the tests in one project.

src/ImageSharp.Formats/ConfigurationExtensions.cs
tocsoft: I'm happy with going another way with this, figured I would chuck it in and it could then be used as a bouncing off point tom do the rest... 👍 for only looking for core modules, we won't have the issue with loading everything up then and we wouldn't even need to scan the assembly for implementations of an interface as we should know exactly which classes in which assemblies we want.

src/ImageSharp.Drawing/DrawImage.cs
```diff
@@ -7,7 +7,7 @@ namespace ImageSharp
 {
     using System;
 
-    using Processors;
+    using Drawing.Processors;
```
dlemstra: Shouldn't we rename this to Drawing.Processing?

build/build.xproj
tocsoft: I found it made it easier to add to the solution, and play better with VS.

src/ImageSharp.Drawing/project.json
tocsoft: totally agree, I think all the project files will defiantly need a sanity pass.

tocsoft:
@Andy-Wilkinson thanks for pointing out the travis build it now build them all, I think we can migrate travis over to a better build script later.

Andy-Wilkinson:
Great... Ultimately it would be nice for the Travis build to run unit tests etc., but that is out of scope for this PR.

build/Program.cs
tocsoft: This means that branches and PRs wont ever produce version numbers that match master ones, if you allow branches to be able to produce packages that could also be produced by master you might end up with 2 different builds producing the same version number and causing accidental confusion if anyone finds/uses the wrong file.

src/ImageSharp/Common/Helpers/Guard.cs
tocsoft: yeah my bad, that shouldn't be public, I had an attempt to make it compile that needed it, must have forgot to change it back.

tocsoft:
UPDATE: 
- Travis now fully builds and tests also runs the unit tests.
- dropped `ImageSharp.Formats` in favour of a reflection based autoload on the default Configuration... we will only auto load out core formats preventing users who don't want to have to think about formats to just be able to reference the formats they want and them just to work. (saves having to find the "right" place to register the formats)

build/Program.cs
JimBobSquarePants: That should never happen. I need to split up the branches so there is a dev and master. Dev builds should always have a build suffix. Alpha, beta, RC etc

tests/ImageSharp.Tests/project.json
JimBobSquarePants: Definitely keep the tests together. I want to throw Stylecop at them though so they are properly formatted and annotated.

src/ImageSharp.Formats.Gif/ImageExtensions.cs
JimBobSquarePants: I didn't know these existed and now I see them I certainly don't like the first one. `ImageBase` shouldn't have a and extension method converting it to an `ImagFrame` as it should have no idea what an `ImageFrame` should be, the calling code should use the constructor. 

Not convinced by the second method etiher. Again the constructor could have been used.

src/ImageSharp/Configuration.cs
```diff
@@ -34,6 +34,30 @@ public class Configuration
...
+        public Configuration()
+            : this(false)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Configuration" /> class.
+        /// </summary>
+        /// <param name="autoloadWellknownFormats">if set to <c>true</c> [autoload wellknown formats].</param>
+        internal Configuration(bool autoloadWellknownFormats)
```
antonfirsov: :+1:  for the idea of having this "smart autoconfiguration"
I think it would be better to have this code in a  static `Configuration.CreateDefaultInstance()` method instead of a parametric base constructor.

antonfirsov:
Shouldn't `Configuration.SyncRoot` be an instance member now? 
OFF: Is it really not allowed to comment on unchanged lines in a PR or did I miss something?

src/ImageSharp/Configuration.cs
```diff
@@ -70,6 +94,32 @@ public void AddImageFormat(IImageFormat format)
...
+            if (type != null)
+            {
+                IImageFormat format = Activator.CreateInstance(type) as IImageFormat;
+                if (format != null
+                    && format.Encoder != null
+                    && format.Decoder != null
+                    && !string.IsNullOrEmpty(format.MimeType)
+                    && format.SupportedExtensions?.Any() == true)
+                {
+                    // we can use the locked version as we have already validated in the if.
```
dlemstra: I think it is cleaner to use the AddImageFormat method instead of duplicating the checks.

src/ImageSharp/Configuration.cs
dlemstra: I have to agreed with @antonfirsov on this, cleaner to have a method for this.

src/ImageSharp/Configuration.cs
antonfirsov: @dlemstra these are non-throwing checks, while throws an exception when something is missing. I see no straightforward way to refactor this without catching first-chance exceptions on *normal* execution path.

src/ImageSharp.Formats.Bmp/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Extension methods for the <see cref="Configuration"/> type.
+    /// </summary>
+    public static partial class ConfigurationExtensions
+    {
+        /// <summary>
+        /// Adds the BMP format.
+        /// </summary>
+        /// <param name="config">The config.</param>
+        /// <returns>The config</returns>
+        public static Configuration AddBmpFormat(this Configuration config)
```
JimBobSquarePants: We don't need this nor the others.  The default behaviour is now for `Configuration.Default` to automagically load the types and I want to use the same syntax for adding all image formats.

build/Program.cs
```diff
@@ -0,0 +1,331 @@
...
+using System;
+using System.IO;
+using System.Linq;
+using NuGet.Versioning;
+using System.Collections.Generic;
+using LibGit2Sharp;
+using Newtonsoft.Json;
+using System.Text;
+
+namespace ConsoleApplication
```
JimBobSquarePants: StyleCop please.

What's the output of this process? i.e What version numbers are generated?

build/Program.cs
```diff
@@ -0,0 +1,331 @@
...
+
+        private static void CreateBuildScript(IEnumerable<SourceProject> projects)
+        {
+            var sb = new StringBuilder();
+            foreach (var p in projects)
+            {
+
+                sb.AppendLine($@"dotnet pack --configuration Release --output ""artifacts\bin\ImageSharp"" ""{p.ProjectFilePath}""");
+            }
+            File.Copy("build-inner.cmd", "build-inner.bak", true);
```
JimBobSquarePants: What does build-inner do? The file is empty.

src/ImageSharp.Processing/Transforms/ResizeProcessor.cs
```diff
@@ -3,7 +3,7 @@
 // Licensed under the Apache License, Version 2.0.
 // </copyright>
 
-namespace ImageSharp.Processors
```
JimBobSquarePants: Can we put all these processors back please. I don't understand why they were all moved.

tests/ImageSharp.Tests/project.json
antonfirsov: @JimBobSquarePants do you want to StyleCop test projects with the current SC rules? :cry: 

Keep in mind, that the current **actual** test coverage of this project is very low. The ~80% coverage shown by codecov is brought mostly by "integration" tests executing everything on every image without any automatic assertions. These are **not unit tests**. If you remove this cheat from the coverage metric, the results will be very disappointing.

Unit tests should be small, self-documenting pieces of code with a **very low** barrier to write them, to encourage developers to write more and more of them. By forcing to xml-doc everything, and forcing developers to focus on formatting details, you will discourage people to do TDD!

Just look at [corefx test code](https://github.com/dotnet/corefx/blob/master/src/System.Memory/tests/Span/Slice.cs). There are no xml docs at all. And these are microsoft people spending their **job time** to write them. How can you expect people who are spending their free time for this project to do an extra +30% work to formatting & "documenting" test cases? (Well-named, clean AAA test cases are already documented!)

I'm happy to add some comments on non trivial parts, specially if asked for. But making a general overrestricted rule from this, overseen by dumb robots will be a crime against TDD & will act against the goal of having more real test coverage.

TDD is my main methodology to develop software, and I think I'm not alone with this. I **strongly** disagree with this idea :(

JimBobSquarePants: I guarantee there's some coding standard applied to those tests also. Stylecop or not they need to be cleaner, conform to a basic set of rules and naming standards and be instantly obvious what they are testing.

src/ImageSharp.Drawing/DrawImage.cs
JimBobSquarePants: Nope... Processing is a library name, processors should be anything inheriting ImageProcessor<TColor>

build/Program.cs
tocsoft: it got populated by the build project after it discovered all the things to build... i've figured out a cleaner way for it to do it so this can now be dropped

tocsoft: scratch that I was wrong globing doesn't seem to work with the pack command... the empty file can be dropped but the build project will generate a pack command for each project it discovered in the src folder.

build/Program.cs
tocsoft: added a description in the file to help describe how versioning.

src/ImageSharp.Processing/Transforms/ResizeProcessor.cs
tocsoft: done... I'm sure It had some logic when I moved them... can't remember what it was now 😕

tocsoft:
Think I've covered all the feedback now.

tests/ImageSharp.Tests/project.json
antonfirsov: > they need to be cleaner, conform to a basic set of rules and naming standards and be instantly obvious what they are testing.

Agreed. There are many test classes being non-conform to the basic formatting rules of the library, and this is not good. I just wanted to warn, that simply copy-pasting the default StyleCop behavior to the tests would be harmful to the project. 

Shouldn't we open a discussion about this? Or as an alternative you can just be a Wise and Merciful Dictator, and everything should be OK :laughing:

JimBobSquarePants: Happy to be merciful if it helps productivity. Let's all try to clean up as we update tests and make sure we don't introduce new formatting issues and we'll be fine 😄.

JimBobSquarePants:
I'm tentatively approving this. Can everyone cast one final look at the updates and also read the versioning info in `build/Program.cs` to double check the logic. If we're all happy we can then merge and carry on. @tocsoft You can do the honours 😄 

Great work!!

tocsoft:
Awesome... I'll give everyone till [07:00GMT](http://everytimezone.com/#2017-1-3,1140,c8l) for anyone to raise any final issues and then I'll merge. _(Will merge sooner if I get ok from everyone sooner)_

JimBobSquarePants:
Let's make sure we add the library names to the readme also.

antonfirsov:
On my machine manually executing `build.cmd` results in:
```
CS2001: Source file 'C:\.....\tocsofg.split-projects\build\../src/ImageSharp/stylecop.json' could not be found.
```

But I'm not a build system expert, and it's late here, so if there's no one else to reproduce it, or it was not designed for local execution, don't listen to me.

:+1: for the versioning strategy

tocsoft:
@antonfirsov that was a bug, i'd moved the file and missed a reference... I'm surprised it still built.

tocsoft:
all merged 🎉

PR #71, opened 2017-01-04
Optimise polygon fill

tocsoft:
Significantly improves the speed for filling polygons.

|                                     Method |      Mean |    StdDev | Scaled | Scaled-StdDev |    Gen 0 |    Gen 1 |    Gen 2 | Allocated |
|------------------------------------------- |---------- |---------- |------- |-------------- |--------- |--------- |--------- |---------- |
|          'Fill Polygon' | 7.6604 ms | 0.2209 ms |   1.00 |          0.00 |  62.5000 |  62.5000 |  62.5000 |    2.6 MB |
|            'Optimised Fill Polygon' | 4.4004 ms | 0.1079 ms |   0.57 |          0.02 | 173.9583 | 173.9583 | 173.9583 |   2.59 MB |
| 'Fill Polygon - anti-alias disabled' | 7.5998 ms | 0.2941 ms |   0.99 |          0.05 |  77.2059 |  77.2059 |  77.2059 |   2.59 MB |
|  'Optimised Fill Polygon - anti-alias disabled' | 2.2995 ms | 0.0660 ms |   0.30 |          0.01 | 244.2708 | 243.2292 | 243.2292 |   2.59 MB |


antonfirsov:
Yeahh, well done! Can't wait to review this!

JimBobSquarePants:
Ah great stuff! I'll do a quick review now.

src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
```diff
@@ -32,15 +32,15 @@ public ImageBrush(IImageBase<TColor> image)
         }
 
         /// <inheritdoc />
-        public IBrushApplicator<TColor> CreateApplicator(PixelAccessor<TColor> sourcePixels, RectangleF region)
+        public BrushApplicator<TColor> CreateApplicator(PixelAccessor<TColor> sourcePixels, RectangleF region)
```
JimBobSquarePants: Much prefer seeing the concrete implementation.

src/ImageSharp.Drawing/Paths/InternalPath.cs
```diff
@@ -164,6 +169,45 @@ public PointInfo DistanceFromPath(Vector2 point)
...
+            int position = 0;
+            for (int i = 0; i < polyCorners && count > 0; i++)
+            {
+                int next = i + 1;
+                if (this.closedPath && next == polyCorners)
+                {
+                    next = 0;
+                }
+
+                var point = FindIntersection(this.points[i], this.points[next], start, end);
```
JimBobSquarePants: `var`... Incidentally the rule used for `var` in the Microsoft libraries is that it can only be used when the type is present on the same line. I don't think many people know this.

src/ImageSharp.Drawing/Paths/InternalPath.cs
```diff
@@ -203,6 +247,92 @@ public bool PointInPolygon(Vector2 point)
             return oddNodes;
         }
 
+        private static bool BoundingBoxesIntersect(Vector2 line1Start, Vector2 line1End, Vector2 line2Start, Vector2 line2End)
```
JimBobSquarePants: Can we use the XML docs for private methods also please. Some (many) might see it as overkill but we do most of our really complicated stuff in private methods so it's good to add additional documentation and really helps with intellisense internally

JimBobSquarePants: Oh and all those vars below please.

src/ImageSharp.Drawing/Shapes/Polygon.cs
```diff
@@ -127,5 +135,21 @@ public Vector2[] AsSimpleLinearPath()
         {
             return this.innerPath.Points;
         }
+
+        /// <summary>
+        /// Finds the intersections.
```
JimBobSquarePants: Can we expand this description slightly please. It's all still rocket science to me 🚀

tests/ImageSharp.Tests/Drawing/SolidPolygonTests.cs
```diff
@@ -31,7 +31,7 @@ public void ImageShouldBeOverlayedByFilledPolygon()
             {
                 image
                     .BackgroundColor(Color.Blue)
-                    .FillPolygon(Color.HotPink, simplePath)
+                    .FillPolygon(Color.HotPink, simplePath, new GraphicsOptions(true))
```
JimBobSquarePants: I'm assuming this is just to explicitly test the overload. Default is anti-alias yeah?

src/ImageSharp.Drawing/Processors/FillShapeProcessor.cs
```diff
@@ -61,77 +62,398 @@ protected override void OnApply(ImageBase<TColor> source, Rectangle sourceRectan
...
+            using (BrushApplicator<TColor> applicator = this.fillColor.CreateApplicator(sourcePixels, rect))
             {
                 Parallel.For(
                 minY,
                 maxY,
                 this.ParallelOptions,
                 y =>
                 {
-                    int offsetY = y - polyStartY;
+                    var buffer = arrayPool.Rent(maxIntersections);
```
antonfirsov: Should be safer to move the buffer outside the using block and the loop. It's also wise to have a try {} finally{} around them.

src/ImageSharp.Drawing/Processors/FillShapeProcessor.cs
```diff
@@ -61,77 +62,398 @@ protected override void OnApply(ImageBase<TColor> source, Rectangle sourceRectan
...
+        {
+            if (lo < hi)
+            {
+                int p = PartitionY(data, lo, hi);
+                QuickSortY(data, lo, p);
+                QuickSortY(data, p + 1, hi);
+            }
+        }
+
+        private static void QuickSortX(Vector2[] data, int size)
```
antonfirsov: Cool to have this non-allocating, delegatele-free quicksort!
However, there is some bloated code. @JimBobSquarePants  I think we should consider using T4 for this, or for stuff like `IPackedBytes`. I'm definitely sure, there will be more cases like this!

src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
```diff
@@ -86,7 +86,7 @@ public ImageBrushApplicator(IImageBase<TColor> image, RectangleF region)
             /// <returns>
             /// The color
             /// </returns>
-            public TColor GetColor(Vector2 point)
+            public override TColor GetColor(Vector2 point)
```
antonfirsov: + I think it makes no sense to use `Vector2` here. For brushes you are casting int-s to cast them back to float-s. It's expensive.
+ Need to profile code, but I'm quite sure we can make significant performance gain relatively cheaply by having a batched "big brother" for `GetColor`:
```
public virtual void PopulateRow(int y, int fromX, int toX, Span<TColor> destination)
```
Maybe this story is about a future PR.

src/ImageSharp.Drawing/Processors/FillShapeProcessor.cs
JimBobSquarePants: Missed this. Agreed, should be faster also.

src/ImageSharp.Drawing/Processors/FillShapeProcessor.cs
JimBobSquarePants: Not a fan of T4 to be honest. Another syntax to learn that lingers.

src/ImageSharp.Drawing/Processors/FillShapeProcessor.cs
tocsoft: the buffer is used in parallel... it needs to be rented inside the parallel loop otherwise multiple threads will be competing over the same buffer.

src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
tocsoft: I tried a version with a `PopulateRow(..)` style interface it slowed it down filling `SolidBrush`es in the simple benchmarks I tried... I didn't look too hard at it as I figured that could be a follow up to properly investigate the issue.

tests/ImageSharp.Tests/Drawing/SolidPolygonTests.cs
tocsoft: yes, the default is to anti-alias

src/ImageSharp.Drawing/Paths/InternalPath.cs
tocsoft: yeah my bad, i'm trying to catch myself using `var` but I do miss them.

We should think about trying to find an analyzer that can catch them.

tocsoft:
@JimBobSquarePants I think I've covered all your changes.

src/ImageSharp.Drawing/Processors/FillShapeProcessor.cs
antonfirsov: Maybe this is not the best example to "lobby" for T4. However in more complex cases we can use it as a replacement for OOP & functional abstractions to apply the DRY principle. 

Unfortunately this is the simpliest tool for this in the .NET world at the moment.

PR #75, opened 2017-01-11
Replace floating point color translations in Jpeg format

JimBobSquarePants:
Replaces the floating point calculations with scaled integers and bit-shifting to improve the performance of the jpeg format.

I've included some benchmarks for comparison.

Please have a look at it and test against any images that you know can cause issues. 

Still not happy with that decoder performance on my machine. We'll figure it out though I'm sure. 

### Jpeg Encode

**Before**

Method |      Mean |    StdDev | Scaled | Scaled-StdDev |    Gen 0 | Allocated |
---------------------- |---------- |---------- |------- |-------------- |--------- |---------- |
'System.Drawing Jpeg' | 4.9274 ms | 0.1229 ms |   1.00 |          0.00 | 271.8750 | 894.51 kB |
'ImageSharp Jpeg' | 9.2119 ms | 0.4067 ms |   1.87 |          0.09 |        - |  66.96 kB |

**After**

Method |      Mean |    StdDev | Scaled | Scaled-StdDev |    Gen 0 | Allocated |
---------------------- |---------- |---------- |------- |-------------- |--------- |---------- |
 'System.Drawing Jpeg' | 4.9281 ms | 0.0531 ms |   1.00 |          0.00 | 271.8750 | 894.51 kB |
'ImageSharp Jpeg' | 7.9708 ms | 0.2411 ms |   1.62 |          0.05 |        - |  66.96 kB |

### Jpeg Decode

**Before**

Method |       Mean |    StdDev | Scaled | Scaled-StdDev | Allocated |
---------------------- |----------- |---------- |------- |-------------- |---------- |
'System.Drawing Jpeg' | 11.1743 ms | 0.7354 ms |   1.00 |          0.00 | 260.74 kB |
'ImageSharp Jpeg' | 67.4747 ms | 1.6753 ms |   6.07 |          0.45 |   3.87 MB |

**After**

Method |       Mean |    StdDev | Scaled | Scaled-StdDev | Allocated |
---------------------- |----------- |---------- |------- |-------------- |---------- |
'System.Drawing Jpeg' | 11.3669 ms | 0.7688 ms |   1.00 |          0.00 | 260.74 kB |
'ImageSharp Jpeg' | 63.8410 ms | 0.4163 ms |   5.64 |          0.40 |   3.87 MB |


antonfirsov:
@JimBobSquarePants I still hope it can be done better, maybe using integers is not the best. Want to play a bit with benchmarks first to elaborate the most optimal way. (around 11pm CET)
Is it ok if I push benchmark changes directly to the master?

I know the `(Encode|Decode)JpegMultiple` benchmarks demand more patience, but running them could provide more information on this :smile:

JimBobSquarePants:
Using scaled integers with bit-shifting like this is a pretty standard way to optimize code like this. The original DCT code did this for example. 

Yeah push benchmarks to master if you like.

antonfirsov:
Everything depends on the context. You need to be very cautious with float <-> int, float <-> byte, int <-> byte conversions, because they are quite expensive.

Regarding accuracy: Maybe we could adopt an image comparsion techniques from [MonoGame tests](https://github.com/MonoGame/MonoGame/blob/develop/Test/Framework/Visual/VisualTestFixtureBase.cs) in the future.

JimBobSquarePants:
@antonfirsov True that... I wonder sometimes whether we are, to coin a phrase my mother uses trying to "make a silk purse out of a sows ear". basing our decoder on the Golang implementation. Yeah it works but it's so messy it feels like it has to be fundamentally architecturally flawed. 

I was very briefly looking at the codebase for Google's new jpeg decoder/encoder and it's a joy to read the code in comparison with Golangs. Everything is right where I'd expect it to be. 

https://github.com/google/guetzli

The way they color transform is interesting also. 
https://github.com/google/guetzli/blob/master/guetzli/color_transform.h#L211

I wonder whether we should try and build an alternative as an experiment?

antonfirsov:
I'm also thinking about a full replacement sometimes, and this library looks really nice. I'm not happy with the current results, because on large images System.Drawing is still much more faster. Finding the reason is on my list, hope it's just multithreading.

However doing the "refactor to patterns" approach in the past 2 months, I realized that that if we do it the right way, in the end will be impossible to recognize the original golang mess. Your mom is right, evolution works like this! :smile:

JimBobSquarePants:
I think multithreading will be a patch rather than a solution. We can beat System.Drawing with our Bmp format on a single thread so it should theoretically be possible to do the same (or at least come close with Jpeg) 

The refactoring work you've done so far is amazing but I don't want you to go so far down a path trying to refactor that codebase that it becomes a chore.

antonfirsov:
Improved the benchmarks so they reflect the context of the problem better. Played with other algorithms, none of them can beat the simplest scalar integer approach :smile: 

`Vector<T>` really sucks. It can't even perform a vectorized dot product :cry: I wish I went to bed earlier instead.

I think you should merge your changes!

```
                                                  Method |          Mean |     StdDev | Scaled | Scaled-StdDev |
-------------------------------------------------------- |-------------- |----------- |------- |-------------- |
                             'Floating Point Conversion' |   537.3523 ns |  2.3920 ns |   1.00 |          0.00 |
                        'Simd Floating Point Conversion' |   767.8779 ns |  1.9016 ns |   1.43 |          0.01 |
                             'Scaled Integer Conversion' |   497.1790 ns |  1.9141 ns |   0.93 |          0.01 |
               'Scaled Integer Conversion + Vector<int>' |   604.9561 ns |  1.8976 ns |   1.13 |          0.01 |
 'Scaled Integer Conversion + Vector<int> + Dot Product' | 7,053.9099 ns | 31.2488 ns |  13.13 |          0.08 |
                         'Scaled Integer LUT Conversion' |   715.6920 ns |  1.9436 ns |   1.33 |          0.01 |
```

antonfirsov:
Tested your changes the `*JpegMultiple` benchmarks. The optimization factor is significant. On small images now we are faster than System.Drawing!

[ORIGINAL](Jpeg-ScaledInteger-Optimization-ORIGINAL.md)

[OPTIMIZED](https://gist.github.com/antonfirsov/01743d39d53cb68f7071d935308c8ab7)

Regarding the deep refactor of the Jpeg codec:
I discourage now to start a full port-to-C#-from-scratch approach: at the current state it can bring much more "chore" work than a full clenup of our current codebase. The most (performance) critical part of the codec is the scan encoder/decoder part (DCT algorithms zigzagging etc.). It's now optimized for the CLR, covered with tests, and in the case of the decoder, moved to it's own class (JpegScanDecoder). It's prepared for further optimizations like parallelization.

I'd rather suggest to follow a mixed approach: Maybe we can port an existing library like guetzli (or parts of it), and replace the scan part with our already CLR-optimized one. 
I spent large amounts of time digging for ideas to speedup the DCT part. Have not found anything that could be used with CLR. Maybe there is something "around" the SOS part, I've missed, but unlikely.

However, I will continue my research. Are there any specific features you want to see in our codecs other than ICC profiles and progressive encoding?

JimBobSquarePants:
@anton That's great news on the performance front!

I would certainly also prefer to keep our current work if we can continue to refactor the code to increase performance and reduce overheads and pinch bits from other libraries. We do seem to have a pressing issue with the jpeg encoder though. 

Download and take a look at https://github.com/bleroy/core-imaging-playground and run the LoadResizeSave method in Program.cs (Don't bother running the benchmarks)

Our jpeg output is awful compared to the others even if I set the supsampling to 4:4:4 and the quality to 100% (Oddly 4:4:4 is worse than 4:2:0 which it simply shouldn't be)

I'm thinking of digging out the old encoder I wrote in the [jpeg branch](https://github.com/JimBobSquarePants/ImageSharp/tree/2fb98f9642167e8556638d89b79e9c453950b45a/src/ImageProcessorCore/Formats/Jpg/New) updating the code and having a look at the output quality there and, if the quality is better, attempt to discern what the differences are.

antonfirsov:
Wanted to play with gif octrees and other perf related stuff, but it seems I gonna have a jpeg weekend :D

JimBobSquarePants:
Grim eh? I was pretty gutted to discover this.

antonfirsov:
Okkay I have to confess! 

The quality drop has been introduced to the encoder with the `Block8x8F` refactor :(
Don't know yet, if it's an essential weakness of the floating point FDCT implementation or it's a mistake somewhere around it's application.

I'm going to investigate this issue. If it's not possible to use a floating point SIMD alogorithm without a quality drop, I will refactor the code to use integer blocks again. 

I will also think about the guetzli port.

JimBobSquarePants:
J'accuse! 😝 

Thanks for investigating this and I'm glad you found a cause. I'll leave it in your capable hands for now and I won't merge this in either. (We'll see whether we need it first)

antonfirsov:
@JimBobSquarePants I made a rounding bug in `Block8x8F.UnZigDivRound()`. Fixed and pushed it. I see no quality difference compared to the old integer-based version now. However, `System.Drawing`  quality is still better.

Check out `JpegTests.LoadResizeSave()` if you want to compare our current results to bleroy-s ones. Added 2 of it's test images.

JimBobSquarePants:
Good work. I've just pushed a change that changes our subsampling rules to match those of Photoshop 4:4:4 default falling to 4:2:0 at <=50% quality. That improves the output quality but increases the size. Gonna experiment with the quality + this PR's changes now.

JimBobSquarePants:
I'm finding the output is actually more pleasing to my eye (I tiny bit sharper for bleroys images; closer to ImageMagick) so once the CI checks pass I'm merging it.

JimBobSquarePants:
I can actually pretty much mimic ImageMagick using the Lanczos3Resampler, our filesize is larger though as  we use standard Huffman tables rather than optimized. 

http://www.impulseadventure.com/photo/optimized-jpeg.html

antonfirsov:
Yeah, in Jpeg we can spend endless amount of time to add features :)

If we ever start implementing/porting these, it's important to provide configuration and scaling options for the encoder. There is a  speed VS compression level VS image quality tradeoff. It won't be nice to enforce the same encoder characteristics for all users. 

An example: guetzli [uses](https://github.com/google/guetzli/blob/master/guetzli/dct_double.cc) double precision numbers in their DCT implementation for scaled blocks. I think it's important to achieve their image quality. Made no benchmarks yet, but it should be slower than a single-precision SIMD implementation, like ours.

JimBobSquarePants:
Interesting, yeah that should definitely be slower. 

The main feature I'd like to see really is progressive encoding. Any other features I expect developer to help add if they really require the functionality. 

I want to revisit the color code again for CMYK and YccK decoding myself. Looking at BitMiracle they do a lot of table building (I think guetzli precalculates them as I recall) and our output seems over saturated.

https://github.com/BitMiracle/libjpeg.net/blob/master/LibJpeg/Classic/Internal/jpeg_color_deconverter.cs

Incidentally I think we should add BitMiracle to our decoding encoding benchmarks.

antonfirsov:
Isn't it just standard (poorly ported) libjpeg code, or have they added some special stuff?

By my opinion, it would be nice to have separate unit tests for these color converter pieces of code, so we can compare, refactor and optimize different implementations systematically and independently from all the other jpeg mess. I developed and optimized `Block8x8F` this way, it's a very efficient methodology.

antonfirsov:
Regarding progressive encoding:

have you seen any library out there that's not based on libjpeg and is capable to do it? (I don't care about the language.)

antonfirsov:
Just pushed a change, that to optimizes yesterdays rounding algorithm. (It became quite slow with the integers)

[Before](https://i.imgsafe.org/d9a35d9324.png)

[After](https://i.imgsafe.org/d9a34bc9fb.png)

JimBobSquarePants:
Oh wow! Yeah, big difference. I'll have a good dig through your changes tonight to get my head round them.

JimBobSquarePants:
Re progressive, not yet. Will keep looking.

JimBobSquarePants:
That code is poorly ported, and I suspect very slow but the lookup principle is one I've seen in various implementations now. 

I've been benchmarking different approaches but agree we could test more.

PR #84, opened 2017-01-19
Add a .NET 4.6.1 profile to use its System.Numerics.Vectors

### LINKED ISSUE #81: net45-profile builds depend on System.Numerics.Vectory 4.1.1.0 from CoreFX
The binary packages targeting the full .NET framework (net45) depend on System.Numerics.Vectors 4.1.1.0, which is from the CoreFX. While it's possible to work around that with an assembly redirect, it shouldn't be necessary.

ibauersachs:
The downgrade from 4.6.1 to 4.6 in the benchmark is required to force
the System.Numerics.Vectors dependency to NuGet. The 4.6.1 version
does not contain the used generic vector type.

I have no idea if additional changes are required so that this profile ends up in the NuGet packages.

Closes #81

### DIFF
--- a/src/ImageSharp.Drawing/project.json
+++ b/src/ImageSharp.Drawing/project.json
@@ -51,7 +51,6 @@
       "type": "build"
     },
     "System.Buffers": "4.0.0",
-    "System.Numerics.Vectors": "4.1.1",
     "System.Runtime.CompilerServices.Unsafe": "4.0.0"
   },
   "frameworks": {
@@ -63,6 +62,7 @@
         "System.IO": "4.1.0",
         "System.IO.Compression": "4.1.0",
         "System.Linq": "4.1.0",
+        "System.Numerics.Vectors": "4.1.1",
         "System.ObjectModel": "4.0.12",
         "System.Resources.ResourceManager": "4.0.1",
         "System.Runtime.Extensions": "4.1.0",
@@ -76,7 +76,21 @@
     },
     "net45": {
       "dependencies": {
-        "System.Runtime": "4.0.0"
+        "System.Numerics.Vectors": "4.1.1",
+        "System.Threading.Tasks.Parallel": "4.0.0"
+      },
+      "frameworkAssemblies": {
+        "System.Runtime": "4.0.0.0"
+      }
+    },
+    "net461": {
+      "dependencies": {
+        "System.Threading.Tasks.Parallel": "4.0.0"
+      },
+      "frameworkAssemblies": {
+        "System.Runtime": "4.0.20.0",
+        "System.Numerics": "4.0.0.0",
+        "System.Numerics.Vectors": "4.0.0.0"
       }
     }
   }
--- a/src/ImageSharp.Formats.Bmp/project.json
+++ b/src/ImageSharp.Formats.Bmp/project.json
@@ -47,7 +47,6 @@
       "type": "build"
     },
     "System.Buffers": "4.0.0",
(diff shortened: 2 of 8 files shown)


JimBobSquarePants:
Thanks for this. I haven't forgotten it and once the team has a look we'll see about getting it merged 👍

antonfirsov:
Sandbox46 has `net461` as target framework. After altering the references in it's `csproj` to the new dll-s this PR provides, and uninstalling the `System.Numerics.Vectors` nuget package it Just Worked (TM).

No trouble in other test projects, everything seems fine to me!

JimBobSquarePants:
Sweet. I'll merge it then.

mellinoe:
I would recommend just using version `4.3.0` of System.Numerics.Vectors in every case here; the package is compatible with all of the frameworks used and should just "do the right thing" when used in each different framework.

JimBobSquarePants:
Will give it a shot. Thanks!

ibauersachs:
I had two rationales:
1. Use the system library where possible. There's no need for additional libraries that are already shipped as part of the full framework.
2. I needed to reference ImageSharp in a plugin of a legacy application. The main application already uses System.Numerics and loading the NuGet 4.3 version is not possible. It's possible to solve that conflict with assembly redirection, but I don't have control over that in some deployments and because of 1. it shouldn't be (and isn't) necessary.

If this change really is a problem, feel free to revert it and I'll build a custom version for our use.

mellinoe:
The other option is to remove "System.Numerics.Vectors" from the "frameworkassemblies" section for net461. It should not be necessary (and doesn't really do anything here), unless I'm missing something.

KLuuKer:
@mellinoe this is a direct link to the current project.json https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/project.json
maybe you should have a look there to see what "we" are doing

mellinoe:
This is the line I'm referring to. I think it should be removed:

https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/project.json#L85

KLuuKer:
the generic dependencies https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/project.json#L40
do not have any reference to numerics nor vectors tough...

mellinoe:
On .NET Framework, the System.Numerics.Vectors assembly does not contain any types. We call it a "facade assembly", meaning it only contains type-forwards. In this case, it just type forwards things into System.Numerics.dll. So you only need to list System.Numerics.dll there, and that is done on [line 84](https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/project.json#L84).

KLuuKer:
@JimBobSquarePants ^ you wanna try that?

ibauersachs:
@mellinoe Yes, this should work. Sorry, I don't have time for an updated PR right now.
@KLuuKer Putting System.Numerics into the generic dependencies would mess up the versioning for the different framework targets.

JimBobSquarePants:
I've just pushed the new project.json files. Everything seems to work locally. 

https://github.com/JimBobSquarePants/ImageSharp/commit/75b541b33fae28e256d2ef72b95aa39e23895247

KLuuKer:
@JimBobSquarePants same for me, everything is working fine now

JimBobSquarePants:
Great stuff. Thanks @mellinoe for the pointers!

PR #85, opened 2017-01-19
Load none-progressive images with missing EOI markers.

### LINKED ISSUE #18: Progressive jpeg with missing EOI marker throws EOFException
_From @gdoron on October 31, 2016 12:57_

Hi James @JimBobSquarePants,

This specific image: 
image
fails when trying to load it.

Stacktrace:

>ImageProcessorCore.Formats.JpegDecoderCore+EOFException: Exception of type 'ImageProcessorCore.Formats.JpegDecoderCore+EOFException' was thrown.
   at ImageProcessorCore.Formats.JpegDecoderCore.Fill()
   at ImageProcessorCore.Formats.JpegDecoderCore.ReadByte()
   at ImageProcessorCore.Formats.JpegDecoderCore.ReadByteStuffedByte()
   at ImageProcessorCore.Formats.JpegDecoderCore.EnsureNBits(Int32 n)
   at ImageProcessorCore.Formats.JpegDecoderCore.DecodeHuffman(Huffman huffman)
   at ImageProcessorCore.Formats.JpegDecoderCore.Refine(Block b, Huffman h, Int32 zigStart, Int32 zigEnd, Int32 delta)
   at ImageProcessorCore.Formats.JpegDecoderCore.ProcessStartOfScan(Int32 n)
   at ImageProcessorCore.Formats.JpegDecoderCore.Decode[TColor,TPacked](Image`2 image, Stream stream, Boolean configOnly)
   at ImageProcessorCore.Image`2.Decode(Stream stream)
   at ImageProcessorCore.Image`2.Load(Stream stream)
   at Yooocan.Logic.ImageLogic.Resize(Stream stream, Int32 quality, Nullable`1 width, Nullable`1 height)

Any clue?

By the way, checkout this story: https://yooocan.com/Story/329/Having-Wheels-Has-Its-Perks
Amazing people, and you're helping them James!

THANKS A LOT!

_Copied from original issue: JimBobSquarePants/ImageProcessor#505_

### LINKED ISSUE #83: none-progressive jpeg with missing EOI marker throws EOFException
I'm on 1.0.0-alpha1-00049, .NET Core 1.1.0. [An image I'm testing with (attached)](https://github.com/JimBobSquarePants/ImageSharp/files/715185/BadImage.zip) is throwing the below exception when I call `var image = new Image(stream);`. All other images I've tested work properly, but I'm not sure what's happening with this one.

(In case GitHub altered it, I've uploaded the image [here](https://vaindil.com/04-30-2011%20Section%20Conclave%20Ceremony%20%2835%29.JPG).)

Exception:

> EOFException: Exception of type 'ImageSharp.Formats.JpegDecoderCore+EOFException' was thrown.

My code:

```csharp
[HttpPost]
public async Task<IActionResult> Upload(IFormFile image)
{
    using (var stream = new MemoryStream())
    {
        await image.CopyToAsync(stream);

        // image verification, stream variable is not touched/used

        stream.Seek(0, SeekOrigin.Begin);

        StripExif(stream);

        // do stuff
    }

    return Ok();
}

public static void StripExif(Stream stream)
{
    // exception thrown here
    var image = new Image(stream);
    
    stream.Seek(0, SeekOrigin.Begin);

    if (image.ExifProfile == null)
        return;

    // a ton of these follow
    image.ExifProfile.RemoveValue(ExifTag.GPSLatitude);

    // ...

    image.Save(stream);
    stream.Seek(0, SeekOrigin.Begin);
}
```

Beginning of stack trace:

```
ImageSharp.Formats.JpegDecoderCore+EOFException: Exception of type 'ImageSharp.Formats.JpegDecoderCore+EOFException' was thrown.
   at ImageSharp.Formats.JpegDecoderCore.Decode[TColor](Image`1 image, Stream stream, Boolean configOnly)
   at ImageSharp.Formats.JpegDecoder.Decode[TColor](Image`1 image, Stream stream)
   at ImageSharp.Image`1.Decode(Stream stream)
   at ImageSharp.Image`1.Load(Stream stream)
   at [removed].StripExif(Stream stream) in [removed]\UploadHelpers.cs:line 39
   at [removed].ImagesController.<Upload>d__5.MoveNext() in [removed]\ImagesController.cs:line 66
```

tocsoft:
**UPDATE** : this only fixes none-progressive jpegs with the missing EOI issue e.g. #83 not progressive ones e.g. #18.

fixes #83 

OK guys, I'm not entirely sure I've fixed this properly but thought someone could check my logic and fix/point in the right direction as required.

I'm stopping processing of the the JPEG once the total number of MCUs processed is at the target number for this image.(i think).

This allows me to process an image with a missing EOI marker, badeof.jpg. I originally ran this on the image attached to #83 and then made my own (smaller) broken file that errors in the same way and my fix opens them both but i'm not sure if I've inadvertently missed something. Made it so it doesn't processes stuff it needs to.

cc @JimBobSquarePants @antonfirsov


JimBobSquarePants:
I think this is exactly how you are supposed to do this. Nice work! @tocsoft

dlemstra:
Couldn't you use one variable that contains the count and decrement that with the result of `scan.ProcessBlocks` and change the while to count being bigger than zero?

tocsoft:
nope... i've defiantly broken something. the fix is currently not working for progressive images.

tocsoft:
OK fixed the issue with progressive images by not fixing them 😏.
Basically this fix will only work with none-progressive images at the moment, added a comment to help with any follow ups but I think the patch has value even if progressive images remain untreated.

antonfirsov:
Seems OK to me. We can take care of the progressive stuff later.

src/ImageSharp.Formats.Jpeg/JpegDecoderCore.cs
```diff
@@ -294,6 +302,9 @@ public void Decode<TColor>(Image<TColor> image, Stream stream, bool configOnly)
                             return;
                         }
 
+                        // when this is a progressive image this gets called a number of times
+                        // need to know how many times this should be called in total.
+                        mcuSet = true;
```
JimBobSquarePants: I think the number depends on the individual encoding. https://msdn.microsoft.com/en-us/library/windows/desktop/ee720036(v=vs.85).aspx#_jpegprogressivedecoding

JimBobSquarePants:
I'm happy to merge this as is. We should test to see if system.drawing can decode a broken progressive jpeg some time before attempting a fix for that. If not, we'll just skip it.

tocsoft:
OK just refactored the PR a little as the call to `ProcessStartOfScan()` always ended in a 0 `mcuCounter` so instead of counting MCUs we should be counting levels processed... this still works with out test images.

src/ImageSharp.Formats.Jpeg/JpegDecoderCore.cs
```diff
@@ -187,7 +187,10 @@ public void Decode<TColor>(Image<TColor> image, Stream stream, bool configOnly)
             }
 
             // Process the remaining segments until the End Of Image marker.
-            while (true)
+            bool processBytes = true;
```
JimBobSquarePants: You're not counting the mcu's anymore? This doesn't fix the issue now with the image in #18

JimBobSquarePants:
@tocsoft  I had a play around with this as I realized that the image in #18 is progressive so I thought we should at least try to get something going for it. I've added the image to the repo and here's the current status. I'm not sure what causes that output, maybe @antonfirsov knows?

badeofprog

antonfirsov:
No idea yet, maybe I can dive into the progressive stuff after some sleep.

Regarding @tocsoft -s latest commit: I think it's better to have an exception, than having this output. People will keep complaining, and the code is quite messy with these try-catches.

antonfirsov:
Wait ... it was your commit @JimBobSquarePants :laughing: My concerns are the same though.

JimBobSquarePants:
Yeah, not my finest work. The offending exception in Bytes seems to get called everywhere though.

antonfirsov:
Yeah, no idea how to implement this in a clean way. It seems to be an actually exceptional exception. 

Except for the users having these buggy inputs regularly.

tocsoft:
yeah I think @JimBobSquarePants you should roll back your commit, we should get the first none-progressive fix in first but leave #18 open and then take a look at progressive as a separate fix, as this at least fixes #83

JimBobSquarePants:
No worries, will do.

antonfirsov:
Okkay, started to read jpeg articles to figure out the progressive stuff, hope I can deal with it.

I think we should merge this (without d05b06b), I will create a new PR, if I manage to fix #18

Issue #88, opened 2017-01-21
Passing Unmanaged Pointer as Pixels buffer in ImageSharp


ghost:
Part of Wayland development is that I create a Shared Memory Buffer that is being used between processes and I wanted to eliminate the amount of buffer copying between ImageSharp and Shared Memory Buffer. If implemented, we may actually be able to see a relatively fast User Interface on Linux for C# that is easy to use for everyone.

Can ImageSharp implement a safe api that accepts IntPtr for Argb or other image format pixel buffer? (Width/height/strides can be provided if needed.)

Please and thank you

BernhardGlueck:
In addition especially for server side processing, as right now images allocate array on the .NET Heap ( which inevitable end up on the LOH etc.. ) it would be great to use this to store the image data in off heap memory using unsafe operations... maybe introduce a base interface like IPixelStorage<TColor> with a provider IPixelStorageProvider<TColor> that allows different strategies to be put int

antonfirsov:
We are currently waiting for reliable VS2017 tooling (for C#7 ref returns) and for new official `corefx` packages. (for `Span<T>` and other low level memory management stuff). 

I think we could introduce design changes based on these features. :+1: for something like `IPixelStorageProvider`.

CsharpOnLinuxDev:
I see, so it potentially possible that Mono would not be supported in this case. (I don't know if it's still maintained to backport Corefx.)

JimBobSquarePants:
No that should be fine. Mono supports NetStandard

https://blog.xamarin.com/net-standard-library-support-for-xamarin/

antonfirsov:
For implementations plans see: #565.

PR #97, opened 2017-02-01
Make all images implement IDisposable

JimBobSquarePants:
### Prerequisites

### Description

This is a biggun' that's been a long time coming. 

The PR changes anything implementing `IImageBase<TColor>` to now also implement `IDisposable`.

There's a little bit of slight of hand going on within `SetPixels` and `ClonePixels` that allows all this to happen. Nothing too fancy though so it should all be easy to follow. The API is still simple but images should now be wrapped in `using` statements.

```
using (FileStream stream = File.OpenRead("foo.jpg"))
using (FileStream output = File.OpenWrite("bar.jpg"))
using (Image image = new Image(stream))
{
         .Grayscale()
         .Save(output);
}
```

Renting and returning of `TColor[]` instances is managed through a single place. `PixelPool<TColor>`. This public static class should always be the single point where all pixel array pooling is managed. (Note, you'll see an empty catch in `ReturnPixels()` that allows us to handle cases where someone doesn't rent from that pool which should be rare enough to not affect performance.)

The memory differences are astounding. Check out the resize benchmark.

```
```

### Before

Method |        Mean |    StdErr |    StdDev |      Median | Scaled | Scaled-StdDev |    Gen 0 |    Gen 1 |    Gen 2 | Allocated |
---------------------------- |------------ |---------- |---------- |------------ |------- |-------------- |--------- |--------- |--------- |---------- |
'System.Drawing Resize' |  76.8127 ms | 1.5556 ms | 9.9606 ms |  71.4049 ms |   1.00 |          0.00 |        - |        - |        - |     512 B |
'ImageSharp Resize' |  74.0430 ms | 0.1690 ms | 0.6095 ms |  73.8233 ms |   0.98 |          0.11 | 329.1667 | 266.6667 | 266.6667 |  20.05 MB |
'ImageSharp Compand Resize' | 126.1365 ms | 0.1987 ms | 0.7435 ms | 125.9578 ms |   1.67 |          0.19 | 333.3333 | 270.8333 | 270.8333 |  20.05 MB |

### After

Method |        Mean |    StdDev | Scaled | Scaled-StdDev | Allocated |
---------------------------- |------------ |---------- |------- |-------------- |---------- |
'System.Drawing Resize' |  72.3372 ms | 1.5962 ms |   1.00 |          0.00 |     512 B |
'ImageSharp Resize' |  70.6951 ms | 0.3417 ms |   0.98 |          0.02 | 186.24 kB |
'ImageSharp Compand Resize' | 120.9624 ms | 0.6795 ms |   1.67 |          0.04 | 186.62 kB |

Could you all please go through this with a fine-tooth comb and ensure I haven't missed anything that could lead to memory leaks. All my tests are passing and I can't find any leaks so far.

### DIFF
--- a/src/ImageSharp/Image/ImageBase{TColor}.cs
+++ b/src/ImageSharp/Image/ImageBase{TColor}.cs
@@ -163,7 +163,7 @@ namespace ImageSharp
             int newWidth = pixelSource.Width;
             int newHeight = pixelSource.Height;
 
-            // push my memory into the accessor (which in turn unpins the old puffer ready for the images use)
+            // Push my memory into the accessor (which in turn unpins the old puffer ready for the images use)
             TColor[] newPixels = pixelSource.ReturnCurrentPixelsAndReplaceThemInternally(this.Width, this.Height, this.pixelBuffer, true);
             this.Width = newWidth;
             this.Height = newHeight;
--- a/src/ImageSharp/Image/ImageBase{TColor}.cs
+++ b/src/ImageSharp/Image/ImageBase{TColor}.cs
@@ -15,14 +15,9 @@ namespace ImageSharp
     /// </summary>
     /// <typeparam name="TColor">The pixel format.</typeparam>
     [DebuggerDisplay("Image: {Width}x{Height}")]
-    public abstract class ImageBase<TColor> : IImageBase<TColor> // IImageBase implements IDisposable
+    public abstract class ImageBase<TColor> : IImageBase<TColor>
         where TColor : struct, IPackedPixel, IEquatable<TColor>
     {
-        /// <summary>
-        /// The <see cref="ArrayPool{TColor}"/> used to pool data. TODO: Choose sensible default size and count
-        /// </summary>
-        private static readonly ArrayPool<TColor> ArrayPool = ArrayPool<TColor>.Create(int.MaxValue, 50);
-
         /// <summary>
         /// The image pixels
         /// </summary>
@@ -92,14 +87,6 @@ namespace ImageSharp
                 // Check we can do this without crashing
                 sourcePixels.CopyTo(target);
             }
-        }
-
-        /// <summary>
-        /// Finalizes an instance of the <see cref="ImageBase{TColor}"/> class.
-        /// </summary>
-        ~ImageBase()
-        {
-            this.Dispose(false);
         }
 
         /// <inheritdoc/>
--- a/src/ImageSharp/Image/ImageBase{TColor}.cs
+++ b/src/ImageSharp/Image/ImageBase{TColor}.cs
@@ -146,49 +146,28 @@ namespace ImageSharp
         }
 
         /// <inheritdoc/>
-        public void SetPixels(int width, int height, TColor[] pixels)
-        {
-            Guard.MustBeGreaterThan(width, 0, nameof(width));
-            Guard.MustBeGreaterThan(height, 0, nameof(height));
-            Guard.NotNull(pixels, nameof(pixels));
-
-            if (!(pixels.Length >= width * height))
-            {
-                throw new ArgumentException($"Pixel array must have the length of at least {width * height}.");
-            }
-
-            this.Width = width;
-            this.Height = height;
-
-            this.ReturnPixels();
-            this.pixelBuffer = pixels;
-        }
-
-        /// <inheritdoc/>
-        public void ClonePixels(int width, int height, TColor[] pixels)
-        {
-            Guard.MustBeGreaterThan(width, 0, nameof(width));
-            Guard.MustBeGreaterThan(height, 0, nameof(height));
-            Guard.NotNull(pixels, nameof(pixels));
-
-            if (!(pixels.Length >= width * height))
-            {
-                throw new ArgumentException($"Pixel array must have the length of at least {width * height}.");
-            }
-
-            this.Width = width;
-            this.Height = height;
-
-            // Copy the pixels. TODO: use Unsafe.Copy.
-            this.ReturnPixels();
-            this.RentPixels();
-            Array.Copy(pixels, this.pixelBuffer, width * height);
-        }
-
-        /// <inheritdoc/>
         public virtual PixelAccessor<TColor> Lock()
         {
             return new PixelAccessor<TColor>(this);
+        }
+
+        /// <summary>
+        /// Switches the buffers used by the image and the PixelAccessor meaning that the Image will "own" the buffer from the PixelAccessor and the PixelAccessor will now own the Images buffer.
+        /// </summary>
+        /// <param name="pixelSource">The pixel source.</param>
+        internal void SwapPixelsBuffers(PixelAccessor<TColor> pixelSource)
+        {
+            Guard.NotNull(pixelSource, nameof(pixelSource));
+            Guard.IsTrue(pixelSource.PooledMemory, nameof(pixelSource.PooledMemory), "pixelSource must be using pooled memory");
+
+            int newWidth = pixelSource.Width;
+            int newHeight = pixelSource.Height;
+
+            // push my memory into the accessor (which in turn unpins the old puffer ready for the images use)
+            TColor[] newPixels = pixelSource.ReturnCurrentPixelsAndReplaceThemInternally(this.Width, this.Height, this.pixelBuffer, true);
+            this.Width = newWidth;
+            this.Height = newHeight;
+            this.pixelBuffer = newPixels;
         }
 
         /// <summary>
--- a/src/ImageSharp/Image/ImageBase{TColor}.cs
+++ b/src/ImageSharp/Image/ImageBase{TColor}.cs
@@ -6,7 +6,6 @@
 namespace ImageSharp
 {
     using System;
-    using System.Buffers;
     using System.Diagnostics;
 
     /// <summary>
--- a/src/ImageSharp/Image/ImageBase{TColor}.cs
+++ b/src/ImageSharp/Image/ImageBase{TColor}.cs
@@ -6,6 +6,7 @@
 namespace ImageSharp
 {
     using System;
+    using System.Buffers;
     using System.Diagnostics;
 
     /// <summary>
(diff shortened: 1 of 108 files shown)


src/ImageSharp/Image/ImageBase{TColor}.cs
```diff
@@ -67,15 +84,24 @@ protected ImageBase(ImageBase<TColor> other)
...
             {
+                // Check we can do this without crashing
                 sourcePixels.CopyTo(target);
             }
         }
 
+        /// <summary>
+        /// Finalizes an instance of the <see cref="ImageBase{TColor}"/> class.
+        /// </summary>
+        ~ImageBase()
```
benaadams: Don't have finalizer, else if the user doesn't call dispose the allocated memory will hang around in the finalizer queue. Better to allow it to be picked up by GC early. 

Ideally the user would call Dispose, but if they don't without the finalizer the allocations will be the same as vanilla new; with the finalizer the allocations will increase.

JimBobSquarePants: Great. Easy change to make. 👍

tocsoft:
My initial thought would be that SetPixels should probably copy the passed in TColor[] into is internal memory that means that we never have to deal with pixel data that is owned by someone else when disposing of `ImageBase` which would prevent the need to catch the return in the array pool. it might be a tiny bit slower but it would be a lot safer.

It would also mean that the worry about leaks would be massively reduced as the rule of thumb can then be if you rent it you return it unlike in Convolution2DProcessor.cs where your renting a TColor[] but there's no obvious way in which its collected.

You could also have `PixelPool<TColor>` be able to return a `PixelAccessor<TColor>` would prevent the need to create an array then lock it and then return the rented array back, it would all be encapsulated in the `PixelAccessor<TColor>` which can get its own memory and release it in its dispose.

tocsoft:
you could even have a version of setpixels that is optimized to just swap the backing data from the new pooled pixel accessor and the image that would prevent the copy and mean that the old data would be freed up with the pooled accessor (would have to be careful not to do this when both sets of data are needed to be retained, but if the method was internal and was called something else then it shouldn't be a problem)

JimBobSquarePants:
>you could even have a version of setpixels that is optimized to just swap the backing data from the new pooled pixel accessor and the image that would prevent the copy and mean that the old data would be freed up with the pooled accessor (would have to be careful not to do this when both sets of data are needed to be retained, but if the method was internal and was called something else then it shouldn't be a problem)

@tocsoft  Isn't that exactly what I'm doing?

https://github.com/JimBobSquarePants/ImageSharp/blob/41f96faa7b23fb51463541231f0b0faf5aaa1f1c/src/ImageSharp/Image/ImageBase%7BTColor%7D.cs#L163

tocsoft:
you are sort of the problem is your doing it in such a way that its hard to track who is currently in charge of what memory... at the moment it looks like Convolution2DProcessor.cs is leaking because your renting but never releasing within the same class/method. Your obviously not leaking there, but it looks like a leak, thus making it harder to maintain.

I'll code up what I mean shortly and show you...seeing code is easier that trying to describe it. 😄

JimBobSquarePants:
Please do, looking forward to seeing what you come up with.

KLuuKer:
@JimBobSquarePants maybe put a debug only line in that empty catch to throw when someone is returning non-pooled array's, so they will know that they are doing something wrong

JimBobSquarePants:
@KLuuKer Good thinking!

antonfirsov:
@KLuuKer I don't think it's a good idea. `SetPixels()` is a public function, users consuming it from a release binary will not notice they are doing something wrong.

@JimBobSquarePants 
Looking forward for @tocsoft -s gist, maybe his work fill change my mind a bit but currently I'm very concerned, this PR would leave a fundamental problem that must be fixed in the near future: 

*`SetPixels()` is no longer maintainable as a safe API in it's current form! Specially not as a public API.*

I would rather delegate the ownership management of the buffers to a separate `PixelStorageProvider`-like abstract class as commented under #88. It should care about the concern of safely swapping the buffer under an image. It will also enable supporting many custom scenarios in the future.

I would be happy to create a PR for this after #97 is merged!

JimBobSquarePants:
So it turns out there is no error thrown after all! I could have sworn it used to. That means the catch block is gone.

>SetPixels() is no longer maintainable as a safe API in it's current form! Specially not as a public API.

@antonfirsov Could you expand upon that?

antonfirsov:
It's a leaky abstraction now:
It's intended to be used used with arrays rented from `PixelPool<TColor>`, but it's not enforced, nor verified when we call `SetPixels()`. We will know we made something wrong only when Dispoese() throws an exception internal to `ArrayPool`. (Worse, if not throwing anything ...)

antonfirsov:
@JimBobSquarePants it's not well documented but as I understand there [must be an exception thrown](https://github.com/dotnet/corefx/blob/master/src/System.Buffers/src/System/Buffers/DefaultArrayPoolBucket.cs#L91), when you return a non-pooled array.

JimBobSquarePants:
@antonfirsov I was convinced it was but it isn't and I don't know why

https://github.com/JimBobSquarePants/ImageSharp/blob/94ca0779e3e0c7842742df9bb64c6a405af08a74/tests/ImageSharp.Tests/Image/PixelPoolTests.cs#L48

benaadams:
It throws if the array length is not a power of 2; as its definitely not from the pool - otherwise the pool can't tell

tocsoft:
ok had my play see 7940683d37b9259871d527f578b324d3ced9082d for how I propose we change this PR.

This way it means that most people don't even need to think about the PixelPool it just work for them.

Sorry but there ended being a lot more changes than I wanted but it ends up dropping the CopyPixels/SetPixel calls all together.

tocsoft:
basically my changes moves the responsibility for Renting and Returning the `TColor[]` arrays to the `PixelAccessor<TColor>` & `ImageBase<TColor>` classes thus leaving us with alot less chance of leaking them, plus as they are both IDisposable now you can do away with the `try{}finaly{}`'s as long as you wrap them in `using(..){...}` statements.

JimBobSquarePants:
@tocsoft I like that, much less leaky. We'd want `internal void SwapPixelsBuffers` to be protected rather than internal so devs can use it though would we not?

tocsoft:
Not sure about internal vs protected. However my gut says make everything internal/private until a third party dev needs access to it then decide on the best way to make it public then.

JimBobSquarePants:
@tocsoft Ok then, we've still got plenty of time to review that anyway. Can you post the resize benchmark results under your version please (should be the same) and if everyone's happy then you can update the PR with your changes.

antonfirsov:
@JimBobSquarePants a non-throwing could be even worse: the user makes a mistake by using [SetPixels() with a non buffered array](https://github.com/JimBobSquarePants/ImageSharp/blob/IDisposable/src/ImageSharp.Processing/Processors/Convolution/ConvolutionProcessor.cs#L47) and he will not notice it at all!

@tocsoft Good job, `SwapPixelsBuffers()` is much better than `SetPixels()` was! This way it's not blinking in red in my eyes anymore! :) 

However I feel buffer management/swapping is not a `PixelAccessor` concern, that's why I want to think about moving this logic to a separate class later.

tocsoft:
### Before - master 8e56aaaf68177d6e6a14c5dee2a7d4772452895c
``` ini

Processor=Intel(R) Core(TM) i7-4770 CPU 3.40GHz, ProcessorCount=8
Frequency=3312645 Hz, Resolution=301.8736 ns, Timer=TSC


```
|                      Method |       Mean |    StdDev | Scaled | Scaled-StdDev |    Gen 0 |    Gen 1 |    Gen 2 | Allocated |
|---------------------------- |----------- |---------- |------- |-------------- |--------- |--------- |--------- |---------- |
|     'System.Drawing Resize' | 59.0057 ms | 2.2540 ms |   1.00 |          0.00 |        - |        - |        - |     512 B |
|         'ImageSharp Resize' | 41.1233 ms | 0.5927 ms |   0.70 |          0.03 | 195.8333 | 195.8333 | 195.8333 |  20.08 MB |
| 'ImageSharp Compand Resize' | 64.2315 ms | 0.6072 ms |   1.09 |          0.04 | 350.0000 | 350.0000 | 350.0000 |  20.08 MB |

### After - 7940683d37b9259871d527f578b324d3ced9082d

``` ini

Processor=Intel(R) Core(TM) i7-4770 CPU 3.40GHz, ProcessorCount=8
Frequency=3312645 Hz, Resolution=301.8736 ns, Timer=TSC


```
|                      Method |       Mean |    StdDev | Scaled | Scaled-StdDev | Allocated |
|---------------------------- |----------- |---------- |------- |-------------- |---------- |
|     'System.Drawing Resize' | 57.2351 ms | 0.6135 ms |   1.00 |          0.00 |     512 B |
|         'ImageSharp Resize' | 39.3286 ms | 1.0688 ms |   0.69 |          0.02 | 186.33 kB |
| 'ImageSharp Compand Resize' | 63.1998 ms | 0.9299 ms |   1.10 |          0.02 | 186.36 kB |

JimBobSquarePants:
Any reason I shouldn't merge this now?

tocsoft:
can't see one...go for it 👍

antonfirsov:
Lets get it finished! ;) This `IDisposable` is really not that :bat:`IDisposable`:bat: !

JimBobSquarePants:
@antonfirsov I'm taking that as a yes 😉

tocsoft:
ooo just remembered don't forget to update the readme so show your supposed to dispose of the images.

JimBobSquarePants:
I'll do that in a mo and skip CI.

gitviper:
Hi, I've just began giving a try to ImageSharp, mostly because the pure managed approach. I was surprised to see the IDisposable in the Image object. Isn't this supposed to be used only when your class requires unmanaged resources only?

Is it really used somewhere in the codebase? if not, is it supposed to be for when someone is willing to extend ImageBase in a way that might require unmanaged resources? And if that's the case, is it worth to put the extra burden of require the "using" everywhere just for the rare case of someone going in that direction?

I think using IDisposable defeats the whole idea of pure managed. In my case, I wanted to use imageSharp in a project that moves images from one filter to another, in a way that's not easy to tell when to dispose an image.

So if there's a place for it, I give my vote to remove IDisposable

antonfirsov:
@vpenades 
- Large managed buffers mostly end up on [Large Object Heap](https://www.simple-talk.com/dotnet/net-framework/the-dangers-of-the-large-object-heap/).
- Literally speaking it's just managed memory, like basic (small objects), but in practice they are really expensive to  manage by the GC. It's a [leaky abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction), and a major performance bottleneck for an imaging library.
- The .NET team addressed this issue by defining a utility called [ArrayPool](https://github.com/dotnet/corefx/issues/4547). It follows a Rent/Return logic which is very similar to the logic we use for unmanaged resources. Much safer though: if you forget to return the array it will be GC-d, no memory leak will happen.
- The backing buffers of ImageSharp images are pooled now: arrays are rented in the constructor, and returned in `Dispose()`

antonfirsov:
> In my case, I wanted to use imageSharp in a project that moves images from one filter to another, in a way that's not easy to tell when to dispose an image.

I suggest you to describe your issue in a more specific way on our gitter channel! We can help you to find the best solution for your use case :)

PR #99, opened 2017-02-05
Move meta data properties to new classes.

dlemstra:
### Prerequisites

### Description
This pull requests moves out all the meta data properties to two new classes. One is called `ImageMetaData` and this is used in the `Image<TColor>` class and the other one is called `ImageFrameMetaData` and is used in the `ImageFrame` class. The latter only contains a FrameDelay value because we only need this at the moment. This could have been added to the ImageFrame class but the TIFF format will not use this.


src/ImageSharp.Formats.Png/PngEncoderCore.cs
```diff
@@ -126,12 +126,12 @@ internal sealed class PngEncoderCore
...
 
         /// <summary>
-        /// Encodes the image to the specified stream from the <see cref="ImageBase{TColor}"/>.
+        /// Encodes the image to the specified stream from the <see cref="Image{TColor}"/>.
         /// </summary>
         /// <typeparam name="TColor">The pixel format.</typeparam>
         /// <param name="image">The <see cref="ImageBase{TColor}"/> to encode from.</param>
         /// <param name="stream">The <see cref="Stream"/> to encode the image data to.</param>
-        public void Encode<TColor>(ImageBase<TColor> image, Stream stream)
+        public void Encode<TColor>(Image<TColor> image, Stream stream)
```
JimBobSquarePants: Oops! Yeah that should definitely be `Image<TColor>`

src/ImageSharp/Metadata/ImageMetaData.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// <summary>
+    /// Encapsulates the metadata of an image.
+    /// </summary>
+    public sealed class ImageMetaData
+    {
+        /// <summary>
+        /// The default horizontal resolution value (dots per inch) in x direction.
+        /// <remarks>The default value is 96 dots per inch.</remarks>
+        /// </summary>
+        public const double DefaultHorizontalResolution = 96;
```
JimBobSquarePants: Do we still need these constants?

We can probably do `public double HorizontalResolution {get; set;} = 96`

src/ImageSharp/Metadata/ImageMetaData.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ImageMetaData"/> class
+        /// by making a copy from other metadata.
+        /// </summary>
+        /// <param name="other">
+        /// The other <see cref="ImageMetaData"/> to create this instance from.
+        /// </param>
+        internal ImageMetaData(ImageMetaData other)
+        {
+            Debug.Assert(other != null);
```
JimBobSquarePants: Now I know why you were asking...

src/ImageSharp/Metadata/ImageMetaData.cs
dlemstra: They are public so that's why I did not remove them. I did not use auto properties with HorizontalResolution  because I want to protect them from being set to a value lower or equal to zero. I am okay with removing the constants and setting it in the constructor. Can I go ahead and change this?

src/ImageSharp/Metadata/ImageMetaData.cs
dlemstra: Yeah, code talks better than twitter 😄.

src/ImageSharp/MetaData/ImageMetaData.cs
```diff
@@ -0,0 +1,169 @@
...
+        public double HorizontalResolution
+        {
+            get
+            {
+                return this.horizontalResolution;
+            }
+
+            set
+            {
+              if (value > 0)
```
antonfirsov: I wonder why do we silently ignore `<=0` values instead of throwing an exception. Is this some kind of standard imaging behavior I don't know about? :)

tests/ImageSharp.Tests/MetaData/ImageMetaDataTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            image.MetaData.VerticalResolution = 300;
+
+            image.MetaData.HorizontalResolution = 100;
+
+            Assert.Equal(200, ((Rational)image.MetaData.ExifProfile.GetValue(ExifTag.XResolution).Value).ToDouble());
+            Assert.Equal(300, ((Rational)image.MetaData.ExifProfile.GetValue(ExifTag.YResolution).Value).ToDouble());
+
+            image.MetaData.SyncProfiles();
+
+            Assert.Equal(100, ((Rational)image.MetaData.ExifProfile.GetValue(ExifTag.XResolution).Value).ToDouble());
```
antonfirsov: I really believe that unit tests are more valuable, if they contain a single AAA block / test case and have a descriptive name saying an actual Fact about the component under test. 

A little bit longer to write, easier to maintain and understand.

src/ImageSharp/MetaData/ImageMetaData.cs
dlemstra: Personally not a big fan of throwing exceptions inside properties. Rather ignore invalid values. What do you prefer @JimBobSquarePants?

src/ImageSharp/MetaData/ImageMetaData.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        /// <summary>
+        /// Synchronizes the profiles with the current meta data.
+        /// </summary>
+        internal void SyncProfiles()
+        {
+            this.SyncExifProfile();
+        }
+
+        private void SyncExifProfile()
```
antonfirsov: Wouldn't it be better to move this method into `ExifProfile` and use it as `this.?ExifProfile.Sync(this)`?

tests/ImageSharp.Tests/MetaData/ImageMetaDataTests.cs
dlemstra: As you have noticed I have a different way of writing the tests. I use Arrange, Act, Assert, Act, Assert ... a lot of times. But I do think we need some common way of how we set up tests. We have a lot of different styles now. I think we need some help from our Grand High Eternal Dictator @JimBobSquarePants

src/ImageSharp/MetaData/ImageMetaData.cs
dlemstra: To me it felt like the logic should be inside the ImageMetaData class.

tests/ImageSharp.Tests/MetaData/ImageMetaDataTests.cs
JimBobSquarePants: Tests need a cleanup to be honest. We've got a lot of integration tests but we really need a lot more simple AAA unit tests. 

We should look at CoreFX for inspiration. Their tests are a great example of how to keep the tests concise and descriptive.

src/ImageSharp/MetaData/ImageMetaData.cs
JimBobSquarePants: I guess all profiles should probably have a common Sync method enforced by an interface.

tests/ImageSharp.Tests/MetaData/ImageMetaDataTests.cs
antonfirsov: I also think corefx tests are good example, I think we should follow their standard:
- One test case is testing only one aspect of a "thing" (proving one fact), no long test cases containing multiple AAA blocks
- All tests have a descriptive name, no need for additional comments
- If the fact (to be proven by the test) is coplex enough, [it's encouraged](https://github.com/dotnet/corefx/blob/master/src/System.Collections/tests/Generic/Dictionary/Dictionary.Tests.cs#L44) to use underscores in it's name to improve readability.

src/ImageSharp/MetaData/ImageMetaData.cs
antonfirsov: It seems to be a Profile-concern to sync it from a ImageMetadata.

tests/ImageSharp.Tests/MetaData/ImageMetaDataTests.cs
JimBobSquarePants: @antonfirsov You and your naming! 😝 I'll actually allow that in tests since naming is so totally different from normal source and it does improve readability in these cases.

src/ImageSharp/MetaData/ImageMetaData.cs
dlemstra: Okay i'll move it to the ExifProfile.

src/ImageSharp/MetaData/ImageMetaData.cs
JimBobSquarePants: Meant to respond to this earlier. We're inconsistent in the library. Sometimes we throw, sometimes we clamp. Leave it as is just now but we should readers the library again before final release.

PR #101, opened 2017-02-12
Added file path based APIs to Image & Image<TColor>

tocsoft:
### Prerequisites

### Description

The change entails adding a new target to the core ImageSharp package targeting netstandard .3, and adding a define to the netstandard1.1 target to allow conditional compilation.

In netstandard1.3+ we now have a dependency on `System.IO.FileSystem` this allows us to add a constructors to Image that take in a file path instead of a stream simplifying the boiler plate required to creating a file. It additionally allows us to add `Image.Save(path)` as an API that saves the file to the named path (looking up the correct image format, based on file's extension, from the formats loaded in to the `Configuration`)

This will simplify how a proportion of users will make use of the API but doesn't break compatibility with netstandard1.1 just doesn't have the nice path based APIs available.

How the API will change:
#### Stream Only API 
The API you can use today and will continue to be able to use when targeting netstandard < 1.3
```csharp
using (FileStream stream = File.OpenRead("foo.jpg"))
using (FileStream output = File.OpenWrite("bar.jpg"))
using (Image image = new Image(stream))
{
         .Grayscale()
         .Save(output);
}
```
#### File path API
Added APIs to dealing with file paths and not having to deal with streams directly.
```csharp
using (Image image = new Image("foo.jpg")) // file only locked while loading data into memory not during entire lifetime of Image
{
         .Grayscale()
         .Save("bar.jpg"); // automatic encoder selected based on extension.
}
```

### DIFF
--- a/src/ImageSharp/Image/Image{TColor}.cs
+++ b/src/ImageSharp/Image/Image{TColor}.cs
@@ -63,6 +63,28 @@ namespace ImageSharp
             this.Load(stream);
         }
 
+#if !NO_FILE_IO
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Image{TColor}"/> class.
+        /// </summary>
+        /// <param name="filePath">
+        /// The file containing image information.
+        /// </param>
+        /// <param name="configuration">
+        /// The configuration providing initialization code which allows extending the library.
+        /// </param>
+        /// <exception cref="System.ArgumentNullException">Thrown if the <paramref name="filePath"/> is null.</exception>
+        public Image(string filePath, Configuration configuration = null)
+            : base(configuration)
+        {
+            Guard.NotNull(filePath, nameof(filePath));
+            using (var fs = File.OpenRead(filePath))
+            {
+                this.Load(fs);
+            }
+        }
+#endif
+
         /// <summary>
         /// Initializes a new instance of the <see cref="Image{TColor}"/> class.
         /// </summary>
@@ -189,11 +211,12 @@ namespace ImageSharp
         /// </summary>
         /// <param name="stream">The stream to save the image to.</param>
         /// <param name="format">The format to save the image as.</param>
-        /// <exception cref="System.ArgumentNullException">Thrown if the stream is null.</exception>
+        /// <exception cref="System.ArgumentNullException">Thrown if the stream or format is null.</exception>
         /// <returns>The <see cref="Image{TColor}"/></returns>
         public Image<TColor> Save(Stream stream, IImageFormat format)
         {
             Guard.NotNull(stream, nameof(stream));
+            Guard.NotNull(format, nameof(format));
             format.Encoder.Encode(this, stream);
             return this;
         }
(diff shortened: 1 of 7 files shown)


dlemstra:
Cool idea to have support for both 1.0 and 1.3. When my project targets 1.4 it will pick the 1.3 version instead of the 1.1 one? I don't like the double negative in `!NO_FILE_IO`. And you could use `NETSTANDARD1_3` instead.

antonfirsov:
My only concern is compilation speed.

Maybe I'm not using Visual Studio and the projects in a proper way, but currently when I need to run a single test after a small change in the `ImageSharp` core project `Projects ✕ Targets` compilations are executed. Adding a new target will make Edit+Compile+Run cycles even longer, having a bad effect on productivity.

Is it possible to solve this?

src/ImageSharp/Image/Image{TColor}.cs
```diff
@@ -63,6 +63,28 @@ public Image(Stream stream, Configuration configuration = null)
             this.Load(stream);
         }
 
+#if !NO_FILE_IO
```
JimBobSquarePants: I wonder whether we should use netstandard names as directives e.g. 

```
#if  (!NETSTANDARD11 && !NETSTANDARD12) 
    // Do your thang!
#endif
```

My reasoning is that we know what API's are available for each standard.

src/ImageSharp/Image/Image{TColor}.cs
```diff
@@ -221,6 +245,58 @@ public Image<TColor> Save(Stream stream, IImageEncoder encoder)
...
+        /// <summary>
+        /// Saves the image to the given stream using the currently loaded image format.
+        /// </summary>
+        /// <param name="filePath">The file path to save the image to.</param>
+        /// <exception cref="System.ArgumentNullException">Thrown if the stream is null.</exception>
+        /// <returns>The <see cref="Image{TColor}"/></returns>
+        public Image<TColor> Save(string filePath)
+        {
+            var ext = Path.GetExtension(filePath).Trim('.');
```
JimBobSquarePants: `var`

src/ImageSharp/project.json
```diff
@@ -46,7 +46,31 @@
     "System.Runtime.CompilerServices.Unsafe": "4.0.0"
   },
   "frameworks": {
+    "netstandard1.3": {
```
JimBobSquarePants: I think it's possible to simplify this. VS2017 projects seem to reference neststandard1.6 for each standard implementation. We'll figure that out when we switch in a couple of weeks though when the RTM hits.

JimBobSquarePants:
@dlemstra I thought the same thing. 

@antonfirsov VS keeps telling me in the output window that we can speed up our compilation. I haven't followed the given instructions though yet.

src/ImageSharp/project.json
dlemstra: We might want to wait a while before we take the RTM version? Been working on it all day on Friday and it was not very friendly. Had a couple of unexpected crashes.

src/ImageSharp/Image/Image{TColor}.cs
dlemstra: Maybe !NETSTANDARD1_1 ?

src/ImageSharp/project.json
JimBobSquarePants: RTM's not out until 7th March as far as I was aware?

src/ImageSharp/Image/Image{TColor}.cs
JimBobSquarePants: Maybe... I'd like to have a dig around some of the MS stuff to see what their convention is.

tocsoft: I still think I prefer feature based directives rather than platform based ones, but saying that I would be willing to accept NETSTANDARDX_X based ones to.

I originally went with the positive HAS_FILE_IO on all targets except the 1.1. target but I changed my mind last minute.

tocsoft:
@antonfirsov We could probably drop all the net451/net46 targets as they will just pickup the netstandard ones. We would only need to target them at the edges, in the test project etc. the core could/should probably be netstandard only.

We would then only have the 2 targets, netstandard 1.1 and 1.3 for FileSystem and everything else.. and even then it would only need to be the core ImageSharp package that needs to target 1.3 everything else would then only target 1.1.

JimBobSquarePants:
Looks good to me now.

antonfirsov:
I think we can address the compilation speed issue later. Maybe after switching to VS2017.

PR #106, opened 2017-02-15
Add dithering binarization to methods and color dithering to quantizers

JimBobSquarePants:
### Prerequisites

### Description
There's a fair whack here but I couldn't really make the PR any smaller.

This PR does the following.

Adds ordered dithering methods with the following matrices.

- Bayer
- Ordered

Adds error diffusion methods with the following matrices

- Atkinson
- Burks
- FloydSteinberg
- JarvisJudiceNinke
- Sierra2
- Sierra3
- SierraLite
- Stucki

Adds optional dithering to the Palette and Octree quantizers. Note: WuQuantizer does not seem to work well with dithering. I ended up with random spots of odd color and the only other [example](https://www.codeproject.com/Articles/66341/A-Simple-Yet-Quite-Powerful-Palette-Quantizer-in-C) I found turned off their dithering for that quantizer.

The output quality of gifs is insane now. This is only 256 colors using the Octree quantizer.

calliphora

Testing the algorithms is tricky. I'd have to create pixel arrays to compare to and since there is no reference to work from I don't know how I would manage to create them in a manner that wouldn't be incredibly time consuming. I've opted for visual tests just now.


src/ImageSharp/Dithering/ErrorDiffusion/ErrorDiffuser.cs
```diff
@@ -0,0 +1,107 @@
...
+                if (matrix[0, i] != 0)
+                {
+                    this.startingOffset = (byte)(i - 1);
+                    break;
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public byte[,] Matrix { get; }
```
JimBobSquarePants: Oh yeah! We switched to `float[][]` in the edge detection algorithms.

tests/ImageSharp.Tests/Formats/GeneralFormatTests.cs
```diff
@@ -111,26 +111,27 @@ public void ImageCanConvertFormat()
...
             foreach (TestFile file in Files)
             {
-                Image image = file.CreateImage();
-
-                using (FileStream output = File.OpenWrite($"{path}/{file.FileNameWithoutExtension}.gif"))
+                using (Image image = file.CreateImage())
                 {
-                    image.SaveAsGif(output);
-                }
+                    using (FileStream output = File.OpenWrite($"{path}/{file.FileNameWithoutExtension}.bmp"))
```
JimBobSquarePants: Good to know! Just merged master back in.

tests/ImageSharp.Tests/Formats/GeneralFormatTests.cs
```diff
@@ -142,22 +143,25 @@ public void ImageShouldPreservePixelByteOrderWhenSerialized()
...
+                        memoryStream.Flush();
+                        serialized = memoryStream.ToArray();
+                    }
                 }
 
                 using (MemoryStream memoryStream = new MemoryStream(serialized))
                 {
-                    Image image2 = new Image(memoryStream);
-                    using (FileStream output = File.OpenWrite($"{path}/{file.FileName}"))
+                    using (Image image2 = new Image(memoryStream))
```
JimBobSquarePants: Aye this was original written before we could do that. Will update.

tests/ImageSharp.Tests/Formats/GeneralFormatTests.cs
JimBobSquarePants: Tests target netstandard1.1 and net4.5.1 just now. Won't switch till we move to VS2017

tocsoft: It targets netcoreapp1.1 which is actually netstandard 1.6 compatible.

src/ImageSharp/Dithering/ErrorDiffusion/ErrorDiffuser.cs
antonfirsov: `float[][]` is still crazy slow. We need at least [something like this](https://gist.github.com/antonfirsov/6c11a2f0a1048e8bcca455c072d14810). 

Even better to have simple on-stack value types for these small matrices. (Need to invent a trick for this).

antonfirsov:
@JimBobSquarePants your issue with tests is exactly the reason why I'm dreaming about statistical image comparer and reference output images in test cases :)

JimBobSquarePants:
@antonfirsov We'll get there eventually. I was looking at https://github.com/JaCraig/Structure.Sketching to see if there was any indication on how the expected results there were generated but I couldn't see anything.

tests/ImageSharp.Tests/Formats/GeneralFormatTests.cs
JimBobSquarePants: Odd.. I was only getting those options and no file overloads in my VS instance.

antonfirsov:
Seems [they are checking for exact binary equality](https://github.com/JaCraig/Structure.Sketching/blob/master/Structure.Sketching.Tests/BaseClasses/TestBaseClass.cs#L23).

Not a roboust solution once you have floating point algorithms.

antonfirsov:
To generate the expected results:
- We can use a tool like ImageMagick
- Or sometimes maybe we can generate them with our own implementations after we made sure visually the output is OK. - Not the best solution, but still protects us from breaking something unintentionally.

src/ImageSharp.Formats.Png/PngEncoderCore.cs
```diff
@@ -495,7 +495,7 @@ private QuantizedImage<TColor> WritePaletteChunk<TColor>(Stream stream, PngHeade
 
             if (this.Quantizer == null)
             {
-                this.Quantizer = new WuQuantizer<TColor>();
+                this.Quantizer = new OctreeQuantizer<TColor>();
```
antonfirsov: Hmm, this has nothing to do with the Dithering. 
I think we need a benchmark measuring the perf effects of switching quantizers. OctreeQuantizer seemed to be really slow when I tested it with Gif.
Certainly a winner in quality though.

JimBobSquarePants: It's doing a lot more work than before.... I added caching in the distance lookup but maybe there is more we can do. 

The Quantizer can be set so dithering could be turned off if they don't want it. I'm making it opt-out since the quality difference is so dramatic.

JimBobSquarePants:
I think we might be stuck with the latter for result comparison as we are going to generate different results from ImageMagick and others at times.

src/ImageSharp/Dithering/ErrorDiffusion/IErrorDiffuser.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// <summary>
+    /// Encapsulates properties and methods required to perfom diffused error dithering on an image.
+    /// </summary>
+    public interface IErrorDiffuser
+    {
+        /// <summary>
+        /// Gets the dithering matrix
+        /// </summary>
+        byte[][] Matrix { get; }
```
antonfirsov: @JimBobSquarePants why are you using `byte`-s here? I think perf would be much better with floats.
I also keep promoting [my idea](https://gist.github.com/antonfirsov/6c11a2f0a1048e8bcca455c072d14810) to flatten these arrays.

src/ImageSharp.Formats.Png/PngEncoderCore.cs
JimBobSquarePants: Ah sorry, misread that. Slower without dithering? Hmmm....

src/ImageSharp/Dithering/ErrorDiffusion/IErrorDiffuser.cs
JimBobSquarePants: Laziness 😈 I was reading from the samples I found that were using bytes. Ordered dithering requires bytes though since we are using to/.from bytes to reduce complexity for all pixel formats. Happy to switch to fast flat array. Throw one into this PR if you like?

src/ImageSharp.Formats.Png/PngEncoderCore.cs
antonfirsov: I just wondered why did you change PngDecoder in the Dithering PR :)
But at least it helped noticed this change :)
I made this benchmark a month ago:
```
                               Method | InputCategory |          Mean |   Scaled | Scaled-StdDev | Allocated |
------------------------------------- |-------------- |-------------- |--------- |-------------- |---------- |
     'EncodeGifMultiple - ImageSharp' |     AllImages | 8,283.9980 ms | 3,691.26 |          0.00 |  72.11 MB |
 'EncodeGifMultiple - System.Drawing' |     AllImages |     2.2442 ms |     1.00 |          0.00 |   3.22 MB |
```
Ok, it's running on multiple Gif frames, while in PNG we have only one, but still very strange.

src/ImageSharp/Dithering/ErrorDiffusion/IErrorDiffuser.cs
antonfirsov: Flat arrays make CPU caches happy!

antonfirsov: Ehh not getting answers on discussions if I don't keep pressing F5. 
It's too late here to write working code :( I can do it next time. Will sketch BulkOperations instead, keep your eyes on gitter ;)

JimBobSquarePants: Will do 😄... We can enhance after this PR is merged. 

Shall I make a start on our new Color struct or do you want to do that?

Get some sleep! 🛌

tests/ImageSharp.Tests/Formats/GeneralFormatTests.cs
tocsoft: ok I think I've figured out how to fix the intellisense in VS.

1. close VS
2. delete `.vs` folder
3. run `dotnet restore` in the root repo folder
4. restart VS 

You should now see the file path based APIs, it looks like VS is caching the reference to the netstandard1.1 output instead of moving to the netstandard1.3 one it should now be using.

src/ImageSharp/Dithering/ErrorDiffusion/IErrorDiffuser.cs
JimBobSquarePants: Some `Fast2DArray<T>` benchmarks. Faster than a jagged array for most situations (including ours) and between 50-60% faster than 2D array across the board. Safer than jagged array also API -wise.

`Fast2DArray<T>`  FTW! Great suggestion @antonfirsov After I merge this PR i'll apply it to our convolution filters also.

```

Allocated=0 B
```

Method | Count |      Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev |
------------------------------------ |------ |---------- |---------- |---------- |------- |-------------- |
'Array access using 2D array' |    10 | 1.3680 ns | 0.0110 ns | 0.0395 ns |   1.00 |          0.00 |
'Array access using a jagged array' |    10 | 0.5169 ns | 0.0121 ns | 0.0453 ns |   0.38 |          0.03 |
'Array access using Fast2DArray' |    10 | 0.5076 ns | 0.0332 ns | 0.1243 ns |   0.37 |          0.09 |
'Array access using 2D array' |   100 | 1.0366 ns | 0.0370 ns | 0.1432 ns |   1.00 |          0.00 |
'Array access using a jagged array' |   100 | 0.4906 ns | 0.0101 ns | 0.0365 ns |   0.48 |          0.08 |
'Array access using Fast2DArray' |   100 | 0.5683 ns | 0.0240 ns | 0.0931 ns |   0.56 |          0.12 |
'Array access using 2D array' |  1000 | 0.9641 ns | 0.0405 ns | 0.1569 ns |   1.00 |          0.00 |
'Array access using a jagged array' |  1000 | 0.6315 ns | 0.0350 ns | 0.1308 ns |   0.67 |          0.16 |
'Array access using Fast2DArray' |  1000 | 0.5521 ns | 0.0307 ns | 0.1147 ns |   0.58 |          0.14 |
'Array access using 2D array' | 10000 | 1.0596 ns | 0.0380 ns | 0.1470 ns |   1.00 |          0.00 |
'Array access using a jagged array' | 10000 | 0.5105 ns | 0.0173 ns | 0.0624 ns |   0.49 |          0.08 |
'Array access using Fast2DArray' | 10000 | 0.4680 ns | 0.0211 ns | 0.0819 ns |   0.45 |          0.09 |

tests/ImageSharp.Tests/Formats/GeneralFormatTests.cs
JimBobSquarePants: 👍 Why the hell that should be required though is a mystery. Mental.

src/ImageSharp.Formats.Png/PngEncoderCore.cs
JimBobSquarePants: You mean the PNG Encoder ? 

The `WuQuantizer` doesn't support dithering (for now) as I couldn't get the output correct so I switched it for the octree one which does. If you're encoding indexed png's you really want the output to be top notch so I thought the switch was important.

PR #107, opened 2017-02-17
Add NamedColors<TColor> and ColorBuilder<TColor>

tocsoft:
### Prerequisites

### Description

I noticed during the dithering PR that we had usages of the below patter for wellknown colors.
```c#
TColor packed = default(TColor);
packed.PackFromBytes(255, 255, 255, 255);
```
I've added 2 new classes, and fix up all the places that are using the above syntax where I can find them to use the improved syntax.

Additionally fixed up the Glow & Vignette usages to remove the usage of passing a default(TColor) around and then trying to compare to it, it seemed dangerous to me as whatever color default(TColor) actually ends up as could be what you want the glow to be and as it stood you could never use it.

#### `ColorBuild<TColor>`
Provides convenience methods for creating new `TColor`s from RGB byte and Hex strings allowing users to create any color using those mechanisms. 

#### `NamedColors<TColor>`
provides static readonly representations of wellknown colors.

### DIFF
--- a/src/ImageSharp.Processing/Processors/Overlays/GlowProcessor.cs
+++ b/src/ImageSharp.Processing/Processors/Overlays/GlowProcessor.cs
@@ -17,12 +17,11 @@ namespace ImageSharp.Processing.Processors
         where TColor : struct, IPackedPixel, IEquatable<TColor>
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="GlowProcessor{TColor}"/> class.
+        /// Initializes a new instance of the <see cref="GlowProcessor{TColor}" /> class.
         /// </summary>
-        public GlowProcessor()
+        /// <param name="color">The color or the glow.</param>
+        public GlowProcessor(TColor color)
         {
-            TColor color = default(TColor);
-            color.PackFromVector4(Color.Black.ToVector4());
             this.GlowColor = color;
         }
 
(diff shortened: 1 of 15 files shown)


src/ImageSharp.Processing/Processors/Overlays/GlowProcessor.cs
```diff
@@ -17,12 +17,11 @@ public class GlowProcessor<TColor> : ImageProcessor<TColor>
         where TColor : struct, IPackedPixel, IEquatable<TColor>
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="GlowProcessor{TColor}"/> class.
+        /// Initializes a new instance of the <see cref="GlowProcessor{TColor}" /> class.
         /// </summary>
-        public GlowProcessor()
+        /// <param name="color">The color or the glow.</param>
+        public GlowProcessor(TColor color)
```
JimBobSquarePants: Glad you fixed this and Vignette.

src/ImageSharp/Colors/Color.cs
```diff
@@ -70,28 +70,6 @@ public Color(byte r, byte g, byte b, byte a = 255)
         /// <summary>
         /// Initializes a new instance of the <see cref="Color"/> struct.
         /// </summary>
-        /// <param name="hex">
```
JimBobSquarePants: This is great! Being able to create any color from a hex is a powerful addition.

JimBobSquarePants:
No risk here so merging.

antonfirsov:
Awesome idea!

tocsoft:
@JimBobSquarePants It was mostly done with regex find and replaces, then around of trying to finding all the places they could be used.

PR #108, opened 2017-02-18
Use Fast2DArray for all convolution algorithms.

JimBobSquarePants:
### Prerequisites

### Description
This PR replaces all `float[][]` instances within the convolution algorithms with `Fast2DArray<float>` instances. We get a slight performance boost across the board due to the increased ability to use CPU caches but also a nicer API for the developer to consume as jagged arrays are clunky to create.

```

Allocated=0 B
```

Method | Count |      Mean |    StdErr |    StdDev |    Median | Scaled | Scaled-StdDev |
------------------------------------ |------ |---------- |---------- |---------- |---------- |------- |-------------- |
'Array access using 2D array' |    10 | 1.2493 ns | 0.0454 ns | 0.1698 ns | 1.2941 ns |   1.00 |          0.00 |
'Array access using a jagged array' |    10 | 0.6094 ns | 0.0475 ns | 0.1839 ns | 0.5901 ns |   0.50 |          0.16 |
'Array access using Fast2DArray' |    10 | 0.5754 ns | 0.0433 ns | 0.2289 ns | 0.4887 ns |   0.47 |          0.20 |
'Array access using 2D array' |   100 | 0.9539 ns | 0.0403 ns | 0.1560 ns | 0.9137 ns |   1.00 |          0.00 |
'Array access using a jagged array' |   100 | 0.5623 ns | 0.0360 ns | 0.1393 ns | 0.5224 ns |   0.60 |          0.17 |
'Array access using Fast2DArray' |   100 | 0.5368 ns | 0.0313 ns | 0.1172 ns | 0.5280 ns |   0.58 |          0.15 |
'Array access using 2D array' |  1000 | 0.9406 ns | 0.0472 ns | 0.1829 ns | 0.8760 ns |   1.00 |          0.00 |
'Array access using a jagged array' |  1000 | 0.4004 ns | 0.0393 ns | 0.1471 ns | 0.3516 ns |   0.44 |          0.18 |
'Array access using Fast2DArray' |  1000 | 0.0606 ns | 0.0207 ns | 0.1190 ns | 0.0000 ns |   0.07 |          0.13 |
'Array access using 2D array' | 10000 | 1.4894 ns | 0.0533 ns | 0.2612 ns | 1.4763 ns |   1.00 |          0.00 |
'Array access using a jagged array' | 10000 | 0.4642 ns | 0.0468 ns | 0.1813 ns | 0.3666 ns |   0.32 |          0.14 |
'Array access using Fast2DArray' | 10000 | 0.4734 ns | 0.0423 ns | 0.1639 ns | 0.4667 ns |   0.33 |          0.13 |

### DIFF
--- a/src/ImageSharp.Processing/Convolution/BoxBlur.cs
+++ b/src/ImageSharp.Processing/Convolution/BoxBlur.cs
@@ -7,7 +7,6 @@ namespace ImageSharp
 {
     using System;
 
-    using Processing;
     using Processing.Processors;
 
     /// <summary>
@@ -41,7 +40,7 @@ namespace ImageSharp
         public static Image<TColor> BoxBlur<TColor>(this Image<TColor> source, int radius, Rectangle rectangle)
             where TColor : struct, IPackedPixel, IEquatable<TColor>
         {
-            source.ApplyProcessor(new BoxBlurProcessor<TColor>(), rectangle);
+            source.ApplyProcessor(new BoxBlurProcessor<TColor>(radius), rectangle);
             return source;
         }
     }
--- a/src/ImageSharp.Processing/Processors/Convolution/BoxBlurProcessor.cs
+++ b/src/ImageSharp.Processing/Processors/Convolution/BoxBlurProcessor.cs
@@ -57,8 +57,8 @@ namespace ImageSharp.Processing.Processors
         {
             int size = this.kernelSize;
             Fast2DArray<float> kernel = horizontal
-                ? new Fast2DArray<float>(new float[1, size])
-                : new Fast2DArray<float>(new float[size, 1]);
+                ? new Fast2DArray<float>(size, 1)
+                : new Fast2DArray<float>(1, size);
 
             float sum = 0F;
             for (int i = 0; i < size; i++)
--- a/src/ImageSharp.Processing/Processors/Convolution/BoxBlurProcessor.cs
+++ b/src/ImageSharp.Processing/Processors/Convolution/BoxBlurProcessor.cs
@@ -35,12 +35,12 @@ namespace ImageSharp.Processing.Processors
         /// <summary>
         /// Gets the horizontal gradient operator.
         /// </summary>
-        public float[][] KernelX { get; }
+        public Fast2DArray<float> KernelX { get; }
 
         /// <summary>
         /// Gets the vertical gradient operator.
         /// </summary>
-        public float[][] KernelY { get; }
+        public Fast2DArray<float> KernelY { get; }
 
         /// <inheritdoc/>
         protected override void OnApply(ImageBase<TColor> source, Rectangle sourceRectangle)
@@ -52,46 +52,42 @@ namespace ImageSharp.Processing.Processors
         /// Create a 1 dimensional Box kernel.
         /// </summary>
         /// <param name="horizontal">Whether to calculate a horizontal kernel.</param>
-        /// <returns>The <see cref="T:float[][]"/></returns>
-        private float[][] CreateBoxKernel(bool horizontal)
+        /// <returns>The <see cref="Fast2DArray{T}"/></returns>
+        private Fast2DArray<float> CreateBoxKernel(bool horizontal)
         {
             int size = this.kernelSize;
-            float[][] kernel = horizontal ? new float[1][] : new float[size][];
+            Fast2DArray<float> kernel = horizontal
+                ? new Fast2DArray<float>(new float[1, size])
+                : new Fast2DArray<float>(new float[size, 1]);
 
-            if (horizontal)
-            {
-                kernel[0] = new float[size];
-            }
-
-            float sum = 0.0f;
-
+            float sum = 0F;
             for (int i = 0; i < size; i++)
             {
                 float x = 1;
                 sum += x;
                 if (horizontal)
                 {
-                    kernel[0][i] = x;
+                    kernel[0, i] = x;
                 }
                 else
                 {
-                    kernel[i] = new[] { x };
+                    kernel[i, 0] = x;
                 }
             }
 
-            // Normalise kernel so that the sum of all weights equals 1
+            // Normalize kernel so that the sum of all weights equals 1
             if (horizontal)
             {
                 for (int i = 0; i < size; i++)
                 {
-                    kernel[0][i] = kernel[0][i] / sum;
+                    kernel[0, i] = kernel[0, i] / sum;
                 }
             }
             else
             {
                 for (int i = 0; i < size; i++)
                 {
-                    kernel[i][0] = kernel[i][0] / sum;
+                    kernel[i, 0] = kernel[i, 0] / sum;
                 }
             }
 
--- a/src/ImageSharp.Processing/Processors/Convolution/Convolution2DProcessor.cs
+++ b/src/ImageSharp.Processing/Processors/Convolution/Convolution2DProcessor.cs
@@ -21,7 +21,7 @@ namespace ImageSharp.Processing.Processors
         /// </summary>
         /// <param name="kernelX">The horizontal gradient operator.</param>
         /// <param name="kernelY">The vertical gradient operator.</param>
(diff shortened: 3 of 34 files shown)


antonfirsov:
It was very strange to me the difference difference between jagged and Fast2DArray is that slight, so I looked at your benchmark, and was like 
oh nooo,

Nanobenchmarks

There is simply too much noise: JIT is a stupid guy, it will generate tons of useless instructions around the real stuff we want to measure. So I [made a modified version](https://gist.github.com/antonfirsov/dd26929efeb62ad177a0bc55779e2931) of the benchmarks, also including a raw 1D variant.

With these we have results that make sense!

Method | Count |       Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev |
------------------------------------ |------ |----------- |---------- |---------- |------- |-------------- |
FlatArrayIndex |     4 | 12.4112 ns | 0.1515 ns | 0.6061 ns |   0.63 |          0.03 |
'Array access using 2D array' |     4 | 19.6437 ns | 0.0178 ns | 0.0689 ns |   1.00 |          0.00 |
'Array access using a jagged array' |     4 | 16.3029 ns | 0.0202 ns | 0.0781 ns |   0.83 |          0.00 |
'Array access using Fast2DArray' |     4 | 12.5566 ns | 0.0190 ns | 0.0658 ns |   0.64 |          0.00 |
 | | | | | | |
FlatArrayIndex |    16 | 12.8350 ns | 0.1215 ns | 0.4707 ns |   0.63 |          0.03 |
'Array access using 2D array' |    16 | 20.3895 ns | 0.2246 ns | 0.8700 ns |   1.00 |          0.00 |
'Array access using a jagged array' |    16 | 16.2835 ns | 0.0276 ns | 0.1068 ns |   0.80 |          0.03 |
'Array access using Fast2DArray' |    16 | 12.9458 ns | 0.0320 ns | 0.1240 ns |   0.64 |          0.03 |
 | | | | | | |
FlatArrayIndex |   128 | 12.0066 ns | 0.0161 ns | 0.0625 ns |   0.60 |          0.01 |
'Array access using 2D array' |   128 | 19.8725 ns | 0.0476 ns | 0.1844 ns |   1.00 |          0.00 |
'Array access using a jagged array' |   128 | 16.7680 ns | 0.0909 ns | 0.3522 ns |   0.84 |          0.02 |
'Array access using Fast2DArray' |   128 | 12.9790 ns | 0.0630 ns | 0.2270 ns |   0.65 |          0.01 |

We can notice, that there is almost no difference between the flat and the Fast2DArray variant.

For contrast: on my machine the original benchmark results are [really strange and misleading](https://gist.github.com/antonfirsov/97649cc7d069d47e4fce411ff0d08401)!

Should I add these benc

antonfirsov:
Okkay, partially undoing my previous comment ...

Made furthere experiments, the situation is more complicated.
[If I complicate the array access](https://gist.github.com/antonfirsov/3a9564f1f4f01f7f403d3c19344cca04), results are more similar to your original ones:
Method | Count |        Mean |    StdDev | Scaled | Scaled-StdDev |
-------------------------------------------- |------ |------------ |---------- |------- |-------------- |
'Emulated 2D array access using flat array' |     4 |  12.0685 ns | 0.1085 ns |   0.64 |          0.01 |
'Array access using 2D array' |     4 |  18.9276 ns | 0.0675 ns |   1.00 |          0.00 |
'Array access using a jagged array' |     4 |  14.7283 ns | 0.2659 ns |   0.78 |          0.01 |
'Array access using Fast2DArray' |     4 |  13.7405 ns | 0.0810 ns |   0.73 |          0.00 |
'Emulated 2D array access using flat array' |    16 |  77.2414 ns | 1.1006 ns |   0.65 |          0.01 |
'Array access using 2D array' |    16 | 119.5516 ns | 1.4198 ns |   1.00 |          0.00 |
'Array access using a jagged array' |    16 |  90.2572 ns | 0.7899 ns |   0.76 |          0.01 |
'Array access using Fast2DArray' |    16 |  89.7342 ns | 2.7663 ns |   0.75 |          0.02 |
'Emulated 2D array access using flat array' |   128 |  76.6352 ns | 0.4883 ns |   0.65 |          0.01 |
'Array access using 2D array' |   128 | 117.3613 ns | 1.9309 ns |   1.00 |          0.00 |
'Array access using a jagged array' |   128 |  88.4352 ns | 1.0197 ns |   0.75 |          0.01 |
'Array access using Fast2DArray' |   128 |  85.6849 ns | 1.1572 ns |   0.73 |          0.02 |

My explanation is that JIT fails to apply the bounds check elimination here. (As it does in your original benchmark.) Confused now about my knowledge about JIT, but still beeing untrustful towards small nanobenchmarks anyways...

JimBobSquarePants:
You're absolutely right there, this was a lazy benchmark. 😈 

Was trying to push the changes before you did your work to change the typeparams so we didn't get merge conflicts. 

I would explain the closeness in that despite us using the 1D array we can never eliminate the bounds checks since the calling method has no knowledge of the underlying array bounds despite inlining.

For me this PR is more about API improvements than anything. Creating and populating a 5x1 jagged array is horrible compared to a 2D array and we are able to better the perf and provide the nicer API.

You wanna push your better benchmark in and we can merge this?

tocsoft:
I felt the syntax around using the Fast2DArray was a little clunky, so I've added in an implicit cast that reduces the boilerplate used everywhere.

tocsoft:
or almost did until @antonfirsov beat me to pushing 😛

JimBobSquarePants:
@tocsoft can we revert this change please and introduce any changes incrementally. The strict was static for a reason. We're now creating a new 1D array per instance.

tocsoft:
I'm confused everything is still static just as static, all the change does is use the implicit conversion instead of  just uses the implicit conversion instead of the constructor directly?

This change isn't instantiating anything any more than the original code did. In fact adding the width/height constructor actually saves us some memory as we don't need to create the `new float [1, size]` arrays at all.

tocsoft:
Lets try this again in english (I fail at typing today.):

I'm confused everything is still just as static. All my change does is use the implicit conversion instead of the constructor directly, thusly is just syntactic sugar around your original code.

The only addition is to also have a constructor which take a height & width to same the additional allocation of the `new float[1, size]` arrays which are immediately discarded.

JimBobSquarePants:
The implicit operator creates a new instance of Fast2DArray. This is only called per instance of the processors. Before we created only one instance of Fast2DArray to use for all. Each Fast2DArray constructor creates a new 1D array also.

JimBobSquarePants:
The width height constructor is fine. It's the private static changes are the issue.

tocsoft:
No the implicit operator is being called per static instance of the class exactly the same as the construct was before it... It would only be called per instance if I had changed the `private static` to just `private` that would be a different matter.

For example 

```c#
private static readonly Fast2DArray<float> KayyaliX =
             new float[,]
             {
                 { 6, 0, -6 },
                 { 0, 0, 0 },
                 { -6, 0, 6 }
             };
```
The above line will only be executed once, and after it is executed you will be left with a `Fast2DArray<float>` in a static field.. it will never be called again, it will never run the implicit operator again.... by shear fact that it is a `static readonly` field means that it would be impossible for the implicit operator to be called to set it again.

tocsoft:
@JimBobSquarePants try out this https://dotnetfiddle.net/EnPIaw to demonstrate that the implicit operator is only called once even for 100 instantiations of a processor

JimBobSquarePants:
Och @tocsoft sorry! Totally misread the diff!

Thought you had changed the static versions to this and was casting in the constructor. YOu change is great!
```

private static readonly new float[,] KirschWest =
              new float[,]
              {
                 { 5, -3, -3 },
                 { 5,  0, -3 },
                 { 5, -3, -3 }
              });
              };
```

Never try to read diffs on your mobile...

JimBobSquarePants:
Is everyone happy to merge?

antonfirsov:
Seems ok to me.

tocsoft:
@JimBobSquarePants no problem... you did have me doubting myself for a moment.

PR #109, opened 2017-02-18
Simplified common pixel interfaces, introducing ArrayPointer<T> 

antonfirsov:
### Prerequisites

### Description
To make the changes more incremental, decided to push the pixel interface refactor before starting the `BulkPixelOperations<T>` madness.
- Simplified interfaces of "packed" pixels (`IPixel`, `IPixel<TSelf>`) ---> Simplified where constraints: `where TColor : struct, IPixel<TColor>`
- Removed all unused interfaces (also `IPackedVector<T>`!)
- Introduced `ArrayPointer<T>`.


src/ImageSharp/Colors/PackedPixel/IPixel.cs
```diff
@@ -0,0 +1,68 @@
+﻿namespace ImageSharp
+{
+    using System;
+    using System.Numerics;
+
+    public interface IPixel
```
JimBobSquarePants: What happens to `PackedValue` ? We need this to be enforced by an interface so that we can provide signature compatibility to MonoGame types. That's why `IPackedVector<T>` still existed despite us not using it.

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "projects": [ "src" ],
     "sdk": {
-    "version": "1.0.0-preview2-003131"
+    "version": "1.0.0-preview2-003121"
```
JimBobSquarePants: Everyone should probably have the later version now. Was thinking of bumping dependencies to 4.3.0 also soon.

src/ImageSharp/Colors/PackedPixel/IPixel.cs
antonfirsov: The generic version does not seem to provide any kind of useful polymorphic behaviour to me. Looks like a legacy concept even in monogame. Having interfaces only to enforce signature (not for the sake of polymorphism) ...  not my favorite stuff.

antonfirsov:
@dlemstra now the exception is thrown only in debug mode, which makes one of the tests failing on CI.  Theoretical question of the day: do we need coverage for debug only stuff?

src/ImageSharp/Colors/PackedPixel/IPixel.cs
JimBobSquarePants: We don't use the property but I'm sure Monogame must. We, of course,  don't want to implement it for our optimized types nor use the interface for typeparams. Time to check the Monogamr source.

dlemstra:
@antonfirsov I don't think we need coverage for debug only stuff. I did not run the tests because I thought they would not be affected.

src/ImageSharp/Colors/PackedPixel/IPixel.cs
antonfirsov: Browsing it online it seemed to me they use the generic interface only in reflection code. Have not checked with VS yet.

JimBobSquarePants: The only direct reference to the property is here

https://github.com/MonoGame/MonoGame/blob/7c3d6870a38f8a5e479e64d935d692f2610e1cda/MonoGame.Framework/Net/PacketReader.cs#L84

https://github.com/MonoGame/MonoGame/blob/7c3d6870a38f8a5e479e64d935d692f2610e1cda/MonoGame.Framework/Net/PacketWriter.cs#L74

All I'm thinking is that all the existing `IPackedVector<TPacked>` implementations could continue as is but with the interface separate from `IPixel<TColor>`. I know it seems really odd but I really want to ensure that we have the greatest range of coverage possible when considering compatibility.

antonfirsov:
@JimBobSquarepants @dlemstra @tocsoft Any other change requests except bringing back  IPackedVector<T>?

JimBobSquarePants:
@antonfirsov None. all looks great to me.

tocsoft:
looks good to me, definitely like the reduced generic constraints.

PR #110, opened 2017-02-19
Support for VSCode

Toxantron:
Because I develop on Linux I noticed the library builds, but came without support for the Linux .Net Core IDE. I added two tasks: _build_ and _test_ and linked it to `ctrl + shift + b`.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,24 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
+  // for the documentation about the tasks.json format
+  "version": "0.1.0",
+  "command": "dotnet",
+  "isShellCommand": true,
+  "args": [],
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Good idea! I'm assuming this is all tested and works? 

I don't have a Linux install to test on and even if I did I'd probably immediately break it with my amateurish keyboard banging.

antonfirsov:
I think a VSCode installation is enough to test this,  we don't even need linux.

Toxantron:
@JimBobSquarePants I use it with my installation and assume it should work for everyone else that can compile the projects with Visual Studio or the CLI.

tocsoft:
Thanks @Toxantron 🎉 

All merged in, nice little addition. Works on VSCode on windows, and it has no effect on Visual studio or the CI build 👍

PR #111, opened 2017-02-20
Standardize to netstandard1.6

Toxantron:
## Suggestion only
I am not sure if this matches your vision of this library. I figured since it is relatively new and targeted at .Net Core and Cross-Platform Development it makes sense to drop older .NetStandard targets and focus on 1.6 or even better 2.0. [This blog](https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/) illustrates the topic a little bit.

I noticed a lot of package downgrades and version inconsistencies using the packages in ASP.Net Core on Linux and thought since you have dedicated targets for the full .Net Frameworks you might be interested in an updated version that seamlessly integrates into ASP.Net Core 1.1

I ran all builds and unit tests on my machine, but am curious what your CI has to say about it.

```sh
$ dotnet --version
1.0.0-preview2-1-003177
```

Created:

```sh
Project ImageSharp (.NETStandard,Version=v1.6) was previously compiled. Skipping compilation.
Project ImageSharp.Processing (.NETStandard,Version=v1.6) was previously compiled. Skipping compilation.
Project ImageSharp.Drawing (.NETStandard,Version=v1.6) was previously compiled. Skipping compilation.
Project ImageSharp.Drawing.Paths (.NETStandard,Version=v1.6) was previously compiled. Skipping compilation.
Project ImageSharp.Formats.Png (.NETStandard,Version=v1.6) was previously compiled. Skipping compilation.
Project ImageSharp.Formats.Jpeg (.NETStandard,Version=v1.6) was previously compiled. Skipping compilation.
Project ImageSharp.Formats.Bmp (.NETStandard,Version=v1.6) was previously compiled. Skipping compilation.
Project ImageSharp.Formats.Gif (.NETStandard,Version=v1.6) was previously compiled. Skipping compilation.
Project ImageSharp.Tests (.NETCoreApp,Version=v1.1) was previously compiled. Skipping compilation.
xUnit.net .NET CLI test runner (64-bit .NET Core ubuntu.16.04-x64)
  Discovering: ImageSharp.Tests
  Discovered:  ImageSharp.Tests
  Starting:    ImageSharp.Tests
  Finished:    ImageSharp.Tests
=== TEST EXECUTION SUMMARY ===
   ImageSharp.Tests  Total: 1585, Errors: 0, Failed: 0, Skipped: 0, Time: 307.683s
```

If this does not match your vision or you dislike it for other reasons, don't hesitate to close this.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -8,7 +8,7 @@
   "tasks": [
     {
       "taskName": "build",
-      "args": [ "src/*/project.json", "-f", "netstandard1.1" ],
+      "args": [ "src/*/project.json", "-f", "netstandard1.6" ],
       "isBuildCommand": true,
       "showOutput": "always",
       "problemMatcher": "$msCompile"
--- a/src/ImageSharp.Drawing.Paths/project.json
+++ b/src/ImageSharp.Drawing.Paths/project.json
@@ -49,28 +49,28 @@
       "version": "1.0.0",
       "type": "build"
     },
-    "System.Buffers": "4.0.0",
-    "System.Runtime.CompilerServices.Unsafe": "4.0.0"
+    "System.Buffers": "4.3.0",
+    "System.Runtime.CompilerServices.Unsafe": "4.3.0"
   },
   "frameworks": {
-    "netstandard1.1": {
+    "netstandard1.6": {
       "dependencies": {
-        "System.Collections": "4.0.11",
-        "System.Diagnostics.Debug": "4.0.11",
-        "System.Diagnostics.Tools": "4.0.1",
-        "System.IO": "4.1.0",
-        "System.IO.Compression": "4.1.0",
-        "System.Linq": "4.1.0",
-        "System.Numerics.Vectors": "4.1.1",
-        "System.ObjectModel": "4.0.12",
-        "System.Resources.ResourceManager": "4.0.1",
-        "System.Runtime.Extensions": "4.1.0",
-        "System.Runtime.InteropServices": "4.1.0",
-        "System.Runtime.Numerics": "4.0.1",
-        "System.Text.Encoding.Extensions": "4.0.11",
-        "System.Threading": "4.0.11",
-        "System.Threading.Tasks": "4.0.11",
-        "System.Threading.Tasks.Parallel": "4.0.1"
+        "System.Collections": "4.3.0",
+        "System.Diagnostics.Debug": "4.3.0",
+        "System.Diagnostics.Tools": "4.3.0",
+        "System.IO": "4.3.0",
+        "System.IO.Compression": "4.3.0",
+        "System.Linq": "4.3.0",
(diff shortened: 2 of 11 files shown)


dlemstra:
The library is specifically designed to run on as many platforms as possible. We would like to keep it at netstandard 1.1.

JimBobSquarePants:
First of I'll just note. Gitter is the correct channel for discussions like this as per the contribution guidelines. We now have a code coverage bot joining in.

I have plans for fixing the dependency graph when we switch to VS2017. I can reference netstandard 1.6 yet create our 1.1, 1.3 etc targets much more easily there. Doing so before that switch is a waste of resources.

Toxantron:
@dlemstra I understand that. But the .NetStandard1.1 target only covers Framework 4.5 and above which is also  included explicitly in the `project.json`. The only thing you lose is compatibility with approximately 12 Windows Phones and Silverlight which will be probably be replaced by the time this "goes public".

I apoligize for putting this in the wrong place. ;-)

JimBobSquarePants:
No worries 😄 

The way we have it now is definitely incorrect and needs to be changed to netstandard1.6 and it will. I will be keeping the low targets though. I have one of those 12 phones. 😝

Toxantron:
If we find the other 11 we might get this issue resolve quicker than I thought. :-P

Toxantron:
Oh and by the way. You already helped my project a lot which I switched from ImageMagiQ to your library for cross-platform compatibility. ;-)

dlemstra:
Yeah mean the other 10 right? I also got one of those. And I really hope someone will come and help me with making Magick.NET cross-platform.

Toxantron:
Well, if I was as talented with image processing as I apparently am with picking my jokes, I would love to. 

You did do a hell of a job on Windows which is why I picked it out of habit, but all I needed was resizing jpegs without any native platform dependencies and this just does the job.

Toxantron:
@dlemstra unless there is something concrete I could help out with?

dlemstra:
I could use help with this issue: https://github.com/dlemstra/Magick.NET/issues/16. This is for creating a Linux compilation script that I can use to build the native library. Would love to get some help with that 😃

Toxantron:
Sure, I'll take a look tomorrow night. Unfortunately if this works I am back to square one having to choose between two libraries. :-P

Toxantron:
@JimBobSquarePants back to this issue. Would you be OK with a PR that only adds what I did instead of replacing **netstandard1.1**? This would let you two and your ten friends use Windows phone while .Net Core projects have a clean 1.6 reference.

JimBobSquarePants:
@Toxantron  As long as you can assure me that we don't lose the ability to install against our currently defined targets and also that the test runner will still run. Every time I've tried to upgrade our dependencies the test runner has broken in VS2015

PR #112, opened 2017-02-21
Update dependencies to 4.3

Toxantron:
As discussed in #111 this adds **netstandard1.6** to all packages without removing the old packages. I assume @JimBobSquarePants will clean-up the remaining inconsistencies between **netstandard1.1** and **netstandard1.3** in the next few weeks with VS2017. Till then this could be a compromise for pure .NetCore projects.

Building for **netstandard1.6** gives a few warning and package downgrades but it works.

**Edit:** This was a pretty dumb idea.


Toxantron:
Well, that worked like a charm...

tocsoft:
My vote is we don't merge this at all and feel we don't need a netstandard1.6 target.

You should always try and target the lowest netstandard version that your codebase needs not the highest. Ours is currently fully compatible with netstandard1.1 when we don't use the filesystem APIs and netstandard1.3 with them.

I don't believe we should even be targeting the full frameworks at all (but that's a fight for another day) so I defiantly disagree with the netstandard1.6 target.

With regards upping the dependency versions to 4.3 that's a separate issue and is on the road map to be done with the port to the 2017/csproj based build system and is something I feel we should do.

All that said, I would like to thank you for your contribution and ultimately its @JimBobSquarePants decision with him being 'Grand High Eternal Dictator' 😁.

Toxantron:
@tocsoft I am starting to agree. While this works just fine for me it seems to cause more harm than good.

Toxantron:
I might have gone at this the wrong way. Instead of upping the target framework upping the dependencies makes more sense. Well, a lot of trouble for nothing. :-)

Toxantron:
@JimBobSquarePants I am starting to understand why you postponed the whole version thing. ;-)

JimBobSquarePants:
:smile: Yeah it's been troublesome. Certain we can sort it all soon though, thanks for trying!

PR #117, opened 2017-02-27
Benchmark running on .NetCore1.1

Toxantron:
### Prerequisites

### Description
The minimal changes required for the project to build and execute benchmarks on Linux with NetCore 1.1

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
-  "projects": [ "src" ],
-    "sdk": {
+  "projects": [ "src", "tests" ],
+  "sdk": {
     "version": "1.0.0-preview2-003121"
   }
 }
--- a/tests/ImageSharp.Benchmarks/project.json
+++ b/tests/ImageSharp.Benchmarks/project.json
@@ -14,7 +14,6 @@
     "allowUnsafe": true
   },
   "dependencies": {
-    "BenchmarkDotNet.Diagnostics.Windows": "0.10.1",
     "ImageSharp": {
       "target": "project"
     },
@@ -46,9 +45,22 @@
   "frameworks": {
     "net46": {
       "dependencies": {
+        "BenchmarkDotNet.Diagnostics.Windows": "0.10.1"
       },
       "frameworkAssemblies": {
         "System.Drawing": ""
+      }
+    },
+    "netcoreapp1.1": {
+      "dependencies": {
+        "Microsoft.NETCore.App": {
+          "type": "platform",
+          "version": "1.1.0-*"
+        },
+        "BenchmarkDotNet": "0.10.2",
+        "CoreCompat.System.Drawing": "1.0.0-beta006",
+        "runtime.linux-x64.CoreCompat.System.Drawing": "1.0.0-beta009",
+        "System.Reflection": "4.3.0"
       }
     }
   }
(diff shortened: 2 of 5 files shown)


antonfirsov:
Pushed a small change:
Needed `"BenchmarkDotNet.Core": "0.10.2"` reference to make it work on Windows, dropped it into this PR. Hope it's still working fine on your linux machine.

global.json
```diff
@@ -1,6 +1,6 @@
 {
-  "projects": [ "src" ],
-    "sdk": {
+  "projects": [ "src", "tests" ],
```
antonfirsov: @JimBobSquarePants does this change break something? Do you know why "tests" was not present in the original list?

antonfirsov:
I have no idea why did appveyor fail on my commit :(

With these set of changes, the only strange thing I got is the following message when I run the benchmarks from command line with `dotnet`:
```
`Assembly ImageSharp.Benchmarks, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null which defines benchmarks references non-optimized CoreCompat.System.Drawing`
```
Have no explanation for this, results are almost the same.

global.json
JimBobSquarePants: Ummmmm.... I think it was because I was having versioning problems with xunit. This might be ok now.

tests/ImageSharp.Benchmarks/project.json
```diff
@@ -46,10 +45,23 @@
...
+    },
+    "netcoreapp1.1": {
+      "dependencies": {
+        "Microsoft.NETCore.App": {
+          "type": "platform",
+          "version": "1.1.0-*"
+        },
+        "BenchmarkDotNet": "0.10.2",
+        "CoreCompat.System.Drawing": "1.0.0-beta006",
+        "runtime.linux-x64.CoreCompat.System.Drawing": "1.0.0-beta009",
```
JimBobSquarePants: I think it might pull for all.

Toxantron:
@antonfirsov that is weird because it is a dependency and should be pulled automatically.

The error message is OK, I get the same. It is due to the missing optimizer flag in their project.

tests/ImageSharp.Benchmarks/project.json
Toxantron: It does. Maybe I should fix that. Because there is a version for Mac as well.

global.json
Toxantron: I needed it for BenchmarkDotNet to reference the Benchmark project.json.

Toxantron:
But @antonfirsov to answer your question. The change still works on Linux. ;-)

Toxantron:
As for appveyor that makes no sense - unless it failed in the packing step because of the new dependency. Which makes only little sense.

tests/ImageSharp.Benchmarks/project.json
Toxantron: After a little research it seems that we can safely reference all platforms it shall run on. It will simply choose the right native libs at runtime to bind to.

JimBobSquarePants:
It failed due to 160 fails unit tests. They all seem to involve the Drawing package. 

See the console here https://ci.appveyor.com/project/JamesSouth/imagesharp/build/1.0.0.570

antonfirsov:
Burlesque mode on: Reverted the change, now it seems to be OK without the additional dependency. 
I've been playing with Job attributes `[CoreJob, ClrJob]` - maybe there was a configuration where they were working only with the extra dependency.
If one other contributor could try this on Windows, and confirm that everything is OK, I think we should merge this, and continue our experiments from there. @JimBobSquarePants thoughts?

JimBobSquarePants:
The appveyor build is windows so we know it works for now. Happy to merge and experiment in the future

PR #124, opened 2017-03-05
Add ImageSharp.Drawing.Text package

### LINKED ISSUE #38: Font Support
The library will at some point need to add font support.

Libraries to consider basing the system on.

- https://github.com/MikePopoloski/SharpFont
- https://github.com/LayoutFarm/Typography

Ideally we shouldn't have any dependencies and absolutely no unmanaged dependencies as we need to deploy easily on web applications.

tocsoft:
### Prerequisites

Fixes #38 
### Description

This provides a new extension method. `DrawText(....)` by introducing a dependency onto [SixLabors.Fonts](https://github.com/sixlabors/fonts) which handles most of the heavy lifting.

The way the code works is first we use SixLabors.Fonts to deal with the fonts, it render to a specific interface that renders lines and curves to generate figures, ImageSharp has a thin layer which proxies those calls to SixLabors.Shapes to build the shapes before passing the final shapes over to the ImageSharp.Drawing.Paths package to render to the image.

### How to use

Before you can draw text you need to create a `FontCollection` and install your font files then create a new font instances with your required settings (see http://fonts.sixlabors.com/docs/getting-started for details)

```c#
FontCollection collection = new FontCollection(); 
Font details = collection.Install("path/to/font.ttf");
```
> Installing fonts is an expensive operation you will want to keep the `FontCollection` around for a long as possible preferably for the lifetime of your application. 

You can then render text using the `DrawText` extensions.

```c#
Font specificFont = new Font(details, 40); //copy the default font but set the size to 40pt
using (var img = new Image(200, 50))
{
    img.DrawText("Hello World", specificFont, Color.Red, new Vector2(10, 10));
    img.Save("output.png");
}
```


src/ImageSharp.Drawing.Text/GlyphBuilder.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        /// <summary>
+        /// Gets the paths that have been rendered by this.
+        /// </summary>
+        public IEnumerable<IPath> Paths => this.paths;
+
+        /// <summary>
+        /// Begins the glyph.
+        /// </summary>
+        void IGlyphRenderer.BeginGlyph()
```
JimBobSquarePants: Prefix with private here please. Not a fan of using the implicit declaration.

Toxantron: I think you missed, that it is an explicit interface implementation. They don't have access modifiers.

JimBobSquarePants: Indeed I did. Not used to seeing them. I'm assuming we want to keep them private then?

Toxantron: They are neither private nor public. They are as accessible as the interface that declares the method because you have to cast the object to the interface to access the method.

tocsoft: I did it to make the class only exposes the Paths property by default as there is no code inside ImageSharp that should ever call those methods (the only thing that should call them is something deep inside SixLabors.Fonts).

I can make them public implicit implementations if you prefer (can't make them private as they are interface methods after all).

JimBobSquarePants: @tocsoft No, keep it as-is with the required cast. Best not to expose anything we don't explicitly want to.

JimBobSquarePants:
@tocsoft You happy to merge this? j'approuve.

tocsoft:
And there was go, we now have text drawing. 😀

PR #125, opened 2017-03-05
Introducing PinnedBuffer<T> to manage pinned (and optionally pooled) arrays as disposable resources + Introducing BulkedPixelOperations API.

antonfirsov:
## Prerequisites

#### `PinnedBuffer<T>`
The scenario described in the title is a common concern across the library. Classes like `PixelAccessor<T>` or `PixelArea<T>` seemed to duplicate the same logic (with all the complicated managed/unmanaged dispose switch and the `isPooled` member). It seemed to be wise to move this concern to it's own class. 

Other benefits:
- It can be reused in many other places, like pinning stuff in `JpegPixelArea`
- `ArrayPointer<T>` could be seen as a slice over `PinnedBuffer<T>`. It will be useful to simplify code around bulk operations.

This change implicated the need for generelizing `PixelPool<T>` to non `IPixel<T>`-bound `PixelDataPool<T>`. It's main resposibility is to provide a simple way to access the library's common array pools ensuring that `Rent()` will always return an array initilized to `default(T)`.

#### `BulkPixelOperations<T>`
The API is introduced, but not used yet. I wanted to PR the `PinnedBuffer<T>` changes first to keep my work as incremental as possible.


antonfirsov:
Can anyone help me investigating why do my changes break `NamedColors<Color>` stuff on appveyor? It is totally OK on may machine. I have no idea what should I do :(

src/ImageSharp/Colors/Color.cs
```diff
@@ -112,6 +112,9 @@ public Color(uint packed)
             this.packedValue = packed;
         }
 
+        /// <inheritdoc />
+        public BulkPixelOperations<Color> BulkOperations => new BulkPixelOperations<Color>();
```
tocsoft: shouldn't these be reading a static instance of `BulkPixelOperations<T>` ? otherwise it'll be really easy to allocate multiple instances from a public API.

The way I see it is it should be more like this
``` c#
private static readony BulkPixelOperations<Color> PrivateBulkOperations = new BulkPixelOperations<Color>();

public BulkPixelOperations<Bgra5551> BulkOperations => PrivateBulkOperations;
```

That way we still get to use the hack to allow for swapping out to more efficiant versions but we wont accidentally allocate a new `BulkPixelOperations<T>` every time you allocate the property on this or any other instance of the struct.

This obviously applies to all `TColor`s

antonfirsov: Yeah it would better to allocate only one instance. 
BulkPixelOperations<T> is stateless + an instance is rarely requested so allocating it multiple times did not seemed to be a big issue for me.

antonfirsov: Wouln't it be better to add a comment into `IPixel.BulkOperations` instead, to point the user to `BulkPixelOperations<T>.Instance`?
I seems unlikely to me this will ever cause a trouble, and the current solution is much better in terms of simplicity.

tocsoft: As we are are newing something up, maybe make the Interface a method named `CreateBulkOperations()` instead of a property, makes people realise that it has something they shouldn't just dot into to use.

And the comment pointing people the the preferred accessor would also help.

JimBobSquarePants:
@antonfirsov You've been busy!

I had a look at the two builds to see if I could figure out why the tests were failing the first time and not the second. No idea!

src/ImageSharp/Colors/PackedPixel/BulkPixelOperations{TColor}.cs
```diff
@@ -0,0 +1,256 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Bulk version of <see cref="IPixel.ToXyzBytes(byte[], int)"/>.
+        /// </summary>
+        /// <param name="sourceColors">The <see cref="ArrayPointer{TColor}"/> to the source colors.</param>
+        /// <param name="destBytes">The <see cref="ArrayPointer{Byte}"/> to the destination bytes.</param>
+        /// <param name="count">The number of pixels to convert.</param>
+        internal virtual void PackToXyzBytes(ArrayPointer<TColor> sourceColors, ArrayPointer<byte> destBytes, int count)
```
JimBobSquarePants: Drop the word "Pack" here as per my commit comments.

src/ImageSharp/Colors/PackedPixel/Byte4.cs
```diff
@@ -70,6 +70,9 @@ public Byte4(float x, float y, float z, float w)
         }
 
         /// <inheritdoc />
+        public BulkPixelOperations<Byte4> CreateBulkOperations() => new BulkPixelOperations<Byte4>();
```
JimBobSquarePants: Is this enforced by the IPackedPixel interface or any other interfaces?

src/ImageSharp/Common/Memory/PinnedBuffer{T}.cs
```diff
@@ -0,0 +1,164 @@
...
+    using System;
+    using System.Buffers;
+    using System.Runtime.InteropServices;
+
+    /// <summary>
+    /// Manages a pinned buffer of value type data 'T' as a Disposable resource.
+    /// The backing array is either pooled or comes from the outside.
+    /// </summary>
+    /// <typeparam name="T">The value type.</typeparam>
+    internal class PinnedBuffer<T> : IDisposable
```
JimBobSquarePants: I like this. Definitely best to reuse code.

src/ImageSharp/Common/Memory/PinnedBuffer{T}.cs
```diff
@@ -0,0 +1,164 @@
...
+        where T : struct
+    {
+        /// <summary>
+        /// A handle that allows to access the managed <see cref="Array"/> as an unmanaged memory by pinning.
+        /// </summary>
+        private GCHandle handle;
+
+        /// <summary>
+        /// A value indicating wether this <see cref="PinnedBuffer{T}"/> instance should return the array to the pool.
```
JimBobSquarePants: "whether"

src/ImageSharp/Common/Memory/PixelDataPool{T}.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Heuristically calculates a reasonable maxArrayLength value for the backing <see cref="ArrayPool"/>.
+        /// </summary>
+        /// <returns>The maxArrayLength value</returns>
+        internal static int CalculateMaxArrayLength()
+        {
+            // ReSharper disable once SuspiciousTypeConversion.Global
+            if (default(T) is IPixel)
+            {
+                const int MaximumExpectedImageSize = 16384;
```
JimBobSquarePants: Same as commit query, how was this value picked? It's gonna be something I probably should know I'm sure.

src/ImageSharp/Image/ImageBase{TColor}.cs
```diff
@@ -162,13 +162,15 @@ public virtual PixelAccessor<TColor> Lock()
         internal void SwapPixelsBuffers(PixelAccessor<TColor> pixelSource)
         {
             Guard.NotNull(pixelSource, nameof(pixelSource));
-            Guard.IsTrue(pixelSource.PooledMemory, nameof(pixelSource.PooledMemory), "pixelSource must be using pooled memory");
 
+            // TODO: Was this check really useful? If yes, we can define a bool PixelAccessor<TColor>.IsBoundToImage to re-introduce it:
```
JimBobSquarePants: @tocsoft What was the reasoning behind this?

src/ImageSharp/Common/Memory/PixelDataPool{T}.cs
antonfirsov: It's 2^14.
Never seen an image larger than 8192*8192, so picked the next power-of-two value.

src/ImageSharp/Colors/PackedPixel/Byte4.cs
antonfirsov: Yeah, it's defined by `IPixel<T>`.

src/ImageSharp/Common/Memory/PixelDataPool{T}.cs
JimBobSquarePants: Ah right thanks... Anything bigger than that they would be better off using something taht supported file based paging anyway.

src/ImageSharp/Image/ImageBase{TColor}.cs
tocsoft: The logic was that there was nothing stopping you passing a PixelAcessor that was created by calling  Lock on an array and you shouldn't be able to switch out the backing buffer from the accessor and the image then as one wouldn't have originated from the array pool thus it shouldn't later be returned to it.

src/ImageSharp/Colors/PackedPixel/Byte4.cs
JimBobSquarePants: Sweet. Didn't see.

src/ImageSharp/Image/ImageBase{TColor}.cs
JimBobSquarePants: Thought as much thanks! @antonfirsov I've we are still able to lock an array then we should reintroduce the check in the manner you suggested in your comments.

antonfirsov: The only place where `PixelAccessor<T>` is constructed from an outsider array is `ArrayExtensions.Lock(this TColor[] pixels, int width, int height)`. That method is currently unused, and I don't think it's useful at all. Shouln't we remove it instead?

JimBobSquarePants: If it's unused then yes. I introduced it ages ago and the functionality has since been superseded with better objects.

JimBobSquarePants:
I'm happy with this now. 👍

antonfirsov:
Ok, then I would merge this soon.
FYI: Renamed `ArrayPointer<T>` to `BufferPointer<T>` to match `PinnedBuffer<T>`-s name. 
There would be a third member in this family called `BufferArea<T>`, it seemed to be better to follow a common naming convention.

JimBobSquarePants:
Yusssssss! Can we make our new color types now or shall we wait til I do the 2017 build?

antonfirsov:
I'm still working on making bulkpixels used. Very close to finish it. Need 1 or 2 days until the last PR comes :)

JimBobSquarePants:
Okily dokily.... Great to hear. I'll sit on my hands til then 😄

PR #126, opened 2017-03-08
Replaced PixelAccessor specialization with BulkPixelOperations

antonfirsov:
### Prerequisites

### Description

- Implemented `Color.BulkOperations : BulkPixelOperations<Color>`
- `PixelAccessor<TColor>` uses `BulkPixelOperations<TColor>.Instance` now
- Optimized `BufferPointer.Copy()` implementations
- Removed the `PixelAccessor : PixelAccessor<Color>` and `ImageFrame : ImageFrame<TColor>` specializations. The base classes are sealed now.
- ~`PixelDataPool<T>` is no longer a static class. There are two flawors instead: `PixelDataPool<T>.Clean`  and `PixelDataPool<T>.Dirty`~ **`PixelDataPool<T>` is no longer clearing it's arrays. We have `ImageBase.ClearArray<T>()`, `PinnedBuffer<T>.Clear()` and `BufferPointer<T>.Clear(int) instead**

### Benchmarks
- [Png](https://gist.github.com/antonfirsov/b7be634873179cb4bf6c5863362baf43) Note: System.Drawing.Png was running slower in the before benchmark
- [ToVector4](https://gist.github.com/antonfirsov/f0de504d784d0cf412d7bffc0ca8a1f5)

~Work is still in progress, currently I'm benchmarking and looking for issues / improvement oppurtunities.~
`ResizeProcessor` optimizations will arrive with the next PR.


antonfirsov:
@drunkappveyor

JimBobSquarePants:
I've given the build server a kick for you. Some proper voodoo in your optimized operations so expect question 😄

src/ImageSharp/Colors/Color.BulkOperations.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                int vecSize = Vector<uint>.Count;
+
+                DebugGuard.IsTrue(
+                    count % vecSize == 0,
+                    nameof(count),
+                    "Argument 'count' should divisible by Vector<uint>.Count!");
+
+                Vector<float> bVec = new Vector<float>(256.0f / 255.0f);
+                Vector<float> magicFloat = new Vector<float>(32768.0f);
```
JimBobSquarePants: We should probably put an explanation here as to what is going on. After reading this http://lolengine.net/blog/2011/3/20/understanding-fast-float-integer-conversions I still don't get the last value.

src/ImageSharp/Colors/Color.BulkOperations.cs
```diff
@@ -0,0 +1,238 @@
...
+                    for (; src < srcEnd; src++, dst++)
+                    {
+                        dst->Load(*src);
+                    }
+
+                    for (int i = 0; i < unpackedRawCount; i += vecSize)
+                    {
+                        Vector<uint> vi = new Vector<uint>(temp, i);
+
+                        vi &= mask;
```
JimBobSquarePants: I need to get better at reading this kind of stuff.

src/ImageSharp/Common/Memory/BufferPointer.cs
```diff
@@ -36,7 +30,17 @@ public static unsafe BufferPointer<T> Slice<T>(this PinnedBuffer<T> buffer)
         public static unsafe void Copy<T>(BufferPointer<T> source, BufferPointer<T> destination, int count)
             where T : struct
         {
-            Unsafe.CopyBlock((void*)source.PointerAtOffset, (void*)destination.PointerAtOffset, USizeOf<T>(count));
+            uint byteCount = USizeOf<T>(count);
+
+            if (byteCount > ByteCountThreshold)
+            {
+                if (TryMarshalCopy(source, destination, count))
```
JimBobSquarePants: I'm assuming this is benchmarked since you're always teaching me 😄

src/ImageSharp/Common/Memory/BufferPointer.cs
```diff
@@ -87,5 +110,62 @@ public static int SizeOf<T>(int count)
...
+            // Note: The following patterns are used throughout the code here and are described here
+            //
+            // PATTERN:
+            //    if (typeof(T) == typeof(Int32)) { ... }
+            //    else if (typeof(T) == typeof(Single)) { ... }
+            // EXPLANATION:
+            //    At runtime, each instantiation of BufferPointer<T> will be type-specific, and each of these typeof blocks will be eliminated,
+            //    as typeof(T) is a (JIT) compile-time constant for each instantiation. This design was chosen to eliminate any overhead from
+            //    delegates and other patterns.
+            if (typeof(TSource) == typeof(long))
```
JimBobSquarePants: I remember thinking I was going to have to do something like this before I discovered the unsafe type. Interesting to see it come back.

src/ImageSharp/Common/Memory/PixelDataPool{T}.cs
```diff
@@ -9,39 +9,69 @@ namespace ImageSharp
...
+        private PixelDataPool(ArrayPool<T> arrayPool, bool clearArray)
+        {
+            this.clearArray = clearArray;
+            this.arrayPool = arrayPool;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="PixelDataPool{T}"/> which will always return arrays initialized to default(T)
+        /// </summary>
+        public static PixelDataPool<T> Clean { get; } = new PixelDataPool<T>(CleanPool, true);
```
JimBobSquarePants: For Perf?

src/ImageSharp/Image/PixelAccessor{TColor}.cs
```diff
@@ -270,24 +283,15 @@ internal void CopyTo(PixelAccessor<TColor> target)
...
-                {
-                    packed.PackFromBytes(*(source + 2), *(source + 1), *source, 255);
-                    Unsafe.Write(destination, packed);
+                BufferPointer<byte> source = area.GetRowPointer(y);
+                BufferPointer<TColor> destination = this.GetRowPointer(targetX, targetY + y);
 
-                    source += 3;
-                    destination += size;
-                }
+                Operations.PackFromZyxBytes(source, destination, width);
```
JimBobSquarePants: I like that this is all encapsulated away now.

src/ImageSharp/Common/Memory/BufferPointer.cs
```diff
@@ -5,25 +5,19 @@
...
-        /// <typeparam name="T">The element type</typeparam>
-        /// <param name="buffer">The input <see cref="PinnedBuffer{T}"/></param>
-        /// <returns>The <see cref="BufferPointer{T}"/></returns>
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static unsafe BufferPointer<T> Slice<T>(this PinnedBuffer<T> buffer)
-            where T : struct
-        {
-            return new BufferPointer<T>(buffer.Array, (void*)buffer.Pointer);
-        }
+        private const uint ByteCountThreshold = 1024u;
```
tocsoft: Why is it better to use `Marshal.Copy()` over that size? why wouldn't you use it less that that?

How was this number decided?

src/ImageSharp/Colors/Color.BulkOperations.cs
antonfirsov: To be honest, I don't understand it either :D 
Here's what's happened: Tried the code in the arcticle -> works. Optimized it with SIMD -> still works.

src/ImageSharp/Common/Memory/BufferPointer.cs
antonfirsov: I found out since that I can simplify this code without a perf drop.

src/ImageSharp/Common/Memory/BufferPointer.cs
antonfirsov: Just run the current version of the `ArrayCopy` benchmark :)

src/ImageSharp/Common/Memory/BufferPointer.cs
antonfirsov: Empirically, using the [ArrayCopy benchmark](https://github.com/JimBobSquarePants/ImageSharp/blob/master/tests/ImageSharp.Benchmarks/General/ArrayCopy.cs).
Would be nice check it on other machines / architectures / frameworks though.

src/ImageSharp/Common/Memory/PixelDataPool{T}.cs
tocsoft: why do we need arrays defaulted to default(T) at all? Do we have logic that is dependent on the data being default? if we do how many places? could we not just reset them ourselves if required and save the added complexity of having more than one data pool? 

My guess is this is for the empty buffer for `new Image(height, width)` could we not just add a simple T[].Clear() extension method, or `RentClean(int minimumLength)` method that can be called to set the default value if required. I feel creating a new images is probably going to happen less than opening an existing file which shouldn't need a clear array anyway as it will be setting all the pixels itself.

antonfirsov: Yeah. `PixelDataPool<T>.Dirty` saves 25% in`ToVector4SimdAligned()`.
I suppose the clean method in `ArrayPool` is suboptimal, maybe we shouldn't use it at all, and do PinnedBuffer.Clean() instead.

src/ImageSharp/Common/Memory/BufferPointer.cs
tocsoft: 👍

src/ImageSharp/Common/Memory/PixelDataPool{T}.cs
antonfirsov: @JimBobSquarePants I can investigate this as part of this PR.

src/ImageSharp/Colors/Color.BulkOperations.cs
antonfirsov: If you want to understand this stuff, you need to dig yourself into IEEE specs counting bits on paper. Alternatively, we can just accept it works, like we did with the DCT implementation. 
Its like accepting a complex math thesis and not bothering with the proof.

src/ImageSharp/Common/Memory/PixelDataPool{T}.cs
antonfirsov: Okkay, it's a static class again, no longer clearing, we can call explicit clear methods instead:
- `ImageBase<T>.ClearPixels()`
- `PinnedBuffer<T>.Clear()`
- `ArrayPointer<T>.Clear(int)`

JimBobSquarePants: 👍

src/ImageSharp/Colors/Color.BulkOperations.cs
JimBobSquarePants: 😄 I'll read up then. Those API additions look interesting! Definitely needed.

antonfirsov: Yeah, all this mess could be replaced with a simple Widen() --> ConvertToSingle() chain.

antonfirsov:
@JimBobSquarepants should I merge this now?

JimBobSquarePants:
Yep. :shipit: I wanna start my VS2017 work.

PR #131, opened 2017-03-11
Upgrade solution to run in vs2017 & update dependencies

tocsoft:
### Prerequisites

### Description

- This updated the project to vs2017
- It consolidates the packages down to into only 2 projects ImageSharp & ImageSharp.Drawing

Only issue I couldn't resolve was when building via the command line stylecop.analyzer couldn't find the stylecop.json file so I had to disable the header stylecop rules, all the other stylecop rules seem to work fine.


JimBobSquarePants:
Oh... I was gonna kill that branch as it's based on the RC and the includes weren't quite right. I've already started redoing the project and have stylcop etc working perfectly so far.

tocsoft:
make sure your trying from the build script as stylecop will work fine in VS but from the command line/build script it wouldn't find the stylecop.json file and ended up spending a a day or 2 trying to solve it and failed.

You also need to make sure the codecov/opencover stuff works too, that's what I spent most of my time getting working. Getting it to work with netcoreapp's was the challenge but I managed to solve it, I found I had to do a few msbuild tricks to make it output full pdb for the codecov build, but not for the package builds(so that debugging would work on linux)

My version here got the number of targets down to 2 for ImageSharp, 1 for ImageSharp.Drawing & 1 for the ImageSharp.Testing additionally I reenginered the whole set of project files so the includes seem to be right to me.

JimBobSquarePants:
Ok... I'll have a look through everything. Cheers!

JimBobSquarePants:
Just pushed a change to the stylecop.json include which should fix your issue. 

I'm wondering, why wouldn't we have all the drawing code in the drawing project? Brushes, Pens, etc in the core don't make sense to me.

JimBobSquarePants:
We also shouldn't need the package references to buffers, vectors etc in the drawing package since we consume them via core.

tocsoft:
To test the issue you will need to reenable the headers rules in `ImageSharp.ruleset` by disabling rules SA1636 & SA1633. I just tried it and it didn't seem to make a difference for me and still failed 😞

I had tried various combination of `AdditionalFiles`/`None` etc and none of them worked.

Yeah I suppose that there's no reason at all that the brushes etc couldn't be in the Drawing package. The reason I did it that way as that code has no external dependencies, where as everything in ImageSharp.Drawing has some external dependency.

JimBobSquarePants:
OK. I'll give it a pop tonight; `AdditionalFiles` is the correct include method. Just pushed another change to drawing to remove the package references. If you have the chance to move that code please do.

tests/ImageSharp.Sandbox46/ImageSharp.Sandbox46.csproj
```diff
@@ -350,6 +350,40 @@
     <Analyzer Include="..\..\packages\Microsoft.CodeAnalysis.Analyzers.1.1.0\analyzers\dotnet\cs\Microsoft.CodeAnalysis.Analyzers.dll" />
     <Analyzer Include="..\..\packages\Microsoft.CodeAnalysis.Analyzers.1.1.0\analyzers\dotnet\cs\Microsoft.CodeAnalysis.CSharp.Analyzers.dll" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\ImageSharp.Drawing\ImageSharp.Drawing.csproj">
+      <Project>{2e33181e-6e28-4662-a801-e2e7dc206029}</Project>
+      <Name>ImageSharp.Drawing</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\src\ImageSharp.Formats.Bmp\ImageSharp.Formats.Bmp.csproj">
```
antonfirsov: These .csproj-s no longer exist. Why are you referencing them?
Also: we should remove the old .dll references in this project, and make sure everything works with the new netstandard project references.

JimBobSquarePants:
@tocsoft Looks like we're not alone re the copyright header issue. https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2310

JimBobSquarePants:
I think we're almost there. C#7 specific things can be performed after the merge.

antonfirsov:
Updated my VS2017, and my R# to 2017.1 EAP.  Everything seems to work fine what I need for a smooth workflow, including performance profiling with Sandbox46. Well done! :+1: 

IDE stability is very far from perfect though, at least for R# users.

JimBobSquarePants:
Great stuff! I'm super happy that we've pulled this off. Thanks for figuring out the tricky bits @tocsoft 

Regarding IDE stability, That'll come both Microsoft and Jetbrains are working hard on it. 

I'll merge this at lunch. 😄

PR #135, opened 2017-03-15
Encoder options being passed during path based saves

tocsoft:
### Prerequisites

### Description

The file path based save operations where failing to pass along the IEncoderOptions. This PR ensure they do and add tests to verify that all the base save operations pass along as expected.

I also Introduced a simple IFileSystem interface to easy testing the file system based codepaths.


tocsoft:
@dlemstra If I don't have the virtual `SaveInternal()` then I can't verify in my tests that all the same methods are doing the right thing in terms of defaults and passing options along etc. (well not anywhere nearly as easily)... I'll have a think to see if I can find a cleaner/different way to do this without the virtual method.

#### RE: Resetting the stream position.
I just copied that from the first same methods I migrated which was [`public Image<TColor> Save(Stream stream, IImageEncoder encoder, IEncoderOptions options)`](https://github.com/JimBobSquarePants/ImageSharp/pull/135/files#diff-ccb32d856499767c70b37d5e6ad54137L406) where originally we where resetting it back to 0 which I thought was really wrong so instead I reset back to start position. I'll remove this and leave it at the end of the stream.

antonfirsov:
@tocsoft how about an internal `IImageTestCallbacks` interface and an optional `internal Image<T>.TestCallbacks` property?

It's a common issue with TDD, that it has an effect on the design, but I generally support everything, that improves testability! We need to find clever ways to restrict these effects.

antonfirsov:
Looks good to me. @dlemstra do you approve this?

src/ImageSharp/IO/IFileSystem.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace ImageSharp.IO
+{
+    using System.IO;
+
+ #if !NETSTANDARD1_1
+    /// <summary>
+    /// A simple interface representing the filesystem.
+    /// </summary>
+    public interface IFileSystem
```
JimBobSquarePants: Love the flexibility this brings!

dlemstra:
Not sure if I like this way of testing. We are now testing if private method calls are done correctly. I rather have a test that makes sure that the options are being. We could probably do that by setting IgnoreMetadata to true/false and check the size of the resulting stream?

tocsoft:
@dlemstra I've figured a way to do away to alter my tests to do away with the callbacks class.

dlemstra:
Much better now. You can probably move the code from `SaveInternal` back to `Save`.

PR #140, opened 2017-03-17
improve vector/glyph rendering quality using a sub-pixel rendering techinique

tocsoft:
### Prerequisites

### Description
Swap out the fill region processor to use a much more accurate sub-pixel rendering technique.

This new renderer is a tiny bit slower than the old one, but (and its a big but) it totally removes the Parrallel looping making it entirely single threaded.

#### New Sub-pixel renderer
new
``` ini

Processor=Intel(R) Core(TM) i5-6200U CPU 2.30GHz, ProcessorCount=4
Frequency=2343752 Hz, Resolution=426.6663 ns, Timer=TSC
  [Host]     : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1637.0 [AttachedDebugger]
  DefaultJob : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1637.0


```
|                        Method |       Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev |     Gen 0 |    Gen 1 |    Gen 2 | Allocated |
|------------------------------ |----------- |---------- |---------- |------- |-------------- |---------- |--------- |--------- |---------- |
| 'System.Drawing Fill Polygon' |  4.0584 ms | 0.0485 ms | 0.1879 ms |   1.00 |          0.00 | 2069.5043 | 428.8793 | 428.8793 |   5.15 MB |
|     'ImageSharp Fill Polygon' | 13.4327 ms | 0.1318 ms | 0.7220 ms |   3.32 |          0.22 |  512.5000 | 262.5000 | 262.5000 |    5.3 MB |

#### Old renderer
old
``` ini

Processor=Intel(R) Core(TM) i5-6200U CPU 2.30GHz, ProcessorCount=4
Frequency=2343752 Hz, Resolution=426.6663 ns, Timer=TSC
  [Host]     : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1637.0 [AttachedDebugger]
  DefaultJob : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1637.0


```
|                        Method |      Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev |     Gen 0 |    Gen 1 |    Gen 2 | Allocated |
|------------------------------ |---------- |---------- |---------- |------- |-------------- |---------- |--------- |--------- |---------- |
| 'System.Drawing Fill Polygon' | 4.0581 ms | 0.0440 ms | 0.2108 ms |   1.00 |          0.00 | 2003.1250 | 362.5000 | 362.5000 |   5.15 MB |
|     'ImageSharp Fill Polygon' | 9.2461 ms | 0.0911 ms | 0.4988 ms |   2.28 |          0.16 |  765.6250 | 640.6250 | 640.6250 |   5.11 MB |

### DIFF
--- a/src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
+++ b/src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
@@ -116,9 +116,9 @@ namespace ImageSharp.Drawing.Brushes
             {
                 using (PinnedBuffer<float> buffer = new PinnedBuffer<float>(scanline))
                 {
-                    var slice = buffer.Slice(offset);
+                    BufferPointer<float> slice = buffer.Slice(offset);
 
-                    for (var xPos = 0; xPos < scanlineWidth; xPos++)
+                    for (int xPos = 0; xPos < scanlineWidth; xPos++)
                     {
                         int targetX = xPos + x;
                         int targetY = y;
--- a/src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
+++ b/src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
@@ -112,9 +112,12 @@ namespace ImageSharp.Drawing.Brushes
                 this.source.Dispose();
             }
 
-            internal override void Apply(float[] scanline, int scanlineWidth, int offset, int x, int y)
+            /// <inheritdoc />
+            internal override void Apply(float[] scanlineBuffer, int scanlineWidth, int offset, int x, int y)
             {
-                using (PinnedBuffer<float> buffer = new PinnedBuffer<float>(scanline))
+                DebugGuard.MustBeGreaterThanOrEqualTo(scanlineBuffer.Length, offset + scanlineWidth, nameof(scanlineWidth));
+
+                using (PinnedBuffer<float> buffer = new PinnedBuffer<float>(scanlineBuffer))
                 {
                     BufferPointer<float> slice = buffer.Slice(offset);
 
--- a/src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
+++ b/src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
@@ -34,7 +34,7 @@ namespace ImageSharp.Drawing.Brushes
         /// <inheritdoc />
         public BrushApplicator<TColor> CreateApplicator(PixelAccessor<TColor> sourcePixels, RectangleF region)
         {
-            return new ImageBrushApplicator(this.image, region);
+            return new ImageBrushApplicator(sourcePixels, this.image, region);
         }
 
         /// <summary>
@@ -71,7 +71,11 @@ namespace ImageSharp.Drawing.Brushes
             /// <param name="region">
             /// The region.
             /// </param>
-            public ImageBrushApplicator(IImageBase<TColor> image, RectangleF region)
+            /// <param name="sourcePixels">
+            /// The sourcePixels.
+            /// </param>
+            public ImageBrushApplicator(PixelAccessor<TColor> sourcePixels, IImageBase<TColor> image, RectangleF region)
+                : base(sourcePixels)
             {
                 this.source = image.Lock();
                 this.xLength = image.Width;
@@ -87,7 +91,7 @@ namespace ImageSharp.Drawing.Brushes
             /// <returns>
             /// The color
             /// </returns>
-            public override TColor this[int x, int y]
+            internal override TColor this[int x, int y]
             {
                 get
                 {
@@ -95,10 +99,10 @@ namespace ImageSharp.Drawing.Brushes
 
                     // Offset the requested pixel by the value in the rectangle (the shapes position)
                     point = point - this.offset;
-                    x = (int)point.X % this.xLength;
-                    y = (int)point.Y % this.yLength;
+                    int srcX = (int)point.X % this.xLength;
+                    int srcY = (int)point.Y % this.yLength;
 
-                    return this.source[x, y];
+                    return this.source[srcX, srcY];
                 }
             }
 
@@ -106,6 +110,34 @@ namespace ImageSharp.Drawing.Brushes
             public override void Dispose()
             {
                 this.source.Dispose();
+            }
+
+            internal override void Apply(float[] scanline, int scanlineWidth, int offset, int x, int y)
+            {
+                using (PinnedBuffer<float> buffer = new PinnedBuffer<float>(scanline))
+                {
+                    var slice = buffer.Slice(offset);
+
+                    for (var xPos = 0; xPos < scanlineWidth; xPos++)
+                    {
+                        int targetX = xPos + x;
+                        int targetY = y;
+
+                        float opacity = slice[xPos];
+                        if (opacity > Constants.Epsilon)
+                        {
+                            Vector4 backgroundVector = this.Target[targetX, targetY].ToVector4();
+
+                            Vector4 sourceVector = this[targetX, targetY].ToVector4();
+
+                            Vector4 finalColor = Vector4BlendTransforms.PremultipliedLerp(backgroundVector, sourceVector, opacity);
+
+                            TColor packed = default(TColor);
+                            packed.PackFromVector4(finalColor);
+                            this.Target[targetX, targetY] = packed;
+                        }
+                    }
+                }
             }
         }
     }
--- a/src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
+++ b/src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
@@ -115,7 +115,7 @@ namespace ImageSharp.Drawing.Brushes
             /// <inheritdoc />
             internal override void Apply(float[] scanlineBuffer, int scanlineWidth, int offset, int x, int y)
             {
-                DebugGuard.MustBeGreaterThanOrEqualTo(scanlineBuffer.Length, offset + scanlineWidth, nameof(scanlineWidth));
+                Guard.MustBeGreaterThanOrEqualTo(scanlineBuffer.Length, offset + scanlineWidth, nameof(scanlineWidth));
 
                 using (PinnedBuffer<float> buffer = new PinnedBuffer<float>(scanlineBuffer))
                 {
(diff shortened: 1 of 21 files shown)


src/ImageSharp.Drawing/Brushes/ImageBrush{TColor}.cs
```diff
@@ -107,6 +111,34 @@ public override void Dispose()
             {
                 this.source.Dispose();
             }
+
+            internal override void Apply(float[] scanline, int scanlineWidth, int offset, int x, int y)
```
JimBobSquarePants: inheritdoc

src/ImageSharp.Drawing/Brushes/PatternBrush{TColor}.cs
```diff
@@ -47,39 +38,43 @@ public class PatternBrush<TColor> : IBrush<TColor>
...
+        }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="PatternBrush{TColor}"/> class.
         /// </summary>
         /// <param name="foreColor">Color of the fore.</param>
         /// <param name="backColor">Color of the back.</param>
         /// <param name="pattern">The pattern.</param>
-        public PatternBrush(TColor foreColor, TColor backColor, bool[,] pattern)
+        internal PatternBrush(TColor foreColor, TColor backColor, Fast2DArray<bool> pattern)
```
JimBobSquarePants: Good to see `Fast2DArray<bool>` introduced here.

src/ImageSharp.Drawing/Brushes/Processors/BrushApplicator.cs
```diff
@@ -17,14 +18,64 @@ public abstract class BrushApplicator<TColor> : IDisposable // disposable will b
...
+        internal abstract TColor this[int x, int y] { get; }
 
         /// <inheritdoc/>
         public abstract void Dispose();
+
+        /// <summary>
+        /// Applies the opactiy weighting for each pixel in a scanline to the target based on the pattern contained in the brush.
+        /// </summary>
+        /// <param name="scanline">The a collection of opacity values between 0 and 1 to be merged with the burshed color value before being applied to the target.</param>
+        /// <param name="scanlineWidth">The number of pixels effected by this scanline.</param>
```
JimBobSquarePants: brushed

src/ImageSharp.Drawing/Brushes/Processors/BrushApplicator.cs
```diff
@@ -17,14 +18,64 @@ public abstract class BrushApplicator<TColor> : IDisposable // disposable will b
...
+        /// <param name="offset">The offset fromthe begining of <paramref name="scanline" /> the opacity data starts.</param>
+        /// <param name="x">The x position in the target pixel space that the start of the scanline data corresponds to.</param>
+        /// <param name="y">The y position in  the target pixel space that whole scanline corresponds to.</param>
+        internal virtual void Apply(float[] scanline, int scanlineWidth, int offset, int x, int y)
+        {
+            using (PinnedBuffer<float> buffer = new PinnedBuffer<float>(scanline))
+            {
+                BufferPointer<float> slice = buffer.Slice(offset);
+
+                for (int xPos = 0; xPos < scanlineWidth; xPos++)
```
JimBobSquarePants: Does scanlineWidth represent the width of the full scanline here or the slice length? It looks on the surface here like you could potentially overrun your slice.

src/ImageSharp.Drawing/Text/TextGraphicsOptions.cs
```diff
@@ -39,6 +44,7 @@ public TextGraphicsOptions(bool enableAntialiasing)
             this.Antialias = enableAntialiasing;
             this.ApplyKerning = true;
             this.TabWidth = 4;
+            this.AntialiasSubpixelDepth = 16;
```
JimBobSquarePants: How is this default calculated?

src/ImageSharp/Common/Memory/BufferPointer{T}.cs
```diff
@@ -74,6 +74,20 @@ public BufferPointer(T[] array, void* pointerToArray)
         public IntPtr PointerAtOffset { get; private set; }
 
         /// <summary>
+        /// Gets the element at the specified position.
+        /// </summary>
+        /// <param name="index">The index from the start of this Pointer to the required element.</param>
+        /// <returns>The <see typeparam="T"/> at the specified position.</returns>
+        public T this[int index]
```
JimBobSquarePants: This could potentially use ref returns + `Unsafe.Add<T>` @antonfirsov Am i correct?

antonfirsov: It should! Can be done with `Unsafe.AsRef<T>(ptr)`. 
This does not effect C# < 7 users here, because BufferPointer is internal.

Happy you spotted this change because, I'm turning `BufferPointer <T>` into `BufferSpan<T>` at the moment. Also wanted to add this indexer, but let's merge this first then to have less conflicts :)

BTW: `Unsafe.AsRef<T>(ptr)` is not visible for C#6 users. It means that `Span<T>` will be a ... VS2017-only data type!!! I have no idea how does Microsoft plan to manage this....

JimBobSquarePants: Great stuff! Yeah, I dunno how they'll do that. Odd

JimBobSquarePants:
Is anyone else concerned about the sheer amount of GC in the benchmarks? I haven't done any investigation at all but the numbers seem very high to me.

src/ImageSharp.Drawing/Brushes/Processors/BrushApplicator.cs
tocsoft: Its the width of the scanline data not the size of the buffer, the buffer should always be >= offset + scanlineWidth, I've added a debug guard to help demonstrate this fact.

antonfirsov:
I'll checkout the code and do a quick investigation + review now.

src/ImageSharp.Drawing/Text/TextGraphicsOptions.cs
tocsoft: Entirely subjectively it was the point I felt that that the image quality at the size feature I was testing at didn't seem to degrade any.

You'll all so notice the Text has a different default to general shapes, the reason for this is the glyphs will most likely be rendered at smaller sizes that general shapes and they will have more finer detail what needs preserving.

tocsoft:
I suspect the allocations are in the SixLabors.Shapes side of the process, this can probably be elevated by caching the shape objects and asking them to be filled directly instead of using the helper methods that take points and build the shape each time.

antonfirsov:
Managed to track down allocations with the basic VS memory profiler! :)
The most allocations are done by 'SixLabors.Shapes.InternalPath.FindIntersections()`. I suspect it's [this line](https://github.com/SixLabors/Shapes/blob/c38e6d7bdef4ccee7a3d5fdfd7f8d0fb75f86484/src/SixLabors.Shapes/InternalPath.cs#L218).

@tocsoft I think you can solve internally without using heap objects :)

Some usefult info on call times, if you want to speed up things a bit:
image
I think applying `Vector2` SIMD operations and some inlining in intersection calculator could help here ;)

I don't think we should bother with optimization right in this PR, but I'm adding my comments for information.

tocsoft:
awesome 👍 look at my code its not even needed at all anymore. Should be a super easy win 😁

src/ImageSharp.Drawing/Brushes/SolidBrush{TColor}.cs
```diff
@@ -72,13 +76,43 @@ public SolidBrushApplicator(TColor color)
...
+                {
+                    BufferPointer<float> slice = buffer.Slice(offset);
+
+                    for (int xPos = 0; xPos < scanlineWidth; xPos++)
+                    {
+                        int targetX = xPos + x;
+                        int targetY = y;
+
+                        float opacity = slice[xPos];
+                        if (opacity > Constants.Epsilon)
```
antonfirsov: These ifs are distracting CPU branch predictors. Isn't it possible solve this with pure multiplications instead? Same for `PremultipliedLerp()` implementation.

src/ImageSharp.Drawing/Brushes/SolidBrush{TColor}.cs
```diff
@@ -72,13 +76,43 @@ public SolidBrushApplicator(TColor color)
...
             /// <inheritdoc />
             public override void Dispose()
             {
                 // noop
             }
+
+            /// <inheritdoc />
+            internal override void Apply(float[] scanlineBuffer, int scanlineWidth, int offset, int x, int y)
+            {
+                DebugGuard.MustBeGreaterThanOrEqualTo(scanlineBuffer.Length, offset + scanlineWidth, nameof(scanlineWidth));
```
antonfirsov: I think it's better to use (non-debug) `Guard` here for more safety. `DebugGuard` is useful on :fire: :fire: execution paths, like indexer implementations.

antonfirsov:
Personally I would like to see this merged ASAP, so I can resolve conflicts with changes I made on the `speedup-resize` branch. I can't see any fundamental problem here, we can do optimization later.

src/ImageSharp.Drawing/Brushes/PatternBrush{TColor}.cs
```diff
@@ -156,6 +143,38 @@ public override void Dispose()
             {
                 // noop
             }
+
+            /// <inheritdoc />
+            internal override void Apply(float[] scanlineBuffer, int scanlineWidth, int offset, int x, int y)
```
antonfirsov: Ok, one change serious thing: `PatternBrushAppicator.Apply()` seems to be uncovered!

tocsoft: yeah glad you spotted that there was a bug i'd missed.. but all fixed now and covered

tocsoft:
OK, so everything's green on the build server and I've addressed all the immediate issues (more optimisations can come later) so I'm merging this to unblock @antonfirsov work.

src/ImageSharp.Drawing/Brushes/PatternBrush{TColor}.cs
antonfirsov: All hail @codecov-io!

src/ImageSharp.Drawing/Text/TextGraphicsOptions.cs
JimBobSquarePants: Sounds legit. 👍

JimBobSquarePants:
It's sooooo good finally having the tooling to profile properly again. 😄

Issue #139, opened 2017-03-17
Improve ResizeProcessor performance


antonfirsov:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
This issue has been opened to track and discuss perf improvements that could be made to ResizeProcessor:
- [x] Use `PinnedBuffer<Vector4>` for `firstPassPixels` and utilize bulk conversion
- [x] Apply microoptimizations to `ResamplingWeightedProcessor<TColor>.Weights` to make it more cache friendly

~Investigate possibilities to reduce the size of resampler windows without loss of quality.~ 
- [ ] Operate on a transposed variant of [firstPassPixels](https://github.com/JimBobSquarePants/ImageSharp/blob/c0042041fcf6e7bfed445d5c606bcebb2d3b8896/src/ImageSharp/Processing/Processors/Transforms/ResizeProcessor.cs#L109) to utilize `BulkPixelOperations<TColor>.PackFromVector4()`.
- [ ] Drop repetitive weight values in `WeightsBuffer`
- [ ] Use Value Type Kernels as Weights windows
- [ ] Investigate possibilities for faster [Compress](https://github.com/JimBobSquarePants/ImageSharp/blob/760439185e24b5fa562cce06a67a42e9cf60ccf0/src/ImageSharp/Common/Extensions/Vector4Extensions.cs#L24) and [Expand](https://github.com/JimBobSquarePants/ImageSharp/blob/760439185e24b5fa562cce06a67a42e9cf60ccf0/src/ImageSharp/Common/Extensions/Vector4Extensions.cs#L37) execution to speedup the Compand variant.

## Update
After analyzing the precomputed weights data, I found out the windows are much smaller than I thought (I misread the code first time), so we can not reduce the window size.

However:
- [This is](https://gist.github.com/antonfirsov/39760ca9cef7c66f281a4a20ce4d06a1) how the `WeightsBuffer` looks like with a Bicubic 500->200 resize operation. 90% of the data is repetitive. We can allocate + precalculate less!
- The size of `IResampler.Radius` is known at compile time for each reasmpler. It means we can achieve significant improvements by utilizing Value Type Kernels. (See #142)

[...]

PR #141, opened 2017-03-18
Buffer API changes, Optimization and refactor of ResizeProcessors

antonfirsov:
### Prerequisites

## Buffer API changes
- Refactored `BufferPointer<T>` to `BufferSpan<T>`. It's API is very similar to `System.Span` now.
- Introduced `PinnedImageBuffer<T>`: a specialization of `PinnedBuffer<T>` that holds 2D area of data. It's very similar to `PixelAccessor<T>`, but the elements are not constrained to be `IPixel<T>`-s.
- Introduced an `IPinnedImageBuffer<T>` interface for common functionalities of `PixelAcccessor<T>` and `PinnedImageBuffer<T>`. (See `PinnedImageBufferExtensions`)

## ResizeProcessor changes
- Implemented first 2 points of #139
  - Weights data is stored in a contigous memory area now (see `WeightsWindow` and `WeightsBuffer`)
  - In the final implementation the execution time dropped about 30%
- Merged `ResizeProcessor` and `CompandResizeProcessor`
- Removed several code duplications

## Benchmarks results

**CPU specs:**
Processor=Intel(R) Core(TM) i7-4810MQ CPU 2.80GHz, ProcessorCount=8
(4 physical cores)

#### Before:
Method |       Mean |    StdDev | Scaled | Scaled-StdDev | Allocated |
---------------------------- |----------- |---------- |------- |-------------- |---------- |
'System.Drawing Resize' | 52.4481 ms | 0.7776 ms |   1.00 |          0.00 |     512 B |
'ImageSharp Resize' | 35.2544 ms | 1.1755 ms |   0.67 |          0.02 |     17 MB |
'ImageSharp Compand Resize' | 65.6250 ms | 1.3946 ms |   1.25 |          0.03 |     17 MB |

#### After:
Method |       Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev | Allocated |
---------------------------- |----------- |---------- |---------- |------- |-------------- |---------- |
'System.Drawing Resize' | 51.4692 ms | 0.1478 ms | 0.5330 ms |   1.00 |          0.00 |     512 B |
'ImageSharp Resize' | 24.4749 ms | 0.3067 ms | 1.2647 ms |   0.48 |          0.02 |  17.02 MB |
'ImageSharp Compand Resize' | 43.8705 ms | 0.1408 ms | 0.5451 ms |   0.85 |          0.01 |  17.02 MB |

#### A fully sequential version of "after" on the same machine (no Parallel.For-s at all):
Method |        Mean |    StdDev |      Median | Scaled | Scaled-StdDev | Allocated |
---------------------------- |------------ |---------- |------------ |------- |-------------- |---------- |
'System.Drawing Resize' |  52.3989 ms | 1.3285 ms |  52.0098 ms |   1.00 |          0.00 |     512 B |
'ImageSharp Resize' |  58.7112 ms | 1.5875 ms |  57.6037 ms |   1.12 |          0.04 |   16.9 MB |
'ImageSharp Compand Resize' | 149.1699 ms | 0.1799 ms | 149.1820 ms |   2.85 |          0.07 |   16.9 MB |

**Conclusion:** considering that I have 4 physical cores, the `2.85/0.85 = 3.35` parallel speedup is quite good on the Compand stuff. It means, we should find an efficient way to keep the parallelism for non-server users.


tocsoft:
I can't see anything thing in here that worries me. Looking good 👍

src/ImageSharp/Processing/Processors/Transforms/ResamplingWeightedProcessor.Weights.cs
```diff
@@ -0,0 +1,175 @@
...
+            private PinnedImageBuffer<float> dataBuffer;
+
+            /// <summary>
+            /// Initializes a new instance of the <see cref="WeightsBuffer"/> class.
+            /// </summary>
+            /// <param name="sourceSize">The size of the source window</param>
+            /// <param name="destinationSize">The size of the destination window</param>
+            public WeightsBuffer(int sourceSize, int destinationSize)
+            {
+                this.dataBuffer = new PinnedImageBuffer<float>(sourceSize, destinationSize);
```
JimBobSquarePants: Any reason for not using the static method `CreateClean` to generate the clean buffer?

antonfirsov: No it was occasional. Could be more readable with CreateClean().

antonfirsov:
I think it's enough to move out if-s from the inner loop. We can benchmark a version where they are outside the outer loops, but I don't think it's worth the efforts.

src/ImageSharp/Processing/Processors/Transforms/ResamplingWeightedProcessor.Weights.cs
JimBobSquarePants: Cool. I would say so. Intent is really obviously then.

JimBobSquarePants:
I'll have a wee go just in case.

PR #143, opened 2017-03-21
Move image loading out of constructors and into static methods

tocsoft:
### Prerequisites

### Description
I've refactored Image loading (from stream, byte[] and files) to move the logic out of the `Image<TColor>` constructor and into static methods.

In the process I was able to clean up a a chunk of the Image data loading internals and additionally a shifted the responsibility of constructing the image object into the `IImageEncoder`s instead of user code. This allowed me to remove the `InitPixels(..)` method as they will always be initialised on construction.

Additionally I removed the optional parameters in from the constructors and now they are using proper overloads.

### DIFF
--- a/src/ImageSharp/Image.Create.cs
+++ b/src/ImageSharp/Image.Create.cs
@@ -1,0 +1,46 @@
+﻿// <copyright file="Image.cs" company="James Jackson-South">
+// Copyright (c) James Jackson-South and contributors.
+// Licensed under the Apache License, Version 2.0.
+// </copyright>
+
+namespace ImageSharp
+{
+    using System;
+    using System.Diagnostics;
+    using System.IO;
+
+    using Formats;
+
+    /// <summary>
+    /// Represents an image. Each pixel is a made up four 8-bit components red, green, blue, and alpha
+    /// packed into a single unsigned integer value.
+    /// </summary>
+    public sealed partial class Image
+    {
+        /// <summary>
+        /// Create a new instance of the <see cref="Image{TColor}"/> class
+        /// with the height and the width of the image.
+        /// </summary>
+        /// <typeparam name="TColor">The pixel format.</typeparam>
+        /// <param name="width">The width of the image in pixels.</param>
+        /// <param name="height">The height of the image in pixels.</param>
+        /// <param name="configuration">
+        /// The configuration providing initialization code which allows extending the library.
+        /// </param>
+        /// <returns>
+        /// A new <see cref="Image{TColor}"/> unless <typeparamref name="TColor"/> is <see cref="Color"/> in which case it returns <see cref="Image" />
+        /// </returns>
+        internal static Image<TColor> Create<TColor>(int width, int height, Configuration configuration)
+            where TColor : struct, IPixel<TColor>
+        {
+            if (typeof(TColor) == typeof(Color))
+            {
+                return new Image(width, height, configuration) as Image<TColor>;
+            }
+            else
+            {
+                return new Image<TColor>(width, height, configuration);
+            }
+        }
+    }
+}
--- a/src/ImageSharp/Image.Create.cs
+++ b/src/ImageSharp/Image.Create.cs
@@ -24,6 +24,33 @@ namespace ImageSharp
         /// <typeparam name="TColor">The pixel format.</typeparam>
         /// <param name="width">The width of the image in pixels.</param>
         /// <param name="height">The height of the image in pixels.</param>
+        /// <param name="metadata">The images matadata to preload.</param>
+        /// <param name="configuration">
+        /// The configuration providing initialization code which allows extending the library.
+        /// </param>
+        /// <returns>
+        /// A new <see cref="Image{TColor}"/> unless <typeparamref name="TColor"/> is <see cref="Color"/> in which case it returns <see cref="Image" />
+        /// </returns>
+        internal static Image<TColor> Create<TColor>(int width, int height, ImageMetaData metadata, Configuration configuration)
+            where TColor : struct, IPixel<TColor>
+        {
+            if (typeof(TColor) == typeof(Color))
+            {
+                return new Image(width, height, metadata, configuration) as Image<TColor>;
+            }
+            else
+            {
+                return new Image<TColor>(width, height, metadata, configuration);
+            }
+        }
+
+        /// <summary>
+        /// Create a new instance of the <see cref="Image{TColor}"/> class
+        /// with the height and the width of the image.
+        /// </summary>
+        /// <typeparam name="TColor">The pixel format.</typeparam>
+        /// <param name="width">The width of the image in pixels.</param>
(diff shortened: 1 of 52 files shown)


antonfirsov:
Makes sense to me :+1: 

We can even consider the following: With static Load() methods we are capable to pass an `IImageDecoder` or an `IImageFormat` explicitly:
- Users might restrict their load call to a specific decoder
- Better encapsulation for tests: the decoder selection logic won't be executed when we are testing a specific decoder.

tocsoft:
This update will actually allow you to do this for testing 
```c#
Image<Color> image =new JpegDecoder().Decode<Color>(stream, options);
```

antonfirsov:
Hmm ... you're right, but maybe it's worth to define an `Image.Load(stream, decoder, options)` wrapper for the sake of API consistency.

tocsoft:
Overloads added
 You can now call 

```c#
byte[] bytes = .......;
Image img = Image.Load(bytes, new PngDecoder());
```

JimBobSquarePants:
@tocsoft I'll do a review of this asap. I'm reading up on chromatic adaption just now so I can write unit tests.

tocsoft:
I just realised that when returning an `Image` as aposed to `Image<TColor>` at the moment I'm duplicating the pixel data... small change to come which should resolve them.

tocsoft:
@JimBobSquarePants fixed the problem I spotted

tocsoft:
Is everyone now happy with this?

src/ImageSharp/Image.Create.cs
```diff
@@ -0,0 +1,46 @@
...
+    using System.Diagnostics;
+    using System.IO;
+
+    using Formats;
+
+    /// <summary>
+    /// Represents an image. Each pixel is a made up four 8-bit components red, green, blue, and alpha
+    /// packed into a single unsigned integer value.
+    /// </summary>
+    public sealed partial class Image
```
JimBobSquarePants: This feels odd. We're extending the non-generic Image class with Image<TColor> methods. I get that it works, it just feels odd.

src/ImageSharp/Image.Decode.cs
```diff
@@ -0,0 +1,71 @@
...
+    using System.Text;
+    using Formats;
+
+    /// <summary>
+    /// Represents an image. Each pixel is a made up four 8-bit components red, green, blue, and alpha
+    /// packed into a single unsigned integer value.
+    /// </summary>
+    public sealed partial class Image
+    {
+        private static IImageFormat DiscoverFormat(Stream stream, Configuration config)
```
JimBobSquarePants: XML docs please

src/ImageSharp/Image.FromFile.cs
```diff
@@ -0,0 +1,218 @@
+﻿// <copyright file="Image.FromStream.cs" company="James Jackson-South">
+// Copyright (c) James Jackson-South and contributors.
+// Licensed under the Apache License, Version 2.0.
+// </copyright>
+
+namespace ImageSharp
+{
+    using System;
```
JimBobSquarePants: Using's should be inside `#if`

src/ImageSharp/Image.FromStream.cs
```diff
@@ -0,0 +1,248 @@
+﻿// <copyright file="Image.FromStream.cs" company="James Jackson-South">
+// Copyright (c) James Jackson-South and contributors.
+// Licensed under the Apache License, Version 2.0.
+// </copyright>
+
+namespace ImageSharp
+{
+    using System;
```
JimBobSquarePants: Unused references in here. Buffer, Diagnostics, Linq

src/ImageSharp/Image.FromStream.cs
```diff
@@ -0,0 +1,248 @@
...
+        /// <exception cref="NotSupportedException">
+        /// Thrown if the stream is not readable nor seekable.
+        /// </exception>
+        /// <returns>The image</returns>
+        public static Image<TColor> Load<TColor>(Stream stream, IDecoderOptions options, Configuration config)
+            where TColor : struct, IPixel<TColor>
+        {
+            config = config ?? Configuration.Default;
+
+            Image<TColor> img = WithSeekableStream(stream, s =>
```
JimBobSquarePants: Can be. 

```
Image<TColor> img = WithSeekableStream(stream, s => Decode<TColor>(stream, options, config));
```

src/ImageSharp/Image.FromBytes.cs
```diff
@@ -0,0 +1,216 @@
+﻿// <copyright file="Image.FromStream.cs" company="James Jackson-South">
+// Copyright (c) James Jackson-South and contributors.
+// Licensed under the Apache License, Version 2.0.
+// </copyright>
+
+namespace ImageSharp
+{
+    using System;
```
JimBobSquarePants: Unused references in here. Buffer, Diagnostics, Linq, Text

src/ImageSharp/Image.cs
```diff
@@ -219,7 +49,7 @@ public Image(byte[] bytes, IDecoderOptions options, Configuration configuration)
         /// </summary>
         /// <param name="other">The other image, where the clone should be made from.</param>
         /// <exception cref="System.ArgumentNullException"><paramref name="other"/> is null.</exception>
-        public Image(Image other)
+        internal Image(Image<Color> other)
```
JimBobSquarePants: How do I now deep clone an image?

tocsoft: not sure why I did that. 😕

src/ImageSharp/Image.FromBytes.cs
tocsoft: theses have been removed

src/ImageSharp/Image.FromStream.cs
tocsoft: done

src/ImageSharp/Image.FromFile.cs
tocsoft: done

src/ImageSharp/Image.Create.cs
tocsoft: It felt the best from the public api point of view (felt it ended up being the most discoverable via intellisense).

JimBobSquarePants: Aye, Enumerable does the same.

JimBobSquarePants:
I'm happy to have this merged now.

dlemstra:
I rather have the copying of the metadata in the copy constructor. Don't really like the LoadFrom method because you can call it multiple times. But it makes no sense to me to have this option. Also leaving a minor comment about the argument order. My OCD says configuration should be first.

tocsoft:
Had a play with the argument ordering and totally agree moved the config args to the start.

tocsoft:
@dlemstra I've taken your feedback into account you happier with this now?

PR #144, opened 2017-03-22
WIP ICC Support

JimBobSquarePants:
### Prerequisites

### Description
This is a work in progress PR that will add full ICC profile support plus a comprehensive color conversion system.

The work will comprise of two parts.

1. Reading and writing the ICC profiles.
2. Enhancing our colorspaces to allow converting colors using those profiles.  

We will try to follow the established EXIF process as much as possible during this process, potentially reusing objects

**ICC profile support completed so far:**
- [x] ICC Reader and Writer supporting the [ICC profile specification v4.3.0.0](http://www.color.org/icc_specs2.xalter) 
- [x] Wire up Reader/Writer in Jpeg and other formats that support the specification.

**Color Spaces supported so far:**
- [x] CIELab
- [x] CIELch
- [x] CIELuv
- [x] CIELchuv
- [x] CIEXyy
- [x] CIEXyz
- [x] Cmyk
- [x] Hsl
- [x] Hsv (Also Hsb)
- [x] HunterLab
- [x] LinearRgb
- [x] Lms
- [x] Rgb
- [x] YCbCr (Jpeg)


src/ImageSharp/MetaData/Profiles/ICC/Curves/IccFormulaCurveElement.cs
```diff
@@ -0,0 +1,84 @@
...
+            this.B = b;
+            this.C = c;
+            this.D = d;
+            this.E = e;
+        }
+
+        /// <summary>
+        /// Gets the type of this curve
+        /// </summary>
+        public ushort Type { get; }
```
dlemstra: Should this not be an enum?

src/ImageSharp/MetaData/Profiles/ICC/Curves/IccOneDimensionalCurve.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// </summary>
+        /// <param name="breakPoints">The break points of this curve</param>
+        /// <param name="segments">The segments of this curve</param>
+        public IccOneDimensionalCurve(float[] breakPoints, IccCurveSegment[] segments)
+        {
+            Guard.NotNull(breakPoints, nameof(breakPoints));
+            Guard.NotNull(segments, nameof(segments));
+
+            bool isWrongSize = breakPoints.Length != segments.Length - 1;
+            Guard.IsTrue(isWrongSize, $"{nameof(breakPoints)},{nameof(segments)}", "Number of BreakPoints must be one less than number of Segments");
```
dlemstra: This guard feels wrong? You are guarding that isWrongSize is true?

src/ImageSharp/MetaData/Profiles/ICC/Curves/IccSampledCurveElement.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="IccSampledCurveElement"/> class.
+        /// </summary>
+        /// <param name="curveEntries">The curve values of this segment</param>
+        public IccSampledCurveElement(float[] curveEntries)
+            : base(IccCurveSegmentSignature.SampledCurve)
+        {
+            Guard.NotNull(curveEntries, nameof(curveEntries));
+            Guard.IsTrue(curveEntries.Length < 1, nameof(curveEntries), "There must be at least one value");
```
dlemstra: Should this not be `curveEntries.Length > 0`?

src/ImageSharp/MetaData/Profiles/ICC/IccDataReader.cs
```diff
@@ -0,0 +1,1727 @@
...
+
+        /// <summary>
+        /// Reads an ASCII encoded string
+        /// </summary>
+        /// <param name="length">number of bytes to read</param>
+        /// <returns>The value as a string</returns>
+        public string ReadASCIIString(int length)
+        {
+            // Encoding.ASCII is missing in netstandard1.1, use UTF8 instead because it's compatible with ASCII
+            string value = Encoding.UTF8.GetString(this.data, this.AddIndex(length), length);
```
dlemstra: You could to `Encoding.GetEncoding("ASCII")` instead: https://github.com/JimBobSquarePants/ImageSharp/blob/a82d28f563279c433d9474981774c1745f8d25ca/src/ImageSharp/Formats/Png/PngDecoderOptions.cs#L15

src/ImageSharp/MetaData/Profiles/ICC/IccDataReader.cs
```diff
@@ -0,0 +1,1727 @@
...
+
+            // remove data after (potential) null terminator
```
dlemstra: Why do you want to do this?

src/ImageSharp/MetaData/Profiles/ICC/IccProfile.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        /// <param name="data">The raw ICC profile data</param>
+        public IccProfile(byte[] data)
+        {
+            Guard.NotNull(data, nameof(data));
+
+            this.data = data;
+            IccReader reader = new IccReader();
+            this.header = reader.ReadHeader(data);
+            this.entries = new List<IccTagDataEntry>(reader.ReadTagData(data));
```
dlemstra: Maybe you could do this lazy and only read them when the property is requested?

src/ImageSharp/MetaData/Profiles/ICC/IccProfile.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        /// <summary>
+        /// Calculates the MD5 hash value of an ICC profile header
+        /// </summary>
+        /// <param name="data">The data of which to calculate the hash value</param>
+        /// <returns>The calculated hash</returns>
+        public static IccProfileId CalculateHash(byte[] data)
+        {
+            Guard.NotNull(data, nameof(data));
+            Guard.IsTrue(data.Length < 128, nameof(data), "Data length must be at least 128 to be a valid profile header");
```
dlemstra: Also the wrong way around? `data.Length >= 128` Maybe use `MustBeGreaterThanOrEqualTo`? You probably want to check all the Guard.IsTrue checks? There are more incorrect ones.

src/ImageSharp/MetaData/Profiles/ICC/TagDataEntries/IccDataTagDataEntry.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// If <see cref="IsAscii"/> is false, returns null
+        /// </summary>
+        public string AsciiString
+        {
+            get
+            {
+                if (this.IsAscii)
+                {
+                    // Encoding.ASCII is missing in netstandard1.1, use UTF8 instead because it's compatible with ASCII
+                    return Encoding.UTF8.GetString(this.Data, 0, this.Data.Length);
```
dlemstra: See my other comment

src/ImageSharp/MetaData/Profiles/ICC/IccDataReader.cs
JBildstein: do you mean removing data after \0? If so, I've had some issues in the past when displaying strings without doing that. I think I had some profiles that had some nonsense characters after the first null terminator.
It's also unnecessary to keep it in, .Net doesn't need it and in the ICC profile it's just used to mark the end or for padding.

src/ImageSharp/MetaData/Profiles/ICC/Curves/IccFormulaCurveElement.cs
JBildstein: It should be yes, I remember wanting to do this but didn't know how to name it. Implemented it now with member names Type1-3. Maybe I should come back to this and give them some better names.
(IccParametricCurve had the same problem, did that too now)

src/ImageSharp/MetaData/Profiles/ICC/IccDataReader.cs
JBildstein: cool thanks! was looking for an alternative but only found the UTF8 one. Fix it now everywhere I used it.

src/ImageSharp/MetaData/Profiles/ICC/IccProfile.cs
JBildstein: Yeah, I got the IsTrue/IsFalse the wrong way around...
Thanks for noticing! I fixed it now everywhere I used it.

src/ImageSharp/MetaData/Profiles/ICC/IccProfile.cs
JBildstein: did that now like it's done with Exif

JimBobSquarePants:
@JBildstein I think merging in your BitConverter changes might have broken the ICC reader tests. Could you possibly have a look once you have a moment.

JBildstein:
@JimBobSquarePants very sorry about that! all tests pass again with #157 merged

JimBobSquarePants:
No worries :smile: Thanks for fixing it so quickly!

JimBobSquarePants:
@JBildstein I think we're in a place now where we can actually start to do some conversion within jpeg. Any idea where to start?

JimBobSquarePants:
@JBildstein Ok, so jpeg now loads and stores the profile now. We need to now wire up the conversion code. I think I have everything we need there but could really do with your input when you have a moment.

JimBobSquarePants:
@JBildstein I'm able to read and write the ICC profile data from most of our images now. 

There is a CMYK image in the test solution though that throws an error when trying to parse/write the entries in it's 29 ICC profile markers. I've written the wrapping reading/writing code correctly, something is going wroing in either ICCDataReader or ICCDataWriter

JimBobSquarePants:
Stack trace on that error. Happens with out YCCK and CMYK image. Something to do with the Entries collection.
```
System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values.
Parameter name: startIndex
   at ImageSharp.IO.EndianBitConverter.CheckByteArgument(Byte[] value, Int32 startIndex, Int32 bytesRequired) in C:\development\github\ImageSharp\src\ImageSharp\IO\EndianBitConverter.cs:line 127
   at ImageSharp.IO.BigEndianBitConverter.ToInt16(Byte[] value, Int32 startIndex) in C:\development\github\ImageSharp\src\ImageSharp\IO\BigEndianBitConverter.cs:line 63
   at ImageSharp.IO.EndianBitConverter.ToUInt16(Byte[] value, Int32 startIndex) in C:\development\github\ImageSharp\src\ImageSharp\IO\EndianBitConverter.ToType.cs:line 47
   at ImageSharp.IccDataReader.ReadUInt16() in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\DataReader\IccDataReader.Primitives.cs:line 22
   at ImageSharp.IccDataReader.ReadClut16(Int32 inChannelCount, Int32 outChannelCount, Byte[] gridPointCount) in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\DataReader\IccDataReader.Lut.cs:line 136
   at ImageSharp.IccDataReader.ReadLut16TagDataEntry() in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\DataReader\IccDataReader.TagDataEntry.cs:line 280
   at ImageSharp.IccDataReader.ReadTagDataEntry(IccTagTableEntry info) in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\DataReader\IccDataReader.TagDataEntry.cs:line 42
   at ImageSharp.IccReader.ReadTagData(IccDataReader reader) in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\IccReader.cs:line 90
   at ImageSharp.IccReader.ReadTagData(Byte[] data) in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\IccReader.cs:line 55
   at ImageSharp.IccProfile.InitializeEntries() in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\IccProfile.cs:line 168
   at ImageSharp.IccProfile.get_Entries() in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\IccProfile.cs:line 89
   at ImageSharp.IccWriter.Write(IccProfile profile) in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\IccWriter.cs:line 27
   at ImageSharp.IccProfile.ToByteArray() in C:\development\github\ImageSharp\src\ImageSharp\MetaData\Profiles\ICC\IccProfile.cs:line 134
   at ImageSharp.Formats.JpegEncoderCore.WriteICCProfiles(IList`1 iccProfiles) in C:\development\github\ImageSharp\src\ImageSharp\Formats\Jpeg\JpegEncoderCore.cs:line 722
   at ImageSharp.Formats.JpegEncoderCore.WriteProfiles[TPixel](Image`1 image) in C:\development\github\ImageSharp\src\ImageSharp\Formats\Jpeg\JpegEncoderCore.cs:line 777
   at ImageSharp.Formats.JpegEncoderCore.Encode[TPixel](Image`1 image, Stream stream) in C:\development\github\ImageSharp\src\ImageSharp\Formats\Jpeg\JpegEncoderCore.cs:line 222
   at ImageSharp.Formats.JpegEncoder.Encode[TPixel](Image`1 image, Stream stream, IJpegEncoderOptions options) in C:\development\github\ImageSharp\src\ImageSharp\Formats\Jpeg\JpegEncoder.cs:line 37
   at ImageSharp.Formats.JpegEncoder.Encode[TPixel](Image`1 image, Stream stream, IEncoderOptions options) in C:\development\github\ImageSharp\src\ImageSharp\Formats\Jpeg\JpegEncoder.cs:line 23
   at ImageSharp.Image`1.Save(Stream stream, IImageEncoder encoder, IEncoderOptions options) in C:\development\github\ImageSharp\src\ImageSharp\Image\Image{TPixel}.cs:line 238
   at ImageSharp.Image`1.Save(Stream stream, IEncoderOptions options) in C:\development\github\ImageSharp\src\ImageSharp\Image\Image{TPixel}.cs:line 181
   at ImageSharp.Image`1.Save(Stream stream) in C:\development\github\ImageSharp\src\ImageSharp\Image\Image{TPixel}.cs:line 169
   at ImageSharp.Tests.GeneralFormatTests.DecodeThenEncodeImageFromStreamShouldSucceed() in C:\development\github\ImageSharp\tests\ImageSharp.Tests\Formats\GeneralFormatTests.cs:line 59
```

JimBobSquarePants:
I really wanna merge this now. Any objections?

src/ImageSharp/MetaData/Profiles/ICC/IccProfile.cs
```diff
@@ -0,0 +1,193 @@
...
+                return this.header;
+            }
+
+            set => this.header = value;
+        }
+
+        /// <summary>
+        /// Gets the actual profile data
+        /// </summary>
+        public List<IccTagDataEntry> Entries
```
antonfirsov: I'm totally unfamiliar with the ICC stuff, and I'm not sure if it's worth to overthink it's API, but considering the fact that the API is public I think it's worth to ask a few questions:

- Are we sure the lazy initialization way is the best way here? For a big header it could produce an overhead on an unexpected point from the users POV (reading a property).
- Is it really a good way to expose this stuff through `List<T>`? Do intend it to be writable?

src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.Adapt.cs
```diff
@@ -0,0 +1,198 @@
...
+{
+    using System;
+
+    using ImageSharp.ColorSpaces;
+    using ImageSharp.ColorSpaces.Conversion.Implementation.Rgb;
+
+    /// <content>
+    /// Performs chromatic adaptation on the various color spaces.
+    /// </content>
+    public partial class ColorSpaceConverter
```
antonfirsov: I see many optimization possibilities regarding the `ColorSpaceConverter` API. Are we sure we need it to be public for the beta/1.0?

JimBobSquarePants: I'm assuming you mean bulk operations etc? I can internalize this for now. It's not something that is explicitly required for the moment.

src/ImageSharp/MetaData/Profiles/ICC/IccProfile.cs
JimBobSquarePants: The API is actually very similar to the EXIF profile API. If we don't lazy initialize the properties we could potentially really hurt decoding performance. It should never be slow enough to negatively affect the user once we are past that point though.

I've always assumed we want this to be writable so this looks correct to me.

src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.Adapt.cs
antonfirsov: Yeah, a general bulk conversion could be really useful in some scenarios (like Jpeg), but I see very high chance that it's also possible to optimize the simple conversion. There are chains of calls on heap object instances. There is an `IColorConversion<in T, out TResult>` interface implemented by stateless objects, and used 99% oft the time through the concrete instances. I'm sure it's possible to find a better way to do this. 

Some concreate ideas:
- Making `IColorConversion.Convert()` implementations invokable through static methods. 
- Review of `IChromaticAdaptation`. Currently it introduces lot's of non-inlinable virtual method calls.
- Evalute possibilities to turn some of the classes into `struct`-s, and play with generic constraints to make more of the invocations inlineable

JimBobSquarePants: All cool. 👍  I've internalized all of it for now. 

The only reason `IColorConversion.Convert()` is not static is because some converters require the specific whitepoint to be passed. I guess we could refactor that to pass the whitepoint. 

`IChromaticAdaptation` exists as there are multiple possible adaption types, we only use `VonKriesChromaticAdaptation` for now though. However, there's no need for us to be passing `IColorConversion` etc through the constructor, they're concrete with no alternative methods.

Bear in mind it's already 96% faster than the original code from Colourful! 😄

antonfirsov: Really cool to have all these new features, and the implementation is good and clean for the first iteration! :smile: 
Also good to see it being internal for now, so we don't have to freeze it with the beta! ;)

JimBobSquarePants:
Merging in now that I've internalized the colorspace code and disabled related tests.

PR #156, opened 2017-03-28
Update to use MathsF

JimBobSquarePants:
### Prerequisites

### Description
This update updates all existing code that used floating point `System.Math` methods to use our own internal `MathF` class. This reduces the noise in the codebase, makes intent more obvious and improves testability of the libraries.

I've aggressively inlined each method so there should be no slowdown.

I can add extra tests for the new class but I think the increased coverage from it's use should probably be enough.


tocsoft:
looking good to me, only thing this PR makes me notice is we now have a lot of instances of `Math.Abs(var) > Constants.Epsillon` I realise that those are outside the scope of the PR but we might want to normalise them down to use a static `ApproximateFloatComparer`.

JimBobSquarePants:
Great minds think alike. 😄 

Yeah, Out of scope for this PR but let's definite refactor that soon.

PR #167, opened 2017-04-06
Fix for off by one for indexed pngs

### LINKED ISSUE #166: Off by one Issue decoding indexed png
### Prerequisites

- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Another png issue, when decoding an index png the pixels seem to be off by one when its resaved.

## Source image
before

## Index encoded image
cansaveindexedpng_standardimageclass_testpattern100x100

## After decoding
This is generated by opening the indexed png then encoded it back back as a bmp (converted to png using MS paint as github doesn't support bmp files)
after

### Steps to Reproduce

```c#
        [Theory]
        [WithTestPatternImages(100, 100, PixelTypes.StandardImageClass)]
        public void CanSaveIndexedPng<TColor>(TestImageProvider<TColor> provider)
            where TColor : struct, IPixel<TColor>
        {
            // does saving a file then repoening mean both files are identical???
            using (Image<TColor> image = provider.GetImage())
            using (MemoryStream ms = new MemoryStream())
            {
                image.MetaData.Quality = 256;
                image.Save(ms, new PngEncoder());
                ms.Position = 0;
                image.DebugSave(provider, extension: "bmp");

                using (Image<TColor> img2 = Image.Load<TColor>(ms, new PngDecoder()))
                {
                    img2.Save(provider, new { decoded = true }, extension: "bmp");

                    using (PixelAccessor<TColor> accessor1 = image.Lock())
                    using (PixelAccessor<TColor> accessor2 = img2.Lock())
                    {
                        for (int y = 0; y < image.Height; y++)
                        {
                            for (int x = 0; x < image.Width; x++)
                            {
                                Assert.Equal(accessor1[x, y], accessor2[x, y]);
                            }
                        }
                    }
                }
            }
        }
```
### System Configuration

- ImageSharp version: (master branch) f91d6b5
- .NET Framework version:
- Additional information:

tocsoft:
### Prerequisites

### Description
Fixes #166 meaning that index pngs offset the first pixel of each row and fill that value with what ever color is in index 0 of the palette.

I think this is correct based on the fact we always add an extra byte to the scanline and when we encode the bytes with skip the first index.

See https://github.com/JimBobSquarePants/ImageSharp/blob/7d348667d6341e6a770e607cae94ff68df0176bc/src/ImageSharp/Formats/Png/Filters/NoneFilter.cs#L40
and https://github.com/JimBobSquarePants/ImageSharp/blob/7d348667d6341e6a770e607cae94ff68df0176bc/src/ImageSharp/Formats/Png/PngDecoderCore.cs#L350

_**DISCLAIMER** I don't know enough about the png format to know which side is correct should the first item be skipped or is it a bug in the encoder putting writing the scanline at the wrong location._

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -579,7 +579,7 @@ namespace ImageSharp.Formats
                         // channel and we should try to read it.
                         for (int x = 0; x < this.header.Width; x++)
                         {
-                            int index = newScanline[x];
+                            int index = newScanline[x + 1];
                             int pixelOffset = index * 3;
 
                             byte a = this.paletteAlpha.Length > index ? this.paletteAlpha[index] : (byte)255;
@@ -603,7 +603,7 @@ namespace ImageSharp.Formats
                     {
                         for (int x = 0; x < this.header.Width; x++)
                         {
-                            int index = newScanline[x];
+                            int index = newScanline[x + 1];
                             int pixelOffset = index * 3;
 
                             byte r = this.palette[pixelOffset];


JimBobSquarePants:
I'm gonna merge this now. We need to have a good look through png for performance boosters at some point though, I think it's a prime candidate for our buffer slicing tools.

Issue #132, opened 2017-03-14
error resizing image


niksan321:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I have an error while resizing this image:
![nakladki_na_porogi_ladya_mazda_6_32](https://cloud.githubusercontent.com/assets/2101530/23914799/1d96ba34-08f8-11e7-83af-006dbca07a69.jpg)

Unable to cast object of type 'System.Byte[]' to type 'System.Byte'

### Steps to Reproduce

in params:
name = nakladki_na_porogi_ladya_mazda_6_32.jpg
pathIn = c:\in\
pathOut = c:\out\
maxWidth = 250
maxHeight = 190

        public void GenerateOptimizedJpg(string name, string pathIn, string pathOut, int maxWidth, int maxHeight, int quality = 60)
        {
            var imgIn = $"{pathIn}{name}";
            var imgOut = $"{pathOut}{Path.GetFileNameWithoutExtension(name)}.jpg";
            var photoBytes = File.ReadAllBytes(imgIn);

            var resizeOptions = new ResizeOptions
            {
                Mode = ResizeMode.Max,
                Size = new Size(maxWidth, maxHeight),
                Position = AnchorPosition.Center,
            };

            using (var inStream = new MemoryStream(photoBytes))
            using (var output = File.OpenWrite(imgOut))
            {
                var options = new JpegEncoderOptions { Quality = quality };
                var image = new Image(inStream);
                image.BackgroundColor(Color.White)
                    .Resize(resizeOptions)
                    .SaveAsJpeg(output, options);
            }
        }

### System Configuration

- ImageSharp version: 
    "ImageSharp.Drawing": "1.0.0-alpha2-00158",
    "ImageSharp.Formats.Bmp": "1.0.0-alpha2-00158",
    "ImageSharp.Formats.Gif": "1.0.0-alpha2-00158",
    "ImageSharp.Formats.Jpeg": "1.0.0-alpha2-00158",
    "ImageSharp.Formats.Png": "1.0.0-alpha2-00154",
[...]

PR #168, opened 2017-04-06
Issue #132: Handle undefined EXIF data type where no number of compon…

blackcity:
### Prerequisites

### Description
#132. Some applications do not set `NumberOfComponents `in EXIF data even if the type is `Undefined`. 

EXIF data of type `Undefined `with number of components not set (numberOfComponents == 0) is handled differently between ExifReader and ExifWriter. ExifReader handles it as byte array (4 bytes) while ExifWriter handles it as single byte. This results in an InvalidTypeCast exception.
<pre>
       private int WriteValue(ExifDataType dataType, object value, byte[] destination, int offset)
        {
             switch (dataType)
            {
                case ExifDataType.Byte:
                case ExifDataType.Undefined:
                    destination[offset] = (byte)value;
                    return offset + 1;
                [...]
            }
        }
</pre>

This PR sets the number of components to 4 (Exif inline value length) if type is Undefined and numberOfComponents is 0.


JimBobSquarePants:
Is this a fix for #132 ? Awesome and thank you! I'll look at it asap.

blackcity:
Yes, fixes #132 and related issues with those weird EXIF values.

JimBobSquarePants:
Great stuff! Thanks for taking the time to submit a fix.

dlemstra:
This project is using the same code as my Magick.NET project and I tried to apply your fix there. But your unit test passes without any problems without the patch applied. You can see my test in the referenced commit.

blackcity:
I thought the same the first time I saw this error. But, there is no error during **reading** the EXIF data. The exception occurs later when the image is saved. Then ExifWriter tries to cast the byte array to a single byte:
<pre>
private int WriteValue(ExifDataType dataType, object value, byte[] destination, int offset)
{
     switch (dataType)
    {
        case ExifDataType.Byte:
        case ExifDataType.Undefined:
            destination[offset] = (byte)value;   // value is a byte Array -> InvalidCastException
            return offset + 1;
        [...]
    }
}
</pre>

But source of the problem is in ExifReader, because the Exif type `Undefined` should not have `NumberOfComponents` set to 0. And this is what the test does, it checks that `NumberOfComponents != 0`:
<pre>
if (entry.DataType == ExifDataType.Undefined)
{
    Assert.NotEqual(0, entry.NumberOfComponents);
}
</pre>

Btw, `NumberOfComponents` can have any value greater than 0 with ExifType `Undefinded`. But if it's 0 it must be set to 4 as a fallback, we handle it now as number of bytes. So in my opinion your test is not correct because NumberOfComponents can have any value except 0. Your test:
<pre>
if (value.DataType == ExifDataType.Undefined) 
{ 
    Assert.AreEqual(4, value.NumberOfComponents);
}
</pre> 

If you want to have the exception, just save the image in the unpatched version and ExifWriter will throw an `InvalidCastException`.

dlemstra:
Maybe I was not being clear enough. When I only add the unit test to my project it passes but I would expect it to fail. I ran the test through the debugger and there are two values that are ` ExifDataType.Undefined`. One has the value `1` for `NumberOfComponents` and the other one has `4` as the value. This means it will pass the unit test if you check if the value is not `0`. This is before applying your fix in `ExifReader.cs`.

blackcity:
Ok, misread your note. Thanks.

blackcity:
Unit test updated. Since `NumberOfComponents` has a default value of 1 the previous version always passed the test.

JimBobSquarePants:
@dlemstra As soon as you're happy with this could you do the merge please. You're the boss of this part.

PR #171, opened 2017-04-11
Fix alpha channel generation for indexed PNGs

### LINKED ISSUE #170: Very bad result after resize.
Hi, I use ImageSharp 1.0.0-alpha5-00054 with ASP.NET Core 1.1

I use the following code to resize the image:

	var result = new MemoryStream();

	var options =
		new ResizeOptions
		{
			Size = new Size(230, 140),
			Mode = resizeMode
		};

	Image.Load(input).Resize(options).Save(result);

And the result is very bad, see the attached source image (login-icon) and resized image:

login-icon
8bf49d97-ba7a-4eec-b056-b9b9700298e3_0_230_140_boxpad

I have no idea what I do wrong.

tocsoft:
### Prerequisites

### Description

Fixes how we generate the alpha channel for Indexed PNGs.

Based on info from [the W3C png spec](https://www.w3.org/TR/PNG/#11transinfo) the alpha channel should have an entry per index entry (if any alpha required).

This leaves me with the understanding that a `Threshold` amount doesn't make much sense but left it in for now but changed its default so it retains the entire set of values.

### Observations

Applying this fix makes me notice 2 things;
1. The OctreeQuantizer seems to introduce some odd difference in the alpha values of the indexed 
pixels.
2. The OctreeQuantizer will produce compound errors between encodings even if the the number of different colors in the image is already below the threshold. i.e. encode indexed, decode, and with no changes encode as indexed again does not produce exactly the same pixel data. 

### Related Issues
Fixes #170

### DIFF
--- a/src/ImageSharp/Formats/Png/IPngEncoderOptions.cs
+++ b/src/ImageSharp/Formats/Png/IPngEncoderOptions.cs
@@ -41,11 +41,6 @@ namespace ImageSharp.Formats
         IQuantizer Quantizer { get; }
 
         /// <summary>
-        /// Gets the transparency threshold.
-        /// </summary>
-        byte Threshold { get; }
-
-        /// <summary>
         /// Gets a value indicating whether this instance should write
         /// gamma information to the stream.
         /// </summary>
--- a/src/ImageSharp/Formats/Png/IPngEncoderOptions.cs
+++ b/src/ImageSharp/Formats/Png/IPngEncoderOptions.cs
@@ -41,6 +41,11 @@ namespace ImageSharp.Formats
         IQuantizer Quantizer { get; }
 
         /// <summary>
+        /// Gets the transparency threshold.
+        /// </summary>
+        byte Threshold { get; }
+
+        /// <summary>
         /// Gets a value indicating whether this instance should write
         /// gamma information to the stream.
         /// </summary>
--- a/src/ImageSharp/Formats/Png/PngEncoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngEncoderCore.cs
@@ -495,7 +495,7 @@ namespace ImageSharp.Formats
 
             // Grab the palette and write it to the stream.
             TColor[] palette = quantized.Palette;
-            byte pixelCount = (byte)palette.Length;
+            byte pixelCount = palette.Length.ToByte();
 
             // Get max colors for bit depth.
             int colorTableLength = (int)Math.Pow(2, header.BitDepth) * 3;
@@ -517,6 +517,11 @@ namespace ImageSharp.Formats
                         colorTable[offset] = bytes[0];
                         colorTable[offset + 1] = bytes[1];
                         colorTable[offset + 2] = bytes[2];
+
+                        if (alpha > this.options.Threshold)
+                        {
+                            alpha = 255;
+                        }
 
                         anyAlpha = anyAlpha || alpha < 255;
                         alphaTable[i] = alpha;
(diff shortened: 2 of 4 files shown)


JimBobSquarePants:
On your observations @tocsoft 

> Observations
> 
> Applying this fix makes me notice 2 things;
> 
> 1. The OctreeQuantizer seems to introduce some odd difference in the alpha values of the indexed
> pixels.
> 2. The OctreeQuantizer will produce compound errors between encodings even if the the number of different colors in the image is already below the threshold. i.e. encode indexed, decode, and with no changes encode as indexed again does not produce exactly the same pixel data.

The Octree isn't a true Octree, rather a hacky attempt to allow full range alpha. It almost works but as you say some of the alpha differences are odd. This is probably due to the way I am applying the mask to provide a pixel index. (I need it to be less than 8).

https://github.com/JimBobSquarePants/ImageSharp/blob/13a4a2613feef13c473a3f2688febb539c3c4902/src/ImageSharp/Quantizers/Octree/OctreeQuantizer.cs#L457

```
private static readonly int[] Mask = { 0x100, 0x80, 0x40, 0x20, 0x10, 0x08, 0x04, 0x02, 0x01 };

int index = ((buffer[3] & Mask[0]) >> (shift - 3)) |
            ((buffer[2] & Mask[level + 1]) >> (shift - 2)) |
            ((buffer[1] & Mask[level + 1]) >> (shift - 1)) |
            ((buffer[0] & Mask[level + 1]) >> shift);
```

The original code would have been.

```
private static readonly int[] Mask = { 0x80, 0x40, 0x20, 0x10, 0x08, 0x04, 0x02, 0x01 };

int index = ((buffer[2] & Mask[level + 1]) >> (shift - 2)) |
            ((buffer[1] & Mask[level + 1]) >> (shift - 1)) |
            ((buffer[0] & Mask[level + 1]) >> shift);
```
I had a quick go just then at altering the allowable max depth to be 15 and using the correct `((buffer[3] & Mask[level + 1]) >> (shift - 3))` (essentially producing what seems to be called a [hextree](https://github.com/mapnik/mapnik/blob/master/include/mapnik/hextree.hpp)) which almost worked but there's something amiss with white pixels causing them to become transparent. 

I'm not sure whether that is of any use though to be honest as memory usage will shoot up and speed down.

I had a play around using the Wu Quantizer with your changes and the output was excellent even without dithering. We could match the quality of the image from issue #170 yet produce smaller output than tinypng. I might have another go at producing dithering in that quantizer.

If that succeeds I'll probably return the Octree to it's original state which serves Gif well and would boost performance there. We'll then make Wu the default for png and carefully comment both explaining the difference in transparency ranges.

Either way, decoding and encoding an indexed png will not produce identical output unless we followed the quantization method exactly used on the original image.

What do you think?

tocsoft:
sounds good to me.

Can we also add a first step that will check if the picture we are encoding even needs its pixel colors reducing? for example if I've opened and indexed png, cropped out a portion and then resaved it the image won't suddenly have more colors than before.

I should be able to save that png without requiring a full Quantization pass just a partial one to create the palette lookup. For simple pre-processed images should save us having to perform a full  Quantization pass unless we added enough colors to each the palette limit.

JimBobSquarePants:
We'd have to store the palette and somehow keep track of the operations performed on the image to determine whether there were changes to the color palette. I'm not sure that would be possible. 

Personally I'd expect slight differences with indexed png since they are lossy by nature.

BTW - Here's the original and output from the Wu Quantizer. The difference is minimal really considering we saved .65kb

2486674a-1d70-11e7-8a42-15109d3dd408
splash - copy

tocsoft:
no I just ment we should take the output of the first pass and see if the number of distinct colors is under the palette size threshold and if it is don't bother doing the lossy reduce stage and just return the first pass data as a QuantizizedImage.

tocsoft:
also side note, the opacity of the 2nd one seems to be wrong, the blue of the octopus has an alpha of 254 instead of 255.

JimBobSquarePants:
Ah right.. so count unique colors then make no attempt to reduce the palette if it's smaller. We'd probably have to set a size limit on that calculation if we were to do it so not to slow things down too much...

On the difference in opacity... ¯\_(ツ)_/¯

PR #173, opened 2017-04-13
Feature/better quantization

JimBobSquarePants:
### Prerequisites

### Description
This PR updates the quantizers in the following manner.

1. Adds dithering support to the **WuQuantizer\<TColor\>**
2. Refactors **WuQuantizer\<TColor\>** to inherit **Quantizer\<TColor\>**
3. Strips out incorrect **OctreeQuantizer\<TColor\>**  multiple color alpha support.
4. Changes the default quantizer in Png to **WuQuantizer\<TColor\>**

See issue #170 and PR #171 for related conversation.


JimBobSquarePants:
We get absolutely fantastic quality from the WuQuantizer now. Here's us vs pngquant

**Original 73.8kb**
ducati_side_shadow

**ImageSharp 21.8kb**
ducati_side_shadow

**pngquant 22.8kb**
ducati_side_shadow-fs8

src/ImageSharp/Quantizers/Wu/WuQuantizer.cs
```diff
@@ -118,13 +103,13 @@ public sealed class WuQuantizer<TColor> : IQuantizer<TColor>
...
         public WuQuantizer()
         {
-            this.vwt = LongPool.Rent(TableLength);
-            this.vmr = LongPool.Rent(TableLength);
-            this.vmg = LongPool.Rent(TableLength);
-            this.vmb = LongPool.Rent(TableLength);
-            this.vma = LongPool.Rent(TableLength);
-            this.m2 = FloatPool.Rent(TableLength);
-            this.tag = BytePool.Rent(TableLength);
+            this.vwt = WuArrayPool.LongPool.Rent(TableLength);
```
antonfirsov: I think we should aim to replace these individual pools with `PinnedBuffer`-s when touching different implementations along the library to centralize the buffer logic, and make future refactor work easier. 
(It won't be pinned I'm almost done removing all the pointers.)

JimBobSquarePants: Agreed. I'd really like to see a more centralized approach. Glad you're making progress!

oscarld:
@JimBobSquarePants Thanks for the great work.

I was trying to reduce the size in KB (not in pixels of course) of png images with ImageSharp 1.0.0-beta0002, but the result was not successfully.

Next, My tries (The input image was [this](https://imgcdn.larepublica.co/cms/test/santorini.png) and has 1363433 Bytes of size):

1. 1112342 Bytes [this](https://imgcdn.larepublica.co/cms/test/santorinifirsttry.png)
```cs
...
using (var image = Image.Load(inputStream))
{
   image.Mutate(x => x
      .Quantize(Quantization.Wu)
   );
   image.SaveAsPng(outputStream);
}
...
```

2. 1560018 Bytes [this](https://imgcdn.larepublica.co/cms/test/santorinisndtry.png)

```cs
using (var image = Image.Load(inputStream))
{
   image.SaveAsPng(cachedStream, new PngEncoder
   {
      Quantizer = new WuQuantizer<Rgba32>()
   });
}
```

Please, can you share an example?

JimBobSquarePants:
Hi @oscarld 

What you need to do is set the `PngColorType` to `PngColorType.Palette`

``` c#
using (var image = Image.Load(inputStream))
{
   image.SaveAsPng(cachedStream, new PngEncoder
   {
      PngColorType = PngColorType.Palette
   });
}
```

An explanation:

In your first code example you have a slightly smaller than the original because you are reducing the number of colors in the image. There is only a slight decrease as you are still saving the image using a 32bit RGBA pixel format.

In your second example Your image is larger because the original must have been previous optimized for web and the quantizer is not actually being used in the encoding process because, again, you are still saving the image using a 32bit RGBA pixel format. (Our encoder outputs smaller images than `System.Drawing` but cannot match a dedicated optimizer due to time constraints)

Here's the output of the image where the `PngEncoder` has been passed the `PngColorType` option. As you can see it is much smaller at 457,910 bytes

Santorini as indexed png

However, I would try to consider not using png as a format for saving photographs when possible unless you specifically require alpha transparency since you will never achieve good compression. Rather you should a codec designed specifically for photographs like jpeg.

Here's your image saved as a jpeg at 85% Quality.
Santorini as jpeg

That wipes the floor with png producing a high quality image at only 189,010 bytes.

Hope that makes sense

If you have any more questions, don't hesitate to visit our Gitter channel so we can help there.

Cheers

James

PR #174, opened 2017-04-14
Drop all pointers

antonfirsov:
### Prerequisites

### Description
This PR replaces pointer usages with `ref` locals without significant effects on performance.  `Unsafe.Add(ref T, int)` is utilized to do "pointer arithmetic" on managed `ref`-s.

- `PinnedBuffer<T>` => `Buffer<T>`: the main responsibility of this class is managing the (optional) pooling of the backing array from now. Pinning is still possible with explicit call (currently it's used in benchmarks only.)
- `PinnedImageBuffer<T>` => `Buffer2D<T>`
- `BufferSpan<T>` is pointer-free now, it's API got more closer to `System.Span<T>`
- Removed pointers from `PixelAccessor<T>` and `PixelArea<T>`. Indexers are using basic array indexing.

Everything is ready to drop `PixelAccessor<T>` now!

### Benchmarks
[Bulk operations BEFORE](https://gist.github.com/antonfirsov/4ebcd74daff24ddf8366a81376823edc)
[Bulk operations AFTER](https://gist.github.com/antonfirsov/42535e71812f62a55970bf8b06f2d554)


src/ImageSharp/Common/Memory/BufferSpan.cs
```diff
@@ -16,59 +15,48 @@ namespace ImageSharp
...
+            ref byte destRef = ref Unsafe.As<T, byte>(ref destination.DangerousGetPinnableReference());
+
+            int byteCount = Unsafe.SizeOf<T>() * count;
+
+            // TODO: Use unfixed Unsafe.CopyBlock(ref T, ref T, int) for small blocks, when it gets available!
+            fixed (byte* pSrc = &srcRef)
+            fixed (byte* pDest = &destRef)
+            {
+#if NETSTANDARD1_1
+                Unsafe.CopyBlock(pDest, pSrc, (uint)byteCount);
```
antonfirsov: @JimBobSquarePants this line is uncovered.
Currently we are not able to test code on `#if NETSTANDARD1_1` path. We should change our test project configuration, or drop the 1.1 target.

JimBobSquarePants: I say drop it. 1.3 covers everything we had before except NET 4.5 Which I don't want to support anyway.

PR #179, opened 2017-04-17
Add ColorVector, refactor Color

JimBobSquarePants:
### Prerequisites

### Description
This change does three things.

1. Removes manual packing and `IPackedVector` implementation from `Color`
2. Introduces a packed `Rgba32` implmenting `IPackedVector`
3. Introduces a new floating point `Vector4` backed `ColorVector` struct + initial optimized bulk operations.

Currently the `ColorVector` struct is still slower than our default `Color` struct (possibly due to cache misses) but I've a suspicion the current copying when reading/writing to the array is making a big difference also. Adding ref returns and Get/Set Pixel will possibly help.


src/ImageSharp/Colors/ColorVector.Definitions.cs
```diff
@@ -0,0 +1,728 @@
+﻿// <copyright file="ColorVector.Definitions.cs" company="James Jackson-South">
```
tocsoft: You can just drop this entire file. You can access the same data from `NamedColors<TColor>.ColorName`

JimBobSquarePants: I wanted `Color` and `ColorVector` to have the same available static properties without having to directly call that file.

tocsoft: Ah ok,  that's fine then.

JimBobSquarePants: Sweet. I'll be duplicating the `Color.Transforms` class also eventually. Just wanna get this in first.

antonfirsov:
It's merged now, so I'm a bit late, but I have a few remarks here:
- There is no real difference between the new `Color` struct and the old one which is `Rgba32` now. The memory layout is the same, the API-s seem to be the same, only the implementations are different. I wonder if we need both of them.
- I think we should find a better name for `ColorVector`. The name should reflect the semantic details of that struct, like `ArgbVector`. (To be honest I also don't like the name `Color` for the same reason, but I understand if it's here for API simplicity.)

JimBobSquarePants:
@antonfirsov That's all cool. 😎 

- `Rgba32` and `Color` are actually different in that `Rgba32` implements `IPackedVector` and has the `PackedValue` property. I created it for interoperability with MonoGame should anyone ever want to use it.
- ColorVector fels clumsy name-wise to me also but I wanted something that differentiated it from the "Packed" types. Open to suggestions! 😄

antonfirsov:
Its trivial to make Color implementing `IPackedVector<uint>`, and they will be exactly the same!
Its ok to maintain monogame compatibility, but monogame actually doesn't do anything through that interface. The compatibility is here by having identical binary buffer formats, no one will use an interface with small value types (boxing!), they will just memcpy the buffers. So I dont think "packed types" is a real category, and its not worth to bring noise to our colorspace model by having two identical structures with an artifical difference between their implemented interfaces. 

I think `VectorRgba` or `RgbaVector` is a good name for `ColorVector`. Its rgba and its a vector :)

JimBobSquarePants:
>Its trivial to make Color implementing IPackedVector<uint>, and they will be exactly the same!

Is it? I must be missing something then. If I'm explicitly declaring the struct property order `PackedValue` would have to be a field which I can't use as that can't implement an interface, only a property or method can.

antonfirsov:
There are two options:
1. Having an "additional" uint field with [FieldOffset(0)]  that overlaps the stuct. It can be returned by the property getter then. 
2. `return Unsafe.As<Color, uint>(ref this)` should also do the job.

PR #180, opened 2017-04-18
Cleanup color

JimBobSquarePants:
### Prerequisites

### Description
Fixing a few things that weren't quite right about #179 

- `PackFromVector4` was returning a new instance
- Inlining instructions wer missing from `Color`
- `Color` can now haz `PackedValue`
- `Rgba32` was removed.

I've kept `ColorVector` naming as-is for now. I want the relationship to Color to be obvious.

I may yet still rename `Color` to `Rgba32` and `ColorVector` to `RgbaVector`, depend how much people can bribe me to do so.


antonfirsov:
@JimBobSquarePants much better! 👍 
Could be even better if I had the money for that bribe! 😄

>  I want the relationship to Color to be obvious.

The only relationship is the order of the components. Even if you keep the name `Color` for API simplicity, for me it feels cleaner to provide descriptive names for the rest, including the Vector4 pixel type.

JimBobSquarePants:
`ColorVector` actually supports the same transforms and static members as `Color` also differentiating it from the rest. I'll keep mulling it over but merge this for now.

PR #187, opened 2017-04-24
Performance improvements for PNG Decoding

Drawaes:
Stop the copy from deframing to inflate
Removed the pixel by pixel processing for RgbWithAlpha
Cleaned up the Crc check on the inflated stream

Down from ~4x slower than the native System.Drawing method to ~2.5x slower.

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -24,7 +24,14 @@ namespace ImageSharp.Formats
         /// <summary>
         /// The dictionary of available color types.
         /// </summary>
-        private static readonly Dictionary<int, byte[]> ColorTypes = new Dictionary<int, byte[]>();
+        private static readonly Dictionary<PngColorType, byte[]> ColorTypes = new Dictionary<PngColorType, byte[]>()
+        {
+            [PngColorType.Grayscale] = new byte[] { 1, 2, 4, 8 },
+            [PngColorType.Rgb] = new byte[] { 8 },
+            [PngColorType.Palette] = new byte[] { 1, 2, 4, 8 },
+            [PngColorType.GrayscaleWithAlpha] = new byte[] { 8 },
+            [PngColorType.RgbWithAlpha] = new byte[] { 8 },
+        };
 
         /// <summary>
         /// The amount to increment when processing each column per scanline for each interlaced pass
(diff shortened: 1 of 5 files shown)


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -294,6 +316,39 @@ private void ReadPhysicalChunk(ImageMetaData metadata, byte[] data)
...
+            pixels = image.Lock();
+            this.bytesPerPixel = this.CalculateBytesPerPixel();
+            this.bytesPerScanline = this.CalculateScanlineLength(this.header.Width) + 1;
+            this.bytesPerSample = 1;
+            if (this.header.BitDepth >= 8)
+            {
+                this.bytesPerSample = this.header.BitDepth / 8;
+            }
+
+            this.previousScanline = ArrayPool<byte>.Shared.Rent(this.bytesPerScanline);
```
antonfirsov: We have `Buffer<byte> : IDisposable` class now to encapsulate renting/returning and possible future memory operations.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -637,20 +672,30 @@ private void ProcessDefilteredScanline<TPixel>(byte[] defilteredScanline, int ro
...
+        /// <typeparam name="TPixel">The type of pixel</typeparam>
+        /// <param name="defilteredScanline">The completed scanline</param>
+        /// <param name="pixels">The pixel accessor</param>
+        private unsafe void RgbWithAlpha<TPixel>(byte[] defilteredScanline, PixelAccessor<TPixel> pixels)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            int offset = this.bytesPerSample * 4;
+            int width = this.header.Width * offset;
+            int pixelId = this.currentRow * this.header.Width;
+            Rgba32[] pixelArray = pixels.PixelArray as Rgba32[];
```
antonfirsov: This works only with `Image<Rgba32>`! (Our fault the non-rgba case is not covered by tests though.)

A general and simple way to do this conversion:
`BulkPixelOperations<TPixel>.Instance.PackFromXyzwBytes(sourceColorSpan, destByteSpan, count)`
In case of `Rgba32` it will delegate to an [optimized memory copy call](https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/Common/Memory/BufferSpan.cs#L25).

Drawaes: fixed

src/ImageSharp/Formats/Png/PngDecoderCore.cs
antonfirsov: Nevermind this point, if you don't have time to figure it out :) I'm planning a library-wide refactor to replace all individual ArrayPool usages.

JimBobSquarePants:
Fantastic work. Really appreciated! 💯

antonfirsov:
Just checked some stuff manually:
Decoding ImageSharp-encoded Png-s seem to work correctly with all `PngColorType`-s after this change.

Drawaes:
Cool, good news

Issue #155, opened 2017-03-26
JPEG decoder will generate wrong pixel colors sometime


sgjsakura:
### Description
The JPEG Decoder will generate pixel colors with slight changes compared with the original color.

### Steps to Reproduce
Using the following simple code to reproduce the problem, here the test image "test1.jpg" can be downloaded via http://galnote.com/StaticFiles/attachments/1703/djqanebq.jpg
```C#
var inputFile = @"image\test1.jpg";

using (var image = new Image(inputFile))
using (var accessor = image.Lock())
{
	var c = accessor[1000, 555];
}
```
The above code will get the color of the pixel at (1000, 555) location, and the value of `c` will be RGB(249,243,222). However the actual color in this point should be RGB(250, 243, 224). This problem will further image processings such as saving it using `SaveAsBmp` method. However Photoshop and Micorosft Paint app can correctly decode the pixels.

### System Configuration

- ImageSharp version: 1.0.0-alpha4-00048
- Other ImageSharp packages and versions: None
- Environment (Operating system, version and so on): Windows 10 Build 1607 with Jan 2017 Update
- .NET Framework version: .NET Core App 1.1 (the version of `Micorosft.NETCore.App` is 1.1.1)
- Additional information: None

tocsoft:
I guess will be this is due to the fact that this image has an ICC profile associated with it.

Currently we don't have ICC profile support, but we are currently actively working on getting it in.

blackcity:
@tocsoft We had a similar issue in our testing. I don't think it is an ICC issue, because ICC is non-destructive, so the underlying pixel is the same with or without ICC profile, only the monitor display color  changes. I didn't test the image, but i bet if we open it in Photoshop and remove the ICC profile we will always get the same color values for this pixel.

JimBobSquarePants:
I'd like to do a proper test against the supplied image comparing multiple decoders to see what the result is. There may be more variance than we think.

antonfirsov:
Thanks for the detailed feedback!
[...]

PR #188, opened 2017-04-26
Use same conversion methods as libjpeg

### LINKED ISSUE #82: Difference between resize results between System.Drawing & Imagesharp
Hi I'm using this file as the input:

orgtest

You probably allready are aware of that but with System.Drawing it takes on avg: 300ms and with ImageSharp arround 1300ms. Anyway to speed it up?

I also noticed a considerable difference in the resized image.
Done with System.Drawing with these settings:
```cs
                        graphics.InterpolationMode = InterpolationMode.HighQualityBicubic;
                        graphics.CompositingQuality = CompositingQuality.HighQuality;
                        graphics.SmoothingMode = SmoothingMode.HighQuality;
                        graphics.PixelOffsetMode = PixelOffsetMode.HighQuality;
```
This is the result from System.Drawing:
currently

With ImageSharp
```cs
var im = new Image(stream);
im.Resize(1300, 866, new BicubicResampler()).Save(targetStream);
```
It looks like the ImageSharp Resized image contains a bit more green. And the filesize is 64kb bigger then the System.Drawing example. I get comparable results with other samplers. 

Am I doing something wrong. Is this as designed?

PS: ImageMagick (Magick.Net or Node.JS) and OpenCV return a result very similar to System.Drawing.

JimBobSquarePants:
### Prerequisites

### Description
This PR improves the accuracy of our YCbCr->Rgb->YcbCr conversions by using the same approach to building lookup tables that libjpeg turbo does. There's very little speed difference to our prior methods ~0.5ms on encode. 

See the following for reference.
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/077e5bb4e0f6761c034e7c1f7e7a4120f580f1a0/jdcolor.c#L211

https://github.com/libjpeg-turbo/libjpeg-turbo/blob/077e5bb4e0f6761c034e7c1f7e7a4120f580f1a0/jccolor.c#L198

With this merged I believe we can close #82 and #155

Here's a comparision between the System.Drawing and Previous ImageSharp outputs and our new one.

**System.Drawing**
sys-draw

**ImageSharp Current 1.0.0-alpha7-00012**
isharp-current

**This PR**
isharp-pr

You'll notice the green tint is completely gone and the image much more clear.

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/Decoder/YCbCrToRgbTables.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Decoder/YCbCrToRgbTables.cs
@@ -1,0 +1,95 @@
+﻿// <copyright file="YCbCrToRgbTables.cs" company="James Jackson-South">
+// Copyright (c) James Jackson-South and contributors.
+// Licensed under the Apache License, Version 2.0.
+// </copyright>
+
+namespace ImageSharp.Formats.Jpg
+{
+    using System.Runtime.CompilerServices;
+
+    using ImageSharp.PixelFormats;
+
+    /// <summary>
+    /// Provides 8-bit lookup tables for converting from YCbCr to Rgb colorspace.
+    /// Methods to build the tables are based on libjpeg implementation.
+    /// </summary>
+    internal struct YCbCrToRgbTables
+    {
+        // Speediest right-shift on some machines and gives us enough accuracy at 4 decimal places.
+        private const int ScaleBits = 16;
+
+        private const int Half = 1 << (ScaleBits - 1);
+
+        private static readonly int[] CrRTable = new int[256];
+
+        private static readonly int[] CbBTable = new int[256];
+
+        private static readonly int[] CrGTable = new int[256];
+
+        private static readonly int[] CbGTable = new int[256];
+
+        /// <summary>
+        /// Optimized method to pack bytes to the image from the YCbCr color space.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="packed">The packed pixel.</param>
+        /// <param name="y">The y luminance component.</param>
+        /// <param name="cb">The cb chroma component.</param>
+        /// <param name="cr">The cr chroma component.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Pack<TPixel>(ref TPixel packed, byte y, byte cb, byte cr)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            // float r = MathF.Round(y + (1.402F * cr), MidpointRounding.AwayFromZero);
+            byte r = (byte)(y + CrRTable[cr]).Clamp(0, 255);
+
+            // float g = MathF.Round(y - (0.344136F * cb) - (0.714136F * cr), MidpointRounding.AwayFromZero);
+            // The values for the G calculation are left scaled up, since we must add them together before rounding.
+            byte g = (byte)(y + RightShift(CbGTable[cb] + CrGTable[cr])).Clamp(0, 255);
+
+            // float b = MathF.Round(y + (1.772F * cb), MidpointRounding.AwayFromZero);
+            byte b = (byte)(y + CbBTable[cb]).Clamp(0, 255);
+
+            packed.PackFromBytes(r, g, b, byte.MaxValue);
+        }
+
+        /// <summary>
+        /// Initializes the YCbCr tables
+        /// </summary>
+        /// <returns>The intialized <see cref="YCbCrToRgbTables"/></returns>
+        public YCbCrToRgbTables Init()
+        {
+            for (int i = 0, x = -128; i <= 255; i++, x++)
+            {
+                // i is the actual input pixel value, in the range 0..255
+                // The Cb or Cr value we are thinking of is x = i - 128
+                // Cr=>R value is nearest int to 1.402 * x
+                CrRTable[i] = RightShift((Fix(1.402F) * x) + Half);
+
+                // Cb=>B value is nearest int to 1.772 * x
+                CbBTable[i] = RightShift((Fix(1.772F) * x) + Half);
+
+                // Cr=>G value is scaled-up -0.714136286
+                CrGTable[i] = (-Fix(0.714136286F)) * x;
+
+                // Cb => G value is scaled - up - 0.344136286 * x
+                // We also add in Half so that need not do it in inner loop
+                CbGTable[i] = ((-Fix(0.344136286F)) * x) + Half;
+            }
+
+            return this;
+        }
+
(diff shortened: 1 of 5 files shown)


src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrTables.cs
```diff
@@ -0,0 +1,106 @@
...
+            // float b = MathF.Round(y + (1.772F * cb), MidpointRounding.AwayFromZero);
+            crBlockRaw[index] = (YCbCrTable[r + RCrOffset] + YCbCrTable[g + GCrOffset] + YCbCrTable[b + BCrOffset]) >> ScaleBits;
+        }
+
+        /// <summary>
+        /// Initializes the YCbCr tables
+        /// </summary>
+        /// <returns>The intialized <see cref="RgbToYCbCrTables"/></returns>
+        public RgbToYCbCrTables Init()
+        {
```
tocsoft: why not just do this in a constructor? or even better make the whole thing a static class.

src/ImageSharp/Formats/Jpeg/Components/Decoder/YCbCrToRgbTables.cs
```diff
@@ -0,0 +1,95 @@
...
+    using System.Runtime.CompilerServices;
+
+    using ImageSharp.PixelFormats;
+
+    /// <summary>
+    /// Provides 8-bit lookup tables for converting from YCbCr to Rgb colorspace.
+    /// Methods to build the tables are based on libjpeg implementation.
+    /// </summary>
+    internal struct YCbCrToRgbTables
+    {
```
tocsoft: can this not just be a static class? with a static constructor for the stuff in `Init()`?

tocsoft:
Awesome, the colors are looking way better.

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrTables.cs
JimBobSquarePants: I wanted to keep this these as lightweight struct as I was accessing it so much. Since I can't define a parameterless constructor I used a method. I need to fix this actually since I don't think it's threadsafe.

JimBobSquarePants:
Q: If I wanted to fix my arrays, how would I then access them quickly in my Pack method without using fixed wrapper which is slow?

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrTables.cs
antonfirsov: @JimBobSquarePants but you forgot to turn`YCbCrTable` into a fixed instance array. Currently you are initializing a static array in an instance method, which doesn't make too much sense.
Actually, this table is much bigger than Unzig, su I'm not sure if it's worth to move it's data to the stack. It might be OK in it's current form, but in this case you should turn it into a static class as @tocsoft suggests. The hybrid form is useles and confusing.

src/ImageSharp/Formats/Jpeg/Components/Decoder/YCbCrToRgbTables.cs
antonfirsov: @JimBobSquarePants see my comment on `RgbToYCbCrTables`, the same could be applied to `YCbCrToRgbTables`!

An additional thing: it's really important to move these 4 tables into a contigous memory area.  It will reduce the chance of cache misses. You have 2 options for this:
1. Do it like you did in `RgbToYCbCrTables`, and turn `YCbCrToRgbTables` into a static class
2. Turn all the four `static readonly int[]` arrays into `fixed` instance arrays.

I guess in case of `YCbCrToRgbTables` the second approach would be faster (no additions needed).

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrTables.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <param name="crBlockRaw">The blue chroma block.</param>
+        /// <param name="index">The current index.</param>
+        /// <param name="r">The red value.</param>
+        /// <param name="g">The green value.</param>
+        /// <param name="b">The blue value.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public unsafe void Allocate(ref float* yBlockRaw, ref float* cbBlockRaw, ref float* crBlockRaw, int index, int r, int g, int b)
+        {
+            // float y = (0.299F * r) + (0.587F * g) + (0.114F * b);
+            yBlockRaw[index] = (YCbCrTable[r] + YCbCrTable[g + GYOffset] + YCbCrTable[b + BYOffset]) >> ScaleBits;
```
antonfirsov: It's possible to eliminate these `+...Offset` operations.
- If you choose to move `RgbToYCbCrTables` into a fixed array, split it into 8 different arrays!
- You can also pin `YCbCrTable` at the beginning of the `ToYCbCr()` call, and move 8 pointers to a struct that could be passed to `Allocate()`

antonfirsov:
@JimBobSquarePants the answer to your question is [here](https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs#L95)
You need to turn your instance methods into static methods accepting a pointer :(

antonfirsov:
But unlike in `Block8x8F`, you don't need casting in your table structs.

With a `YCbCrToRgbTables* table` you can use `table->CrRTable[i]`

JimBobSquarePants:
I am 99% there with making the changes, I can't figure out how to get a reference to my fixed `YCbCrToRgbTables* table` to pass to the method.

Working in managed code all my career really has left me with knowledge gaps!

JimBobSquarePants:
@antonfirsov @tocsoft  Think I've got it now.

antonfirsov:
@JimBobSquarePants it's never too late late to learn new stuff! :)

Haven't tested it online, but looks good to me now.

JimBobSquarePants:
@antonfirsov Absolutely! I'm off to bed now so will see what everyone thinks in the morning. 

Managed to shave a few ms off the decoder so happy there. Sub 30ms in the DecodeJpeg benchmark for the first time on my machine. 

We should compile a jpeg decoder todo list to see if we can attract passing optimizers. If they know what the issues are they will be in a better position to help out.

antonfirsov:
@JimBobSquarePants  It should be the same thing that happens with other tests involving SIMD. (Like my case with `ColorVectorTransformTests` yesterday.)

I have a pretty long, and detailed TODO list regarding JpegDecoder. I also want to see this written down as an issue, so we can argue about the points/prioritize them/inform potential new contributors.

JimBobSquarePants:
628b747f8ccdfb757062f36a27eedecfc2295f515c0586e05fbfb0620c0571a2

Issue #191, opened 2017-04-26
Image.Load with interlaced PNG files: System.ArgumentOutOfRangeException : Non-negative number required. Parameter name: count


srudenko:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Exception happens when files loaded from stream. It only happens with some interlaced PNG files, particularly with the following 2 files
![invalidpngwithinterlacing](https://cloud.githubusercontent.com/assets/576927/25446457/895306b0-2aa9-11e7-8a5a-c03330d0eca8.png)
![91c3ae_fdc3323aeacf4744b1ff1418c8f100d2](https://cloud.githubusercontent.com/assets/576927/25447459/30920216-2aad-11e7-9236-9ed2b5208499.png)

I have tested the same code with interlaced PNG from your repository [plash-interlaced.png](https://github.com/JimBobSquarePants/ImageSharp/blob/master/tests/ImageSharp.Tests/TestImages/Formats/Png/splash-interlaced.png) and it was successfully loaded.

### Steps to Reproduce

>                    using (var fileStream = File.OpenWrite(filePath))
>                     {
>                               input.Seek(0, SeekOrigin.Begin);> 
>                               var image = Image.Load(input);
>                     }

Exception is:

`System.ArgumentOutOfRangeException : Non-negative number required.
Parameter name: count
   at System.IO.Win32FileStream.Read(Byte[] array, Int32 offset, Int32 count)
   at System.IO.FileStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.Compression.DeflateStream.Read(Byte[] array, Int32 offset, Int32 count)
   at ImageSharp.Formats.ZlibInflateStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at ImageSharp.Formats.PngDecoderCore.DecodeInterlacedPixelData[TPixel](Stream compressedStream, PixelAccessor1 pixels)
   at ImageSharp.Formats.PngDecoderCore.Decode[TPixel](Stream stream)
[...]

PR #196, opened 2017-04-27
Fixed Interlaced decoding for multi frames

Drawaes:
Removed the ZlibInflateStream as extra layer of un-needed indirection
Increased the overall perf by removing the extra Read layer.
Added support back in for multi frame interlaced PNG

Test image from issue
#191


src/ImageSharp/Formats/Png/Zlib/DeframeStream.cs
```diff
@@ -3,6 +3,7 @@
     using System;
```
JimBobSquarePants: Can we now rename this to ZLibInflateStream for consistancy?

Drawaes: Changed the name

JimBobSquarePants:
All good now. Merging, Thanks! 👍

PR #200, opened 2017-04-28
[WIP] Duff-Porter general pixel function for drawing images

vpenades:
### Prerequisites

### Description

I've began using ImageSharp a while ago, primarily for a project that requires blending multiple layers one on top of another, preserving correct alpha channels... pretty much like how potoshop laters work.

I had the effects already written for another library, so I gave it a try to port them to ImageSharp.

I am aware there's already Rgba32.Transforms, but they're not quite the same, because these functions assume, to some degree, the background image is opaque, so, if used with transparent backgrounds, the results are unexpected.

My proposal works this way:

In the class `PorterDuffFunctions<TBckPixel, TsrcPixel>` I wrote some of the main Porter Duff composition functions. They are generic, so they can run with any IPixel type, even the Backdrop and Source pixel formats can be different!

But, instead of calling these functions directly, there's this enumeration:

`enum PixelTransformMode { Normal, Multiply, Add, etc, etc };`

And an extension that allows to do this:

`var pixelFunction = PixelTransformMode.Multiply.GetPixelFunc<Rgba32,Rgba32>();`

From now on, I just need to do:

`var finalColor = pixelFunction(Rgba32.Red,Rgba32.Blue,0.5f);`

Finally, in order to allow drawing images with this feature, I added a processor `DrawImageEffectProcessor `which is a copy, with modifications, of `DrawImageProcessor ` and some more extensions.

And that's it!

Feel free to comment, suggest, rename, or rearrange the code I wrote!

Btw, I've just began using Github, so sorry in advance if I messed something!!! ^_^;


tocsoft:
I've not manage d to review this yet but I have been planning a similar piece of work for a little while so you've just saved me a whole heap of time 👍.

I'll review this over the next few days and get it/a version of it merged in soon. I'll be more looking at it from a vector drawing angle but I'll get the APIs working consistently across all the drawing methods.

JimBobSquarePants:
Thanks @vpenades but before we go any further I have a few questions. 

1. What is wrong with the original equations and if there is something why not fix them? 
2. Where are the corresponding unit tests? My original equations match the results from the W3C specification. 
3. What is the performance of your methods, have you benchmarked them?. I have a feeling invoking the function per pixel will be slow. 

Cheers

James

src/ImageSharp/PixelFormats/PorterDuffFunctions.cs
```diff
@@ -0,0 +1,267 @@
...
+        /// General composition function for all modes, with a general solution for alpha channel
+        /// </summary>
+        /// <param name="backdrop">Original backgrop color</param>
+        /// <param name="source">Original source color</param>
+        /// <param name="xform">Desired transformed color, without taking Alpha channel in account</param>
+        /// <returns>The final color</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static TBckPixel Compose(Vector4 backdrop, Vector4 source, Vector4 xform)
+        {
+            System.Diagnostics.Debug.Assert(source.W > 0, nameof(source) + " Alpha must be non zero");
```
JimBobSquarePants: Use `DebugGuard` here.

vpenades:
@JimBobSquarePants The way I use these blend functions is pretty much the same used by photoshop, that is: I have a collection of layers, all of them with alpha mask, opacity, and blend mode, and I want to "flatten" all of them in one single image, preserving the combined alpha values of all the layers in the final image.

When I needed to do that I also tried the standard W3C formulas, and soon discovered that, of all the formulas, only the AlphaBlend formula does preserve correctly the alpha value. All others, Multiply, Add, Overlay, etc... only work correctly if the target image is completely opaque. After some research, I figured out a general solution that works for all blend modes and treats alpha channel in exactly the same way, regardless of the operation to perform.

These formulas are certainly slower than a plain AlphaBlend, not only because the invoke function, but also because the division inside `Compose` , but I don't (only) seek performance in these formulas, that's why I left the original DrawImage extension (which uses the faster, plain AlphaBlend), In fact... the DrawImage extension I added with the function, lets people experiment with crazy filters, almost as if the function is a pixel shader...

Unit tests.... I wanted to ask about that... how where should I place them ?

JimBobSquarePants:
Thanks for replying @vpenades 

>When I needed to do that I also tried the standard W3C formulas, and soon discovered that, of all the formulas, only the AlphaBlend formula does preserve correctly the alpha value. All others, Multiply, Add, Overlay, etc... only work correctly if the target image is completely opaque.

So does your output produce identical results to Photoshop? I find it odd that the spec [here](https://www.w3.org/TR/compositing-1/#blending) would be incorrect. They describe the correct behavior as the following.

>The result of the mixing function is modulated by the backdrop alpha. A fully opaque backdrop allows the mixing function to be fully realized. A transparent backdrop will cause the final result to be a weighted average between the source color and mixed color with the weight controlled by the backdrop alpha. The value of the new color becomes:

Now I'm not 100% convinced that my work was correct (I am truly terrible at maths) but we should still align with a specification so that results can be compared with that specification. 

Regarding unit tests. Stick them in with all the other color tests in the tests project. I plan to refactor that project soon. 

I'd love to see something working based on your PR, the idea of library-wide composition functionality is truly exciting! 😄

vpenades:
@JimBobSquarePants Yes, the functions I added produce the same results as photoshop (at least, they did a year ago when I programmed them!)

If you notice in the W3C specs, they only do true alpha blending between two transparent images for Alpha Blending AKA Normal mode. For all other modes, they do the maths against a solid color background image.

If you try all the W3C spec functions with two transparent images, you certainly get some nice images, but not what photoshop produces, and certainly not what an artist would expect. Anyway, with the images currently available in the test it's difficult to produce compositions that truly show the difference.

Given that my functions will be slower, It would make sense to keep both models around... the generalized solution for accurate/artistic/complex compositions, and the standard W3C functions for basic composition where performance is more critical.

JimBobSquarePants:
In that case we could probably do a quick check against the background alpha and choose a path as to which method to use. It might be the case though that the difference is negligible and we can just always use yours (if the results are the same for fully opaque backgrounds). We should benchmark both sets of functions to be sure.

vpenades:
@JimBobSquarePants Yes, with this approach, you can choose different strategies to select the best function for every case, for example, if the destination pixel format is know to be opaque, due to not having alpha, you could directly choose the faster functions optimized for opaque backgrounds... maybe it could be interesting to extend IPixel to inform, in some way, it has no alpha, or create another `interface IOpaquePixel : IPixel` that uses Vector3 instead of Vector4 ... I don't know... there's many choices here, but I think it's up to you now to choose the one you like

Btw, I'll be off for a few days... dunno if there's anything else I need to do in my fork...

tocsoft:
@vpenades I'm looking at how to change the internal APIs at the moment to figure out the best/most efficient way of introducing your code. 

My plan is to extend the GraphicsOptions struct allow setting a pixel blender and blend amount on all of the Drawing APIs. 

I'll let you know if there's anything I come across.

vpenades:
Given the pixel blender function is generic, it's going to be difficult to place it on a struct without making it also generic

antonfirsov:
My concerns here:
- Isn't it possible to go without the `PixelTransformMode` enum? It violates the [open/closed](https://en.wikipedia.org/wiki/Open/closed_principle) principle, thus extensibility: for a new pixel function we need to extend the enum touching the library code.
- The design could be changed in a way, that's **much** better in terms of performance, however it's a bigger refactor. @vpenades are you interested making those changes? If not, @JimBobSquarePants how should we manage yet another public API thing, that's slow by design?

antonfirsov:
Here is a proposal, that addresses both of my concerns above:

```C#
public abstract class PixelFunction<TBckPixel, TsrcPixel>
  where TBckPixel : IPixel
  where TsrcPixel : IPixel
{
  public abstract Apply(TBckPixel backdrop, TsrcPixel source, float opacity);
  
  internal virtual BulkApply(BufferSpan<TBckPixel> backdrop, BufferSpan<TsrcPixel> source, float opacity) 
  { 
    // default slow bulk implementation here
  }
}

public class SubstractFunction<TBckPixel, TsrcPixel> : PixelFunction<TBckPixel, TsrcPixel>
  where TBckPixel : IPixel
  where TsrcPixel : IPixel
{
  public abstract Apply(TBckPixel backdrop, TsrcPixel source, float opacity)
  {
    // ...
  }
  
  internal override BulkApply(BufferSpan<TBckPixel> backdrop, BufferSpan<TsrcPixel> source, float opacity)
  {
    // Optimized bulk implementation here
  }
}
```

JimBobSquarePants:
@antonfirsov I would definitely like to avoid enums where possible though the PorterDuff equations are a known set so we should be ok with that here - Same as our Colorblindness methods. That would leave only the enum public.

I am concerned with performance though so I'm interested in anything we can do to approach this in the best possible manner - One that works well with line drawing also. Happy to do the work here of course now we have the correct equations.

With your pattern what actually goes in the `BulkApply` method in the base class? Each function has it's own distinct equation so there wouldn't be anything as far as I can see to go there.

antonfirsov:
@JimBobSquarePants the base class `BulkApply()` calls it's own abstract `Apply` for each pixel in the span.

However this default implementation will be slow, because the `Apply` call won't be inlined here. That's why we need `BulkApply` which will be boilderplate code actually, but run fast.

This pattern is very similar to `BulkPixelOperations<T>`.

src/ImageSharp.Drawing/Processors/DrawImageEffectProcessor.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        /// <summary>
+        /// Gets the image to blend.
+        /// </summary>
+        public Image<TPixel> Image { get; private set; }
+
+        /// <summary>
+        /// Gets The function effect to apply on a per pixel basis
+        /// </summary>
+        public Func<TPixel, TPixel, float, TPixel> PixelFunction { get; private set; }
```
antonfirsov: With my proposal this would be `PixelTransform<TPixel, TPixel> PixelTransform { get; private set; }`

JimBobSquarePants:
I spotted the pattern but I'm not sure we even need the base class - There's nothing for the functions to inherit.

src/ImageSharp.Drawing/Processors/DrawImageEffectProcessor.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            using (PixelAccessor<TPixel> sourcePixels = this.Image.Lock())
+            using (PixelAccessor<TPixel> targetPixels = target.Lock())
+            {
+                Parallel.For(
+                    minY,
+                    maxY,
+                    this.ParallelOptions,
+                    y =>
+                        {
```
antonfirsov: With my proposal, we can do `PixelTransform.BulkApply(target.GetRowSpan(y), source.GetRowSpan(y), alpha)` here instead the nested for loop.

JimBobSquarePants: Ah right!... Sorry, knackered. Late here and I've been working on ICC profile stuff.

antonfirsov: Cool! My continuous typos and edits also did not help to make it understandable :)

tocsoft:
The things you've been saying about bulk apply etc are already things I'm addressing/addressed in my local branch. 
I'll let people see something soon. 

Basically I'm taking queue from things like the `BulkPixelOperations<T>` apis (in fact my plan is to actually rename it `PixelOperations<T>`) and have it be a general pattern/location for doing generic pixel operations that can optionally be optermised on a PixelType by pixel type bases. 

I've just been dragged out for a few hours, but my wife is out on a hen night tonight so I've got all evening and into the night which I'll be devoting to this.

antonfirsov:
@tocsoft I'm sceptical if it's possible to do this without hurting performance. (Actually 90% sure it isn't.) Off topic here, so let's continoue this discussion on gitter :)

vpenades:
Just popping for a second, I'm out and can't follow all the details... Don't disregard a function call as too slow. It was very slow in the first not 1.0 but the performance was improved over time. I don't have benchmarks, but I'm guessing it's probably as fast as any other virtual call. As reference, at work I'm using such functions to process Kinect video buffers in real-time.

antonfirsov:
@vpenades you might be right that the perf is equivalent to a virtual call.
The problem is that when you are processing a large data stream like an image, you even can't afford a virtual call at per item (per pixel) level, because it will lead to frequent cache misses!

We need an inlined call here.

vpenades:
Well... there's always a compromise between performance and flexibility/usability. I understand some people wants the highest performance possible for massive pixel processing... but others like me will priorize having a flexible and easy to use API.

Having said that, it's not my responsability to say where to set the red line between both... I'm just glad to try contribute in any way I can, and even if the proposal is not accepted (I guess will not) I'm just happy some ideas got through... since at the end, I am planning to use ImageSharp for many of my projects.

Regarding performance, There's two ways I can think of that can improve the performance:

First one, is to have a RowBlender instead of a PixelBlender:

```
abstract class RowBlender<TDstPixel,TSrcPixel>
{
public PixelAccessor<TDstPixel> DstImage {get;set;}
public int DstRow {get;set;}
public int DstStart {get;set;}
public PixelAccessor<TSrcPixel> SrcImage {get;set;}
public int SrcRow {get;set;}
public int SrcStart {get;set;}
public int PixelCount {get;set;}
public float Alpha {get;set;}

public abstract void Invoke();
}
```


This function would be called once per row instead of once per pixel, so the performance difference would be negligible.

Second one, the fastest one would be to inline everything, and just generate full blown DrawImageEffectProcessor classes with a T4 template, one per each blend mode.

I don't like the second one because, even it might give performance comparable to the current processor, it removes the possibility to pass user pixel blenders.

The first one, the RowBlender would be a compromise between the easy yo use PixelBlender, and the generated processors... at least it gives users the possibility to program their own pixel blenders.

Sorry I can't follow the thread, I'll be off for 3 or 4 days

Cheers!

antonfirsov:
@vpenades the work your done is great, and it's really valuable, I'm sure we gonna merge it, we just want to refactor it a bit ;)

Unless I'm missing something, my `PixelTransform<T>` proposal is basically equivalent to what you posted as `RowBlender<T>` in terms of performance, but it seems to be more flexible.

@JimBobSquarePants when we've argreed on the final pattern, we need to find the way to manage this refactor. Who's gonna do it? on which branch? etc.

tocsoft:
@vpenades thanks for this, we've now merged in version... We now have a version of `DrawImage` that takes in a enum of type `PixelBlenderModes` that let you pick the type of blending... I the future (post V1 release) we'll introduce a way of providing custom pixel blending functions.

You can specify the blending mode of a majority of Additive operations, drawing images/shapes/text and things like glow etc by setting the `PixelBlender` property and the `PixelBlenderPercentage` property on GraphicOptions that can be passed in.

PR #201, opened 2017-04-28
Reduce register pressure

Drawaes:
Mutlipling

Vector4 * scalar is the same as 

Vector4 * new Vector4(scalar)


JimBobSquarePants:
All good. Thanks!

PR #202, opened 2017-04-29
Pixel blending - Efficient pixel blending + merging Duff-Porter functions

tocsoft:
### Prerequisites

### Description

This merges in #200 

1. The adds the concept of a `PixelBlender<TPixel>` which can be used for blending pixels.
2. Extends/renames `BlukPixelOperations` to `PixelOperations` and adds a `GetBlender()` method to allow us to write pixel type specific implementations of the blenders as required.
3. Migrates all the drawing code both `Brush` based and straight DrawImage based.

`PixelBlender<TPixel>` are internal for now, but then could easily be exposed later.


src/ImageSharp/PixelFormats/PixelBlender{TPixel}.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// Composes 2 pixels together.
+        /// </summary>
+        /// <param name="destination">The destination span.</param>
+        /// <param name="background">The background span.</param>
+        /// <param name="source">The source span.</param>
+        /// <param name="amount">
+        /// A value between 0 and 1 indicating the weight of the second source vector.
+        /// At amount = 0, "from" is returned, at amount = 1, "to" is returned.
+        /// </param>
+        public abstract void Compose(BufferSpan<TPixel> destination, BufferSpan<TPixel> background, BufferSpan<TPixel> source, BufferSpan<float> amount);
```
antonfirsov: I'm not familiar enough with processors, but don't we have naturally a `BufferSpan<Vector4>` at call site instead of `BufferSpan<TPixel>` in some cases? Eg. I know this is the case for certain operations of the `ResizeProcessor`.

If this is true AND our pixel functions operated on `Vector4` instead of `TPixel` we can eliminate lots of useless packing/unpacking calls.

src/ImageSharp/PixelFormats/PorterDuffFunctions.cs
```diff
@@ -0,0 +1,274 @@
...
+    {
+        /// <summary>
+        /// Source over backdrop
+        /// </summary>
+        /// <param name="backdrop">Backgrop color</param>
+        /// <param name="source">Source color</param>
+        /// <param name="opacity">Opacity applied to Source Alpha</param>
+        /// <returns>Output color</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static TPixel NormalBlendFunction(TPixel backdrop, TPixel source, float opacity)
```
antonfirsov: I think it might be worth to provide `Vector4` versions for these functions.
Advantages:
- We can introduce a `BufferSpan<Vector4>`-based bulk-composer method in pixel blender.
- We can optimize `PixelBlender.Compose(BufferSpan<T>, BufferSpan<T>, float)` to use `PixelOperations<T>.ToVector4()` / `PixelOperations<T>.FromVector4()`.
- We could even use `Vector4` based bulk-composer in our processors. (See my comment on `PixelBlender<T>`)

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.PixelBenders.cs
```diff
@@ -0,0 +1,68 @@
...
+#pragma warning disable CS1710 // XML comment has a duplicate typeparam tag
+    /// <summary>
+    /// A stateless class implementing Strategy Pattern for batched pixel-data conversion operations
+    /// for pixel buffers of type <typeparamref name="TPixel"/>.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    public partial class PixelOperations<TPixel>
+#pragma warning restore CS1710 // XML comment has a duplicate typeparam tag
+        where TPixel : struct, IPixel<TPixel>
+    {
```
tocsoft: The single pixel operations are actually on the `PixelBlender<TPixel>` classes themselves

src/ImageSharp/PixelFormats/PorterDuffFunctions.cs
tocsoft: totally agree having Vector4 versions just makes sense.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -0,0 +1,242 @@
...
+
+    /// <summary>
+    /// Collection of Porter Duff alpha blending functions
+    /// </summary>
+    /// <remarks>
+    /// These functions are designed to be a general solution for all color cases,
+    /// that is, they take in account the alpha value of both the backdrop
+    /// and source, and there's no need to alpha-premultiply neither the backdrop
+    /// nor the source.
+    /// Note there are faster functions for when the backdrop color is known
```
JimBobSquarePants: Has anyone benchmarked this? The existing functions that implement  `Vector4BlendTransforms.cs` are per-pixel. I think we should actually scrap that class + `Rgba32` and `RgbaVector` transforms and us a single set of functions throughout.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -0,0 +1,242 @@
...
+            if (source.W == 0)
+            {
+                return backdrop;
+            }
+
+            return Compose(backdrop, source, source);
+        }
+
+        /// <summary>
+        /// Source multiplied by backdrop
```
JimBobSquarePants: Use the method summaries from `Vector4BlendTransforms.cs`

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions{TPixel}.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// <summary>
+    /// Collection of Porter Duff alpha blending functions
+    /// </summary>
+    /// <typeparam name="TPixel">Pixel Format</typeparam>
+    /// <remarks>
+    /// These functions are designed to be a general solution for all color cases,
+    /// that is, they take in account the alpha value of both the backdrop
+    /// and source, and there's no need to alpha-premultiply neither the backdrop
+    /// nor the source.
+    /// Note there are faster functions for when the backdrop color is known
```
JimBobSquarePants: See above

src/ImageSharp/Processing/Processors/Overlays/GlowProcessor.cs
```diff
@@ -65,25 +65,32 @@ protected override void OnApply(ImageBase<TPixel> source, Rectangle sourceRectan
...
-                                packed.PackFromVector4(Vector4BlendTransforms.PremultipliedLerp(sourceColor, glowColor.ToVector4(), 1 - (.95F * (distance / maxDistance))));
-                                sourcePixels[offsetX, offsetY] = packed;
-                            }
-                        });
+                for (int i = 0; i < width; i++)
+                {
+                    rowColors[i] = glowColor;
+                }
+
+                // TODO move GraphicOptions into core so all processes can use it.
```
JimBobSquarePants: Is there anything stopping you doing this as part of the PR?

src/ImageSharp/Processing/Processors/Overlays/GlowProcessor.cs
```diff
@@ -65,25 +65,32 @@ protected override void OnApply(ImageBase<TPixel> source, Rectangle sourceRectan
...
+                PixelBlender<TPixel> blender = PixelOperations<TPixel>.Instance.GetPixelBlender(PixelBlenderMode.Normal);
+                for (int y = minY; y < maxY; y++)
```
JimBobSquarePants: Why can this not be parallel? I see we are using the 3x faster parallel version in `Polaroid()` but this one for `Glow()`

src/ImageSharp/Processing/Processors/Overlays/GlowProcessorParallel.cs
```diff
@@ -0,0 +1,98 @@
...
+    using System.Numerics;
+    using System.Threading.Tasks;
+
+    using ImageSharp.PixelFormats;
+
+    /// <summary>
+    /// An <see cref="IImageProcessor{TPixel}"/> that applies a radial glow effect an <see cref="Image{TPixel}"/>.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    internal class GlowProcessorParallel<TPixel> : ImageProcessor<TPixel>
```
JimBobSquarePants: What's this all about?

src/ImageSharp.Drawing/Processors/DrawImageProcessor.cs
```diff
@@ -57,43 +60,60 @@ public DrawImageProcessor(Image<TPixel> image, Size size, Point location, int al
...
+                Rectangle bounds = this.Image.Bounds;
+                int minX = Math.Max(this.Location.X, sourceRectangle.X);
+                int maxX = Math.Min(this.Location.X + bounds.Width, sourceRectangle.Width);
+                int minY = Math.Max(this.Location.Y, sourceRectangle.Y);
+                int maxY = Math.Min(this.Location.Y + bounds.Height, sourceRectangle.Bottom);
 
-            float alpha = this.Alpha / 100F;
+                int width = maxX - minX;
+                using (Buffer<float> amount = new Buffer<float>(width))
+                using (PixelAccessor<TPixel> toBlendPixels = this.Image.Lock())
```
JimBobSquarePants: There's a bug here. We're not passing the resized image to our blend method.

src/ImageSharp/PixelFormats/PixelBlenders/DefaultHardLightPixelBlender{TPixel}.cs
```diff
@@ -0,0 +1,55 @@
...
+        public static DefaultHardLightPixelBlender<TPixel> Instance { get; } = new DefaultHardLightPixelBlender<TPixel>();
+
+        /// <inheritdoc />
+        public override TPixel Compose(TPixel background, TPixel source, float amount)
+        {
+            return PorterDuffFunctions<TPixel>.HardLightFunction(background, source, amount);
+        }
+
+        /// <inheritdoc />
+        public override void Compose(BufferSpan<TPixel> destination, BufferSpan<TPixel> background, BufferSpan<TPixel> source, BufferSpan<float> amount)
```
antonfirsov: @JimBobSquarePants a long term practical question to think about: 
I think its really worth to DRY away this kind of perf-related boilerplate in the library! Currently it's really hard to refactor this stuff. Optimization and metaprogramming are usually walking hand in hand. Unlsess someone has a better idea, we should T4 this kind of code. (AFAIK the only usable metaprogramming tool in .NET)

src/ImageSharp/PixelFormats/PixelBlenders/DefaultDarkenPixelBlender{TPixel}.cs
```diff
@@ -0,0 +1,55 @@
...
+                BufferSpan<Vector4> destinationSpan = buffer.Slice(0, destination.Length);
+                BufferSpan<Vector4> backgroundSpan = buffer.Slice(destination.Length, destination.Length);
+                BufferSpan<Vector4> sourceSpan = buffer.Slice(destination.Length * 2, destination.Length);
+
+                PixelOperations<TPixel>.Instance.ToVector4(background, backgroundSpan, destination.Length);
+                PixelOperations<TPixel>.Instance.ToVector4(source, sourceSpan, destination.Length);
+
+                for (int i = 0; i < destination.Length; i++)
+                {
+                    destinationSpan[i] = PorterDuffFunctions.DarkenFunction(backgroundSpan[i], sourceSpan[i], amount[i]);
```
antonfirsov: I know this is counter-intuitive, and kills code readability, but we should not use `Span<T>` indexer  in this library on hot path. (Neither the official one!) It will be 30% slower on all runtimes except `corefx`.
`DangerousGetPinnableReference()` + `Unsafe.Add()` FTW!

Regarding the refactor costs of multiple functions, see my next comment!

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.PixelBenders.cs
antonfirsov: I meant operations like `PackFromBytesXyz(byte, byte, byte)`

src/ImageSharp/PixelFormats/PixelBlenders/DefaultHardLightPixelBlender{TPixel}.cs
JimBobSquarePants: I'm gonna have to learn T4 template syntax aren't I? 😝 

I don't find this that hard, just time consuming.

src/ImageSharp/PixelFormats/PixelBlenders/DefaultDarkenPixelBlender{TPixel}.cs
JimBobSquarePants: I thought `BufferSpan<T>` already used that no?

JimBobSquarePants: Or do you mean for when we switch it out?

antonfirsov: Currently BufferSpan indexer has a complicated and slow implementation (in order to correctly handle stuff like AsBytes). Unfortunately the same is true for the official "slow span". You cant implement a fast span without having byref **members**.

src/ImageSharp/Processing/Processors/Overlays/GlowProcessorParallel.cs
tocsoft: I was keeping the old code around while I was rewriting things and forgot to remove. Being removed

src/ImageSharp/PixelFormats/PixelBlenders/DefaultDarkenPixelBlender{TPixel}.cs
JimBobSquarePants: I'm confused.
https://github.com/JimBobSquarePants/ImageSharp/blob/f018838ff98f53573e2761ba7b0d61c44f1951b9/src/ImageSharp/Common/Memory/BufferSpan%7BT%7D.cs#L103

src/ImageSharp/Processing/Processors/Overlays/GlowProcessor.cs
tocsoft: It can be, I forgot to put  `Parallel.For()` back in after testing

src/ImageSharp/Processing/Processors/Overlays/GlowProcessor.cs
tocsoft: nope.. and done

JimBobSquarePants: 👍

src/ImageSharp/Processing/Processors/Overlays/GlowProcessor.cs
JimBobSquarePants: Sweet.

src/ImageSharp/Processing/Processors/Overlays/GlowProcessorParallel.cs
JimBobSquarePants: 👍

src/ImageSharp/PixelFormats/PixelBlenders/DefaultDarkenPixelBlender{TPixel}.cs
antonfirsov: We get a reference (managed pointer) to the first item of the array with `DangerousGetPinnableReference()`, then offset it by `index` to return the indexed position.
My point is that it's enough to call `DangerousGetPinnableReference()` once for the whole loop. We can't implement this short-circuit inside the `Span<T>` itself, `private T ref referenceToFirstElement` is not possible in CLR. 
Only for coreclr codebase owners for [fast span implementation](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/shared/System/Span.cs#L28) owners :)

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.PixelBenders.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        /// <summary>
+        /// Gets the MultiplyBlender.
+        /// </summary>
+        private PixelBlender<TPixel> multiplyBlender = new DefaultMultiplyPixelBlender<TPixel>();
+
+        /// <summary>
+        /// Gets the ScreenBlender.
+        /// </summary>
+        private PixelBlender<TPixel> screenBlender = new DefaultScreenPixelBlender<TPixel>();
```
tocsoft: nope removed, no longer needed they where left over for an earlier pass

src/ImageSharp/PixelFormats/PixelBlenders/DefaultHardLightPixelBlender{TPixel}.cs
antonfirsov: It's not that bad. Quite similar to old ASP/ASPX synthax. The only problem is the lack of builtin intellisense support.

Wanna see a real nightmare from the world of metaprogramming? Check out [this](https://github.com/bioinformatics-centre/BayesTyper/blob/master/bayesTyperUtils/vcf%2B%2B/src/boost/variant/variant.hpp)!

src/ImageSharp/PixelFormats/PixelBlenders/DefaultDarkenPixelBlender{TPixel}.cs
antonfirsov: Nevertheless, if I got the greenlight for T4 from @JimBobSquarePants, I will refactor this later (for all the implementations), so @tocsoft don't bother with it.

tocsoft: awesome, i'll leave that one to you 😉 

I've come across https://github.com/daveaglick/Scripty  as an alternative to T4 before now.

antonfirsov: Yeah I also seen that, and I like it pretty much, but I'm not sure if we can trust Scripty in long term. @daveaglick is practically the only contributor, and he seems to be pretty busy with other projects as well.

tocsoft: actually if we're adding some generation in I'll leave off trying to find an efficient way to wire in the composition modes onto op the blending. At the moment this code is doing the equivalent of 'Over' compositing. But with code generation it will be trivial to expand out the modes to include all the other ones.

daveaglick:
@antonfirsov Is that an offer to help :smile:. Like most OSS, I can't guarantee that Scripty will be viable in the long run. It's certainly possible that at some point I'll pass it on, or a better alternative will come along and it'll stall, or I'll just get burned out. That said, at the moment I'm committed to keeping it going even if the pace of commits isn't super-rapid.

antonfirsov:
@daveaglick Scripty is definitely in the category of projects I'd ❤️ to work on. The .NET world needs better metaprogramming support!
The problem is that currently I'm struggling to keep on with my ImageSharp TODO-s :( Maybe in the next few months I'll have some time to pick something from your issue list 😄

src/ImageSharp/PixelFormats/PixelBlenders/DefaultDarkenPixelBlender{TPixel}.cs
JimBobSquarePants: @antonfirsov Ah yeah, make sense. Didn't realize they were using intrinsics.

Re T4, we're already using it for Block8x8F so I'm not adverse to it I just want use to be kept to a bare minimum.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -0,0 +1,242 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Vector4 NormalBlendFunction(Vector4 backdrop, Vector4 source, float opacity)
+        {
+            source.W *= opacity;
+            if (source.W == 0)
+            {
+                return backdrop;
+            }
+
+            return Compose(backdrop, source, source);
```
JimBobSquarePants: This doesn't need to compose as Normal should return Source if not transparent.

src/ImageSharp/Processing/Processors/Overlays/GlowProcessor.cs
```diff
@@ -65,25 +71,36 @@ protected override void OnApply(ImageBase<TPixel> source, Rectangle sourceRectan
...
+            int width = maxX - minX;
+            using (Buffer<TPixel> rowColors = new Buffer<TPixel>(width))
             using (PixelAccessor<TPixel> sourcePixels = source.Lock())
             {
+                for (int i = 0; i < width; i++)
+                {
+                    rowColors[i] = glowColor;
+                }
+
                 Parallel.For(
```
JimBobSquarePants: Vignette and anything calling it should use the same approach. Time to delete `Vector4BlendTransforms`

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
JimBobSquarePants: Ignore this. Utter nonsense, Head's not in the game this morning.

src/ImageSharp/Processing/Processors/Overlays/GlowProcessor.cs
tocsoft: `Vector4BlendTransforms` is now gone and everything uses the `PixelBlenders` and thus `GraphicOptions` to set them.

JimBobSquarePants: Awesome, J'approuve!

src/ImageSharp/Processing/Effects/Alpha.cs
```diff
@@ -34,12 +34,12 @@ public static Image<TPixel> Alpha<TPixel>(this Image<TPixel> source, int percent
         /// </summary>
         /// <typeparam name="TPixel">The pixel format.</typeparam>
         /// <param name="source">The image this method extends.</param>
-        /// <param name="percent">The new opacity of the image. Must be between 0 and 100.</param>
```
JimBobSquarePants: Mind and update the tests!

JimBobSquarePants:
@tocsoft Do we wanna merge this in now or wait for any T4 stuff? I say now if internal and optimize in beta.

tocsoft:
Merge now, as you said we can worry about optermising later

JimBobSquarePants:
make-it-so

PR #205, opened 2017-05-01
Bug fix gif

### LINKED ISSUE #199: Saving gif gives wrong\missing pallet colors
### Prerequisites

- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
This original image
original-gif

when saving turns into this
bad-gif

looks like a color pallet issue

### Steps to Reproduce
```C#
Image.Load("original-gif.gif").Save("bad-gif.gif");
```

### System Configuration

- ImageSharp version: 1.0.0-alpha7-0031
- Environment (Operating system, version and so on): win10 x64
- .NET Framework version: netcoreapp1.0 and net461
- Additional information:

JimBobSquarePants:
### Prerequisites

### Description
This PR fixes #199 plus a couple of other issues I spotted during testing.


src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -302,16 +312,10 @@ private void WriteGraphicalControlExtension<TPixel>(ImageFrame<TPixel> imageFram
         private void WriteGraphicalControlExtension<TPixel>(ImageBase<TPixel> image, IMetaData metaData, EndianBinaryWriter writer, int transparencyIndex)
             where TPixel : struct, IPixel<TPixel>
         {
-            // TODO: Check transparency logic.
```
dlemstra: This logic seems to be gone because the `DisposalMethod` has been moved to `IMetaData`. Maybe you should make the DisposalMethod nullable so you can check if you can copy the value from the original. But I do wonder if this logic makes any sense though. We can probably just have `DisposalMethod.Unspecified` as the default.

src/ImageSharp/MetaData/ImageMetaData.cs
```diff
@@ -116,14 +118,12 @@ public double VerticalResolution
...
-        /// <summary>
-        /// Gets or sets the frame delay for animated images.
-        /// If not 0, this field specifies the number of hundredths (1/100) of a second to
-        /// wait before continuing with the processing of the Data Stream.
-        /// The clock starts ticking immediately after the graphic is rendered.
-        /// </summary>
+        /// <inheritdoc/>
         public int FrameDelay { get; set; }
 
+        /// <inheritdoc/>
```
dlemstra: Not sure why you want to use `inheritdoc` here but don't do it for the other fields. Think it's better to just put the comments here. Never seen `inheritdoc` do anything in the code completion.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -332,6 +333,7 @@ private void ReadFrame()
         /// </summary>
         /// <param name="imageDescriptor">The <see cref="GifImageDescriptor"/>.</param>
         /// <param name="indices">The pixel array to write to.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
dlemstra: Do we really need this?

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -506,14 +508,20 @@ private void RestoreToBackground(ImageBase<TPixel> frame)
...
         /// <summary>
-        /// Sets the frame delay in the metadata.
+        /// Sets the frames metadata.
         /// </summary>
         /// <param name="metaData">The meta data.</param>
-        private void SetFrameDelay(IMetaData metaData)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void SetFrameMetaData(IMetaData metaData)
         {
             if (this.graphicsControlExtension != null && this.graphicsControlExtension.DelayTime > 0)
```
dlemstra: You probably want to remove the second check from this if? You moved it inside already.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -74,7 +79,13 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
             this.bitDepth = ImageMaths.GetBitsNeededForColorDepth(quality);
 
             // Quantize the image returning a palette.
-            QuantizedImage<TPixel> quantized = ((IQuantizer<TPixel>)this.Quantizer).Quantize(image, quality);
+            this.hasMultipleFrames = image.Frames.Any();
```
dlemstra: I know this is correct but the statement looks confusing when you look at it the first time. Maybe you should call it `hasFrames` instead?

JimBobSquarePants: Yeah, that's clearer, will do.

src/ImageSharp/MetaData/ImageMetaData.cs
JimBobSquarePants: Ugh, just realised that VS doesn't currently support intellisense based on inheritdoc. I don't want to stop using it though as it massively cuts down on errors duplicating XML docs and Wyam already supports it for our API docs. 

It looks like MVC are planning to use post processing to augment their XML output. I think we should follow suite. 
https://github.com/aspnet/Mvc/issues/928

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
JimBobSquarePants: Original logic was incorrect. `DisposalMethod.Unspecified` is already the default and unless we are doing something crazy with the gif between decoding/encoding we should definitely copy the properties across. Devs have the power now to set each frame value independently should they choose.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
JimBobSquarePants: Yeah, good to remove.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
JimBobSquarePants: There's no harm in doing this. We want to inline as many small private methods as we can really.

JimBobSquarePants:
Imma gonna merge this. It all works and any changes can be made internally.

PR #206, opened 2017-05-01
Change BitCovnerter to be processor endianness aware

Drawaes:
### Description
Currently if the library was run on an BigEndian Processor it will not work, instead I have changed the Endianness Converters to Reverse and Native. This is then selected based on the 

BitConverter.IsLittleEndian flag.

### DIFF
--- a/src/ImageSharp/IO/EndianBitConverter.cs
+++ b/src/ImageSharp/IO/EndianBitConverter.cs
@@ -16,12 +16,12 @@ namespace ImageSharp.IO
         /// <summary>
         /// The little-endian bit converter.
         /// </summary>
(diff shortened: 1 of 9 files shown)


JimBobSquarePants:
We've been using these classes to read formats we know to be stored either big or little endian format. How does this affect that?

Drawaes:
Well all the tests pass :) Basically if you run on a little endian processor there is no change. If you read it on a bigendian processor it flips it. For instance if you say "ReadBigEndian" and you are on a bigendian processor it doesn't flip the bits

JimBobSquarePants:
I'll check the physical output of the decode/encode tests to be sure.

Drawaes:
Don't waste your time, I will make a new one side by side first and withdraw these changes

JimBobSquarePants:
Ok coke. Thanks!

Issue #208, opened 2017-05-04
Remove non-generic Image class


JimBobSquarePants:
Opening this issue to track the removal of the non-generic Image class from the API. 

The issue as it stands. 

While this works and is simple to follow. 

```
using (Image image = Image.Load("foo.jpg"))
{
    image.Resize(image.Width / 2, image.Height / 2)
         .Save("bar.jpg"); 
}
```

The following will not compile since we cannot implicitly cast from `Image<Rgba32>` to `Image` and all our methods return `Image<TPixel>`

```
using (Image image = Image.Load("foo.jpg"))
{
    Image image2 = image.Resize(image.Width / 2, image.Height / 2);
}
```
It's infeasible to create overloads where we cast, not just due to the maintenance workload but for the performance implications also so the best bet is to simply remove the class.

I don't consider generics to be a blocker. Linq is hugely popular.

antonfirsov:
Unfortunately the pixel type (`Argb32`) is a noise for most of the server-side users, however it should not be an issue, because people are usually going with `var`.

With a static, `Argb32`-specific `Image` class, the following code should keep compiling:
```C#
using (var image = Image.Load("foo.jpg"))
{
    image.Resize(image.Width / 2, image.Height / 2)
         .Save("bar.jpg"); 
}
```

PS: @JimBobSquarePants I know this is not your favorite argument, but still true 😄

JimBobSquarePants:
😄 Yeah true but hey still clunky! Generics all the way baby!

tocsoft:
Just want to capture this too. When this is done we have some places in ImageSharp.Drawing that have a similar pattern, we have various Brushes, and a Pen class that should also be removed at the same time.

PR #209, opened 2017-05-05
Remove non generic image

JimBobSquarePants:
### Prerequisites

### Description
This PR does two things fixing #208 

1. Removes the non-generic `Image`, `Brushes`, `Brush`, `Pens`, and `Pen `classes in favour of their generic `TPixel` equivalents.
2. Promotes `Rgba32` to the root `ImageSharp` namespace to improve visibility.

Most developers, since `var` is so common, will now do this to load/manipulate images.
```
using (var image = Image<Rgba32>.Load("foo.jpg"))
{
         .Grayscale()
         .Save("bar.jpg"); 
}
```


tocsoft:
In my opinion feels nicer to type `Image.Load<TPixel>(...)` than `Image<TPixel>.Load(...)`

JimBobSquarePants:
Yeah, you're right. That would better reflect existing types like `Enumerable` also.

tocsoft:
Maybe even just leave `Image<TPixel>.Load()` in but also add `Image.Load<TPixel>()` versions in that just call down into the first.

JimBobSquarePants:
Thanks! @dlemstra

dlemstra:
I have to agree with @tocsoft and I also like `Image.Load<TPixel>()` better. Maybe we should just have the following options available:

- `Image<TPixel>.Load()`
- `Image.Load<TPixel>()`
- `Image.Load()` = `Image.Load<Rgba32>()`

I would also like the last one to be available because most people will probably only use the Rgba32 variant.

JimBobSquarePants:
Almost finished making the other changes but the last one would require me to create an overload for every overload of Load. Even saying sounds crazy. Devs should use the generic method.

dlemstra:
I do think we only need that for two overloads. The one with only a file name and the one with only a stream. And if you want more you should use one of the generic overloads.

JimBobSquarePants:
@dlemstra ok coke. Only cos you're awesome 😄

dlemstra:
That should be extra super awesome instead of just awesome. And thanks great leader!

JimBobSquarePants:
OK. So adding the `Rgba32` overloads for all the `Load` methods was less work than I thought.

Rgba32 can use either `Image.Load()`or `Image.Load<Rgba32>()`

All other pixel formats would use `Image.Load<TPixel>()`

I think we're good to go now.

src/ImageSharp/Image/Image.Create.cs
```diff
@@ -7,56 +7,46 @@ namespace ImageSharp
...
         /// <param name="configuration">
         /// The configuration providing initialization code which allows extending the library.
         /// </param>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
         /// <returns>
-        /// A new <see cref="Image{TPixel}"/> unless <typeparamref name="TPixel"/> is <see cref="Rgba32"/> in which case it returns <see cref="Image" />
+        /// A new <see cref="Image{TPixel}"/>.
         /// </returns>
-        internal static Image<TPixel> Create<TPixel>(int width, int height, ImageMetaData metadata, Configuration configuration)
+        internal static Image<TPixel> Create<TPixel>(int width, int height, Configuration configuration)
```
tocsoft: we can get rid of these now, now we don't have to do the `if (typeof(TPixel) == typeof(Rgba32))` thing, and just call the constructor directly from the `IImageDecoder`s

JimBobSquarePants:
Thanks for finishing this up guys, was way past my bedtime!

PR #212, opened 2017-05-09
Use Bufferspan + Unsafe.Add for png filters

JimBobSquarePants:
### Prerequisites

### Description
This PR replaces the old `fixed` png filters with methods utilizing `BufferSpan<byte>` + `Unsafe.Add` to bring the format codebase more inline with the newer memory access struct approach.

### DIFF
--- a/src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
+++ b/src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
@@ -22,22 +22,25 @@ namespace ImageSharp.Formats
         /// <param name="bytesPerScanline">The number of bytes per scanline</param>
         /// <param name="bytesPerPixel">The bytes per pixel.</param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void Decode(ref byte scanline, ref byte previousScanline, int bytesPerScanline, int bytesPerPixel)
+        public static void Decode(BufferSpan<byte> scanline, BufferSpan<byte> previousScanline, int bytesPerScanline, int bytesPerPixel)
         {
+            ref byte scanPointer = ref scanline.DangerousGetPinnableReference();
+            ref byte prevPointer = ref previousScanline.DangerousGetPinnableReference();
+
             // Average(x) + floor((Raw(x-bpp)+Prior(x))/2)
             for (int x = 1; x < bytesPerScanline; x++)
             {
                 if (x - bytesPerPixel < 1)
                 {
-                    ref byte scan = ref Unsafe.Add(ref scanline, x);
-                    ref byte above = ref Unsafe.Add(ref previousScanline, x);
+                    ref byte scan = ref Unsafe.Add(ref scanPointer, x);
+                    byte above = Unsafe.Add(ref prevPointer, x);
                     scan = (byte)((scan + (above >> 1)) % 256);
                 }
                 else
                 {
-                    ref byte scan = ref Unsafe.Add(ref scanline, x);
-                    ref byte left = ref Unsafe.Add(ref scanline, x - bytesPerPixel);
-                    ref byte above = ref Unsafe.Add(ref previousScanline, x);
+                    ref byte scan = ref Unsafe.Add(ref scanPointer, x);
+                    byte left = Unsafe.Add(ref scanPointer, x - bytesPerPixel);
+                    byte above = Unsafe.Add(ref prevPointer, x);
                     scan = (byte)((scan + Average(left, above)) % 256);
                 }
             }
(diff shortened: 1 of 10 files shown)


JimBobSquarePants:
I'm fairly certain codecov is utterly bonkers. Some of the files that are reporting as a loss in coverage are completely unrelated.

src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
```diff
@@ -22,18 +22,23 @@ internal static unsafe class AverageFilter
...
+            for (int x = 1; x < bytesPerScanline; x++)
             {
-                for (int x = 1; x < bytesPerScanline; x++)
+                if (x - bytesPerPixel < 1)
                 {
-                    byte left = (x - bytesPerPixel < 1) ? (byte)0 : scan[x - bytesPerPixel];
-                    byte above = prev[x];
-
-                    scan[x] = (byte)((scan[x] + Average(left, above)) % 256);
+                    ref byte scan = ref Unsafe.Add(ref scanline, x);
```
antonfirsov: We are not writing to `scan` and `above`, I have the feeling, it would be cleaner to go without `ref` with them:
```C#
byte scan = Unsafe.Add(ref scanline, x);
```

src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
```diff
@@ -22,18 +22,23 @@ internal static unsafe class AverageFilter
         /// <param name="bytesPerScanline">The number of bytes per scanline</param>
         /// <param name="bytesPerPixel">The bytes per pixel.</param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void Decode(byte[] scanline, byte[] previousScanline, int bytesPerScanline, int bytesPerPixel)
+        public static void Decode(ref byte scanline, ref byte previousScanline, int bytesPerScanline, int bytesPerPixel)
```
antonfirsov: Not entirely sure on this point, but:
Wouldn't it be better to take two `BufferSpan<byte>` parameters on these methods, and keep the "reference extracting" step as an implementation detail?
Advantage: cleaner method declaration.
Disadvantage: code duplication in method bodies.

src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
JimBobSquarePants: Yeah that's cool. Can be done.

src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
JimBobSquarePants: The reason I kept it like that was due to the encoder versions requiring the refs (We don't want to declare them twice for encoding then counting) so it felt better to use the same pattern. It does all feel hella clunky though. 

I really wish we could do some sort of clever wrapper using `Span<T>` and `#ifdefs`

src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
antonfirsov: It will be copied by value in all cases to make the necessary calculations. For me it feels better to do it right in the variable declaration.

src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
antonfirsov: Not sure if I understand this right now, I'll have another look tomorrow. If possible, I prefer span on the API-s.

src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
```diff
@@ -22,18 +22,26 @@ internal static unsafe class AverageFilter
         /// <param name="bytesPerScanline">The number of bytes per scanline</param>
         /// <param name="bytesPerPixel">The bytes per pixel.</param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void Decode(byte[] scanline, byte[] previousScanline, int bytesPerScanline, int bytesPerPixel)
+        public static void Decode(BufferSpan<byte> scanline, BufferSpan<byte> previousScanline, int bytesPerScanline, int bytesPerPixel)
         {
+            ref byte scanPointer = ref scanline.DangerousGetPinnableReference();
```
KodrAus: Hi there! :)

Just curious how you know for sure these buffers will be pinned? If the default `ArrayPool` allocates it's slabs on LOH then it'd be safe so long as callers haven't tweaked the GC config. Not sure to be honest, I haven't actually checked that, but AFAIK the `ArrayPool` implementation in CoreFX doesn't guarantee the buffers it gives you will be pinned.

In CoreFXLab for instance, there's [a pattern for retaining a buffer](https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs#L62-L71), which uses a `GcHandle` to pin if it's not expected to be already.

JimBobSquarePants: Yeah very little pinning going on in the library now.

antonfirsov: Maybe we need a common naming rule for these variables. 
I have to confess that I'm also mixing stuff: sometimes I name them `blahFooBarPointer` on other places I use `blahFooBarBase`. This is confusing.

KodrAus: I don't know why I didn't notice the `ref` part first time I looked. I just saw:

```diff
+ DangerousGetPinnableReference()
- fixed (byte*)
```
And thought they were still raw pointers. A common naming rule sounds like a good call 👍

JimBobSquarePants: @antonfirsov  `blahFooBarPointer` would be the technically correct name methinks. I reckon this is good to go now.

@KodrAus Easy enough to miss when you're not used to the `Span<T>` style api

antonfirsov: @JimBobSquarePants `fooBarPointer` suggests it's a native pointer. `fooBarBaseReference` seems to be the most correct name (but too verbose?).

antonfirsov: OFF:
@KodrAus really cool to see a Rust developer around! Learning Rust is on my long list of things I want to do one time! :)

JimBobSquarePants: `fooBarBaseRef`?

JimBobSquarePants: Shortening to Ref is fine since `ref` is a known keyword.

src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
```diff
@@ -44,23 +52,34 @@ public static void Decode(byte[] scanline, byte[] previousScanline, int bytesPer
         /// <param name="scanline">The scanline to encode</param>
         /// <param name="previousScanline">The previous scanline.</param>
         /// <param name="result">The filtered scanline result.</param>
+        /// <param name="bytesPerScanline">The number of bytes per scanline</param>
         /// <param name="bytesPerPixel">The bytes per pixel.</param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void Encode(byte[] scanline, byte[] previousScanline, byte[] result, int bytesPerPixel)
+        public static void Encode(BufferSpan<byte> scanline, BufferSpan<byte> previousScanline, BufferSpan<byte> result, int bytesPerScanline, int bytesPerPixel)
```
antonfirsov: There is no need to pass `bytesPerScanline` as a 3rd parameter, if the `scanline` span had the lenght `bytesPerScanline`.
Can I drop this in into this PR? I would also replace `this.scanLine` and `this.previousScanline` with `Buffer<T>`.

JimBobSquarePants: Go for it. Just realised that we're not actually renting those arrays.

antonfirsov:
@JimBobSquarePants needed to do a bit more refactor than I expected. Now your turn to review my changes!

src/ImageSharp/Formats/Png/Filters/AverageFilter.cs
```diff
@@ -12,28 +12,37 @@ namespace ImageSharp.Formats
...
         /// </summary>
         /// <param name="scanline">The scanline to decode</param>
         /// <param name="previousScanline">The previous scanline.</param>
-        /// <param name="bytesPerScanline">The number of bytes per scanline</param>
         /// <param name="bytesPerPixel">The bytes per pixel.</param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void Decode(byte[] scanline, byte[] previousScanline, int bytesPerScanline, int bytesPerPixel)
+        public static void Decode(BufferSpan<byte> scanline, BufferSpan<byte> previousScanline, int bytesPerPixel)
         {
+            Guard.MustBeSameSized(scanline, previousScanline, nameof(scanline));
```
antonfirsov: @JimBobSquarePants maybe DebugGuard could be enough here?

JimBobSquarePants:
Maaan that's so much cleaner! I totally forgot about the implicit cast from Buffer to BufferSpan! 😄 

I think we could get away with DebugGuard, everything else looks sweet!

PR #213, opened 2017-05-11
Replace BufferSpan<T> with Span<T>

JimBobSquarePants:
### Prerequisites

### Description
This is quite a biggie so I'd like you all to double check what I have done. The list is as follows:

- Add System.Memory
- Update System.Runtime.CompilerServices.Unsafe
- Move memory classes/structs to ImageSharp.Memory (I want devs to use when they know what they are doing plus all internal for now)
- Replace BufferSpan<T> with Span<T> and adjust methods accordingly

It's all one big fat check-in. Apologies for that.


src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
```diff
@@ -85,28 +88,29 @@ internal virtual void PackFromXyzBytes(BufferSpan<byte> sourceBytes, BufferSpan<
...
         /// Bulk version of <see cref="IPixel.ToXyzBytes(byte[], int)"/>.
         /// </summary>
-        /// <param name="sourceColors">The <see cref="BufferSpan{T}"/> to the source colors.</param>
-        /// <param name="destBytes">The <see cref="BufferSpan{T}"/> to the destination bytes.</param>
+        /// <param name="sourceColors">The <see cref="Span{T}"/> to the source colors.</param>
+        /// <param name="destBytes">The <see cref="Buffer{T}"/> to the destination bytes.</param>
+        /// <param name="startIndex">The starting index of the <paramref name="destBytes"/>.</param>
         /// <param name="count">The number of pixels to convert.</param>
-        internal virtual void ToXyzBytes(BufferSpan<TPixel> sourceColors, BufferSpan<byte> destBytes, int count)
+        internal virtual void ToXyzBytes(Span<TPixel> sourceColors, Buffer<byte> destBytes, int startIndex, int count)
```
tocsoft: why can't `destBytes` be a span? can't ToXyzBytes just be updated to take in a `Span<byte>` too?

_this also applies to the other similar methods too_

antonfirsov: @tocsoft the problem is with the [`IPixel.To***Bytes()` API-s](https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/PixelFormats/IPixel.cs#L59). They consume an array now.
I think we definitely need to change that, but this is a more complex topic, outside the scope of this refactor.
Temporally it can be `void ToXyzBytes(Span<byte> bytes, int startIndex)`, but I have an idea for a better solution :smile:

tocsoft: A span inherently has a start index tho(ie the start of the span) just pass in a new span that's been offset correctly from the source one... Or is there some perf issues here I can't see?.. but yes we can easily defer that to a later change.

antonfirsov: Yes, there will be perf. issues. Even with the `void ToXyzBytes(Span<byte> bytes, int startIndex)` version.

These API-s should be optimized further in terms of performance and type safety:
We don't want to set random bytes in a span or an array here. We only want to set a single `Rgb24` value.

JimBobSquarePants: Thanks for explaining @antonfirsov Yeah, limited by current IPixel API and how we then communicate with streams. This is a workaround to get things working at least.

JimBobSquarePants:
@dlemstra I moves it cos these classes/structs require a specific knowledge of the code base and are not commonly used.

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
antonfirsov: Still, I don't think it's a good API in it's current form. 
1. As described in this issue it can't be used with `Span<T>` without having a perf loss.
2. There are [several](https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/Quantizers/OctreeQuantizer%7BTPixel%7D.cs#L29) [places](https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/Formats/Png/PngEncoderCore.cs#L544) where we are allocating 4-sized arrays to extract RGBA values.

I wanted to propose changing this for a long time. I want to use this occasion, and @JimBobSquarePants's fresh head, so it's time for a gist! ;)

JimBobSquarePants: The API was a good start but we have better toys to play with now so sounds good to me. Let's see what you come up with.

antonfirsov: [Here it is!](https://gist.github.com/antonfirsov/a5e8c373f56b5f6696071cf555216f87)

JimBobSquarePants: YES!! Love it!

tocsoft: Love it to. love how it removes any chance of index bound checking, and array allocations.

antonfirsov: I will do it soon then. Is it possible to temporally change `IPixel` API to consume `Span<byte>`+ index and rollback changes on `PixelOperations<T>` to use spans again? 
Seems easier to make the `PixelOperations<T>` rollback it in this PR, than doing it later.

JimBobSquarePants: Do what you need on this PR.

antonfirsov:
I think I'm done here (unless someone finds a mistake).
`Span<T>` is now literally everywhere!

JimBobSquarePants:
Great work!

PR #220, opened 2017-05-19
Load from pixel data methods

tocsoft:
### Prerequisites

### Description

We've have multiple people recently asking how to push the raw pixel data into imagesharp.

So I've added a couple of new methods (with multiple overloads) of `Image.LoadPixelData(...)`

There are 2 variants one that takes `TPixel[]` and another taking `byte[]`. The version taking `byte[]` processes the bytes as raw memory holding pixel data based on which ever `TPixel` is being used, thus if you have a `byte[]` holding data in argb32 you would have to load that using `Image.LoadPixelData<Argb32>(...)` etc.

### DIFF
--- a/src/ImageSharp/Image/Image.LoadPixelData.cs
+++ b/src/ImageSharp/Image/Image.LoadPixelData.cs
@@ -14,49 +14,36 @@ namespace ImageSharp
     using ImageSharp.PixelFormats;
 
     /// <content>
-    /// Adds static methods allowing the creation of new image from a byte array.
+    /// Adds static methods allowing the creation of new image from raw pixel data.
     /// </content>
     public static partial class Image
     {
         /// <summary>
-        /// Create a new instance of the <see cref="Image{TPixel}"/> class from the given pixel data.
+        /// Create a new instance of the <see cref="Image{TPixel}"/> class from the raw <typeparamref name="TPixel"/> data.
         /// </summary>
         /// <param name="data">The byte array containing image data.</param>
         /// <param name="width">The width of the final image.</param>
         /// <param name="height">The height of the final image.</param>
         /// <typeparam name="TPixel">The pixel format.</typeparam>
         /// <returns>A new <see cref="Image{TPixel}"/>.</returns>
-        public static Image<TPixel> LoadPixelData<TPixel>(TPixel[] data, int width, int height)
+        public static Image<TPixel> LoadPixelData<TPixel>(Span<TPixel> data, int width, int height)
             where TPixel : struct, IPixel<TPixel>
             => LoadPixelData(Configuration.Default, data, width, height);
 
         /// <summary>
-        /// Create a new instance of the <see cref="Image{TPixel}"/> class from the given pixel data.
-        /// </summary>
-        /// <param name="config">The config for the decoder.</param>
-        /// <param name="data">The byte array containing image data.</param>
-        /// <param name="width">The width of the final image.</param>
-        /// <param name="height">The height of the final image.</param>
-        /// <typeparam name="TPixel">The pixel format.</typeparam>
-        /// <returns>A new <see cref="Image{TPixel}"/>.</returns>
-        public static Image<TPixel> LoadPixelData<TPixel>(Configuration config, TPixel[] data, int width, int height)
-            where TPixel : struct, IPixel<TPixel>
-            => LoadPixelData(config, new Span<TPixel>(data), width, height);
-
-        /// <summary>
-        /// Create a new instance of the <see cref="Image{TPixel}"/> class from the given byte array as raw pixel data.
+        /// Create a new instance of the <see cref="Image{TPixel}"/> class from the given byte array in <typeparamref name="TPixel"/> format.
         /// </summary>
         /// <param name="data">The byte array containing image data.</param>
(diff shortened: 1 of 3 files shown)


JimBobSquarePants:
@dlemstra You mean a RAW format? https://en.wikipedia.org/wiki/Raw_image_format

This is different though, there's no format only bytes.

dlemstra:
Thats not what I mean't. I mean raw pixel data in bytes. And make it possible to both read and write it using an encoder and decoder. Instead of having a new Load overload for this.

src/ImageSharp/Image/Image.LoadPixelData.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <param name="height">The height of the final image.</param>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <returns>A new <see cref="Image{TPixel}"/>.</returns>
+        public static Image<TPixel> LoadPixelData<TPixel>(Configuration config, byte[] data, int width, int height)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            int size = width * height;
+            using (var sourceBuffer = new Buffer<TPixel>(size))
+            {
+                PixelOperations<TPixel>.Instance.PackFromRawBytes(new Span<byte>(data), sourceBuffer.Span, size);
```
antonfirsov: What was your goal with `LoadPixelData<TPixel>(Configuration config, byte[] data, int width, int height)` ? Is it meant to be a stream in `TPixel` format or is it meant to be a stream of `Rgba32`.

- It's not well documented, the user needs to guess.
- If the `sourceBytes` meant to have it's memory layout in `TPixel` format, we don't need `PackFromRawBytes()`. It's always a memcpy, you can just call `SpanHelpers.Copy(data, buffer.Span.AsBytes(), size)`
- If `sourceBytes` meant to be in `Rgba32`, your `PackFromRawBytes()` implementation is wrong!

Personally I think we need the first variant here: loading byte data having identical memory layout to `TPixel[]`. In this case you just need to unsafe cast your `byte` span to `TPixel` span, no need for an extra copy.

tocsoft:
The problem is, it would be very easy to miss the fact that the width and height are required values if its a Format. 

Additionally a format implies the input/output is round trappable but that wouldn't be the case as the data would be missing the important dimension data.

src/ImageSharp/Image/Image.LoadPixelData.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <summary>
+        /// Create a new instance of the <see cref="Image{TPixel}"/> class from the given pixel data.
+        /// </summary>
+        /// <param name="config">The config for the decoder.</param>
+        /// <param name="data">The byte array containing image data.</param>
+        /// <param name="width">The width of the final image.</param>
+        /// <param name="height">The height of the final image.</param>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <returns>A new <see cref="Image{TPixel}"/>.</returns>
+        public static Image<TPixel> LoadPixelData<TPixel>(Configuration config, TPixel[] data, int width, int height)
```
antonfirsov: I wonder if we even need the array variant of the API-s. There is an implicit cast from `TPixel[]` to `Span<TPixel>`.

src/ImageSharp/Image/Image.LoadPixelData.cs
tocsoft: its supposed to be the first variant... i'll try to improve comments, and switch out the span copy.

antonfirsov: [NonPortableCast](https://github.com/dotnet/corefx/blob/ea67b20d33f240c6de7dba12a81f21de06f188eb/src/System.Memory/src/System/Span.cs#L554) is your friend!

src/ImageSharp/Image/Image.LoadPixelData.cs
antonfirsov: @JimBobSquarePants thoughts?

dlemstra:
Using an decoder/encoder for this would also make it possible to add some specific options in the future. That would make it possible to read only the data of the red channel as a double (in bytes). That might be a step to far at this point but it won't be that easy to remove your new load methods from the public API when we would want to do that.

antonfirsov:
@dlemstra I think implementing this feature as a decoder would be misleading to the users. It's a very special case.

antonfirsov:
> That would make it possible to read only the data of the red channel as a double (in bytes).

antonfirsov:
> That would make it possible to read only the data of the red channel as a double (in bytes).

This can be solved by `LoadFromPixelData` (super fast) + a pixel type conversion call. I think this API is easier to understand, than exposing complicated DecoderOptions.

src/ImageSharp/Image/Image.LoadPixelData.cs
JimBobSquarePants: Agree. Drop it, the implicit cast is enough and I think MS want to push people to use `Span<T>` as much as possible.

JimBobSquarePants:
This is the best approach IMO. It's clean and conveys the right message.

PR #221, opened 2017-05-24
Internalize PixelAccessor<T>

JimBobSquarePants:
### Prerequisites

### Description

This PR internalizes `PixelAccesor<TPixel>` allowing us to provide a more sensible API and further optimize our various algorithms. 

Image pixels can now be accessed using:

1. `image[x, y]`;
2. `ref image.GetPixelReference(x, y)` C#7+
3. `image.GetRowSpan(y)`
4. `image.GetRowSpan(x, y)`

Which feels much cleaner to me.

I've also optimized some low hanging fruit in the codebase and am going through all the existing processors at least to replace existing code. This benchmark shows the obvious benefits.
```
Frequency=2742191 Hz, Resolution=364.6719 ns, Timer=TSC
  [Host]     : Clr 4.0.30319.42000, 64bit RyuJIT-v4.7.2046.0
  DefaultJob : Clr 4.0.30319.42000, 64bit RyuJIT-v4.7.2046.0
```

Method |      Mean |    StdDev | Scaled | Scaled-StdDev | Allocated |
-------------------------------- |---------- |---------- |------- |-------------- |---------- |
'PixelAccessor Copy by indexer' | 2.3211 ms | 0.0466 ms |   1.00 |          0.00 |   2.65 kB |
'PixelAccessor Copy by Span' | 1.1721 ms | 0.0235 ms |   0.51 |          0.01 |   2.52 kB |
'Copy by indexer' | 1.7352 ms | 0.0472 ms |   0.75 |          0.02 |   2.41 kB |
'Copy by Span' | 1.1033 ms | 0.0228 ms |   0.48 |          0.01 |   2.27 kB |

I'm sure there's more than can be done internally in places with `PixelOperations` but this is a very healthy start.

### DIFF
--- a/src/ImageSharp/Image/ImageBase{TPixel}.cs
+++ b/src/ImageSharp/Image/ImageBase{TPixel}.cs
@@ -32,15 +32,12 @@ namespace ImageSharp
         /// </summary>
         public const int MaxHeight = int.MaxValue;
 
+#pragma warning disable SA1401 // Fields must be private
         /// <summary>
-        /// The image pixels
+        /// The image pixels. Not private as Buffer2D requires an array in its constructor.
         /// </summary>
-        private TPixel[] pixelBuffer;
-
-        /// <summary>
-        /// The span representing the pixel buffer
-        /// </summary>
-        private Span<TPixel> span;
+        internal TPixel[] PixelBuffer;
+#pragma warning restore SA1401 // Fields must be private
 
         /// <summary>
         /// A value indicating whether this instance of the given entity has been disposed.
@@ -116,7 +113,7 @@ namespace ImageSharp
         }
 
         /// <inheritdoc/>
-        public TPixel[] Pixels => this.pixelBuffer;
+        public Span<TPixel> Pixels => new Span<TPixel>(this.PixelBuffer, 0, this.Width * this.Height);
 
         /// <inheritdoc/>
         public int Width { get; private set; }
@@ -147,14 +144,14 @@ namespace ImageSharp
             get
             {
                 this.CheckCoordinates(x, y);
-                return this.pixelBuffer[(y * this.Width) + x];
+                return this.PixelBuffer[(y * this.Width) + x];
             }
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             set
             {
                 this.CheckCoordinates(x, y);
-                this.pixelBuffer[(y * this.Width) + x] = value;
+                this.PixelBuffer[(y * this.Width) + x] = value;
             }
         }
 
@@ -168,7 +165,7 @@ namespace ImageSharp
         public ref TPixel GetPixelReference(int x, int y)
         {
             this.CheckCoordinates(x, y);
-            return ref this.pixelBuffer[(y * this.Width) + x];
+            return ref this.PixelBuffer[(y * this.Width) + x];
         }
 
         /// <summary>
@@ -180,7 +177,7 @@ namespace ImageSharp
         public Span<TPixel> GetRowSpan(int y)
         {
             this.CheckCoordinates(y);
-            return this.span.Slice(y * this.Width, this.Width);
+            return this.Pixels.Slice(y * this.Width, this.Width);
         }
 
         /// <summary>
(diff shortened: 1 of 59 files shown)


src/ImageSharp/Image/ImageBase{TPixel}.cs
```diff
@@ -37,6 +38,11 @@ public abstract class ImageBase<TPixel> : IImageBase<TPixel>
         private TPixel[] pixelBuffer;
 
         /// <summary>
+        /// The span representing the pixel buffer
+        /// </summary>
+        private Span<TPixel> span;
```
antonfirsov: [Span could be used as member in only in stack-only structs!](https://github.com/dotnet/corefxlab/blob/master/docs/specs/span.md#spant-will-be-stack-only)

MS is planning to provide a Roslyn analyzer to ensure this.

I think we should change the `Pixels` property to return a `Span<TPixel>` instead having this member! There are several reasons to do this:

- The "tail" of the `pixelBuffer` array is junk (extra bytes allocated by `ArrayPool`), so current `Pixels.Length` returns a random power-of-2 number unrelated to `Width*Height`
- Exposing pixel data as an array won't be possible after introducing pluggable memory management. (The source of the `Pixels` span might be unmanaged memory, an MMF buffer or anything.

src/ImageSharp/Processing/Processors/Convolution/Convolution2DProcessor.cs
```diff
@@ -56,17 +57,19 @@ protected override void OnApply(ImageBase<TPixel> source, Rectangle sourceRectan
             int maxY = endY - 1;
             int maxX = endX - 1;
 
-            using (PixelAccessor<TPixel> targetPixels = new PixelAccessor<TPixel>(source.Width, source.Height))
-            using (PixelAccessor<TPixel> sourcePixels = source.Lock())
+            using (var targetPixels = new PixelAccessor<TPixel>(source.Width, source.Height))
```
antonfirsov: We can use `Buffer2D` on these places, possibly replacing all usages of `PixelAccessor`. 
But this change might be big enough to keep it for another PR ..

src/ImageSharp/Image/ImageBase{TPixel}.cs
JimBobSquarePants: Ah yes, of course!

I'm happy to change pixels to `Span<TPixel>` makes sense and certainly does no harm.

JimBobSquarePants: I'm assuming I'm able to assign that once vs evaluating it on each request?

e.g.

```
this.pixelBuffer = newPixels;
this.Pixels = new Span<TPixel>(this.pixelBuffer, 0, this.Width * this.Height);
```

vs

```
public Span<TPixel> Pixels => new Span<TPixel>(this.pixelBuffer, 0, this.Width * this.Height)
```

src/ImageSharp/Processing/Processors/Convolution/Convolution2DProcessor.cs
JimBobSquarePants: Yeah, there's a few places I'd like to visit in the codebase re `PixelAccessor`. 

e.g
https://github.com/JimBobSquarePants/ImageSharp/blob/631545a7561817fb7aff2defb4422dbff41c96d6/src/ImageSharp/Formats/Gif/GifDecoderCore.cs#L477

But I think I'll leave that till after this PR. We've touched a lot of files already and I'd like to get the public API out there first.

src/ImageSharp/Image/ImageBase{TPixel}.cs
JimBobSquarePants: I went with evaluating each time to follow existing patterns.

antonfirsov: Assigning it to an auto property would be the same as having it as a field, because autoprops have implicit backing fields. Spans have to be always created on the fly.

JimBobSquarePants: Of course, yeah. Thanks!

Issue #227, opened 2017-05-26
Inconsistent Primitives


willdean:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

ImageSharp has `Rectangle` and `RectangleF` (and `SixLabors.Shapes.Rectangle`) but no `PointF` - instead Vector2 is used in some places.

I'm new to ImageSharp, and have experimentally ported a .NET System.Drawing-based chart drawing routine to ImageSharp (which has gone sort of OK-ish)

A first impression which I feel is worth expressing, because (if it isn't already) it will soon be too late to fix is that the primitive classes feel rather inconsistent - you can draw a Rectangle but you can't draw a RectangleF, if you want a non-integer Point then you need a Vector2 not a PointF.

Could you get rid of the integer types and just standardise on floats everywhere?  Things like Rectangle often get cast to `SixLabors.Shapes.Rectangle` anyway, which is floating point, and both SixLabors and ImageSharp publish `Size` structs, one floating point and one integer.

None of this is show-stopping - I've written a few more extension methods on `Image<>` - but it does feel a bit rough, and it's perhaps the kind of thing only a newcomer would notice - after a while everyone gets numbed to this sort of stuff.

JimBobSquarePants:
Yeah you're absolutely right there. We need to organize the primitives. 

Dropping integer based ones is probably infeasible since there are places outside the library that with expect integer values and unless they are careful they can fall fowl of incorrect rounding behaviour. We should create the missing primitives and offer implicit casting throughout and make sure our API looks correct from the outside.

```
// Implicit casting works in reverse also.
Rectange -> RectangleF 
Point -> PointF -> Vector2
Size -> SizeF
```

This is a pretty easy task if someone wants to take it on?

willdean:
I'll do this if you want - like you say I don't think it's a huge task.

[...]

PR #230, opened 2017-05-29
Expand and test primitives

JimBobSquarePants:
### Prerequisites

### Description
This PR expands on the core primitives `Point`, `PointF`, `Size`, `SizeF`, `Rectangle` and `RectangleF` to add equivalent functionality from System.Drawing. Fixing #227 

Additional the following casting is possible.

- `Rectangle` -> implicit `RectangleF`, `Vector4`
- <strike>`Vector4` <-> implicit `RectangleF`</strike>
- `RectangleF` -> explicit `Rectangle`

- `Point` -> implicit `PointF`, `Vector2`
- `Point` -> explicit `Size`
- `Vector2` <-> implicit `PointF`
- `PointF` -> explicit `Point`, `SizeF`

- `Size` -> implicit `SizeF`
- `Size` -> explicit` Point`
- `SizeF` -> explicit `Size`

All explicit casting from `F` equivalents uses truncation to match existing behavior when explicitly casting from real to integer types.

I've not touched up on changing API calls since we do not use, nor require `RectangleF` in public methods. All drawing methods continue to use `Vector2[]` since the underlying **SixLabors.Shapes** types expect that type and casting to `PointF[]` creates new allocations. 

We could however use `Unsafe.As<T>` to convert them in overloads? What do you think?

### DIFF
--- a/src/ImageSharp/Numerics/RectangleF.cs
+++ b/src/ImageSharp/Numerics/RectangleF.cs
@@ -165,20 +165,6 @@ namespace ImageSharp
         }
 
         /// <summary>
-        /// Creates a <see cref="RectangleF"/> with the coordinates of the specified <see cref="Vector4"/>.
-        /// </summary>
-        /// <param name="vector">The rectangle</param>
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static implicit operator RectangleF(Vector4 vector) => new RectangleF(vector.X, vector.Y, vector.Z, vector.W);
-
-        /// <summary>
-        /// Creates a <see cref="Vector4"/> with the coordinates of the specified <see cref="RectangleF"/>.
-        /// </summary>
-        /// <param name="rectangle">The rectangle</param>
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static implicit operator Vector4(RectangleF rectangle) => new Vector4(rectangle.X, rectangle.Y, rectangle.Width, rectangle.Height);
-
-        /// <summary>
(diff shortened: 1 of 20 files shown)


src/ImageSharp/Numerics/RectangleF.cs
```diff
@@ -37,221 +33,291 @@ public struct RectangleF : IEquatable<RectangleF>
...
+        }
 
         /// <summary>
-        /// Gets the y-coordinate of the top edge of this <see cref="RectangleF"/>.
+        /// Creates a <see cref="RectangleF"/> with the coordinates of the specified <see cref="Vector4"/>.
         /// </summary>
-        public float Top => this.Y;
+        /// <param name="vector">The rectangle</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static implicit operator RectangleF(Vector4 vector) => new RectangleF(vector.X, vector.Y, vector.Z, vector.W);
```
tocsoft: Not convinced these should implicitly (or even explicitly) convert to/from `Vector4` is this conversion used anywhere? 

The main reason is its not certain what will be mapped where, should a vector represent `x`,`y`,`width`,`height` or should it represent `x`,`y`,`right`,`left` because it ambiguous I would drop it, we can add some extension methods if there's anything that needs this.

JimBobSquarePants:
Yeah happy to do that.

iamcarbon:
Have we considered using System.Drawing.Primitives (implementing: Point, PointF, Size, SizeF, Rectangle, and RectangeF) as a shared library between SixLabors & ImageSharp?

JimBobSquarePants:
Answered in Gitter also but will here also.

>I did but decided not to since I didn't have control of inlining plus I've no idea how long they will actually stick around the framework for.

PR #232, opened 2017-05-30
Cleanup Processing tests

JimBobSquarePants:
### Prerequisites

### Description
Updates the processing unit tests to use new test image providers and improve consistency.


tests/ImageSharp.Tests/Processing/Overlays/VignetteTest.cs
```diff
@@ -0,0 +1,70 @@
...
+            using (Image<TPixel> source = provider.GetImage())
+            using (var image = new Image<TPixel>(source))
+            {
+                var bounds = new Rectangle(10, 10, image.Width / 2, image.Height / 2);
+
+                image.Vignette(bounds)
+                    .DebugSave(provider, null, Extensions.Bmp);
+
+                // Draw identical shapes over the bounded and compare to ensure changes are constrained.
+                image.Fill(NamedColors<TPixel>.HotPink, bounds);
```
antonfirsov: Wouldn't it be better to DRY all these Fill+Fill+CheckSimilarity calls away into a common `EnsureProcessorChangesAreConstrained()` method somewehere?

tests/ImageSharp.Tests/FileTestBase.cs
```diff
@@ -13,6 +13,54 @@ namespace ImageSharp.Tests
...
+
+        /// <summary>
+        /// A collection of all the gif test images
+        /// </summary>
+        public static IEnumerable<string> AllGifFiles = TestImages.Gif.All;
+
+        /// <summary>
+        /// The standard pixel format enumeration
+        /// </summary>
+        public const PixelTypes StandardPixelType = PixelTypes.StandardImageClass;
```
antonfirsov: We should remove `PixelTypes.StandardImageClass` and replace all usages with `PixelTypes.Rgba32`. `StandardImageClass` was here to test the behaviour of the `Image : Image<Color>` subclass, now outdated.

tocsoft:
We should probably add some Traits to the test to define which ones are purely visual tests that need a human eye to those that have real asserts, we can then have the test runner to skip the Visual tests on the build server as no one can look at them anyway up there and should speed up our test execution, will also give us a more realistic representation of test coverage.

antonfirsov:
@tocsoft I think we should extend these visual tests later to compare the output to a reference output instead (stored in a test image + reference image repostory / NuGet package). These updated tests should be executed on CI too.

Even without proper assertions these tests still cover some stuff, at least from making exception throwing bugs.

tests/ImageSharp.Tests/FileTestBase.cs
JimBobSquarePants: Cool. Will do.

tests/ImageSharp.Tests/Processing/Overlays/VignetteTest.cs
JimBobSquarePants: Yeah, good idea will do that.

antonfirsov:
Everything looks good to me now!

JimBobSquarePants:
@antonfirsov Great! Once this builds with the updates I'm gonna merge the PR. 

We still require more changes within the tests to improve consistency and coverage but this is a healthy start. 

@tocsoft I think the tests are still very useful and provide important coverage. I'd like to do comparison with example files in the future to ensure we don't regress at any point though.

PR #234, opened 2017-05-30
Use SixLabors.Sharp outliner code in place in ImageSharp

### LINKED ISSUE #229: ArgumentOutOfRangeExceptions when drawing lines with negative Y coordinates
### Prerequisites

- [X] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I tried to draw simple lines on an Image and am constantly getting ArgumentOutOfRangeExceptions depending on the coordinates.

The issue is clearly if Y coordinate goes to negative. With a pen size 3 I can draw line from Y coordinate 2 but not 1 or 0 because it goes out of range. Does this mean ImageSharp doesn't do any kind of clipping to anything it draws and one can't draw arbitrary lines on an image?

> ImageSharp.ImageProcessingException occurred
>   HResult=0x80131500
>   Message=An error occured when processing the image using DrawPathProcessor'1. See the inner exception for more detail.
>   Source=ImageSharp
>   StackTrace:
>    at ImageSharp.Processing.ImageProcessor'1.Apply(ImageBase'1 source, Rectangle sourceRectangle)
>    at ImageSharp.Image'1.ApplyProcessor(IImageProcessor'1 processor, Rectangle rectangle)
>    at ImageSharp.ImageExtensions.Draw[TPixel](Image'1 source, IPen'1 pen, IPath path)
>    at ImageSharp.ImageExtensions.DrawLines[TPixel](Image'1 source, IPen\`1 pen, Vector2[] points)
>    at ConsoleApp1.Program.Main(String[] args) in C:\temp\ConsoleApp1\ConsoleApp1\Program.cs:line 15
> 
> Inner Exception 1:
> AggregateException: One or more errors occurred.
> 
> Inner Exception 2:
> ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
> 

Inner exceptions are "Specified argument was out of the range of valid values.\r\nParameter name: start" 
> at ImageSharp.Drawing.Processors.DrawPathProcessor'1.<>c__DisplayClass12_0.<OnApply>b__0(Int32 y)
>   at System.Threading.Tasks.Parallel.<>c__DisplayClass17_0'1.<ForWorker>b__1()
>   at System.Threading.Tasks.Task.InnerInvoke()\r\n   at System.Threading.Tasks.Task.InnerInvokeWithArg(Task childTask)
>   at System.Threading.Tasks.Task.<>c__DisplayClass176_0.<ExecuteSelfReplicating>b__0(Object )"

### Steps to Reproduce

```
using ImageSharp;
using ImageSharp.Drawing.Pens;
using ImageSharp.PixelFormats;
using System.Numerics;

namespace ConsoleApp1
{
	class Program
	{
		{
			var image = new Image<Argb32>(256, 256);
			var pen = new Pen<Argb32>(new Argb32(31, 31, 31), 4.5f);

			image.DrawLines(pen, new Vector2[] { new Vector2(10, 10), new Vector2(100, 100) }); // Works
			image.DrawLines(pen, new Vector2[] { new Vector2(-10, -10), new Vector2(100, 250) }); // Exceptions
		}
	}
}

```

### System Configuration

Using latest ImageSharp from MyGet (ImageSharp 1.0.0-alpha9-00084, ImageSharp.Drawing 1.0.0-alpha9-00079). Windows 10 v1703 (15063.296), .NET Framework 4.6.2

tocsoft:
### Prerequisites

### Description

Removes ImageSharps code around Pens/Strokes/Outlining in favour of the much more powerful outliner capabilities from SixLabors.Fonts. This now creates a new `IPath` from the outline of another `IPath` thus allowing us to consolidate the rasterization code so a single more efficient processor for both Fill and Draw.

Fixes #229 by removing the buggy code entirely.


src/ImageSharp.Drawing/Pens/Pen{TPixel}.cs
```diff
@@ -26,16 +27,16 @@ namespace ImageSharp.Drawing.Pens
     public class Pen<TPixel> : IPen<TPixel>
         where TPixel : struct, IPixel<TPixel>
     {
-        private static readonly float[] EmptyPattern = new float[0];
-        private readonly float[] pattern;
+        private static readonly ReadOnlySpan<float> EmptyPattern = new float[0];
+        private readonly ReadOnlySpan<float> pattern;
```
antonfirsov: Span-:policeman: says:
[`Span<T>` is stack only!](https://github.com/dotnet/corefxlab/blob/master/docs/specs/span.md#spant-will-be-stack-only) Should never be used as member in classes!

tocsoft: Doh... Your right of course, I'll fix it in the morning.

antonfirsov: The situation raises an API question:

- You can take a `ReadOnlySpan<T>` and copy its data into a `Buffer<T>` inside.
- But in an other overload you can take a `float[]` without copying anything (faster).

I wonder if it's OK to implement a heterogenous API, or should we just ignore span for now on places like this.

FYI What you really need here is called [`ReadOnlyBuffer<T>`](https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs) in the future API-s. (Kind-of `ReadOnlySpan<T>` that be used as member.)

JimBobSquarePants: Is there an eta on `ReadOnlyBuffer<T>`?

antonfirsov: Don't think so. `System.Buffers.Primitives` looks very experimental to me.

The `S.B.Buffer<T>` and `S.B.ReadOnlyBuffer<T>` structs are really simple cousins of `Span<T>`, but the other stuff happening here ... really strange. Still haven't figured out how exactly they do refcounting. (It's ways more simple in C++!)

Maybe I'm gonna try to introduce some of them as an ImageSharp internal as I did with `BufferSpan<T>`, but I'm not sure if their design fits our needs. Have to go deeper.

JimBobSquarePants: OK... I just want to be careful about anything we add to a public API.

tocsoft: figured I would just use the array version for now and worry about other overloads later, once real buffers land

JimBobSquarePants: Cool. I'll have a good dig through asap.

tocsoft:
merging this now.

PR #242, opened 2017-06-05
Refactor pixel conversion methods

antonfirsov:
### Prerequisites

### Description
- Introduced new pixel types: `Rgb24`, `Bgr24`, `Bgra32`
- `IPixel` methods now pack to/from structs of these types instead of `Span<byte>`
- Renamed all methods containing `Xyz`, `Xyzw` etc. component order codes to refer to these structs
- Generate `PixelOperations` boilerplate code with T4 to follow DRY principle
- Refactor codecs, quantizers etc. to use the new API-s

**Unrelated chore**
- Additional minor refactors in codecs (eg. utilizing `.GetRowSpan()`)
- Eliminate xUnit analyzer warnings in tests


src/ImageSharp/PixelFormats/PixelOperations{TPixel}.Generated.cs
```diff
@@ -0,0 +1,302 @@
...
+            this.ToBgra32(sourceColors, destBytes.NonPortableCast<byte, Bgra32>(), count);
+        }
+		
+		/// <summary>
+        /// Converts 'count' elements in 'source` span of <see cref="Rgb24"/> data to a span of <typeparamref name="TPixel"/>-s.
+        /// </summary>
+        /// <param name="source">The source <see cref="Span{T}"/> of <see cref="Rgb24"/> data.</param>
+        /// <param name="destPixels">The <see cref="Span{T}"/> to the destination pixels.</param>
+        /// <param name="count">The number of pixels to convert.</param>
+        internal virtual void PackFromRgb24(Span<Rgb24> source, Span<TPixel> destPixels, int count)
```
tocsoft: we don't need `count` here `source` is a `Span` of known length, we should just push all source values into the dest.... if you want to copy less then the actually source data you should call `source.Slice(0, amountToCopy)` and pass that in, but I suspect that all places call this currently called the source's `Span` is already the correct length.

JimBobSquarePants: It does certainly seem that way. Every reference I checked the span length is sufficient. 

This is a fantastic piece of work btw @antonfirsov The API is so much better now and all tests are passing. Other than this it's got the full seal of approval. 

Do you think the next move after this would be to pass the `Vector4` instances by ref? 4 floats must be expensive copying around. Actually surprised `System.Numerics` doesn't use ref more.

antonfirsov: It's a bit more complicated because of the `Span<byte>` overloads (`PackTo/PackFrom****Bytes`), where `Span<byte>` could be both source and destination. Because of them, the simplest solution was to specify the pixel count explicitly everywhere.

I think I can manage it without `count`, but need to have a deeper look at the call sites.

JimBobSquarePants: If you think it's best to keep count then I support your judgement.

tocsoft: That's fine then, keeping the suite of APIs consistent with each other is a perfectly agreeable reason 👍

antonfirsov: Alright, let's keep the current version for now, maybe someone will find a better option later as the code evolves.

antonfirsov:
Pushed another change that T4-s most of the `Rgba32.PixelOperations`. This way it will be easier to apply any refactors (eg signature changes) in the future.  It's enough to modify 6 method templates now instead of 22 methods with docs.

@JimBobSquarePants RE `ref Vector4`: [It doesn't look like it's worth it](https://github.com/JimBobSquarePants/ImageSharp/blob/pixel-conversion-refactor/tests/ImageSharp.Benchmarks/General/PixelConversion_ConvertFromVector4.cs#L144).

JimBobSquarePants:
Ah good man. Yeah, not worth the effort. This all looks great then!

JimBobSquarePants:
Imma gonna merge this. 👍

PR #243, opened 2017-06-05
Repair Benchmarks for VSCode/Linux

Toxantron:
### Prerequisites

### Description

After some inactivity in the project I wanted to continue contributing and noticed that since moving to VS2017 the Benchmarks no longer worked under Linux with .Net Core and VSCode.

* I added the target framework _netcoreapp1.1_
* I split the dependencies and restored the references from _project.json_
* I upgraded BenchmarkDotNet since **0.10.2** does not work with _csproj_ and _netcoreapp1.1_
* After the upgrade I replaced the deprecated attributes and performed a fex checks

**TODO:** Someone with Windows (and VS2017) should confirm this still works and runs properly on _net461_

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -16,7 +16,7 @@
     {
       "taskName": "build benchmark",
       "suppressTaskName": true,
-      "args": [ "build", "tests/ImageSharp.Benchmarks/project.json", "-f", "netcoreapp1.1", "-c", "Release" ],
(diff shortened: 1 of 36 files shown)


antonfirsov:
Thanks for fixing this! I really missed the multi-framework benchmarking feature.

It works for me on Windows with both `core` and `clr 4.6.1`!

JimBobSquarePants:
Excellent! That makes life much easier! 💯

PR #249, opened 2017-06-18
Read and write duplicate ICC tag data entries only once

JBildstein:
### Prerequisites

### Description
An ICC profile can reuse the same data for multiple tag data entries. This pull request changes the reader to only read each entry once and improves how the writer finds and writes those duplicate entries.
For reference: https://github.com/JimBobSquarePants/ImageSharp/issues/217#issuecomment-301795153


JimBobSquarePants:
Excellent! Thanks for the update. 👍

PR #250, opened 2017-06-18
Additional pixel blenders + refactor generators

tocsoft:
### Prerequisites

### Description

I've added some missing pixel blenders/composition models. The added compositors unblock us being able to draw transparent regions onto existing pixels (this enabled me to create the `AvatarWithRoundedCorner` sample project.)

Refactored the default versions of  `PixelBlender<TPixel>` so they are all generated via a template.

Included a new `samples` folder where we can start padding out with small sample applications.


samples/AvatarWithRoundedCorner/Program.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace AvatarWithRoundedCorner
+{
+    using System;
+    using System.Numerics;
+    using ImageSharp;
+    using SixLabors.Shapes;
+
+    class Program
+    {
+        static void Main(string[] args)
```
antonfirsov: Isn't it better to have a single samples project, with different code examples residing in different classes? (Launching them with some kind of sample switcher in `Main()`)

JimBobSquarePants: Yeah, that makes sense to me.

src/ImageSharp/PixelFormats/PixelBlenders/DefaultPixelBlenders.Generated.tt
```diff
@@ -0,0 +1,118 @@
+﻿<#
```
JimBobSquarePants: Sucking up to @antonfirsov  now! 😛 

Good to see!

samples/AvatarWithRoundedCorner/Program.cs
tocsoft: I disagree (obviously 😉 ) My logic for having lot of individual projects is for the fact they are supposed to be self contained demonstrations on how a user could implement such a thing... if they are all bundled into a larger single project then 1 its harder to see what the dependencies are, and 2 it becomes tempting to share code across the samples which, in my opinion, shouldn't happen so as to make it simpler to follow (preferably without even having to download from github) but If people feel strongly enough I'll to change it.

JimBobSquarePants: The great thing about source control is we can always change it in the future. It's not a blocker for this PR as far as I'm concerned.

src/ImageSharp/PixelFormats/PixelBlenders/DefaultPixelBlenders.Generated.tt
antonfirsov: Would be much easier to refactor/optimize them!

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.PixelBenders.cs
```diff
@@ -22,25 +22,30 @@ internal virtual PixelBlender<TPixel> GetPixelBlender(PixelBlenderMode mode)
...
-                    return DefaultHardLightPixelBlender<TPixel>.Instance;
+                case PixelBlenderMode.Multiply: return DefaultPixelBlenders<TPixel>.Multiply.Instance;
+                case PixelBlenderMode.Add: return DefaultPixelBlenders<TPixel>.Add.Instance;
+                case PixelBlenderMode.Substract: return DefaultPixelBlenders<TPixel>.Substract.Instance;
+                case PixelBlenderMode.Screen: return DefaultPixelBlenders<TPixel>.Screen.Instance;
+                case PixelBlenderMode.Darken: return DefaultPixelBlenders<TPixel>.Darken.Instance;
+                case PixelBlenderMode.Lighten: return DefaultPixelBlenders<TPixel>.Lighten.Instance;
+                case PixelBlenderMode.Overlay: return DefaultPixelBlenders<TPixel>.Overlay.Instance;
+                case PixelBlenderMode.HardLight: return DefaultPixelBlenders<TPixel>.HardLight.Instance;
+                case PixelBlenderMode.Src: return DefaultPixelBlenders<TPixel>.Src.Instance;
```
antonfirsov: Is this now a closed set of well known and commonly used pixel blenders?

JimBobSquarePants: It should be. I think we're implementing all the Porter Duff equations.

tocsoft: Yeah.. I believe this covers all the standard composition/blending types. We can worry about making it pluggable later.

tocsoft:
merging this... we can fight out the samples structure later.

JimBobSquarePants:
Good man. Meant to approve it before.

PR #254, opened 2017-06-24
Refactor encoder/decoder image format APIs

tocsoft:
### Prerequisites

### Description
Refactors the the way encoders and decoders are registered into the system.

This removed `IDecoderOptions`/`IEncoderOptions` concept, setting configurations for encoding/decoding is done directly on the actual encoders/decoders.

You can register encoders/decoders per image mime-type, you can register file extension-> mime-type mapping. You can also replace the registered encoder for a specific mime-type. 
``` c#
// lets switch out the default encoder for jpeg to one 
// that saves at 90 quality and ignores the matadata
Configuration.Default.SetMimeTypeEncoder("image/jpeg", new ImageSharp.Formats.JpegEncoder()
{
    Quality = 90,
    IgnoreMetadata = true
});
```
Also introduced a new type `IMimeTypeDetector` this is the interface thats responsible for detecting the mime-type of an image file based on detecting characteristics in the file header.

Changed `IImageFormat` to be `IImageFormatProvider` that instead of exposing encoders/decoders directly instead has a callback that allows it to register one or more image encoders, decoders, mimetype detectors and ext->mime-type mapings. 

Renamed quality to PaletteSize on png & gif for a more accurate name. Also removed quality from metadata.


src/ImageSharp/Configuration.cs
```diff
@@ -81,19 +115,53 @@ public Configuration(params IImageFormat[] providers)
...
+        /// </summary>
+        /// <param name="formatProvider">The format providers to call configure on.</param>
+        public void AddImageFormat(IImageFormatProvider formatProvider)
+        {
+            Guard.NotNull(formatProvider, nameof(formatProvider));
+            formatProvider.Configure(this);
+        }
+
+        /// <inheritdoc />
+        public void SetMimeTypeEncoder(string mimeType, IImageEncoder encoder)
```
tocsoft: I disagree it should be fully public and discoverable otherwise you can't do :
```c#
Configuration.Default.SetMimeTypeEncoder("image/jpeg", new JpegEncoder()
{
    Quality = 90,
    IgnoreMetadata = true
});
```

dlemstra: Fair point. Better keep it public then.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -363,8 +380,6 @@ private unsafe void ReadFrameColors(byte[] indices, byte[] colorTable, int color
 
             if (this.previousFrame == null)
             {
-                this.metaData.Quality = colorTableLength / 3;
```
tocsoft: Its not used anywhere (as I removed the usage of the metadata quality from the encoders).

The reason i removed using it tn the encoders was that it causes issues when switching formats, if you decoded a jpeg its 'quality' could be 70, but then encoding a gif would have caused a palette size of 70 which is drastically different. So now you have to be explicit but manually setting it on the encoders.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -352,6 +362,13 @@ private void ReadFrameIndices(GifImageDescriptor imageDescriptor, byte[] indices
         /// <param name="descriptor">The <see cref="GifImageDescriptor"/></param>
         private unsafe void ReadFrameColors(byte[] indices, byte[] colorTable, int colorTableLength, GifImageDescriptor descriptor)
         {
+            if (this.IgnoreFrames && this.image != null)
```
tocsoft: yeah I'll drop this change out... and add a ticket

src/ImageSharp/Formats/Bmp/BmpImageFormatProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <summary>
+    /// Detects gif file headers
+    /// </summary>
+    public class BmpImageFormatProvider : IImageFormatProvider
+    {
+        /// <inheritdoc/>
+        public void Configure(IImageFormatHost host)
+        {
+            var encoder = new BmpEncoder();
+            foreach (string mimeType in BmpConstants.MimeTypes)
```
tocsoft: I've decided to leave them for a time being, doesn't make a huge difference either way and at least the current way its obvious where the static data relating to a format is.

src/ImageSharp/Formats/Gif/GifConstants.cs
```diff
@@ -90,6 +91,16 @@ internal static class GifConstants
         /// <summary>
         /// Gets the default encoding to use when reading comments.
         /// </summary>
-        public static Encoding DefaultEncoding { get; } = Encoding.GetEncoding("ASCII");
+        public static readonly Encoding DefaultEncoding = Encoding.GetEncoding("ASCII");
```
tocsoft: Left it separate for now, its safer in case we want to switch out the encoder to be different based on format. I favour having the formats as isolated from each other as we can. Less chance on causing cross encoder bugs during fixes then.

tocsoft:
8bd785a switches things up somewhat and reintroduces `IImageFormat`s but changes how you deal with them, ImageFormats and decoders etc are still divorced form each other and only linked together at the Configuration level. Also renamed the `IImageFormatProvider`s to `IConfigurationModule` to better relay what they are for. And not lockout other config level tweeks people might want to inject, eg FileSytem (if we ever make the API public).

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
JimBobSquarePants: 👍 Quality should be a property of the encoder. Having it global just confuses everybody.

src/ImageSharp/Configuration.cs
```diff
@@ -81,126 +115,147 @@ public Configuration(params IImageFormat[] providers)
...
+            this.mimeTypeDecoders.AddOrUpdate(imageFormat, decoder, (s, e) => decoder);
         }
 
         /// <summary>
-        /// Adds image format. The class is locked to make it thread safe.
+        /// Removes all the registered mime type detectors.
         /// </summary>
-        /// <param name="format">The image format.</param>
-        private void AddImageFormatLocked(IImageFormat format)
+        public void ClearMimeTypeDetectors()
```
antonfirsov: should be `ClearImageFormatDetectors()`

src/ImageSharp/Configuration.cs
```diff
@@ -44,12 +60,15 @@ public Configuration()
         /// <summary>
         /// Initializes a new instance of the <see cref="Configuration" /> class.
         /// </summary>
-        /// <param name="providers">The inital set of image formats.</param>
-        public Configuration(params IImageFormat[] providers)
+        /// <param name="providers">A collection of providers to configure</param>
```
antonfirsov: The argument name should be refactored as well.

JimBobSquarePants:
shia-labeouf-applause

# Bravissimo!!

I absolutely love this, from the static format detection to the way you've split out the encoders and decoders. Really, really nice! I think we have our format API. 👍

tocsoft:
@dlemstra have you had a chance to cast you eye over this yet, just waiting on your final pass to approve.

Issue #245, opened 2017-06-07
Pixeldata not correct when reading certain JPEG


devedse:
### Description
I'm writing a small tool that compares images after doing lossless optimizations on them using another tool: https://encode.ru/threads/1589-FileOptimizer

I took a picture with my own Camera (6000x4000) and ran it through the FileOptimizer. After that I used the tool I created to compare the source and target image to see if there were any pixel differences. The comparison failed on pixel X: 5990 Y: 3992

For image1 the debugger shows: 36, 15, 10, 255 (RGBA)
For image2 the debugger shows: 36, 15, 12, 255 (RGBA)

The strange this is, if I open the same images in Paint.NET the pixel value is: 38, 14, 12, 255 (RGBA). (And it's the same for both images).

![image](https://user-images.githubusercontent.com/2350015/26861652-39e3880e-4b13-11e7-802d-dd30c64ffddd.png)

I also tested what the pixel value would be of the images saved as BMP. In Paint.NET they read again as the same values we saw before:
1. Image1: 36, 15, 10, 255 (RGBA)
2. Image2: 36, 15, 12, 255 (RGBA)

### Steps to Reproduce

Images that should be equal:
[SourceImagesThatShouldBeEqual.zip](https://github.com/JimBobSquarePants/ImageSharp/files/1056844/SourceImagesThatShouldBeEqual.zip)

Code I'm using to reproduce the bug:
```
private static bool AreImagesEqual(string image1Path, string image2Path)
{
    var w = Stopwatch.StartNew();

    var image1 = Image.Load(image1Path);
    var image2 = Image.Load(image2Path);


    if (image1.Width != image2.Width || image1.Height != image2.Height)
    {
        return false;
    }

    int width = image1.Width;
    int height = image1.Height;

    using (var fs = new FileStream("image1.bmp", FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite))
    {
        image1.SaveAsBmp(fs);
    }

    using (var fs = new FileStream("image2.bmp", FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite))
    {
        image2.SaveAsBmp(fs);
    }

    for (int y = 0; y < height; y++)
    {
        for (int x = 0; x < width; x++)
        {
            var pixel1 = image1.Pixels[y * width + x];
            var pixel2 = image2.Pixels[y * width + x];

            if (pixel1 != pixel2)
            {
[...]

Issue #224, opened 2017-05-25
jpg.decodedBlock should it release memory after usage?


mhamri:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I was using imagesharp 1 alpha 4 48 previousely and my app for a normal task took up to 4gb, today I updated to latest version and saw it decreased to 3gb. thanks for the improvement. 

I just ran a profiler and saw that the type of decoded block and color block from **static** type changed to normal property so I got some concern if i'm doing something wrong. I assume that it should release the resources when I call the dispose method or it reach end of the using statement.

I checked almost all of my code, i have using and dispose for every usage. also the class that wrapp the imagesharp is implementing IDisposable.

![2017-05-24 18_49_26-pbww imageserver running - microsoft visual studio](https://cloud.githubusercontent.com/assets/8991783/26434181/c2b080b2-4139-11e7-9e4f-6c24f0c32145.png)

### Steps to Reproduce
i'm using imagesharp with DI in startup.cs, configServices method: 

`services.AddTransient<ImageSharpHelper>();`

and the `imageSharpHelper` is like bellow:

```
using ImageSharp;
using ImageSharp.Formats;
using Microsoft.Extensions.Logging;
using System;
using System.Collections.Generic;
using System.IO;

namespace namespace1.namespace2.Image
{
    public class ImageSharpHelper : IDisposable
    {
        private readonly ILogger<ImageSharpHelper> _logger;

        public ImageSharpHelper(ILogger<ImageSharpHelper> logger)
        {
            _logger = logger;
        }

        public Image<Rgba32> Client { get; set; }
        public int? CompressRate { get; set; }
        public string Mime { get; set; }
        public ImageOutPutTypes OutPutTypes { get; set; }
        public int? QualityRate { get; set; }

        private Dictionary<ImageOutPutTypes, Func<MemoryStream, string>> SaveImageMapper => new Dictionary<ImageOutPutTypes, Func<MemoryStream, string>>()
        {
            [ImageOutPutTypes.Bmp] = SaveAsBmp,
[...]

Issue #214, opened 2017-05-13
Decoding JPEG puts the process to infinite loop if there is an error on JPEG bytes


xeora:
The problem appears when you want to resize a JPEG with crop options. 

Problematic code piece is the following;
https://github.com/JimBobSquarePants/ImageSharp/blob/master/src/ImageSharp/Formats/Jpeg/Components/Decoder/InputProcessor.cs#L156

I made a hotfix for myself by changing the code like

`while (length > 0 && errorCode == DecoderErrorCode.NoError)`

But surely, I didn't test anything else if it breaks something more. Currently working okay in my environment. 

It requires deep investigation if it is the correct fix or not... 

thanks...

JimBobSquarePants:
Hi @xeora 

Could you please resubmit this with version numbers etc and the offending image? We've nothing to go on since you deleted the template for some reason.

Cheers

James

tocsoft:
Also could you provide a code example when you provide version details as crop/resize functions have no impact the decoder they are entirely separate systems.

xeora:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When you are loading a jpg file with some broken bytes in it, process gets in to infinite loop.

### Steps to Reproduce
Try to load the jpg with broken bytes or falsy EOF. 
ex: `ImageSharp.Image.Load(InputImage)`

### System Configuration
Windows 10 IIS server but it is happening also linux mono or windows 2012 srd srv with .net 4.5.2 installed. But it is not .net or OS related problem..

- ImageSharp version: Current Git hub version.
- Other ImageSharp packages and versions: - 
- Environment (Operating system, version and so on): Windows 10 IIS
- .NET Framework version: .net 4.5.2
- Additional information: Sorry for my first entry, I already customize your code a lot to work faster. That's why, I just jump in to the point where problem is exists. However, I didn't modified the related part and when i test it with the current version on github, problem is still there. I said while croping / resizing because my code is joined the load and the resize together. I thought it was encoding problem but when i check deeper, it is decoding problem. 

Thanks...

[...]

Issue #159, opened 2017-03-29
Load Image Exception Jpg ImageSharp.Formats.Jpg.MissingFF00Exception


vinhhrv:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Load Image Error

![jpg_bug](https://cloud.githubusercontent.com/assets/769541/24437694/c78c32ec-146c-11e7-9712-de85d73bd467.jpg)

ImageSharp.Formats.Jpg.MissingFF00Exception

### Steps to Reproduce

`using (var file = File.OpenRead("files/jpg_bug.jpg"))
 { 
      var image = Image.Load(file);
}`

Attach file

### System Configuration

- ImageSharp version: 1.0.0-alpha5-00046
- Other ImageSharp packages and versions:
- Environment (Operating system, version and so on): win10
- .NET Framework version: netcoreapp
- Additional information:

HelmuthWcs:
I'm getting the same error on some pictures taken with the samsung gear 360. Failed with alpha2 so I updated, and still failing with latest alpha5
![4](https://cloud.githubusercontent.com/assets/2522424/24452199/608beeb4-1482-11e7-96d7-c07b14925182.jpg)

JimBobSquarePants:
@vinhhrv @HelmuthWcs 

The causes that throw that exception are different in both images. 

The first image is actually correct and highlight a bug in the decoder.

The seconds image however is buggy and will require a workaround. Here's the output from JpegSnoop

```
*** Decoding SCAN Data ***
  OFFSET: 0x0000D03B
  Scan Decode Mode: Full IDCT (AC + DC)

  Scan Data encountered marker   0xFFD9 @ 0x00785631.0
*** ERROR: Can't find huffman bitstring @ 0x0078562F.1, table 1, value [0xffffb200]
*** ERROR: Bad huffman code @ 0x0078562F.0
*** ERROR: Bad scan data in MCU(449,224): Chr(Cr) CSS(0,0) @ Offset 0x0078562F.1
           MCU located at pixel=(7184,3584)

```

[...]

PR #274, opened 2017-07-10
WIP Replace Jpeg Decoder

JimBobSquarePants:
### Prerequisites

### Description

This is a WIP pull request to allow us to work collaboratively on a replacement for the current jpeg decoder.  

While baseline encoding is fairly straightforward, decoding is much more difficult due to the required support for both baseline and progressive jpegs + correct handling of images encoded using buggy encoders. 

Our current decoder can handle some cases but struggles with others. It's fairly fast but the code is extremely difficult to navigate due to the nature of the original [Golang](https://github.com/golang/go/tree/123fd4640ae27e7f179f84cd7efece07ab954b45/src/image/jpeg) implementation. It also produces different output to other implementations with an error in the +-3 range per pixel component. There are hints of a memory leak there also. #151 #224 

The new jpeg decoder is based upon a jpeg implementation found in Mozilla's [PDF.js](https://github.com/mozilla/pdf.js/blob/85161fa6dc81c3df921b27a378fbaf82491aa9ee/src/core/jpg.js) with several key enhancements. 

1. Fast table lookups for colorspace transforms
2. Better handling of broken images
3. Faster Huffman and IDCT implementations

We can now decode the images in #159 correctly throw in #214 and produce much closer output to the reference implementations in #155 and #245 

However, the new decoder is 2X slower that our original and will require optimization. To do that we need to do the following:

1. Modularize the code more to allow improved testing
2. Perform unit and load testing on the decoder components
3. Optimize the decoder based on test findings. 

Decoding a jpeg has two areas of high complexity which require extreme optimization: Huffman decoding and Inverse Discrete Cosine Transform; we can make large gains in performance in focusing our efforts there. 

The Huffman decoder requires an accurate fast lookup table implementation See [here](https://github.com/mattdesl/slim/blob/8c2b061c3c1cfb3bd6490a1eb0a9713e94ed0d14/slim/src/slim/texture/io/Huffman.java#L88) and [here](https://github.com/mattdesl/slim/blob/8c2b061c3c1cfb3bd6490a1eb0a9713e94ed0d14/slim/src/slim/texture/io/JPEGDecoder.java#L554) for a reference implementation in Java and  IDCT can be sped up using SIMD. (I would like to see if we can do it using `Vector<short>` rather than the `Block8x8F` implementation we have in the current decoder to keep memory usage down. Perhaps @mellinoe can give us some hints there?)

Please have a good dig through the code. If you have any questions please ask. I'd like us to get a much better understanding of the format than we currently have.

### Update by @antonfirsov
I'm on a bit different and more critical opinion towards this change, and I also have a plan to speed up the decoder while keeping it correct and memory friendly. [I'm quite sure there is no easy way for this, and I really appreciate any help you can provide with the tasks I managed to identify.](https://github.com/JimBobSquarePants/ImageSharp/pull/274#issuecomment-314193953)


src/ImageSharp/Common/Extensions/ListExtensions.cs
```diff
@@ -0,0 +1,48 @@
+﻿// <copyright file="ListExtensions.cs" company="James Jackson-South">
```
JimBobSquarePants: I don't think I'm using any of this anymore. It can go.

src/ImageSharp/Common/Extensions/StreamExtensions.cs
```diff
@@ -27,7 +27,7 @@ public static void Skip(this Stream stream, int count)
 
             if (stream.CanSeek)
             {
-                stream.Position += count;
+                stream.Seek(count, SeekOrigin.Current); // Position += count;
```
JimBobSquarePants: This is supposedly faster

tests/ImageSharp.Benchmarks/Image/DecodeJpeg.cs
```diff
@@ -23,21 +24,21 @@ public void ReadImages()
         {
             if (this.jpegBytes == null)
             {
-                this.jpegBytes = File.ReadAllBytes("../ImageSharp.Tests/TestImages/Formats/Jpg/Baseline/Calliphora.jpg");
+                this.jpegBytes = File.ReadAllBytes("../../../../../../../../ImageSharp.Tests/TestImages/Formats/Jpg/Baseline/Calliphora.jpg");
```
JimBobSquarePants: This needs fixing properly

tests/ImageSharp.Tests/TestFile.cs
```diff
@@ -140,7 +138,15 @@ public Image<Rgba32> CreateImage()
         /// </returns>
         public Image<Rgba32> CreateImage(IImageDecoder decoder)
         {
-            return Image.Load(this.image.Configuration, this.Bytes, decoder);
+            return Image.Load(this.GetImage().Configuration, this.Bytes, decoder);
+        }
+
+        private Image<Rgba32> GetImage()
```
JimBobSquarePants: This means that the test files are now lazy loaded. No more mental CPU after every build!

tests/ImageSharp.Benchmarks/Image/DecodeJpeg.cs
antonfirsov: I have plans to fix stuff like this. The test image repository provider code should be shared between tests and benchmarks.

antonfirsov:
### Modularization 

As far as I understand the Jpeg decoding process could be broken down to the following well defined main steps:

Stream processing logic:

1.1.  Building up the huffman tables
*Input*: The stream
*Output*: the `HuffmanTables` data structure.

1.2. Decoding the content of the stream (scans) into 8x8 blocks based on the tables
*Input*: The stream + `HuffmanTables`
*Output*: Decoded blocks.

Transformation logic:

2. Apply IDCT on the blocks
*Input*: A DCT encoded block
*Output*: A block of color values

3. Transform YCbCr/Ycck/Cmyk etc. blocks into `TPixel` data
*Input*: 1-3 blocks in jpeg color spaces
*Output*: A span of `TPixel` colors

The granularity for steps 1.1-3. should be one MCU/Block, so we can properly test + reorganize the code according to steps described at the end of my comment.

I think we should find a way to independently unit test these steps. For steps 2-3 this is relatively easy, I suggest to follow the "Reference implementation" approach I used with `Block8x8F` tests. In order to make step 3. testable we should clean up the `LinearizeBlockData()` logic. It's really hard to follow what's going on here. We should test the color transformation with reference implementations at block level.

Modularizing + testing Step 1. is a bit harder though. As a first step, we should try to encapsulate the stream processing logic into a separate `InputProcessor` class decoupling it's logic from `ScanDecoder`.

It's worth to define a separate structure for integer blocks (`Block8x8Int`), it's more type-safe and makes it easier to replace it with `Block8x8F`.

As far as I know integer type `Vector<T>` lacks SIMD support for operations as basic as multiplication. Just run [this benchmark](https://github.com/JimBobSquarePants/ImageSharp/blob/master/tests/ImageSharp.Benchmarks/General/Vectorization/MulUInt32.cs) if you don't believe me :P
I wonder if @mellinoe or someone else has an idea regarding `Vector<short>`, but I'm 90% sure, if we want a fast decoder we have no other options than:
1. Implement this modularization
2. Refactor the decoder to process blocks incrementally instead of processing them into large output pixel areas (-> saving memory)
3. Migrate to floats. (-> speed up!)

I know it's a hard job, but  I would be really happy to get help on this.

src/ImageSharp/Formats/Jpeg/Port/Components/JpegPixelArea.cs
```diff
@@ -0,0 +1,147 @@
...
+            this.componentData = new Buffer<byte>(width * height * numberOfComponents);
+            Span<byte> componentDataSpan = this.componentData;
+            const uint Mask3Lsb = 0xFFFFFFF8; // Used to clear the 3 LSBs
+
+            using (var xScaleBlockOffset = new Buffer<int>(width))
+            {
+                Span<int> xScaleBlockOffsetSpan = xScaleBlockOffset;
+                for (int i = 0; i < numberOfComponents; i++)
+                {
+                    ref Component component = ref components.Components[i];
```
antonfirsov: The inner part of the loop could be an instance method of `Component`.

antonfirsov:
## Unit Testing

### The concept

image

[A sample unit test following this concept.](https://github.com/JimBobSquarePants/ImageSharp/blob/master/tests/ImageSharp.Tests/Formats/Jpg/Block8x8FTests.cs#L416) (Ok, it lacks the corner cases :stuck_out_tongue:)

The reference implementation lives in the Test project, does not get changed, and does not depend on the tested code, so if we know that the reference implementation is correct at the point when the test was implemented, we can assume it keeps it's correctness, so we can use it to cross-validate the actual implementation any time in the future.
With this approach we do not have to capture meaningless blobs of data, copying it to the code. The reference implementation produces validation data in a way that actually *makes sense*. 

I prefer to implement these tests before doing any changes to the code. Well worth the efforts!

*Ps.:*
I haven't read this concept in books, it was just a straightforward thing that came into my mind when I was about to optimize DCT & IDCT. So I did a quick sanity check now, by googling ["unit testing FFT"](https://www.google.hu/search?q=unit+testing+fft&oq=unit+testing+fft&aqs=chrome..69i57.2920j0j7&sourceid=chrome&ie=UTF-8) which is a similar, but widely used mathematical method. The results show that I was not the only one to came to this conclusion, looks like the authors of a [popular FFT library](http://www.fftw.org/) used a similar (but more sophisticated) approach.

### Example: Testing the YCbCr -> TPixel colorspace conversion
- **Input:** Random-generated `Y`, `Cb` and `Cr` blocks.
- **Output:** `Span<TPixel>` of 64 color values.
- **Reference impl.:** The simplest possible YCbCr->TPixel conversion code that operates on 3 input blocks and outputs `Span<TPixel>`. No optimizations, no LUT-s!
- **Ąctual impl.:** The related method in the decoder. *

 **Ooops ... looks like we don't have that method at the moment ... that's why I call this a modularization refactor :stuck_out_tongue_winking_eye: Good testing improves the code, and our understanding of the code as well! :guitar: We really need that logic to be understood and separated to enable future refactors.*

src/ImageSharp/Formats/Jpeg/Port/Components/FrameComponent.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        /// Gets or sets the identifier
+        /// </summary>
+        public byte QuantizationIdentifier;
+
+        /// <summary>
+        /// Gets or sets the block data
+        /// </summary>
+        public Buffer<short> BlockData;
```
antonfirsov: It's really worth to introduce `Block8x8Int` and refactor this into `Buffer<Block8x8Int>`.

A quick candidate implementation utilizing [Fixed-size buffers](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/unsafe-code-pointers/fixed-size-buffers) and having an `.AsSpan()` method:
```C#
internal unsafe struct Block8x8Int
{
    private fixed short data[64];

    public Span<short> AsSpan() => new Span<short>(this.data, 64);

    public ref short this[int idx] => ref Unsafe.Add(ref Unsafe.As<Block8x8Int, short>(ref this), idx);
}

```

antonfirsov: Or .. maybe it's better to allocate it as a private member `Buffer<short> blockData` to be `ArrayPool`-friendly, but we should expose it only through a `public Span<Block8x8Int> BlockData { get; }` property utilizing `NonPortableCast`.

JimBobSquarePants: You wanna give that a pop as one of the initial refactorings?

mellinoe:
> As far as I know integer type Vector<T> lacks SIMD support for operations as basic as multiplication. Just run this benchmark if you don't believe me :P
I wonder if @mellinoe or someone else has an idea regarding Vector<short>, but I'm 90% sure, there is no easy win here. If we want a fast decoder we have no other options than following this plan:

Actually, we should support multiply on `Vector<short>`. The unfortunate part of the API currently is that there isn't a good way to figure out which primitive types support which individual operations. If you look at [this inscrutable chart](https://github.com/dotnet/coreclr/blob/master/src/jit/simdintrinsiclist.h#L37), you can get an idea of which mappings exist for different intrinsics. [This line](https://github.com/dotnet/coreclr/blob/master/src/jit/simd.cpp#L2420-L2429) is also of interest. Note that most "missing intrinsics" are that way because the underlying hardware doesn't support it. In this particular case, I'm pretty positive that SSE does have instructions specifically for 16-bit signed integer multiplication. It should work.

I would like to help give more pointers, although I'm not sure how deep I can get with the free time I have. Is there a particular spot that's been identified as a bottleneck I could try to analyze?

JimBobSquarePants:
Thanks @mellinoe for your comments. That chart is very useful!

There's work for us to do before you could really make any changes `Buffer<short>` to `Block8x8` refactoring but by my understanding there are two places libjpeg turbo uses SIMD to speed up operations which we should also do. I've highlighted the relevant parts of our codebase.

1. DCT transforms forward and [Inverse](https://github.com/JimBobSquarePants/ImageSharp/blob/ae6d91a556eedb7ed489254ae558d89fb98d64ef/src/ImageSharp/Formats/Jpeg/Port/Components/IDCT.cs).
2. [Colorspace transforms](https://github.com/JimBobSquarePants/ImageSharp/blob/ae6d91a556eedb7ed489254ae558d89fb98d64ef/src/ImageSharp/Formats/Jpeg/Port/JpegDecoderCore.cs#L879)

I'm way out of my comfort zone here with `Vector<T>`, I just don't "get it" to be honest so any insight there would be really useful.

antonfirsov:
@mellinoe this is a really useful piece of information, thank you! I'm already researching integer DCT algorithms.

@JimBobSquarePants 1.-2. points of my plan are still valid though, and we need that coverage, modularity + cleaning up the `LinearizeBlockData()` stuff before starting any vectorization. Color transformations will be much faster with SIMD.

antonfirsov:
@JimBobSquarePants just found a standard-compliant [IDCT reference implementation](https://github.com/ernesto-jimenez/midenuncia/blob/master/call_receiver/tmndec/idctref.c) to test our code against.

antonfirsov:
Bad news ... looks like [neither integer division](https://github.com/dotnet/coreclr/blob/master/src/jit/simdintrinsiclist.h#L90), [nor bit shifting](https://github.com/dotnet/coreclr/blob/master/src/jit/simdintrinsiclist.h#L147) is supported. @mellinoe can you confirm or deny this?

I don't think it's possible to build efficient integer DCT and colorspace conversion algorithms without having at least bit shifting supported by SIMD.

src/ImageSharp/Formats/Jpeg/Port/JpegDecoderCore.cs
```diff
@@ -0,0 +1,968 @@
...
+                {
+                    multiplierSpan[i] = (short)IDCT.Descale(quantizationTable[i] * IDCT.Aanscales[i], 12);
+                }
+
+                for (int blockRow = 0; blockRow < blocksPerColumn; blockRow++)
+                {
+                    for (int blockCol = 0; blockCol < blocksPerLine; blockCol++)
+                    {
+                        int offset = GetBlockBufferOffset(ref component, blockRow, blockCol);
+                        IDCT.QuantizeAndInverseFast(ref frameComponent, offset, ref computationBufferSpan, ref multiplierSpan);
```
antonfirsov: This method is not fast at all :(
image

I'd rather separate the quantaziation logic from IDCT to make it modular and testable. See no real benefits from doing them in one step. We can optimize later.

src/ImageSharp/Formats/Jpeg/Port/Components/FrameComponent.cs
antonfirsov: Okkay, this one seems easy for me ... but I really don't want to code all the refactors and tests by my own. You have better knowledge on the new decoder code now.
We also got the integration testing task as a jpeg optimization prerequisite. I want to focus on that in the next 1-2 weeks.

src/ImageSharp/Formats/Jpeg/Port/JpegDecoderCore.cs
JimBobSquarePants: Yeah, definitely a misnomer. It's a port of the libjpeg turbo method plus libjpeg.net. I'm not sure what is causing the poor performance (I was certainly expecting better)

If you can separate that out, great! All the examples I've found do it as a single process though.

mellinoe:
@antonfirsov That is correct. We have a proposal for Shift operations here: https://github.com/dotnet/corefx/issues/16835, but we haven't committed to it yet. Knowing that you would use it is very valuable, hopefully we can move forward with it.

Integer division indeed isn't supported, but I don't believe it is supported by the hardware. An option is to convert the values to floats, and then convert back. It's likely you'd want to wrap that up into some kind of helper method. We've recently added the integer and floating-point conversion methods necessary for this, but those will only be accelerated on .NET Core 2.0.

antonfirsov:
For everyone's information:
The new decoder should be based on this port, because the stream parsing and the CPU-intensive processing logic is clearly separated here. This makes work on progressive jpeg-s easier. So good job @JimBobSquarePants, this decoder is a really-really good thing! (Despite it's messy and slow colorspace transformation logic, which has to be replaced :P)

However, I suggest to open a new clean WIP PR, removing the noise we gathered here. I could describe my implementation plan there, and track the work.

There are some possibilities for incremental work. We can release a beta with the initial slow variant of the PdfJs decoder, and improve performance later. We need to keep all the classes of the original golang port in the repository though, to keep the useful logic and tests refactored as we add changes.

@JimBobSquarePants thoughts?

JimBobSquarePants:
Hey @antonfirsov ! The colorspace stuff I thought was pretty good! 😝 

I agree, Let's close this PR and open a new one. Make sure we capture the EXIF fixes I added also though.

I think the PdfJS based decoder is good enough for beta despite the comparative decrease in speed. It can already successfully decode every single rogue image across all our jpeg issues and produces a very similar output to libjpeg. Make it work; make it work fast.

During beta we can tackle performance issues and ensure we don't regress using your new qa lab code. 

So let's do two PR's, one to switch out the decoder, then a second WIP one for performance improvements.

antonfirsov:
@JimBobSquarePants I just merged `jpeg-port` into my `jpeg-lab` branch, organizing tests+classes+namespaces in a cleaner way. 

For a while I'm planning to do refactors that keep both decoders working, so we can just merge back `jpeg-lab` into `jpeg-port` and use this (#274) PR for the switch-out process.

JimBobSquarePants:
Great! 👍

Issue #142, opened 2017-03-18
Proposal: Value Type Kernels


antonfirsov:
### Prerequisites
- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### The problem
- There are many places in the library where we operate on relatively small matrices known at compile time using a dynamically allocated heap array (wrapped in `Fast2DArray`):
  - `ConvolutionProcessor`-s 
  - `PatternBrush`-es
  - `ResizeProcessor` weight windows
- This approach bings in implicit overhead, having a significant negative effect on performance:
  - Indexing heap arrays increases the probabilty of cache misses
  - Calculations are iterating using nested for loops instad of having a simple expression

### The solution
- Define `KernelMatrix*x*` structures for common kernel sizes (Like `KernelMatrix4x3`, `KernelMatrix4x1` etc.)
- Use specific kernel matrix structures (known at compile time!) as a generic arguments in implementation related classes (`ConvolutionProcessor<TColor, TKernelMatrix>` !)
  - Using `KernelMatrix9x1`-like constructs as resampler windows could bring significant performance improvements for ResizeProcessor (see #139) too!
- Implement all kernel-related calculations as simple expressions without for loops, etc (Like implementations of `Matrix4x4`)
- Use a generic matrix for unknown sizes (something like `Fast2DArray<float>`)

### TODO
- [ ] Prototype a benchmark comparing a heap matrix VS a value matrix convolution
- [ ] Implement stuff based on benchmark conclusions

JimBobSquarePants:
@antonfirsov I think we should try and make this a priority now if we're gonna get it in for V1.

antonfirsov:
@JimBobSquarePants hmm .. originally I wanted to polish + standardize our memory management in the next few months, but there are many uncertainties around that topic, so maybe I'll grab this one instead.

There is a lot of work here though, but maybe we can make it into 1.0.

JimBobSquarePants:
@antonfirsov Have a look and estimate it then. Looking at it myself I'm envisaging quite a lot of API changes to allow it which makes me think maybe it's for V2 instead.

JimBobSquarePants:
[...]

PR #275, opened 2017-07-12
Refactored Mutate/Clone APIs

### LINKED ISSUE #265: Option to have image operations return new images instead of mutating source image
We should add an option for image operations to switch between mutating the original image and returning a new image while leaving the source image untouched.

tocsoft:
### Prerequisites

### Description
This is a rather significatant change to our public APIs the proposal is thus.

Instead of `Image<TPixel>` having a range of extension methods on it to do all sorts of mutations that return the save instance I've introduced `IImageOperations<TPixel>` which is used to queue/indirecty apply `IImageProcessors` to the source image.

The way you get a handle on an `IImageOperations<TPixel>` is by using the only 2 remaining extension APIs from `Image<TPixel>` (should theses be moved directly onto `Image<TPixel>` instead of being extension methds?).

The driver for this is I wanted the mutateing APIs to be obvious that they are a fluent chaining/mutating set of operations that return the source image again instead of returning a image each time.

Additionally I wanted to enable a version of all the `ImageOperation`s that did return a new instance for those times you wanted to generate multiple outputs from the same input.

### So some code examples.

So to perform a simple set of operation you now do this:

```c#
image.Mutate(x=>x.Grayscale().Resize(10,10).Pixelatte()); // returns void
image.Save("mutatedImage.png");
```
this will mutate the pixel data stored in `image` first making it grayscale, then resizing it, and finaly pixelatting it.

now to do all that again by not alter the source image would be a simple as

```c#
using(var new_image = image.Clone(x=>x.Grayscale().Resize(10,10).Pixelatte())) // returns `Image`
{
    // deal with clone in here
    new_image.Save("newImage.png");
}

image.Save("origionalImage.png");
```

second example has left `image` in exactly the same unaltered state is was in before starting, and instead has created a new image `new_image` with all the mutations applied.

Fixes #265


src/ImageSharp/ApplyProcessors.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// </summary>
+    public static partial class ImageExtensions
+    {
+        /// <summary>
+        /// Mutates the image by applying the operations to it.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="source">The image to rotate, flip, or both.</param>
+        /// <param name="operations">The operations to perform on the source.</param>
+        public static void Mutate<TPixel>(this Image<TPixel> source, Action<IImageOperations<TPixel>> operations)
```
tocsoft: Question : should theses be extension methods or exposed instance methods of `Image<TPixel>`?

The more I think about it I feel these should actually be moved onto Image<TPixel> directly thus exposing this APIs without needing to add the using statements.

antonfirsov: We should be consistent:
- Either by redesigning our whole API to be fully extension method focused, so not only `Mutate()`, but all the methods/properties producing derived results from a narrow core API, should be extension methods. So even `image.Bounds` would become an `image.Bounds()` extension method.
- Or just turn both `Mutate(operations)` and `Clone(operations)` to instance methods as well

src/ImageSharp/IImageOperationsProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// Called during Mutate operations to generate the imageoperations provider.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format</typeparam>
+        /// <param name="source">The source image.</param>
+        /// <returns>A new IImageOPeration</returns>
+        IImageOperations<TPixel> CreateMutator<TPixel>(Image<TPixel> source)
+            where TPixel : struct, IPixel<TPixel>;
+    }
+
+    /// <summary>
```
antonfirsov: The naming seems inconsistent here.

It should be probably `IImageOperationsFactory` + `CreateImageOperations()`, or `IImageOperationsProvider` + `GetImageOperations()`.

src/ImageSharp/ImageOperations.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            return this.ApplyProcessor(processor, this.image.Bounds);
+        }
+
+        /// <summary>
+        /// Applies a bluck colelctino of pressorce at once
+        /// </summary>
+        /// <param name="processors">Processors to apply</param>
+        /// <returns>this </returns>
+        public IImageOperations<TPixel> ApplyProcessors(IEnumerable<IImageProcessor<TPixel>> processors)
```
antonfirsov: This method is unused.

src/ImageSharp/ApplyProcessors.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="source">The image to rotate, flip, or both.</param>
+        /// <param name="operations">The operations to perform on the source.</param>
+        /// <returns>Anew Image which has teh data from the <paramref name="source"/> but with the <paramref name="operations"/> applied.</returns>
+        public static Image<TPixel> Clone<TPixel>(this Image<TPixel> source, Action<IImageOperations<TPixel>> operations)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            Guard.NotNull(operations, nameof(operations));
+            var generated = new Image<TPixel>(source);
```
antonfirsov: For most processors we (eg. `Resize()`) we are allocating a new `Width`*`Height` image area twice!
- First calling this copy constructor
- Second time [in the processor itself](https://github.com/SixLabors/ImageSharp/blob/tocsoft/mutate-api/src/ImageSharp/Processing/Processors/Transforms/ResizeProcessor.cs#L77)

So the issue Eric Mellino pointed out on gitter is still here. We should find a way to avoid this during this API validation process. I'm not saying that we should refactor everything now, but we need to ensure it is possible with the new API!

src/ImageSharp/ApplyProcessors.cs
tocsoft: Bounds() should be an extension method I feel, as its only really used as a helper for image operations. (in fact it should probably just be internal)

We should also probably move out all all the overloads of `Save()` into extension methods too `Image` would just have `Save(Stream, Encoder)`.

Also things like `To<TPixel2>()` could also probably be a candidate for changing. 

And now the `ApplyProcessor()` is basically redundant.

src/ImageSharp/ApplyProcessors.cs
tocsoft: The internal parts of the should be refactorable to make it possible to prevent the double allocation.

antonfirsov: I'm not sure if it's possible without changing `IImageProcessor<TPixel>` design.

The processors doing allocation of a new image buffer should be able to take the source and destination image buffers from the outside. It would make `SwapPixelBuffers()` unnecessary:
https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Processing/Processors/Transforms/ResizeProcessor.cs#L97

@JimBobSquarePants thoughts?

antonfirsov: I may play around with this, making a proposal in the weekend.

tocsoft: I was planning to create a new internal interface `ICloneableImageProcessor<T> : IImageProcessor<TPixel>` (or similare ) then in the `DefaultImageOperator` check if the passed in `IImageProcessor` is in fact an `ICloneableImageProcessor` and if it is then instead of calling `IImageProcessor.Apply()` then it instead call `IImageProcessor.CloneAndApply()` which returns the cloned image. All it would require is the `DefaultImageOperator` cloning the source image lazerly the first time we encounter an `IImageProcesssor` and its not an `ICloneableImageProcessor` and if it is on then it will retain the cloned image and just call Apply with it on all remaining `IImageProcesssor`

tocsoft: feel free to have a go... basically my ideal situation (for v1) is to do it without having to change the public API any even if that means that only our internal processors have to extra power of generating the clone.

antonfirsov:
@JimBobSquarePants @tocsoft
finding a clean & maintainable solution for the [double cloning issue](https://github.com/SixLabors/ImageSharp/pull/275#discussion_r127365978) seems really hard, I'm struggling since hours with no success. One thing seems pretty sure: the `IImageProcessor<T>` API has to be changed to manage the copying family of processors in a clean way.

I don't think any of our users are implementing their own processors though, so probably a minor change on that API somewhere between the beta and the RC should not be an issue. Thoughts? :)

tocsoft:
@antonfirsov c421e6d is basically what I was thinking with regards the fix for double cloning... none of those changes actually alter the public API... so it is possible without having to change `IImageProcessor<>` true it doesn't expose the ability to third party developers but that can be done an a version or 2 once other devs start wanting to make processors for it.

src/ImageSharp/IImageProcessorApplicator{TPixel}.cs
```diff
@@ -0,0 +1,50 @@
...
+    using ImageSharp.Formats;
+    using ImageSharp.PixelFormats;
+    using ImageSharp.Processing;
+    using SixLabors.Primitives;
+
+    /// <summary>
+    /// An interface to queue up image operations.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format</typeparam>
+    public interface IImageProcessorApplicator<TPixel>
```
antonfirsov: Shouldn't we find a less verbose name for this? How about `IImageProcessingContext<TPixel>`? //`image.Mutate(ctx => ctx.Resize(..) )`

src/ImageSharp/Image/ICloningImageProcessor.cs
```diff
@@ -0,0 +1,37 @@
...
+    using System.Threading.Tasks;
+
+    using ImageSharp.PixelFormats;
+    using SixLabors.Primitives;
+
+    /// <summary>
+    /// Encapsulates methods to alter the pixels of an image.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    public interface ICloningImageProcessor<TPixel> : IImageProcessor<TPixel>
```
antonfirsov: Wasn't your intention to internalize this?

src/ImageSharp/Processing/Processors/CLoneingImageProcessor.cs
```diff
@@ -0,0 +1,166 @@
...
+    using System.Threading.Tasks;
+
+    using ImageSharp.PixelFormats;
+    using SixLabors.Primitives;
+
+    /// <summary>
+    /// Allows the application of processors to images.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    internal abstract class CloneingImageProcessor<TPixel> : IImageProcessor<TPixel>, ICloningImageProcessor<TPixel>
```
antonfirsov: Typo: `Cloneing...`

Can't we avoid code duplication by subclassing `ImageProcessor<TPixel>`?

tocsoft: there is no duplication really, they both follow the same patterns but the code is different, one takes 2 images the other takes 1. Basically `CloneingImageProcessor` is for image processors that require a 2nd buffer to apply changes to before overwriting the pixel data in the original. `ImageProcessor` if for processors that don't require the 2nd buffer and can apply there changes directly to the source.

src/ImageSharp/Image/ICloningImageProcessor.cs
tocsoft: yep i was, must have missed it

src/ImageSharp/IImageProcessorApplicator{TPixel}.cs
tocsoft: I like `IImageProcessingContext`

JimBobSquarePants:
@tocsoft How big a change to the `IImageProcessor<T>` API are we talking here? Ideally I'd rather not have to have something that developers cannot implement themselves with our primitives. 

Happy with the rest of the changes btw. It'll take me a while to get used to `Mutate()`returning `void` though.

tocsoft:
The major difference to `IImageProcessor<T>` is that it now takes an `Image<T>`  instead of an `ImageBase<T>` and now its responsible for making sure all the frames are processed too.

JimBobSquarePants:
Sorry, I meant future changes for preventing double cloning.

tocsoft:
ah there are no changes directly to `IImageProcessor<T>` but we add a new interface (currently internal) `ICloningImageProcessor<T>` that we detect and use its `CloneAndApply()` (which takes in the source and returns the cloned image) method instead of  `IImageProcessor<T>.Apply()`. Basically its optional and if you want to opt in implement the other interface. Additionally we have a base class `CloningImageProcessor` that we could expose that would make it quite easy to implement a processor that always requires a fresh/scratch pixel area to run its process. 

https://github.com/SixLabors/ImageSharp/pull/275/files#diff-515733e0588e848c4a58efcf630b0eef is an example of it being used for resizeing.

JimBobSquarePants:
@tocsoft I can't make my mind up about this. 

`Mutate()` and `Clone()` look good on the surface but we also have the `Apply` (Formerly `Run` - I just changed the name) delegate. It feels like we've added a level of abstraction and misdirection that makes the code more difficult to follow + adds new overheads (heap allocation due to capturing of params). What benefit does it bring? 

I also find the processing context stuff confusing, calls like: 
`source.Configuration.ImageOperationsProvider.CreateImageProcessingContext` are a red flag to me. Discovery of methods four levels deep is virtually nonexistent.

The internal interface `IInternalImageProcessingContext<TPixel>` and its default implementation are a flag also. What other implementations are we expecting? Does this abstraction need to happen?

I'm struggling a little to recognize the codebase tbh, that could be the flu I have just now limiting me but it all feels more complicated than it should be.... and the more complicated it is, the slower it is and the slower we, and others can develop against it. One of my primary goals with this library was readability. I don't want to sacrifice that.

@antonfirsov @dlemstra What are your thoughts?

tocsoft:
> Mutate() and Clone() look good on the surface but we also have the Apply (Formerly Run - I just changed the name) delegate. It feels like we've added a level of abstraction and misdirection that makes the code more difficult to follow + adds new overheads (heap allocation due to capturing of params). What benefit does it bring?

Benefit is that its obvious to all consumers of image sharp about image operations being mutative and not confusing users into thinking that can dispose of the returned image and expect the original back (i've seen this confusion many times thus a huge red flag to me)

> adds new overheads (heap allocation due to capturing of params). What benefit does it bring?

we where capturing the params anyway, each IImageProcessor should have all the info it needs to alter an image in a specific way.

> I also find the processing context stuff confusing, calls like:
source.Configuration.ImageOperationsProvider.CreateImageProcessingContext are a red flag to me. Discovery of methods four levels deep is virtually nonexistent.

this method is only used internally and only exists for testing purposes, it allow us to actually unit test out extension method logic without the stupid overhead of trying to run the processor and guess if the settings looked with based on pixel data returning.

>The internal interface IInternalImageProcessingContext<TPixel> and its default implementation are a flag also. What other implementations are we expecting? Does this abstraction need to happen?

Again these are interfaces etc for testing purposes , and allows us flexibility in the future if we need it.

> I'm struggling a little to recognize the codebase tbh, that could be the flu I have just now limiting me but it all feels more complicated than it should be.... and the more complicated it is, the slower it is and the slower we, and others can develop against it. One of my primary goals with this library was readability. I don't want to sacrifice that.

The process of developing new `IImageProcessor`s hasn't changed... its just that they should really be standalone and not require any processing logic in the extension methods like we have at the moment (and the only reason I had to add `Run()`) which to a massive architectural/design red flag separating the logic across many file.

JimBobSquarePants:
> `Mutate()` and `Clone()` look good on the surface but we also have the `Apply` (Formerly Run - I just changed the name) delegate. It feels like we've added a level of abstraction and misdirection that makes the code more difficult to follow + adds new overheads (heap allocation due to capturing of params). What benefit does it bring?

>**Benefit is that its obvious to all consumers of image sharp about image operations being mutative and not confusing users into thinking that can dispose of the returned image and expect the original back (i've seen this confusion many times thus a huge red flag to me)**

Sorry... I means just the `Apply()` bit. `Mutate()` and `Clone()` are great!


>I also find the processing context stuff confusing, calls like: source.Configuration.ImageOperationsProvider.CreateImageProcessingContext are a red flag to me. Discovery of methods four levels deep is virtually nonexistent.

>**this method is only used internally and only exists for testing purposes, it allow us to actually unit test out extension method logic without the stupid overhead of trying to run the processor and guess if the settings looked with based on pixel data returning.**

We should document those classes then and if there's only one context implementation simplify the methodology for now. I can't see us introducing new contexts.

>**The process of developing new IImageProcessors hasn't changed... its just that they should really be standalone and not require any processing logic in the extension methods like we have at the moment (and the only reason I had to add Run()) which to a massive architectural/design red flag separating the logic across many file.**

`Run/Apply` shouldn't be public then. Let's get rid of that logic then and move the offending code to the processors. I meant to do it ages ago.

samples/AvatarWithRoundedCorner/Program.cs
```diff
@@ -8,62 +8,74 @@ namespace AvatarWithRoundedCorner
...
-                    Size = size,
-                    Mode = ImageSharp.Processing.ResizeMode.Crop
-                });
-
-                ApplyRoundedCourners(image, cornerRadius);
-                image.Save(destination);
-            }
+                Size = size,
+                Mode = ImageSharp.Processing.ResizeMode.Crop
+            }).Apply(i => ApplyRoundedCorners(i, cornerRadius));
```
antonfirsov: I'm also really confused about the `Run()` / `Apply()` + `DelegateProcessor<T>` stuff. Is the purpose of this to allow users to access image data in the processor (sizes etc?).

I think we should find a way to enable `ApplyRoundedCorners()` to work on `IImageProcessingContext<T>`, it's just too much indirection now.

tocsoft: I personally preferred `Run()` over apply but eather way. 

Basically its an escape hatch for developers to be able to access the correctly cloned/mutated image without having to go though the rigmarole of writing an entire `IImageProcessor`.

antonfirsov: It still makes things more complicated which were simple before the Mutate/Clone API change. I had to debug through the code to understand it, and it also confused @JimBobSquarePants, so I think he is right we should find a simpler way to manage accessing the image inside the processing lambda.

How about `IImageProcessingContext<T>.CurrentImage`?

tocsoft: `CurrentImage` is exactly what we shouldn't have, adding that would prevent us easily being able to do optimizations around rolling up similar processors etc into single set of pixel operations or even things like detecting that some operations will have no effect on the final image thus can be skipped (eg after a user applies a lot of operations then calls `Fill()` with a solid color thus we could skip all the previous operations and just call fill as it'll fully set every pixel anyway.)

Or for rolling up we could have multiple calls to draw text we would be able to roll all the calls into a single operation that only renders the glyphs as a single pass instead of multiple operations.

JimBobSquarePants: I don't think we should be trying to predict behavior. If we want grouped operations we should create processors to do that. I.E anything using matrices can actually be combined (we should actually do that), maybe even affine transforms (I tried combining rotate and skew once, it didn't go well 😝).

To me, an imaging API should be imperative, We should optimize the building blocks yeah, but we should strictly follow the input instructions.

antonfirsov: I think speeding up individual processors by eliminating per-pixel virtual calls everywhere + introducing #142 could bring much more at this point, than any chaining optimization.

Optimizing processor chains sounds like a 2.0 feature for me. At this point having a straightforward API is more important IMO.

antonfirsov: @JimBobSquarePants shouln't we try to make a decision on this before @tocsoft leaves on holidays?
To avoid conflicts in the Tests project, it would be wise to build my regression testing PR based on top of the `mutate-api`. I need the test stuff to proceed with jpeg.

JimBobSquarePants: Yes we should!

I like `Mutate()` and `Clone()` but I'm still not comfortable with the `Apply()` lambda indirection, much preferring the simplicity that was present before when invoking the processors. Writing an `IImageProcessor` should be easy enough for anyone with the requirement to do so. 

I still feel strongly also that our API should be imperative and we should always only operate on what instructions we are given. 

So I'm voting to remove `Apply()` for now.

antonfirsov: Yeah, but that means, we need something like `IImageProcessingContext<T>.CurrentImage` to extract the image dimensions etc. for the next processing step.
Unless someone finds a better solution, I'm totally fine with it.

antonfirsov: @JimBobSquarePants If we just remove `Apply()` without an easy replacement, that means the user has to implement his own ImageProcessor-s just to enable very basic use cases.

JimBobSquarePants: I don't fancy `CurrentImage` much either.

What use cases are we thinking of? Before there was just a simple loop so each processor always had access to the up-to-date image dimensions from the image passed to the processor.

antonfirsov: Or we can probably just update this example, and suggest doing operations like this in two steps:
```C#
image.Mutate(ctx => ctx.Resize());
image.Mutate(ctx => DrawTheRoundedCorners(ctx));
```

Sorry for thinking loudly and slowly across 3 comments :laughing:

JimBobSquarePants: Ah... Hold on @antonfirsov , I think this just clicked. 

`Apply()` is good, very good in fact, I'd somehow missed `DelegateProcessor` and how that works. Apologies to all for my slowness here. I can't believe I missed that! 😳 

@tocsoft You have my approval. Let's update this from our main branch and do comment cleanup/namespace changes and get it out there. 👍

antonfirsov: @JimBobSquarePants @tocsoft my only concern: will this be clear for users? (We both needed time to understand it...)
Wouldn't it be better to split the steps in the example instead as I suggested in my previous comment? Or to show multiple variants of `ConvertToAvatar()`: one operating in two steps, another one with `Apply()`.

And finally: Don't we need a more descriptive name for `Apply()`? Like `ApplyDelegateProcessor()` or `RunDelegateProcessor()`?

JimBobSquarePants: @antonfirsov I think the difference is that we were looking at 200+ changes and have visibility of the internals which complicates things. There's a phrase in English "Can't see the wood for the trees" which describes perfectly my issues with understanding here.

We should flesh out and heavily comment our examples to make things easier for consumers to understand. I like the name `ApplyDelegateProcessor()` as it describes exactly what we are doing but do we need the user to be aware of the underlying operation? Maybe with the correct intellisense docs `Apply()` can be enough.

tocsoft: Thats why I called it `Run`  instead of `Apply`... As to me it invokes the though of run this action against the image. Also helps differentiate itself from applying `IImageProcessor`s

JimBobSquarePants:
`Run` makes me think of `Task` which doesn't feel contextually correct. `Apply` makes me think of JavaScript's `function.Apply` which would be contextually correct.

antonfirsov:
@JimBobSquarePants @tocsoft tried to clarify the avatar example by adding some docs, comments + a variant, that does the trick without `Apply()`.

@JimBobSquarePants if you say that users who are not messed with 200 git changes+flu should understand all this, I believe you! :smile:
Everything looks good to me now!

tocsoft:
Those updates look good to me 👍 

I like that we're showing both methods as its demonstrating that we aren't prescribing doing it one way or the other.

antonfirsov:
Here is the only problem left:
- On one hand, we should merge this ASAP to unblock upcoming PR-s.
- On the other hand: this is a big change, and we need to explain it to our users first, together with the reasons driving to this change. I'm afraid, we should write a blog post before merging.

tocsoft:
I could use this as the base branch for doing the package renames/namespace changes. That should cause less disruption on people who inadvertently update to latest as they would be having to to pull in a different package anyway. 

Also means there is only one large braking change instead of 2.

tocsoft:
obviously would still require a blog post but it would be a single one with API, package name in one.

antonfirsov:
:+1: Good thinking!

Issue #258, opened 2017-06-26
Is there any equivalent for System.Drawing.Image.GetPixelFormatSize()?


xakep139:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Just a short question, any help will be appreciated!

tocsoft:
You can use the standard `sizeof(struct)` operator to get the size of whichever pixel format you need.

e.g. to find the size of the `Rgba32` pixel struct you can just use `var size = sizeof(Rgba32);`.

xakep139:
Great! But if I don't know pixel format of an image? Default format for `Image.Load()` is `Rgba32`, even if I load 8 bpp PNG

tocsoft:
That's right, the pixel format is always whatever the `TPixel` is in `Image<TPixel>` how the underlying file format stores the data is ignored outside the decoder, the decoders convert it to the correct `TPixel` type you asked for..

JimBobSquarePants:
Yes, You load the image as the given `TPixel` type. 

- `Image.Load(stream)` and `Image<Rgba32>.Load` both load the type as `Rgba32`
- `Image<Rgba24>.Load` loads the type as `Rgba24`

and so on...

xakep139:
I got it, but I have to determine what exact BPP has particular image stored in `Stream`/`byte[]`/etc.

antonfirsov:
AFAIK, you can't do that currently. @JimBobSquarePants shouldn't we convert this into a feature request issue or open a new one?

JimBobSquarePants:
Really this is a misuse of the issue tracker. The question should have been asked in Gitter. 

It's a sensible requirement though so let's continue here and add the correct tags.

xakep139:
> You can use the standard sizeof(struct) operator to get the size of whichever pixel format you need.
> e.g. to find the size of the Rgba32 pixel struct you can just use var size = sizeof(Rgba32);

[...]

PR #292, opened 2017-08-10
Basic metadata-only image parsing API

xakep139:
### Prerequisites

### Description

Implemented feature request #258


tests/ImageSharp.Tests/TestFileSystem.cs
```diff
@@ -22,7 +22,7 @@ public class TestFileSystem : ImageSharp.IO.IFileSystem
 
         public static TestFileSystem Global { get; } = new TestFileSystem();
 
-        public static void RegisterGloablTestFormat()
+        public static void RegisterGlobalTestFormat()
```
xakep139: Fix typo

tests/ImageSharp.Tests/TestFormat.cs
```diff
@@ -23,15 +23,15 @@ public class TestFormat : IConfigurationModule, IImageFormat
     {
         public static TestFormat GlobalTestFormat { get; } = new TestFormat();
 
-        public static void RegisterGloablTestFormat()
+        public static void RegisterGlobalTestFormat()
```
xakep139: Fix typo

tests/ImageSharp.Tests/TestFormat.cs
```diff
@@ -200,10 +200,15 @@ public TestDecoder(TestFormat testFormat)
...
                 });
 
-                // TODO record this happend so we an verify it.
+                // TODO record this happend so we can verify it.
                 return this.testFormat.Sample<TPixel>();
             }
 
+            public int DetectPixelSize(Configuration configuration, Stream stream)
+            {
+                throw new NotImplementedException();
```
xakep139: Not implemented for test format

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -1238,7 +1254,7 @@ private void ProcessStartOfFrameMarker(int remaining)
 
                 if (h == 3 || v == 3)
                 {
-                    throw new ImageFormatException("Lnsupported subsampling ratio");
+                    throw new ImageFormatException("Unsupported subsampling ratio");
```
xakep139: Fix typo

tests/ImageSharp.Tests/Formats/Gif/GifDecoderTests.cs
```diff
@@ -80,5 +81,19 @@ public void Decode_TextEncodingSetToUnicode_TextIsReadWithCorrectEncoding()
                 Assert.Equal("浉条卥慨灲", image.MetaData.Properties[0].Value);
             }
         }
+
+        [Theory]
+        [InlineData(TestImages.Gif.Cheers, 8)]
+        [InlineData(TestImages.Gif.Giphy, 8)]
```
xakep139: For these two GIFs `Image.GetPixelFormatSize()` returns 32 which is not correct

tests/ImageSharp.Tests/Image/ImageLoadTests.cs
```diff
@@ -55,28 +55,28 @@ public ImageLoadTests()
...
 
-            TestFileSystem.RegisterGloablTestFormat();
+            TestFileSystem.RegisterGlobalTestFormat();
             TestFileSystem.Global.AddFile(this.FilePath, this.DataStream);
         }
 
         [Fact]
         public void LoadFromStream()
         {
-             Image<Rgba32> img = Image.Load<Rgba32>(this.DataStream);
```
xakep139: Indentation here

antonfirsov:
@xakep139 nice job!
I wonder if instead of having a dedicated `DetectPixelSize()` method, wouldn't it be better to introduce and return a `PixelTypeInfo` object? We could bridge it to our `PixelFormats` later.

xakep139:
@antonfirsov I apologize for delayed response. Your idea is pretty good, I'll implement it.
Initially I thought about detecting `TPixel`, but some bpp's couldn't be converted into it.

xakep139:
@antonfirsov do you have other suggestions?

tests/ImageSharp.Tests/TestImages.cs
```diff
@@ -25,6 +25,9 @@ public static class Png
             public const string Powerpoint = "Png/pp.png";
             public const string SplashInterlaced = "Png/splash-interlaced.png";
             public const string Interlaced = "Png/interlaced.png";
+            public const string Palette8Bpp = "Png/Palette-8bpp.png";
```
antonfirsov: Unix systems are case-sensitive, so you have to rename this to `palette-8pp.png`, to match the filename. There is a [test failure](https://travis-ci.org/SixLabors/ImageSharp/builds/267901746#L2360) on travis because of that.

antonfirsov:
@xakep139 no, the `PixelTypeInfo` thing is the only one, but we are releasing our first beta soon, which means that we need to be very strict defining our public API. I think we need to implement this with the `PixelTypeInfo` stuff now, because there is no chance to change it later.

@JimBobSquarePants any suggestions to manage this?

tests/ImageSharp.Tests/TestImages.cs
xakep139: fixed in https://github.com/SixLabors/ImageSharp/pull/292/commits/d6a2d40a80a60e0eb25861d7714f4cfd5b8cc90d

xakep139:
@antonfirsov I've synced the branch with upstream

antonfirsov:
@xakep139 thanks! (And also thanks for your patience!)

Someone has to refactor your work to the `PixelTypeInfo` style so we can merge it. I'll try to be that someone ...
Can I push right to your branch?

xakep139:
@antonfirsov, yes, sure

xakep139:
@antonfirsov any plans on this PR?

antonfirsov:
@xakep139 Sorry for the late response! I gonna deal with this during my ImageSharp hackaton in Feburary. It could be related to #430, because we should probably use the same `PixelFormatInfo` class on both API-s.

I would probably move the method `DetectPixelType()` into another interface to reduce API pollution, because not every `IImageDecoder` can implement it now:

```C#
internal interface IPixelTypeDetector
{
    PixelTypeInfo DetectPixelType(Configuration configuration, Stream stream)
}

public class PngDecoder : IImageDecoder, IPixelTypeDetector
{
   ...
}
```

@JimBobSquarePants I really think this is a valuable feature, and we should merge this. Thoughts about my ideas?

xakep139:
In our system we should check bit depth of uploaded images, it's technical requirement of another system

xakep139:
In general it could be usefull to get metadata without loading (decoding) the whole image in memory. E.g. to check image size (width, height), get pixel type (RGB, RGBA, etc.) and bit depth (or a pixel size).
For now we could only get image format with `Image.DetectFormat()` without loading.
Such feature (like `magick identify` in [ImageMagick](https://www.imagemagick.org/script/identify.php)) could be very helpfull.

JimBobSquarePants:
Thanks @xakep139 
Ok that makes sense, as long as we are not trying to use it to automatically determine the `PixelFormat` to use.

@antonfirsov Would we then use reflection to determine which decoders support what we need?

tocsoft:
we would register them the same we we do with the encoders/decoders i.e. we would add `Configuration.SetPixelTypeInfoProvider(ImageFormat format, IPixelTypeDetector detectorInstance)`

antonfirsov:
Actually I wanted to keep things as dumb as:

```C#
static class Image
{
    ...

    static bool TryDetectPixelType(Configuration config, Stream stream, out PixelTypeInfo pixelTypeInfo)
    {
        IImageDecoder decoder = DiscoverDecoder(stream, Configuration config);
        IPixelTypeDetector detector = decoder as IPixelTypeDetector;
        if (detector == null) return false;
        pixelTypeInfo = detector.DetectPixelType(config, stream);
        return true;
    }

    ...
}
```

I'm afraid that going any further, and exposing `IPixelTypeDetector` publicly + registering it into `Configuration` really means starting the design of a public metadata API, which I think is outside the scope of this PR.

JimBobSquarePants:
Yeah let's leave it simple there and try not to expose another public API yet.

denisivan0v:
@antonfirsov @JimBobSquarePants Me and @xakep139 are working on the same project and now we extremely need an API that would provide base image information without decoding it. 
So, I've made changes that was mentioned in https://github.com/SixLabors/ImageSharp/pull/292#issuecomment-357602218:

* introduced base [`IImage`](https://github.com/SixLabors/ImageSharp/pull/292/files#diff-70bebbed1a8bf52bdf94497d7856525eR9) interface (as it was requested in #430)
* introduced [`IImageInfoDetector`](https://github.com/SixLabors/ImageSharp/pull/292/files#diff-d4db53c88c24b507d6189745936daeeeR11) interface (following to this https://github.com/SixLabors/ImageSharp/pull/292#issuecomment-357518960)
* `Image.DetectPixelType` method expanded to [`Image.Identify`](https://github.com/SixLabors/ImageSharp/pull/292/files#diff-66eb52dcbd709d404f9026dd97d7ce79R65) method that returns `IImage`

Hope it will also help to resolve #430.

I'd be much appreciated for any suggestions and comments.

antonfirsov:
@denisivan0v thanks for joining in!

I really hope we'll be able to figure out how to provide the feature you need. But let's start with a very important disclaimer:
For the ImageSharp team, maintaining a consistent & clean API has a higher priority than individual features! So expect us to be very strict at the review process.

The major issue is that ImageSharp has not been designed to support metadata-only scenarios. I understand however that many users need it, and I think it's worth for us to find a way for supporting them, so they don't have to look for other libraries. (@JimBobSquarePants do you agree?)

However, to avoid the super-hard design questions I foresee here, this feature set should be limited & isolated at this point, with future-proof API modifications only.

src/ImageSharp/Image/IImage.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// Represents the base image abstraction.
+    /// </summary>
+    public interface IImage
+    {
+        /// <summary>
+        /// Gets information about pixel.
+        /// </summary>
+        PixelTypeInfo PixelType { get; }
```
antonfirsov: I suppose this value holds the native pixel type in the stream, after `Identify()`.
What's the deal with `Image.Load<TPixel>()`? What if `typeof(TPixel)` ≠ `the native pixel type in the stream` ?

JimBobSquarePants: That situation will happen more often than not since 3 out of 4 of our currently supported image formats are commonly RGB.

If it's going to be part of the interface then both The property and its type should be renamed as it's just confusing now. What would the equivalent information be called in other libraries?

src/ImageSharp/Image/IImage.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace SixLabors.ImageSharp
+{
+    /// <summary>
+    /// Represents the base image abstraction.
+    /// </summary>
+    public interface IImage
+    {
+        /// <summary>
+        /// Gets information about pixel.
```
JimBobSquarePants: Lets flesh out the property descriptors here. This doesn't tell me anything and wont help with docs.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -242,6 +295,7 @@ private void ReadLogicalScreenDescriptor()
             {
                 Width = BitConverter.ToInt16(this.buffer, 0),
                 Height = BitConverter.ToInt16(this.buffer, 2),
+                BitsPerPixel = (this.buffer[4] & 0x07) + 1,  // The lowest 3 bits represent the bit depth minus 1
```
JimBobSquarePants: Was hoping this wouldn't trip you up. Good research. 👍

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -327,7 +385,7 @@ private void ReadFrame()
                 indices = Buffer<byte>.CreateClean(imageDescriptor.Width * imageDescriptor.Height);
 
                 this.ReadFrameIndices(imageDescriptor, indices);
-                this.ReadFrameColors(indices, localColorTable ?? this.globalColorTable, imageDescriptor);
+                this.ReadFrameColors(ref image, ref previousFrame, indices, localColorTable ?? this.globalColorTable, imageDescriptor);
```
JimBobSquarePants: `Image<T>` and `ImageFrame<T>` are already reference types?

src/ImageSharp/Image/ImageInfo.cs
```diff
@@ -0,0 +1,24 @@
+﻿using SixLabors.ImageSharp.Formats;
+using SixLabors.ImageSharp.MetaData;
+
+namespace SixLabors.ImageSharp
+{
+    internal sealed class ImageInfo : IImage
```
JimBobSquarePants: Please add xml docs

src/ImageSharp/Formats/PixelTypeInfo.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace SixLabors.ImageSharp.Formats
+{
+    /// <summary>
+    /// Stores information about pixel.
```
JimBobSquarePants: We need a better description than this.

I don't like the name. `RawPixelTypeInfo` perhaps? Naming is hard!

src/ImageSharp/Image/IImage.cs
antonfirsov: However, if we rename it to something like `OriginalPixelType`, while keeping it in `IImage` means that we added a stream-specific metadata-like property to the `Image<T>` class which is mostly intended to represent pixel data.

My suggestion:
Let's name it `OriginalPixelType` and move it into the `ImageMetaData` class.

src/ImageSharp/Formats/PixelTypeInfo.cs
antonfirsov: @JimBobSquarePants I proposed this class and naming in order to make it extendable in a way like this:

```C#
public class PixelTypeInfo
{
    public System.Type PixelType { get; }
    public int BitsPerPixel { get; }

    internal PixelTypeInfo(int bitsPerPixel, System.Type pixelType = null)
    {
        this.BitsPerPixel = bitsPerPixel;
    }

    public static PixelTypeInfo Create<TPixel>() where TPixel 
        :  IPixel<TPixel> => 
        new PixelTypeInfo(Unsafe.SizeOf<TPixel>(), typeof(TPixel));
}
```

This will enable integrating it deeper with our pixel type infrastructure, and provide non-generic pixel-type related API-s. As a bridge between native (image format/stream bound) pixel type data and our internal pixel types, this class doesn't necessarily represent "raw"/native pixel type information.

src/ImageSharp/Image/IImage.cs
tocsoft: Image<T> shouldn't have any knowledge of its original loaded data.. this is why we removed the original image format from Image<T> I think we need this API to return an `ImageMetadata` object instead that exposes all this info and not the based `IImage` interface.

antonfirsov: Okkay, let's sum up the situation, so we can figure it out:
- @tocsoft Not sure if I'm comfortable with this fact, but we already [have `MetaData` exposed](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Image/Image%7BTPixel%7D.cs#L101) on `Image<TPixel>`.
- We need to include Width+Height (or maybe `SixLabors.Size`?) into the metadata-only result. Currently it's not exposed on `ImageMetadata`.
  - @denisivan0v solved this by defining `ImageInfo : IImage` as the metadata-only result
  - But maybe these are just different things, and sharing a common interface is an LSP violation.

@denisivan0v @xakep139 do you actually need the `ImageInfo`/`Image<T>` polymorphism? Eg: do you want to mix them in heterogenous containers? For me it feels like an unrealistic scenario.

tests/ImageSharp.Tests/TestUtilities/ReferenceCodecs/SystemDrawingReferenceDecoder.cs
```diff
@@ -1,19 +1,17 @@
...
 
 using SixLabors.ImageSharp.Formats;
 using SixLabors.ImageSharp.PixelFormats;
 
-using PixelFormat = System.Drawing.Imaging.PixelFormat;
-
 namespace SixLabors.ImageSharp.Tests.TestUtilities.ReferenceCodecs
 {
-    public class SystemDrawingReferenceDecoder : IImageDecoder
+    using SixLabors.ImageSharp.MetaData;
```
JimBobSquarePants: Why were the namespaces altered in this class?

src/ImageSharp/Image/IImage.cs
JimBobSquarePants: >But maybe these are just different things, and sharing a common interface is an LSP violation.

I'm not sure... Reading through the code again I now see why I'm uneasy and suggested the change. It's because it's merging the two identities as one.

See https://github.com/SixLabors/ImageSharp/pull/292/files/9f719d5b9423083c112a69144e98a15bbb8c9409#diff-b4f626cf509fdbafd46cc45da8aed941R429 for example. The code takes the raw input format and passes it into the image, which now contains conflicting information. 

My Suggestion:

- Drop that new overload to the `Image<TPixel>` constructor and calculate the `PixelTypeInfo` property automatically in our current constructor.
- Keep naming as-is, including interface sharing.
- Write decent intellisense documentation describing what `ImageInfo` is, clearly highlighting that it contains the raw info.

That way we preserve two separate identities. The current raw identity that can be used for making informed decisions as to what `TPixel` to use etc, and our point-in-time representation of the image pixels.

antonfirsov: @JimBobSquarePants I'm not sure if I'm following you correctly. 
Do you suggest to keep the `PixelType` property on the `IImage` interface + keep it in sync with `TPixel` in case of `Image<TPixel>`? Or the opposite: always exposing the raw type after `Image.Load<TPixel>()` regardless of `TPixel`?

JimBobSquarePants: @antonfirsov The former.

For `Image<TPixel>` the `PixelType` property represents `TPixel`

For `ImageInfo` the `PixelType` property represents bits per pixel of the source image.

antonfirsov: Makes sense, and looks like being in sync with my vision about the role of `PixelTypeInfo` as bridge between `TPixel` and raw formats, and future pixel-type agnostic Image API-s. (System.Drawing is decoding images into their native formats, we might implement this feature with `IImage`.)

However I'm still unsure if the result of `Identify()` is actually an **image**. Maybe we should rename to `IImageInfo`, and reserve `IImage` for future use cases: 
`IImage` should represent an image that *actually has pixels*, so it could be used with processors, `.Mutate()`, `.Clone()` etc.

I think we should implement the following hierarchy in long term:
```C#
// Has Dimensions + PixelTypeInfo + Metadata
public interface IImageInfo { ... }

// Also has pixels, could be used with ImageProcessor-s.
public interface IImage : IImageInfo { ... }

internal class ImageInfo : IImageInfo { ... }

public class Image<TPixel> : IImage { ... }
```

JimBobSquarePants: Yeah, let's do the rename. 👍

tests/ImageSharp.Tests/TestUtilities/ReferenceCodecs/SystemDrawingReferenceDecoder.cs
denisivan0v: That's wasn't my intension. I guess it's because of my incorrect tooling settings (I'm using JetBrains Rider on Mac). Reverted it in 61c9caf.

src/ImageSharp/Image/ImageInfo.cs
denisivan0v: Done in 61c9caf

src/ImageSharp/Image/IImage.cs
denisivan0v: Done in 61c9caf.

tests/ImageSharp.Tests/ImageSharp.Tests.csproj
```diff
@@ -17,6 +17,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="CoreCompat.System.Drawing" Version="1.0.0-beta006" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
```
JimBobSquarePants: Do we need this?

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
denisivan0v: I had to remove `TPixel` type parameter from this class and move it `Decode` method. So `Image<TPixel> image` and `ImageFrame<TPixel> previousFrame` fields also removed and became local variables in `Decode<TPixel>` method. They marked with `ref` because they are reassigning down the call stack.

tests/ImageSharp.Tests/ImageSharp.Tests.csproj
denisivan0v: This enables unit tests execution in JetBrains Rider, but it's not necessary.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
denisivan0v: Actually, this was initially done by @xakep139 👍

tests/ImageSharp.Tests/ImageSharp.Tests.csproj
JimBobSquarePants: Let's keep it then. I want everyone to be able to test the project in their desired IDE.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
xakep139: Initially done in https://github.com/SixLabors/ImageSharp/pull/292/commits/f8573e334a4673a31665ee280a10b0e4eb96a15a#diff-7ae85077a97a9237d124f0ddfbe4752c

src/ImageSharp/Image/IImage.cs
denisivan0v: Great design review, thanks!

I also wasn't fully satisfied that `Image<T>` and internal `ImageInfo` shares the same `IImage` interface. But `IImageInfo` and suggested hierarchy solved the problem and made things clear.

Done in 61c9caf.

src/ImageSharp/Formats/PixelTypeInfo.cs
denisivan0v: See this https://github.com/SixLabors/ImageSharp/pull/292#discussion_r162216541.

JimBobSquarePants:
Many, many thanks @xakep139 and @denisivan0v ! 🥇 

This is an incredibly valuable addition to the library and we're so happy you stuck with us to get it merged in!

PR #296, opened 2017-08-16
Added more failing tests for some PNG and GIF images

devedse:
### Prerequisites

### Description

Added some new tests for images that don't get decoded correctly.


antonfirsov:
@devedse so these images should be identical if decoded with a correct PNG decoder?

We should add these tests to my new integration test suite, the current test architecture on `master` is not suitable for such reference image based tests. You have images "1" and "2". Which one should be considered as "expected" and which one as "actual"?

devedse:
Yeah adding them to your suite is fine with me :smile:. Not sure where I can find it though.

One problem with actual & expected is that the actual could also be wrong if the PNG decoder fails. So ideally we would have some kind of reference decoder that could be used for loading reference images. (But since we don't really have that I think, both images are actually "Actual", and there is no such thing as "Expected").

We could ofcourse just rename the image that is currently failing to "Actual". (The way I could find this out is by simply running the unit test, then checking which pixel it fails for, then opening both images in Paint.NET, and then opening the image in Paint.NET and seeing what the value is of that pixel, and then seeing which of the 2 images loaded with ImageSharp matches this pixel value.)

antonfirsov:
@devedse I'm closing this in favor of #301. We have some skipped failing tests indicating bugs in PNG decoder on the `beta-1` branch. 
Let's see what will be the result with the rest of your images, when someone manages to fix these bugs!

devedse:
Alright, let me know when it's fixed. I'll rerun all my tests then :smile:.

PR #297, opened 2017-08-17
Update namespaces & package names

tocsoft:
This PR does 6 things;

1. updates the namespace to SixLabors.ImageSharp.*
2. updates package names to SixLabors.IamgesSharp.*
3. updates the style of file headers to the new SixLabors agreed style used by other projects
4. updates code style for using statements to the SixLabors agreed style used by other projects.
5. update build process so that it use gitversion to generate CI version numbers and allow release version numbers to be controlled by tags
6. updates some dependencies to 4.4.0 versions instead of 4.3.0 versions.

This PR targets #275


JimBobSquarePants:
Ooh this is exciting! 😁

PR #299, opened 2017-08-19
Beta 1

JimBobSquarePants:
### Prerequisites

### Description
This PR collects the following PR's together so we can release our first beta. I really want to get this out so I took the bull by the horns and did some of the manual merging required to put it all together.

- #275
- #297 
- #298 

Which means we have the new API, namespace, and jpeg decoder included in the codebase.

The project builds but I couldn't get the reference image tests running locally for the new `qa-lab` stuff. 

@antonfirsov How does the external image references thing work? 
@tocsoft I seem to have broken AppVeyor, no idea how.

Please have a good look through this and make sure all is well. I'll get cracking on a blog post asap unless someone volunteers.


antonfirsov:
@JimBobSquarePants hmm .. interesting approach. I thought we will just disable MyGet releases, and merge+review the PR-s one-by one, but OK, maybe we can speed up things a bit! :) There are several  issues though we need to deal with:
1. Rightfully failing tests introduced by `qa-lab` indicating bugs (OilPaint, Png). What should we do with those?
2. We need to find a good place for the [external image repository](https://github.com/antonfirsov/Imagesharp.Tests.Images). Currently it's under my personal account. I'd create a clean repository under SixLabors and just copy all the stuff here to make it smaller, by removing history. Can I go on with this?
3. #298 was WIP, so I disabled some of the StyleCop rules, beacuse I don't want to document stuff that's gonna be eventually refactored after a few days (or hours). I need to re-enable it and fix the issues.

All in all: I know how to deal with (2)-(3), if I have the greenlight, I'll start working on this. I need your decision on (1).

#### Re external images
They are pulled in with [git submodules](https://www.atlassian.com/blog/git/git-submodules-workflows-tips). This is basically ...
repository-within-a-repository

The main points:
- You need to tell git explicitly to pull in the submodule repositories as well. In my client it's configured to automatically do the job. We probably need to extend the build code to make it work on CI as well.
- A submodule reference contains a specific revision number, you need to explicitly update it.

Soo .. it's quite similar to NuGet, but it's 100% git.

tocsoft:
i;m in the process of fixing the build atm

don't worry about releases.. i'm on it the new rename branch will release stuff to a new package name and to a different feed.

re the images put them in the https://github.com/SixLabors/TestImages repo and kill the branch that's already in there.

JimBobSquarePants:
Yeah, too slow I thought, I liked your qa code so went for it :smile:

I've already fixed the oil paint issue. png.... I'm hoping someone can easily fix the interlace bug as we had it working before. I was thinking we could get the hackfest pointing at this branch? 

Ah... That makes sense, yeah let's get that setup then.

antonfirsov:
@tocsoft I think you are aware of this, but I need to emphasize: if you have big images in the history, git is gonna pull them in the future regardless of being deleted (git is all about changes..)

So what I would do is to:
- Create an orphan branch in SixLabors/TestImages and copy the stuff here + delete the old `master`
- Rename `TestImages`: The actual input images ("TestImages") are still kept in the `ImageSharp` repo after my QA work. The purpose of the repository is actually "Various kinds of external images" (reference images, load test images etc.). That's why I named it `Imagesharp.Tests.Images`, but I'm open for better names.
- OR: go the simple way, and create a new repository.

Thoughts?

tocsoft:
seems like it'll be easier to just delete that repo and create a new one, just go for that instead 👍

tocsoft:
@SixLabors/core I believe I have the changes in to let the build servers run.

We now have 2 issues one seems to be just broken stuff the other relatively major,

1. The simpler one, we have some broken tests  6 that fail in VS and a lot of `ImageShouldApplyDiffusionFilterInBox` test that don't seem to fail in VS but do fail for me locally on the command line.
2. the potentially major one, we have issues with the way we are using `CoreCompat.System.Drawing` for the travis build, it doesn't like some of the image types https://travis-ci.org/SixLabors/ImageSharp/builds/266268165#L2469

antonfirsov:
@tocsoft 
1. No idea
2. 48bpp png issue: should be a bug in PNG decoder in mono System.Drawing. The simplest workaround would be to ignore them if we detect mono. Will think on ways to implement this.

+1: Should we keep the failing PNG tests failing, or shuld we `Skip` them for now?

JimBobSquarePants:
Skip for now and raise an issue.

antonfirsov:
The failing `DitherTests` indicated a bug. Skipped them for now together with the Png stuff. Opened #300 and #301.

antonfirsov:
Okkay guys, looks like I managed to green out everything.

Let's have a quick last-minute API Review! This is our last chance to introduce breaking changes.

@JimBobSquarePants a few comments on Dither API: 
- I think you should apply postfixes on Ditherer and ErrorDiffuser class names. We really should not have a class named just `Ordered` on our public API, even if it's nested deeply into other namespaces.
- Is it OK to have the ditherers in `SixLabors.ImageSharp.Dithering`? Shouldn't they belong to `SixLabors.ImageSharp.Processing.Dithering` instead? They are used by an ImageProcessor as well.

antonfirsov:
@JimBobSquarePants bad news:
I covered the progressive cases with reference images produced by ImageMagick, and discovered that the PDF.js decoder is actually significantly less accurate for progressive images. 
3 of our tests are failing now, because the difference is above the 0.5% threshold I defined.

JimBobSquarePants:
That's annoying re progressive jpeg but my thoughts on that are it will be the effect of accumulative IDCT error so something we can fix when we use floating point. 

Re dither, I'll revisit the class design, the bug may not be one, dithering operates on the pixels on the right side and below the original so that first pixel will likely be unchanged. There might be a bounds check bug like in oil pain so I'll check it out.

antonfirsov:
@JimBobSquarePants Re progressive Jpeg: 
I want to make sure this is the case. We need to define regression tests for the spectral data being produced by (PdfJs) `JpegDecoderCore.ParseStream()` as I described in #298. 

I want to use BitMiracle to produce reference data, because their output seems to be correct. As far as I understand,, we should expect *exact equality* at this point.

JimBobSquarePants:
Yeah, before IDCT I would expect everything to be exact. Hopefully the issue becomes obvious.  Good luck reading that code though!

antonfirsov:
@JimBobSquarePants bad news again:
Looks like the PdfJs port produces wrong results for progressive jpeg-s even before IDCT. Check out `SpectralJpegTests.CompareSpectralResults()` on the `jpeg-lab` branch. The verification data is produced by BitMiracle libjpeg.NET. I don't think the data is wrong, because the resulting image is much more similar to the `libjpeg-turbo` output.

JimBobSquarePants:
@antonfirsov I'll have a look. Fingers crossed it's something obvious to me. 

I still think the PdfJS port is still the more viable option for the beta as it stands despite progressive issues since it can handle all the broken images.

antonfirsov:
@JimBobSquarePants it's up to you to decide which decoder is better for users in it's *current* state :) We can switch them again or even combine them later.

However, I just realized, I can achieve the same low-memory two-pass behavior with the old one by reordering a few operations. Would be an easier start for me, if the spectral data is correct in the old one in the progressive case. 

For PDF.js I think we need to verify if the original javascript implementation is correct for progressive images. It's easy to do with `ImageComparer.Tolerant(0,0)`.

JimBobSquarePants:
I've just fixed the error diffusor so that test now passes.

I think `Ordered` is fine as a class name because that is a type of dither algorithm (See DITHER.TXT) in the solution.

I'll have a look at wiring up a webpage to allow us to save pdf.js images from for comparison.

tocsoft:
just added some API cleanup, cleaned up some namespaces, removed some stuff from the public API that's just no longer needed or was not being used.

tests/ImageSharp.Tests/Image/PixelAccessorTests.cs
```diff
@@ -42,7 +42,7 @@ public static Image<TPixel> CreateTestImage<TPixel>()
         [WithMemberFactory(nameof(CreateTestImage), PixelTypes.All, ComponentOrder.Zyx)]
         [WithMemberFactory(nameof(CreateTestImage), PixelTypes.All, ComponentOrder.Xyzw)]
         [WithMemberFactory(nameof(CreateTestImage), PixelTypes.All, ComponentOrder.Zyxw)]
-        public void CopyTo_Then_CopyFrom_OnFullImageRect<TPixel>(TestImageProvider<TPixel> provider, ComponentOrder order)
+        internal void CopyTo_Then_CopyFrom_OnFullImageRect<TPixel>(TestImageProvider<TPixel> provider, ComponentOrder order)
```
antonfirsov: @tocsoft are you sure `internal` is detected by xUnit?

tocsoft: They where showing up in my test explorer... just checked they seem to still be running on the build server

tests/ImageSharp.Sandbox46/ImageSharp.Sandbox46.csproj
```diff
@@ -20,6 +20,10 @@
     <PackageReference Include="BitMiracle.LibJpeg.NET" Version="1.4.280" />
     <PackageReference Include="xunit" Version="2.3.0-beta4-build3742" />
     <PackageReference Include="Moq" Version="4.7.99" />
+    <!--<PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />-->
+    <PackageReference Include="Moq" Version="4.7.99" />
+    <PackageReference Include="xunit" Version="2.3.0-beta4-build3742" />
```
antonfirsov: duplicate package references!!! :skull_and_crossbones:

JimBobSquarePants: I just his upgrade. Blame MS. :tongue:

JimBobSquarePants:
@tocsoft Is there any CI config stuff we need to do after we merge? I'd like to knock out a blog post to go with it also since the first beta is such a big deal. Also, git tagging, we should be doing that I think?

tocsoft:
There shouldn't be any extra CI stuff needed this branch has already been updated. What we can do is get it merged in get a version built it won't automatically create release build until we add a git tag/github release. 

We can draft up the blogpost (we should probably start drafting that up now.) then once we are all happy with the post we upload to nuget.org + publish the post.

antonfirsov:
please, lets do the span api cleanup before we merge!

see gitter for my suggestion.

antonfirsov:
I mean my newest proposal: hiding everything including .GetPixelReference(), exposing GetPixelSpan() as an extension method

antonfirsov:
Or if we lack the time to find consensus, lets just hide everything for now, I can PR my idea later.

tocsoft:
I vote we hide everything we currently have returning a Span internally for now but create the 2 new extension methods `.GetPixelSpan()` & `.GetPixelRowSpan(int row)` (this one is used extensively for creating processors) and put them into an ImageSharp.Unsafe namespace so you have to actively add the using statement. 

Additionally we need to ass `.SavePixelData(streamByteArrayOrSpan)` (to mimic `Image.LoadPixelData()` to save pixel data to a stream or byte array).

I'll work this up now.

tocsoft:
Done. `Span<TPixel>` has been removed to the easily accessible public API. 

To access the pixel data you have to use the extension methods exposed in the `SixLabors.ImageSharp.Advanced.Unsafe` namespace (the reason it is `Advanced.Unsafe` is because the Unsafe namespace by itself was clashing with the `Unsafe` helper class in from the framework.)

Also I've added `Image.SavePIxelData(Span<byte> buffer)`.

antonfirsov:
@tocsoft good job! The only problem I have with this is naming. The namespace should not be `*.Advanced.Unsafe` for two reasons:
- Span is not unsafe! It's exactly the opposite: overflow errors avoided by bounds checnking + the CoreCLR 2.0 is actually guaranteeing 100% thread safety by throwing an exception on a heap capture of a span.
- The the word `Unsafe` could still conflict with `SRCS.Unsafe` even if prefixed by `Advanced.`.

Let's just name it `SixLabors.ImageSharp.Advanced` or `SixLabors.ImageSharp.Dangerous`!

tocsoft:
source reasonable to me, I'll drop the Unsafe part and just leave it as Advanced.

antonfirsov:
:+1:  cool! I take back my second `*.Dangerous` proposal anyways, because the word dangerous means a different thing in MS terminology. `SixLabors.ImageSharp.Advanced` is the best!

antonfirsov:
Pushed my YCbCr SIMD optimization right into beta-1. Benchmarked with unit tests for now, the speedup is significant.
Before:
image
After:
image

JimBobSquarePants:
Holy Moly! 👍

JimBobSquarePants:
2x Faster than before.

```
BenchmarkDotNet=v0.10.9, OS=Windows 10 Redstone 2 (10.0.15063)
Processor=Intel Core i7-6600U CPU 2.60GHz (Skylake), ProcessorCount=4
Frequency=2742187 Hz, Resolution=364.6724 ns, Timer=TSC
.NET Core SDK=2.0.0
  [Host]     : .NET Core 1.1.2 (Framework 4.6.25211.01), 64bit RyuJIT
  DefaultJob : .NET Core 1.1.2 (Framework 4.6.25211.01), 64bit RyuJIT


            Method |     Mean |    Error |   StdDev | Allocated |
------------------ |---------:|---------:|---------:|----------:|
 'ImageSharp Jpeg' | 74.50 ms | 1.445 ms | 1.351 ms |   7.87 KB |
```

tocsoft:
woops, fixing the build now

src/ImageSharp/Image/ImageExtensions.cs
```diff
@@ -138,7 +138,7 @@ public static void Save<TPixel>(this Image<TPixel> source, Stream stream, IImage
         /// <param name="source">The source image</param>
         /// <param name="buffer">The buffer to save the raw pixel data to.</param>
         /// <exception cref="System.ArgumentNullException">Thrown if the stream is null.</exception>
-        public static void SavePixelData<TPixel>(this Image<TPixel> source, Span<byte> buffer)
+        internal static void SavePixelData<TPixel>(this Image<TPixel> source, byte[] buffer)
```
antonfirsov: We could probably expose a `SavePixelData(byte[] buffer, int offset)` variant publicly.

tocsoft: i've made it public in the Image-Frame branch

antonfirsov:
Made some further decoder optimazation by optimizing `PackFromVector4()`.
[My current results](https://gist.github.com/antonfirsov/7e3545ed77fa863a58a75d255483e086#gistcomment-2202944) are much better than yesterday. @JimBobSquarePants being keen of the results on your machine.

Issue #314, opened 2017-08-28
Image.Load() with IndexOutOfRangeException "Index was outside the bounds of the array."


xakep139:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

I try to load 8-bit colormap interlaced PNG image that displays normally. It loads normally with .NET Framework's `System.Drawing.Image.FromFile()` and also with ImageMagick.
This bug could be related to #301.

### Steps to Reproduce

Simple code:
```
Image.Load("banner.png");
```
with image [banner_index.zip](https://github.com/SixLabors/ImageSharp/files/1256118/banner_index.zip) cause an `System.IndexOutOfRangeException` with stacktrace:
```
   at ImageSharp.Formats.PngDecoderCore.ProcessInterlacedDefilteredScanline[TPixel](Byte[] defilteredScanline, Span`1 rowSpan, Int32 pixelOffset, Int32 increment)
   at ImageSharp.Formats.PngDecoderCore.DecodeInterlacedPixelData[TPixel](Stream compressedStream, Image`1 image)
   at ImageSharp.Formats.PngDecoderCore.Decode[TPixel](Stream stream)
   at ImageSharp.Image.Decode[TPixel](Stream stream, Configuration config)
   at ImageSharp.Image.WithSeekableStream[T](Stream stream, Func`2 action)
   at ImageSharp.Image.Load[TPixel](Configuration config, Stream stream, IImageFormat& format)
   at ImageSharp.Image.Load[TPixel](Configuration config, String path)
   at ImageSharp.Image.Load(String path)
   at ImageSharpTest.Program.Main(String[] args) in C:\Users\user\Documents\Visual Studio 2017\Projects\ImageSharpTest\Program.cs:line 12
```

### System Configuration

- ImageSharp version: 1.0.0-alpha9-00187
- Environment (Operating system, version and so on): Windows 10 Enterprise Version 1703 Build 15063.540
- Visual Studio Professional 2017 Version 15.3.2
- `dotnet --info` output:
```
.NET Command Line Tools (2.0.0)

Product Information:
 Version:            2.0.0
[...]

Issue #313, opened 2017-08-28
Image.Load() with System.ArgumentException "Not a valid pass index: 7"


xakep139:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

I try to load PNG image that displays normally. It also loads normally with .NET Framework's `System.Drawing.Image.FromFile()` and with ImageMagick.

### Steps to Reproduce

Simple code:
```
Image.Load("banner.png");
```
with image [banner_adam7.zip](https://github.com/SixLabors/ImageSharp/files/1255625/banner_adam7.zip) cause an `System.ArgumentException` with stacktrace:
```
   at ImageSharp.Formats.PngDecoderCore.ComputeColumnsAdam7(Int32 passIndex)
   at ImageSharp.Formats.PngDecoderCore.DecodeInterlacedPixelData[TPixel](Stream compressedStream, Image`1 image)
   at ImageSharp.Formats.PngDecoderCore.Decode[TPixel](Stream stream)
   at ImageSharp.Image.Decode[TPixel](Stream stream, Configuration config)
   at ImageSharp.Image.WithSeekableStream[T](Stream stream, Func`2 action)
   at ImageSharp.Image.Load[TPixel](Configuration config, Stream stream, IImageFormat& format)
   at ImageSharp.Image.Load[TPixel](Configuration config, String path)
   at ImageSharp.Image.Load(String path)
   at ImageSharpTest.Program.Main(String[] args) in C:\Users\user\Documents\Visual Studio 2017\Projects\ImageSharpTest\Program.cs:line 13
```

### System Configuration

- ImageSharp version: 1.0.0-alpha9-00187
- Environment (Operating system, version and so on): Windows 10 Enterprise Version 1703 Build 15063.540
- Visual Studio Professional 2017 Version 15.3.2
- `dotnet --info` output:
```
.NET Command Line Tools (2.0.0)

Product Information:
 Version:            2.0.0
 Commit SHA-1 hash:  cdcd1928c9

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
[...]

Issue #301, opened 2017-08-19
PNG decoder result is incorrect for certain images


antonfirsov:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
The decoder result has incorrect pixels with:
- [ ] Interlaced images (`TestImages.Png.Interlaced`)
- [ ] Certain other images from #296, like (`TestImages.Png.SnakeGame`)

### Steps to Reproduce
Uncomment commented test images, and unskip skipped tests in `PngDecoderTests` on the `beta-1` branch.

JimBobSquarePants:
Fix for snakegame in beta-1, other fixes in master

PR #315, opened 2017-08-28
Several fixes in the PNG decoder.

dlemstra:
### Prerequisites

### Description

This pull request fixes a couple of issue that occur when decoding a PNG file. With these patches the images from #313 and #314 can be read. 

I have not tested if the images from #301 can be read on the beta-1 branch. Could you check if these patches also fix the tests in the beta-1 branch @JimBobSquarePants?

It is probably also wise to investigate if we don't have the same issues in the PNG encoder. Anyone?


antonfirsov:
Shouldn't we merge this right into beta-1 and add new PNG tests together with the images from #313 #314?

JimBobSquarePants:
Depends how quickly we can get this beta out I guess, we can merge this in here and beta-1 and add the addition tests/images to beta-1

Issue #178, opened 2017-04-17
Error with Image.Load


KANekT:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode

### Steps to Reproduce
` 
var image = Image.Load(file.OpenReadStream());
`

I have error Too many coefficients 64 > 63 for [file/link](https://data.yougile.com/user-data/c989b56d-b0b7-4e1d-a048-c8f22ad0ec44/lemon.jpg)

`   at ImageSharp.Formats.Jpg.JpegScanDecoder.Refine(InputProcessor& bp, HuffmanTree& h, Int32 delta)
   at ImageSharp.Formats.Jpg.JpegScanDecoder.DecodeBlock(JpegDecoderCore decoder, Int32 scanIndex)
   at ImageSharp.Formats.Jpg.JpegScanDecoder.DecodeBlocks(JpegDecoderCore decoder)
   at ImageSharp.Formats.JpegDecoderCore.ProcessStartOfScan(Int32 remaining)
   at ImageSharp.Formats.JpegDecoderCore.ProcessStream(ImageMetaData metadata, Stream stream, Boolean metadataOnly)
   at ImageSharp.Formats.JpegDecoderCore.Decode[TColor](Stream stream)
   at ImageSharp.Formats.JpegDecoder.Decode[TColor](Configuration configuration, Stream stream, IDecoderOptions options)
   at ImageSharp.Image.Decode[TColor](Stream stream, IDecoderOptions options, Configuration config)
   at ImageSharp.Image.Load[TColor](Configuration config, Stream stream, IDecoderOptions options)
   at ImageSharp.Image.Load(Stream stream)
   at Biggab.Game.Education.Backend.ViewModels.Words.CreateOrEditViewModelFactory.<SaveFile>d__5.MoveNext() in C:\Projects\BiggabSocial\src\Biggab.Game.Education.Backend\Areas\Backend\ViewModels\Words\CreateOrEditViewModelFactory.cs:line 99`
### System Configuration

- ImageSharp version: 1.0.0-alpha9-00025
[...]

PR #322, opened 2017-09-02
Reimplemented Jpeg postprocessor logic + fixed bugs in stream parsing

antonfirsov:
### Prerequisites

### Description
Should fix stability and memory issues #224 #214 #178 #159 #151 and provide better accuracy (#245 #155) without significant performance drop. For some images it produces almost identical results to libjpeg-turbo (run `CompareJpegDecoders_Baseline()` for comparison results).


src/ImageSharp/Formats/Jpeg/Common/FastFloatingPointDCT.cs
```diff
@@ -3,47 +3,47 @@
...
     /// <summary>
-    /// Contains forward and inverse DCT implementations
+    /// Contains inaccurate, but fast forward and inverse DCT implementations.
     /// </summary>
-    internal static class DCT
+    internal static class FastFloatingPointDCT
     {
 #pragma warning disable SA1310 // FieldNamesMustNotContainUnderscore
-        private static readonly float C_1_175876 = 1.175876f;
+        private static readonly float C_1_175876 = 1.175875602f;
```
tocsoft: wouldn't theses be better as `const` values instead of `readonly static` ones?

antonfirsov: Wow, worked with this code a lot but haven't spotted this all the time! I think this is because these values were originally defined as `Vector4`.

JimBobSquarePants: Yeah, we can fix that as we optimize post beta.

JimBobSquarePants:
Very happy with this as is. Well done @antonfirsov on this and in getting real tests in there.

 Will help out post beta to introduce more constants into the decoding process. `E X I F`, `A d  o b e` etc and any optimization I can get my head around.

JimBobSquarePants:
```
BenchmarkDotNet=v0.10.9, OS=Windows 10 Redstone 2 (10.0.15063)
Processor=Intel Core i7-6600U CPU 2.60GHz (Skylake), ProcessorCount=4
Frequency=2742191 Hz, Resolution=364.6719 ns, Timer=TSC
.NET Core SDK=2.0.0
  [Host]     : .NET Core 1.1.2 (Framework 4.6.25211.01), 64bit RyuJIT
  DefaultJob : .NET Core 1.1.2 (Framework 4.6.25211.01), 64bit RyuJIT


            Method |     Mean |    Error |   StdDev | Allocated |
------------------ |---------:|---------:|---------:|----------:|
 'ImageSharp Jpeg' | 143.0 ms | 2.778 ms | 3.307 ms |   7.87 KB |
```

Just leaving the above so we have a known reference point to optimize from.

Issue #284, opened 2017-07-29
Allow encoding directly from frames


PazerOP:
It'd be nice to have `IImageEncoder` also have the overload
```C# 
void Encode<TPixel>(ImageBase<TPixel> imageBase, Stream stream) where TPixel : struct, IPixel<TPixel>
```

That way, you can directly save frames of an image as separate images without having to clone them as you have to now:

```C#
var img = Image.Load(source);

foreach (var frame in img.Frames.Select((f, i) => (f, i)))
{
	var frameImg = new Image<Rgba32>(frame.Item1);

	var outputFile = File.Create($"frames/{frame.Item2}.png");
	frameImg.SaveAsPng(outputFile);
}
```

JimBobSquarePants:
Your wish is granted! :smile:

The latest builds on MyGet now allow exporting of frames 

``` c#
using( var img = Image.Load(source))
{
    foreach (var frame in img.Frames.Select((f, i) => (f, i)))
    {
        using (Image<TPixel> cloned = img.Frames.ExportFrame(i))
        {
         // Save cloned frame
        }
    }
}
```

antonfirsov:
@JimBobSquarePants what does `.Select((f, i) => (f, i))` do in your sample code? Or is it just a typo?

JimBobSquarePants:
I just copy/pasted the surrounding code above to be honest. It doesn't do anything now I look at it.

PR #326, opened 2017-09-09
 API Image frames refactor 

tocsoft:
### Prerequisites

### Description

This PR has a few API changed and a significate internal refactor of where pixel data is stored.

1. refactors the internals of where image data is stored such that all pixel data is now help within `ImageFrame`s, every Image will always have at least one image frame (unless its been disposed). 
2.  hides accesses Span<TPixel> apis into the `SixLabor.ImageSharp.Advanced` namespace as these APIs are "Use at your own risk".
3. Adds the missing `SavePixelData()` apis to pair with the `LoadPixelData` apis to access the raw byte array holding the image pixels data.


tocsoft:
I think it would be almost impossible at the moment to make that work for the current version... we would have to make a heap more of the internals public. 

I feel we should leave it sealed for v1, we can always relax things as people hit the limits of our current implementation, I feel moving over to an interface based version of the API would end you being the better choice than just unsealing `Image`.

JimBobSquarePants:
@tocsoft I can live with that for now.

antonfirsov:
@JimBobSquarePants I think we should keep our stuff sealed. People should  follow practices like [composition over inheritance](https://en.m.wikipedia.org/wiki/Composition_over_inheritance) for these cases like integral images.

JimBobSquarePants:
Just noticed this issue, can we handle this better now?

https://github.com/SixLabors/ImageSharp/issues/284

tocsoft:
sort of we would still need and `Image<T>`  to encode from so we could create a temp Image<T> with just the one frame (which due to the fact we store all data in frames now be a relatively cheep operation) and then just encode that... would need to think about the best way though... would probably be better to have configuration on the encoders to let use specify which FrameIndex to treat as the source of data to encode (excluding gif of course)

src/ImageSharp/Image/ImageFrameCollection.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        /// <summary>
+        /// Gets the count.
+        /// </summary>
+        public int Count => this.frames.Count;
+
+        /// <summary>
+        /// Gets the root frame.
+        /// </summary>
+        public ImageFrame<TPixel> RootFrame => this.frames.Count > 0 ? this.frames[0] : null;
```
antonfirsov: Is it valid to have 0 frames? Shouldn't we throw in the constructor instead?

src/ImageSharp/Image/ImageFrameCollection.cs
```diff
@@ -0,0 +1,166 @@
...
+        /// <param name="frame">The <seealso cref="ImageFrame{TPixel}"/> to locate in the <seealso cref="Image{TPixel}"/>.</param>
+        /// <returns>The index of item if found in the list; otherwise, -1.</returns>
+        public int IndexOf(ImageFrame<TPixel> frame) => this.frames.IndexOf(frame);
+
+        /// <summary>
+        ///  Inserts the <paramref name="frame"/> to the <seealso cref="Image{TPixel}"/> at the specified <paramref name="index"/>.
+        /// </summary>
+        /// <param name="index"> The zero-based index at which item should be inserted..</param>
+        /// <param name="frame">The <seealso cref="ImageFrame{TPixel}"/> to insert into the <seealso cref="Image{TPixel}"/>.</param>
+        public void Insert(int index, ImageFrame<TPixel> frame)
```
antonfirsov: Is it really a good idea to expose a mutable List API on `ImageFrameCollection`? Feels like the API surface is to wide now.
I'd rather keep it as an immuatble collection. If someone wants to implement a gif frame manipulator tool with ImageSharp (looks like a very uncommon use case), he should probably construct new image instances after calculating the frames.

antonfirsov:
Everything LGTM except the mutable `ImageFrameCollection` API. It's much easier to maintain a consistent `Image<T>` state if we close down that list.

src/ImageSharp/Image/ImageFrameCollection.cs
tocsoft: yes you can have zero frames, it will have zero when the ImageFrameCollection as been disposed.

src/ImageSharp/Image/ImageFrameCollection.cs
tocsoft: if its immutable then you can't create an animated gif.. also the public api doesn't allow users to specify frames so the only way for a user to go from zero to animated gif is to create a series of images, merge their frames together into a single image instance and then save that.

src/ImageSharp/Image/ImageFrameCollection.cs
antonfirsov: Q1: Shouldn't we throw `ObjectDisposedException` instead of returning null instead?
Q2: Looks like it's possible to dispose an `ImageFrameCollection` without disposing the underlying image bringing `Image<T>` into an invalid state. Shouldn't we drop `IDisposable` from `ImageFrameCollection<T>` API and control it directly from `Image<T>`?

src/ImageSharp/Image/ImageFrameCollection.cs
antonfirsov: It's probably matter of API taste, but I would utilize [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) instead, having some kind of `ImageBuilder` object to construct multi-frame images. This looks like a more convenient way to preserve valid object state. Constructing `ImageFrame` instances of a certain size, and verifying it against the image dimensions when calling `ImageFrameCollection.Add()` looks a bit clanky to me.

But looks like ImageSharp API follows mutable style on many other places (like `ImageMetaData`) so maybe it's easier to conform to this style everywhere for 1.0.

How about encapsulating the consistance rules into `ImageFrameCollection` by hiding `ImageFrame<T>` constructors and define methods on `ImageFrameCollection<T>`:
```C#
public ImageFrame<T> AppendNewFrame() { }
public ImageFrame<T> InsertNewFrame(int index) { }
```
?

@JimBobSquarePants thoughts?

src/ImageSharp/Image/ImageFrameCollection.cs
tocsoft: 1. yes we should probably throw an `ObjectDisposedException`.

2. I don;t think its really an issue that you can dispose of the ImageFrameCollection without disposing of the image. You can do the same for ImageFrame too.  I suppose we could have the Image expose and interface of `IImageFrameCollection` that doesn't support `IDisposeable` then the users can't call dispose of the collection separately to the Image.

src/ImageSharp/Image/ImageFrameCollection.cs
tocsoft: I had thought about adding those as APIs (well in concept not name, couldn't think of what to call them) but I like it.

This makes me think we should find a way to extend the `Mutate()` apis to allow us to apply mutations to single frames... might need to extend expand `IImageProcessors`, as we will need  `IFrameProcessor`s as well as `IImageProcessor`s will also have to scope some of the APIs to allow targeting only images and not frames (resize, crop, rotate etc) unless we allow images to have frames with multiple dimensions and handle normalising frame sizes on a per encoder bases instead of during add frame (i.e. update gif to draw based on the first frame dimensions)

JimBobSquarePants: @antonfirsov I see where you're going there and it does seem safer. As far as I can see we don't actually new up a frame anywhere within our codebase.

@tocsoft I'd rather we treated images as a whole and avoid adding `IFrameProcessor`'s. Seems like we'd be creating a complexity level that would be difficult to follow.

tocsoft: I think we should leave it as is for now then... we can worry about that later if/when people hit issues with the API.

antonfirsov:
We are really short on time, so OK for now.
In long term I'm planning to suggest deep changes to the frame API, because I'm not happy with the  "greatest common divisor" approach we used to pick the gif model as an universal model for `Image<TPixel>`.
Hope we can go for an early 2.0 ...

tocsoft:
ok merging this.. we can continue any final tweeks on the beta branch

PR #329, opened 2017-09-15
Expose advanced pixel reference API.

tocsoft:
Exposes an extension method for exposing a pinnable `ref` to the start of the pixel buffer.

To access this extensions can be accessed by adding `using SixLabors.ImageSharp.Advanced;`


tocsoft:
@mellinoe can you confirm that if we added this extension method to the public API it would unblock you? 

Basically if we can give you a `ref` to pixel 0,0 that will let you do the unsafe stuff you need to do to get to the pixel data you want?

antonfirsov:
@tocsoft I don't think `GetPixelReference(x, y)` is useful as a general pixel manipulator API. It's actually a false promise to users who want fast pixel access, because it's still applying the `y*Width` multiplication for each call. I think we should remove this method to have a cleaner API in long term. People who need fast pixel access in safe ways should use the `Span<T>` methods in the future.

What we need *now* is a simple method for users who want to go unsafe and do pointer arithmetic with `SRCS.Unsafe()`. I think this is what @mellinoe needs. For this purpose having something like a `.DangerousGetPinnableReferenceToPixelBuffer()` extension method is much cleaner. (Less verbose name proposal: `GetReferenceToFirstPixel()`.)

tocsoft:
@antonfirsov its not about fast pixel access its for getting `ref` pixel access

antonfirsov:
@tocsoft but why do you need `ref` access to *all/random* pixels? I can't imagine any real-life use case utilizing `ref` locals for any other reason than performance.

tocsoft:
no idea... if you think its overkill then feel free to change it.. I was just trying to prevent us adding a new API then later having to add this version (that accesses any pixel) which could be used interchangeably for the older one which we'd now be stuck with.

mellinoe:
The key thing needed is a pinnable reference to the beginning of the pixel data, for use with PInvokes. Without it, I'm not able to upload the pixel data to OpenGL/Direct3D/Vulkan/etc. without copying out everything into an auxiliary buffer.

Essentially, I need something that would satisfy this test:

```C#
public void CopyToOtherBuffer(Image<Rgba32> image, IntPtr otherBuffer, int size)
{
    Rgba32* pixelStart = PinImagePixelsSomehow(); // This is the function needed
    int imageSizeInBytes = image.Width * image.Height * sizeof(Rgba32);
    Buffer.MemoryCopy(pixelStart, otherBuffer.ToPointer(), size, imageSizeInBytes);
    // otherBuffer should now contain the raw pixel data from image.
}
```

mellinoe:
Basically, the current version would work fine, but so would a `GetReferenceToFirstPixel` method -- I wouldn't need to get a reference to some other pixel generally speaking. If I did, I could just do some pointer arithmetic myself. If you are more comfortable exposing one or the other, then that is fine with me.

tocsoft:
@mellinoe thanks for that confirmation.
@antonfirsov I've been mulling on this today and I think I've decided I agree with you and `image.DangerousGetPinnableReferenceToPixelBuffer()` is the way to go

tocsoft:
_for future me when i can't remember why i decided this_ 😀 
Went for this name as it aligns with the name on `Span<T>` which we are ultimately calling.

mellinoe:
@tocsoft I think the class itself also needs to be made public.

tocsoft:
@mellinoe fixed... thanks for the spot.

mellinoe:
This is up to you of course, but it may be worth adding a small test case for it, roughly like what I wrote above. It could create some native memory (`Marshal.AllocHGlobal`), call `CopyToOtherBuffer` (above), and then assert that all the pixels are the same, in order. (Then free the native memory :smile:) That's roughly the use case that I'm interested in anyways.

antonfirsov:
Modified + extended the docs on `DangerousGetPinnableReferenceToPixelBuffer` giving the user some info on the underlying buffer.

> Such a reference can be used for pinning but must never be dereferenced.
This would be only true, if the image had 0 pixels, which is not allowed.

Everything looks good to me now, but agree with @mellinoe, a test might be useful. I think it's totally fine to use `Unsafe.CopyBlock()` + an array in the tests, no need to create an unsafe buffer.

Should we add the test in this PR?

tocsoft:
@antonfirsov yes we should, can you do me a favour and write the test for me please 😃. Your our resident expert on these things (can't get my head around how to convert the `ref TPixel` into something Unsafe.CopyBlock can use.) 

_I have a feeling i'm going to feel pretty stupid after I see the test in seeing how obvious the answer is_

mellinoe:
https://github.com/mellinoe/ImageSharp/commit/e82f57bc5b375d2b7cabb00998c297a603acf6c9

Feel free to cherry-pick it or just copy-paste -- whichever is more convenient.

tocsoft:
@mellinoe I've pulled in your commit.. thanks for that your a star can I ask you to please sign our CLA?

mellinoe:
Done

antonfirsov:
Thanks @mellinoe!

@tocsoft I think we need to expose a `TPixel[]` equivalent of [this method](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Image/ImageExtensions.cs#L132). Would be wise to add it to this PR.

Gonna do it today, and I think we can merge this afterwards.

antonfirsov:
Wait, I found a serious issue!
There would be a name collision for users using both:
`SixLabors.ImageSharp.ImageExtensions`
`SixLabors.ImageSharp.Advanced.ImageExtensions`

Let's rename the second one! What should it be? `AdvancedImageExtensions`?

tocsoft:
`AdvancedImageExtensions` sounds fine to me.

You might as well make the change while your adding the `TPixel[]` methods.

antonfirsov:
Changes in progress ... will also polish the test cases a bit.

JimBobSquarePants:
Naming is hard 😄

antonfirsov:
Done!

I know we are in beta now, but shouldn't we rename [this method](https://github.com/SixLabors/ImageSharp/blob/63fa1fdbe1c44fb72a96ceb3bde316637da97528/src/ImageSharp/Image/ImageExtensions.cs#L121)?

I think that the zero-argument `SavePixelData()` overload should return `TPixel[]` instead of `byte[]`!

tocsoft:
It depends on how most people want the raw data.. I suspect most people will want the raw data as a byte[] so they can either save it or integrate with some other system that won't understand our pixel format but will usually support byte[]'s

tocsoft:
I've never seen anyone asking for an array TPixel[] that disconnected from the pixel buffer, but I have seen requests for a way to get the byte[]

antonfirsov:
You might be right, let's keep the current names then! 

Everything LGTM for me now.

PR #337, opened 2017-09-17
Accuracy bugfix + further optimizations on JpegImagePostprocessor

antonfirsov:
### Prerequisites

### Description
- Fixed an accuracy bug in `JpegColorConverter.FromYCbCrSimd`: for libjpeg compatibility we need to *actually* round the values before downscaling them.
- Improved `JpegDecoder` regression testing by specifying the expected similarity per-image
- Implemented additional AVX2-specific optimizations [improving `JpegImagePostProcessor` performance](https://gist.github.com/antonfirsov/7e3545ed77fa863a58a75d255483e086#gistcomment-2205292).
We are starting to reach the limitations of the current .NET SIMD API, It's seems really hard to optimize the standard YCbCr color path any further. We should focus on the stream parser (Huffman decoder) and on other color models from now.


src/ImageSharp/Formats/Jpeg/Common/Decoder/JpegBlockPostProcessor.cs
```diff
@@ -11,145 +11,80 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Common.Decoder
...
-            /// </summary>
-            public UnzigData Unzig;
-
-            /// <summary>
-            /// Creates and initializes a new <see cref="ComputationData"/> instance
-            /// </summary>
-            /// <returns>The <see cref="ComputationData"/></returns>
-            public static ComputationData Create()
+            // To be "more accurate", we need to emulate this by rounding!
+            if (SimdUtils.IsAvx2CompatibleArchitecture)
```
tocsoft: for readability shouldn't this be a single call to  `this.WorkspaceBlock1.NormalizeColorsAndRoundInplace();` with the if check inside the callled method.

src/ImageSharp/Formats/Jpeg/Common/Decoder/JpegColorConverter.FromYCbCr.cs
```diff
@@ -121,63 +122,32 @@ internal static void ConvertCore(ComponentValues values, Span<Vector4> result)
                     tmp.MultiplyInplace(1.772F);
                     b.AddInplace(ref tmp);
 
-                    r.RoundAndDownscale();
-                    g.RoundAndDownscale();
-                    b.RoundAndDownscale();
+                    if (Vector<float>.Count == 4)
+                    {
+                        // TODO: Find a way to properly run & test this path on AVX2 PC-s! (Have I already mentioned that Vector<T> is terrible?)
+                        r.RoundAndDownscalePreAvx2();
```
tocsoft: again I feel we should have a single call here called `.RoundAndDownscale()` that independently has the checks in. As long as `SimdUtils.IsAvx2CompatibleArchitecture` is a const then the jitter will remove the dead code for us but it will be much more readable as the callsite .

tocsoft: But I feel that `SimdUtils.IsAvx2CompatibleArchitecture` would have to be an aggressively inlined getter only property for the jitter to follow the fact that the count is a jitter time constant and cause the dead code to be eliminated.

tocsoft: I could be totally wrong and it's not the case, if its not then feel free to ignore.

antonfirsov: There are 2 solutions:
1. a static method taking 3 tuples by ref.
2. Defining a new struct for 3 tuples. 

Not sure if its worth to bother with this at the moment.

src/ImageSharp/Formats/Jpeg/Common/Decoder/JpegBlockPostProcessor.cs
antonfirsov: Agree, will do it.

src/ImageSharp/Formats/Jpeg/Common/Decoder/JpegColorConverter.FromYCbCr.cs
tocsoft: If you don't think its worth the effort then feel free to leave it for now then.

antonfirsov: +Yeah the jitter tends to eliminate dead code afaik.

src/ImageSharp/Formats/Jpeg/Common/Block8x8F.CopyTo.cs
```diff
@@ -0,0 +1,146 @@
...
+            Unsafe.Add(ref Unsafe.Add(ref destLocalOrigo, destStride + 8), 2) = selfRight.Y;
+            Unsafe.Add(ref Unsafe.Add(ref destLocalOrigo, destStride + 8), 3) = selfRight.Y;
+            Unsafe.Add(ref Unsafe.Add(ref destLocalOrigo, destStride + 8), 4) = selfRight.Z;
+            Unsafe.Add(ref Unsafe.Add(ref destLocalOrigo, destStride + 8), 5) = selfRight.Z;
+            Unsafe.Add(ref Unsafe.Add(ref destLocalOrigo, destStride + 8), 6) = selfRight.W;
+            Unsafe.Add(ref Unsafe.Add(ref destLocalOrigo, destStride + 8), 7) = selfRight.W;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static void WidenCopyImpl(ref Vector4 s, ref float destBase)
```
JimBobSquarePants: This isn't currently being used?

src/ImageSharp/Formats/Jpeg/Common/Block8x8F.Generated.cs
```diff
@@ -96,51 +96,139 @@ public void TransposeInto(ref Block8x8F d)
...
-
         /// <summary>
-        /// Level shift by +128, clip to [0, 255]
+        /// AVX2-only variant for executing <see cref="NormalizeColorsInplace"/> and <see cref="RoundInplace"/> in one step.
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal void NormalizeColorsInplace()
+        public void NormalizeColorsAndRoundInplaceAvx2()
+        {
+            Vector<float> off = new Vector<float>(128f);
```
JimBobSquarePants: I know this doesn't appear to speed things up in the benchmarks but I think these should probably be static. It would reduce memory pressure and every little helps.

antonfirsov: @JimBobSquarePants I don't think there is a difference in this case.
A non-static, non-virtual method acts likes a static method taking `this` as the first argument. You should turn a method to `static` when you realize that you actually do not use `this` inside it. This is not the case here. (Or I'm just missing something?)

JimBobSquarePants: Sorry, I mean the `Vector<T>` instance. I've seen similar elsewhere with `Vector4` also.

src/ImageSharp/Formats/Jpeg/Common/Block8x8F.CopyTo.cs
antonfirsov: I got it inlined by into the methods above with ReSharper while trying to optimize stuff. Unfortunately it made only a very tiny difference. 

I want to leave it here for now for documentation purposes / future use.

src/ImageSharp/Formats/Jpeg/Common/Block8x8F.Generated.cs
antonfirsov: The behaviour of JIT intrinsics is quite different to the behaviour we used to with normal data types / instructions.
- `new Vector<float>(128f)` will translate into a single CPU instruction initializing an AVX register from a constant
- `new Vector4(128f)` will do the same but with an SSE register AFAIK

If you make it static it will be ... well, actually what you gonna see depends on the quality of the JIT code generator. By my experience, for `Vector4` it tends to generate optimized code, but I never trust what it's doing with `Vector<T>`.

TL;DR: The simplest way to make sure that optimized code will be generated is to new it up, like I did 😄

src/ImageSharp/Formats/Jpeg/Common/Block8x8F.CopyTo.cs
JimBobSquarePants: No worries.

src/ImageSharp/Formats/Jpeg/Common/Block8x8F.Generated.cs
JimBobSquarePants: Ok, lets leave it as it is just now then. I'll let you do the merge 👍

PR #342, opened 2017-09-19
Fix for Little/BigEndianBitConverter.ToInt64 reading wrong value in some cases

JBildstein:
### Prerequisites

### Description
The ToInt64 of the Little- and BigEndianConverter classes read wrong values when the least significant half of the number was negative. A negative 32bit integer extended to 64bit is filled up with 1s. Then when combining both halfs of the number with an OR, the other half is effectively ignored. By zeroing out those filled up 1s the bug gets fixed.


JimBobSquarePants:
Thanks @JBildstein Appreciate it!

PR #344, opened 2017-09-19
Add Extract/Clone apis for manipulating deconstructing images from from

### LINKED ISSUE #340: Clone/Extract frames from image
Provide 2 new public methods on `Image<TPixel>` called `CloneFrame` and `ExtractFrame`.

### Methods signature
`public Image<T> CloneFrame(int frameIndex);`
### Expected behavior
A new `Image<TPixel>` instance is created with a clone of the frame at index `frameIndex` as its `RootFrame`.
Throws `IndexOutOfRangeException` if `frameIndex` >= `ImageFrameCollection.Count` or `frameIndex` < `0`
The cloned `Image<TPixel>` should have the same metadata as the source image

### Methods signature
`public Image<T> ExtractFrame(int frameIndex);`
### Expected behavior
A new `Image<TPixel>` instance is created with the actual frame at index `frameIndex` as its `RootFrame`. _this is an instance move not a clone operation._
The frame from the source image will be removed from the source images `ImageFrameCollection`
Throws `IndexOutOfRangeException` if `frameIndex` >=` ImageFrameCollection.Count` or `frameIndex` < `0`
Throws `InvalidOperationException` if the user is trying to extract the last frame from the collection.
The cloned `Image<TPixel>` should have the same metadata as the source image.

tocsoft:
### Prerequisites

### Description

This adds the new APIs to `Image<TPixel>` for cloning and extracting frames from an image.

New APIs
- `Image<TPixel>.Clone(int frameIndex)`
- `Image<TPixel>.CloneAs<TPixel2>(int frameIndex)`
- `Image<TPixel>.Extract(int frameIndex)`

### TODO 

- [x] **Add Tests**
- [x] **Resolve API location** 
   should these APIs be members of `Image<T>` of `ImageFrameCollection`? 
   i.e. do you call `image.Clone(index)` or `image.Frames.CloneFrame(index)`
- [x] **Resolve API names**
   should the apis have the `Frame` postfix? _I feel this depended on if the API lives on the frames collection or the image_

resolves #340


JimBobSquarePants:
The Frame suffix seems nice and explicit to me but I don't think that the APIs should live in the ImageFrameCollection. (Don't ask me why it just doesn't feel right)

antonfirsov:
@JimBobSquarePants I think you have this feeling because all of them are returning an `Image<T>` instance. This makes them more bound to `Image<T>` than `ImageFrameCollection<T>` where everything operates on frame instances only.

JimBobSquarePants:
@antonfirsov Yeah, that describes it well.

antonfirsov:
@tocsoft All LGTM. Like that we manage a consistent state for frames through our API now.

We probably need a gif manipulator example, because the behavior of some functions (eg. `AddFrame()`) is not trivial now.

PR #346, opened 2017-09-21
AVX2 optimization for Jpeg YCbCr color conversion

antonfirsov:
### Prerequisites

### Description
- Managed to fight `Vector<T>` and now we have a special `JpegColorConverter.FromYCbCrSimdAvx2` execution path for AVX2-enabled (Haswell+) CPU-s with `Vector<float>.Count == 8`.
- The "old" `JpegColorConverter.FromYCbCrSimd` is capable of handling the `Vector<float>.Count == 4` path even when `Vector.IsHardwareAccelerated == false`. It's covered with it's own unit test.
- All converters were moved to the `ColorConverters` subfolder.

### Benchmarks

`JpegColorConverter.FromYCbCrSimdAvx2` is much faster on CPU-s supporting it:
```
      Method |       Mean |     Error |    StdDev | Scaled | Allocated |
------------ |-----------:|----------:|----------:|-------:|----------:|
      Scalar | 9,396.9 ns | 696.05 ns | 39.328 ns |   1.00 |       0 B |
 SimdVector4 | 1,687.6 ns |  77.52 ns |  4.380 ns |   0.18 |       0 B |
    SimdAvx2 |   410.9 ns | 116.94 ns |  6.608 ns |   0.04 |       0 B |
```

The [total decoder performance improvement](https://gist.github.com/antonfirsov/7e3545ed77fa863a58a75d255483e086#gistcomment-2208993) is also significant. The postprocessing step takes now less than `50%` of the full decoder CPU load.

### DIFF
--- a/src/ImageSharp/Common/Tuples/Vector4Pair.cs
+++ b/src/ImageSharp/Common/Tuples/Vector4Pair.cs
@@ -71,5 +71,10 @@ namespace SixLabors.ImageSharp.Common.Tuples
             v *= new Vector<float>(1 / 255f);
             self = v;
         }
+
+        public override string ToString()
+        {
+            return $"{this.A}, {this.B}";
+        }
     }
 }
--- a/src/ImageSharp/Formats/Jpeg/Common/Block8x8.cs
+++ b/src/ImageSharp/Formats/Jpeg/Common/Block8x8.cs
@@ -1,3 +1,6 @@
+// Copyright (c) Six Labors and contributors.
+// Licensed under the Apache License, Version 2.0.
+
 using System;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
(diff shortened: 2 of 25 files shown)


JimBobSquarePants:
TIL. `8388608` is the number of bits in a megabyte. How the number works in `FastRound` I still have no idea.

antonfirsov:
@JimBobSquarePants some good points to start:
- [An online toy](https://www.h-schmidt.net/FloatConverter/IEEE754.html) to with the bits of an IEEE-754 single precision float
- [The matching Wikipedia article](https://en.wikipedia.org/wiki/Single-precision_floating-point_format)

TBH I haven't took the time to understand `FastRound` yet. For me, the only way is to take a pen, a paper + at least a half free hour to play with the bits.

PR #348, opened 2017-09-21
WIP - Js/cleanup jpeg

JimBobSquarePants:
### Prerequisites

### Description
This PR aims to make the jpeg decoder methods more readable and consistent with other parts of the codebase.

I'm introducing byte marker constants for JFIF, EXIF etc and am planning on simplifying the file marker discovery code. If possible I'd like to use the raw input stream for reading out with the ScanDecoder to reduce overheads.

<strike>If anyone can spot why I'm suddenly allocating 5MB on my machine during the Jpeg decoder benchmarks please let me know. I've done nothing I can see which could possibly cause this.</strike>

Turns out the allocations were already there in master!


antonfirsov:
@JimBobSquarePants just noticed your update. What test have you been running to spot the allocations?

tests/ImageSharp.Tests/FileTestBase.cs
```diff
@@ -70,17 +70,17 @@ public static class Extensions
...
-            // TestFile.Create(TestImages.Jpeg.Baseline.Cmyk), // Perf: Enable for local testing only
-            // TestFile.Create(TestImages.Jpeg.Baseline.Floorplan), // Perf: Enable for local testing only
-            // TestFile.Create(TestImages.Jpeg.Progressive.Festzug), // Perf: Enable for local testing only
-            // TestFile.Create(TestImages.Jpeg.Baseline.Bad.MissingEOF), // Perf: Enable for local testing only
-            // TestFile.Create(TestImages.Jpeg.Baseline.Bad.ExifUndefType), // Perf: Enable for local testing only
-            // TestFile.Create(TestImages.Jpeg.Progressive.Fb), // Perf: Enable for local testing only
-            // TestFile.Create(TestImages.Jpeg.Progressive.Progress), // Perf: Enable for local testing only
-            // TestFile.Create(TestImages.Jpeg.Baseline.GammaDalaiLamaGray), // Perf: Enable for local testing only
-            // TestFile.Create(TestImages.Jpeg.Progressive.Bad.BadEOF), // Perf: Enable for local testing only
+             //TestFile.Create(TestImages.Jpeg.Baseline.Turtle), // Perf: Enable for local testing only
```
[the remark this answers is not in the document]
JimBobSquarePants: Some of them (the non-buggy ones) I would say we could get rid of.

JimBobSquarePants:
@antonfirsov I ran the benchmarks on master and my branch. The allocations are now present in both.

src/ImageSharp/Formats/Jpeg/GolangPort/OrigJpegDecoderCore.cs
```diff
@@ -529,11 +496,20 @@ private void ProcessApp2Marker(int remaining)
...
             byte[] identifier = new byte[Icclength];
             this.InputProcessor.ReadFull(identifier, 0, Icclength);
-            remaining -= Icclength; // we have read it by this point
-
-            if (identifier[0] == 'I' && identifier[1] == 'C' && identifier[2] == 'C' && identifier[3] == '_'
-                && identifier[4] == 'P' && identifier[5] == 'R' && identifier[6] == 'O' && identifier[7] == 'F'
-                && identifier[8] == 'I' && identifier[9] == 'L' && identifier[10] == 'E' && identifier[11] == '\0')
+            remaining -= Icclength; // We have read it by this point
+
+            if (identifier[0] == OrigJpegConstants.ICC.I &&
```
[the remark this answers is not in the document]
JimBobSquarePants: I like that a lot!

antonfirsov:
But they weren't there in my previous benchmark, and [the regression is quite dramatic](https://gist.github.com/antonfirsov/7e3545ed77fa863a58a75d255483e086#gistcomment-2209986).

It must be introduced by #344 then. @tocsoft any idea?

JimBobSquarePants:
@antonfirsov Yeah, It had me worried I'd done something. Will have a look at that other branch. Maybe we're not cleaning up properly.

src/ImageSharp/Formats/Jpeg/GolangPort/OrigJpegDecoderCore.cs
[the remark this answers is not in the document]
JimBobSquarePants: Bugger... Turns out the equality operators do not compare the sequence. https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Span.cs#L308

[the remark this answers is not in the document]
antonfirsov: Oops. Hope there is an easy to use helper somewhere around. AFAIK Span is used to operate on string slices in Kestrel.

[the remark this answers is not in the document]
JimBobSquarePants: Span has a specialized `SequenceEquals` for bytes. Was gonna use it but now can't get `ASCII` encoding on 1.1

src/ImageSharp/Formats/Jpeg/GolangPort/OrigJpegConstants.cs
```diff
@@ -26,64 +26,32 @@ internal static class OrigJpegConstants
...
             /// </summary>
-            public const byte Cb = 2;
+            public static readonly byte[] JFifMarker = { JFif.J, JFif.F, JFif.I, JFif.F, JFif.Null };
 
             /// <summary>
-            /// The YCbCr chroma component id.
+            /// Describes the EXIF specific markers
             /// </summary>
-            public const byte Cr = 3;
+            public static readonly byte[] IccMarker = { ICC.I, ICC.C, ICC.C, ICC.UnderScore, ICC.P, ICC.R, ICC.O, ICC.F, ICC.I, ICC.L, ICC.E, ICC.Null };
```
antonfirsov: I always forget, everything is NuGet package now. It should be very simple to get the ASCII bytes for these characters though:
```C#
static byte[] ToAsciiBytes(string str) => str.Select(c => (byte) c).ToArray();
```

I can add these changes this in the weekend.

JimBobSquarePants: Yeah, thought just grabbing the package would be easy and I rushed a workaround so fixing it would be great! 

Only thing I'd like to do now is split the `ParseStream` method into two methods separating out the marker searching and the switch statement. 

I experimented with using the raw stream for all the bits outside the ScanDecoder but never got that working... We'd be able to associate the InputProcessor with the decoder only and possibly speed that portion of the code up. Another time perhaps.

src/ImageSharp/Formats/Jpeg/GolangPort/OrigJpegConstants.cs
```diff
@@ -227,6 +209,32 @@ internal static class Markers
         }
 
         /// <summary>
+        /// Contains JFIF specific markers
+        /// </summary>
+        public static class JFif
```
tocsoft: is this still needed now you've moved to the `ToAsciiByte` style?

antonfirsov:
Everything looks good now but there is one thing I miss very much: 
We should add proper test cases to ensure we actually can detect the header markers. (Exif, JFif, Icc, Adobe). [What we currently have](https://github.com/SixLabors/ImageSharp/blob/master/tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.cs#L303) is very limited and doesn't really help us keeping our code correct while doing refactors.

@JimBobSquarePants you look to understand this stuff well now, you probably know what we should assert to to have a better coverage. I think it's worth to do it in this PR.

src/ImageSharp/Formats/Jpeg/GolangPort/OrigJpegConstants.cs
JimBobSquarePants: No. Good spot. I'll delete it, thought I already had.

JimBobSquarePants:
@antonfirsov You're right. Will figure out something better.

antonfirsov:
@JimBobSquarePants actually, I was thinking about integration-like tests:
Something like parsing image streams in `MetaDataOnly` mode, and asserting that the information in the headers is correctly fetched. We have [something for Exif & JFif](https://github.com/SixLabors/ImageSharp/blob/master/tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.cs#L303), but the rest is at 0% coverage AFAIK. 

This stuff is less trivial, but this is exactly why we need it much more!

Maybe there is some implicit coverage caused by having some test images with Icc and Adobe markers properly color processed *because* we can correctly read the headers, but I think it's time to make it explicit, defining `[Fact]`-s saying *"we can process this/that header"*.

JimBobSquarePants:
@antonfirsov We already have a comprehensive suite of [tests](https://github.com/SixLabors/ImageSharp/tree/master/tests/ImageSharp.Tests/MetaData/Profiles) for EXIF metadata parsing and ICC profile parsing though so as far as I'm concerned we cover that well enough. All we really need to do with that is ensure it's captured correctly which I've asserted with my refactor.

antonfirsov:
@JimBobSquarePants  I'm not talking about full Exif & ICC metadata parsing tests. What we need to do in the Jpeg test code is to ensure that we can detect the profile, and run the [necessary initialization blocks](https://github.com/SixLabors/ImageSharp/blob/js/cleanup-jpeg/src/ImageSharp/Formats/Jpeg/GolangPort/OrigJpegDecoderCore.cs#L530) for **a given input stream**.

I do not agree that your new tests are good enough for this, because `ProfileResolver.IsProfile()` is just very small and quite trivial part of the logic.  There is a lot more stuff happening around it which is much easier to break: reading the necessary amount of bytes, skipping bytes etc.

What we need to do:
- Pick an image for each of the application headers (Exif, JFif, ICC, Adobe)
- Run `JpegDecoderCore` with `MetaDataOnly == true` (to exclude the stuff that's not needed for these tests)
- Ensure that the critical internal `JpegDecoderCore` properties (indicating we have read the markers) are correctly set

This is critical for having a correct coverage for the code you refactored now + only developing these tests can help us evolving towards a better & more testable design.

JimBobSquarePants:
Testing internal fields is very brittle and exposing them purely for testing breaks encapsulation. Where do we draw the line?

I'm all for adding a test that ensures the metadata EXIF, ICC properties are populated but the rest... I'm not convinced yet.

antonfirsov:
@JimBobSquarePants I am testing internals (classes structs properties sometimes fields) all the way in Jpeg & all our other internal utils and I'm on the strong opinion that it's worth all the tradeoffs! 

For code with complicated internal logic & narrow public API (like Jpeg) black box testing is exactly the opposite of TDD. And I believe that TDD brings a huge value into our codebase by 
- Helping to keep our code correct between refactors 
- Pushing it towards modularity 
- Better maintainability
- Reduced number of bugs.

It means that we should aim for [White-box testing](https://en.wikipedia.org/wiki/White-box_testing) indicating that we should not be afraid to expose internals to tests. Everything that's important enough to be tested independently should be made internal and tested! It's a "by design" characteristic of white-box testing.

JimBobSquarePants:
@antonfirsov I think we might be talking about slightly different things here. 

I also believe we should be breaking up our code into smaller, more testable encapsulated objects. However, I believe we should be testing those objects themselves, not their state within a larger class. I.E. I shouldn't need a full integration test when I know my individual components are tested. 

White box testing requires explicit knowledge of the internal behaviour. We control the codebase yes but we, IMO, should be aiming to allow refactoring of those internals while keeping maintenance churn of the tests low. I shouldn't be avoiding the use of private fields.

Our public API is narrow but that's not a bad thing.

In this particular case, If I reintroduce the [`AdobeMarker`](https://github.com/SixLabors/ImageSharp/blob/1987eae9fbb798bf10cbc48ecce9cf07c2e13fcf/src/ImageSharp/Formats/Jpeg/Common/Decoder/AdobeMarker.cs) and [`JFifMarker`](https://github.com/SixLabors/ImageSharp/blob/1987eae9fbb798bf10cbc48ecce9cf07c2e13fcf/src/ImageSharp/Formats/Jpeg/Common/Decoder/JFifMarker.cs) structs, alter them slightly, and add tests for them I should be confident that the individual components work well while keeping them private fields.

antonfirsov:
@JimBobSquarePants in ideal case you are right, but you need to take into consideration the following points which make things really different:
- This is mostly ported code
- The time/timing factor

It's just impossible to turn these ported monsters into proper modular code in a few commits, not even in a few PR-s. What we can do is to cover them *as they are*, and use this coverage to support our current and future refactoring. With time, not only our code gets better but also our tests.

The result is the same: better, modular, and tested code, but I believe that the road leading here is less painful with the test-first approach. Actually: based on my previous Jpeg experience, it depends on our testing strategy if we ever get here!

In this particular case the most important is to test if the outcome of the `ProcessApp1Marker`, `ProcessApp2Marker` etc. functions is correct (the proper marker/profile data is initialized). To move these methods into independent modules you need to do *lots* of error prone refactors.

Why not adding the tests before all the refactors bringing bugs into our code?

antonfirsov:
Haven't read this book yet, but I beleive [this is the game we play here](https://www.amazon.com/Working-Effectively-Legacy-Michael-Feathers/dp/0131177052), and I'm sure it has a bit different winning strategy than the one when you develop modular components from scratch.

JimBobSquarePants:
I'm gonna push a few changes later that will help us with this particular scenario but I agree yeah, there is a balance.

antonfirsov:
@JimBobSquarePants good job, it's much better now! I like that now we have modular stuff that can be used by both PdfJsPort and GoLangPort decoders.

I see further possibilities to improve both the code and the tests though (the coverage of `Process****Marker` methods is still implicit), but we can fix this later. For mow everything LGTM.

src/ImageSharp/Formats/Jpeg/GolangPort/OrigJpegDecoderCore.cs
```diff
@@ -505,8 +455,7 @@ private void ProcessApp1Marker(int remaining)
             byte[] profile = new byte[remaining];
```
antonfirsov: I wonder if we can reuse `this.Temp`  on these locations.

JimBobSquarePants: Unfortunately not, it's far too large. I'd like to investigate pooling and copying the stream though, to an instance contained within the `ExifProfile` class.

JimBobSquarePants:
@antonfirsov I'm gonna merge this for now. I agree though that I'd like to get the testing better per marker by creating a type per marker. One thing that would really help would be restricting the `InputProcessor` to the `ScanDecoder`. It's only there that we really need it.

PR #354, opened 2017-09-27
Js/improve png encoder

JimBobSquarePants:
### Prerequisites

### Description
Improves the performance of the png encoder by combining variance summing with the filter encoding rather than looping through each row twice.


JimBobSquarePants:
@antonfirsov Nothing dramatic, Only a couple of milliseconds as those loops would break out early before anyway. 

I'd love to see if we could simdify that section one day.

JimBobSquarePants:
Imma gonna merge this as it's good.

antonfirsov:
@JimBobSquarePants it's possible to simdify it. We need to [eliminate this `if`](https://github.com/SixLabors/ImageSharp/blob/571b47f8f6293833f42014977963293928d625d3/src/ImageSharp/Formats/Png/Filters/AverageFilter.cs#L33) by splitting it into 2 loops first.

An untrivial if inside a loop has a very bad effect on CPU pipelines + branch prediction, so I expect significant speedup from removing it.

PR #357, opened 2017-10-05
Update EXIF on rotate. Fix #268

### LINKED ISSUE #268: Rotating an image does not update relevant EXIF data
Right now using image.RotateFlip doesn't update the Rotation EXIF data, which means that if it's present the image will still appear rotated in modern image viewers that automatically rotate the images.

image.Rotate should maybe update the EXIF data automatically to the new rotation value, of course it should be optional in case the "wrong" rotation is intended

JimBobSquarePants:
### Prerequisites

### Description
Since we don't know the updated rotation value and EXIF tags only cover certain angles the best fix is to remove the rotate tag. We also update height width if the canvas has been expanded.


src/ImageSharp/Processing/Processors/Transforms/RotateProcessor.cs
```diff
@@ -87,6 +88,30 @@ protected override void BeforeApply(ImageFrame<TPixel> source, Rectangle sourceR
...
+                // No need to do anything so return.
+                return;
+            }
+
+            profile.RemoveValue(ExifTag.Orientation);
+
+            if (this.Expand && profile.GetValue(ExifTag.PixelXDimension) != null)
+            {
+                profile.SetValue(ExifTag.PixelXDimension, source.Width);
+                profile.SetValue(ExifTag.PixelYDimension, source.Height);
```
dlemstra: It is possible that we add the `ExifTag.PixelYDimension` value. And I wonder if we should do this operation. Otherwise we need to this in a lot of spots. I personally think we shouldn't do this.

JimBobSquarePants: I made the assumption that if one exists, both will. Why wouldn't you do this? I had users asking for EXIF parity with ImageProcessor.

dlemstra: Doesn't this mean that you would also need to add something to the resize processor? Maybe we could do this logic inside `Sync` method of the ExifProfile if you really think we need this?

JimBobSquarePants: We would yeah, crop also. I'd love to have it centralized but how would it be aware of the required changes?

JimBobSquarePants:
I'm gonna merge this. It fixes our issue for now and we can always think of something better in the future.

PR #361, opened 2017-10-06
Js/fix diffused animations

JimBobSquarePants:
### Prerequisites

### Description
When encoding animated gifs we weren't resetting the colormap cache between frames. This caused banding for the Octree Quantizer and out of range exceptions in the Wu Quantizer. 

I've also switched out the default SierraLite Diffuser in QuantizerBase for the Floyd-Steinburg one despite it being slower as it yields much reduced file sizes. 30kb against one test image.


antonfirsov:
Wonder if we can find a way to regression-test gif encoding. ImageSharp Encoder VS ImageSharp decoder as a start?

JimBobSquarePants:
We could maybe do it but I don't know what the acceptable level of variance would be and how much it would differ between images since quantization is lossy.

PR #365, opened 2017-10-08
Enable decoding  only single frame from animated gif + fix quantization issue

JimBobSquarePants:
### Prerequisites

### Description
This PR was originally designed to simply add a `FrameDecodingMode` enumeration for decoding animated gifs but I discovered an inconsistency in quantization while I was cleaning up the codebase. 

As such I've combined the fix + tests to this PR.

### DIFF
--- a/src/ImageSharp/Formats/Gif/GifEncoderCore.cs
+++ b/src/ImageSharp/Formats/Gif/GifEncoderCore.cs
@@ -154,18 +154,14 @@ namespace SixLabors.ImageSharp.Formats.Gif
         {
             // Transparent pixels are much more likely to be found at the end of a palette
             int index = -1;
+            var trans = default(Rgba32);
             for (int i = quantized.Palette.Length - 1; i >= 0; i--)
             {
-                quantized.Palette[i].ToXyzwBytes(this.buffer, 0);
+                quantized.Palette[i].ToRgba32(ref trans);
 
-                if (this.buffer[3] > 0)
-                {
-                    continue;
-                }
-                else
+                if (trans.Equals(default(Rgba32)))
                 {
                     index = i;
-                    break;
                 }
             }
 
--- a/tests/ImageSharp.Tests/Formats/Gif/GifDecoderTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Gif/GifDecoderTests.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0.
 
 using System.Text;
-using SixLabors.ImageSharp.Formats;
 using SixLabors.ImageSharp.Formats.Gif;
 using SixLabors.ImageSharp.PixelFormats;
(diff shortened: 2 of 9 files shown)


antonfirsov:
@JimBobSquarePants I really want to review this, but I'm being short on time now. Do we still have a few days for this, or do you want to go fast because of beta-2?

JimBobSquarePants:
@antonfirsov I'm happy to delay beta-2 for this. The fix is an important one.

tests/ImageSharp.Tests/Formats/Gif/GifDecoderTests.cs
```diff
@@ -79,18 +78,40 @@ public void Decode_IgnoreMetadataIsTrue_CommentsAreIgnored()
...
+        }
+
+        [Theory]
+        [WithFile(TestImages.Gif.Giphy, PixelTypes.Rgba32)]
+        public void CanDecodeAllFrames<TPixel>(TestImageProvider<TPixel> provider)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            using (Image<TPixel> image = provider.GetImage(new GifDecoder { DecodingMode = FrameDecodingMode.All }))
+            {
+                Assert.True(image.Frames.Count > 1);
```
antonfirsov: Maybe not for this PR but ...
there are only 5 frames in this gif. By saving all of them as PNG we can introduce quite good regression testing for the gif decoder!

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -154,18 +154,14 @@ private int GetTransparentIndex<TPixel>(QuantizedImage<TPixel> quantized)
         {
             // Transparent pixels are much more likely to be found at the end of a palette
             int index = -1;
+            var trans = default(Rgba32);
             for (int i = quantized.Palette.Length - 1; i >= 0; i--)
             {
-                quantized.Palette[i].ToXyzwBytes(this.buffer, 0);
+                quantized.Palette[i].ToRgba32(ref trans);
```
antonfirsov: Cool to see more of these dropped!

antonfirsov:
Also, I'm copying my gitter comment here, because it's relevant:
> shouldn't we add 300x300-c.gif to our test images?

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
JimBobSquarePants: Yeah, I wanna replace all of them and remove the method very soon.

tests/ImageSharp.Tests/Formats/Gif/GifDecoderTests.cs
JimBobSquarePants: Indeed! I'll have a look

PR #367, opened 2017-10-09
Added RLE8 BMP decoder support. Fix #323

### LINKED ISSUE #323: BMP with 8-bit pixel depth and RLE won't load
### Prerequisites

### Description

Doing Image.Load() on the following file fails:

[TestImageBMP.zip](https://github.com/SixLabors/ImageSharp/files/1272446/TestImageBMP.zip)

### Steps to Reproduce

Do Image.Load on the included image.

### System Configuration

- ImageSharp version: 00191 and 00194
- .NET Framework version: .NET Core 1.1
- Additional information:

georgexcollins:
### Prerequisites

- [x ] I have written a descriptive pull-request title
- [x ] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x ] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x ] I have provided test coverage for my change (where applicable)

### Description
Added methods to BmpDecoderCore to add support for decoding RLE8 format. Added sample image from github issue #323 as a unit test case.


src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -209,6 +233,114 @@ private static int CalculatePadding(int width, int componentCount)
...
+        /// <param name="colors">The <see cref="T:byte[]"/> containing the colors.</param>
+        /// <param name="width">The width of the bitmap.</param>
+        /// <param name="height">The height of the bitmap.</param>
+        /// <param name="inverted">Whether the bitmap is inverted.</param>
+        private void ReadRle8<TPixel>(PixelAccessor<TPixel> pixels, byte[] colors, int width, int height, bool inverted)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            var color = default(TPixel);
+            var rgba = default(Rgba32);
+            byte[] data = this.UncompressRle8(width, height);
```
JimBobSquarePants: I think this could be better using `Buffer2D<Bgr24>` you would pool the array then and avoid the multiplier.

JimBobSquarePants:
Thanks for this; really appreciate it! Just one small comment where I think we can reduce allocations.

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
antonfirsov: @JimBobSquarePants good point! Don't you mean `Buffer2D<byte>` instead? I'm not familiar with the format, but [line 258](https://github.com/georgexcollins/ImageSharp/blob/b2dea16ab5a2b6e55e170ddec3622d8faecd23be/src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs#L258) suggests that values in `data` does some kind of indexing into the `colors` palette.

@georgexcollins `Buffer<T>` and `Buffer2D<T>` is our centralized memory management helper to overcome GC limitations. Currently it abstracts away pooling, but it will be capable of more magic in the future, so it's important to use it everywhere for large buffers.

tests/ImageSharp.Tests/TestImages.cs
```diff
@@ -134,8 +134,9 @@ public static class Bmp
             public const string NegHeight = "Bmp/neg_height.bmp";
             public const string CoreHeader = "Bmp/BitmapCoreHeaderQR.bmp";
             public const string V5Header = "Bmp/BITMAPV5HEADER.bmp";
-            
-            public static readonly string[] All = { Car, F, NegHeight, CoreHeader, V5Header };
+            public const string RLE = "Bmp/RunLengthEncoded.bmp";
+
+            public static readonly string[] All = { Car, F, NegHeight, CoreHeader, V5Header, RLE };
```
antonfirsov: This has nothing to do with this PR, but I leave my notes here as a reminder:
Adding an image to `TestImages.Bmp.All` adds a new test case to `BmpTests` (so 👍 for adding test coverage for your work @georgexcollins), but we should make `TestImages.Bmp.All` local to the `BmpTests` class, because the current situation is hard to follow.

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
JimBobSquarePants: @antonfirsov Sorry yeah, I meant `Buffer2D`, looking below though at the decompression routing I don't know if it's possible to use it though. Depends on what `cmd[0]` yields. 

@georgexcollins What do you think? You're the only one out of us who knows this part of the format 😄

georgexcollins:
That makes sense, I thought there would be a better way. I have updated to use Buffer<byte>. I had some issues with Git where I originally committed using Github client with wrong email address, hopefully it's now sorted.

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -209,6 +233,114 @@ private static int CalculatePadding(int width, int componentCount)
...
+            {
+                this.UncompressRle8(width, buffer);
+
+                for (int y = 0; y < height; y++)
+                {
+                    int newY = Invert(y, height, inverted);
+                    Span<TPixel> pixelRow = pixels.GetRowSpan(newY);
+                    for (int x = 0; x < width; x++)
+                    {
+                        rgba.Bgr = Unsafe.As<byte, Bgr24>(ref colors[buffer[(y * width) + x] * 4]);
```
antonfirsov: By using `Buffer2D<byte>` instead of `Buffer<byte>` you can easily avoid the repeating the `y * width` multiplications for each pixel: 
- Use `Span<byte> row = buffer.GetRowSpan(y)` right before the inner loop
- You can index int you row Span by then: `row[x]`

JimBobSquarePants: I don't think we can do that actually since we do not know which row to write to in `UncompressRle8`

If merged as-is for now.

JimBobSquarePants:
Thanks again @georgexcollins Everything was great! 👍

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
georgexcollins: I was going to write to buffer as before (it's still a Buffer<T>), and as Anton says, read whole lines.

Are the CreateClean static methods on Buffer2D<T> missing a 'new' modifier, I couldn't create a clean instance of Buffer2D

JimBobSquarePants: @georgexcollins The [`CreateClean`](https://github.com/SixLabors/ImageSharp/blob/09729940effcd25c385bd62b24f0b64b144a1fa2/src/ImageSharp/Memory/Buffer2D%7BT%7D.cs#L77) methods should work fine to create an instance that is clean for the required number of elements. 

It is possible however that the underlying array could have populated param type elements at the tail but they are outwith the maximum possible position in the 2D Buffer.

Feel free to submit another PR if you think you can improve on what you have. 😄

antonfirsov: @JimBobSquarePants I'm talking about reading the uncompressed `buffer` inside the for loops. The inner loop is always using the `y`-th row of `buffer` -> it can read a row span instead.

`UncompressRle8` can process the data into `buffer.Span` as if it was an array.

JimBobSquarePants: Fool of a Took! @antonfirsov You're absolutely right of course, I can't believe I missed that! 

@georgexcollins I'm gonna add that optimisation as part of #368 to save you the effort.

PR #368, opened 2017-10-11
Fix inverted RGB palette decoding plus 32bit RLE decoding

JimBobSquarePants:
### Prerequisites

### Description
While merging #367 I noticed that the assigned inverted `newY` property was never used in `ReadRgbPalette`.

I've also added an optimisation opportuinty @antonfirsov [suggested](https://github.com/SixLabors/ImageSharp/pull/367#discussion_r144064115) 

And fixed decoding of Rgba32 image formats. #369


src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -246,19 +247,21 @@ private void ReadRle8<TPixel>(PixelAccessor<TPixel> pixels, byte[] colors, int w
...
+            var rgba = new Rgba32(0, 0, 0, 255);
 
-            using (var buffer = Buffer<byte>.CreateClean(width * height))
+            using (var buffer = Buffer2D<byte>.CreateClean(width, height))
             {
                 this.UncompressRle8(width, buffer);
 
                 for (int y = 0; y < height; y++)
                 {
                     int newY = Invert(y, height, inverted);
```
antonfirsov: I think it's possible to go without the `Invert()` method by managing `newY` as a loop variable, but the benefits are not significant ... just talking my idea.

PR #377, opened 2017-10-30
Rotation by bi-linear interpolation

georgexcollins:
### Prerequisites

### Description
This is a first-draft attempt at bilinear interpolation following the algorithm as described here:
http://supercomputingblog.com/graphics/coding-bilinear-interpolation/

Before I go any further, could this be useful? There are various things that I know still need to be done and probably several that I don't know about yet.

This branch currently replaces the original RotateProcessor with the new interpolating version when the Rotate extension method is called, so the extension methods will need a way for the user to specify a resampling algo. There are likely optimisation to be made. I noticed in the resize processor there is a fork in the code based on 'Compand', I'm guessing this is probably applicable here, but not sure.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Transforms/RotateInterpolationProcessor.cs
+++ b/src/ImageSharp/Processing/Processors/Transforms/RotateInterpolationProcessor.cs
@@ -1,0 +1,97 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Numerics;
+using System.Text;
+using System.Threading.Tasks;
+using SixLabors.ImageSharp.Helpers;
+using SixLabors.ImageSharp.Memory;
+using SixLabors.ImageSharp.PixelFormats;
+using SixLabors.Primitives;
+
+namespace SixLabors.ImageSharp.Processing.Processors
+{
+    internal class RotateInterpolationProcessor<TPixel> : RotateProcessor<TPixel>
+        where TPixel : struct, IPixel<TPixel>
+    {
+        /// <inheritdoc/>
+        protected override void OnApply(ImageFrame<TPixel> source, Rectangle sourceRectangle, Configuration configuration)
+        {
+            if (this.OptimizedApply(source, configuration))
+            {
+                return;
+            }
+
+            int height = this.CanvasRectangle.Height;
+            int width = this.CanvasRectangle.Width;
+            Matrix3x2 matrix = this.GetCenteredMatrix(source, this.ProcessMatrix);
+            Rectangle sourceBounds = source.Bounds();
+
+            using (var targetPixels = new PixelAccessor<TPixel>(width, height))
+            {
+                Parallel.For(
+                    0,
+                    height,
+                    new ParallelOptions { MaxDegreeOfParallelism = 1 }, //configuration.ParallelOptions,
+                    y =>
+                    {
+                        Span<TPixel> targetRow = targetPixels.GetRowSpan(y);
+
+                        for (int x = 0; x < width; x++)
+                        {
+                            var transformedPoint = PointF.Rotate(new Point(x, y), matrix);
+                            var rounded = Point.Round(transformedPoint);
+                            if (sourceBounds.Contains(rounded.X, rounded.Y))
+                            {
+                                int ceilX = unchecked((int)MathF.Ceiling(transformedPoint.X));
+                                int ceilY = unchecked((int)MathF.Ceiling(transformedPoint.Y));
+                                int floorX = unchecked((int)MathF.Floor(transformedPoint.X));
+                                int floorY = unchecked((int)MathF.Floor(transformedPoint.Y));
+
+                                // Clamp sampling pixels to the source image edge
+                                ceilX = Math.Min(ceilX, sourceBounds.Width - 1);
+                                floorX = Math.Max(0, floorX);
+                                ceilY = Math.Min(ceilY, sourceBounds.Height - 1);
+                                floorY = Math.Max(0, floorY);
+
+                                // At the image edge, take the whole value of source
+                                float wx1, wx2, wy1, wy2;
+                                if (ceilX == floorX)
+                                {
+                                    wx1 = 1;
+                                    wx2 = 0;
+                                }
+                                else
+                                {
+                                    wx1 = ceilX - transformedPoint.X;
+                                    wx2 = transformedPoint.X - floorX;
+                                }
+                                if (ceilY == floorY)
+                                {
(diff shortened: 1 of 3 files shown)


JimBobSquarePants:
Hi @georgexcollins 

Sorry for the delay. Thanks for submitting this. 

Before we can consider accepting the PR we'll have to get a few things StyleCop things sorted out so it will build so if you can try building it in release mode locally and fix that, that would be great.

I'll do a full PR review as soon as possible. 

Cheers

James

src/ImageSharp/Processing/Processors/Transforms/RotateInterpolationProcessor.cs
```diff
@@ -0,0 +1,97 @@
...
+            int width = this.CanvasRectangle.Width;
+            Matrix3x2 matrix = this.GetCenteredMatrix(source, this.ProcessMatrix);
+            Rectangle sourceBounds = source.Bounds();
+
+            using (var targetPixels = new PixelAccessor<TPixel>(width, height))
+            {
+                Parallel.For(
+                    0,
+                    height,
+                    new ParallelOptions { MaxDegreeOfParallelism = 1 }, //configuration.ParallelOptions,
```
JimBobSquarePants: If it doesn't work in parallel there's little point using `Parallel.For`

src/ImageSharp/Processing/Processors/Transforms/RotateInterpolationProcessor.cs
```diff
@@ -0,0 +1,97 @@
...
+                    {
+                        Span<TPixel> targetRow = targetPixels.GetRowSpan(y);
+
+                        for (int x = 0; x < width; x++)
+                        {
+                            var transformedPoint = PointF.Rotate(new Point(x, y), matrix);
+                            var rounded = Point.Round(transformedPoint);
+                            if (sourceBounds.Contains(rounded.X, rounded.Y))
+                            {
+                                int ceilX = unchecked((int)MathF.Ceiling(transformedPoint.X));
```
JimBobSquarePants: Why are these unchecked? Can we really go outwith int bounds?

src/ImageSharp/Processing/Processors/Transforms/RotateInterpolationProcessor.cs
```diff
@@ -0,0 +1,97 @@
...
+                                int floorY = unchecked((int)MathF.Floor(transformedPoint.Y));
+
+                                // Clamp sampling pixels to the source image edge
+                                ceilX = Math.Min(ceilX, sourceBounds.Width - 1);
+                                floorX = Math.Max(0, floorX);
+                                ceilY = Math.Min(ceilY, sourceBounds.Height - 1);
+                                floorY = Math.Max(0, floorY);
+
+                                // At the image edge, take the whole value of source
+                                float wx1, wx2, wy1, wy2;
```
JimBobSquarePants: I'd really like to get this working for any passed in resampling algorithm rather than just bilinear.  Could we refactor this to use weighting algorithms? @antonfirsov What do you think?

src/ImageSharp/Processing/Transforms/Rotate.cs
```diff
@@ -47,6 +47,6 @@ public static IImageProcessingContext<TPixel> Rotate<TPixel>(this IImageProcessi
         /// <returns>The <see cref="Image{TPixel}"/></returns>
         public static IImageProcessingContext<TPixel> Rotate<TPixel>(this IImageProcessingContext<TPixel> source, float degrees, bool expand)
             where TPixel : struct, IPixel<TPixel>
-        => source.ApplyProcessor(new RotateProcessor<TPixel> { Angle = degrees, Expand = expand });
+        => source.ApplyProcessor(new RotateInterpolationProcessor<TPixel> { Angle = degrees, Expand = expand });
```
JimBobSquarePants: How do I choose between rotation sampling algorithm?

src/ImageSharp/Processing/Processors/Transforms/RotateInterpolationProcessor.cs
georgexcollins: Temporary commented out to aid debugging... It works fine in parallel.

src/ImageSharp/Processing/Processors/Transforms/RotateInterpolationProcessor.cs
georgexcollins: I can add an overload to the ImageExtensions.Rotate to include a IRotationResampler, and then factor out the resampling logic into a bi-linear interpolation implementation of that interface. This would be much like the Resize setup. I thought I should be able to use the resamplers that already exist there but I cannot get my head around how the resize processor works. Any pointers welcome.

georgexcollins:
Thanks for the comments. As I said this is very much in progress, just checking I'm doing something worthwhile. I'll probably get back to this in a couple of weeks time.

JimBobSquarePants:
Definitely something worthwhile so we should try to get it working well. 👍 

I wonder whether we could adapt the work done here (originally a port of ImageSharp)?  It's based upon a similar weights principle. I really like the way affine transforms have been implemented here.  

https://github.com/JaCraig/Structure.Sketching/blob/3ef7cf83a825b7bd962dfe6682475e8fceade597/Structure.Sketching/Filters/Resampling/BaseClasses/AffineBaseClass.cs#L132

src/ImageSharp/Processing/Processors/Transforms/RotateInterpolationProcessor.cs
JimBobSquarePants: 👍

src/ImageSharp/Processing/Processors/Transforms/RotateInterpolationProcessor.cs
JimBobSquarePants: Resize essentially calculates a collection of sampled weights for rows and columns based on the given algorithm and the input output size, then a window (of length determined by the algorithm) within that collection is summed and the input vector is then multiplied by that sum. 

Our implementation is complicated to follow but it's a good sacrifice to make for the superior performance it delivers.

There's a good introduction to resampling here.

http://entropymine.com/imageworsener/resample/

JimBobSquarePants:
Hi @georgexcollins 

Many, many thanks for your help with this but I'm going to close this PR in favour of #386 

You invigorated me to have another crack at the problem and as such I'm managed to create a solution that handles all the different `IResampler` implementations. 

I couldn't have done it without you.

Cheers

James

PR #384, opened 2017-11-21
Remove PixelConversionExtensions

JimBobSquarePants:
### Prerequisites

### Description
Removes the old `ToXyzBytes` method from the source and tests. They've served us well but the new `ToRgba32` and other methods are faster, more semantic, and less brittle.

I've also added an indexer to `Rgba32` to allow us to test the correct component during ordered dithering.


src/ImageSharp/PixelFormats/Rgba32.cs
```diff
@@ -220,6 +220,32 @@ public uint PackedValue
             set => this.Rgba = value;
         }
 
+        /// <summary>
+        /// Gets the component value at the given index
+        /// </summary>
+        /// <param name="index">The component index</param>
+        /// <returns>The <see cref="byte"/></returns>
+        public byte this[int index]
```
antonfirsov: We need a more consistent behavior if we want to expose this publicly. Currently it:
- Throws in debug mode for an index being out of range (indicating an incorrect usage)
- Silently, returns Alpha in release mode (this behavior is undocumented!)

JimBobSquarePants: I'll refactor to not use `default` and throw after the switch

tocsoft: do we event need this at all? its only used here https://github.com/SixLabors/ImageSharp/blob/65b419085131fd6b7ba32d2709d0ce3bb61a4d3b/src/ImageSharp/Dithering/Ordered/OrderedDitherBase.cs#L29-L34 

I feel its probably better to just moving the switch inline to that method and not add it unnecessarily to the public api

JimBobSquarePants:
Thanks @tocsoft , @antonfirsov we good?

antonfirsov:
@JimBobSquarePants yeah, everything looks fine for me for now. 

However that ditherer code could be optimized a lot later. Being curious: do we have (at least integration) tests for it now?

JimBobSquarePants:
@antonfirsov Yeah we do in the unit test project, it's been around for ages but it was depending on the `Xyzw` method. 

The problem optimising something like this is it has to consumable both per-pixel and per row plus ordered dither implementations can have different matrices of different dimensions.

PR #386, opened 2017-11-23
Add Interpolated Affine Transforms

### LINKED ISSUE #31: Rotation output quality isn't very nice
Hi,

It looks like there's no interpolation happening for rotation - can I add that in as a feature request? The quality of images, when rotated by arbitrary degrees, are quite poor.

Thanks,

JimBobSquarePants:
### Prerequisites

### Description
Fixes #31 and replaces #377 

This PR adds interpolation options to affine transforms. 🎉 

- Adds new base class `AffineProcessor` which is used as a base to `RotateProcessor`, `SkewProcessor`, and `TransformProcessor`
- Adds a new `TransformProcessor` class and calling methods which can be used to apply combined mutliple transforms such as `Rotate`, `Skew`, and `Scale`.
- Adds overloads to `Rotate` and `Skew` that accepts any `IResampler` implementation
- Aggregated known `IResampler` implementations into a single class `KnownResamplers` to reduce the allocation on each run. 

affine

Here's example code combining rotation and scale transforms to rotate the image by 45 degrees and reduce the size by two using the `BicubicResampler`.

``` c#
    // Load image
    Matrix3x2 rotate = Matrix3x2Extensions.CreateRotationDegrees(45F);
    Matrix3x2 scale = Matrix3x2Extensions.CreateScale(new SizeF(.5F, .5F));

    image.Mutate(i => i.Transform(rotate * scale, KnownResamplers.Bicubic));

    // Save image
```


antonfirsov:
@JimBobSquarePants Just pushed a modification so `RotateWithSampler` is executed with all know resamplers (with a single angle).

By scrolling through the result files you can notice a "shape shifting" behavior, that seems strange to me. Is this the expected behavior of the resamplers?

tests/ImageSharp.Tests/Processing/Processors/Transforms/RotateTests.cs
```diff
@@ -39,6 +61,21 @@ public void Rotate<TPixel>(TestImageProvider<TPixel> provider, float value)
...
+        [WithTestPatternImages(nameof(ResmplerNames), 50, 100, DefaultPixelType)]
+        public void RotateWithSampler<TPixel>(TestImageProvider<TPixel> provider, string resamplerName)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            IResampler resampler = GetResampler(resamplerName);
+
+            using (Image<TPixel> image = provider.GetImage())
+            {
+                image.Mutate(x => x.Rotate(50, resampler));
+                image.DebugSave(provider, resamplerName);
```
antonfirsov: If we are fine with the current results, it would be wise, to push the results into the reference image repository, and add `image.CompareToReferenceOutput(provider, ImageComparer.Tolerant());` here. At least for a few resamplers.

src/ImageSharp/Processing/Processors/Transforms/RotateProcessor.cs
```diff
@@ -17,81 +16,129 @@ namespace SixLabors.ImageSharp.Processing.Processors
...
                     configuration.ParallelOptions,
                     y =>
-                    {
-                        Span<TPixel> targetRow = targetPixels.GetRowSpan(y);
+                        {
+                            Span<TPixel> destRow = destination.GetPixelRowSpan(y);
+
+                            for (int x = 0; x < width; x++)
+                            {
+                                var transformedPoint = Point.Rotate(new Point(x, y), matrix);
```
antonfirsov: It's actually a `SixLabors.Core` issue, but I noticed that `Point.Rotate()` is the same as `Point.Skew()`, the code should be moved into a single `Point.Transfrom()` method IMO, and we should add new overloads, accepting matrix by `ref` to reduce the copy operations.

I wonder if it's possible to do a better resampling strategy. Currently our weights are centered at a rounded position. I should really read more image processing literature.

tests/ImageSharp.Tests/Processing/Processors/Transforms/RotateTests.cs
JimBobSquarePants: I think you've found an issue. I'm not sampling the offset correctly .

I'm starting at zero when I should be starting at -radius. I'll have a look.

JimBobSquarePants: Much better results now for the larger window samplers. We still get slight offsetting for those samplers but that's simply how they work and our output images are tiny so we notice it more. 

No-one in their right mind would use a Lanczos8 sampler for rotation anyway 😛.

src/ImageSharp/Processing/Processors/Transforms/RotateProcessor.cs
JimBobSquarePants: PR on Core submitted https://github.com/SixLabors/Core/pull/11

antonfirsov:
@JimBobSquarePants I compared our results to ImageMagick output I think there are some inaccuracies (which were there before this PR as well).

#### ImageMagick output ([with Paeth filter](https://www.imagemagick.org/script/command-line-options.php#rotate)):
01_imagemagick_paeth

#### Our output with NearestNeighbour filter (same as the parameterless rotation before this PR):
02_imagesharp_nearestneighbor

#### Our output with Triangle filter:
04_imagesharp_triangle

### Conclusions:
1. Our output dimensions are inaccurate. The right corner of the pattern is bitten off the screen.
2. It looks like there is an unnecessary global translation applied to our output.
3. ImageMagick output looks much smoother compared to all our filters. Is it because the Paeth filter?

src/ImageSharp/Processing/Processors/Transforms/AffineProcessor.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// <returns>The <see cref="Matrix3x2"/></returns>
+        protected abstract Matrix3x2 CreateProcessingMatrix();
+
+        /// <summary>
+        /// Creates a new target canvas to contain the results of the matrix transform.
+        /// </summary>
+        /// <param name="sourceRectangle">The source rectangle.</param>
+        protected virtual void CreateNewCanvas(Rectangle sourceRectangle)
+        {
+            if (this.ResizeRectangle == Rectangles.DefaultRectangle)
```
antonfirsov: Very strange logic through the inheritance hierarchy. Aren't we fine by always calculating `ResizeRectangle` from the affine matrix without specializations?

JimBobSquarePants: In rotate, yeah, i'm gonna refactor that. The potential time saved is not worth the code complexity.

Re our resampling, fixing the dimensions is easy - a missing `MathF.Ceiling` when calculating the bounds. I cannot get it to match the ImageMagick quality though and I don't know why. 😦 

Maybe I'm using the wrong weights or something.

src/ImageSharp/Processing/Processors/Transforms/AffineProcessor.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    internal abstract class AffineProcessor<TPixel> : ResamplingWeightedProcessor<TPixel>
+        where TPixel : struct, IPixel<TPixel>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AffineProcessor{TPixel}"/> class.
+        /// </summary>
+        /// <param name="sampler">The sampler to perform the resize operation.</param>
+        protected AffineProcessor(IResampler sampler)
+            : base(sampler, 1, 1, Rectangles.DefaultRectangle) // Hack to prevent Guard throwing in base, we always set the canvas
```
antonfirsov: I think the we should not force having a know `ResizeRectangle` at constructor time. We should probably turn `CreateNewCanvas()` into a `Rectangle CalculateDestRectangle(sourceRectangle)` calling it, and setting `ResizeRectangle` inside the `BeforeApply()` step.
`ResizeProcessor` could return a static size, known at constructor time then.

JimBobSquarePants: I'll have a look. The `CreateDestination` method is used to determine the size of the destination output in the base class so we require a known height/width then. This takes place before `BeforeApply` so I might not be able to do much.

src/ImageSharp/Processing/Processors/Transforms/AffineProcessor.cs
JimBobSquarePants: I think I'm gonna drop the expand option and always expand the canvas as it's the most common use-case by far. if someone wants to crop it they can do so as a separate operation. 

It's making the maths annoying and overly complicated for our optimized methods.

JimBobSquarePants:
I'm marking this as work-in-progress. I can't figure out what I'm doing wrong and the initial maths were totally incorrect.

JimBobSquarePants:
rotate

It's aliiiiiiive!

JimBobSquarePants:
Go home codecov, you're drunk.

antonfirsov:
Awesome! :) Gonna review this tonight!

JimBobSquarePants:
I'm gonna need a hand designing the Transform API so we put together something really useful for developers to consume.

src/ImageSharp/Processing/Transforms/Resamplers/Lanczos3Resampler.cs
```diff
@@ -14,6 +16,7 @@ public class Lanczos3Resampler : IResampler
         public float Radius => 3;
 
         /// <inheritdoc/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
antonfirsov: Unfortunately, it doesn't help here. These method's are resolved dynamically through the interface, [so they can not be inlined](http://www.cs.technion.ac.il/users/yechiel/c++-faq/inline-virtuals.html).

JimBobSquarePants: Dammit! I knew this, just forgot. wish we had a way around that.

JimBobSquarePants:
Zero idea why the Appveyor build failed. All tests passed.

src/ImageSharp/Processing/Processors/Transforms/AffineProcessor.cs
JimBobSquarePants: We don't use anything useful from the base class now so this will be refactored to stand alone

JimBobSquarePants:
Ok, I think this is ready for review now.

src/ImageSharp/Processing/Processors/Transforms/RotateProcessor.cs
```diff
@@ -47,6 +47,7 @@ public RotateProcessor(float degrees, IResampler sampler)
         /// <inheritdoc/>
         protected override Matrix3x2 GetTransformMatrix()
         {
+            // Tansforms are inverted else the output is the opposite of the expected.
```
Romanx: Just a lerker but spelling mistake: *"Tansforms"* -> Transforms

src/ImageSharp/Processing/Processors/Transforms/SkewProcessor.cs
```diff
@@ -50,6 +50,7 @@ public SkewProcessor(float degreesX, float degreesY, IResampler sampler)
         /// <inheritdoc/>
         protected override Matrix3x2 GetTransformMatrix()
         {
+            // Tansforms are inverted else the output is the opposite of the expected.
```
Romanx: Just a lerker but spelling mistake: *"Tansforms"* -> Transforms

src/ImageSharp/Processing/Processors/Transforms/TransformProcessor.cs
```diff
@@ -30,6 +30,8 @@ public TransformProcessor(Matrix3x2 matrix)
         public TransformProcessor(Matrix3x2 matrix, IResampler sampler)
             : base(sampler)
         {
+
+            // Tansforms are inverted else the output is the opposite of the expected.
```
Romanx: Just a lerker but spelling mistake: *"Tansforms"* -> Transforms

tocsoft:
Moving over the resize to the same code path seems sensible to me. We reduce moving parts for initial release and then we can always fork the resize operation back out later again later if we feel its a place we want to apply any potential optimisations.

src/ImageSharp/Common/Helpers/ImageMaths.cs
```diff
@@ -147,17 +147,26 @@ public static Rectangle GetBoundingRectangle(Point topLeft, Point bottomRight)
...
+            var br = Vector2.Transform(new Vector2(rectangle.Width, rectangle.Height), matrix);
+
+            // Find the minimum and maximum "corners" based on the ones above
+            float minX = MathF.Min(tl.X, MathF.Min(tr.X, MathF.Min(bl.X, br.X)));
+            float maxX = MathF.Max(tl.X, MathF.Max(tr.X, MathF.Max(bl.X, br.X)));
+            float minY = MathF.Min(tl.Y, MathF.Min(tr.Y, MathF.Min(bl.Y, br.Y)));
+            float maxY = MathF.Max(tl.Y, MathF.Max(tr.Y, MathF.Max(bl.Y, br.Y)));
+            var min = new Vector2(minX, minY);
+            var max = new Vector2(maxX, maxY);
+            Vector2 size = max - min;
```
dlemstra: I wonder if you need the Vector2 here. Why is that better than `float sizeX = maxX -minX`?

src/ImageSharp/Processing/Processors/Transforms/RotateProcessor.cs
```diff
@@ -17,95 +16,72 @@ namespace SixLabors.ImageSharp.Processing.Processors
...
+            if (MathF.Abs(this.Degrees) < Constants.Epsilon)
             {
                 // No need to do anything so return.
                 return;
             }
 
             profile.RemoveValue(ExifTag.Orientation);
 
-            if (this.Expand && profile.GetValue(ExifTag.PixelXDimension) != null)
+            if (profile.GetValue(ExifTag.PixelXDimension) != null)
```
dlemstra: I wonder if this will be correct for auto rotated images. This had `this.Expand` set to false which would mean that we would not update the values. But I do think we could update the values though. Not sure why we didn't?

src/ImageSharp/Processing/Transforms/Resamplers/KnownResamplers.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    /// <summary>
+    /// Contains reusable static instances of known resampling algorithms
+    /// </summary>
+    public static class KnownResamplers
+    {
+        /// <summary>
+        /// Gets the Bicubic sampler that implements the bicubic kernel algorithm W(x)
+        /// </summary>
+        public static IResampler Bicubic { get; } = new BicubicResampler();
```
dlemstra: You are now reusing the sampler across threads. Good change 👍

tests/ImageSharp.Tests/Processing/Transforms/TransformTests.cs
```diff
@@ -0,0 +1,76 @@
...
+                  {
+                      { 20, 10, 45 },
+                      { -20, -10, 45 }
+                  };
+
+        public static readonly List<string> ResamplerNames
+            = new List<string>
+                  {
+                      nameof(KnownResamplers.Bicubic),
+                      //nameof(KnownResamplers.Box),
```
dlemstra: Do we need those commented out lines?

src/ImageSharp/Processing/Processors/Transforms/RotateProcessor.cs
JimBobSquarePants: Previously the rotations that fell under that category (90, 180, 270) actually ignored the expanded value and automatically expanded anyway. Not updating them was a mistake on my part.

src/ImageSharp/Common/Helpers/ImageMaths.cs
JimBobSquarePants: Tbh, I don't think it is better.

tests/ImageSharp.Tests/Processing/Transforms/TransformTests.cs
JimBobSquarePants: Depends on how we do the testing. We really should have an expected output + comparison to prevent regression bugs for all these but I wanted approval of the output before I created standard files.

antonfirsov: We probably should keep these here, in the generic transform test class + drop them in `RotateTest` and `SkewTest`.

tests/ImageSharp.Tests/Processing/Processors/Transforms/RotateTests.cs
```diff
@@ -26,6 +29,26 @@ public class RotateTests : FileTestBase
             RotateType.Rotate270
         };
 
+        public static readonly TheoryData<string> ResamplerNames
+            = new TheoryData<string>
+        {
+            nameof(KnownResamplers.Bicubic),
```
antonfirsov: Now that we introduced `TransformProcessor` and `TransformTests`, I think `RotateTests` should focus mostly on corner cases it specializes. (See my comment on `TransformTests`)

tests/ImageSharp.Tests/Processing/Transforms/TransformTests.cs
```diff
@@ -0,0 +1,76 @@
...
+                      //nameof(KnownResamplers.NearestNeighbor),
+                      //nameof(KnownResamplers.Robidoux),
+                      //nameof(KnownResamplers.RobidouxSharp),
+                      //nameof(KnownResamplers.Spline),
+                      //nameof(KnownResamplers.Triangle),
+                      //nameof(KnownResamplers.Welch),
+                  };
+
+        [Theory]
+        [WithFileCollection(nameof(DefaultFiles), nameof(TransformValues), DefaultPixelType)]
```
antonfirsov: Let's do this with the test pattern instead! (-> smaller actual and reference outputs).

I also suggest to have two `[Theory]`-s to cover this stuff:
- One for going through all the ResamplerName-s using it as a theory parameter + fixing a single transform value
- Another one to test different transform parameters with a single resampler.

This way we can cover everything without bloating the test-space.

antonfirsov: I will do this very soon!

antonfirsov:
Pushed my changes to `TransformTests`, haven't played with ImageMagick yet. 
@JimBobSquarePants maybe not that important, but looks like there are artifacts with Lancos5 & Lancos8. If we can't figure it out now, for me it is OK to put them on our known issues list.

JimBobSquarePants:
@antonfirsov I've pushed an update that fixes the Lanczos banding and improves edge output overall. There's scope to improving edge handling to reduce darkening of edge pixels but I have a horrible feeling that will require companding to/from the linear color space to transform with the appropriate gamma value. That would be painfully slow with our per-pixel calculations. 🙁 

You'll see some slight internal ringing with the larger Lancsoz samplers when you zoom right in but that's [expected](https://en.wikipedia.org/wiki/Lanczos_resampling#Limitations).

According to the ImageMagick source (resample.c) this [paper](http://www.cs.cmu.edu/~ph/texfund/texfund.pdf) contains all the secrets to resampling.  but it's a bit too mathsy for me. 

Regarding Image sizes. I've updated the algorthim to +-.5F in the correct manner but our output size is still 1px smaller in both directions than ImageMagick. I have no idea why, as technically everything I have done is correct. It doesn't _seem_ to be cropping anything other than in nearest neighbor mode which is to be expected. (Other samplers definitely hit those edges correctly just a little darker than i'd like).

Regarding tests. Yes, I want samples to compare to but they have to be based upon what we think is the best possible output from our algorithm. Comparing to ImageMagick per pixel will be fruitless so I think we should simply use that for aspirational purposes  re edge handling. 

Oh and Matrix translations. I don't know how to do the maths to adjust the canvas for those.

antonfirsov:
@JimBobSquarePants I didn't mean to use ImageMagick output as reference, I just want to manually compare it to ours before moving on.

JimBobSquarePants:
@antonfirsov Wires crossed, sorry! 😄

antonfirsov:
@JimBobSquarePants now I understand why are ImageMagick dimensions larger: they extend the image with the resampling radius to keep the resampled (blurred) edge pixels on the image. We are currently cropping those 1-3 pixels when rotating an image that has "significant" parts on corners.

I wonder if there are users who will need this, or this something that won't be noticed by anyone ever?

JimBobSquarePants:
@antonfirsov I think since we had to zoom in so much we can probably do without. 

I don't know if you have ever had a good look at System.Drawings rotation but it's pretty ropey zoomed in also. 

The only thing that really bothers me is the darkening of edge pixels. I can't figure out reading the IM source how they manage to keep the component values intact.

antonfirsov:
@JimBobSquarePants I'm also trying to investigate this. Managed to define a failing test with a `5x5` full white image. Should I push it now?
My suspicion is that we are sampling some virtual black pixels from the "outer space".

antonfirsov:
@JimBobSquarePants @tocsoft @dlemstra and **anyone reading this**:
Here is the final dilemma to answer about our TransformProcessor + affine matrix rules:
Looks we are using a pixel origin convention based on image center `(w/2, h/2)`. (Eg. a rotation matrix will rotate around the center of the image (after an automatic translation to the center)). Are we sure this what most "bring your own matrix" users want? ImageMagick affine transforms are based at the top left corner as far as I understand. [Looks like](https://docs.opencv.org/3.0-beta/doc/py_tutorials/py_imgproc/py_geometric_transformations/py_geometric_transformations.html) OpenCV works the same way.

Thoughts?

src/ImageSharp/Processing/Transforms/Transform.cs
```diff
@@ -0,0 +1,39 @@
...
+    public static partial class ImageExtensions
+    {
+        /// <summary>
+        /// Transforms an image by the given matrix.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="source">The image to skew.</param>
+        /// <param name="matrix">The transformation matrix.</param>
+        /// <returns>The <see cref="Image{TPixel}"/></returns>
+        public static IImageProcessingContext<TPixel> Transform<TPixel>(this IImageProcessingContext<TPixel> source, Matrix3x2 matrix)
```
antonfirsov: An API idea:
Maybe we could provide a few shortcuts for common affine transforms:
```C#
public enum TransformOrigin { Center, TopLeft }

public static IImageProcessingContext<TPixel> Transform<TPixel>(this IImageProcessingContext<TPixel> source, float rotationAngle, Vector2 scale, Vector2 translation, TransformOrigin origin = TransformOrigin.Center) { ... }

public static IImageProcessingContext<TPixel> Transform<TPixel>(this IImageProcessingContext<TPixel> source, float rotationAngle = 0, float uniformScale = 1, float translationX = 0, float translationY = 0, TransformOrigin origin = TransformOrigin.Center) { ... }

// etc ...

```

JimBobSquarePants:
@antonfirsov Please push the test. It's be useful for working against.

I know why we get darker pixels (It's due to a lack of samples as we hit the edge and stop) and I'm gonna try a workaround idea I have. 

Re the matrix transforms. I'm going to move automatic centering to the separate `Rotate` and `Skew` processors since this is what people will expect for them. This will make `Transform` itself free-form. 

I'll have a look at the `DrawImageProcessor` to integrate the transforms into there also.

src/ImageSharp/Processing/Processors/Transforms/AffineProcessor.cs
```diff
@@ -0,0 +1,304 @@
...
+
+        /// <summary>
+        /// Creates a new target canvas to contain the results of the matrix transform.
+        /// </summary>
+        /// <param name="sourceRectangle">The source rectangle.</param>
+        private void ResizeCanvas(Rectangle sourceRectangle)
+        {
+            this.transformMatrix = this.GetTransformMatrix();
+
+            this.targetRectangle = Matrix3x2.Invert(this.transformMatrix, out Matrix3x2 sizeMatrix)
```
antonfirsov: Maybe we should allow users to define their own `targetRectangle`. Some of them might need deterministic output sizes instead of automatically fitting the transformed image. This would allow them to crop or extend the result image.

JimBobSquarePants:
@antonfirsov So I'm fairly certain ImageMagick cheats on the rotation to get those super nice edges. 

https://github.com/ImageMagick/ImageMagick/blob/d60778438c241c445ea589a59474750de7bb9a2e/MagickCore/shear.c#L1781

It looks to me like they apply a border matching the background, shear, then crop. 

I've refactored our processors to allow free-form transforms using the `Transform` extension. I've aso enhanced that method to allow the passing of a target rectangle as you suggested.

I've still to add transforms to the DrawImage API but I thought I'd let you have a look at my updates first plus it's getting late.

src/ImageSharp/Processing/Transforms/Transform.cs
JimBobSquarePants: Aye maybe. Let's get the basics in first then expand. Those `Transform.Origin` values could maybe complicate things

JimBobSquarePants:
Well I'll be damned! 😄 

rotate-better

src/ImageSharp/Processing/Transforms/Transform.cs
```diff
@@ -0,0 +1,53 @@
...
+            where TPixel : struct, IPixel<TPixel>
+        => Transform(source, matrix, KnownResamplers.NearestNeighbor);
```
antonfirsov: For Resize, default resampler is `KnownResamplers.Bicubic`. Shouldn't we be consistent here?

antonfirsov:
@JimBobSquarePants finally did some progress with *output verification* point, and found a few minor issues with the output. Reproduction: run the test cases doing 50deg rotation without rescaling or translating the image.

1. When no destination `rectangle` is specified, the calculated one has an ad-hoc clipping:
transform_rotatescaletranslate_autodestrectangle_rgba32_testpattern100x50_r 50 _s 1 1 _t 0 0
I think it should fit the image into the transformed input rectangle instead.

2. When drawing the image into the same rectangle, there is a minor 1px difference compared to the ImageMagick output:
rot50magick
transform_rotatescaletranslate_samedestrectangle_rgba32_testpattern100x50_r 50 _s 1 1 _t 0 0
Doesn't look like a big issue, but I think at this point we should have our output as correct as possible in order to have good regression testing introduced.
The ImageMagick output was produced with the following command:
```
magick convert Orig.png -matte -virtual-pixel Transparent -distort AffineProjection 0.6427876,0.7660444,-0.7660444,0.6427876,0,0 Rot50Magick.png
```
@dlemstra can you confirm whether I'm using ImageMagick correctly? The matrix is exactly the same we produce in the `Matrix3x2 m = rotate * scale * translate;` line.

src/ImageSharp/Processing/Transforms/Transform.cs
JimBobSquarePants: Perhaps, I'd probably choose Triangle though for performance.

JimBobSquarePants:
>1. When no destination `rectangle` is specified, the calculated one has an ad-hoc clipping on the left + padding on the right:
transform_rotatescaletranslate_autodestrectangle_rgba32_testpattern100x50_r 50 _s 1 1 _t 0 0
I think it should fit the image into the transformed input rectangle when `rectangle` is not specified by the user.

When we don't have an explicit rectangle set we default to the world bounds of the matrix transform. (Is that wise? I don't know. Maybe we should simply stick with the source rectangle... Actually, I think I prefer that.)

In either case, since you are rotating at point 0,0 you will always get cut off. This is why both `Rotate` and `Skew` methods specifically recenter the image.

>2. When drawing the image into the same rectangle, there is a minor 1px offset compared to the ImageMagick output:
rot50magick
transform_rotatescaletranslate_samedestrectangle_rgba32_testpattern100x50_r 50 _s 1 1 _t 0 0
Doesn't look like a big issue, but I think at this point we should have our output as correct as possible in order to have good regression testing introduced.
The ImageMagick output was produced with the following command:
```
magick convert Orig.png -matte -virtual-pixel Transparent -distort AffineProjection 0.6427876,0.7660444,-0.7660444,0.6427876,0,0 Rot50Magick.png
```
I don't know If I'm bothered by this. I'm confident in my maths so far.

src/ImageSharp/Processing/Processors/Transforms/AffineProcessor.cs
```diff
@@ -0,0 +1,334 @@
...
+                            {
+                                // Use the single precision position to calculate correct bounding pixels
+                                // otherwise we get rogue pixels outside of the bounds.
+                                var point = Vector2.Transform(new Vector2(x, y), matrix);
+
+                                // Clamp sampling pixel radial extents to the source image edges
+                                Vector2 maxXY = point + radius;
+                                Vector2 minXY = point - radius;
+
+                                var extents = new Vector4(
```
antonfirsov: Is this correct in cases like `maxXY.X < 0`? Isn't it better to have:
```C#
var extents = new Vector4(
    MathF.Round(maxXY.X),
    MathF.Round(maxXY.Y),
    MathF.Round(minXY.X),
    MathF.Round(minXY.Y));
```

JimBobSquarePants: No, the extants are used for getting the sub-pixel values to pass to the resampler. If you round them that's how you end up with the crappy output I had initially.

JimBobSquarePants:
@antonfirsov 

Ok, I've made a few changes based on your suggestions.

- We now use `KnownResamplers.Bicubic` for our default transforms
- When no rectangle is passed to the freeform transforms the processor will use the source value
- I've added a `TransformHelpers` class with a single method `GetTransformedBoundingRectangle` which makes it easier to calculate the bounding box of a transform. Developers can then use that result to calculate the correct translation values based on the rectangle `X` and `Y` values.

antonfirsov:
> When no rectangle is passed to the freeform transforms the processor will use the source value

This makes the most sense to me as well!

> I don't know If I'm bothered by this. I'm confident in my maths so far.

I went on producing test cases, and I'm quite sure now that the ImageMagick output is correct and somehow our result has an unnecessary `(+1, 0)` offset. Can't spot the issue in the code however, and it's late now. Will keep hunting tomorrow.

JimBobSquarePants:
Ok, I'll have another look myself and see if I can find anything.

antonfirsov:
@JimBobSquarePants pushed a better test case for this: `Transform_RotateScale_ManuallyCentered`

Original:
origmanualcenter

ImageMagick:
rot50manualcentermagick

ImageSharp:
transform_rotatescale_manuallycentered_rgba32_testpattern96x96_r 50 _s 0 8

Not sure if the x error is `1px`, migh be actually a sup-pixel value (0.5?).

JimBobSquarePants:
I think what you are seeing here is the difference in resampling algotithms + mathematical approach to the rotation.

Below is an example of three of our resamplers. Bicubic, Robidoux, and Spline, hopefully you can see the subpixel offsetting in the output. They're definitely using something closer to Spline anyway as you can see the smoothing on the patterned gradient

affine-differences

And zoomed.

affine-differences-zoomed

antonfirsov:
@JimBobSquarePants I think we can mark our output correct then, good news!
Tonight I gonna do my final review on tests + push the reference images.

antonfirsov:
@JimBobSquarePants just pushed some new test cases, and it looks like there are issues with the  `rectangle` parameter on `TransformProcessor`, or at least I would expect a different behavior.

There is a test case trying to define the target rectangle with a negative origo `(-5, -5)`. I expect this virtual position to be the origo of the transformed new image, so a `(+5,+5)` translation should be applied to the result, but there is no such translation.

I would find it very useful to define such virtual viewports:
it can handle the situations where a transform moves parts of an image to negative coordinates without applying an additional translation to my affine transform.

src/ImageSharp/Processing/Transforms/TransformHelpers.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        /// <summary>
+        /// Returns the bounding <see cref="Rectangle"/> relative to the source for the given transformation matrix.
+        /// </summary>
+        /// <param name="rectangle">The source rectangle.</param>
+        /// <param name="matrix">The transformation matrix.</param>
+        /// <returns>
+        /// The <see cref="Rectangle"/>.
+        /// </returns>
+        public static Rectangle GetTransformedBoundingRectangle(Rectangle rectangle, Matrix3x2 matrix)
```
antonfirsov: This method needs a review + extensive unit testing including cases having a rectangle with a negative origo.
I suggest to internalize it in this PR instead, so we can focus on the rest.

JimBobSquarePants: OK, will do.

JimBobSquarePants:
@antonfirsov 

Hmmmm. Many of the transforms will have a rectangles with negative origins (any rotate). Any offsetting we do would be the equivalent of going 0->width, 0->height would it not since both values reflect the offset?

That's what we are currently doing anyway.

tests/ImageSharp.Tests/Processing/Transforms/TransformTests.cs
```diff
@@ -142,13 +168,27 @@ public void Transform_WithSampler<TPixel>(TestImageProvider<TPixel> provider, st
...
+        private Matrix3x2 MakeManuallyCenteredMatrix<TPixel>(float angleDeg, float s, Image<TPixel> image)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            Matrix3x2 rotate = Matrix3x2Extensions.CreateRotationDegrees(angleDeg);
+            Vector2 toCenter = 0.5f * new Vector2(image.Width, image.Height);
+            var translate = Matrix3x2.CreateTranslation(-toCenter);
+            var translateBack = Matrix3x2.CreateTranslation(toCenter);
+            var scale = Matrix3x2.CreateScale(s);
+
+            Matrix3x2 m = translate * rotate * scale * translateBack;
```
JimBobSquarePants: I don't think this is correct. The last translation back will be incorrect due to pre-scaling. 

In order to calculate the correct matrix you would need to also alter the last translate back to take into consideration the new dimensions.

`var translateBack = Matrix3x2.CreateTranslation(toCenter * s); `

antonfirsov: @JimBobSquarePants this transform does not care with target image dimensions. An example:
If we have a 10x10 shape in the center of an 50x50 image and we want to scale it down by the factor of `0.5` keeping it in the center of the original 50x50 canvas, we move it to the origo (`(-25,-25)`), scale it, than move it back with a`(+25,+25)` vector.

In my tests the cropping is done manually by defining the new canvas "after" the transform.

antonfirsov:
@JimBobSquarePants pushed another case for `Transform_IntoRectangle`. With an origin of `(5,0)` and dimensions of `W=5,H=10` I expect to keep the right part of the transformed shape, not the left one. Currently the result for this case is the same as the result for `(0,0)` + `W=5,H=10`. This means that the rectangle origin is an ignored information, which makes no sense to me. 

A possible solution:
- We should replace `targetRectangle` with `Size targetDimensions` in `AffineProcessor<T>`, the `Rectangle` parameter in the extension method should be also replaced with a `Size`.
- We can define an additional extension method that takes a `Rectangle` and applies a `Matrix3x2.CreateTranslation(-rectangle.Location)` to the matrix. This variant should behave the way I expect from the current one.

src/ImageSharp/Processing/Processors/Transforms/AffineProcessor.cs
```diff
@@ -0,0 +1,344 @@
...
+namespace SixLabors.ImageSharp.Processing.Processors
+{
+    /// <summary>
+    /// Provides the base methods to perform affine transforms on an image.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    internal abstract class AffineProcessor<TPixel> : CloningImageProcessor<TPixel>
+        where TPixel : struct, IPixel<TPixel>
+    {
+        private Rectangle targetRectangle;
```
antonfirsov: The XY values of `targetRectangle` are actually never used!

JimBobSquarePants: Using a `Rectangle` allows us to use `Rectangle.Intersect` though and easily handle negative values. 

I'm not really following what your expectations are (I must be feeling slow), would it be possible for you to diagram them out? We should see what other libraries do also.

antonfirsov: No worries, it's easy to tell them in a few images. Will do it in the next few days.
The implementation should be also easy, could be done within an hour.

JimBobSquarePants: 👍 Great news, thanks!

antonfirsov:
@JimBobSquarePants okkay, here's a better explanation!

Let's say we have the following input image of dimensions `200x100`:
image

I'm going to apply the following transformation logic:
```C#
var m = Matrix4x4.CreateScale(2.0, 1.0);
var rectangle = new Rectangle(100, 0, 100, 100);
image.Mutate(i => i.Transform(m, KnownResamplers.Spline, rectangle))
```

And here is my point:
image

Gonna push the fix right into this PR, just need your approval :)

JimBobSquarePants:
Ah yes! That makes perfect sense now. Great explanation! :+1:

src/ImageSharp/Processing/Processors/Transforms/NonAffineTransformProcessor.cs
```diff
@@ -0,0 +1,238 @@
...
+using SixLabors.ImageSharp.PixelFormats;
+using SixLabors.Primitives;
+
+namespace SixLabors.ImageSharp.Processing.Processors
+{
+    /// <summary>
+    /// Provides the base methods to perform non-affine transforms on an image.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    internal class NonAffineTransformProcessor<TPixel> : InterpolatedTransformProcessorBase<TPixel>
```
antonfirsov: I'd rather call this `ProjectiveTransformProcessor` or `HomographyTransformProcessor`

src/ImageSharp/Processing/Transforms/Transform.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        /// <summary>
+        /// Transforms an image by the given matrix.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="source">The image to transform.</param>
+        /// <param name="matrix">The transformation matrix.</param>
+        /// <returns>The <see cref="Image{TPixel}"/></returns>
+        internal static IImageProcessingContext<TPixel> Transform<TPixel>(this IImageProcessingContext<TPixel> source, Matrix4x4 matrix)
```
antonfirsov: Good that it's internal for now. How about finishing projective transforms in a follow-up PR? I'd rather have this merged first, it's already *The PR That Has Been Under Review For The Longest Time In History* :)

src/ImageSharp/Processing/Processors/Transforms/NonAffineTransformProcessor.cs
JimBobSquarePants: Either work for me. I'll let you decide.

src/ImageSharp/Processing/Transforms/Transform.cs
JimBobSquarePants: Absolutely! Merge this as soon as you are satisfied with it.

src/ImageSharp/Processing/Transforms/Transform.cs
```diff
@@ -0,0 +1,118 @@
...
+            => Transform(source, matrix, sampler, Size.Empty);
+
+        /// <summary>
+        /// Transforms an image by the given matrix using the specified sampling algorithm.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="source">The image to transform.</param>
+        /// <param name="matrix">The transformation matrix.</param>
+        /// <param name="sampler">The <see cref="IResampler"/> to perform the resampling.</param>
+        /// <param name="sourceRectangle">The rectangle defining the source pixel area to transform. 'sourceRectangle.Location' becomes the origo of the transformed image.</param>
```
antonfirsov: @JimBobSquarePants I probably changed the semantics you originally intended for the `Rectangle` parameter. Check out the tests and my illustration, in [this comment](https://github.com/SixLabors/ImageSharp/pull/386#issuecomment-357104963).
If everyone's happy with it, we can merge the PR.

JimBobSquarePants: Does it though? With rotate/skew this should still mean the output image dimensions.

src/ImageSharp/Processing/Transforms/Transform.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <summary>
+        /// Transforms an image by the given matrix using the specified sampling algorithm
+        /// and a rectangle defining the transform origin in the source image and the size of the result image.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="source">The image to transform.</param>
+        /// <param name="matrix">The transformation matrix.</param>
+        /// <param name="sampler">The <see cref="IResampler"/> to perform the resampling.</param>
+        /// <param name="rectangle">
+        /// The rectangle defining the transform origin in the source image, and the size of the result image.
```
antonfirsov: @JimBobSquarePants I renamed it back to `rectangle`, I hope my comments explain it better now. If still unsatisfied & there is no better idea, we can internalize this overload for now. 
The one with the `destinationSize` parameter is that really matters.

JimBobSquarePants: Yeah that's better. 👍 I'll update the submodule and get this merged.

PR #393, opened 2017-12-07
Fix WuQuantizer output + optimizations

JimBobSquarePants:
### Prerequisites

### Description
I managed to break the Wu Quantizer with #384. This PR fixes that plus adds some optimizations to the code.


src/ImageSharp/Quantizers/WuQuantizer{TPixel}.cs
```diff
@@ -197,19 +196,21 @@ protected override void InitialQuantizePixel(TPixel pixel)
             var rgba = default(Rgba32);
             pixel.ToRgba32(ref rgba);
 
-            int r = rgba.R >> 2; // 8 - IndexBits
-            int g = rgba.G >> 2;
-            int b = rgba.B >> 2;
-            int a = rgba.A >> 5; // 8 - IndexAlphaBits
+            int r = rgba.R >> (8 - IndexBits);
```
JimBobSquarePants: This is where the bug was, I was adding the shifted component instead of the originals, the rest is optimization code.

src/ImageSharp/Quantizers/WuQuantizer{TPixel}.cs
```diff
@@ -328,12 +329,12 @@ private static float Volume(Box cube, long[] moment)
...
         /// <param name="moment">The moment.</param>
         /// <returns>The result.</returns>
-        private static long Bottom(Box cube, int direction, long[] moment)
+        private static long Bottom(ref Box cube, int direction, long[] moment)
         {
             switch (direction)
             {
                 // Red
-                case 0:
+                case 3:
```
JimBobSquarePants: Both the original C codebase and the C# port use ABGR ordering in their calculations. I changed it to RGBA in this one ages ago but that complicates debugging so I changed it back.

PR #404, opened 2017-12-18
Refactor and Enhance Color Matrix Filters

JimBobSquarePants:
### Prerequisites

### Description
This PR does several things.

- Adds adjustable matrix functionality to color matrix filters to match the floating point [specification](https://developer.mozilla.org/en-US/docs/Web/CSS/filter). Code is based on Chrome's [implementation](https://cs.chromium.org/chromium/src/cc/paint/render_surface_filters.cc).
- Fixes any inconsistencies between our original implementation and the spec.
- Replaces `ColorMatrixProcessor` with simpler `FilterProcessor`
- Add the factory class `MatrixFilters` to allow reuse of `Matrix4x4` color matrix algorithms 
- Add the extension `Filter` which accepts a free-form `Matrix4x4` as a parameter so filters can be easily combined into a single operation.

``` c#
using (Image<TPixel> image = Image.Load(instream))
{
    Matrix4x4 brightness = MatrixFilters.CreateBrightnessFilter(0.9F);
    Matrix4x4 contrast = MatrixFilters.CreateContrastFilter(1.2F);
    Matrix4x4 saturation = MatrixFilters.CreateSaturateFilter(1.5F);
    image.Mutate(x => x.Filter(brightness * contrast * saturation));
    image.Save(outstream);
}
```

### Notes
I've had to temporarily disable the verification code from the two `DetectEdgesTest` tests dues to changes in the grayscale processor which no longer uses companding to match the output from Chrome etc. Once we can update those I can reimplement.


tests/ImageSharp.Tests/Processing/Processors/ColorMatrix/FilterTest.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using SixLabors.Primitives;
+using Xunit;
+
+namespace SixLabors.ImageSharp.Tests.Processing.Processors.Filters
+{
+    public class FilterTest : FileTestBase
+    {
+        [Theory]
+        [WithFileCollection(nameof(DefaultFiles), DefaultPixelType)]
```
antonfirsov: I think we should stop using `DefaultFiles` in new test cases & keep it for legacy tests only.
For filters, I suggest to test using `[WithTestPatternImages]` + probably `CalliphoraPartial.png` to have one realistic photo.

tests/ImageSharp.Tests/Processing/Processors/ColorMatrix/FilterTest.cs
```diff
@@ -0,0 +1,48 @@
...
+        public void ImageShouldApplyFilter<TPixel>(TestImageProvider<TPixel> provider)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            using (Image<TPixel> image = provider.GetImage())
+            {
+                Matrix4x4 brightness = MatrixFilters.CreateBrightnessFilter(0.9F);
+                Matrix4x4 contrast = MatrixFilters.CreateContrastFilter(1.2F);
+                Matrix4x4 saturation = MatrixFilters.CreateSaturateFilter(1.5F);
+                image.Mutate(x => x.Filter(brightness * contrast * saturation));
+                image.DebugSave(provider);
```
antonfirsov: If the color filters are correct (which I assume is true), it's safe to add reference images + a `CompareToReferenceOutput()` call here.

src/ImageSharp/Processing/Processors/Filters/FilterProcessor.cs
```diff
@@ -0,0 +1,62 @@
...
+                endY,
+                configuration.ParallelOptions,
+                y =>
+                    {
+                        Span<TPixel> row = source.GetPixelRowSpan(y);
+
+                        for (int x = startX; x < endX; x++)
+                        {
+                            ref TPixel pixel = ref row[x];
+                            var vector = Vector4.Transform(pixel.ToVector4(), matrix);
```
antonfirsov: Optimization opportunity (absolutely not urgent, we can postpone it to a follow-up PR):
Create a temporal `(endX-startX)` -sized `Buffer<Vector4>`, and use `PixelOperations` to pack+unpack.

src/ImageSharp/Processing/Processors/Filters/FilterProcessor.cs
```diff
@@ -0,0 +1,62 @@
...
+        protected override void OnApply(ImageFrame<TPixel> source, Rectangle sourceRectangle, Configuration configuration)
+        {
+            var interest = Rectangle.Intersect(sourceRectangle, source.Bounds());
+            int startY = interest.Y;
+            int endY = interest.Bottom;
+            int startX = interest.X;
+            int endX = interest.Right;
+            Matrix4x4 matrix = this.Matrix;
+
+            Parallel.For(
```
antonfirsov: `Parallel.For` ... I can't live ... with or without you ...

src/ImageSharp/Processing/Processors/Filters/FilterProcessor.cs
JimBobSquarePants: Does this give much of a performance boost? Didn't know if it would be worth the allocation.

src/ImageSharp/Processing/Processors/Filters/FilterProcessor.cs
JimBobSquarePants: I know 🙁

tests/ImageSharp.Tests/Processing/Processors/ColorMatrix/FilterTest.cs
JimBobSquarePants: They're all definitely accurate. I just wanted to wait til We have this in before updating/adding test images.

tests/ImageSharp.Tests/Processing/Processors/ColorMatrix/FilterTest.cs
JimBobSquarePants: Will update.

src/ImageSharp/Processing/Processors/Filters/FilterProcessor.cs
antonfirsov: Bulk packing/unpacking is much faster now for `Rgba32`.

`Buffer<T>` uses a pooled array normally, so it shouldn't be an issue to construct some of them. However, we should not produce an instance for each row, and `Parallel.For` prevents us to new it up it in the outer scope. I have an idea to deal with all these `Parallel.For`-s, but this issue is at the very end of my queue now :(

src/ImageSharp/Processing/Processors/Filters/BlackWhiteProcessor.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// Applies a black and white filter matrix to the image
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    internal class  BlackWhiteProcessor<TPixel> : FilterProcessor<TPixel>
+          where TPixel : struct, IPixel<TPixel>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BlackWhiteProcessor{TPixel}"/> class.
+        /// </summary>
+        public BlackWhiteProcessor()
```
antonfirsov: I wonder if we still need these subclasses?

antonfirsov:
@JimBobSquarePants just made the output images smaller + grouped the output of all filter test classes into a subfolder. If you are happy with the results, I think we can push them to the reference images repository. Note: There is no need for reference images for the `******InBox` variants,

src/ImageSharp/Processing/Processors/Filters/BlackWhiteProcessor.cs
JimBobSquarePants: Some yes, some no. I'm happy to drop the ones we don't leaving only the ones that require extra parameters.

JimBobSquarePants:
@antonfirsov Yeah, happy to see that moved to reference images. I'm 100% confident that the algorithmic output is good.

antonfirsov:
How did I enable StyleCop for `ImageSharp.Tests`? :scream:

antonfirsov:
OK, it was just a small issue in the `ImageSharp` project.

antonfirsov:
Managed to provide proper coverage from only 35.6 KB of test images!

It's really not hard to add these tests now, would be nice if we could add this to the list of responsibilities of PR openers. Can I somehow help to turn this into a rule? Maybe we need an official testing guide + batch scripts to copy images?

JimBobSquarePants:
@antonfirsov I don't understand what the thresholding is doing now. Why is the difference negative? why is it not 0 since we are comparing with our own output? How are the test images generated?

antonfirsov:
@JimBobSquarePants Looks like there are numeric (or reference decoder?) errors under Linux, resulting in differences between the calculated and the expected images, causing the travis build to fail. The hacky stuff you see is a workaround by introducing a relatively higher threshold when detecting Linux.

The trick is ugly and complicated, so I gonna undo it, and skip the `CompareToReferenceOutput()` step on Linux instead. Having the verification done on appveyor-only should be enough to catch regressions.

The reference output is the copy of our own, but in order to allow changes in the implementation, we should always use a (very small) tolerance when testing floating point algorithms.

antonfirsov:
Does it look better now? :)

JimBobSquarePants:
@antonfirsov Much better yeah! I'm gonna merge this now. We should chat sometime  about these Linux issues in Slack to see if we can figure out what is going on.

PR #408, opened 2017-12-20
Update build number generation stratagy

tocsoft:
Due to issues with gitversion not numbering our CI builds to our liking (ie resetting the build counter on a tagged release)  then I've updated the build scripts to no longer use gitversion and instead have written a powershell script that will emulate the gitversion features where where using.

version numbers will be generated using one these of the 3 strategies;

1. tagged builds
   1. if a tagged build happens use the tag as the version number (with auto padding of number to keep lexagraphic sorting working)
2. branch
   1. get the build counter from appvayor and pad it to 6 characters
   2. take the last used tag to get the base version number (X.X.X part)
   3. append a cleaned branch name (master is replaced with 'dev') 
   4. append the padded counter
3. PR builds
   1. get the build counter from appvayor and pad it to 6 characters
   2. take the last used tag to get the base version number (X.X.X part)
   3. append 'PullRequest(PR number)_'
   4. append padded counter


tocsoft:
My first attempt was to try older version of gitversion but It didn't seem to make a difference and after careful looking at gitversion diagnostic output the counter reset seemed correct it just not the effect we wanted.

JimBobSquarePants:
Impressive! 

How did you test this without banging your head against the Appveyor brick wall?

tocsoft:
it was copious amounts of altering local environment variables... and hoping that it ends up working 😉

Issue #423, opened 2018-01-01
"Value should be an unsigned short" exception on image with EXIF rotate.


musukvl:
### Description
The image I want to resize has EXIF and ImageSharp tries to update it with new Width/Height values on rotating.
I have "Value should be an unsigned short" on following code:
```
var input = @"G:\temp\temp-img\1.jpg";
using (var image = SixLabors.ImageSharp.Image.Load(input))
{
       image.Mutate(x => x
            .Rotate((float)90.0)); // Exception goes here
       image.Save(input); 
}
```
![1](https://user-images.githubusercontent.com/3954761/34471309-d6e17a3e-ef56-11e7-8ff1-135a3961998a.jpg)

### Analytics
Looks like problem in RotateProcessor
```
 profile.SetValue(ExifTag.PixelXDimension, source.Width);
```
It expects source.Width as `ushort`, but Image.Width defined as `int`:  
`public int Width => this.frames.RootFrame.Width;`

![image](https://user-images.githubusercontent.com/3954761/34471342-fd01a832-ef57-11e7-93bf-cc2c4613755d.png)

![image](https://user-images.githubusercontent.com/3954761/34471386-24a760a6-ef59-11e7-9d20-57409b920c13.png)

My suggestion is to cast PixelXDimension from int to ushort. It fixes problem but I'm not sure it is the best solution.

### Steps to Reproduce

Use 1.jpg attached and try to rotate it.

### System Configuration

- ImageSharp version: 1.0.0-dev000621   1.0.0-beta0002   
- Other ImageSharp packages and versions: master(6f438dac852c5d7408161e1afa034984671257f4)
- Environment (Operating system, version and so on): Windows 10
- .NET Framework version: 2.0
- Additional information:

### Exception: 

```
System.ArgumentException occurred
  HResult=0x80070057
  Message=Value should be an unsigned short.
Parameter name: value
  Source=<Cannot evaluate the exception source>
  StackTrace:
   at SixLabors.ImageSharp.Guard.IsTrue(Boolean target, String parameterName, String message) in G:\ext\xxxxx\ImageSharp\src\ImageSharp\Common\Helpers\Guard.cs:line 212
```

jsmarcus:
I'm also having this problem.

PR #424, opened 2018-01-01
#423. "Value should be an unsigned short" exception on image rotate

musukvl:
Added safe Int Width and Height values cast to ushort. ExifTag.PixelXDimension and  ExifTag.PixelYDimension will not be updated if it is impossible to cast int to ushort without truncation.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Transforms/RotateProcessor.cs
+++ b/src/ImageSharp/Processing/Processors/Transforms/RotateProcessor.cs
@@ -105,10 +105,10 @@ namespace SixLabors.ImageSharp.Processing.Processors
 
             profile.RemoveValue(ExifTag.Orientation);
 
-            if (this.Expand && profile.GetValue(ExifTag.PixelXDimension) != null)
+            if (this.Expand && profile.GetValue(ExifTag.PixelXDimension) != null && source.Width <= ushort.MaxValue && source.Height <= ushort.MaxValue)
             {
-                profile.SetValue(ExifTag.PixelXDimension, source.Width);
-                profile.SetValue(ExifTag.PixelYDimension, source.Height);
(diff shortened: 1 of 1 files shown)


antonfirsov:
@dlemstra can you confirm if this change is correct and we should merge it?
@musukvl can you sign the CLA so we can merge the change? :)

> Alexey Ryazhskikh seems not to be a GitHub user. 

I don't understand why does CLAassistant claim this :(

/cc @JimBobSquarePants @tocsoft

tocsoft:
It's because the commit has an email address against it that hasn't been linked  to a github user account.

dlemstra:
This is not correct. The value of `ExifTag.PixelXDimension` can either be a `short` or a `long`.

@musukvl It might be better to remove the current exif tag and add a new exif tag that is either a short or a long depending on the value,

JimBobSquarePants:
Hold on! So what is actually causing the error and why can we not use an int to set the value?

The issue here seems to be a larger issue on of a lack of type safety within the API. If the following will throw at runtime then it shouldn't be possible.

``` c#
profile.SetValue(ExifTag.PixelXDimension, source.Width); 
profile.SetValue(ExifTag.PixelYDimension, source.Height); 
```

JimBobSquarePants:
@dlemstra I wonder, do you think it is possible to solve our `ExifValue` needing to support multiple `ExifDataType` values by using flags? 

We could then combine the values like `ExifDataType.Short | ExifDataType.Long` and support both lengths.

dlemstra:
@JimBobSquarePants We should probably add support for that but not sure yet how to implement that. Might have time to look at that this weekend.

JimBobSquarePants:
Closing in favour of #448 

Thanks @musukvl for your help! 👍

Issue #225, opened 2017-05-25
Implement pluggable memory management


antonfirsov:
### Problem
- In some use cases the default `ArrayPool` behavior does not fit user needs (see #222). Users should be able to  configure + reset + disable pools.
- Some users may want to back `Image<T>` by memory resources other than managed memory. (Eg. native buffers).

### Solution
Implement a pluggable `MemoryManager`, make the default pooling memory manager configurable.

**Update:**
[feature/memory-manager](https://github.com/SixLabors/ImageSharp/tree/feature/memory-manager) is the WIP branch for this refactor. It's having #431 merged. If anyone wishes to contribute, please file your PR-s against that branch. The tasks below are being updated according to the progress on that branch.

#### Tasks for beta-3
- [x] Refactor all raw `ArrayPool` and `PixelDataPool` usages to use `Buffer` and `Buffer2D` instead.
- [x] `Buffer2D<T>` should compose `Buffer<T>` (as `IBuffer<T>`) [instead of inheriting it](https://en.wikipedia.org/wiki/Composition_over_inheritance)
- [x] Move the responsibility of providing `Buffer<T>` instances to a generic MemoryManager class. Refactor all `new Buffer<T>` and `new Buffer2D<T>` instantiations to call factory method on `MemoryManager` instead. 
- [x] We need to cover all the encoders/decoders with proper regression tests before continuing the refactor work. (Working on this - would be filed as a separate PR)
- [x] Extensive test coverage for `ArrayPoolMemoryManager`
- [x] Allocations in `DefaultPixelBlenders<TPixel>` should use `MemoryManager` taken from the outside
- [x] Replace the legacy `PixelArea<TPixel>` class with `BufferArea<T>` + extensions methods working on `BufferArea<TPIxel>` and/or `IBuffer2D<TPixel>` using `PixelOperations<TPixel>`
- [x] Replace ~all~ most `PixelAccessor<TPixel>` usages with `Buffer2D<T>`. ~Drop the `IBuffer2D<T>` interface.~ - the full removal could be done later.
- [x] `MemoryManager` should construct `IBuffer<T>` and `IManagedByteBuffer` instead of `Buffer<T>`, codecs should use `IManagedByteBuffer.Array`
- [x] Based on user feedback, there also might be some strange, unexpected leaking when images are created/destroyed continously. **Doing some load testing, identifying+understanding this behavior is inevitable!**

#### Probably 1.0
- [x] Passing custom `MemoryManager` to codecs

[...]

PR #431, opened 2018-01-11
WIP: swappable memory pools

rytmis:
### Prerequisites

### Description

This is very much a work in progress -- so far, I've moved Buffer construction to use a singleton MemoryPool with an ArrayPool backed implementation.


antonfirsov:
@JimBobSquarePants check out #225 and Gitter conversations if you want (+have some time) to keep track of this! :)

JimBobSquarePants:
Will do! :smile:

src/ImageSharp/Memory/ArrayPoolMemoryManager.cs
```diff
@@ -30,16 +30,18 @@ public ArrayPoolMemoryManager(int minSizeBytes = 0)
         }
 
         /// <inheritdoc />
-        internal override Buffer<T> Allocate<T>(int size, bool clear = false)
+        internal override Buffer<T> Allocate<T>(int itemCount, bool clear = false)
```
JimBobSquarePants: Let's use two distinct methods here with overloads. I'm not a big fan of using optional parameters in inherited methods; it can often lead to breaking changes since this could well become a public API.

src/ImageSharp/Configuration.cs
```diff
@@ -83,6 +83,11 @@ public Configuration(params IConfigurationModule[] configurationModules)
         /// </summary>
         public IEnumerable<IImageFormat> ImageFormats => this.imageFormats;
 
+        /// <summary>
+        /// Gets or sets the <see cref="MemoryManager"/> that is currently in use.
+        /// </summary>
+        public MemoryManager MemoryManager { get; set; } = new ArrayPoolMemoryManager(1024 * 80);
```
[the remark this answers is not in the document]
antonfirsov: @rytmis `ArrayPool` is actually really useful in small-mid dynamic array allocation use cases, because it reduces GC pressure + improves data locality!
Check out [it's usages in aspnet mvc](https://github.com/aspnet/Mvc/search?utf8=%E2%9C%93&q=JsonOutputFormatter&type=). 

An example: it's [providing](https://github.com/aspnet/Mvc/blob/b196708a3e0bca4820f2002e063fabe979afde1b/src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs#L48) `ArrayPool<char>` to Newtonsoft.Json [through it's own `IArrayPool<T>` abstraction](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/JsonTextWriter.cs#L72).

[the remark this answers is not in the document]
rytmis: I changed the default value to 512 bytes, but obviously you can decide what you want to use for it -- I don't know what benchmarks to run to find the optimal values, and so on, so I'll just defer to your judgment. 🙂

src/ImageSharp/Formats/Jpeg/GolangPort/Components/Decoder/OrigHuffmanTree.cs
```diff
@@ -246,19 +221,30 @@ public void Dispose()
...
+            fixed (int* lutPtr = this.Lut)
+            {
+                return lutPtr[index];
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public int GetMaxCode(int index)
+        {
+            fixed (int* maxCodesPtr = this.MaxCodes)
```
antonfirsov: `GetMaxCode()`  gonna pin/unpin `this.MaxCodes` several times inside a loop. We need to find a better way. Pinning outside the loop, or applying [`ref` arithmetics with `SRCS.Unsafe`](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Formats/Jpeg/Common/Block8x8.cs#L52) could do the job.

Do you want me to deal with this?

src/ImageSharp/Memory/NullMemoryManager.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace SixLabors.ImageSharp.Memory
+{
+    /// <summary>
+    /// Implements <see cref="MemoryManager"/> by allocating new buffers on every call.
+    /// </summary>
+    public class NullMemoryManager : MemoryManager
```
antonfirsov: I think we need to find a better name for this :) `NullMemoryManager` suggests me, that it's always returning `null` instead of a memory block with a valid memory address.

@JimBobSquarePants @tocsoft suggestions?

tocsoft: `SimpleMemoryManager` maybe?

src/ImageSharp/Common/Extensions/StreamExtensions.cs
```diff
@@ -29,23 +29,16 @@ public static void Skip(this Stream stream, int count)
             }
             else
             {
-                byte[] foo = ArrayPool<byte>.Shared.Rent(count);
-                try
+                byte[] foo = new byte[count];
```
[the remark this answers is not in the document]
JimBobSquarePants: Agreed.

src/ImageSharp/Memory/NullMemoryManager.cs
vpenades: What about ManagedMemoryFactory and PooledMemoryFactory ?

vpenades:
As far as I understand how this configurable memory manager works, it is designed to set a global memory manager before any other call, and then it uses that memory manager on everything; from requesting memory for image buffers to small streaming arrays.

My questions are:

- What happens if you switch memory manager on the fly, when you've already allocated images, or worse, when there's some asynchronous background image processing running.
- It is possible to do this?: To use a "Null" memory manager to hold Image's buffers, but to use ArrayPools locally bounded to the context of encoders/decoders and processors.

If given the choice, I would like to configure ImageSharp like this:
- Images will always use GC managed memory, even for small images.
- Encoders, Decoders and processos would use ArrayPools associated to their lifetime, for example, a JpegEncoder would create an internal ArrayPool to reuse arrays within the context of encoding a given image, but when it's done, the ArrayPool is GC'ed along with the encoder.

Example:
```
var image = new Image<Rgba32>(512,512); // memory held by standard GC'ed array
image.Mutate( dc => dc.Resize(256,256).Crop(48,48).GaussianBlur(2) ); // dc using local ArrayPools, GC'ed at the end of the mutation
image.Save("test.png"); // png encoder using ArrayPools internally, again, GC'ed at the end of encoding
// no need to dispose image since it's buffers are GC managed.
```

In this case, the image's internal buffer would be GC managed, but IImageProcessingContext would have a non static PooledMemoryManager used by all the processors, but GC'ed when the "mutation" is completed.

For me this is the ideal behaviour, because in my case it's a combination of performance, ease of use and guaranteeing when I've finished, all memory used by ImageSharp will be GC'ed

rytmis:
@vpenades, the long-term idea as @tocsoft outlined it is to have "usage hints" at the call sites. This way, you could configure a single composite memory manager that would work the way you describe, since the usage hints for encoders would be different than those used for images.

antonfirsov:
@vpenades

#### re: different memory usage scenarios
The hard part is the context-bound automatical release of pools after each operation, but maybe we can figure this out. It's very easy however to provide a safe API for manual release of these resources: 
When you are finished with all the image manipulation work, you can just call `MemoryManager.Default.ReleaseRetainedResources()`. (It should not affect images/buffers being in use.)

#### re: MemoryManager consistence:
> What happens if you switch memory manager on the fly, when you've already allocated images, or worse, when there's some asynchronous background image processing running.

My suggestion (could work with both `Configuration.Default` and user-managed `Configuration`-s:
```C#
public class Configuration
{
    ...
    
    public IMemoryManager MemoryManager 
    {
    	get { ... } 
    	set 
    	{ 
    		if (this.memoryManager.IsTouched) throw new InvalidOperationException();
    		this.memoryManager = value;
    	}
    }
}
```

vpenades:
@antonfirsov my concern with static configurations is that I'm working with a plugin based system (I've pointed you to [it](https://github.com/vpenades/UberFactory) before) in which I can have more than one ImageSharp based plugin loaded at the same time. (In the scenarios I have in mind, there will be lots!)

So I have these conflicting scenarios when I have two independently developed plugins being loaded at the same time:

- Both plugins trying to configure the memory manager independently: the second setup attempt crashes.
- Both leaving the configuration setup to the host application. In this case, I have to make the host application "ImageSharp aware" which defeats the purpose of being plugin based.

This can also happen with third party libraries extending or wrapping imagesharp, and willing to use it with a specific configuration; whenever you try to use libraries from different sources, we can end having conflicting configurations.

The only way I can think to solve this is to break the APIs and create objects through non static factory instances, but I guess it's a change too big to be considered.

antonfirsov:
@vpenades your scenario is exactly the reason why usin `Configuration.Default` is not mandatory at all! You can have as many instances of the `Configuration` object, thus `MemoryManager` as you want. Just make sure, you use the proper `Image.Load(configuration, etc ...)` overloads.
If it was only up to me to decide, I would maybe even drop the singleton, because I consider it an antipattern, but it would complicate the API, thus raise the barrier of entry for new users.

antonfirsov:
Calling `Configuration.Default.MemoryManager.Anything()` was only demonstrational in my examples. You can operate on the Configuration/MemoryManager instance of your choice.

vpenades:
@antonfirsov mybe the singleton can be dropped if we rename things: instead of calling it "Configuration" let's called "Factory", so the new API would look like this:

```
var image1 = PooledFactory.Load("test.png");
var image2 = GCManagedFactory.CreateImage<Rgba32>(512,512);
```

I think it's quite a huge change, but it's not that ugly. In fact, having the images created from a factories would help future works when we require "interop" or file mapped images, which could be supported transparently.

antonfirsov:
Hey @rytmis! Do you have interest/time to continue working on this?
If you can't make it, I'd prefer pulling your work into a separate branch, finish it there, and open a new PR in the end.

Gonna have time for this in the middle of the next week.

rytmis:
Interest? Sure. Time? Not so sure. Skills? Probably not. :)

If you need someone to do boring grunt work, then I'm your guy, obviously -- but I don't think pulling this to a separate branch is necessarily a problem for that, so do whatever is easiest for you. I'm just glad to have helped at least a little bit.

antonfirsov:
You already helped a lot, thanks!

I have some time, so I'll grab it from now then. In order to stick to my personal plans I need to finish it before the end of the next week. Probably gonna ask for some input.

Don't talk nonsense about your skills, you're doing cool stuff according to your blog and GH contribution history! 😄

antonfirsov:
Closed this in favor of #475

Issue #409, opened 2017-12-21
Out of Memory Exception when resizing image


robertbaker:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
```
Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Microsoft.Azure.WebJobs.Host.FunctionInvocationException : Exception while executing function: ImageProcessorFunctions.ProcessImageConversion ---> SixLabors.ImageSharp.ImageProcessingException : An error occured when processing the image using DelegateProcessor`1. See the inner exception for more detail. ---> SixLabors.ImageSharp.ImageProcessingException : An error occured when processing the image using ResizeProcessor`1. See the inner exception for more detail. ---> System.OutOfMemoryException : Exception of type 'System.OutOfMemoryException' was thrown.
 at System.Buffers.ConfigurableArrayPool`1.Rent(Int32 minimumLength)
 at SixLabors.ImageSharp.Memory.Buffer`1..ctor(Int32 length)
 at SixLabors.ImageSharp.Processing.Processors.ResizeProcessor`1.OnApply(ImageFrame`1 source,ImageFrame`1 cloned,Rectangle sourceRectangle,Configuration configuration)
 at SixLabors.ImageSharp.Processing.CloningImageProcessor`1.CloneAndApply(Image`1 source,Rectangle sourceRectangle) 
 End of inner exception
 at SixLabors.ImageSharp.Processing.CloningImageProcessor`1.CloneAndApply(Image`1 source,Rectangle sourceRectangle)
 at SixLabors.ImageSharp.Processing.CloningImageProcessor`1.Apply(Image`1 source,Rectangle sourceRectangle)
 at SixLabors.ImageSharp.DefaultInternalImageProcessorContext`1.ApplyProcessor(IImageProcessor`1 processor,Rectangle rectangle)
 at SixLabors.ImageSharp.DefaultInternalImageProcessorContext`1.ApplyProcessor(IImageProcessor`1 processor)
[...]

Issue #325, opened 2017-09-08
ImageSharp buffers take up too much memory space


HasanAyan:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I am only resizing and placing single rectangle to a PNG image (725kb). I checked my code and I dispose eveything related to ImageSharp after I am done. Still, I see (dotMemory) that ImageSharp use using a lot of memory

![dotMemory ss](https://i.imgur.com/7FGdlHB.png)

The first time I run my application, I get 5 of these objects and each time I run the application again it adds new one.

Not sure if there is a problem here or not but I would like to know if I am doing something wrong.

### System Configuration

- ImageSharp version: 1.0.0-alpha9-00194
- Other ImageSharp packages and versions: ImageSharp.Drawing 1.0.0-alpha9-00194
- Environment (Operating system, version and so on): windows 10 creators update
- .NET Framework version: .netcore 2.0
- Additional information:

JimBobSquarePants:
Can you supply a code sample please so we can replicate?

HasanAyan:
Sure. 

I need to correct something; it doesn't add a new buffer object each time. It creates 5 when I run it first, adds another one on the second run and it seems to preserve the count for the continuing runs (I tested 4 consequetive runs)

This is the first part of my code, where I basically change the size and try to minimize the size. In my test 
I run this for two different images (~720kb each)

```


public static byte[] OptimizeImage(MemoryStream rawImage)
        {

            rawImage.Position = 0;
            Image.Load(rawImage).To<Rgb24>().Resize(new ResizeOptions()
            {
                Mode = ResizeMode.Max,
                Size = new Size(750, 400),

            }).Save(rawImage, new PngEncoder()
            {
                CompressionLevel = 9,
                PngColorType = PngColorType.Rgb,
                IgnoreMetadata = true,
            }).Dispose();


            return rawImage.ToArray();
        }

```

The second part is where I use copy of one of the two images I optimized before and add a filled rectangle at some coordinates (between 1 and 4, but 90% of times only one rectangle). I call this function around 140 times.

I


```
[...]

Issue #151, opened 2017-03-24
ImageSharp.Formats.Jpg.DecodedBlock ArrayPool is using up WAY too much memory


DeCarabas:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I'm basically reporting issue #123 again, but with a slightly different spin. I'm running my little web app on a VM with 1GB of RAM, and I'm constantly getting killed with out of memory because of the ArrayPool being used for ImageSharp.Formats.Jpg.DecodedBlock[]. 

My app concurrently fetches RSS feeds, then for each entry in each RSS feed concurrently fetches all the images and evaluates all the images for a given entry to see which one will make the best thumbnail. I don't need to get into the large number of feeds before I'm decoding a lot of images in parallel.

After a single fetch session, I dumped the process and examined the heap; I discovered that ImageSharp.Formats.Jpg.DecodedBlockArray.ArrayPool was holding on to 337MB all on its own; the largest entry was 262,144 elements long for a size of 70,254,616. (Obviously that's just the current state of the pool; as currently configured by ImageSharp it can grow way past that.)

I need some way to get a handle on this memory usage; I'm looking at other constraints but some kind of configuration to let me limit the pool would be really nice.

### Steps to Reproduce
N/A

### System Configuration

- ImageSharp version: 1.0.0 alpha 4
- Environment (Operating system, version and so on): linux 
- .NET Framework version: .net core 1.1

antonfirsov:
@DeCarabas Thank you for your detailed report!

@JimBobSquarePants based on these reports it seems obvious to me, that we need to make our pooling configurable. My proposal:
https://gist.github.com/antonfirsov/25301496604015fbfe7f2b90d608a190

tocsoft:
[...]

PR #436, opened 2018-01-15
PixelDataPool<T>: reduce maximum pooled array size

antonfirsov:
### Prerequisites

### Description
It might be a longer process to finish #431, however users experiencing `OutOfMemoryException`-s need a stable ImageSharp for yesterday :) In this PR I try to provide a solution for memory issues reported in #225, #409, #325, #224 and #151 by a simple change:
- The maximum pooled buffer size is now limited to 32MB (equivalent to 8 megapixels of raw `Rgba32` data). It's 32-times lower for `Image<Rgba32>` than before. I think most images are going to fit into this size without eating up the memory of the process while *still using pooled memory*. For outliers however, it might be better new-up those arrays. (*You can still run out of usable memory because of LOH fragmentation issues! I suggest to touch [`GCSettings.LargeObjectHeapCompactionMode`](https://blogs.msdn.microsoft.com/mariohewardt/2013/06/26/no-more-memory-fragmentation-on-the-net-large-object-heap/) regularly in environments with strict memory limits!*)

- The same byte limit is used for non-`IPixel` value type arrays. This should probably fix the issues caused by pooled `Block8x8` buffers in JpegDecoder.

We'll keep working on #431 to provide a full-scale solution. @rytmis already did the majority of the work, big thanks for that! (However we still need to fix the array conversion issues, provide proper test coverage, polish the API further before merging it.)

/cc @denisivan0v @xakep139 @vpenades @JulianRooze


denisivan0v:
Many thanks @antonfirsov, but in my cases it's still a big amount of allocated memory since in [`ArrayPool<T>.Create(MaxArrayLength, 50)`](https://github.com/SixLabors/ImageSharp/pull/436/files#diff-de77e1fc781805efd2c8375e1356c6d4R30) `maxArraysPerBucket = 50` and we end up with 32Mb * 50 * number of buckets of pooled memory.

I'm looking forward to #431 and now rely on these changes to do my tests.
It's also can be very useful to have an API to check image metadata (see https://github.com/SixLabors/ImageSharp/pull/292#issuecomment-357602218) without decoding it and make these allocations.

rytmis:
I've read the ArrayPool API docs and the DefaultArrayPool implementation. Those parameters mean that you've got up to 50 32MB arrays, up to 50 16MB arrays and so on -- the first argument being the max size _within a bucket_ and the second being the maximum number of "similarly-sized" arrays within a single bucket of the pool. The size selection of DefaultArrayPool for a bucket works in powers of two, and you'll always get the size that's the smallest available but at least as large as you asked for.

Out of curiosity, I ran the bucket count and bucket size algorithms for a DefaultArrayPool with those parameters, and the absolute maximum retained size would be just over 3 GB if every single array was in use. Of course I know that it's extremely unlikely that any allocation pattern will hit very close to that mark, but even a 50% fill factor will swallow a large portion of the memory space in 32 bit or in a low-memory container.

Add to this the fact that you've got one ArrayPool for every TPixelData you use within the app: for instance, if all you do is construct Rgba32 images from Jpegs, you'll have one ArrayPool of Block8x8's and another of Rgba32's. -- this is the exact case I had when I came along. This means that the max retained size for the pools is north of 6GB.

I'd suggest that instead of dropping the max size of an individual array, you'd drop the max bucket count instead -- or, possibly, both. If you try to allocate beyond what the pool is capable of holding, it will just return a regular heap-allocated array instead.

(Caveat: I'm currently undergoing a case of the Man Flu, so there's a possibility that I've messed up my math at some point. If you wish to double-check, the sources are at https://github.com/dotnet/corefx/blob/master/src/System.Buffers/src/System/Buffers )

antonfirsov:
@rytmis good experiment and useful calculations! Gonna reduce `maxArraysPerBucket` then, but I think, in usual cases it is very unlikely to get even close to those theoretical limits, specially with the largest 32MB buckets.

I'm also unsure about my math, but if your service is kinda stateless (so buffers are always returned to the pool when a request is finished), you need to stress your application in a way so it reaches a maximum peek of ~10 **concurrent** `~4000x2000` sized image uploads to approach the `50%` `ArrayPool` fill factor.

For me it feels that if this is a realistic scenario in your environment then:
- It is very likely that your gonna experience memory issues even without `ArrayPool` because of [LOH fragmentation](https://blogs.msdn.microsoft.com/maoni/2016/05/31/large-object-heap-uncovered-from-an-old-msdn-article/): The .NET GC [has not been designed](https://github.com/dotnet/corefx/issues/5921#issuecomment-181970876) to follow such a stress.
- For image processing applications, you should really consider using a less limited container, or use load balancing techniques to reduce the load of your individual containers, regardless of the image processing framework you use.

But instead of guessing, I really want to model this by defining proper load testing benchmarks. It's not easy for me to make them realistic, so any input and review would be really appreciated!

rytmis:
My scenario is a batch processor that will download and convert images according to a number of different presets. After each resize, the resized image is stored on disk and the image itself is disposed, so that _should_ mean having _at most_ two images in memory at a time. No concurrent processing.

After running a batch where there were lots of sequential resizes happening, my memory usage peaked somewhere around 1,5 GB and the app began to slow down (I'm guessing GC and disk thrashing due to page file usage).

Note that this is a scenario that has no such issues with System.Drawing. I'm not talking hypotheticals here: what I did was take an existing batch app that targeted the desktop framework and ported it to run on CoreFX, replacing System.Drawing with ImageSharp. The app has been running for over two years without any memory-related issues, but the ported version choked almost immediately.

The main problem the current pool approach is that if you happen to load a single image that is, say, 90 MB in size, you'll _permanently_ retain 128 MB of memory in that pool bracket, even when it has been released back to the pool. If you then load a 130 MB image, you'll permanently retain 256 MB. So going through a series of images that get ever larger, you'll eventually exhaust the heap space even when you'd only need enough capacity for the one you are currently working with. 

In other words, if you ever, at any point, allocate _and then release_ one array from each bucket up until you reach the 130 MB mark, you will have retained at least half a gigabyte until the app terminates. Multiply that by the number of pixel formats, and... you get the picture (pun not intended).

(These calculations assume byte arrays -- I'm not sure, but for TPixel arrays this might change the constants by a little bit. In any case, it's still the same order of magnitude).

vpenades:
@rytmis That's pretty much the same scenario I have. I still believe that using pools for caching lots and lots of very small, similar sized arrays, like handling server messages with lots of small POCOs, it's fine....

But images are too large, if the pool's memory cannot be completely disposed by the GC, or at will, there's many scenarios in which you can end up with your application running with less than half of available memory.

antonfirsov:
@rytmis I think you really need outliers of size `>128MB` to reach that 1.5 GB peek. Now that the maximum pool size is reduced significantly (while still covering typical image sizes), I expect your application's memory usage to be stabilized at much lower levels.

Whether `ArrayPool` is good for imaging or not:
I'm on the strong opinion, that there *always exists an `L > 0` upper limit* for the pool size which leads to better throughput with pooling enabled than without it. In practice, the memory used by the pool should stop growing after a certain number of requests, way beyond it's theoretical limit:

image

I'm quite sure I will be able proove this later with benchmarks.

This limit might be lower for certain applications + scenarios, pretty high for others, but *never zero*, if you optimize for throughput. It's really unfortunate, that bugs we made in ImageSharp (+lack of possibility to customize stuff) led to a debate on ArrayPools with those strong counter-opinions :disappointed: If you are using ASP.NET MVC, you [already have ArrayPools in your application](https://github.com/aspnet/Mvc/search?utf8=%E2%9C%93&q=arraypool&type=)!

vpenades:
@antonfirsov I'm not against ArrayPools, I agree they provide a performance benefit, I also want speed too!. 😄 

What am I against is statically allocated, non disposable array pools. From your chart, what I want is that if you're no longer performing any sort of image processing, the memory usage to drop back to zero.

antonfirsov:
@vpenades totally understand it. [You'll be able to implent this kind of logict](https://github.com/SixLabors/ImageSharp/issues/225#issuecomment-358038796) after merging #431.

rytmis:
Apologies if I sounded like I don't like ArrayPools -- that was never my intention. 🙂 I was just trying to point out potential issues in this particular mitigation strategy, that's all.

antonfirsov:
@rytmis pushed a change based on your observations. Gonna adapt this logic in #431.

antonfirsov:
I'm merging this now. If anyone is about to try our next nightlies because of this change: your feedback is really appreciated.

GeorgePlotnikov:
@antonfirsov below the result

image
the process was finished with 

```
An unhandled exception of type 'SixLabors.ImageSharp.Formats.Jpeg.GolangPort.Components.Decoder.EOFException' occurred in System.Private.CoreLib.dll
Reached end of stream before proceeding EOI marker!
```

Issue #441, opened 2018-01-19
Image.Load() on PNG sometimes results in NULL return, no exception


justintubbs:
I have PNG files that when used with `Image.Load()` return a NULL with no indication of what the error or issue might be.

**Example PNG images:**
[image1.png](http://justintubbs.com/files/image1.png)
[image2.png](http://justintubbs.com/files/image2.png)

Using ImageSharp nightly dev NuGet (also tried stable/release which has same issue)
```
<PackageReference Include="SixLabors.ImageSharp" Version="1.0.0-dev000642" />
```

**Code I'm using:**
```
using SixLabors.ImageSharp;
using SixLabors.ImageSharp.Formats;

public IActionResult UpdateReportBranding([FromForm] IFormFile file)
{
   var ms = new MemoryStream();
   file.OpenReadStream().CopyTo(ms);
   IImageDecoder decoder = new SixLabors.ImageSharp.Formats.Png.PngDecoder();
   var image = Image.Load(ms, decoder);
   if (image == null) << IS NULL HERE
   {
       ModelState.AddModelError("File", "Image was unable to load and may not be properly formatted.");
       return BadRequest();
   }
```

JimBobSquarePants:
Hi @justintubbs 

As with #440 I cannot access the images to investigate them. Could you please allow access.

Cheers

James

dlemstra:
This will happen when there is no data chunk in the file. I think this PR will fix the issue: #442.

justintubbs:
Is anyone else having issues getting to these photos?  I can view them in multiple browsers, even in Incognito/Private browsing mode

tocsoft:
I'm even getting it when just navigating to http://justintubbs.com/

justintubbs:
![image](https://user-images.githubusercontent.com/12716768/35198286-ea87c43e-feb1-11e7-8e06-a2e42e1cab27.png)

justintubbs:
![image](https://user-images.githubusercontent.com/12716768/35198288-f4f8195a-feb1-11e7-8592-7a006718768a.png)

justintubbs:
I'm just using Windows "Snipping Tool" and saving as PNG or JPG format here...you should be able to replicate yourself.

JimBobSquarePants:
I'm unable to replicate this using the current repository codebase with any of your posted images nor with any new images generated using the Window Sipping Tool.

justintubbs:
[...]

PR #442, opened 2018-01-20
Throw exception when the image does not contain a data chunk (#441).

dlemstra:
### Prerequisites

### Description
Instead of returning `null` in the png decoder we now throw an exception when no data chunk was read.


JimBobSquarePants:
We should definitely throw here but I'm not sure this will be the fix for #441 Since the OP would notice if there is no image data. Something nasty is waiting us no doubt.

PR #446, opened 2018-01-23
Premultiply to prevent transparent pixel bleed

### LINKED ISSUE #428: Incorrect color bleed when applying processors to images with alpha
### Prerequisites

### Description

When dealing with images that containt transparent pixels, some filters don't work correctly. Essentially, all processors that take the surrounding of a given pixel as input (the kernel) and return a single pixel color as output.

Usually, in images with transparent pixels, the fully transparent pixels have a black color,

The problem is that the kernels of these processors indiscriminately take all the pixel colors and handle each channel separately, without taking into account if the pixels are transparent, and should be discarded.

I am aware that handling alpha in such important filters as Resize and Blur might be difficult, and if handling transparency correctly, might end slowing down considerably.

Solution1: to make all these processors handle transparency correctly, by weighting in the transparency amount into the kernel solution. For example, if a BoxBlur kernel peeks 25 pixels, and all but 2 pixels are fully transparent, only the colors of these two non transparent pixels should be accounted for in the final solution.

Solution2: before applying any of these filters, do some sort of preprocessing that would "bleed" the non transparent colors to the transparent surrounding. Even if all these pixels remain transparent, when aplying processors like blur or resize, the end result will have much better looking borders.

Here's an example of how the transparent pixels look like when they're all made opaque. This particular solution is typical for texture mipmaps.

nd2cd

### Steps to Reproduce

Case 1: Blur Filters
1- Create a new 256x256 image.
2- Draw a 32x32 white rectangle at the center of the image.
3- Apply a GaussianBlur.

Result: the blur around the white rectangle has a gradient to Black.
preview

Expected: all the visible pixels should be white.

Case 2: Resize Filters
1- Create a new 256x256 image.
2- Draw some thin lines, text, or whatever that has a high frequency detail.
3- Downsize the image to 250x250.

Result: the resize filter has peeked a bit of the surrounding of the text, which is transparent black, that has bleeded to the final image.
preview
Expected: all the pixels should be white.

### System Configuration

- ImageSharp version: 1.0.0 beta 2
- .NET Framework version:
- Additional information:

JimBobSquarePants:
### Prerequisites

### Description
This PR fixes all convolution based processors and the resize processors to use premultiplication when calculating pixel values. This fixes #428


src/ImageSharp/Common/Extensions/Vector4Extensions.cs
```diff
@@ -12,6 +12,32 @@ namespace SixLabors.ImageSharp
...
+        /// Premultiplies the "x", "y", "z" components of a vector by its "w" component leaving the "w" component intact.
+        /// </summary>
+        /// <param name="source">The <see cref="Vector4"/> to premultiply</param>
+        /// <returns>The <see cref="Vector4"/></returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Vector4 Premultiply(this Vector4 source)
+        {
+            float w = source.W;
+            Vector4 premultiplied = source * w;
+            return new Vector4(premultiplied.X, premultiplied.Y, premultiplied.Z, w);
```
antonfirsov: We can drop an unnecessary constructor invocation for perf here:
```C#
premultiplied.W = w;
return premultiplied;
```

JimBobSquarePants: I was actually wondering about that when I wrote it. Should have profiled but I didn't know whether requesting the property twice was slower.

tests/ImageSharp.Tests/Processing/Processors/Transforms/ResizeTests.cs
```diff
@@ -82,14 +82,27 @@ public void Resize_IsNotBoundToSinglePixelType<TPixel>(TestImageProvider<TPixel>
...
+        [Theory]
+        [WithFile(TestImages.Png.Kaboom, DefaultPixelType)]
+        public void Resize_DoesNotBleedAlphaPixels<TPixel>(TestImageProvider<TPixel> provider)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            using (Image<TPixel> image = provider.GetImage())
+            {
+                image.Mutate(x => x.Resize(image.Width / 2, image.Height / 2));
+                image.DebugSave(provider);
+                image.CompareToReferenceOutput(provider);
```
antonfirsov: 👍 for being Kaboom-proof!

antonfirsov:
All looks good to me!

Wish I could find ways to significantly optimize `Premultiply()` by batching + SIMD, but failed to figure out anything worth to give a try at first glance. We need vector element shuffling intrinsics for that, currently not available in .NET vector API-s.

JimBobSquarePants:
@antonfirsov Great! Yeah, I had a look at batching myself and couldn't see anything obvious. Once those API's are in we can revisit.

PR #448, opened 2018-01-24
Fix and standardize EXIF dimension transform updates

### LINKED ISSUE #423: "Value should be an unsigned short" exception on image with EXIF rotate.

JimBobSquarePants:
### Prerequisites

### Description
This PR introduces a helper method and base class to ensure that any EXIF dimensional metadata values are correctly updates should the dimensions of the image change. Fixes #423 and supersedes #424

### DIFF
--- a/src/ImageSharp/DefaultInternalImageProcessorContext.cs
+++ b/src/ImageSharp/DefaultInternalImageProcessorContext.cs
@@ -52,8 +52,8 @@ namespace SixLabors.ImageSharp
             if (!this.mutate && this.destination == null)
             {
                 // This will only work if the first processor applied is the cloning one thus
-                // realistically for this optermissation to work the resize must the first processor
-                // applied any only up processors will take the douple data path.
+                // realistically for this optimization to work the resize must the first processor
+                // applied any only up processors will take the double data path.
                 if (processor is ICloningImageProcessor<TPixel> cloningImageProcessor)
                 {
                     this.destination = cloningImageProcessor.CloneAndApply(this.source, rectangle);
--- a/src/ImageSharp/Processing/Transforms/TransformHelpers.cs
+++ b/src/ImageSharp/Processing/Transforms/TransformHelpers.cs
@@ -32,17 +32,29 @@ namespace SixLabors.ImageSharp
             if (profile.GetValue(ExifTag.PixelXDimension) != null)
             {
                 profile.RemoveValue(ExifTag.PixelXDimension);
-                profile.SetValue(
-                    ExifTag.PixelXDimension,
-                    image.Width <= ushort.MaxValue ? (ushort)image.Width : (uint)image.Width);
+
+                if (image.Width <= ushort.MaxValue)
+                {
+                    profile.SetValue(ExifTag.PixelXDimension, (ushort)image.Width);
+                }
+                else
+                {
+                    profile.SetValue(ExifTag.PixelXDimension, (uint)image.Width);
+                }
             }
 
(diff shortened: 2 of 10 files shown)


src/ImageSharp/Processing/Transforms/TransformHelpers.cs
```diff
@@ -12,6 +14,38 @@ namespace SixLabors.ImageSharp
     /// </summary>
     internal class TransformHelpers
     {
+        /// <summary>
+        /// Updates the dimensional metadata of a transformed image
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="image">The image to update</param>
+        public static void UpdateDimensionalMetData<TPixel>(Image<TPixel> image)
```
dlemstra: Looks like you are missing an `a` here.

src/ImageSharp/DefaultInternalImageProcessorContext.cs
```diff
@@ -52,8 +52,8 @@ public IImageProcessingContext<TPixel> ApplyProcessor(IImageProcessor<TPixel> pr
             if (!this.mutate && this.destination == null)
             {
                 // This will only work if the first processor applied is the cloning one thus
-                // realistically for this optermissation to work the resize must the first processor
-                // applied any only up processors will take the douple data path.
+                // realistically for this optimization to work the resize must the first processor
```
dlemstra: 👍

src/ImageSharp/Processing/Transforms/TransformHelpers.cs
JimBobSquarePants: Where? Am I being blind?

dlemstra: `UpdateDimensionalMetData`
`UpdateDimensionalMetaData`

JimBobSquarePants: Dammit! :smile:

PR #454, opened 2018-02-03
Changed Exifs Long type validation for positive values of int type

JoshaMunnik:
### Description

Added code so that the check is also successful if the value is a positive int (including zero).


JimBobSquarePants:
@dlemstra What are your thoughts here? 

We need to delete the invalid override in `ResizeProcessor` (My bad) but should we also have this logic?

JimBobSquarePants:
Hi @JoshaMunnik 

Thanks for your submission here, we really, really appreciate it! 👍 

At this time though I think it's best that we only support the types as dictated by the EXIF specification to avoid any potential confusion. 

Cheers

James

Issue #439, opened 2018-01-18
Images get loaded incorrectly on Mono


TodesBrot:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When running Mono, images get loaded incorrectly. I have reproduced this bug on Android (although it only happens without a debugger attached) and on Windows using the Linux subsystem. Not all images seem to be affected.

This is one of the affected images:
![menu](https://user-images.githubusercontent.com/8228822/35096374-3b93aba8-fc4c-11e7-96f2-da35deafe8d0.png)

On Android the image looks corrupt:
![android](https://user-images.githubusercontent.com/8228822/35096435-7d616872-fc4c-11e7-84bd-2c59ab39b8e2.png)

On the Linux subsystem it only contains transparent pixels, although it has a size of 1.152 KB:
![output](https://user-images.githubusercontent.com/8228822/35096368-33c779ea-fc4c-11e7-960a-10c5a6679e15.png)

This is another image that gets corrupted on the Linux subsystem too:
![logo](https://user-images.githubusercontent.com/8228822/35096552-e5818810-fc4c-11e7-85ff-89b3a9e7874d.png)

On the Linux subsystem:
![output](https://user-images.githubusercontent.com/8228822/35096546-dc1c7c9e-fc4c-11e7-8553-a3a41be9f798.png)

On Android it gets corrupted differently though (note that this is a low quality screenshot taken from my game, so keep in mind that the purple background is not part of the image):
![image](https://user-images.githubusercontent.com/8228822/35096626-32750eee-fc4d-11e7-937a-b4973d27af1c.png)

On the Linux subsystem, trying to load this image results in an ImageFormatException with the message "Unknown filter type", even though ImageSharp can load it on Windows:
![level tree](https://user-images.githubusercontent.com/8228822/35096749-a4de4a86-fc4d-11e7-8e7a-5abb9b194c5c.png)

I can provide more images if needed.

### Steps to Reproduce
Load any of the source images above on Mono.

### System Configuration
- ImageSharp version: v1.0.0-beta0002
- Other ImageSharp packages and versions:
[...]

PR #455, opened 2018-02-03
Move swap to classes to avoid odd issues on Mono.

dlemstra:
### Prerequisites

### Description

It appears that the issue reported in #439 is happening due to some odd bug in Mono. The problem is this line in `PngDecoderCore.cs`:

```C#
Swap(ref this.scanline, ref this.previousScanline);
```

Swapping these two buffers corrupts the value of `this.bytesPerScanline` and makes it `0`. I am guessing that some pointer magic is happening in the swap method that corrupts the value. I tried to reproduce the issue but I could only make it happen with our code and not a simple sample program. Maybe @akoeplinger from the Mono team could take a look at this and try to figure out why this corruption is happening. It can be reproduced  (at e4293f95b3876e51e9e82b8c47555ec91fec1694) with the first image of #439 and the code below. I used the `ImageSharp.Sandbox46` project to build and test it on Mono.

```C#
var img = Image.Load("35096374-3b93aba8-fc4c-11e7-96f2-da35deafe8d0.png");
img.Save("out.png");
```

Moving the swap inside the class and removing the `ref` seems to resolve the issue.

### DIFF
--- a/src/ImageSharp/Image/ImageFrame{TPixel}.cs
+++ b/src/ImageSharp/Image/ImageFrame{TPixel}.cs
@@ -168,7 +168,9 @@ namespace SixLabors.ImageSharp
         {
             Guard.NotNull(pixelSource, nameof(pixelSource));
 
-            ComparableExtensions.Swap(ref this.pixelBuffer, ref pixelSource.pixelBuffer);
+            Buffer2D<TPixel> temp = this.pixelBuffer;
+            this.pixelBuffer = pixelSource.pixelBuffer;
+            pixelSource.pixelBuffer = temp;
         }
 
         /// <summary>
--- a/src/ImageSharp/Common/Extensions/ComparableExtensions.cs
+++ b/src/ImageSharp/Common/Extensions/ComparableExtensions.cs
@@ -169,19 +169,5 @@ namespace SixLabors.ImageSharp
         {
             return (byte)value.Clamp(0, 255);
         }
-
-        /// <summary>
-        /// Swaps the references to two objects in memory.
-        /// </summary>
-        /// <param name="first">The first reference.</param>
-        /// <param name="second">The second reference.</param>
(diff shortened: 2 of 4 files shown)


antonfirsov:
It's crazy, these fields are not even next to each other! :S Maybe strange behavior was somehow related to generics in `ComparableExtensions.Swap<T>()`.

The change looks good to me!

antonfirsov:
Posted this into mono gitter channel: 
>Looks like a generic Swap<T>(ref T a, ref T b) method is able to corrupt the owner object of a and/or b when a/b comes from the heap.


Hope someone will be able to figure it out

dlemstra:
Turns it is the `[MethodImpl(MethodImplOptions.AggressiveInlining)]` on the Swap method. Details can be found here: https://github.com/mono/mono/issues/6777. Maybe just removing that attribute could solve our issue?

JimBobSquarePants:
Great work! 

I'm happy for the swap methods to remain in the individual classes to be honest. Better localization.

src/ImageSharp/Image/ImageFrame{TPixel}.cs
```diff
@@ -168,7 +168,9 @@ internal void SwapPixelsBuffers(ImageFrame<TPixel> pixelSource)
         {
             Guard.NotNull(pixelSource, nameof(pixelSource));
 
-            ComparableExtensions.Swap(ref this.pixelBuffer, ref pixelSource.pixelBuffer);
+            Buffer2D<TPixel> temp = this.pixelBuffer;
```
jaykrell: Suggest var.

JimBobSquarePants: Why? In the statement you cannot see from the right hand side what the type is. Your suggestion deliberately obfuscates the codebase.

PR #467, opened 2018-02-17
Added actionable error message if image is saved after being disposed.

ardave:
### Prerequisites

### Description
Added an ObjectDisposedException if image is saved after being disposed.  

Previously, I did this on accident, and got null reference exceptions from deep within the png/jpg/etc encoders which made it difficult to translate those exceptions into the simple fix that was needed on my end.  Perhaps future users will figure out what they did wrong more quickly with this ObjectDisposedException.

I added a test but cannot run it, as the test suite blows up with a GDI+ related error.  I guess I would need a Windows machine to run the tests myself?  I don't have one handy.

### DIFF
--- a/src/ImageSharp/Common/Helpers/Guard.cs
+++ b/src/ImageSharp/Common/Helpers/Guard.cs
@@ -85,6 +85,24 @@ namespace SixLabors.ImageSharp
         }
 
         /// <summary>
+        /// Verifies that the provided reference is not yet disposed
+        /// and throws an exception if it is.
+        /// </summary>
+        /// <param name="t">The target value, which should be validated.</param>
+        /// <param name="disposed">A boolean value indicating whether the T has been disposed. </param>
+        /// <typeparam name="T">The type of the value.</typeparam>
+        /// <exception cref="ObjectDisposedException">
+        /// <paramref name="t"/> is disposed.
+        /// </exception>
+        public static void NotDisposed<T>(this T t, bool disposed)
+        {
+            if (disposed)
+            {
+                throw new ObjectDisposedException(t.GetType().Name);
+            }
+        }
+
+        /// <summary>
         /// Verifies that the specified value is less than a maximum value
         /// and throws an exception if it is not.
         /// </summary>
--- a/src/ImageSharp/Image/Image{TPixel}.cs
+++ b/src/ImageSharp/Image/Image{TPixel}.cs
@@ -105,6 +105,11 @@ namespace SixLabors.ImageSharp
         public IImageFrameCollection<TPixel> Frames => this.frames;
 
         /// <summary>
+        /// Gets a value indicating whether this object has yet been disposed.
+        /// </summary>
+        public bool Disposed { get; private set; }
+
+        /// <summary>
         /// Gets the root frame.
         /// </summary>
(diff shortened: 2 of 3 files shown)


src/ImageSharp/Image/Image{TPixel}.cs
```diff
@@ -104,6 +104,11 @@ internal Image(Configuration configuration, ImageMetaData metadata, IEnumerable<
         /// </summary>
         public IImageFrameCollection<TPixel> Frames => this.frames;
 
+        /// <summary>
+        /// Gets a value indicating whether this object has yet been disposed.
+        /// </summary>
+        public bool Disposed { get; private set; }
```
antonfirsov: `IsDisposed` is the usual convention in .NET

ardave:
I will give that a try

src/ImageSharp/Common/Helpers/Guard.cs
```diff
@@ -84,6 +84,24 @@ public static void NotNullOrEmpty<T>(IEnumerable<T> target, string parameterName
...
+        /// <param name="disposed">A boolean value indicating whether the T has been disposed. </param>
+        /// <typeparam name="T">The type of the value.</typeparam>
+        /// <exception cref="ObjectDisposedException">
+        /// <paramref name="t"/> is disposed.
+        /// </exception>
+        public static void NotDisposed<T>(this T t, bool disposed)
+        {
+            if (disposed)
+            {
+                throw new ObjectDisposedException(t.GetType().Name);
```
JimBobSquarePants: `typeof(T)` would be a compile time calculation and less expensive.

ardave:
Am out of town for the weekend - should have more time to tweak this during
the week.    Thanks, everyone!

ardave:
Also, I don't believe I intended to leave that .Disposed property public at
all - I only had it public because I initially added the NotDisposed guard
in the various .Save*() extension methods for the Image<> class, before
realizing I could just guard once in the Image's own .Save(...) function.

I might make my first attempt to keep it private and needing to be known
only by the type that is itself disposed, before to exposing that knowledge
any more broadly.  Unless anyone sees something wrong with that.  Or unless
the compiler sees something wrong with that.

ardave:
So I can't use the ImageExtensions.Save() (following from ) function within ImageToperationTests.cs because of a namespace collision _(The type 'ImageExtensions' exists in both 'SixLabors.ImageSharp, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' and 'SixLabors.ImageSharp.Drawing, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null')_.  

I tried duplicating the tiniest bit of that .Save() extension method's logic from within my test, particularly 

> Configuration.FindEncoder(ImageFormats.Png)

, 
just to satisfy the Image's non-Extension .Save() method.  But that call to .FindEncoder() fails with a nullreferenceexception, but I cannot get the runtime to give me a stack trace for where exactly the exception gets thrown.  And visually, I can't figure out how it's even possible for that exception to be thrown.  _And_, something about that project type is such that System.Console is unavailable for printing diagnostic messages.

So perhaps i might graciously just bow out of this one?  This looks like a cool project to work on, and I'd love to contribute more if I have the time some day, but i have too much going on right now and this is turning out to be just a liiiitle bit too involved for just changing a NullReferenceException to an ObjectDisposedException.

Issue #465, opened 2018-02-14
Create image strip


valse:
Hi,
  I'd like to merge some images (same size side by side) and I tried the DrawImage method:

```c#
image1.Mutate(x => x
   .DrawImage(
      image2,
      1,
      new Size(image1.Width * 2, image1.Height),
      new Point(image1.Width, 0)
   )
);
```

but the output image width is not like expected (image1.Width * 2) but it keeps the image1 width.

To solve this I make first a Resize of first image adding padding on the right side:

```c#
.Resize(new ResizeOptions
   {
      Mode = ResizeMode.Pad,
      Position = AnchorPosition.TopLeft,
      Size = new Size(image1.Width * 2, image1.Height)
   }
)
```

Is it the right solution or the DrawImage must work resizing the final output too?

Thanks
Marco

valse:
Hi,
   I saw a strange memory behavior about DrawImage: in my project I take about 20 Full HD video thumbnails (1920x1080) that I merge together like this result image (that I created empty :-P referring my previous post):

![test](https://user-images.githubusercontent.com/1492995/36471532-0d816724-16ef-11e8-94e8-db6e411632d5.jpg)

Each thumb will be resized to 110x62 but if I let this job to the DrawImage method, the dotnet.exe process grow a lot for each request.

If I resize the thumb before using it in the DrawImage method, the memory keeps low.

This is my snippet code:
```c#
using (var thumb = Image.Load(buffer))
{
	// without this memory leaks
	thumb.Mutate(x => 
		x.Resize(thumbWidth, thumbHeight)
	);

	strip.Mutate(x =>
		x.DrawImage(
			thumb,
			1,
			new Size(thumbWidth, thumbHeight),
			new Point(
				nIndex % thumbsPerColumn * thumbWidth,
				(int) Math.Floor(nIndex / (decimal) thumbsPerColumn) * thumbHeight)
		)
	);                            

	nIndex++;
}
```
Thanks
Marco

antonfirsov:
@valse thanks that's very useful feedback! 👍 Normally, the memory usage should not grow with your use case, so this might be a bug. Gonna try to investigate it during my work on #225.

It's a huge job however, so I really need all the help the community can provide! I you could make a small, self containing console app reproducing this issue, there is higher chance I'll be able to fix the issue and maybe provide a workaround for you that works with the current official beta.

valse:
[...]

PR #471, opened 2018-02-22
Remove Clone from DrawImage to reduce memory consumption

JimBobSquarePants:
### Prerequisites

### Description
This PR removes the internal clone from the `DrawImageProcessor` to reduce memory. See #465 
Preserving the original image is now the responsibility of the consuming developer.


JimBobSquarePants:
@tocsoft I was actually thinking the same when I was out for a walk earlier. I vote drop the size parameter entirely as draw whatever they give us.

JimBobSquarePants:
@tocsoft Ok, I've stripped out the `size` parameter completely, It already feels better to consume. 

I added a couple of quick cleanups to our transform code so we could better test passing a transformed image to the method. Nothing major, just making things more sensible.

PR #472, opened 2018-02-22
Fixing Encoder Tests

antonfirsov:
### Prerequisites

### Description
There were several bugs in the BMP+PNG+Jpeg encoder tests introduced by #469. In fact, those tests weren't verifying anything. This is fixed now.

BMP and Jpeg encoders are now tested without reference images using the `System.Drawing.Common` decoder implementation of the platform running the tests. In case of Jpeg, the tolerance threshold values are set to high values now, the correct values are TBD. (+ I think the encoder's LUT-based RGB->YCbCr conversion is not accurate enough at the moment, making the encoder inaccurate. Like in the decoder, it has to be implemented with `float` + SIMD instead!)

### DIFF
--- a/tests/ImageSharp.Tests/Formats/Jpg/JpegEncoderTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Jpg/JpegEncoderTests.cs
@@ -15,6 +15,7 @@ namespace SixLabors.ImageSharp.Tests.Formats.Jpg
     using SixLabors.ImageSharp.PixelFormats;
     using SixLabors.ImageSharp.Tests.TestUtilities.ImageComparison;
     using SixLabors.ImageSharp.Tests.TestUtilities.ReferenceCodecs;
+    using SixLabors.Primitives;
 
     using Xunit;
     using Xunit.Abstractions;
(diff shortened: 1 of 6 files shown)


tests/ImageSharp.Tests/TestUtilities/TestImageExtensions.cs
```diff
@@ -22,6 +22,38 @@ namespace SixLabors.ImageSharp.Tests
...
+                            foreach (ImageFrame<TPixel> frame in img.Frames)
+                            {
+                                Span<TPixel> pixelSpan = frame.GetPixelSpan();
+
+                                PixelOperations<TPixel>.Instance.ToVector4(pixelSpan, tempSpan, pixelSpan.Length);
+
+                                for (int i = 0; i < tempSpan.Length; i++)
+                                {
+                                    ref Vector4 v = ref tempSpan[i];
+                                    v.W = 1.0f;
```
JimBobSquarePants: This only works if your vector range is `0-1`, for  `Short4` ranges from `-37267` to `37267`. That's why I've relied on the image format to handle transparency so far.

tests/ImageSharp.Tests/Formats/Jpg/JpegEncoderTests.cs
```diff
@@ -55,20 +56,27 @@ public void EncodeBaseline_IsNotBoundToSinglePixelType<TPixel>(TestImageProvider
             TestJpegEncoderCore(provider, subsample, quality);
         }
 
-        private static ImageComparer GetComparer(int quality)
+        /// <summary>
+        /// Anton's SUPER-SCIENTIFIC tolerance threshold calculation
+        /// </summary>
+        private static ImageComparer GetComparer(int quality, JpegSubsample subsample)
```
JimBobSquarePants: The scaled approach is the same I found in the libjpeg source so I wonder why it would produce such different output? Happy for you to replace it in the future with your SIMD wizardry though!

tests/ImageSharp.Tests/TestUtilities/TestImageExtensions.cs
JimBobSquarePants: I do wonder whether we should be using the ScaledVector approach for all our methods. it would slow things down slightly for the more exotic types though.

antonfirsov: Yeah that ranging stuff is quite complicated, I still can't follow it. Maybe we can have `ToVector4` and `ToScaledVector4` side by side, letting the ImageProcessor to pick one of them. The refactor path could be also easier.
Neverheless this is just a basic implementation for testing purposes used by `***IsNotBoundToSinglePixelType` theories, and we are testing just a few pixel types in those.

tests/ImageSharp.Tests/Formats/Jpg/JpegEncoderTests.cs
antonfirsov: Its not a proper correctness testing for now, I'm comparing the orinal image against the one that's saved by our encoder then decoded by S.D. The goal was to cath regressions.

However  there are a few results with 444+Q=100 encoding which should be better IMO. Libjpeg can is able to choose from several code paths AFAIK depending on configuration + architecture, LUT is just one of them and SIMD is preferred whenever available. (At least the decoder works this way.)

tests/ImageSharp.Tests/TestUtilities/TestImageExtensions.cs
antonfirsov: Let's go for it then! 
(Post 1.0 I guess)

PR #474, opened 2018-02-24
Make resize and transform processors immutable #473

### LINKED ISSUE #473: ResizeOptions instance can't be reused
### Prerequisites

### Description
When reusing an instance of the ResizeOptions on two images, one landscape and one portrait, only the first image is resized correctly. This was unexpected behaviour for me, although it is working as programmed, see [https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Processing/Transforms/Resize.cs](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Processing/Transforms/Resize.cs)

### Steps to Reproduce
1. Initialize a ResizeOptions instance.
2. Use that instance to resize two different images. This behaviour is easiest to observe when the images have different orientations.

### Thoughts
I don't know if this behavior is intended. If it is, this issue is moot, though it would be nice to have this documented somewhere.

JimBobSquarePants:
### Prerequisites

### Description
Fixes #473 which was a sharp prod to remind me to fix the immutability issues we have with the resize processor and calling methods and also the transform processors.

Both issues were unsolvable without adding an extra `Bounds()` method to the `IImageProcessingContext` interface which tracks the source image bounds. Additionally I had to merge the `ResizeProcessor` and `ResamplingWeightedProcessor` to allow the calculations required when passing a `ResizeOptions` instance. Nothing else was using that processor anyway and if we need to in the future we can always move the weight calculation methods out into another helper class.


src/ImageSharp/IImageProcessingContext{TPixel}.cs
```diff
@@ -14,6 +14,12 @@ namespace SixLabors.ImageSharp
     public interface IImageProcessingContext<TPixel>
         where TPixel : struct, IPixel<TPixel>
     {
+        /// <summary>
+        /// Gets the source image bounds
+        /// </summary>
+        /// <returns>The <see cref="Rectangle"/></returns>
+        Rectangle Bounds();
```
antonfirsov: As far as I understand, this is always the boundary of the **initial** image.
What happens if someone applies multiple operations, with the sizes `sizeA > sizeB > sizeC` ?
```C#
image.Mutate(c => c.Resize(sizeA).Resize(sizeB).Resize(sizeB) );
```

JimBobSquarePants: Good catch! The default implementation would have only worked when mutating not cloning. I've updated and added tests.

JimBobSquarePants:
@antonfirsov It's actually all really trivial stuff, basically copy/pasting code from one location to another and tweaking the input parameters 😄 . I'm gonna hopefully get a chance to review your PR at lunch, if not then definitely tonight 👍

tests/ImageSharp.Tests/Image/ImageProcessingContextTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            var x400 = new Size(400, 400);
+            var x300 = new Size(300, 300);
+            var x200 = new Size(200, 200);
+            var x100 = new Size(100, 100);
+            using (var image = new Image<Rgba32>(500, 500))
+            {
+                image.Clone(x =>
+                    x.AssertBounds(x500)
+                        .Resize(x400).AssertBounds(x400)
+                        .Resize(x300).AssertBounds(x300)
```
antonfirsov: 👍

src/ImageSharp/IImageProcessingContext{TPixel}.cs
antonfirsov: I'm still a bit worried if this method is good for us:

We have added a visible state to an object that looked kinda stateless towards it's consumers. I remember we did not want to expose source/destination images on the interface for some reason, but actually, if we are exposing `Bounds`, it's almost the same. 

I understand the reason however, that you need this because the initialization logic lives now mostly inside the processor constructors, but maybe there are other solutions to achieve the same goals.

@tocsoft your thoughts?

If we are unsure, how about hiding this behind `IInternalImageProcessingContext<T>` to avoid exposing it to users?

JimBobSquarePants:
Before I merge this, quick question. Since the bounds are always going to be a `Rectangle` of `(0, 0, width, height)` I wonder whether we should just preserve the `Size` instead?

antonfirsov:
Then we should definitely use `Size`.

Thought I added a few more comments here, but can't find it, so I'm going to repeat it:
1. Naming: I think we should name this method `GetCurrentSize()`.
2. I have an idea to internalize this stuff painlessly, maybe I'll send a PR later hiding Size/Bounds and the `MemoryManager` from the `IImageProcessingContext<T>` API surface. We should not block ourselves thinking too much on this, but if we have a bit of extra time, I think it's better to do this work to keep the interface as narrow as possible.

JimBobSquarePants:
Ok, I'm happy with this for now then using `GetCurrentSize()`. Any future changes can be handled in a separate PR.

PR #478, opened 2018-02-28
Refactored Image Formats management into its own class

vpenades:
### Prerequisites

### Description
After adding memory management to Configuration class, it's convenient to create custom configurations, but since the ImageFormats are handled directly by the default configuration, it's tricky to create custom configurations that share the ImageFormats of the default configuration.

In order to solve the problem, I've moved the ImageFormats code to it's own management class, so not it's easy to create a new configuration, and assign the image formats of the default configuration. Also, formats added at runtime to the default configuration will automatically propagate to all other configurations sharing the `ImageFormatsManager`.

Additional suggestions:
- Image Formats API in Configuration could be removed, or at least tagged as Obsolete.
- Right now ConfigurationModules is used only to setup image formats, the whole code could be moved to ImageFormatsManager.


JimBobSquarePants:
Cheers @vpenades !!

@tocsoft Could you please review this PR as you wrote the original configuration code. Thanks!

src/ImageSharp/Configuration.cs
```diff
@@ -1,267 +1,221 @@
...
+
+        /// <summary>
+        /// Gets the currently registered <see cref="IImageFormat"/>s.
+        /// </summary>
+        public IEnumerable<IImageFormat> ImageFormats => this.FormatsManager.ImageFormats;
+
+        /// <summary>
+        /// Gets or sets the <see cref="ImageFormatsManager"/> that is currently in use.
+        /// </summary>
+        public ImageFormatsManager FormatsManager { get; set; } = new ImageFormatsManager();
```
JimBobSquarePants: Can we rename `ImageFormatsManager` to `ImageFormatManager` and the property names to match please.

src/ImageSharp/Configuration.cs
```diff
@@ -1,267 +1,221 @@
...
+        /// Creates the default instance with the following <see cref="IConfigurationModule"/>s preregistered:
+        /// <para><see cref="PngConfigurationModule"/></para>
+        /// <para><see cref="JpegConfigurationModule"/></para>
+        /// <para><see cref="GifConfigurationModule"/></para>
+        /// <para><see cref="BmpConfigurationModule"/></para>
+        /// </summary>
+        /// <returns>The default configuration of <see cref="Configuration"/></returns>
+        internal static Configuration CreateDefaultInstance()
+        {
+            return new Configuration(
```
JimBobSquarePants: Shouldn't the `ImageFormatManager` constructor accept these instead?

src/ImageSharp/Configuration.cs
```diff
@@ -1,267 +1,221 @@
...
-        /// <summary>
-        /// Sets the max header size.
-        /// </summary>
-        private void SetMaxHeaderSize()
-        {
-            this.MaxHeaderSize = this.imageFormatDetectors.Max(x => x.HeaderSize);
-        }
-    }
-}
+﻿// Copyright (c) Six Labors and contributors.
```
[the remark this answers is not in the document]
vpenades: Indeed... I was surprised too see that massive change.

src/ImageSharp/Configuration.cs
vpenades: That's an issue I would like to discuss. The ´IConfigurationModule´ requires to pass a ´Configuration´ instance. If we want to do the formats configuration in 'ImageFormatManager'  the parameter needs to be changed from Configuration to 'ImageFormatManager', which would lead to renaming the interface to 'IImageFormatModule' or something like that.

But I can see the ´IConfigurationModule´ to be useful to intialize things other than image formats, so I believe it's still subject to debate.

vpenades:
How the hell Travis is detecting [5 image formats](https://travis-ci.org/SixLabors/ImageSharp/builds/347233665#L2403)??

src/ImageSharp/Configuration.cs
tocsoft: nope they should be passed into the `Configuration` like they are and not into `ImageFormatManager` manager its just happen by chance that the things they are configuring are image formats rather than anything else.

tocsoft:
@vpenades the test project injects a fake test image format that allows us to test the global default configuration object

https://github.com/SixLabors/ImageSharp/blob/47a8533b7d7b43b490d4958d2702490a580d9d34/tests/ImageSharp.Tests/TestFormat.cs#L19

src/ImageSharp/Configuration.cs
JimBobSquarePants: Great! keep it as-is then

vpenades:
Sooo... anything else?

Should I remove the formats API from configuration? or I leave things are they're and merge?

tocsoft:
sorry, i'm actually in work still at the moment i'll try to do a full review this evening.

vpenades:
@JimBobSquarePants I had to merge some changes from other PRs that have modified the Configuration.cs and caused a conflict. As far as I've seen, the addition has been to add the ReadOrigin property, which I've also added to the copy constructor.

I would humbly request a review for this just in case I didn't resolve the conflict correctly and I left something out.

src/ImageSharp/Configuration.cs
```diff
@@ -1,272 +1,244 @@
...
+        {
+            Guard.NotNull(configuration, nameof(configuration));
+            configuration.Configure(this);
+        }
+
+        /// <summary>
+        /// Registers a new format provider.
+        /// </summary>
+        /// <param name="format">The format to register as a known format.</param>
+        public void AddImageFormat(IImageFormat format)
```
antonfirsov: @tocsoft @JimBobSquarePants wouldn't it cleaner to have our API without these shortcuts, exposing `AddImageFormat()`, `SetEncoder()` etc. only on `ImageFormatManager`?
Or should we keep them for compatibility?

src/ImageSharp/Configuration.cs
```diff
@@ -1,272 +1,244 @@
...
+        /// </summary>
+        public Configuration()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Configuration"/> class.
+        /// </summary>
+        /// <param name="configuration">A configuration instance to be copied</param>
+        public Configuration(Configuration configuration)
```
antonfirsov: Normally I would expect this kind of constructor to do a deep copy. Wouldn't it be better to have a `Configuration.ShallowCopy()` method instead to be more explicit?

src/ImageSharp/Configuration.cs
JimBobSquarePants: I would be inclined to remove them.

src/ImageSharp/Configuration.cs
vpenades: I think the most common use is going to be create a copy of the Default configuration, that keeps all the internal objects shared with Default, and then change exactly what you want to change.

By doing a deep copy you loose the feature of sharing underlaying components between configurations. Sharing components can be really convenient for MemoryManager and ImageFormatManager; for example, we can have a derived configuration with a custom memory manager, but image formats shared with default, so if a new format is added later at runtime, the derived configurations automatically pick that new format.

src/ImageSharp/Configuration.cs
vpenades: I didn't remove them precisely because backwards compatibility. As an intermediate solution, they can be tagged with the Obsolete attribute for a while, so developers will have a cautionary warning to point to the right methods, and they can finally be removed after a while.

JimBobSquarePants: I wouldn't bother, we've broken tons of things recently and we'll be moving things about more when be fix up the namespaces.

src/ImageSharp/Configuration.cs
JimBobSquarePants: >ImageFormatManager; for example, we can have a derived configuration with a custom memory manager, but image formats shared with default, so if a new format is added later at runtime, the derived configurations automatically pick that new format.

But what happens if someone want a more restrictive list in a separate location? It would then be a lot of work creating and populating the new manager.

vpenades: @JimBobSquarePants then what it needs a deep copy is MemoryManager and ImageFormatManager.

In the case of ImageFormatManager, should it also deep copy the formats themselves?

JimBobSquarePants: I would expect so yeah. .NET needs a `DeepClone()` method to make this all easier!

antonfirsov: @vpenades I suggested a `ShallowCopy()` method, not a deep copy. I just don't like a constructor doing a shallow copy implicitly. Being more verbose here is better IMO:
```C#
Configuration myConfig = Configuration.Default.ShallowCopy();
myConfig.MemoryManager = ...;
```

JimBobSquarePants:
LGTM, I've just updated the PR from master and will merge tomorrow if no complaints. 👍

PR #479, opened 2018-02-28
Baseline Jpegs can contain more than one scan

JimBobSquarePants:
### Prerequisites

### Description
@antonfirsov Well the first part was an easy fix. We were stopping processing after reading the initial SOS (Start Of Scan) marker because is was a baseline jpeg. Baseline Jpegs, while generally have only one scan, can actually have more than one. 

That exposed an issue we were having with missing EOF (End of File) markers on baseline files. I was able to copy the behavior from PdfJS port though to counter that.

I'd really like to try and work with you post V1 when we have time to try to bring the best of both decoders together into a better one. I think we have more overhead than we need using the InputProcessor for anything other than spectral decoding within the SOS segment. The byte stuffing and reversal should only be needed there.


src/ImageSharp/Formats/Jpeg/GolangPort/Components/Decoder/InputProcessor.cs
```diff
@@ -107,6 +107,13 @@ public byte ReadByte()
             return this.Bytes.ReadByte(this.InputStream);
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public OrigDecoderErrorCode ReadByteUnsafe(out byte result)
```
JimBobSquarePants: Ah that's better! 👍

antonfirsov: Actually defaulting to this logic would be the best, but I felt it's more safe to continuously refactor stuff towards it, also keeping the exception throwing where we haven't yet implemented the "smart" error management.

JimBobSquarePants:
Glad it wasn't something buried deep in the scan decoder or something else crazy. You're added method is much neater. 

I'll merge once the build is passed, hopefully tweaking the tests will bring up the coverage.

antonfirsov:
@JimBobSquarePants  re mixing the two decoders, this is how I imagine it:
1. Adapt `OrigJpegScanDecoder` into the PdfJs stream parsing logic, this can be tested & debugged independently with the help of `SpectralJpegTests`
2. Implement `IRawJpegData` by the pdfjs port, so we can feed the spectral results to `JpegImagePostProcessor`

PR #480, opened 2018-02-28
Introducing ReadOrigin. Fix #477

### LINKED ISSUE #477: Unexpected Behavior using Image.Load(Stream) with AWS S3 Source
### Description
I am using the .NET AWS SDK to load from S3 using their GetObject, but I had to do "extra" work to get it to work with ImageSharp, while System.Drawing.Image.FromStream worked without that extra work.

### Steps to Reproduce

1. I added a package dep via nuget to the current latest (1.0.0-beta0002).
2. Using v3.3.17 of AWS S3 SDK, I call their GetObject method and attempt to load their ResponseStream into the ImageSharp Image.Load(Stream) method. 

Actual (First Stop):
It says it can't read. I noticed AWS had their own internal "[MD5Stream](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/Internal/Util/HashStream.cs#L393)" so I figured maybe I should copy that to a regular MemoryStream.

New Problem:
Can't load, tried these N formats. (All the default ones..) as in #164. That issue suggests seeking back to the beginning, which I do. And now it works.

However, that issue is closed, saying it was fixed in an alpha, but the issue persists in beta.

More to the point, and the reason I opened a separate issue, is that I don't think I should need to copy to a MemoryStream. It seems like the library should handle that, especially since "it just works" with System.Drawing. 

Basically, can the library handle all this and make it "just work"?

### System Configuration
- ImageSharp version: 1.0.0-beta0002
- Environment (Operating system, version and so on): Win10 x64 
- .NET Framework version: 4.5

JimBobSquarePants:
### Prerequisites

### Description
This PR introduces new overloads for any of our methods that accept a stream. Fixing #477

These overloads contain a new parameter `ReadOrigin` which allows the calling code to instruct us to read from either the beginning or the current position within the stream, defaulting to the beginning.

This ensures that developers are not surprised, nor have to reset the stream position in a common use case.


tocsoft:
It feels like it would make more sense to put the `ReadOrigin` onto the `Configuration` object instead of just passing it into all the places we use streams, thus it can be globally opted into.

antonfirsov:
@JimBobSquarePants I think there is a misunderstanding. Isn't the point of #477 that we are failing to load from a non-seekable stream?

In that case this is only a "fix" for the workaround that asks the user to manually load the image into a memory stream. We are just rewinding the stream to the beginning, so the user shouldn't.

If my understanding is correct, I think we should not do this, this is quite unconventional. We should investigate the issue with non-seekable streams instead.

tocsoft:
if the stream is non-seekable we are already copying it internally into a memory stream then reading it from the beginning... re the original issue is we must be detecting it as seekable thus not doing the internal copy and thus failing to reset it between attempts to detect image format.

JimBobSquarePants:
@antonfirsov There's actually two issues embedded in one there, I missed the second issue that simply passing the stream doesn't work.

@tocsoft I had a look through the AWS code. `MD5Stream` is both non-readable and non-seekable. We're throwing because it's non-readable and `CopyTo()` relies on the stream being readable. 

However, the OP managed to wrap the stream in a `MemoryStream` (I'm assuming through the constructor but I don't want to have to download the SDK to check) so I've asked him how. We should use that approach rather than our existing one if this is the case.

Regarding using `Configuration`, I designed it like `Stream.Seek(SeekOrigin)` which the localized context works well with. I guess I could move it there though, it might actually make the code a little cleaner.

src/ImageSharp/Configuration.cs
```diff
@@ -83,6 +83,11 @@ public Configuration(params IConfigurationModule[] configurationModules)
         /// </summary>
         public IEnumerable<IImageFormat> ImageFormats => this.imageFormats;
 
+        /// <summary>
+        /// Gets or sets the position in a stream to use for reading.
+        /// </summary>
+        public ReadOrigin ReadOrigin { get; set; } = ReadOrigin.Begin;
```
dlemstra: I don't think we should make this the default. We should not seek within the stream unless the user instructs us to do so.

src/ImageSharp/Image/Image.FromStream.cs
```diff
@@ -246,9 +241,19 @@ private static T WithSeekableStream<T>(Stream stream, Func<Stream, T> action)
...
             {
+                if (config.ReadOrigin == ReadOrigin.Begin)
+                {
+                    stream.Position = 0;
+                }
+
                 return action(stream);
             }
 
+            if (config.ReadOrigin == ReadOrigin.Current)
```
dlemstra: Shouldn't this be `config.ReadOrigin != ReadOrigin.Current` or `config.ReadOrigin == ReadOrigin.Begin`?

src/ImageSharp/Configuration.cs
JimBobSquarePants: Based on my normal expectations I would be inclined to think the same. However, I think many other devs would expect it to JustWork™

src/ImageSharp/Image/Image.FromStream.cs
JimBobSquarePants: If a stream is non-seekable then it's always gonna end up at 0 since we copy it to a `MemoryStream`. I thought best to throw rather than silently ignore.

src/ImageSharp/Configuration.cs
dlemstra: I personally think this is a bad decision but you are the "Grand High Eternal Dictator" so you have the final say. I do think we need an option for this and it is a great idea to add this but this should not be the default.

src/ImageSharp/Image/Image.FromStream.cs
dlemstra: Maybe I don't understand what this method should do but shouldn't it copy the data from the stream to a memory stream to make it seekable? The ReadOrigin is an option for the input stream or also for the stream that we create?

src/ImageSharp/Configuration.cs
JimBobSquarePants: Haha! We'll see what the other chaps think. Off to bed now so will come back to it in the morning.

src/ImageSharp/Image/Image.FromStream.cs
tocsoft: I feel we should be using this setting as more of a hint than a hard rule. Basically we use it as permission to reset streams when they are seekable and when they are not then we use our current logic of copy and read from there.

src/ImageSharp/Configuration.cs
antonfirsov: I'm not convinced at all that we need to do this, specially not by default!
See my comment on #477.
Or am I missing something? 😕

JimBobSquarePants:
I'm putting this on hold for the moment. Something funky is going on.
https://github.com/SixLabors/ImageSharp/issues/477#issuecomment-369424708

JimBobSquarePants:
Ok... So it turns out #477 was user error. 

I think there's still merit in adding some granular control to our stream seeking behavior so with that in mind I've changed the default to `ReadOrigin.Current` and removed the extra exception. 

Default behaviour is still the same as prior to the PR, we just have extra options now.

PR #482, opened 2018-02-28
Update System.Memory library

iamcarbon:
### Prerequisites

### Description

Updates ImageSharp to the latest System.Memory preview, annotated readonly structures, and fixed breaking changes. 

All tests pass.


iamcarbon:
@antonfirsov do the current head tests pass on .NETCOREAPP 2.1? are we breaking any new tests with this update?

I'll install the preview bits and see if we can fix the tests without cross-targeting. If we can't, it may be a good time to begin cross targeting the main project to .NETCOREAPP 2.1.

I also think that there's been a few more breaking changes to Memory since the first preview release. We may also want to wait for preview2 when the API  is more stable.

antonfirsov:
@iamcarbon you're most likely right, those test failures will be most likely unrelated to `System.Memory`. I was hoping that with this PR we can introduce actual support for 2.1 but It's getting too hard.

Maybe we should leave it for now, and just update the package. We can figure out the next steps later.

src/ImageSharp/PixelFormats/Generated/PixelOperations{TPixel}.Generated.cs
```diff
@@ -20,8 +21,8 @@ internal virtual void PackFromRgba32(Span<Rgba32> source, Span<TPixel> destPixel
         {
 			GuardSpans(source, nameof(source), destPixels, nameof(destPixels), count);
             
-            ref Rgba32 sourceRef = ref source.DangerousGetPinnableReference();
-            ref TPixel destRef = ref destPixels.DangerousGetPinnableReference();
+            ref Rgba32 sourceRef = ref MemoryMarshal.GetReference(source);
```
antonfirsov: We also need to apply these changes to the `.tt` files. This code is auto-generated.

JimBobSquarePants:
Which tests are failing?

src/ImageSharp/PixelFormats/Generated/PixelOperations{TPixel}.Generated.cs
iamcarbon: Got stuck trying deterministically reference a nuget package in the T4 template. 

something like...  <#@ assembly name="$(NuGetPackageRoot)/system.memory/4.5.0-preview1-26216-02/lib/netstandard2.0/System.Memory.dll" #>

Will take a fresh stab tomorrow.

iamcarbon:
@JimBobSquarePants There are breaking changes on Span in that I would expect to fail but this PR should fix them. Also curious...

antonfirsov:
[My test output](https://gist.github.com/antonfirsov/1098788676073e739dfd24b7862c163c) from the point where stuff starts to fail. Many failing, mostly Drawing stuff.

src/ImageSharp/PixelFormats/Generated/PixelOperations{TPixel}.Generated.cs
antonfirsov: @iamcarbon you don't need that! T4 is just a dumb tool for generating the code, it doesn't need to know what the generated code does or references. Actually you can generate any text with it. The syntax is almost the same as classic ASP.

antonfirsov:
Actually .. they are all drawing related, maybe with the same root cause. Maybe there is a simple fix :)

JimBobSquarePants:
That's good news! I'll let you have a dig around for now, I'm on the day job.

antonfirsov:
Nope .. I really should be in bed now! 😄

src/ImageSharp/PixelFormats/Generated/PixelOperations{TPixel}.Generated.cs
iamcarbon: @antonfirsov ahhhhh! was overthinking this for no reason. updated.

JimBobSquarePants:
Do all the tests pass on 2.1 now? I haven't installed the preview bits yet.

antonfirsov:
No, I just approved the T4 changes I requested. If we want, we can fix the drawing stuff later.
The tooling is not very friendly, or at least not with the VS2017+R# versions I have installed, so I haven't started debugging it.

iamcarbon:
I think 2.1 is a moving target until the Memory stuff stabilizes. Not sure it's worth supporting yet.

JimBobSquarePants:
No worries, let's get this in for now and fix the 2.1 bugs later.

JimBobSquarePants:
Imma gonna merge this now, I hate these things stacking up.

Issue #401, opened 2017-12-15
Refactor solution namespaces and classes


JimBobSquarePants:
### Problem

Currently most of the existing types are located within a flat namespace `SixLabors.ImageSharp`. Many of our extension methods share a single class `ImageExtensions`.

This is detrimental for several reasons:

- Discoverability. Having all the types in a single namespace pollutes intellisense.
- Bugs. Using classes with the same name and namespace in both `ImageSharp` and `ImageSharp.Drawing` causes ambiguous reference issues #391
- Documentation. With clustered method sharing single namespaces and classes we are unable to produce meaningful API documentation using automated means. 

### Solution

Tackling this problem requires some significant refactoring to the solution and tests.

- Refactor extensions and processors into individual namespaces grouped by their operation function. For example:
  - Transforms
  - Effects
  - Convolution
  - Drawing
- Separate the static extension method classes into individual classes. For example: 
  - Resize
  - Crop 
  - Rotate
- Change root namespace for `ImageSharp.Drawing` from `SixLabors.ImageSharp` to `SixLabors.ImageSharp.Drawing` and perform similar refactorings there where necessary.

### Potential Issues

Any open PR's will likely have merging issues after this change. We should strive to close any that we possible can before work commences.

rspeele:
Just want to let you know this might be more important than you think.

I'm in .NET framework land, so it might be different on Core. But as far as I can tell, the ambiguous references make it **impossible** to call `ImageExtensions` methods from any code that references `SixLabors.ImageSharp.Drawing`.

I know in #391 you suggested this was due to calling the methods as regular old static methods instead of using the extension method syntax, but in my experience it doesn't work either way -- you just get a less helpful, more misleading error using the extension method syntax (which of course normally I would use).

#### Example

I make a console app targeting framework 4.5.2 (I doubt the framework version matters though).
I install nightly:

```
Install-Package SixLabors.ImageSharp -Version 1.0.0-dev000733 -Source https://www.myget.org/F/sixlabors/api/v3/index.json 
```

I write this code:

```csharp
using SixLabors.ImageSharp;

namespace Foo
{
    class Program
    {
        static void Main(string[] args)
        {
            var image = Image.Load(new byte[0]);
            image.Resize(0, 0);
        }
    }
}
```

It doesn't compile.

```
[...]

PR #486, opened 2018-03-11
Refactor namespaces to avoid collisions.

JimBobSquarePants:
### Prerequisites

### Description
This PR refactors the entire processing pipeline to a new namespace structure fixing #401 It's huge but most of the changes are simply namespace related.

Processing extension methods and individual processors now follow a pattern that groups the API by functionality.

For example processing methods pertaining to transforms adhere to the following structure.

- `SixLabors.ImageSharp.Processing.Transforms`
- `SixLabors.ImageSharp.Processing.Transforms.Processors`

Base classes and interfaces follow that same pattern by design.

Processors are internal (for now) but I would like to make them public. I can do this as part of this PR if everyone agrees.

Additionally the following optimizations and changes have been made.
- Replaced `FastArray2D<T>` with faster `DenseMatrix<T>` which now matches 1D array for speed.
- Refactored all the processors to be immutable removing initialization code from extension methods
- Renamed `QuantizedImage` to `QuantizedFrame`
- Add new `QuantizeProcessor<TPixel>` processor
- Refactored quantizers to remove per pixel virtual call
- Moved `Rgba32` where it belongs with the other pixel formats
- Moved `Bounds()` and `Size` to `IImage`.
- Fixed a bug in `DrawImage` where the default Graphics options were being overwritten.

Drawing methods follow the same pattern. The main namespaces used are `Overlays` and `Text`.


src/ImageSharp/IImage.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// Encapsulates the properties and methods that describe an image.
+    /// </summary>
+    public interface IImage : IImageInfo, IDisposable
+    {
+        /// <summary>
+        /// Gets the size of the image.
+        /// </summary>
+        /// <returns>The <see cref="Size"/></returns>
+        Size Size();
```
antonfirsov: We can push this up to `IImageInfo`.
Actually there are Width + Height properties here, so maybe it should be an extension method.

src/ImageSharp/IImage.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>
+        /// Gets the bounds of the image.
```
antonfirsov: `Bounds()`  could be an extension method on `IImageInfo`.

JimBobSquarePants: I'm happy to move them there. 👍

antonfirsov:
Another counter-argument against publishing the processors (and I promise, this would be the last one 😄):

Implementing #142 won't be possible without breaking public API. Are we fine with this?
It would also affect `IResampler` API.

JimBobSquarePants:
@antonfirsov You've got me there! 😄  

We can't afford another round of massive churn after this before V1 so yeah, if we want to implement #142 we should keep the processors internal for now.

I imagine value type kernels would be the last big task before V1. Do you think you would be able to allocate (haha) time in order to take that on?

antonfirsov:
Unfortunately, I can't make for 1.0 😞, it's a quite a lot of work & I'm starting my new job in a week.

We need to decide if we should make it possible to implement it in a `1.*`  version, without major braking changes. 
The price is: keeping things internal and/or investing extra effort into making our API more future-proof.

JimBobSquarePants:
>We need to decide if we should make it possible to implement it in a 1.* version, without major breaking changes

It all depends on how much it will affect the `IResampler` API. We've got calling overloads in `Resize` that use the type as a parameter. `Resize` will probably be our most utilized method so we need to be very careful.

I'm happy to wait til V2 for #142 also though. 

Ship a working version, hopefully make enough money to allow us all to live of this should we choose to and then add everything we've learned from V1* to V2

antonfirsov:
Naive apporach:
`IResampler` should be something like `IResampler<TMatrix>` & `ResizeProcessor` should be also generic to utilize JIT-time, boxing & virtual call free optimizations.

Advanced approach:
Finding a way to hide generics behind nested implementation classes using tricks like double dispatch. The public API should be still touched, but in a less intrusive way.

JimBobSquarePants:
@dlemstra Yeah much better I think. What's your thoughts on the rest of the PR?

antonfirsov:
Had another look, and still looks good to me.

I think the only non-trivial decision here is to have the drawing code in `SixLabors.ImageSharp.Processing.Overlays` instead of `SixLabors.ImageSharp.Processing.Drawing`. Can't decide if it's good or not. Thoughts?

JimBobSquarePants:
I've moved `Overlays` back to `Drawing` again. I think devs will expect `Drawing`

PR #487, opened 2018-03-14
Use non-generic IQuantizer for image formats

JimBobSquarePants:
### Prerequisites

### Description
This PR replaces the generic `IQuantizer<TPixel>` used by the image formats with a non-generic implementation  `IQuantizer`.

It also introduces internal `IFrameQuantizer<TPixel>` implementations for the known quantizers and greatly simplifies the internal logic for the quantizing encoders, especially the png encoder.

Additionally I've cleaned up unused encoder options and tweaked the guard range classes to return a more useful message.

I'm really glad @antonfirsov pushed for this, it's much better now. 👍


src/ImageSharp/Processing/Quantization/IQuantizer.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// Gets a value indicating whether to apply dithering to the output image.
+        /// </summary>
+        bool Dither { get; }
+
+        /// <summary>
+        /// Gets the dithering algorithm to apply to the output image.
+        /// </summary>
+        IErrorDiffuser DitherType { get; }
```
tocsoft: not really needed on the interface as its an implementation details inside the `XXXFrameQuantizer`
noting can set it so it will only be configured during construction of a concrete type.

src/ImageSharp/Processing/Quantization/IQuantizer.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// Provides methods for allowing quantization of images pixels with configurable dithering.
+    /// </summary>
+    public interface IQuantizer
+    {
+        /// <summary>
+        /// Gets a value indicating whether to apply dithering to the output image.
+        /// </summary>
+        bool Dither { get; }
```
tocsoft: As with below `DitherType`  property not needed on interface and not even needed on classes as it can be infered by weather the `IErrorDiffuser` is set or not as they are immutable classes.

src/ImageSharp/Processing/Quantization/IQuantizer.cs
JimBobSquarePants: We'll need this to pass to the `FrameQuantizerBase{TPixel}` class as the constructor takes the `IQuantizer` interface. You have got me thinking though about the `Dither` property. We don't need it as long as this is set but we'll have to reverse the default behavior in order to avoid the dev having to pass `null` to a constructor (which I find icky). Currently the default constructor diffuses.

e.g 

- `new OctreeQuantizer()` **255 colors, no diffusion**
- `new OctreeQuantizer(128)` **28 colors, no diffusion**
- `new OctreeQuantizer(DiffuseMode.FloydSteinberg)` **255 colors, FloydSteinberg.Diffusion**
- `new OctreeQuantizer(DiffuseMode.FloydSteinberg, 128)` **128 colors, FloydSteinberg.Diffusion**

I'll probably change the `DitherType` name to `DiffuseMode` also to match the dithering API.

JimBobSquarePants:
@tocsoft   Re. Documentation, Yeah, I would like to write some comprehensive sections on the various image formats, how they work, what they support etc. I really want to make knowledge of things like quantization more accessible to developers since it's so cryptic or absent from System.Drawing.

The maximum palette length really is only a concern of the quantizers and the code we had before was pretty hacky internally.

src/ImageSharp/Processing/Quantization/QuantizationMode.cs
```diff
@@ -4,26 +4,26 @@
 namespace SixLabors.ImageSharp.Processing.Quantization
 {
     /// <summary>
-    /// Provides enumeration over how an image should be quantized.
+    /// Contains reusable static instances of known quantizing algorithms
     /// </summary>
-    public enum QuantizationMode
+    public static class QuantizationMode
```
antonfirsov: :+1: for dropping the enum! I think we should also change the name to `KnownQuantizers` to match our typical convention.

JimBobSquarePants: We don't use the `KnownXXX` anymore within the API, I dropped that with the namespace refactor since it didn't match the normal MS convention.

JimBobSquarePants:
Ok.... 

I've dropped the `Dither` property on `IQuantizer` and renamed the `DitherType` property to `Diffuser` on both that and `IFrameQuantizer<TPixel>`. 

Quantizers dither by default and can be easily set to not do so.

src/ImageSharp/Processing/Quantization/QuantizationMode.cs
antonfirsov: I see. I guess  we do not use simple plural for these classes to avoid collision with namespaces?
`XXXMode` intuitively suggests a closed API to me. Is this the standard MS convention now?

JimBobSquarePants: > `XXXMode` intuitively suggests a closed API to me. 

Yeah, you're right, it does doesn't it. Mode isn't a standard suffix, the naming is usually something that would clash horribly with our namespaces.

I'm happy to change all our static aggregators to use the prefix `Known` again as it does open up the API.

antonfirsov: To me it looks like the `Known` prefix is the least worst of all those options. Any recommendation from MS?

JimBobSquarePants: I'll ask Immo

JimBobSquarePants:
I'm tacking on a naming convention change to the API with this PR.

The convention is as follows.

- `KnownXXX`. Any open-ended collection of algorithms e.g. `KnownQuantizers`, `KnownResamplers`
- `XXXMode`. Any closed operation enumeration  e.g `AnchorPositionMode`, `FlipMode`
- `XXXType`. Any enumeration of a data type e.g `ExifDataType`. 

I also shifted some primitives I'd missed that are shared by the metadata profiles.

I know I shouldn't mix and match in a single PR but I forgot to change branch and I'm too tired to untangle the commits.

src/ImageSharp/Processing/Quantization/OctreeQuantizer.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace SixLabors.ImageSharp.Processing.Quantization
+{
+    /// <summary>
+    /// Allows the quantization of images pixels using Octrees.
+    /// <see href="http://msdn.microsoft.com/en-us/library/aa479306.aspx"/>
+    /// </summary>
+    public class OctreeQuantizer : IQuantizer
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OctreeQuantizer"/> class.
```
antonfirsov: Shouldn the docs mention that dithering is enabled with this ctr? (I'm following the twitter conversaton :) )

JimBobSquarePants: Yeah they should....😋

src/ImageSharp/Processing/Dithering/KnownDitherMatrices.cs
```diff
@@ -8,7 +8,7 @@ namespace SixLabors.ImageSharp.Processing.Dithering
     /// <summary>
     /// Contains reusable static instances of known ordered dither matrices
     /// </summary>
-    public class DitherMode
+    public class KnownDitherMatrices
```
antonfirsov: How about naming this `KnownDitherers` to match `KnowDiffusers`?

src/ImageSharp/Processing/Filters/KnownMatrixFilters.cs
```diff
@@ -9,7 +9,7 @@ namespace SixLabors.ImageSharp.Processing.Filters
     /// <summary>
     /// A collection of known <see cref="Matrix4x4"/> values for composing filters
     /// </summary>
-    public static class MatrixFilters
+    public static class KnownMatrixFilters
```
antonfirsov: Isn't `KnownFilterMatrices` a better naming for this?

JimBobSquarePants:
What's the position on this now. Can I merge it in?

PR #506, opened 2018-03-27
Improve Code Quality

iamcarbon:
### Prerequisites

### Description

This PR cleans up some formatting inconsistencies and removes dead code. It also:

- Makes PngHeader and BmpFileHeader immutable structs (passed by reference)
- Makes ExifValue immutable
- Passes additional readonly struct's by readonly ref (in)
- Moves to a TryRead pattern for reading PngChunks (this is close to being immutable)
- Refactors ExifReader & Writer to use Spans & BinaryPrimitives


src/ImageSharp/Formats/PixelTypeInfo.cs
```diff
@@ -6,10 +6,10 @@ namespace SixLabors.ImageSharp.Formats
     /// <summary>
     /// Contains information about the pixels that make up an images visual data.
     /// </summary>
-    public class PixelTypeInfo
+    public readonly struct PixelTypeInfo
```
antonfirsov: Not sure if it's worth to turn this into a `struct`. The instances are not frequents, so there are no performance benefits, however it makes it easier to users to make mistakes:
Default-constructed instances are invalid (`BitsPerPixel == 0`) + it introduces the cognitive load of struct/value semantics. (Note: we plan to add more fields like a reference to the pixel type etc.)
Thoughts?

antonfirsov:
Great PR!

The question is about the level of impact on public API, because if there are significant breaking changes, we should consider adding them quickly to beta-3. (Maybe separate PR?)

src/ImageSharp/Formats/PixelTypeInfo.cs
iamcarbon: Agreed. If this class grows, we'll want it to be a reference type. Reverting.

JimBobSquarePants:
If you can get a working build in the next ten minutes or so I’m happy to add this to the beta we’re packaging up just  now

iamcarbon:
I think there's only only a single change to the public API (changing an IEnumerable to an IList in the ExifProfile) -- the rest of the changes are internal.

Not critical to get into Beta 3, but can't hurt.

antonfirsov:
Those API changes seem to be really small, maybe they won't gonna hurt anyone, or will they?

TBH, I'm afraid of pushing last-minute changes into releases, even if they seem super-safe, and done by smart people :) It's too late here to trust my code-review skills,

JimBobSquarePants:
Ok let’s wait then and release first. I’m reviewing on my phone.

dlemstra:
Why do we need to change the ExifValue API? Not sure what the benefit of making it immutable is.

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -106,7 +106,7 @@ public ExifParts Parts
         /// <summary>
         /// Gets the values of this EXIF profile.
         /// </summary>
-        public IEnumerable<ExifValue> Values
+        public IList<ExifValue> Values
```
dlemstra: This list should not be mutable.

iamcarbon: @dlemstra -- Agreed.

iamcarbon:
@dlemstra -- I'm seeing if we can make ExifValue a structure -- it's immutability will be important if we do. Will revert it's immutability if this doesn't pan out.

antonfirsov:
@dlemstra How large can the count of `ExifProfile.Values` grow in extreme cases? If it can grow high, I think making it struct might be benefitical from performance point of view. 

I've seen lot's of exif allocations the last time I profiled my LoadTest project, but can't recall where it was exactly.

iamcarbon:
Yep, pulling out of the idea of making ExifValue a struct -- it's too big and we won't be able to access it by reference until we get a list that exposes an ItemRef method. I'm going to revert the immutability.

@dlemstra We currently have a SetValue method on ExifProfile that's equivalent to setting it directly from the values list. Going to revert the immutability & WithValue method, but think we should consider making the setter internal -- forcing everything through ExifProfile.SetValue.  Let me know your thoughts.

iamcarbon:
OK! I think this is ready for review. I'll start diving deeper in the next pull request.

iamcarbon:
OK, done for real. #promise

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -160,6 +160,31 @@ public ExifValue GetValue(ExifTag tag)
...
 
+        /// <summary>
+        /// Conditionally the value of the tag if it exists.
+        /// </summary>
+        /// <param name="tag">The tag of the EXIF value.</param>
+        /// <param name="value">The value of the tag, if found.</param>
+        /// <returns>
+        /// The <see cref="bool"/>.
+        /// </returns>
+        public bool TryGetValue(ExifTag tag, out ExifValue value)
```
JimBobSquarePants: `TryGetValue` is the best pattern. The more of that the better. 👍

src/ImageSharp/Primitives/LongRational.cs
```diff
@@ -276,78 +127,100 @@ public string ToString(IFormatProvider provider)
...
-        /// Finds the greatest common divisor of two <see cref="long"/> values.
+        /// Create a new instance of the <see cref="LongRational"/> struct from a double value.
         /// </summary>
-        /// <param name="left">The first value</param>
-        /// <param name="right">The second value</param>
-        /// <returns>The <see cref="long"/></returns>
-        private static long GreatestCommonDivisor(long left, long right)
+        /// <param name="value">The <see cref="double"/> to create the instance from.</param>
+        /// <param name="bestPrecision">Whether to use the best possible precision when parsing the value.</param>
+        public static LongRational FromDouble(double value, bool bestPrecision)
```
JimBobSquarePants: Much cleaner 👏

JimBobSquarePants:
@antonfirsov @tocsoft @dlemstra Any objections to this? LGTM.

src/ImageSharp/MetaData/Profiles/Exif/ExifReader.cs
```diff
@@ -55,96 +58,99 @@ private int RemainingLength
...
+        private unsafe string ConvertToString(ReadOnlySpan<byte> buffer)
         {
-            string result = Encoding.UTF8.GetString(data, 0, data.Length);
+#if NETSTANDARD1_1
+            byte[] bytes = buffer.ToArray();
+
+            string result = Encoding.UTF8.GetString(bytes, 0, buffer.Length);
+
+#else
+            byte* pointer = (byte*)Unsafe.AsPointer(ref MemoryMarshal.GetReference(buffer));
```
antonfirsov: We need to pin this! (`fixed (...) { ... }`)

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -92,21 +93,17 @@ public ExifProfile(ExifProfile other)
...
-            get;
-            set;
-        }
+        public ExifParts Parts { get; set; }
 
         /// <summary>
         /// Gets the tags that where found but contained an invalid value.
         /// </summary>
-        public IEnumerable<ExifTag> InvalidTags => this.invalidTags;
+        public IReadOnlyList<ExifTag> InvalidTags => this.invalidTags;
```
antonfirsov: @dlemstra are you happy with `IReadOnlyList`? It seems fine to me, because it lets indexing the collection while keeping it immutable.

src/ImageSharp/MetaData/Profiles/Exif/ExifReader.cs
```diff
@@ -15,38 +19,37 @@ namespace SixLabors.ImageSharp.MetaData.Profiles.Exif
...
+            this.exifData = exifData;
+        }
+
+        private delegate TDataType ConverterMethod<TDataType>(ReadOnlySpan<byte> data);
 
         /// <summary>
         /// Gets the invalid tags.
         /// </summary>
-        public IEnumerable<ExifTag> InvalidTags => this.invalidTags;
+        public IList<ExifTag> InvalidTags => this.invalidTags;
```
antonfirsov: Is this list intended to be mutable from the outside?

iamcarbon:
Good catches. I think it’s unlikely the GC will pause the thread during these inclined calls and relocate the array, but unless I can a find a guarantee — will pin.

The Encoder classes are getting Span overloads in 2.1 — hoping to #if def there later.

iamcarbon:
Yep, needs pinned! Thanks for reviewing and catching this @antonfirsov !!!

antonfirsov:
Yeah pinning is the standard practice flwhen consuming pointer API-s, it's cheap for short periods AFAIK.

src/ImageSharp/MetaData/Profiles/Exif/ExifReader.cs
iamcarbon: Nope. Changing to ReadOnly.

JimBobSquarePants:
OK, now the pointer is pinned I'm gonna merge this in. Thanks for the effort @iamcarbon and thanks everyone for reviewing. 👍

PR #511, opened 2018-03-28
Add Scaled Vector Packing/Unpacking

JimBobSquarePants:
### Prerequisites

### Description
This PR adds two new methods to the `IPixel` interface.

```
/// <summary>
/// Sets the packed representation from a scaled <see cref="Vector4"/>.
/// </summary>
/// <param name="vector">The vector to create the packed representation from.</param>
void PackFromScaledVector4(Vector4 vector);

/// <summary>
/// Expands the packed representation into a scaled <see cref="Vector4"/>
/// with values clamped between <value>0</value> and <value>1</value>.
/// The vector components are typically expanded in least to greatest significance order.
/// </summary>
/// <returns>The <see cref="Vector4"/>.</returns>
Vector4 ToScaledVector4();
```
Utilizing these methods we make it much more simple and quicker to clone from one pixel format to another using `CloneAs`.


tocsoft:
do we need both the scaled and non-scaled vector4 options? can't we just ensure that the current vector4 APIs always expect a scaled vector?

JimBobSquarePants:
The current `ToVector4` `PackFromVector4` methods produce the same values as MonoGame. I’m sure there’s good reason in gaming engines for those ranges so I would be inclined to keep both.

src/ImageSharp/ImageFrame{TPixel}.cs
```diff
@@ -246,27 +246,23 @@ internal ImageFrame<TPixel2> CloneAs<TPixel2>()
...
-                        for (int x = 0; x < target.Width; x++)
-                        {
-                            var color = default(TPixel2);
-                            color.PackFromVector4(scaleFunc(pixels[x, y].ToVector4()));
-                            targetPixels[x, y] = color;
-                        }
-                    });
-            }
+                        ref var pixel = ref targetRow[x];
+                        pixel.PackFromScaledVector4(sourceRow[x].ToScaledVector4());
```
antonfirsov: The only change I strongly suggest is to introduce new batch primitives for `PackFromScaledVector4()` and `ToScaledVector4()` in `PixelOperations<T>` .
We could drop the for loop then, and push the data through a temporary `Vector4` buffer.

JimBobSquarePants: I’m not convinced by the need for this. We don’t have it for the existing vector methods and we don’t need any temporary allocations just now.

JimBobSquarePants: My bad, they do exist. Was looking in the wrong place! Will update now

antonfirsov: `ParallelFor.WithTemporaryBuffer` could help providing the buffer efficiently.

JimBobSquarePants: I've pushed the PixelOperations plus tests but in this instance I do not feel that we warrant that allocation at all. We're over-complicating a very simple operation. We're looping in PixelOperations! 

What I have noticed though is that we cannot run a parallel operation within `ImageFrame<T>` using anything other the default configuration `ParallelOptions` values for the action. 

We are passing `MemoryManager` instance to the frame constructor when we should be passing the `Configuration` instance.

antonfirsov: We need to solve this then, but it could be done in a  separate PR then to minimalize the current changeset.

antonfirsov: I'll leave a few comments to clarify my point though:

> We're over-complicating a very simple operation. We're looping in PixelOperations!

The point of `PixelOperations<T>` is to allow SIMD optimized path for specific operations. The [optimized path for unscaled Rgba <-> Vector4 packing](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/PixelFormats/Rgba32.PixelOperations.cs#L39) is about 2x faster than the standard.

> in this instance I do not feel that we warrant that allocation at all

[`ParallelFor.WithTemporaryBuffer`](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Common/Helpers/ParallelFor.cs#L30) allows us to solve the above point from only `ParallelOptions.MaxDegreeOfParallelism` number of buffer "allocations" (which are actually pooled array usages!).

JimBobSquarePants: Ok, forgot about the `Rgba32` optimizations you did. I've made the change and am using the default configuration for now. I'll open another PR for passing the `Configuration` instance

PR #512, opened 2018-03-30
Extend AppVeyor coverage on more frameworks and platforms

antonfirsov:
### Prerequisites

### Description
- `appveyor.yml` is extended to run tests on several platforms & also cover 32bit architecture.
  - Each configuration is executed as a separate job
- Individual configurations are executed by `run-tests.ps1`
- Test execution now targets all of these: `net462`, `net47`, `net471`, `netcoreapp2.0`
- Unfortunately, `netcoreapp1.1` has been left uncovered, because of the `System.Drawing.Common` reference. It should be powered with the same RyuJIT as the old full frameworks, so I hope we are going to be fine.

### How to debug locally
- Set soultion platform to x86 
- Build the project
- Visual Studio test runner will execute the first configuration (`net471`) in the selected platform
- There should be ~30-40 failing tests

### WIP
- ~[ ] Locate and fix the 32 bit bugs~ Sporadic test failures do not seem to occur on AppVeyor now, but I still have them sometimes locally.
- [x] introduce mono execution. (Done, but disabled for now)


antonfirsov:
Good news is that the majority of the 32bit test failures is caused by floating point inaccuracies rooted in architecture mismatch of reference / actual output. This could be fixed by increasing the tolerance.

The most serious difference I noticed is in `BlackWhiteTest.ApplyBlackWhiteFilter()`.

antonfirsov:
And yeah ... one major disadvantage of this change is that appveyor is going to run for ages :\

antonfirsov:
Dropped the `RunExtendedTests.cmd` test execution mode. All configurations are executed as separate job using `run-tests.ps1`. I wish we could execute them in parallel. Is it possible for open-source licenses?

antonfirsov:
I consider this PR ready for review, if the current build passes.

@tocsoft was there any reason for automatic test execution when calling `build.ps1` locally? I removed it, because there are too many runtime+architecture combinations now. It's probably better to run tests explicitly using `run-tests.ps1`.

build.ps1
```diff
@@ -100,9 +100,17 @@ dotnet build -c Release /p:packageversion=$version
 
 if ($LASTEXITCODE ){ Exit $LASTEXITCODE }
 
-if ( $env:CI -ne "True") {
-    dotnet test ./tests/ImageSharp.Tests/ImageSharp.Tests.csproj --no-build -c Release
-}
+#
+# TODO: DO WE NEED TO RUN TESTS IMPLICITLY?
```
tocsoft: nope  feel free to drop

PR #514, opened 2018-04-04
Drop NETSTANDARD 1.1

iamcarbon:
### Prerequisites

### Description

This PR drops support for .NETStandard 1.1. This eliminates the need to conditionally exclude the file system APIs.

Most importantly, this also helps ensure that we don't take on any legacy users as we build for the future.

### DIFF
--- a/src/ImageSharp.Drawing/ImageSharp.Drawing.csproj
+++ b/src/ImageSharp.Drawing/ImageSharp.Drawing.csproj
@@ -5,7 +5,7 @@
         <VersionPrefix Condition="$(packageversion) != ''">$(packageversion)</VersionPrefix>
         <VersionPrefix Condition="$(packageversion) == ''">0.0.1</VersionPrefix>
         <Authors>SixLabors and contributors</Authors>
-        <TargetFrameworks>netstandard1.1;netstandard2.0</TargetFrameworks>
+        <TargetFrameworks>netstandard2.0</TargetFrameworks>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <AssemblyName>SixLabors.ImageSharp.Drawing</AssemblyName>
--- a/src/ImageSharp/Configuration.cs
+++ b/src/ImageSharp/Configuration.cs
@@ -9,9 +9,7 @@ using SixLabors.ImageSharp.Formats.Bmp;
 using SixLabors.ImageSharp.Formats.Gif;
 using SixLabors.ImageSharp.Formats.Jpeg;
 using SixLabors.ImageSharp.Formats.Png;
-#if !NETSTANDARD1_1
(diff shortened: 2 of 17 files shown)


iamcarbon:
We may even consider dropping 1.x entirely and targeting .NETSTANDARD2.0 & .NETCOREAPP2.1 for the 1.0 release. .NETCOREAPP2.1 would be dependency free.

iamcarbon:
Notable: .NETCORE 1.x is supported until June 27, 2019.

antonfirsov:
Originally, I was a big supporter of this move, but now I'm concerned that we can hurt enterprise users with it. I'm fine, if this is only a small group of people, but unfortunately, we don't know for sure.

iamcarbon:
It would be nice to understand the user needing .NET 1.1 / .NET 4.5.1 support, why, and for how long. 

I understand the case where enterprises keep around legacy software running old servers. But are they also deploying new solutions to these same systems? 

We don't have these users yet. Do we want them?

JimBobSquarePants:
I'm also torn on this one. But I'm strongly erring toward keeping the target. I'll explain why.

My plan has always been to support the various target frameworks for as long as Microsoft do and **no longer**. 

Since netstandard 1.1 is the lowest target we can build against that gives us support for .NET 4.5.2 I planned to support that until it reached end-of-life.

According to this StackOverflow [answer](https://stackoverflow.com/a/3013935) though that means supporting 1.1 until 2**023-01-10**

Do I want to support it for that long? That's a long time away!

Realistically we have to ask ourselves. Does supporting 1.1 actively harm us? It is:

- Dramatically increasing the development complexity of the library?
- Severely limiting our ability to improve the performance of the library?
- Preventing future feature development?

I'm inclined to say no to the above questions. 

I absolutely hate having `#ifdef`'s in the codebase (super-super hate) but I think we have gotten off very lightly so far. Microsoft have been brilliant in offering many of the new API's to lower targets so we've been able to capitalize well there. That may change for future performance enhancements (SIMD intrinsics) but I don't know.

I'm not too concerned about slow `Span<T>`. We can use `Unsafe<T>` in hotpaths and keep the performance "_fast enough"_. I doubt very much we shall ever be the bottleneck in a legacy targeted application.

I'm planning on introducing dual licensing very soon and I do want to snag every available potential customer we can. If keeping the target low (lower than any of the competition) helps us do that then great. This project has taken 3 years of my life so far and if I don't get to make a decent living off the back of it I shall be crushed.

iamcarbon:
Good getting the teams thoughts on this! 

Given:

- Microsoft continues to provide new features out of band and target .NET Standard 1.1 and support .NET 4.5.2 until 2020+ -- and things just work;
- These users will provide more resources (read money) to the project than the totality of the effort to support them (agreed on all points that it is currently low);
- We don't compromise on speed (on modern runtimes) [compatibility = bonus];

I agree!

PR #530, opened 2018-04-15
Simplify Bmp Decoding / Encoding

iamcarbon:
### Prerequisites

### Description
- Eliminates the EndianBinaryWriter from the BMP encoder. 
- Writes the packed BmpHeader & BmpInfoHeader directly to the buffer


iamcarbon:
OK, this is ready for review.

Working to kill the EndianBinary(Reader / Writer) now that we have all these nice new memory helpers.

src/ImageSharp/Formats/Bmp/BmpInfoHeader.cs
```diff
@@ -8,28 +12,55 @@ namespace SixLabors.ImageSharp.Formats.Bmp
     /// the screen.
     /// <see href="https://en.wikipedia.org/wiki/BMP_file_format"/>
     /// </summary>
-    internal sealed class BmpInfoHeader
+    [StructLayout(LayoutKind.Sequential, Pack = 1)]
```
JimBobSquarePants: Clever!

src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
```diff
@@ -54,96 +50,58 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
...
             var fileHeader = new BmpFileHeader(
                 type: 19778, // BM
                 offset: 54,
                 reserved: 0,
                 fileSize: 54 + infoHeader.ImageSize);
 
-            WriteHeader(writer, fileHeader);
-            this.WriteInfo(writer, infoHeader);
-            this.WriteImage(writer, image.Frames.RootFrame);
+            byte[] buffer = new byte[40]; // TODO: stackalloc
```
JimBobSquarePants: It's probably better to actually do the stackalloc now. We'll forget otherwise.

iamcarbon: Annoyingly, we can't use stack allocated memory (prior to NETCORE 2.1) for the stream reads. Removing these comments for now.

src/ImageSharp/Formats/Bmp/BmpFileHeader.cs
```diff
@@ -44,12 +48,25 @@ public BmpFileHeader(short type, int fileSize, int reserved, int offset)
...
         /// </summary>
         public int Offset { get; }
+
+        public static BmpFileHeader Parse(Span<byte> data)
+        {
+            return MemoryMarshal.Cast<byte, BmpFileHeader>(data)[0];
+        }
+
+        public unsafe void WriteTo(Span<byte> buffer)
+        {
```
antonfirsov: How about this without pinning? 
```C#
ref BmpFileHeader dest = ref Unsafe.As<byte, BmpFileHeader>(ref MemoryMarshal.GetReference(buffer));
dest = this;
```
Doesn't really matter here, but it's most likely faster :)

src/ImageSharp/Formats/Bmp/BmpInfoHeader.cs
```diff
@@ -91,5 +122,40 @@ internal sealed class BmpInfoHeader
...
+
+        /// <summary>
+        /// Parses the BITMAPCOREHEADER consisting of the headerSize, width, height, planes, and bitsPerPixel fields (12 bytes).
+        /// </summary>
+        /// <param name="data">The data to parse,</param>
+        /// <returns>Parsed header</returns>
+        /// <seealso href="https://msdn.microsoft.com/en-us/library/windows/desktop/dd183372.aspx"/>
+        public static BmpInfoHeader ParseCore(ReadOnlySpan<byte> data)
+        {
+            return new BmpInfoHeader(
```
antonfirsov: Just wondering why don't you do the casting trick here?

iamcarbon: The core header is only 12 bytes -- so  the cast won't work here. Adding one more check for the full cast case to make sure we're always passing 40 bytes...

src/ImageSharp/Formats/Bmp/BmpFileHeader.cs
iamcarbon: Much nicer!

iamcarbon:
@antonfirsov Good feedback w/ suggestions applied!

antonfirsov:
Well done! Should we merge this when appveyor is finished?

iamcarbon:
Travis, are you ok?

PR #534, opened 2018-04-18
Disposing data chunk instead of explicit returning it to ArrayPool in PngDecoderCore.Identify

denisivan0v:
### Prerequisites

### Description
The new `Image.Identify` API was introduced in #292. In #475 memory management was significantly improved, but there is still a case for `png` format when `ArrayPool` is used directly when calling `Image.Identify`, not with `IManagedByteBuffer` API.
This leads to the issue when `ArgumentException` is throwing in `ArrayPool.Return` method with this message:

```
The buffer is not associated with this pool and may not be returned to it.
Parameter name: array
```

This PR fixes this problem.

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -313,11 +313,7 @@ namespace SixLabors.ImageSharp.Formats.Png
                     }
                     finally
                     {
-                        // Data is rented in ReadChunkData()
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Hi @denisivan0v 

Good catch! I'll get this merged once the CI finished 👍

denisivan0v:
Many thanks @JimBobSquarePants! 🎉

JimBobSquarePants:
No worries! Thanks for the PR! 😄

PR #545, opened 2018-04-23
[WIP] Quantization Refactoring

iamcarbon:
### Prerequisites

### Description

This is a WIP to refactor the quantization pipeline to:

- allocate from the memory manager
- share palette and image pixel buffers between frames

TODO

- [ ] Improve test coverage
- [ ] Create a benchmark on a 100+ frame GIF (it should have dramatically fewer allocations and forced GC's)


iamcarbon:
Closing for now...

src/ImageSharp/Processing/Quantization/FrameQuantizers/FrameQuantizerBase{TPixel}.cs
```diff
@@ -51,14 +51,18 @@ protected FrameQuantizerBase(IQuantizer quantizer, bool singlePass)
         public IErrorDiffuser Diffuser { get; }
 
         /// <inheritdoc/>
-        public virtual QuantizedFrame<TPixel> QuantizeFrame(ImageFrame<TPixel> image)
+        public virtual void QuantizeFrame(ImageFrame<TPixel> image, Span<byte> quantizedPixels, Span<TPixel> quantizedPalette, out int quantizedPaletteLength)
         {
             Guard.NotNull(image, nameof(image));
 
+            if (quantizedPixels.Length != image.Width * image.Height)
```
iamcarbon: Should use Guard...

src/ImageSharp/Processing/Quantization/QuantizedFrame{TPixel}.cs
```diff
@@ -1,61 +0,0 @@
...
-using System;
-using SixLabors.ImageSharp.PixelFormats;
-
-namespace SixLabors.ImageSharp.Processing.Quantization
-{
-    /// <summary>
-    /// Represents a quantized image frame where the pixels indexed by a color palette.
-    /// </summary>
-    /// <typeparam name="TPixel">The pixel format.</typeparam>
-    public class QuantizedFrame<TPixel>
```
JimBobSquarePants: I'd actually like to keep this as a separate entity if I can.

iamcarbon: Reverted

iamcarbon:
@JimBobSquarePants This eliminates a ton of garbage encoding long GIF frame sequences.

Still loads more coming from dithering.

iamcarbon:
Also, still ambiently wrapping my brain around all the Huffman logic. Not ready to touch that yet.

JimBobSquarePants:
I'll have to properly download and have a look. 

I envisioned we would simply alter `QuantizedFrame<TPixel>` to have an `IBuffer<TPixel>` for the palette and IBuffer2D<byte> for the pixels while making the frame itself implement `IDisposable`

JimBobSquarePants:
@iamcarbon  Yeah dithering is slow. Anything you can do to improve that would be great. 

Weeping at resize benchmarks just now. We're much slower than SD again because we do the resize correctly with alpha premultiplication which adds an extra step.

antonfirsov:
@JimBobSquarePants @iamcarbon this is a good quick improvement on quantizers, however I know for sure the root cause for the long time and high memory usage of `OctreeQuantizer` (therefore default gif encoder).

There is an opportunity for a **huge** asymptotic optimization, because the current tree is unbalanced + not limited in depth. It's also possible to store the tree in a way that embraces memory locality, but this should be the second step.

I believe that these steps together should bring those 4-6 minutes well below 1min, but this is big a task for a couple of days.

JimBobSquarePants:
@iamcarbon I think we can speed things up further quite simply by introducing a global color palette option. 

Instead of using the Octree (Or whatever) throughout, and generating a local palette for each frame, when in global mode we could use the PaletteQuantizer to quantize the subsequent frames based on the palette created by the first frame. 

This should speed up processing, memory consumption, and reduce the file size of the output images. Making it optional (probably default) we can offer local tables for people who need them.

JimBobSquarePants:
@iamcarbon I'm gonna close in this favour of #637 as I've covered the memory improvements you made plus added another heap of improvements.

I'd love to see how much faster your crazy gif encodes now!

PR #546, opened 2018-04-26
Add Projective Transforms

JimBobSquarePants:
### Prerequisites

### Description
Adds the ability to perform projective transforms (perspective, taper) on images


JimBobSquarePants:
@antonfirsov Surprising comparison results here. 32bit passing 64 bit failing? WTF?

antonfirsov:
@JimBobSquarePants the difference is not big however. I suspect the root cause is a difference between SIMD and software implementations of `Vector4` operations.

Hope it's not CPU dependent & locally reproducible.

tests/ImageSharp.Tests/Processing/Transforms/ProjectiveTransformTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            where TPixel : struct, IPixel<TPixel>
+        {
+            var taperMatrixComparer = ImageComparer.TolerantPercentage(0.2f);
+            using (Image<TPixel> image = provider.GetImage())
+            {
+                Matrix4x4 m = ProjectiveTransformHelper.CreateTaperMatrix(image.Size(), taperSide, taperCorner, .5F);
+                image.Mutate(i => { i.Transform(m); });
+
+                string testOutputDetails = $"{taperSide}-{taperCorner}";
+                image.DebugSave(provider, testOutputDetails);
```
antonfirsov: @JimBobSquarePants added some tests to cover CreateTaperMatrix() parameters. I may get it wrong but the `XxxOrYyy` stuff might be the inverse of the intended. Can you check the output?

src/ImageSharp/Processing/Transforms/ProjectiveTransformHelper.cs
```diff
@@ -0,0 +1,166 @@
...
+        /// <summary>
+        /// Creates a matrix that performs a tapering projective transform.
+        /// <see href="https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/graphics/skiasharp/transforms/non-affine"/>
+        /// </summary>
+        /// <param name="size">The rectangular size of the image being transformed.</param>
+        /// <param name="taperSide">An enumeration that indicates the side of the rectangle that tapers.</param>
+        /// <param name="taperCorner">An enumeration that indicates on which corners to taper the rectangle.</param>
+        /// <param name="taperFraction">The amount to taper.</param>
+        /// <returns>The <see cref="Matrix4x4"/></returns>
+        public static Matrix4x4 CreateTaperMatrix(Size size, TaperSide taperSide, TaperCorner taperCorner, float taperFraction)
```
antonfirsov: Cool suff! Covers basic cases & good for testing without getting outside the scope of the library.

tests/ImageSharp.Tests/Processing/Transforms/ProjectiveTransformTests.cs
JimBobSquarePants: I'll have to do some tests locally to ensure the output is expected. The class is based on a Skia example and and old Charles Petzold blog post.

src/ImageSharp/Processing/Transforms/ProjectiveTransformHelper.cs
JimBobSquarePants: Thanks for adding the extra tests, they affirm the expected output well.

JimBobSquarePants:
@antonfirsov Back to you; do your magic please.

antonfirsov:
Bottom-**Left**OrTop:
transform_withtapermatrix_rgba32_solid30x30_ 255 0 0 255 _bottom-leftortop

Bottom-**Right**OrBottom:
transform_withtapermatrix_rgba32_solid30x30_ 255 0 0 255 _bottom-rightorbottom

@JimBobSquarePants you sure `TaperCorner` works fine? Isn't it the inverse of the intended? Or I'm getting it wrong?

JimBobSquarePants:
>Bottom-**Left**OrTop:
transform_withtapermatrix_rgba32_solid30x30_ 255 0 0 255 _bottom-leftortop
>
>Bottom-**Right**OrBottom:
transform_withtapermatrix_rgba32_solid30x30_ 255 0 0 255 _bottom-rightorbottom
>
>@JimBobSquarePants you sure `TaperCorner` works fine? Isn't it the inverse of the intended? Or I'm getting it wrong?

@antonfirsov I _think_ it's correct. 

In the first example we are choosing to taper the "Bottom" side on either the "Left" or "Top" corner. Since "Left" is possible regardless of whether "Top" is impossible "Left" is chosen. 

In the seconds example we are choosing to taper the "Bottom" side on either the "Right" or "Bottom" corner. Since "Right" is possible regardless of whether "Bottom" is possible "Right" is chosen. 

The original article with source code for the class this is based on can be found here:

http://www.charlespetzold.com/blog/2009/07/Taper-Transforms-with-Matrix3DProjection-An-Analytical-Approach.html

SkiaSharp have an example here:
https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/graphics/skiasharp/transforms/non-affine

I can attempt to compare against the demos. (Feel free to have a look first though) https://developer.xamarin.com/samples/xamarin-forms/SkiaSharpForms/Demos/

antonfirsov:
LOL .. that image!

But isn't the first example tapering towards the **right** corner? Or should I start over my English courses? :D

JimBobSquarePants:
Yeah I felt a little guilty posting that 😄 

Here's the comparative images side by side and scaled.

Bottom-**Left**OrTop:
taper-bottom-left-or-top
taper-bottom-left-or-top


Bottom-**Right**OrBottom:

taper-bottom-right-or-bottom

antonfirsov:
It seems to me that the original definition is wrong, because the "tapered" edge is not on the left (but on the right!) corner of the image.

image


But again, it might be just my wrong understanding of the term "tapering".

JimBobSquarePants:
Ah no... I can see how this is confusing but the language is correct. 

A classic definition of taper would be 

>to become progressively smaller toward one end.

You're tapering towards the bottom edge and choosing which corner to taper with.

antonfirsov:
I understand now.
The PR is in a perfect state, sorry for the long linguistic session! :)

antonfirsov:
Well .. almost. Haven't noticed, the tests are failing again, gonna check it out now.

JimBobSquarePants:
floating-points-from-hell

Haha! 🤣 

It's amazing really the trouble we've had with this. Will be happy when MS push all their `Vector2` fixes.

PR #549, opened 2018-04-28
Better Jpeg Identify

JimBobSquarePants:
### Prerequisites

### Description
Picking up where #547 left off this PR reduces the amount of work both jpeg decoders do when identifying jpegs to a bare minimum. It also fixes a bug where EXIF metadata was not read using the Golang decoder.

The PDFJs Decoder is now stonkingly fast at identifying jpegs

```
         Method | Runtime |                    TestImage |         Mean |        Error |      StdDev |    Gen 0 | Allocated |
--------------- |-------- |----------------------------- |-------------:|-------------:|------------:|---------:|----------:|
 IdentifyGolang |     Clr |  Jpg/baseline/Calliphora.jpg | 2,407.572 us |   955.188 us |  53.9699 us |   7.8125 |   21184 B |
  IdentifyPdfJs |     Clr |  Jpg/baseline/Calliphora.jpg |     6.662 us |     3.154 us |   0.1782 us |   0.3586 |     768 B |
 IdentifyGolang |    Core |  Jpg/baseline/Calliphora.jpg | 2,679.661 us | 4,995.331 us | 282.2455 us |   7.8125 |   21144 B |
  IdentifyPdfJs |    Core |  Jpg/baseline/Calliphora.jpg |     5.735 us |     1.188 us |   0.0671 us |   0.3586 |     760 B |
 IdentifyGolang |     Clr | Jpg/baseline/jpeg420exif.jpg | 8,638.402 us | 4,777.137 us | 269.9171 us | 250.0000 |  533175 B |
  IdentifyPdfJs |     Clr | Jpg/baseline/jpeg420exif.jpg | 1,465.673 us |   219.289 us |  12.3902 us | 244.1406 |  512709 B |
 IdentifyGolang |    Core | Jpg/baseline/jpeg420exif.jpg | 8,968.104 us | 4,480.158 us | 253.1373 us | 250.0000 |  533113 B |
  IdentifyPdfJs |    Core | Jpg/baseline/jpeg420exif.jpg | 1,528.228 us |   698.713 us |  39.4786 us | 244.1406 |  512685 B |
```

I'd really like to start the process of switching the decoders over now. It means rewriting a lot of the unit tests though.


src/ImageSharp/Formats/Jpeg/PdfJsPort/Components/DoubleBufferedStreamReader.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            this.buffer?.Dispose();
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private int ReadByteSlow()
+        {
+            this.stream.Seek(this.position, SeekOrigin.Begin);
```
antonfirsov: Wonder why do we need to seek the stream on every read?

JimBobSquarePants: It was easier than tracking both positions. We could compare I guess then only seek if required.

JimBobSquarePants:
@antonfirsov Can you have a look over my changes here please. This should be good to pull in for now. If you can figure out why bumping up the `MinGetBits` const breaks some images that would be lovely, I've added my thoughts to the codebase.

@iamcarbon Some of this might interest you as the last commit changed the byte reading within the scan decoder to more closely match libjpeg. There's a much better potential for implementing fast Huffman decoding now which could save ~10ms on decoding. I'm stuck but if you are keen to take on the challenge let me know.

src/ImageSharp/Formats/Jpeg/GolangPort/OrigJpegDecoderCore.cs
```diff
@@ -361,12 +361,11 @@ public void ParseStream(Stream stream, bool metadataOnly = false)
 
                         break;
                     case OrigJpegConstants.Markers.SOS:
-                        if (metadataOnly)
+                        if (!metadataOnly)
```
antonfirsov: After "ignoring" SOS, the decoder will start scanning through the contents of the SOS block using tens (hundreds? thousands?) of `FindNextFileMarker()` calls returning in invalid "file markers". I don't think this is the expected behavior.

Can't this behavior push the decoder into a faulted state leading to ignore some further file markers it shouldn't ignore, even in metadata-only state?

antonfirsov: Accidentally commented this on `OrigJpegDecoderCore` instead of `PdfJsDecoderCore`. 

Actually I only investigetad the latter, and the issue is present here! Haven't checked `OrigJpegDecoderCore`.

src/ImageSharp/Formats/Jpeg/PdfJsPort/PdfJsJpegDecoderCore.cs
```diff
@@ -206,119 +208,122 @@ public IImageInfo Identify(Stream stream)
...
+                    switch (fileMarker.Marker)
+                    {
+                        case PdfJsJpegConstants.Markers.SOF0:
+                        case PdfJsJpegConstants.Markers.SOF1:
+                        case PdfJsJpegConstants.Markers.SOF2:
+                            this.ProcessStartOfFrameMarker(remaining, fileMarker, metadataOnly);
+                            break;
+
+                        case PdfJsJpegConstants.Markers.SOS:
+                            if (!metadataOnly)
```
antonfirsov: .. so the issue happens here actually.

src/ImageSharp/Formats/Jpeg/GolangPort/OrigJpegDecoderCore.cs
JimBobSquarePants: Does this marker declare it's length? If so we might be able to simply skip the length.

antonfirsov: I guess [it should](
http://vip.sugovica.hu/Sardi/kepnezo/JPEG%20File%20Layout%20and%20Format.htm).

However, `ProcessSOS` does not depend on it as far as I understand.

JimBobSquarePants: We can just skip the remaining value same as the other markers.

src/ImageSharp/Formats/Jpeg/PdfJsPort/Components/PdfJsScanDecoder.cs
```diff
@@ -432,66 +412,86 @@ private void DecodeMcuACSuccessive(ref PdfJsHuffmanTable acHuffmanTable, PdfJsFr
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private int ReadBit(DoubleBufferedStreamReader stream)
+        private bool TryReadBit(DoubleBufferedStreamReader stream, out int bits)
```
antonfirsov: Why this change?

Exchanging an `int` + a `bool` through the stack is slower than doing the same using only an `int`, and we are on a hot path here AFAIK.

antonfirsov:
#### TryReadBit
Ok, the `TryReadBit()` call is not that hot and it might be worth to communicate the result through a return bool for better readability.

#### metadataOnly
Unfortunately, the `metadataOnly` mode is still results in reading 721 invalid "file markers" for *Snake.jpg*. I think the semantic of `remaining` might be different for SOS. If you debug through `MetaDataIsParsedCorrectly_PdfJs` with `useIdentify == false` VS `useIdentify == true` the issue becomes clear!

Is it possible to place an APP marker (or any other metadata-related marker) after a SOS? If not, we are probably fine if we just break out of the while loop instead of continuing in a faulted state.

antonfirsov:
I think it's very unlikely (if not impossible) to find an APP marker after a Scan. If you can't find abetter solution, I vote for jumping out of the loop when `metaData == true` and we reach SOS. Feels like, the chance of incorrect behavior is lower if we do so.

JimBobSquarePants:
Ok, I’m happy to go with that. Will change in a mo

antonfirsov:
I guess [this article](http://dev.exiv2.org/projects/exiv2/wiki/The_Metadata_in_JPEG_files) confirms my assumption. Of course, nothing is official in Jpegland :)

JimBobSquarePants:
Ok @antonfirsov now we're cooking!

```
         Method | Runtime |                    TestImage |         Mean |       Error |     StdDev |    Gen 0 | Allocated |
--------------- |-------- |----------------------------- |-------------:|------------:|-----------:|---------:|----------:|
 IdentifyGolang |     Clr |  Jpg/baseline/Calliphora.jpg |     7.552 us |   1.2161 us |  0.0687 us |  10.0937 |   21190 B |
  IdentifyPdfJs |     Clr |  Jpg/baseline/Calliphora.jpg |     1.797 us |   0.3151 us |  0.0178 us |   0.3605 |     760 B |
 IdentifyGolang |    Core |  Jpg/baseline/Calliphora.jpg |     7.843 us |   0.2068 us |  0.0117 us |  10.0403 |   21144 B |
  IdentifyPdfJs |    Core |  Jpg/baseline/Calliphora.jpg |     1.913 us |   0.3518 us |  0.0199 us |   0.3548 |     752 B |
 IdentifyGolang |     Clr | Jpg/baseline/jpeg420exif.jpg | 1,683.887 us | 849.9576 us | 48.0242 us | 253.9063 |  533227 B |
  IdentifyPdfJs |     Clr | Jpg/baseline/jpeg420exif.jpg | 1,465.061 us | 404.3200 us | 22.8448 us | 244.1406 |  512705 B |
 IdentifyGolang |    Core | Jpg/baseline/jpeg420exif.jpg | 1,602.222 us | 234.3564 us | 13.2416 us | 253.9063 |  533091 B |
  IdentifyPdfJs |    Core | Jpg/baseline/jpeg420exif.jpg | 1,597.570 us | 482.7360 us | 27.2755 us | 244.1406 |  512677 B |
```

antonfirsov:
Looks much better!
Gonna do a final review tomorrow.

JimBobSquarePants:
I'm gonna create a new issue once we get his merged regarding the scan decoder. 

I want to get it replaced eventually and will put my thoughts in the issue.

antonfirsov:
Everything is fine now, we should do the merge, and implement further improvements in new PR-s.

PR #552, opened 2018-04-29
Issue 551 solidbrush blend performance

woutware:
Please check thoroughly, I think I might have overlooked some conditions for skipping the blending in the FillProcessor.

Also I haven't done the FillRegionProcessor yet, I'm gonna need help with that.

The speedup with the image constructor with clear color param compared to the fill is pretty nice! My whole test draw is about 10% faster now overall.


src/ImageSharp.Drawing/Processing/Drawing/Processors/FillProcessor.cs
```diff
@@ -49,38 +50,57 @@ protected override void OnFrameApply(ImageFrame<TPixel> source, Rectangle source
...
-            using (BrushApplicator<TPixel> applicator = this.brush.CreateApplicator(
-                source,
-                sourceRectangle,
-                this.options))
-            {
-                amount.Span.Fill(this.options.BlendPercentage);
+            var solidBrush = this.brush as SolidBrush<TPixel>;
 
+            // If there's no reason for blending, then avoid it.
+            if (solidBrush != null && this.options.BlendPercentage == 1f && solidBrush.Color.ToVector4().W == 1f)
```
tocsoft: fast path should be activated when `solidBrush != null` and any of these things are true:

`this.options.BlenderMode == PixelBlenderMode.Normal && this.options.BlendPercentage == 1f && solidBrush.Color.ToVector4().W == 1f`
OR

`this.options.BlenderMode == PixelBlenderMode.Over && this.options.BlendPercentage == 1f && solidBrush.Color.ToVector4().W == 1f`

OR

`this.options.BlenderMode == PixelBlenderMode.Src`

tocsoft:
you will find its nigh on impossible to apply the same sort of fast path logic to the region filling, as we need to take into account anti-aliasing at region edges which in turn requires applying pixel blending.

src/ImageSharp/Image{TPixel}.cs
```diff
@@ -104,6 +139,11 @@ internal Image(Configuration configuration, ImageMetaData metadata, IEnumerable<
         /// </summary>
         public IImageFrameCollection<TPixel> Frames => this.frames;
 
+        /// <summary>
+        /// Gets the clear color to initialize the image frame pixels with.
+        /// </summary>
+        internal TPixel? ClearColor => this.clearColor;
```
antonfirsov: Not sure if I like this.

Does it worth to pollute our object model in this manner just to make it easier to add frames of a "default" color when producing gif-s manually? (0.01% of ImageSharp use-cases I guess).
@JimBobSquarePants @tocsoft thoughts?

src/ImageSharp.Drawing/Processing/Drawing/Processors/FillProcessor.cs
```diff
@@ -49,38 +50,61 @@ protected override void OnFrameApply(ImageFrame<TPixel> source, Rectangle source
...
-                source,
-                sourceRectangle,
-                this.options))
-            {
-                amount.Span.Fill(this.options.BlendPercentage);
+            var solidBrush = this.brush as SolidBrush<TPixel>;
 
+            // If there's no reason for blending, then avoid it.
+            if (solidBrush != null &&
+                (
```
antonfirsov: I refactor this logic into an `IsSolidBrushWithoutBlending` property or similar for better readability.

antonfirsov:
Opened #553 to deal with the coverage issues. Hope it can be merged quickly.

src/ImageSharp/Image{TPixel}.cs
JimBobSquarePants: I'm not a fan of nullable type properties on classes. I'm voting no.

woutware: The thought behind it is that if you create an image with a background color, all its frames will have automatically that background color, which could be reasonably expected from a user stand point. Without it you'd have the first frame with the background color, and if the user adds another frame with ImageFrameCollection.CreateFrame(), it will also have that background color.

What's the reason for not having nullable type properties on a class?

tocsoft: if we have this atall you would want to distinguish between default and not set because default(TPixel) would be == Transparent and you want to skip the code path if its not set at all.

JimBobSquarePants: Transparent isn’t the same as default. Rgb = 255.

tocsoft: oh yeah.. your right, ignore that in that case we can just skip the logic on `default(TPixel)` 👍

woutware: Agreed with the `default(TPixel)` part, so you suggest always initializing the image frame with the clearColor then instead of using `memoryManager.AllocateClean2D`?

I don't agree with the nullable reasoning, if it's properly documented and default values are not illogical I don't see a reason to refrain from using certain language features, if TPixel were a class it'd have been the same. In this case non-nullable works as well, but in general if nullable is better suitable for a particular situation I wouldn't handicap the design/api out of fear users are not capable enough of using it.

woutware: Alright, committed the change to non-nullable clear color.

src/ImageSharp/ImageFrame{TPixel}.cs
```diff
@@ -52,6 +53,39 @@ internal ImageFrame(MemoryManager memoryManager, int width, int height, ImageFra
...
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ImageFrame{TPixel}" /> class.
+        /// </summary>
+        /// <param name="configuration">The <see cref="Configuration"/> to use for buffer allocation and parallel options to clear the buffer with.</param>
+        /// <param name="width">The width of the image in pixels.</param>
+        /// <param name="height">The height of the image in pixels.</param>
+        /// <param name="clearColor">The color to clear the image with.</param>
+        internal ImageFrame(Configuration configuration, int width, int height, TPixel clearColor)
```
JimBobSquarePants: Could you please rename this to `backgroundColor`. Clear would be a verb.

src/ImageSharp/Image{TPixel}.cs
JimBobSquarePants: It wasn't a question of handicapping the API nor one of fear. Nullable properties have their place, just not in any API where you hold all the cards.

Take for example the new `ResponseHeaders` class from `Microsoft.Http.Abstractions`. The `ContentLength` property there is of type `long?` this makes perfect sense since the `HttpResponse` might not provide a content-length header.  

We have absolute control over that property and assigning it so Nullable would have been the incorrect option design wise.

On that note... I wonder whether this should be a property of the `ImageFrame<T>` itself?  That way we base any non-background color assigned frames on the root frame which would keep the property in line with the Width/Height properties.

src/ImageSharp/ImageFrame{TPixel}.cs
```diff
@@ -52,6 +53,39 @@ internal ImageFrame(MemoryManager memoryManager, int width, int height, ImageFra
...
+        internal ImageFrame(Configuration configuration, int width, int height, TPixel clearColor, ImageFrameMetaData metaData)
+        {
+            Guard.NotNull(configuration, nameof(configuration));
+            Guard.MustBeGreaterThan(width, 0, nameof(width));
+            Guard.MustBeGreaterThan(height, 0, nameof(height));
+            Guard.NotNull(metaData, nameof(metaData));
+
+            this.MemoryManager = configuration.MemoryManager;
+            this.PixelBuffer = this.MemoryManager.Allocate2D<TPixel>(width, height, false);
+            this.Clear(configuration.ParallelOptions, clearColor);
```
JimBobSquarePants: `this.PixelBuffer.Span.Fill()`?

src/ImageSharp/ImageFrame{TPixel}.cs
```diff
@@ -267,6 +301,23 @@ internal ImageFrame<TPixel2> CloneAs<TPixel2>()
             return target;
         }
 
+        /// <summary>
+        /// Clears the bitmap.
+        /// </summary>
+        /// <param name="parallelOptions">The parallel options.</param>
+        /// <param name="value">The value to initialize the bitmap with.</param>
+        public void Clear(ParallelOptions parallelOptions, TPixel value) {
```
JimBobSquarePants: Clearing is a process and belongs in the processing namespace; we don't want to clutter the Image class. We also never want to expose the ParallelOptions as a field. They are part of the Configuration Class and are globally set.

src/ImageSharp/Image{TPixel}.cs
```diff
@@ -37,6 +38,21 @@ public Image(Configuration configuration, int width, int height)
...
+        /// Initializes a new instance of the <see cref="Image{TPixel}"/> class
+        /// with the height and the width of the image.
+        /// </summary>
+        /// <param name="configuration">
+        /// The configuration providing initialization code which allows extending the library.
+        /// </param>
+        /// <param name="width">The width of the image in pixels.</param>
+        /// <param name="height">The height of the image in pixels.</param>
+        /// <param name="clearColor">The color to initialize the pixels with.</param>
+        public Image(Configuration configuration, int width, int height, TPixel clearColor)
```
JimBobSquarePants: `clearColor` => `backgroundColor`

src/ImageSharp/Image{TPixel}.cs
```diff
@@ -104,6 +139,11 @@ internal Image(Configuration configuration, ImageMetaData metadata, IEnumerable<
         /// </summary>
         public IImageFrameCollection<TPixel> Frames => this.frames;
 
+        /// <summary>
+        /// Gets the clear color to initialize the image frame pixels with.
+        /// </summary>
+        internal TPixel ClearColor => this.clearColor;
```
JimBobSquarePants: Move this to the `ImageFrame<TPixel>` and add to the constructor there. Use RootFrame as a default for subsequent frames

src/ImageSharp/Image{TPixel}.cs
```diff
@@ -22,6 +22,7 @@ public sealed class Image<TPixel> : IImage, IConfigurable
     {
         private readonly Configuration configuration;
         private readonly ImageFrameCollection<TPixel> frames;
+        private readonly TPixel clearColor;
```
JimBobSquarePants: No need for field. Assign straight to the property. (When you move all this to `ImageFrame`)

src/ImageSharp/Memory/BufferExtensions.cs
```diff
@@ -52,6 +52,18 @@ public static void Clear<T>(this IBuffer<T> buffer)
             buffer.Span.Clear();
         }
 
+        /// <summary>
+        /// Fills the contents of this buffer.
+        /// </summary>
+        /// <param name="buffer">The buffer</param>
+        /// <param name="value">The value to fill the buffer with.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void Fill<T>(this IBuffer<T> buffer, T value)
```
JimBobSquarePants: Does this have a caller? The above code works directly on the Span

src/ImageSharp/ImageFrame{TPixel}.cs
woutware: I left in the parallel clearing, on my laptop it's still 10%-20% faster and I can really use all the speed there is as the drawing bit is quite a bit slower at the moment. The user can always set the parallel options to use 1 thread if he insists.

src/ImageSharp/ImageFrame{TPixel}.cs
woutware: Changed method from public to internal. I'd just see this as initialization rather than processing, akin to how Allocate2D allows initialization, which isn't in processing either, but doesn't allow a default value.

src/ImageSharp/Image{TPixel}.cs
woutware: Done!

src/ImageSharp/Image{TPixel}.cs
woutware: Done, good point there.

src/ImageSharp/Image{TPixel}.cs
woutware: Done.

src/ImageSharp/Memory/BufferExtensions.cs
woutware: Removed again, used this in one of my intermediate experiments I think.

src/ImageSharp.Drawing/Processing/Drawing/Processors/FillProcessor.cs
woutware: Changed it, thanks for the help here!

src/ImageSharp/Image{TPixel}.cs
woutware: I've moved it to ImageFrame, it's indeed more consistent with the Width/Height properties.

antonfirsov:
From the Separation of Concerns point of view, I'm still not a big fan of introducing a `BackgroundColor` property on an imaging API. (It would be fine on a canvas API!)

But if you guys find it powerful and useful & vote to keep it, I would not resists.

One question:
- Do we still need `BackgroundColorProcessor`?
- If yes:
  - Shouln't we optimize it as well?
  - Shouldn't it change the `ImageFrame.BackgroundColor` for all frames?

woutware:
@antonfirsov I can't comment on the `BackgroundColorProcessor`, but remember that before this change the pixels were initialized with zeroes by AllocateClean2D. If you're 100% puristic, you'd be making a processor for this as well, which I think at this level is overkill. You can consider the one time initializion of the pixels as integral part of the allocation similarly to AllocateClean2D does it.

The change is a tremendous speedup: previously you'd always have 1) the initialization with zeroes, and 2) the slow Fill processor. Now you're skipping 2), and made 1) configurable with a default value. So you're basically twice as fast.

Some measured timings:

```
alloc + fill 00:00:00.0205244, alloc with background color 00:00:00.0093514

When not disposing the bitmap, so with a fresh alloc each time the difference becomes pretty much zero by the way, I'd have expected more difference here:
alloc + fill 00:00:00.0755986, alloc with background color 00:00:00.0741654
```

antonfirsov:
@woutware the constructor taking `backgroundColor` and the fill optimization are cool things, I have no issue with them! 

The purist part of me just wants to drop the background color parameter after initialization, and leave the post-constructor job of initializing new frames with the proper color to the user. Only very a small minority of our users do manual multi-frame gif manipulation!

woutware:
I've no preference either way, I'll be using single frame myself only, so I'm fine with whatever fits the current philosophy best. Having said that, if you create an Image with a background color, it would feel more logical to me that new image frames have that same background color automatically. But again it's pretty much equal to me either way.

antonfirsov:
It's @JimBobSquarePants's job to decide it from now. He is the Grand High Eternal Dictator! 😆 

(+ my question abut `BackgroundColorProcessor` still stands.)

JimBobSquarePants:
@antonfirsov `BackgroundColor` does operate on all frames.

https://github.com/SixLabors/ImageSharp/blob/01d04268d6cc93ea9936274d84397fa9b1b2b0d4/src/ImageSharp/Processing/Overlays/Processors/BackgroundColorProcessor.cs#L43

We need it because it performs a different action, doesn't require a brush (therefore can live in ImageSharp) and can operate only on the background layer (i.e you cannot overwrite original color data, only blend against it.)

It's super useful for padding out areas of non-transparent image formats following `ResizeMode.Pad`

src/ImageSharp/ImageFrame{TPixel}.cs
JimBobSquarePants: 👍

src/ImageSharp/ImageFrame{TPixel}.cs
JimBobSquarePants: 👍

src/ImageSharp/ImageFrame{TPixel}.cs
```diff
@@ -52,15 +51,47 @@ internal ImageFrame(MemoryManager memoryManager, int width, int height, ImageFra
...
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ImageFrame{TPixel}" /> class.
+        /// </summary>
+        /// <param name="configuration">The <see cref="Configuration"/> to use for buffer allocation and parallel options to clear the buffer with.</param>
+        /// <param name="width">The width of the image in pixels.</param>
+        /// <param name="height">The height of the image in pixels.</param>
+        /// <param name="backgroundColor">The color to clear the image with.</param>
+        internal ImageFrame(Configuration configuration, int width, int height, TPixel backgroundColor)
```
JimBobSquarePants: This is good! I actually wanted to introduce the configuration as a parameter and drop the `MemoryManager` overload so we're halfway there.

antonfirsov:
@JimBobSquarePants thanks, I understand now!

Shouldn't `BackgroundColorProcessor` alter `ImageFrame.BackgroundColor` (if we keep the property)?

JimBobSquarePants:
I think I get @antonfirsov 's issue with the `BackgroundColor` as a property. As soon as anyone writes to the image then it becomes invalid. 

@woutware I think we'll have to drop the convenience of the property and simply rely on setting the background color per frame. (This is seriously fine grained stuff anyway so I doubt it will be an issue.)

JimBobSquarePants:
We're nearly there with this, looking forward to seeing it merged! 👍

woutware:
Alright, removed the ImageFrame.Background property, and just using `default` if the user creates a new frame after the fact.

@JimBobSquarePants probably add some comments to the `BackgroundColorProcessor` code that you wrote here explaining the thoughts behind it, that'll be super useful for future users.

src/ImageSharp/ImageFrameCollection.cs
```diff
@@ -143,7 +143,7 @@ public Image<TPixel> CloneFrame(int index)
         /// <inheritdoc/>
         public ImageFrame<TPixel> CreateFrame()
```
JimBobSquarePants: Let's create an overload for this that accepts a color, then we cover all bases.

woutware: I see the method implements a member from interface IImageFrameCollection, wouldn't it be nice to change it to:
```
CreateFrame(TPixel backgroundColor = default)
```

JimBobSquarePants: I think we don’t actually need the interface. I can’t see why someone would need to implement it themselves.

JimBobSquarePants: Yeah, always choose overloads.

woutware:
Yay! Thanks guys!

JimBobSquarePants:
Thanks @woutware for taking the time to put this together. Much appreciated! 👍

maxvoxel8:
> you will find its nigh on impossible to apply the same sort of fast path logic to the region filling, as we need to take into account anti-aliasing at region edges which in turn requires applying pixel blending.

@tocsoft could the same logic be used if antialiasing is off?

Issue #518, opened 2018-04-10
Bad RST marker


mabead:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Loading image throws exception.

### Steps to Reproduce
With 1.0.0-beta0003, calling Image.Load on the attached file throws the following exception:

```
SixLabors.ImageSharp.ImageFormatException
  HResult=0x80131500
  Message=Bad RST marker
  Source=SixLabors.ImageSharp
```

[25.zip](https://github.com/SixLabors/ImageSharp/files/1892447/25.zip)

### System Configuration

- ImageSharp version: 1.0.0-beta0003
- Environment: Windows
- .NET Framework version: dotnet core 2.0.5

JimBobSquarePants:
Fixed with #571

softsprocket:
Do you have an idea when the nuget will be updated? Thanks.

JimBobSquarePants:
Last week

softsprocket:
Ah right. I had to check 'Include prerelease' to see the update. Sorry (and thanks).

JimBobSquarePants:
No worries, hopefully not too long before we can do a release proper.

PR #571, opened 2018-05-11
Swap jpeg decoders

### LINKED ISSUE #517: Reached end of stream before proceeding EOI marker!
### Prerequisites

### Description
Calling Image.Load results in exception.

### Steps to Reproduce

With 1.0.0-beta0003, calling Image.Load on the attached file throws the following exception:

```
SixLabors.ImageSharp.Formats.Jpeg.GolangPort.Components.Decoder.EOFException
  HResult=0x80131500
  Message=Reached end of stream before proceeding EOI marker!
  Source=SixLabors.ImageSharp
```

[REPRO copy.zip](https://github.com/SixLabors/ImageSharp/files/1892442/REPRO.copy.zip)

### System Configuration

- ImageSharp version: 1.0.0-beta0003
- Environment: Windows
- .NET Framework version: dotnet core 2.0.5

JimBobSquarePants:
### Prerequisites

### Description

This PR switches out the old Golang based Jpeg decoder with the newer PdfJs decoder.

This gives us a robust decoder that produces 100% accurate spectral output. Performance matches the Golang decoder with less memory usage. There is excellent scope for further improving performance also.

We'll keep both decoders in the codebase for now as there might be ideas we can still pinch from the Golang decoder. 

Fixes #517 and #518

I've also expanded the unit tests to better cover both decoders and reorganized the internal namespaces to clean up the code.


tests/ImageSharp.Tests/Formats/Jpg/JpegImagePostProcessorTests.cs
```diff
@@ -71,15 +72,70 @@ public void DoProcessorStep<TPixel>(TestImageProvider<TPixel> provider)
             }
         }
 
+        [Theory]
+        [WithFile(TestImages.Jpeg.Baseline.Calliphora, PixelTypes.Rgba32)]
+        [WithFile(TestImages.Jpeg.Baseline.Testorig420, PixelTypes.Rgba32)]
+        public void DoProcessorStepPdfJs<TPixel>(TestImageProvider<TPixel> provider)
```
antonfirsov: `VerifySpectralCorrectness_PdfJs/Golang` should test the first step, here we only need to test `JpegImagePostProcessor`, so parsing input with both decoders doesn't add value here. 
It's totally fine to run `JpegImagePostProcessorTests` with the default jpeg decoder only, which is `PdfJsDecoderCore` now.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
```diff
@@ -80,9 +80,9 @@ internal partial struct Block8x8F
             set => this[(y * 8) + x] = value;
         }
 
-        public static Block8x8F operator *(Block8x8F block, float value)
+        public static Components.Block8x8F operator *(Components.Block8x8F block, float value)
```
antonfirsov: I'm getting a "Qualifier is redundant" tip here.
I wonder which analyzer added the namespace qualifier?

JimBobSquarePants: It was Resharper, probably due to the struct being a partial.

PR #579, opened 2018-05-18
Delete SpanHelper.

JimBobSquarePants:
### Prerequisites

### Description
No need for SpanHelper now so let's delete it.


antonfirsov:
I don't think AppVeyor the test failures are related to the changes in this PR. Created #580 to check it in a quick and dirty way.

antonfirsov:
I think the issue has been introduced by using the newest "VS 2017 Preview" AppVeyor image.

I think we can *workaround* (😞)  this by [using that image only for 2.1](https://www.appveyor.com/docs/build-environment/#using-multiple-images-for-the-same-build).

PR #580, opened 2018-05-18
Quick & dirty empty test PR to trigger AppVeyor PR build

antonfirsov:
Just a quick test, something may have went wrong on AppVeyor. See #579.


antonfirsov:
Quirky behavior proven.

Issue #562, opened 2018-05-08
JpgExport -> System.ArgumentOutOfRangeException: Index and count must refer to a location within the buffer. Parameter name: bytes


alex-birch:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Similar to [this](https://github.com/SixLabors/ImageSharp/issues/559) issue I raised yesterday but with a different stack trace (thanks for looking at it so quickly!). I'm using ImageSharp to open all these images from [mscoco](http://cocodataset.org/) - there are thousands of images and it's doing well on most of them.

Loading and export this image as a Jpg ByteArray causes the following stack trace (again, PNG works okay):

### Steps to Reproduce
![coco_val2014_000000260238](https://user-images.githubusercontent.com/4734530/39756400-cfdb786a-530c-11e8-9fb8-296a7f2429da.jpg)

`Unhandled Exception: System.ArgumentOutOfRangeException: Index and count must refer to a location within the buffer.
Parameter name: bytes
   at System.Text.UnicodeEncoding.GetString(Byte[] bytes, Int32 index, Int32 count)
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccDataReader.ReadTextDescriptionTagDataEntry()
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccDataReader.ReadTagDataEntry(IccTagTableEntry info)
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccReader.ReadTagData(IccDataReader reader)
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccProfile.InitializeEntries()
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccWriter.Write(IccProfile profile)
   at SixLabors.ImageSharp.Formats.Jpeg.GolangPort.JpegEncoderCore.WriteIccProfile(IccProfile iccProfile)
   at SixLabors.ImageSharp.Formats.Jpeg.GolangPort.JpegEncoderCore.Encode[TPixel](Image`1 image, Stream stream)
[...]

Issue #559, opened 2018-05-07
JpgExport -> System.OverflowException: Array dimensions exceeded supported range


alex-birch:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported
Closest I could find was [this](https://github.com/SixLabors/ImageSharp/issues/217)

### Description

Loading and export this image as a Jpg ByteArray causes the following stack trace (PNG works okay):

`Unhandled Exception: System.OverflowException: Array dimensions exceeded supported range.
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccReader.ReadTagTable(IccDataReader reader)
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccReader.ReadTagData(IccDataReader reader)
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccProfile.InitializeEntries()
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccWriter.Write(IccProfile profile)
   at SixLabors.ImageSharp.Formats.Jpeg.GolangPort.JpegEncoderCore.WriteIccProfile(IccProfile iccProfile)
   at SixLabors.ImageSharp.Formats.Jpeg.GolangPort.JpegEncoderCore.Encode[TPixel](Image`1 image, Stream stream)
   at WebApplicationBasic.Program.Main(String[] args) in Program.cs:line 18`

### Steps to Reproduce

![coco_val2014_000000395978](https://user-images.githubusercontent.com/4734530/39688369-bbadfefa-5215-11e8-8c08-e6fa7f16afa5.jpg)

I loaded the image and exported to JPG like so:
`            var img = SixLabors.ImageSharp.Image.Load<Rgba32>(args[0]);
            img.ToByteArray(new JpegEncoder());`

### System Configuration

- ImageSharp version:
SixLabors.ImageSharp 1.0.0-beta0003
- Other ImageSharp packages and versions:
- Environment (Operating system, version and so on):
MacOS High Sierra
- .NET Framework version:
Dotnet 2.1.3
- Additional information:

JBildstein:
[...]

PR #588, opened 2018-05-22
ICC Improvements

### LINKED ISSUE #578: System.ArgumentOutOfRangeException for Parameter name: start loading a jpg
### Prerequisites

### Description

When i try to load the attached image, i get the following error

```
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: start
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument)
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccDataReader.ReadUInt32()
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccReader.ReadTagTable(IccDataReader reader)
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccReader.ReadTagData(IccDataReader reader)
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccProfile.InitializeEntries()
   at SixLabors.ImageSharp.MetaData.Profiles.Icc.IccWriter.Write(IccProfile profile)
   at SixLabors.ImageSharp.Formats.Jpeg.GolangPort.JpegEncoderCore.WriteIccProfile(IccProfile iccProfile)
   at SixLabors.ImageSharp.Formats.Jpeg.GolangPort.JpegEncoderCore.Encode[TPixel](Image`1 image, Stream stream)
   at SixLabors.ImageSharp.ImageExtensions.Save[TPixel](Image`1 source, String filePath, IImageEncoder encoder)
   at SixLabors.ImageSharp.ImageExtensions.Save[TPixel](Image`1 source, String filePath) 
```

### Steps to Reproduce

using (Image<Rgba32> image = SixLabors.ImageSharp.Image.Load(fileName))
{
    // note: issue does not happen during Load (e.g. decoding), but rather in the Encoding part when trying to save it
    image.Save("bar.jpg"); // Automatic encoder selected based on extension.
}

with this image:

wrong_icc_profile_size

### System Configuration

- ImageSharp version: 1.0.0-beta0003
- Environment (Operating system, version and so on): windows 10
- .NET Framework version: netcoreapp2.0
- Additional information:

I have tried to check the image with imagemagick identify and JPEGsnoop, both can decode image.

What i have noticed are some weird ICC Profile Sizes, for example at the first Marker: APP2:
The ICC Profile size is stated to be 594832749 bytes, which seems to be wrong.

```
*** Marker: APP2 (xFFE2) ***
  OFFSET: 0x00000012
  Length          = 32016
  Identifier      = [ICC_PROFILE]
    ICC Profile:
      Marker Number = 1 of 1
        Profile Size                      : 594832749 bytes
        Preferred CMM Type                : 'ﾱﾼ(ﾲ' (0xB1BC28B2)
        Profile Version                   : 6.13.0.11 (0x6D0BA39C)
        Profile Device/Class              : ? (0x2D606CB4) ('-`lﾴ' (0x2D606CB4))
        Data Colour Space                 : ? (0x96F23188) ('ﾖ￲1ﾈ' (0x96F23188))
        Profile connection space (PCS)    : 'lgﾋﾩ' (0x6C678BA9)
        Profile creation date             : 13617-27684-33198 14436:27625:30956
        Profile file signature            : ';(kﾷ' (0x3B286BB7)
        Primary platform                  : ? (0x714F3D87) ('qO=ﾇ' (0x714F3D87))
        Profile flags                     : 0x6B8C6AC3
        Profile flags                       > Embedded profile
        Profile flags                       > Profile can be used independently of embedded
        Device Manufacturer               : '?ﾇkh' (0x3F876B68)
        Device Model                      : 'e3A0' (0x65334130)
        Device attributes                 : 0x6B4A608C_428C6B32
        Device attributes                   > Reflective
        Device attributes                   > Matte
        Device attributes                   > Media polarity = negative
        Device attributes                   > Black & white media
        Rendering intent                  : 0x5CB843A2
        Profile creator                   : 'kUh' (0x6B055568)
        Profile ID                        : 0x458D6AFE_541545DA_6AF9532A_46166AF5
```

JBildstein:
### Prerequisites

### Description
This PR improves the handling of ICC profiles, specifically:
 - fixes profile ID calculation
 - add check for the validity of a profile and use it in the jpeg decoders
 - change the ToByteArray method to use the underlying buffer instead of parsing and then writing the profile
 - make parsing of profile data more defensive to guard against corrupt data

That last point is not extensive and could need some more work. Currently only the tag table is checked for invalid values but it would make sense for many values to check them during parsing as well.
For example, often a number of array items or a position within the profile is read. I'm not sure how to best handle it yet and may need some medium changes and a separate PR.
Until the profiles are actually used for color management, it's not really relevant though. The issues that surfaced with corrupt profiles only occured because of the way ToByteArray was implemented.

This PR fixes #578, #559 and #562


src/ImageSharp/Formats/Jpeg/PdfJsPort/PdfJsJpegDecoderCore.cs
```diff
@@ -345,6 +345,11 @@ public void ParseStream(Stream stream, bool metadataOnly = false)
                 // Read on.
                 fileMarker = FindNextFileMarker(this.markerBuffer, this.InputStream);
             }
+
+            if (this.MetaData.IccProfile?.CheckIsValid() == false)
```
JimBobSquarePants: Can we move this to `AssignResolution` and rename that to `InitDerivedMetaDataProperties`, The `return` above prevents us reading more than we need to when we are identifying images.

src/ImageSharp/Formats/Jpeg/GolangPort/GolangJpegDecoderCore.cs
```diff
@@ -413,6 +413,11 @@ public void ParseStream(Stream stream, bool metadataOnly = false)
             }
 
             this.InitDerivedMetaDataProperties();
+
+            if (this.MetaData.IccProfile?.CheckIsValid() == false)
```
JimBobSquarePants: Can we move this to `InitDerivedMetaDataProperties`

src/ImageSharp/Formats/Jpeg/PdfJsPort/PdfJsJpegDecoderCore.cs
JBildstein: I initially wanted to put it there, but it's only called from Decode and Identify and not in ParseStream. In the go port it's called at the end of the ParseStream method (and by extension in Decode and Identify).
Should we move `AssignResolution` there as well? I changed the `return` you mentioned to a `break` so things after the `while` loop will be executed.

JimBobSquarePants: Both `Decode` and `Identify` call `ParseStream` in both decoders. We want to return at that point in PdfJs decoder to avoid reading the entire SOS segment when we don’t need to. So using break slows us down. 

I can move the method if you like? Save you the bother.

JBildstein: yes they do, but `ParseStream` is public so it can be called directly, that's the only reason I wanted to do it there.
Sorry about the return thing, I thought break behaves differently at that point. We could use a 
`goto` to jump out of the `while` (it's evil, I know :P)

and no worries, it's a small change, I can do it.

JimBobSquarePants: 👍 It's only public as we use it for tests. The class is internal so not available to consumers.

JBildstein: ok great, then there's no reason to change anything in that regard. Will push changes in a couple minutes.

PR #589, opened 2018-05-23
Fix running tests locally in 32 bit mode

dlemstra:
### Prerequisites

### Description
This PR fixes locally running the unit tests in 32 bit mode. This is done by doing the following:
- Added extra files to the 32 bit exclude list in the JPEG unit tests.
- Disabled caching of images in the `FileProvider` to lower memory usage.

And this PR also includes a fix of disposing the objects in the finally block of the `WuFrameQuantizer`.


tests/ImageSharp.Tests/TestUtilities/ImageProviders/FileProvider.cs
```diff
@@ -133,15 +133,14 @@ public override Image<TPixel> GetImage(IImageDecoder decoder)
             {
                 Guard.NotNull(decoder, nameof(decoder));
 
-                Key key = new Key(this.PixelType, this.FilePath, decoder);
+                if (!TestEnvironment.Is64BitProcess)
```
antonfirsov: Can't we keep caching smaller images on 32bit?
(By providing an upper megapixel limit.)

JimBobSquarePants: It doesn't seem to affect our testing timings much so I'm inclined to keep it simple for now.

PR #604, opened 2018-06-05
Cross target NETCOREAPP2.1

iamcarbon:
### Prerequisites

### Description

This eliminates the reference to System.Memory and allows us to use the new apis in NETCOREAPP2.1.


iamcarbon:
@antonfirsov We're still using dotnet-sdk-2.1.300-rc1-008673 for the Travis build. Can we update this to the final release?

brianpopow:
@iamcarbon yes you can, i have tried it in a different branch already

antonfirsov:
> [..]  allows us to use the new apis in NETCOREAPP2.1.

@iamcarbon we need to pollute our code with `#ifdef`-s if we want to go this way. I think it's not a good strategy from the maintenance point of view. 

The PR also introduces a maintenance cost of keeping an extra target compiled in the core library (slower compilations locally + on CI).

Questions:
- Are there any .NET Core 2.1 API-s available at the moment we **really** want to use because of the benefits they bring?
- Is there any killer benefit from users POV if we target .NET Core 2.1 directly?

iamcarbon:
Being able to read and write spans directly to and from streams is a big improvement.

But I agree, the ifdefs get gross.

JimBobSquarePants:
If we believe that the Span API's will provide us with enough performance improvements in the other codecs I'm happy for this to go in. Eventually we would be able to remove the definitions as the API's eventually becomes standard which I am sure they will.

We do have to be careful though. A scatter of ifdefs in the codecs are ok, elsewhere though I think is a definite no-no.

JimBobSquarePants:
@antonfirsov Agree with both points. 👍

iamcarbon:
Added extension methods to span to reduce the #ifdef statements along with a few more examples of where they can be used to eliminate allocations.

Ready for review!

antonfirsov:
@iamcarbon well done thanks!
Can you validate if `dotnet xunit -f netcoreapp2.1` covers the new `#if`-ed code paths? Any appveyor configuration suggestions if it doesn't?

src/ImageSharp/Formats/Bmp/BmpInfoHeader.cs
```diff
@@ -132,6 +132,11 @@ internal struct BmpInfoHeader
         /// <seealso href="https://msdn.microsoft.com/en-us/library/windows/desktop/dd183376.aspx"/>
         public static BmpInfoHeader Parse(ReadOnlySpan<byte> data)
         {
+            if (data.Length != Size)
+            {
+                throw new ArgumentException(nameof(data), $"Must be 40 bytes. Was {data.Length} bytes.");
```
JimBobSquarePants: Interpolate the constant `Size` here.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -736,16 +737,16 @@ private void WriteProfiles<TPixel>(Image<TPixel> image)
         private void WriteStartOfFrame(int width, int height, int componentCount)
         {
             // "default" to 4:2:0
-            byte[] subsamples = { 0x22, 0x11, 0x11 };
+            Span<byte> subsamples = stackalloc byte[] { 0x22, 0x11, 0x11 };
             byte[] chroma = { 0x00, 0x01, 0x01 };
```
JimBobSquarePants: Why do we `stackalloc` one but not the other?

antonfirsov:
Can't remember for sure, but [this line](https://github.com/SixLabors/ImageSharp/blob/master/run-tests.ps1#L75) is probably a workaround for a similar issue.

iamcarbon:
No clue why downgrading the Unsafe package fixed the build. Also, not sure if a fix is imminent with the release of the 4.5.1 packages. Going to look into this more on Monday.

antonfirsov:
Pushed a few changes, because I'm quite OCD-ish about really having the coverage we have. @JimBobSquarePants hope you're fine with them.

I'd say let's merge this, I'm curious about the new NuGet package being generated by our nightlies!

However we should expect super long and boring test execution from now. It might be worth to disable some of the `target_framework`-s  in `appveyor.yml` (which?) or ask Santa Claus to rent us a test machine up in the clouds. ☁️ 🌨

JimBobSquarePants:
Looks good though me, hopefully we can beef up our CI server soon. I’ll chase up our DNF application.

Issue #565, opened 2018-05-09
System.Memory API integration


antonfirsov:
### Goals
Thanks to #558, we are now able to expose `System.Memory` API-s allowing efficient interop operations. However, we need to distinguish between different use-cases, because they raise design problems of different difficulty levels. In some use-cases we have to decide whether we should expose `Span<T>`, `Memory<T>` or both.

### "Official" guideline on System.Memory
Managed to find a WIP (?) documentation on `Memory<T>`. (It was mentioned in [coreclr/#17340](https://github.com/dotnet/coreclr/pull/17340))
- [Memory<T> API documentation and samples](https://gist.github.com/GrabYourPitchforks/8efb15abbd90bc5b128f64981766e834)
- [Memory<T> usage guidelines](https://gist.github.com/GrabYourPitchforks/4c3e1935fd4d9fa2831dbfcab35dffc6)

### Tasks
- [x] Basic `ReadOnlySpan<T>` overloads of [LoadPixelData](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Image.LoadPixelData.cs) (done in #567)
- [x] `ReadOnlySpan<T>` overloads for [`Image.Load()` and `Image.DetectFormat()`](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Image.FromBytes.cs#L44). See @jherby2k's [proposal](https://github.com/SixLabors/ImageSharp/issues/565#issuecomment-388057174).
  - We can pin the span and use [`UnmanagedMemoryStream`](https://msdn.microsoft.com/en-us/library/system.io.unmanagedmemorystream%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) for this.
- [x] Integrate `System.Memory.IMemoryOwner<T>` and/or `System.Memory.MemoryManager<T>` with [`SL.ImageSharp.Memory.IBuffer<T>`](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Memory/IBuffer%7BT%7D.cs)
  	- Prerequisite for using `Memory<T>`, thus for the following 2 points
  	- ~Notice the name collision with `SL.ImageSharp.Memory.MemoryManager<T>` :P~ **UPDATE:** SixLabors stuff is now called `MemoryAllocator`
- [x] `Image.Wrap(existingMemoryArea)` or similar.
  - Very useful for interop scenarios, see #435 or #88
[...]

PR #607, opened 2018-06-10
Introduce basic Memory<T> API-s

antonfirsov:
#### Prerequisites

### Description
This PR introduces some memory API-s specified in #565. I think I managed to understand and follow the [ownership management rules](https://gist.github.com/GrabYourPitchforks/4c3e1935fd4d9fa2831dbfcab35dffc6) of `Memory<T>`. beta5 was originally intended to be a hotfix release, but I think it's worth to add these API-s.

#### Features
- `AdvancedImageExtensions`: 
    - `GetPixelMemory()` / `GetPixelSpan()` for both `Image<T>` and `ImageFrame<T>`
    - `GetPixelRowMemory()` / `GetPixelRowSpan()` for both `Image<T>` and `ImageFrame<T>`
- `Image.WrapMemory(Memory<T>)` for wrapping an existing external memory area (eg. contents of a `System.Drawing.BitmapData`)
    - This is kept internal for now. We need to validate it by testing it against an actual interop scenario (eg. `System.Drawing`)

#### Internal Refactors
- Adapting `System.Buffers.MemoryManager<T>` (formerly known as `OwnedMemory<T>`) by `IBuffer<T>` implementations
- Refactor `ShapeRegion` to allow dropping `AllocateFake`

#### Breaking Changes
- Renamed `MemoryManager` to `MemoryAllocator` to avoid name conflicts with `System.Buffers.MemoryManager<T>` 
- Memory management primitives should be moved to **SixLabors.Core**, which means that they should be defined under the `SixLabors.Memory` namespace instead of `SixLabors.ImageSharp.Memory`. In order to concentrate the breaking changes into a single release, I decided to do the namespace move in this PR.
- `DangerousGetPinnableReferenceToPixelBuffer()` is marked obsolete. `MemoryMarshal.GetReference(source.GetPixelSpan())` should be used instead


src/ImageSharp/PixelAccessor{TPixel}.cs
```diff
@@ -5,8 +5,8 @@
 using System.Diagnostics;
```
JimBobSquarePants: When do we get to ditch this class? I don't think it's used much now.

antonfirsov:
@tocsoft after thinking a lot it seems still most reasonable to manage the ownership move/swap logic in `Buffer2D<TPixel>`. This way all `IBuffer<T>`-s are returned to their own allocators/pools on `.Dispose()`.

The logic now is implemented by `Buffer2D<TPixel>.SwapOrCopyContent()` with the following rules:
- We can always swap the contents of two `Buffer2D<T>` instances if their inner buffers both *own* their memory (regardless of their original `MemoryAllocator`-s).
- If one of the inner buffers is only *consumed*, and the sizes are same, we can copy the contents to the destination buffer
- If one of the inner buffers is only *consumed*, and the sizes are different, the operation is invalid.

For me it seems we are good with these rules, added a plenty of tests to ensure they work.

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
```diff
@@ -24,78 +24,134 @@ public static Configuration GetConfiguration<TPixel>(this Image<TPixel> source)
...
-        /// <param name="source">The source image frame</param>
-        /// <returns>A pinnable reference the first root of the pixel buffer.</returns>
-        public static ref TPixel DangerousGetPinnableReferenceToPixelBuffer<TPixel>(this ImageFrame<TPixel> source)
+        /// <param name="source">The source <see cref="ImageFrame{TPixel}"/></param>
+        /// <returns>The <see cref="Memory{T}"/></returns>
+        public static Memory<TPixel> GetPixelMemory<TPixel>(this ImageFrame<TPixel> source)
             where TPixel : struct, IPixel<TPixel>
-         => ref DangerousGetPinnableReferenceToPixelBuffer((IPixelSource<TPixel>)source);
+        {
+            return source.PixelBuffer.Buffer.Memory;
```
tocsoft: we shouldn't expose the buffer as a Memory<T> as we then can't swap the buffer and return the memory to the Array pool other wise someone could have a reference to a piece of memory that has been returned to someone else or was owned by an external process and it could leak dangerous things

tocsoft: The image itself should be treated as Memory<T> of sorts but can never share the hand over directly indefinitely.

antonfirsov: @tocsoft the same could be true for `Span<T>`:

```C#
Span<int> testSpan = default;

using (IBuffer<T> buffer = allocator.Allocate<int>(10))
{
    // span is assigned, we are on the same method stack in the using block:
    testSpan = buffer.GetSpan(); 
}

using (IBuffer<T> somethingElse= allocator.Allocate<int>(10))
{
     somethingElse.GetSpan()[5] = 666;
}

// we are out of the intended scope here:
int naiveUsersValue = testSpan[5];
```

Despite the dangers I think both `image.GetPixelMemory()` and `image.GetPixelRowMemory(y)` are useful methods for advanced users, because they allow heaping pixel rows, so users can use them in async API-s etc.

What we can do is:
1. Document these pitfals
2. Introduce a new level of indirection: Decouple `IBuffer<T>` from `MemoryManager<T>`, so we can swap the buffer behind the `Memory<T>` instance.

Implementing point 2. would take many hours, delaying the open-up of memory API-s further. We can actually do it later, as an additional increment.

tocsoft: We have no one actively after Memory<T> when Span<T> as having a reference to the image is there. I vote we remove it from the API until we can expose it safely or at least its requested. Until then its an API we have to support which very easily (unlike span which is much harder and can only leak within a single stack) can leak the contents of memory at any random point in the future of the whole application.

I think its too much risk with near to zero benefits that can be handled by keeping a reference to the Image.

antonfirsov: @tocsoft you can't do the following with span:
```C#
async Task DoOperationOnPixelsAsync(Span<Rgba32> pixels)
{
    await .....;
    await ........;
}
```

One of our users (on gitter) walked into this pitfall resulting in a runtime failure, and he blamed ImageSharp for it. (This might be fine, we can't expect everyone to understand stack-only types.) We decided to hide the `Span<T>` API-s because of this story. There was a quite valid reasoning, that we should expose only the `Memory<T>` variant.

I say we should expose either both the Span and the Memory API-s or neither of them. If we can't agree, I say let's internalize everything for now, turning this PR into a refactor one.

@JimBobSquarePants your thoughts?

antonfirsov: Actually ... you don't even need to have a goal of implementing something in a separate async operation for let's say perf. If you are implementing an ASP.NET controller, being in an async context most of the time, it's very easy to walk into a pitfall of this kind.

tocsoft: 1.  `Span<T>` is now hidden away in the advanced namespace so we are much less likely to get people confused about using Span with async etc. 
2. you won't be doing pixel level operations over spans in a controller in MVC.. they will be advised to create a Processor and do there stuff there (MVC sorted)
3. if you changes `Span<T>`  in `DoOperationOnPixelsAsync` with `Image<T>` you can now work async and only have a manor (aggressively in lines called over to get the `Span<T>` as required)

If you are using the API's that expose `Span<T>` we are deeming the user knowledgeable enough about the framework to handle/understand how async and Spans don't mix well.

If we don't expose `Span<T>` then users can't write new `IImageProcessor`s they can do that without `Memory<T>` as that's exactly what we are doing internally.

antonfirsov: @mellinoe sorry for pulling you in, but I think your opinion might be very useful here!
@iamcarbon - same for you :)

TLDR:
- Exposing only `Span<TPixel>` API-s is dangerous, because it may lead to heap captures for users who don't follow (or don't know about) coding patterns we suggest.
- Exposing `Memory<TPixel>` API-s is dangerous, because the memory may point to an "outdated" location after certain processors being executed.

antonfirsov: @tocsoft if no one disagrees, I'm being tempted to accept your reasoning.

Exposing `Memory<T>` could be an incremental addition later.

antonfirsov:
@tocsoft made `GetPixelMemory()` and `GetPixelRowMemory(y)` internal for now. Gonna do the `IBuffer<T>`  :left_right_arrow: `MemoryManager<T>` separation in a separate PR (very soon).

Gonna merge this, if you are fine with the current status.

tocsoft:
I can't see anything I'm not happy with now. 👍

PR #613, opened 2018-06-12
Add 16 bit png support

### LINKED ISSUE #610: Feature Request: Support for 16 Bit Greyscale PNG's
### Prerequisites

### Description

Decoding 16-Bit Greyscale PNG's currently is not possible with imagesharp. This would be nice to have.

### Steps to Reproduce

Load an 16 Bit greyscale PNG, like the attached image:
bit-depth-not-supported

### System Configuration

- ImageSharp version: beta4
- Environment (Operating system, version and so on): windows 10
- .NET Framework version:  does not matter
- Additional information:

image properties:
```
Properties:
    date:create: 2018-06-11T14:00:27+02:00
    date:modify: 2018-06-11T14:00:27+02:00
    png:IHDR.bit-depth-orig: 16
    png:IHDR.bit_depth: 16
    png:IHDR.color-type-orig: 0
    png:IHDR.color_type: 0 (Grayscale)
    png:IHDR.interlace_method: 0 (Not interlaced)
    png:IHDR.width,height: 772, 579
```

JimBobSquarePants:
### Prerequisites

### Description
This WIP PR will add proper 16 bit png support to the library. Fixes #610 

Currently we support decoding from 16 Rgb and Rgba pngs but we cheat while doing it by removing the least significant bits and preserving only 8 bits of data. To do this properly we need to be able to use `Rgba64` to losslessly read and write the pixel data. 

My plan is as follows:
- [x] Refactor `Rgba64` to use `StructLayout.Sequential` and add `R,G,B,A` fields.
- [x] Add To/From `Rgba64` methods to `IPixel`
- [x] Use the new methods to read/write the pixel data. Conversion will be automatic to other formats via scaling.
- [x] Add tests

Additionally I would like to:
- [x] Create a new `Rgb48` pixel format

Potentially we could create new pixel formats for greyscale and greyscale + alpha  data but that could happen post launch if the workload becomes too much.


JimBobSquarePants:
@antonfirsov @tocsoft @dlemstra and anyone else reading this.

I've got the fundamentals in place for reading 16 bit png's but need some help on the writing side. 

I need to figure out a way to set both the bit depth and the color type when encoding without complex sanitation to ensure that the values are correct. 

e.g A 2 color palette image is never going to be a 16 bit image. 

I'm stuck figuring it out.

JimBobSquarePants:
Failing tests are due to me changing png encoder grayscale conversion algorithm to ITU-R recommendation 709  matching the libpng implementation.

Will reference images tomorrow and add specific bitdepth reference images plus tests.

JimBobSquarePants:
@antonfirsov Yup, as long as the image pixel format is of high enough resolution then the accuracy is maintained. We’ll have `Rgb48` and `Rgba64` now to make that easier.

JimBobSquarePants:
This is ready to review now. 

Sorry about the size of this chaps but it turned out adding two new pixel formats with associated operations was a big job touching everywhere.

I had a hell of a time with image comparison as I had to switch to comparing using `Rgba64` and I had to get rid of the System.Drawing comparer for pngs as it was losing accuracy on decode. 

Anyway, it works, we now have full 16 bit support.

tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
```diff
@@ -20,74 +20,94 @@ public class PngDecoderTests
     {
         private const PixelTypes PixelTypes = Tests.PixelTypes.Rgba32 | Tests.PixelTypes.RgbaVector | Tests.PixelTypes.Argb32;
 
+        // This should be exact but for some reason it fails in some build environments.
+        private static readonly ImageComparer ValidatorComparer = ImageComparer.TolerantPercentage(0.0001F, 26);
```
antonfirsov: Is it travis? Can't we make it environment specific?

tests/ImageSharp.Tests/TestUtilities/TestEnvironment.Formats.cs
```diff
@@ -60,11 +60,10 @@ private static Configuration CreateDefaultConfiguration()
...
             {
-                configuration.ConfigureCodecs(
-                    ImageFormats.Png,
-                    SystemDrawingReferenceDecoder.Instance,
-                    SystemDrawingReferenceEncoder.Png,
-                    new PngImageFormatDetector());
+                // System.Drawing on Windows can decode 48bit and 64bit pngs but
+                // it doesn't preserve the accuracy we require for comparison.
+                // This makes CompareToOriginal method non-useful.
+                configuration.Configure(new PngConfigurationModule());
```
antonfirsov: We should find a different testing method for those bit depths then, otherwise we are testing `PngDecoder` against itself in all our test cases, loosing the point of the regression testing.

tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
```diff
@@ -142,20 +162,66 @@ public void Decode_Interlaced_ImageIsCorrect<TPixel>(TestImageProvider<TPixel> p
...
             {
-                image.DebugSave(provider);
+                var encoder = new PngEncoder { ColorType = PngColorType.Rgb, BitDepth = PngBitDepth.Bit16 };
 
-                // Workaround a bug in mono-s System.Drawing PNG decoder. It can't deal with 48Bpp png-s :(
-                if (!TestEnvironment.IsLinux && !TestEnvironment.IsMono)
+                if (!SkipVerification(provider))
                 {
-                    image.CompareToOriginal(provider, ImageComparer.Exact);
+                    image.VerifyEncoder(provider, "png", null, encoder, customComparer: ImageComparer.Exact);
```
antonfirsov: For me it looks like you are testing the decoders result against itself, running it one another time inside `VerifyEncoder()`. Or am I missing something?

tests/ImageSharp.Tests/TestUtilities/ImageComparison/TolerantImageComparer.cs
```diff
@@ -105,12 +106,12 @@ public TolerantImageComparer(float imageThreshold, int perPixelManhattanThreshol
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static int GetManhattanDistanceInRgbaSpace(ref Rgba32 a, ref Rgba32 b)
+        private static int GetManhattanDistanceInRgbaSpace(ref Rgba64 a, ref Rgba64 b)
```
antonfirsov: 👍 seems good, but it might be worth to have another look tomorrow. All our tests depend on this stuff 😄

antonfirsov:
We can even skip the CRC point in this PR, adding an issue for it.

The point is that if our 64bpp/48bpp output is **most likely** fine, and we are sure we do not break existing code/test logic, that is an increment compared to what we have in the main branch.

tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
JimBobSquarePants: Seems to be both CI environments.

tests/ImageSharp.Tests/TestUtilities/ImageComparison/TolerantImageComparer.cs
JimBobSquarePants: Defo would like a second opinion on this. I had to change some tolerances, I'm assuming code round actually hid differences before.

dlemstra:
@antonfirsov You can probably do this with Magick.NET (Q16) by loading an image and getting the pixels of that image.

tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
JimBobSquarePants: Yeah, this is useless, but it was the only way I could do any testing just now by round tripping. We need a better reference encoder for png

tests/ImageSharp.Tests/TestUtilities/TestEnvironment.Formats.cs
JimBobSquarePants: Yup, very much agree.

src/ImageSharp/PixelFormats/Rgb48.cs
```diff
@@ -0,0 +1,257 @@
...
+            dest.G = (byte)MathF.Round(vector.Y);
+            dest.B = (byte)MathF.Round(vector.Z);
+        }
+
+        /// <inheritdoc />
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void ToRgba32(ref Rgba32 dest)
+        {
+            Vector4 vector = this.ToVector4() * 255F;
+            dest.R = (byte)MathF.Round(vector.X);
```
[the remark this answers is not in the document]
JimBobSquarePants: Maybe.... We'd probably have to scale up and down via shifting to preserve accuracy though...

JimBobSquarePants:
@antonfirsov The CRC won't tell us enough to determine things are correct, only that the compressed data integrity is ok so we can ignore that. I'll revert the reference decoder changes though and do some cleanup there and I'll use tolerance for the 16 bit ones. Using `Rgba32` as our pixel format might actually preserve the same data.

I also just double checked the rgb-16-alpha.png image using BeyondCompare and it's an exact match so that's ok.

@dlemstra If you could rig something in the tests to bridge between an Magick.NET image and ours that would be amazing!

antonfirsov:
@dlemstra do I get the full Rgba64 value from 16bit png images? Is it possible without writing C++ code (eg. using Magick.NET)? Any hint/link for start?

@JimBobSquarePants 👍 for the revert, we should not risk our existing test coverage. If we could load a raw pixel data dump, and do a comparison against it, then we are good with the 16bit stuff! I'll see what I can do tonight.

`Decode_64Bpp_Rgba64_rgb-16-alpha.png` is fine, it just looked like memory garbage to me 😄

JimBobSquarePants:
Hey @antonfirsov Can you do me a biggie and have a look at the image comparison code? The new font tests are failing with some pretty big differences after the merge. It could be rounding errors creeping in during conversion to/from `Rgba64` but it could be something different. 

Perhaps we should use different comparers for 16 bit and 8 bit pixels formats?

e.g.
```
FontShapesAreRenderedCorrectly_LargeText<Rgba32>(provider: Solid2480x3508_(255,255,255,25

SixLabors.ImageSharp.Tests.TestUtilities.ImageComparison.ImageDifferenceIsOverThresholdException : Image difference is over threshold!
Report ImageFrame 0: 
Total difference: 0.1764%
[Δ(1542,1542,1542,0) @ (819,6)];
[Δ(-1542,-1542,-1542,0) @ (822,6)];
[Δ(1285,1285,1285,0) @ (832,6)];
[Δ(2570,2570,2570,0) @ (833,6)];
[Δ(1542,1542,1542,0) @ (834,6)]...
```

antonfirsov:
@JimBobSquarePants I'll have a look tomorrow!

src/ImageSharp/PixelFormats/Rgb48.cs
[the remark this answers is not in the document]
JimBobSquarePants: Check out the new method using offset + shift I nicked from libpng. Much better!

JimBobSquarePants:
@antonfirsov No need to look over the `ImageComprarer` it's correct. There was something, not quite right about the text drawing reference images that made them different. I recreated them and everything works.

antonfirsov:
Good news!

I won't be able do make progress with Magick.NET testing now, so I think it's fine to merge this as-is + add an issue for the 16bit testing topic.
I want to do a second review however in the next few days, hope I can do it by Saturday night.

JimBobSquarePants:
Merging this, it's solid and blocking other work.

src/ImageSharp/PixelFormats/Rgb48.cs
```diff
@@ -0,0 +1,256 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool Equals(Rgb48 other)
+        {
+            return this.R == other.R
+                && this.G == other.G
+                && this.B == other.B;
+        }
+
+        /// <inheritdoc />
+        public override string ToString() => this.ToVector4().ToString();
```
antonfirsov: "({R},{G},{B})" is much better for debugging.

tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
```diff
@@ -142,20 +168,66 @@ public void Decode_Interlaced_ImageIsCorrect<TPixel>(TestImageProvider<TPixel> p
...
             {
-                image.DebugSave(provider);
+                var encoder = new PngEncoder { ColorType = PngColorType.Rgb, BitDepth = PngBitDepth.Bit16 };
 
-                // Workaround a bug in mono-s System.Drawing PNG decoder. It can't deal with 48Bpp png-s :(
-                if (!TestEnvironment.IsLinux && !TestEnvironment.IsMono)
+                if (!SkipVerification(provider))
                 {
-                    image.CompareToOriginal(provider, ImageComparer.Exact);
+                    image.VerifyEncoder(provider, "png", null, encoder, customComparer: ValidatorComparer);
```
antonfirsov: Two calls to `DebugSave()` and `CompareToOriginal()` would more straightforward and readable in a decoder test. ( (`VerifyEncoder()` does the same though.)

tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
```diff
@@ -20,79 +20,105 @@ public class PngDecoderTests
     {
         private const PixelTypes PixelTypes = Tests.PixelTypes.Rgba32 | Tests.PixelTypes.RgbaVector | Tests.PixelTypes.Argb32;
 
+        // TODO: Cannot use exact comparer since System.Drawing doesn't preserve more than 32bits.
+        private static readonly ImageComparer ValidatorComparer = ImageComparer.TolerantPercentage(0.1302F, 2134);
```
antonfirsov: Even if S.D decoder is inaccurate, and the output seems visually perfect, this high level of tolerance seems worrying for me. It might be a sign that something is wrong somewhere (maybe the comparison logic). What is the percentage difference if you lower the threshold?

tests/ImageSharp.Tests/Processing/Processors/Transforms/ResizeTests.cs
```diff
@@ -17,7 +17,7 @@ public class ResizeTests : FileTestBase
     {
         public static readonly string[] CommonTestImages = { TestImages.Png.CalliphoraPartial };
 
-        private static readonly ImageComparer ValidatorComparer = ImageComparer.TolerantPercentage(0.005f);
+        private static readonly ImageComparer ValidatorComparer = ImageComparer.TolerantPercentage(0.069F);
```
antonfirsov: Why do we need this dramatic (~50x) increase in tolerance?

PR #614, opened 2018-06-12
Add optimised drawing path with cached glyph rendering

### LINKED ISSUE #421: Inconsistent behaviour between TextMeasurer and the result text size when DPI is specified
### Prerequisites

### Description
I'm trying to draw a text on an image where its size is determined from the size of the text plus some margin.

For this, I'm using `TextMeasurer.Measure(Font font, float dpi)` to measure the text size then create the image as shown in the following code:

```csharp
private static void Main(string[] args)
{
    var rawImg72 = WriteTextToImage(72);
    System.IO.File.WriteAllBytesAsync(@".\output\img72.png", rawImg72).Wait();

    var rawImg96 = WriteTextToImage(96);
    System.IO.File.WriteAllBytesAsync(@".\output\img96.png", rawImg96).Wait();
}

private static byte[] WriteTextToImage(int dpi)
{
    var text = "SixLabors.ImageSharp";
    var font = SystemFonts.CreateFont("Arial", 16);
    var sizeF = TextMeasurer.Measure(text, new RendererOptions(font, dpi));
    var textSize = new Size(Convert.ToInt32(sizeF.Width), Convert.ToInt32(sizeF.Height));
    var margin = 4 * 2;
    var imageSize = new Size(textSize.Width + margin, textSize.Height + margin);
    var textPosition = new PointF(
        (imageSize.Width - textSize.Width) / 2,
        (imageSize.Height - textSize.Height) / 2);

    using (var img = new Image<Rgba32>(imageSize.Width, imageSize.Height))
    {
        // not sure these are needed or doing anything; the result with or without them is the same
        img.MetaData.HorizontalResolution = dpi;
        img.MetaData.VerticalResolution = dpi;

        img.Mutate(operation => operation
            .Fill(Rgba32.WhiteSmoke)
            .DrawText(text, font, Rgba32.Black, textPosition));

        using (var stream = new System.IO.MemoryStream())
        {
            img.Save(stream, ImageFormats.Png);

            return stream.ToArray();
        }
    }
}
```

The result of this code is the following images:

**72 DPI**
img72

**96 DPI**
img96

Note the following:

- The first image (72 DPI): the size is perfect and the image is horizontally centred, but the vertically is off.
- The second image (96 DPI): the image is vertically centred, but the size is incorrect and the text isn't horizontally centred (or maybe it is because it thinks the image size is the same as with 72 DPI).

### Steps to Reproduce
See Description section.

### System Configuration

- ImageSharp version: 1.0.0-beta0002 (from NuGet)
- Environment (Operating system, version and so on): Windows 10 Pro [Version 10.0.16299.125]
- .NET Framework version: .NET Core 2.0

tocsoft:
### TODO
- [x] Expand test coverage if required (waiting on code cove)
- [x] Update reference images to match slightly updated rendering output.

### Description
I've added a much more optimised text drawing path which makes use of caching individual glyphs the first time they are rendered. This has 2 benefits; 
1. we only have to create a glyph vector once per font/dpi/size.
2. we only have to do the Intersection scanning once per unique glyph.

_The cache could also be expanded in the future to live for a variety of extended life times (instead of the per drawing operation its using now) i.e. we could cache the glyphs for the life time of the `ImageProcessingOperation` or allow for a user to pass into the mutate/clone method some form of lifetime scope that can be used to manage the lifetime of the cache. Or could this be an operation on `MemoryManager` even??_

This drastically reduce rendering time on large amount of text, with larger gains the large the corpus. In addition we also drastically reduce our allocations (we will only now allocate a ComplexPath once per unique glyph now as apposed to before we would allocate one per visible character in the text).

These are the numbers,  in each case we are rendering a word wrapped piece of text with "Hello World" rendered "TextIterations" number of times as a single space separated string.
``` ini

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i5-4690 CPU 3.50GHz (Haswell), 1 CPU, 4 logical and 4 physical cores
.NET Core SDK=2.1.300
  [Host]     : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 64bit RyuJIT
  DefaultJob : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 64bit RyuJIT


```
|                                 Method | TextIterations |         Mean |       Error |      StdDev | Scaled | ScaledSD |      Gen 0 |    Gen 1 |    Gen 2 |  Allocated |
|--------------------------------------- |--------------- |-------------:|------------:|------------:|-------:|---------:|-----------:|---------:|---------:|-----------:|
|             **&#39;System.Drawing Draw Text&#39;** |             **10** |     **961.9 us** |    **19.00 us** |    **30.14 us** |   **1.00** |     **0.00** |          **-** |        **-** |        **-** |      **680 B** |
| &#39;ImageSharp Draw Text - Cached Glyphs&#39; |             10 |   2,385.1 us |    11.65 us |    10.90 us |   2.48 |     0.08 |    58.5938 |        - |        - |   192955 B |
|         &#39;ImageSharp Draw Text - Nieve&#39; |             10 |  19,288.9 us |   121.60 us |   113.75 us |  20.07 |     0.63 |   750.0000 | 281.2500 |        - |  3920624 B |
|                                        |                |              |             |             |        |          |            |          |          |            |
|             **&#39;System.Drawing Draw Text&#39;** |            **100** |   **2,811.3 us** |    **56.13 us** |    **52.50 us** |   **1.00** |     **0.00** |          **-** |        **-** |        **-** |     **7506 B** |
| &#39;ImageSharp Draw Text - Cached Glyphs&#39; |            100 |  10,094.0 us |    45.29 us |    42.37 us |   3.59 |     0.07 |   265.6250 |        - |        - |   861721 B |
|         &#39;ImageSharp Draw Text - Nieve&#39; |            100 | 210,848.5 us | 1,273.35 us | 1,191.10 us |  75.03 |     1.42 | 10250.0000 | 812.5000 | 250.0000 | 40888369 B |

``` ini

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i5-4690 CPU 3.50GHz (Haswell), 1 CPU, 4 logical and 4 physical cores
.NET Core SDK=2.1.300
  [Host]     : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 64bit RyuJIT
  DefaultJob : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 64bit RyuJIT


```
|                                         Method | TextIterations |       Mean |     Error |    StdDev | Scaled | ScaledSD |      Gen 0 |     Gen 1 |    Gen 2 |   Allocated |
|----------------------------------------------- |--------------- |-----------:|----------:|----------:|-------:|---------:|-----------:|----------:|---------:|------------:|
|             **&#39;System.Drawing Draw Text Outline&#39;** |             **10** |   **3.612 ms** | **0.0227 ms** | **0.0201 ms** |   **1.00** |     **0.00** |          **-** |         **-** |        **-** |       **728 B** |
| &#39;ImageSharp Draw Text Outline - Cached Glyphs&#39; |             10 |   6.741 ms | 0.0208 ms | 0.0195 ms |   1.87 |     0.01 |   273.4375 |         - |        - |    865368 B |
|         &#39;ImageSharp Draw Text Outline - Nieve&#39; |             10 |  60.993 ms | 0.2344 ms | 0.1957 ms |  16.89 |     0.10 |  4000.0000 |  250.0000 |        - |  15793256 B |
|                                                |                |            |           |           |        |          |            |           |          |             |
|             **&#39;System.Drawing Draw Text Outline&#39;** |            **100** |  **31.409 ms** | **0.1844 ms** | **0.1725 ms** |   **1.00** |     **0.00** |          **-** |         **-** |        **-** |      **7460 B** |
| &#39;ImageSharp Draw Text Outline - Cached Glyphs&#39; |            100 |  28.109 ms | 0.1337 ms | 0.1250 ms |   0.89 |     0.01 |   656.2500 |         - |        - |   2047190 B |
|         &#39;ImageSharp Draw Text Outline - Nieve&#39; |            100 | 622.104 ms | 3.0993 ms | 2.8991 ms |  19.81 |     0.14 | 48250.0000 | 1437.5000 | 437.5000 | 160002918 B |

### Before 
image
### After 
image

# Issues fixed
fixes #421


JimBobSquarePants:
Looks like the `FontShapesAreRenderedCorrectly` test is failing.

antonfirsov:
@tocsoft I think what you need is to increase the tolerance a bit for the tests still failing. There are floating point implementation differences, leading to super-minor differences in the output.

tocsoft:
when outlining large numbers of characters we now even beat system.drawing 😃

src/ImageSharp.Drawing/Processing/Text/TextGraphicsOptions.cs
```diff
@@ -90,6 +99,16 @@ public TextGraphicsOptions(bool enableAntialiasing)
         /// </summary>
         public float WrapTextWidth { get => this.wrapTextWidth; set => this.wrapTextWidth = value; }
 
+        /// <summary>
+        /// Gets or sets a value indicating the DPI to render text along the X axis.
+        /// </summary>
+        public float DpiX { get => this.dpiX ?? DefaultTextDpi; set => this.dpiX = value; }
```
tocsoft: I snuck this one in... this will allow users to render at DPIs other that 72.

tests/ImageSharp.Tests/Drawing/Text/DrawTextOnImageTests.cs
```diff
@@ -101,6 +107,103 @@ public class DrawTextOnImageTests
...
+            where TPixel : struct, IPixel<TPixel>
+        {
+
+            Font font = CreateFont(fontName, fontSize);
+            string fnDisplayText = text.Replace("\n", "");
+            fnDisplayText = fnDisplayText.Substring(0, Math.Min(fnDisplayText.Length, 4));
+            TPixel colorOutline = NamedColors<TPixel>.Black;
+            TPixel colorFill = NamedColors<TPixel>.Gray;
+
+            provider.VerifyOperation(
```
antonfirsov: This is the expected output right?

image

Don't we need another test? More complex path (eg. circle) but simpler brush/pen. I'm happy to add it.

src/ImageSharp.Drawing/Processing/Text/Processors/DrawTextOnPathProcessor.cs
```diff
@@ -0,0 +1,123 @@
...
+                this.fillRegionProcessor.Brush = this.Brush;
+
+                foreach (IPath p in glyphs)
+                {
+                    this.fillRegionProcessor.Region = new ShapeRegion(p);
+                    this.fillRegionProcessor.Apply(source, sourceRectangle);
+                }
+            }
+
+            if (this.Pen != null)
```
antonfirsov: What if `Brush` and `Pen` are both null? Shouldn't we throw?

(Same for all text processors.)

src/ImageSharp.Drawing/Processing/Text/Processors/DrawTextProcessor.cs
```diff
@@ -0,0 +1,509 @@
...
+            public IPen Pen { get; internal set; }
+
+            public GraphicsOptions Options { get; internal set; }
+
+            public void BeginFigure()
+            {
+                this.builder.StartFigure();
+            }
+
+            public bool BeginGlyph(RectangleF bounds, int cacheKey)
```
antonfirsov: Don't we need a dedicated `GlyphParameters` struct/class with proper equals + gethashcode implementations  instead of an integer key here? Collisions may lead to bugs!

tests/ImageSharp.Tests/Drawing/Text/DrawTextOnImageTests.cs
```diff
@@ -101,6 +107,103 @@ public class DrawTextOnImageTests
...
+            string text)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            Font font = CreateFont(fontName, fontSize);
+            string fnDisplayText = text.Replace("\n", "");
+            fnDisplayText = fnDisplayText.Substring(0, Math.Min(fnDisplayText.Length, 4));
+            TPixel color = NamedColors<TPixel>.Black;
+
+            provider.VerifyOperation(
+                ImageComparer.Tolerant(imageThreshold: 0.1f, perPixelManhattanThreshold: 20),
```
antonfirsov: I think this threshold is too high now, will do some experiments lowering it.

tests/ImageSharp.Tests/Drawing/Text/DrawTextOnImageTests.cs
```diff
@@ -101,8 +107,151 @@ public class DrawTextOnImageTests
...
+            TPixel colorFill = NamedColors<TPixel>.Black;
+            IBrush<TPixel> fillBrush = Brushes.Solid(colorFill);
+            
+            provider.VerifyOperation(
+                TextDrawingComparer,
+                img =>
+                {
+                    int w = (int)(img.Width * 0.6);
+                    int h = (int)(img.Height * 0.6);
+                    IPath path = new EllipsePolygon(img.Width/2, img.Height/2, w, h);
```
antonfirsov: @tocsoft can you confirm, if the output is correct?

image

The first point in `path` is `X=120, Y=300`, but the text starts on the top, so I'm a bit confused. Is it center aligned?

src/ImageSharp.Drawing/Processing/Drawing/Processors/FillRegionProcessor.cs
```diff
@@ -37,22 +37,29 @@ public FillRegionProcessor(IBrush<TPixel> brush, Region region, GraphicsOptions
...
+        /// </summary>
+        public FillRegionProcessor()
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets the brush.
         /// </summary>
-        public IBrush<TPixel> Brush { get; }
+        public IBrush<TPixel> Brush { get; set; }
```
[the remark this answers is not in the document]
JimBobSquarePants: Processors should be immutable. I've already gone through and changed all others before.

tests/ImageSharp.Tests/Drawing/Text/DrawTextOnImageTests.cs
tocsoft: I've been thinking about this and think we should just remove this set of APIs... its just a thin wrapper around `TextBuilder.GenerateGlyphs` from Shapes and you can already draw shapes... it would also encourage users to investigate the shapes apis and also allow us to drop the SixLabors.Shapes.Text package and switch down to just Fonts & Shapes directly, then is users want to do this we can direct them to the optional package that they can use.

tocsoft: it doesn't harm that its one less API to make sure is displaying correctly before we launch too.

src/ImageSharp.Drawing/Processing/Text/Processors/DrawTextProcessor.cs
tocsoft: your right... i'll add it to Fonts get it wired up.

src/ImageSharp.Drawing/Processing/Drawing/Processors/FillRegionProcessor.cs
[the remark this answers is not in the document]
tocsoft: ok.. i'll make them immutable again... was trying to find ways of reducing allocations by allow users to reuse as required, but its not a major gain really

src/ImageSharp.Drawing/Processing/Text/Processors/DrawTextOnPathProcessor.cs
tocsoft: yep

src/ImageSharp.Drawing/Processing/Drawing/Processors/FillRegionProcessor.cs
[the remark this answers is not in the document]
antonfirsov: Most of the processors were not designed to be reusable, they cache stuff of their current execution. Making them reusable would mean that we need to implement and unit test proper reset logic for them, which is a statefulness hell IMO.

tests/ImageSharp.Tests/Drawing/Text/DrawTextOnImageTests.cs
tocsoft: The TextBuilder.GenerateGlyphs are really tests that belong in the Shapes project and not really Imagesharp concerns... as far and Imagesharp is concerned they will be paths and that it renders paths correctly.

antonfirsov:
@tocsoft I pushed a few refactors, if you're fine with them, I'm happy to merge this.

antonfirsov:
Btw. defining a library-wide cache infrastructure is a good idea! (Not easy though.)

A won't mix it with allocators however. A short proposal:

'''C#
public interface ICache
{
	bool TryLookup<TKey, TResult>(TKey key, out TResult);
	void Clear();
}

public class Configuration
{
	// ...
	public ICache Cache { get; set; } = new NullCache(); // no caching by default?
	// ...
}

public interface IImageProcessingContext<TPixel>
{
	// ...
	public ICache Cache { get; set; } = new DefaultCache();
	// ...
}
'''

JimBobSquarePants:
`where TKey : IEquatable<TKey>` nothing worse that slow lookups.

PR #616, opened 2018-06-16
EXIF Support for PNG's

brianpopow:
### Prerequisites

### Description

This MR adds reading and writing eXIf chunks for PNG files. The basic difference between JPG's and PNG's EXIF is, that PNG does not include the Exif code (e. g. "Exif00") in the exif chunk. 

I had to change the ExifWriter a little for that. The GetData method now has a parameter, which indicates, if the Exif Code should be included (defaults to true). 

Please review the changes and let me know what you think.


JimBobSquarePants:
I'll leave this one for @dlemstra to review but I'm thinking we should make the `EXIF00` bytes the sole responsibility of the jpeg format if it's unused in png.

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -233,8 +232,10 @@ public void SetValue(ExifTag tag, object value)
         /// <summary>
         /// Converts this instance to a byte array.
         /// </summary>
+        /// <param name="includeExifIdCode">Indicates, if the Exif ID code should be included.
+        /// This Exif ID code should not be included in case of PNG's. Defaults to true.</param>
         /// <returns>The <see cref="T:byte[]"/></returns>
-        public byte[] ToByteArray()
+        public byte[] ToByteArray(bool includeExifIdCode = true)
         {
             if (this.values == null)
```
dlemstra: You should also check if `this.data` starts with the exif id and skip the first 6 bytes when includeExifIdCode is true.

src/ImageSharp/MetaData/Profiles/Exif/ExifWriter.cs
```diff
@@ -86,23 +84,51 @@ public byte[] GetData()
...
-            result[7] = (byte)'I';
-            result[8] = 0x2A;
-            result[9] = 0x00;
-
-            int i = 10;
-            uint ifdOffset = ((uint)i - StartIndex) + 4;
+            int i = 0;
+            if (includeExifIdCode)
+            {
+                result[0] = (byte)'E';
```
dlemstra: You could also increment `i` when you set the values so you don't need the duplicate code block in the else part.

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
brianpopow: good point, i missed that one. I have added a unit test for that case.

src/ImageSharp/MetaData/Profiles/Exif/ExifWriter.cs
brianpopow: i have changed that

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -247,7 +255,33 @@ public byte[] ToByteArray()
...
+        /// <returns>True, if the byte array starts with the Exif Code</returns>
+        private bool StartsWithExifIdCode(byte[] exifBytes)
+        {
+            if (exifBytes.Length < 6)
+            {
+                return false;
+            }
+
+            // The EXIF ID code: ASCII "Exif" followed by two zeros.
+            byte[] exifCode = { 69, 120, 105, 102, 0, 0 };
```
JimBobSquarePants: Can we avoid this allocation per iteration. Do we know yet whether that identifier is specific to jpeg?

brianpopow: i have changed that. 

I am not sure if its jpeg specific or other formats which support Exif also use this. I only can tell for sure that PNG does not use it. Maybe @dlemstra knows more about this?

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -233,11 +238,19 @@ public void SetValue(ExifTag tag, object value)
         /// <summary>
         /// Converts this instance to a byte array.
         /// </summary>
+        /// <param name="includeExifIdCode">Indicates, if the Exif ID code should be included.
+        /// This Exif ID code should not be included in case of PNG's. Defaults to true.</param>
         /// <returns>The <see cref="T:byte[]"/></returns>
-        public byte[] ToByteArray()
+        public byte[] ToByteArray(bool includeExifIdCode = true)
```
JimBobSquarePants: I want to move the Exif identifier out of the ExifProfile as that's defined as part of the App1 segment itself in section 4.7.2 of the specification. 

http://www.exif.org/Exif2-2.PDF

Ideally we should pass the marker to the `ToByteArray` method when exporting. The `ExifProfile` constructor should have an overload accepting the marker.

JimBobSquarePants: Fantastic, thanks!

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder;
```
JimBobSquarePants: The ExifProfile needs to be format agnostic.

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -239,7 +235,8 @@ public void SetValue(ExifTag tag, object value)
         /// Converts this instance to a byte array.
         /// </summary>
         /// <param name="includeExifIdCode">Indicates, if the Exif ID code should be included.
-        /// This Exif ID code should not be included in case of PNG's. Defaults to true.</param>
+        /// The Exif Id Code is part of the JPEG APP1 segment. This Exif ID code should not be included in case of PNG's.
+        /// Defaults to true.</param>
         /// <returns>The <see cref="T:byte[]"/></returns>
         public byte[] ToByteArray(bool includeExifIdCode = true)
```
JimBobSquarePants: Pass through the code as a `ReadonlySpan<bye>` copying that to the output array.

src/ImageSharp/MetaData/Profiles/Exif/ExifWriter.cs
```diff
@@ -42,13 +43,14 @@ public ExifWriter(IList<ExifValue> values, ExifParts allowedParts)
...
-        /// This Exif ID code should not be included in case of PNG's. Defaults to true.</param>
+        /// The Exif Id Code is part of the JPEG APP1 segment. This Exif ID code should not be included in case of PNG's.
+        /// Defaults to true.</param>
         /// <returns>
         /// The <see cref="T:byte[]"/>.
         /// </returns>
         public byte[] GetData(bool includeExifIdCode = true)
         {
-            uint startIndex = 6;
+            uint startIndex = (uint)ProfileResolver.ExifMarker.Length;
```
JimBobSquarePants: Again, pass this to the method

src/ImageSharp/MetaData/Profiles/Exif/ExifWriter.cs
```diff
@@ -86,14 +85,27 @@ public byte[] GetData()
                 return null;
```
brianpopow: im not sure about this one, i think this was supposed to return null, if only the Exif Id code is present. Is that correct?

brianpopow:
@JimBobSquarePants i tried to follow your suggestions, can you please review the changes and see if that's how you wanted it to be. 

Also any suggestions for more / better unit tests are welcome.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -613,6 +614,21 @@ private void WritePhysicalChunk<TPixel>(Stream stream, Image<TPixel> image)
...
+        /// <summary>
+        /// Writes the eXIf chunk to the stream, if any EXIF Profile values are present in the meta data.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="stream">The <see cref="Stream"/> containing image data.</param>
+        /// <param name="image">The image.</param>
+        private void WriteExifChunk<TPixel>(Stream stream, Image<TPixel> image)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            if (image.MetaData.ExifProfile?.Values.Count > 0)
```
JimBobSquarePants: We need to alter this slightly as there's extra checks and balances.

`image.MetaData.SyncProfiles();` is required before writing according to the jpeg encoder and we need additional length checks as the spec dictates it's length should match jpeg limitations. 

https://github.com/SixLabors/ImageSharp/blob/d171acba8051063a68467e912f1a81fae1126ed8/src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs#L608

src/ImageSharp/MetaData/Profiles/Exif/ExifWriter.cs
JimBobSquarePants: @dlemstra What do you think?

tests/ImageSharp.Sandbox46/ImageSharp.Sandbox46.csproj
```diff
@@ -20,9 +20,9 @@
   <ItemGroup>
     <PackageReference Include="BitMiracle.LibJpeg.NET" Version="1.4.280" />
     <PackageReference Include="xunit" Version="2.3.1" />
-    <PackageReference Include="Moq" Version="4.8.2" />
-    <!--<PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />-->
-    <!--<PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />-->
+    <PackageReference Include="Moq" Version="4.8.3" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.2" />
```
JimBobSquarePants: Do we need these references? They were commented out.

brianpopow: no we do not need them, i have removed it again

src/ImageSharp/Formats/Png/PngEncoderCore.cs
brianpopow: i have added the SyncProfile before writing the chunk. 

For the length check: I could add a check, if the length exceeds 2^16, like the JpegEncoder does, and throw an exception, but that feels wrong to me. This restriction is Jpeg specific not PNG specific. 

In the recommendations they state, when writing PNG Exif to Jpeg: 
> the total length of the eXIf chunk data may need to be adjusted

But how to decide what to cut off here is not clear to me. I really would like to hear @dlemstra opinion on that.

JimBobSquarePants: Looking at this it appears we should simply the split data into 64k chunks.  http://dev.exiv2.org/issues/1232

That means we need an extend method, like the ICC readeron decoding.
https://github.com/SixLabors/ImageSharp/blob/d7bd82b79ee396a42b13a83563e4fbdd951c2a55/src/ImageSharp/Formats/Jpeg/PdfJsPort/PdfJsJpegDecoderCore.cs#L508

JimBobSquarePants:
@brianpopow Looks great to me 👍 thanks for persevering!

@dlemstra Would you be able to cast your eye over it just to be certain I haven't missed anything?

src/ImageSharp/MetaData/Profiles/Exif/ExifReader.cs
```diff
@@ -79,6 +79,7 @@ public List<ExifValue> ReadValues()
 
             if (this.ReadString(4) == "Exif")
             {
+                // two zeros are expected to follow the Exif Id code
```
dlemstra: Do we really need this comment?

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -639,7 +671,7 @@ private void WriteExifProfile(ExifProfile exifProfile)
         /// </exception>
         private void WriteIccProfile(IccProfile iccProfile)
         {
-            // Just incase someone set the value to null by accident.
+            // Just in-case someone set the value to null by accident.
```
dlemstra: Can you remove this comment instead? Not every JPEG file has an IccProfile.

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -230,11 +229,26 @@ public void SetValue(ExifTag tag, object value)
...
+        }
+
         /// <summary>
         /// Converts this instance to a byte array.
         /// </summary>
+        /// <param name="exifIdCode">The Exif Id Code is part of the JPEG APP1 segment (Exif00). Those bytes will be written at
+        /// the beginning of the array. This Exif ID code should not be included in case of PNG's.</param>
         /// <returns>The <see cref="T:byte[]"/></returns>
-        public byte[] ToByteArray()
+        public byte[] ToByteArray(ReadOnlySpan<byte> exifIdCode = default)
```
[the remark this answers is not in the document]
brianpopow: yes i think you are right, it feels a bit odd. This is jpeg specific and should be handled by the encoder/decoder

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
```diff
@@ -230,11 +229,26 @@ public void SetValue(ExifTag tag, object value)
             this.values.Add(newExifValue);
         }
 
+        /// <summary>
+        /// Extends the profile with additional data.
+        /// </summary>
+        /// <param name="bytes">The array containing addition profile data.</param>
+        public void Extend(byte[] bytes)
```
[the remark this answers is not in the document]
brianpopow: while this extending is also something jpeg specific and therefore should be handled by the decoder, i could not figure out a way to change it without making it more complicated. The problem is, that we do not know in advance how many APP1 marker there will be for very large EXIF data.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -607,27 +609,57 @@ private void WriteDefineQuantizationTables()
...
             {
                 return;
             }
 
-            if (data.Length > Max)
+            int remaining = data.Length;
+            int bytesToWrite = remaining > MaxBytesApp1 ? MaxBytesApp1 : remaining;
+            int app1Length = bytesToWrite + 2;
+
+            // write the app1 header
```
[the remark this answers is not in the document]
brianpopow: i have removed the unnecessary comments

src/ImageSharp/MetaData/Profiles/Exif/ExifProfile.cs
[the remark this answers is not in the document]
brianpopow: @dlemstra: i have moved the extending of the exif data to the jpeg decoder now, please let me know if you think this is a viable solution.

JimBobSquarePants:
@brianpopow I'm going to have a look at this PR this evening and fix the merge conflicts and any final tweaks, thanks for being so patient with us and helping out! 👍

JimBobSquarePants:
Ok merging this now as it all looks good to me.

PR #617, opened 2018-06-16
Get rid of PixelAccessor<T>

antonfirsov:
### Prerequisites

### Description
`PixelAccessor<T>` has been originally used to allow efficient pixel manipulation through pinning + pointer arithmetics. Thanks to `Unsafe.Add()` and `Span<T>` pinning is no longer necessary, but we kept `PixelAccessor<T>` as a legacy component (mostly used in tests) to minimize refactor efforts.

- This PR replaces all `PixelAccessor<T>` usages with `Buffer2D<T>` and removes the class from the codebase. 
- The `IBuffer2D<T>` abstraction is no longer necessary, so it has been replaced with `Buffer2D<T>`.

I think the changes are trivial, so I plan to merge it as soon as I get at least one quick review.


src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -574,6 +580,7 @@ private void ReadImageHeaders(Stream stream, out bool inverted, out byte[] palet
                 this.stream.Read(palette, 0, colorMapSize);
             }
 
+            // TODO: ReSharper tells this expression is always true, looks like he's pretty right about it:
             if (this.infoHeader.Width > int.MaxValue || this.infoHeader.Height > int.MaxValue)
```
JimBobSquarePants: You mean always false yeah? Been struggling to find a definitive answer but I think 65535x65535 is the maximum size on a 64 bit machine.

https://stackoverflow.com/questions/29175585/what-is-the-maximum-resolution-of-c-sharp-net-bitmap

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -103,36 +103,42 @@ public Image<TPixel> Decode<TPixel>(Stream stream)
                 this.ReadImageHeaders(stream, out bool inverted, out byte[] palette);
 
                 var image = new Image<TPixel>(this.configuration, this.infoHeader.Width, this.infoHeader.Height);
-                using (PixelAccessor<TPixel> pixels = image.Lock())
+
+                Buffer2D<TPixel> pixels = image.Frames.RootFrame.PixelBuffer;
```
JimBobSquarePants: You've defined `GetRootFramePixelBuffer()` below but are not using it here.

src/ImageSharp/ImageFrame{TPixel}.cs
```diff
@@ -216,19 +216,10 @@ internal ImageFrame(Configuration configuration, ImageFrame<TPixel> source)
         /// It is imperative that the accessor is correctly disposed off after use.
         /// </remarks>
         /// </summary>
-        /// <returns>The <see cref="PixelAccessor{TPixel}"/></returns>
-        internal PixelAccessor<TPixel> Lock()
+        /// <returns>The <see cref="Buffer2D{TPixel}"/></returns>
+        internal Buffer2D<TPixel> Lock()
```
JimBobSquarePants: `Lock` is an odd name now that we are not pinning. Wouldn't `GetRootFramePixelBuffer()` also call this? Is it required anymore?

antonfirsov: Forgot to remove this, good catch!

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
antonfirsov: Always false of course :)

`65535 * 65535 = 4 294 836 225` and `int.MaxValue = 2 147 483 647` so `Width` and `Height` can't exceed that value in neither case.

`Width` and `Height` should be defined as `long` if we want to allow those sizes. (But I think we don't.)

JimBobSquarePants: I think we're crossing wires here. `65535` is the max value for a `ushort` so not anywhere close to `int.MaxValue`. Incidentally that's the maximum length of any dimensions for a jpeg also.

antonfirsov: Omg true that, I should not make statements before lunch!

Maybe we should define something like an `Image.MaximumDimension` constant.

JimBobSquarePants: I have thought about that before but it varies from format to format so we can't. 

I should print this out and frame it btw, I think it's the first time I've ever corrected you! 😝

antonfirsov: I think it already happened a few times! 😄
Fixed this stuff in my newest commit.

JimBobSquarePants:
All good now, @antonfirsov I'll leave you to merge. 👍

PR #618, opened 2018-06-17
Add span overloads for Image.Load() and Image.DetectFormat()

antonfirsov:
### Prerequisites

### Description
This PR implements API-s proposed in #565 by @jherby2k with the help of [System.IO.UnmanagedMemoryStream](https://www.nuget.org/packages/System.IO.UnmanagedMemoryStream/) which is a new dependency for targets > netstandard 1.1.

Related tests have been refactored heavily (isolation, cleanup, code share).


src/ImageSharp/ImageSharp.csproj
```diff
@@ -44,6 +44,9 @@
     <PackageReference Include="System.Memory" Version="4.5.0" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.0" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' != 'netstandard1.1' ">
```
tocsoft: you can skip this for our net standard 2.0 > targets too so you probably want to change this to `'$(TargetFramework)' == 'netstandard1.3'` instead

antonfirsov:
@tocsoft good catch, fixed!

src/ImageSharp/Image.FromBytes.cs
```diff
@@ -171,5 +172,125 @@ public static Image<TPixel> Load<TPixel>(Configuration config, byte[] data, IIma
...
+        /// By reading the header on the provided byte array this calculates the images format.
+        /// </summary>
+        /// <param name="config">The configuration.</param>
+        /// <param name="data">The byte array containing encoded image data to read the header from.</param>
+        /// <returns>The mime type or null if none found.</returns>
+        public static unsafe IImageFormat DetectFormat(Configuration config, ReadOnlySpan<byte> data)
+        {
+            fixed (byte* ptr = &data.GetPinnableReference())
+            {
+                using (var stream = new UnmanagedMemoryStream(ptr, data.Length))
```
davidfowl: It's unfortunate to allocate a heap object to wrap the span...

antonfirsov: Makes sense in the `DetectFormat()` use case, thanks!

When doing a full image decoding (`Image.Load(...)`), that allocation is a drop in the bucket however. Using `UnmanagedMemoryStream` seemed to be the best way for "converting" a span into a `Stream`.

PR #631, opened 2018-06-24
Remove NotNull guard on value type

Jjagg:
### Prerequisites

### Description

I came across this `NotNull` guard on a `DenseMatrix` while checking out the dithering code ;) (JetBrains' Heap Allocations Viewer plugin for Rider drew my attention to it, because it got boxed!)


antonfirsov:
Good catch! After seeing this PR and #627 I think we should turn `Guard.NotNull()` into a generic method to prevent the incorrect usage:

```C#
public static Guard.NotNull<T>(T obj) where T : class { ... }
```

Issue #632, opened 2018-06-25
Diffuser modifies quantized image with same palette


Jjagg:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Running the [`ErrorDiffusionPaletteProcessor`](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Processing/Dithering/Processors/ErrorDiffusionPaletteProcessor.cs) on a quantized image with the same palette that was used for quantization modifies the image. This is not desirable since the point of dithering is to improve visual fidelity with the limited palette, while in this case a perfectly faithful quantized image is modified and image quality is degraded.

Of course a dithering algorithm can't work well for 100% of cases, but bad quality on perfectly quantized images can prove an issue in practice. With a two-pass quantization algorithm on an image with few distinct colors this results in bad image quality.

The code responsible is [here](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Processing/Dithering/Processors/ErrorDiffusionPaletteProcessor.cs#L99-L107):
```
pair = this.GetClosestPixelPair(ref sourcePixel, this.Palette);
sourcePixel.ToRgba32(ref rgba);
luminance = isAlphaOnly ? rgba.A : (.2126F * rgba.R) + (.7152F * rgba.G) + (.0722F * rgba.B);
TPixel transformedPixel = luminance >= threshold ? pair.Second : pair.First;
```

Since all references I found just got the closest color in the palette, I tried changing the implementation to get only the closest color. I didn't see worse visual quality in the test images I ran the diffusion on, though I must say I don't have a trained eye (and I guess this might be subjective too).

The single-pixel implementation also provides a major performance benefit compared to the current implementation. When dithering we can check if there is no error and skip the dither for the pixel. For images that have few colors or all colors in the palette this provides a major speed boost. I added some tests to the benchmark project with 3 images; the lake image from the test set (lots of colors), the lake image quantized with 256 colors and processed with a palette of 128 colors, and the lake image quantized and processed with the full quantization palette.

``` ini
[...]

Issue #630, opened 2018-06-24
GIF dithering causes massive encoding slowdown


TodesBrot:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
There are three main points of concern:
Saving GIFs with dithering enabled massively slows down the encoding time.
When dithering is disabled, decoding and reencoding GIFs is much slower than encoding them for the first time.
This is not directly an issue but seems to be related: Setting every frame's DisposalMethod to RestoreToBackground has a positive impact on the time it takes to reencode the resulting image.

### Steps to Reproduce
![spritesheet](https://user-images.githubusercontent.com/8228822/41818157-fc42e12a-77a8-11e8-81ae-08c39c177dc6.png)
Save this image as "Spritesheet.png" and run the following code:
```
bool ditheringenabled = true; //Dithering causes a massive encoding slowdown.
bool restoretobackground = false; //This makes reencoding faster.

int spritecount = 10;

Image<Rgba32> img = SixLabors.ImageSharp.Image.Load("Spritesheet.png");
int w = img.Width / spritecount;
Image<Rgba32> gif = new Image<Rgba32>(w, img.Height);

for (int i = 0; i < spritecount; i++)
{
	gif.Frames.AddFrame(img.Clone(ctx => ctx.Crop(new SixLabors.Primitives.Rectangle(w * i, 0, w, img.Height))).Frames[0]);
	
	//RestoreToBackground causes the reencoding to be much faster.
	gif.Frames[i].MetaData.DisposalMethod = restoretobackground ? DisposalMethod.RestoreToBackground : DisposalMethod.NotDispose;
}

//Disabling the dithering gets completely rid of the initial encoding slowdown.
GifEncoder encoder = new GifEncoder() { Quantizer = new OctreeQuantizer(dither: ditheringenabled) };

DateTime start1 = DateTime.Now;
gif.Save("Result.gif", encoder);
TimeSpan time1 = (DateTime.Now - start1);

Image<Rgba32> reencode = SixLabors.ImageSharp.Image.Load("Result.gif");
DateTime start2 = DateTime.Now;
[...]

PR #637, opened 2018-06-26
Gif and Quantization Improvements

JimBobSquarePants:
### Prerequisites

### Description
These changes do a few things (All related) Fixing #630 and #632 

- Adds palette options to gif encoding Global (default) vs Local. This means smaller animated gifs.
- Improves performance and memory usage of quantizers ~3x faster
- Improves performance and quality of dithering algorithms.


JimBobSquarePants:
Ok. Good to review once tests have passed.

tests/ImageSharp.Tests/Formats/Png/PngEncoderTests.cs
```diff
@@ -19,7 +19,9 @@ namespace SixLabors.ImageSharp.Tests
 
     public class PngEncoderTests
     {
-        private const float ToleranceThresholdForPaletteEncoder = 0.2f / 100;
+        // This is bull. Failing online for no good reason. 
+        // The images are an exact match. Maybe the submodule isn't updating?
+        private const float ToleranceThresholdForPaletteEncoder = 0.2273F;
```
antonfirsov: Super-strange. I hope it's not my algorithm that sucks.

Maybe we could print the "last modified" timestamp of submodule files, or something like that.

JimBobSquarePants: It’s weird. The exact same difference before and after the update.

JimBobSquarePants:
Hey @iamcarbon Check this, 6x speedup!

With the new global palette mode we're down to 56 seconds with 11.3Mb output.

global color palette performance

With the local palette mode it's 1 minute with 13Mb output.

local color palette performance

Here's the two images to compare. Local, of course is flawless, global suffers slightly since the image reuses the palette from the first frame.

[gifs.zip](https://github.com/SixLabors/ImageSharp/files/2141080/gifs.zip)

iamcarbon:
@JimBobSquarePants Heck yes! This is a gigantic speedup!!!!

tests/ImageSharp.Tests/Formats/Png/PngEncoderTests.cs
antonfirsov: @JimBobSquarePants I think I figured it out! 

Actually, it fails for me locally when I change the tolerance back to the original value. This is my actual output image for palette size 80:
palettecolortype_wuquantizer_palette-8bpp__palettesize-80

It differs from both the current and the past version of the same image in the ReferenceOutput repository. Can you confirm that it means that the actual output of the execution on my PC differs from yours? This could be only explained by hardware (CPU) differences. I suspect a method like [`WuFrameQuantizer<T>.Volume()`](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Processing/Quantization/FrameQuantizers/WuFrameQuantizer%7BTPixel%7D.cs#L323) as the source.

If that is the case, this is exactly the reason why we have a tolerance in our comparer :) I think you have overshoot the tolerance however, based on my local tests `1.5F/100` should be good enough. It's not a percentage value in this class, you defined `22.73%` actually! (I usually postfix my stuff with ***Percent when I mean percentage.)

src/ImageSharp/Formats/Gif/GifColorTableMode.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace SixLabors.ImageSharp.Formats.Gif
+{
+    /// <summary>
+    /// Provides enumeration for the available Gif color table modes.
+    /// </summary>
+    public enum GifColorTableMode
+    {
+        /// <summary>
+        /// A single color table is calculated from the first frame and reused for subsequent frames.
```
antonfirsov: We should probably allow the user to select the baseline frame.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -72,45 +81,95 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
...
-            int index = this.GetTransparentIndex(quantized);
-
             // Write the header.
             this.WriteHeader(stream);
 
-            // Write the LSD. We'll use local color tables for now.
-            this.WriteLogicalScreenDescriptor(image, stream, index);
+            // Write the LSD.
+            int index = this.GetTransparentIndex(quantized);
+            bool useGlobalTable = this.colorTableMode.Equals(GifColorTableMode.Global);
```
antonfirsov: Do you swear you never coded in Java? 😄

src/ImageSharp/Processing/Dithering/ErrorDiffusion/ErrorDiffuserBase.cs
```diff
@@ -74,9 +74,21 @@ public void Dither<TPixel>(ImageFrame<TPixel> image, TPixel source, TPixel trans
         {
             image[x, y] = transformed;
 
+            // Equal? Break out as there's nothing to pass.
```
antonfirsov: There are still many optimization opportunities left here, but we need to change (extend?) the `IErrorDiffuser` API for that.

Not related to this PR but `[MethodImpl(MethodImplOptions.AggressiveInlining)]` is unnecessary on `Dither<TPixel>()`. The compiler won't be able to inline it.

src/ImageSharp/Formats/Gif/GifColorTableMode.cs
JimBobSquarePants: I did think about this but if someone creates a custom `PaletteQuantizer` implementation then it doesn't matter what frame is used. Plus it would vary from image to image.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
JimBobSquarePants: I've read a lot of Java source 😄

tests/ImageSharp.Tests/Formats/Png/PngEncoderTests.cs
JimBobSquarePants: Wow yeah! It's definitely different to the output on my machine. That explains that then.

I need to get my head around the tolerance values. At the moment writing `1.5F/100` (Is that 1% over 100 pixels?) isn't immediately intuitive and we should write a helper that makes it more clear.

src/ImageSharp/Processing/Dithering/ErrorDiffusion/ErrorDiffuserBase.cs
JimBobSquarePants: I thought there were improvements to what can be inlined that would allow this. I could be wrong though. 🤷‍♂️

tests/ImageSharp.Tests/Formats/Png/PngEncoderTests.cs
antonfirsov: @JimBobSquarePants `tolerance == 0.01` means 1%, which can be interpreted as the following:
- Either: An RGB image should be white, but 1% of white pixels are black (the rest is white)
- Or: **all** the pixels in the image have a 1% difference compared to the expected values

I tried to explain this [here](https://github.com/SixLabors/ImageSharp/blob/master/tests/ImageSharp.Tests/TestUtilities/ImageComparison/TolerantImageComparer.cs#L49) but I think I wasn't clear enough.

[`ImageComparer.TolerantPercentage()`](https://github.com/SixLabors/ImageSharp/blob/master/tests/ImageSharp.Tests/TestUtilities/ImageComparison/ImageComparer.cs#L31) is a helper that expects tolerance value scaled into `0-100` it will convert it to the `0-1.0` interval. It's useful because tolerance values are typically very small.

src/ImageSharp/Processing/Dithering/ErrorDiffusion/ErrorDiffuserBase.cs
antonfirsov: If you invoke a method through interface or virtual call on a base class, it can't be inlined, because the code is not known at JIT time. (The virtual invocation is responsible to dispatch the method at runtime.)

tests/ImageSharp.Tests/Formats/Png/PngEncoderTests.cs
JimBobSquarePants: That's much more clear, thanks!

Issue #635, opened 2018-06-26
GIF Encoder: Exploit transparency


Jjagg:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Feature

The GIF format supports reusing pixels from previous frames by setting the Disposal Method to 1 ("Do not dispose") in the previous frame and using a transparent pixel in the current frame. This feature can be exploited to produce better quality GIFs for images where subsequent frames keep the same color values.  My suggestion is to expose this as an option for the `GifEncoder`. To take full advantage the disposal method for all frames should be set to 1, so maybe that should be overriden when the option is enabled.

The benefit to doing this is that we can reduce the number of colors in the `ImageFrame` and get better quantization results and ultimately a better quality GIF. We can optionally also check if we only need to redraw a subarea of the entire frame. I don't think that will be often applicable for typical use-cases. Though when it can be applied it will reduce file size (and quantization cost if quantization can happen on a subset of an image -> #633).

### Implementation

Below I refer to an `ImageFrame` as just frame and to a frame as it will be visible in the GIF as a rendered frame.

The implementation is straightforward. Before quantizing a frame (if the previous frame is preserved) we can replace each pixel that has the same value as (or is within some distance of) the pixel at the same position in the last rendered frame with a transparent pixel. We need to allocate two frames for this process, one to keep track of the rendered frame and one to be a copy of the current frame that we can make transparent. After quantizing we can update the rendered frame by copying non-transparent pixels over from the `QuantizedFrame`.

I have a WIP branch that implements this behavior in the GIF encoder (not yet as an option). The process of comparing and replacing pixels is implemented as an `ImageProcessor` that takes a reference `ImageFrame` (or a `QuantizedFrame` or a single color), a maximum distance and a replacement color.
I can set up a PR to further discuss this and polish the implementation if this proposal is accepted :)

JimBobSquarePants:
Thanks @Jjagg this would be a great improvement.

Just a quick note. IImageProcessor implementations should be separate from codecs so any behaviour should be written as part of the codec codebase itself.

Jjagg:
[...]

PR #638, opened 2018-06-26
WIP GIF Encoder: option to reuse repeated pixels

Jjagg:
### Prerequisites

### Description

This adds an option to the GIF encoder that can improve image quality. Explanation is in #635. 
I didn't write tests yet.


src/ImageSharp/Formats/Gif/GifEncoder.cs
```diff
@@ -30,6 +30,14 @@ public sealed class GifEncoder : IImageEncoder, IGifEncoderOptions
         /// </summary>
         public IQuantizer Quantizer { get; set; } = new OctreeQuantizer();
 
+        /// <summary>
+        /// Gets or sets a value indicating whether pixels that stay the same between two frames should be reused
+        /// by forcing the <see cref="DisposalMethod"/> of all frames to <see cref="DisposalMethod.NotDispose"/>
+        /// and making the repeated pixel transparent. This improves GIF quality when pixels are
+        /// repeated across frames.
```
Jjagg: This is kind of verbose, but I think this is all relevant information to the user so not sure how to improve.

src/ImageSharp/Formats/Gif/GifEncoder.cs
```diff
@@ -30,6 +30,14 @@ public sealed class GifEncoder : IImageEncoder, IGifEncoderOptions
...
+        /// </summary>
+        public bool CutRepeatedPixels { get; set; } = false;
```
Jjagg: I defaulted this to `false` because it overrides the `DisposalMethod` of frames which might be unexpected. Not sure about this behavior though.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -347,5 +388,61 @@ private void WriteImageData<TPixel>(QuantizedFrame<TPixel> image, Stream stream)
...
+        /// <summary>
+        /// Makes pixels that are repeated transparent.
+        /// </summary>
+        /// <param name="renderedFrame">The currently rendered frame.</param>
+        /// <param name="sourceFrame">The source frame.</param>
+        /// <param name="cutFrame">
+        /// The output frame. Pixels will be made transparent if the source color is the same as the rendered color
+        /// and get the same color as in the <paramref name="sourceFrame"/> if it is not.
+        /// </param>
+        private void DoCutRepeatedPixels<TPixel>(Buffer2D<TPixel> renderedFrame, ImageFrame<TPixel> sourceFrame, ImageFrame<TPixel> cutFrame)
```
Jjagg: Might actually be nice to put these methods and the memory management in a class (or partial class) to keep it out of the core encoder. This kind of stuff makes it harder to see what's in the GIF spec and what's extra. Also cleaner if more optimizations like this are added later.

src/ImageSharp/Formats/Gif/GifEncoder.cs
JimBobSquarePants: I don't think this should be a member of the options and should only be determined by the `DisposalMethod` enumeration.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -57,6 +65,7 @@ public GifEncoderCore(MemoryAllocator memoryAllocator, IGifEncoderOptions option
             this.textEncoding = options.TextEncoding ?? GifConstants.DefaultEncoding;
             this.quantizer = options.Quantizer;
             this.ignoreMetadata = options.IgnoreMetadata;
+            this.cutRepeatedPixels = options.CutRepeatedPixels;
```
JimBobSquarePants: Set via `DisposalMethod`

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -94,21 +103,51 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
...
                 this.WriteImageDescriptor(frame, stream);
                 this.WriteColorTable(quantized, stream);
                 this.WriteImageData(quantized, stream);
 
                 quantized = null; // So next frame can regenerate it
             }
 
+            if (this.cutRepeatedPixels)
+            {
+                renderedFrame.Dispose();
```
JimBobSquarePants: We tend to just use the elvis operator in cases like this.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
JimBobSquarePants: I think they're ok in this class. It appears to me that what you're doing actually conforms to the specification no?

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -94,21 +103,51 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
...
             foreach (ImageFrame<TPixel> frame in image.Frames)
             {
                 if (quantized == null)
                 {
-                    quantized = this.quantizer.CreateFrameQuantizer<TPixel>().QuantizeFrame(frame);
+                    ImageFrame<TPixel> frameToQuantize = frame;
+
+                    if (this.cutRepeatedPixels)
+                    {
+                        this.DoCutRepeatedPixels(renderedFrame, frame, cutFrame);
```
JimBobSquarePants: Isn't this unnecessary for the first frame?

src/ImageSharp/Formats/Gif/IGifEncoderOptions.cs
```diff
@@ -25,5 +25,13 @@ internal interface IGifEncoderOptions
...
         /// </summary>
         IQuantizer Quantizer { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether pixels that stay the same between two frames should be reused
+        /// by forcing the <see cref="DisposalMethod"/> of all frames to <see cref="DisposalMethod.NotDispose"/>
+        /// and making the repeated pixel transparent. This improves GIF quality when pixels are
+        /// repeated across frames.
+        /// </summary>
+        bool CutRepeatedPixels { get; }
```
JimBobSquarePants: Remove this please.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -144,22 +147,52 @@ private void EncodeGlobal<TPixel>(Image<TPixel> image, QuantizedFrame<TPixel> qu
...
+
+                    if (lastFrameDisposalMethod == DisposalMethod.NotDispose)
+                    {
+                        this.DoCutRepeatedPixelsGlobal(renderedFrame, paletteQuantized, transparencyIndex);
+                    }
+                    else
                     {
-                        this.WriteImageData(paletteQuantized, stream);
+                        renderedFrame.Dispose();
+                        renderedFrame = paletteQuantized;
```
Jjagg: Pretty sure this is incorrect for `DisposalMethod.RestoreToBackground` and `RestoreToPrevious` since the quantized palette may have transparent pixels. Not entirely sure how to handle this, because I don't have a good grip on which pixels are used when `DisposalMethod`s are mixed.

Jjagg:
For the global palette, the cutting step happens after quantization. That's because we don't benefit from cutting out colors since the palette stays the same anyway. The potential benefit for global palette encoding is that we can have a bunch of transparent indices, so the compression can do a better job. Not sure how effective this will be in practice though.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
Jjagg: This is only an issue if the image frames are already optimized for GIF encoding by the user. So if they take advantage of the different disposal methods and transparency. Should that be supported or can I assume each frame stands on its own while encoding?

Jjagg: Okay, thinking about this some more it really doesn't matter :p If pixels are already transparent we can't optimize them anyway. So I think this is completely fine. Still gotta write the tests though!

Jjagg:
@JimBobSquarePants I don't have time to finish this PR for a while (probably for about another month or so). Feel free to close it in the meantime if you want if you don't want it to clutter the PR list. Or if anyone wants to take over that's fine too. I'll get back to this when I can :)

Jjagg:
I'm gonna close this for now. I'll open a new PR when I manage to pick this back up :)

JimBobSquarePants:
Ok chum, looking forward to seeing a new one in the future, I think it's a great addition to the library!

PR #644, opened 2018-07-02
Feature: histogram equalization

brianpopow:
### Prerequisites

### Description

This PR adds a HistogramEqualization Processor for enhancing the global contrast of an image.

Example image:

unequalized_hawkes_bay_nz

After Equalization: 

unequalized_hawkes_bay_nz_equalized


src/ImageSharp/Processing/Contrast/HistogramEqualizationExtension.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Six Labors and contributors.
+// Licensed under the Apache License, Version 2.0.
+
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Processing.Contrast
```
JimBobSquarePants: `SixLabors.ImageSharp.Processing.Normalization`?

brianpopow: i have changed that

src/ImageSharp/Processing/Contrast/HistogramEqualizationProcessor.cs
```diff
@@ -0,0 +1,117 @@
...
+        protected override void OnFrameApply(ImageFrame<TPixel> source, Rectangle sourceRectangle, Configuration configuration)
+        {
+            var rgb48 = default(Rgb48);
+            var rgb24 = default(Rgb24);
+            int numberOfPixels = source.Width * source.Height;
+            bool is16bitPerChannel = typeof(TPixel) == typeof(Rgb48) || typeof(TPixel) == typeof(Rgba64);
+
+            // build the histogram of the grayscale levels
+            int luminanceLevels = is16bitPerChannel ? 65536 : 256;
+            int[] histogram = new int[luminanceLevels];
```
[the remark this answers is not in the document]
brianpopow: im using MemoryAllocator now, please check if im using it correctly.

src/ImageSharp/Processing/Contrast/HistogramEqualizationProcessor.cs
```diff
@@ -0,0 +1,117 @@
...
+            for (int i = 0; i < histogram.Length; i++)
+            {
+                if (histogram[i] != 0)
+                {
+                    cdfMin = cdf[i];
+                    break;
+                }
+            }
+
+            int[] lut = new int[luminanceLevels];
```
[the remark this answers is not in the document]
brianpopow: removed this allocation, its not necessary, using the cdf array instead

src/ImageSharp/Processing/Contrast/HistogramEqualizationProcessor.cs
```diff
@@ -0,0 +1,117 @@
...
+            double numberOfPixelsMinusCdfMin = (double)(numberOfPixels - cdfMin);
+            int luminanceLevelsMinusOne = luminanceLevels - 1;
+            for (int y = 0; y < source.Height; y++)
+            {
+                Span<TPixel> row = source.GetPixelRowSpan(y);
+                for (int x = 0; x < source.Width; x++)
+                {
+                    TPixel sourcePixel = row[x];
+
+                    int luminance = this.GetLuminance(sourcePixel, is16bitPerChannel, ref rgb24, ref rgb48);
```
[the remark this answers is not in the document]
brianpopow: yes and i see no good way around this. 

First i need to access all pixels to calculate the histogram. 
After that i need the luminance of each pixel again to remap it to the new value.

src/ImageSharp/Processing/Normalization/HistogramEqualizationProcessor.cs
```diff
@@ -0,0 +1,120 @@
...
+        where TPixel : struct, IPixel<TPixel>
+    {
+        /// <inheritdoc/>
+        protected override void OnFrameApply(ImageFrame<TPixel> source, Rectangle sourceRectangle, Configuration configuration)
+        {
+            var rgb48 = default(Rgb48);
+            var rgb24 = default(Rgb24);
+            MemoryAllocator memoryAllocator = configuration.MemoryAllocator;
+            int numberOfPixels = source.Width * source.Height;
+            bool is16bitPerChannel = typeof(TPixel) == typeof(Rgb48) || typeof(TPixel) == typeof(Rgba64);
```
tocsoft: we generally don't have ImageProcesses having any logic dependent on the `PixelType` do we really  need this? do we even need to know that its a 16bit image? normally we would convert the pixel to a Vector4 and process the buffer with no need for branching.

src/ImageSharp/Processing/Normalization/HistogramEqualizationProcessor.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            // build the histogram of the grayscale levels
+            int luminanceLevels = is16bitPerChannel ? 65536 : 256;
+            Span<int> histogram = memoryAllocator.Allocate<int>(luminanceLevels, clear: true).GetSpan();
+            for (int y = 0; y < source.Height; y++)
```
tocsoft: this can be done with a single loop of 
```c#
var pixels = source.GetPixelSpan();
for (int i = 0; i < pixels.Length; i++){
    int luminance = this.GetLuminance(in pixels[i], is16bitPerChannel, ref rgb24, ref rgb48);
    histogram[luminance]++;
}
```

src/ImageSharp/Processing/Normalization/HistogramEqualizationProcessor.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            var rgb48 = default(Rgb48);
+            var rgb24 = default(Rgb24);
+            MemoryAllocator memoryAllocator = configuration.MemoryAllocator;
+            int numberOfPixels = source.Width * source.Height;
+            bool is16bitPerChannel = typeof(TPixel) == typeof(Rgb48) || typeof(TPixel) == typeof(Rgba64);
+
+            // build the histogram of the grayscale levels
+            int luminanceLevels = is16bitPerChannel ? 65536 : 256;
+            Span<int> histogram = memoryAllocator.Allocate<int>(luminanceLevels, clear: true).GetSpan();
```
tocsoft: should be a called with a using statment 

```c#
using(IBuffer<int> buffer = memoryAllocator.AllocateClean<int>(luminanceLevels))
{
// wrap remaining code in here
}
```

src/ImageSharp/Processing/Normalization/HistogramEqualizationProcessor.cs
```diff
@@ -0,0 +1,120 @@
...
+            // creating the lookup table: subtracting cdf min, so we do not need to do that inside the for loop
+            for (int i = 0; i < histogram.Length; i++)
+            {
+                cdf[i] = cdf[i] - cdfMin;
+            }
+
+            // apply the cdf to each pixel of the image
+            double numberOfPixelsMinusCdfMin = (double)(numberOfPixels - cdfMin);
+            int luminanceLevelsMinusOne = luminanceLevels - 1;
+            for (int y = 0; y < source.Height; y++)
```
tocsoft: as above, as this doesn't actually depend on the relative position of the pixel then this can be done is a single loop over all the pixels.

src/ImageSharp/Processing/Normalization/HistogramEqualizationProcessor.cs
brianpopow: ah nice, i did not know that

src/ImageSharp/Processing/Normalization/HistogramEqualizationProcessor.cs
brianpopow: i have changed that

src/ImageSharp/Processing/Normalization/HistogramEqualizationProcessor.cs
brianpopow: i am using now vector4, but still i need to determine how many gray levels there are with the pixel type, because i need that to create the histogram.

tocsoft: Looking at the code the number of levels looks like its something that the user should be in control of. We should allow it to be passed in to the constructor of the processor but have a default value of 65536  this will give a good high quality default plus allow users to adjust as they desire to tweaking the output.

brianpopow: it is important for the algorithm to know how many possible grey levels there are. If the input image only has 256 possible grey levels, creating a histogram with 65536 would not change the quality of the outcome. 
The algorithm just maps each input grey level to a new output grey level with the goal to use all possible grey values equally. For example the picture i have attached to this PR has only grey values of 120 to 200. After the equalization it uses the the complete range from 0 to 256, but the total number of different grey levels will not change after the equalization. 

here is the histogram of the input image: 
before_eq

and after the equalization:
after_eq

tocsoft: So it sounds like my suggestion will allow that and will work fine it will just over allocate on smaller pixel sizes but will allow users with higher quality pixel definitions to fully make use of the processor. 

Ultimately we can't have processors that have logic dependent on the pixel type. All processors in the core of imagesharp have to have to work consistently no matter which pixel-type is used this includes custom user defined pixel types that might not even map to any sort of bit depth.... they could be using some high precision floating point value for each channel that doesn't map to any of those grayscale depths you are expecting thus we need to allow the user to influence the mapping.

brianpopow: ok, i understand now your reasoning why you do not want pixel type related code in the processor. 

The luminance levels is now a parameter of the constructor as you suggested.

brianpopow:
@JimBobSquarePants would think an adaptive histogram equalization could be a useful feature for imagesharp? I think it can produce much better results then the global histogram equalization in some cases.  

I have hacked together a first rough version in in ImageSharp. Here is an example result where is works pretty good: 

adaptive_side_by_side

(note: click on the image to see it in full size, the preview does not look correct)

Let me know, if you think that this is something you would like to have for imagesharp, then i would look further into it.

JimBobSquarePants:
@brianpopow Definitely! I'd love to get a common set of histogram based operation put together.

PR #654, opened 2018-07-10
Improve CQ6

iamcarbon:
### Prerequisites

### Description

- Removes a few virtual calls through sealing and the removal of the IRgbWorkingSpace abstraction
- Updates Colourful in our benchmarks to 2.0.0 (simplified dependencies + allocation free)


iamcarbon:
@JimBobSquarePants It makes sense to get these remaining colorspaces into Colourful using double precision math and compare benchmarks. That would be a good starting point to consider replacing our implementation.

JimBobSquarePants:
We're still 4x faster currently. Here's `CieXyz` to `CieLab`.

```
               Method |     Mean |     Error |    StdDev |   Median | Scaled |
--------------------- |---------:|----------:|----------:|---------:|-------:|
  'Colourful Convert' | 862.0 ns | 16.673 ns | 16.375 ns | 857.9 ns |   1.00 |
 'ImageSharp Convert' | 233.0 ns |  4.665 ns |  7.533 ns | 229.3 ns |   0.27 |
```

iamcarbon:
@JimBobSquarePants And that's before any SIMD optimizations!

PR #660, opened 2018-07-22
Replace IBuffer<T> with IMemoryOwner<T> + publish Image.WrapMemory()

### LINKED ISSUE #88: Passing Unmanaged Pointer as Pixels buffer in ImageSharp

antonfirsov:
### Prerequisites

### Problems solved by this PR
- `IBuffer<T>` used to mix responsibilities for owning vs. consuming buffers
- The main responsibility of `IBuffer<T>` (providing buffers that own a `Memory<T>`) is actually a duplicate of `System.Buffers.IMemoryOwner<T>`
- `Image.WrapMemory<T>` has not yet been validated by a real-life use case

### PR details
- `IBuffer<T>` has been replaced with the following:
  - `System.Buffers.IMemoryOwner<T>` for of the 99% usages: it represents a "buffer" which is able to provide `Memory<T>`
  - ~`SixLabors.Memory.BufferManager<T>`~ `SixLabors.Memory.MemorySource<T>`: an internal wrapper to manage the OWNED vs CONSUMED logic for ImageSharp. `SwapOrCopyContent` is now implemented here.
- Made `Image.WrapMemory<T>` public
- [Unit test](https://github.com/SixLabors/ImageSharp/blob/af/memory-bridge/tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs#L66) demonstrating how to draw on a `System.Drawing.Bitmap` instance with ImageSharp!

Fixes #88.


tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
```diff
@@ -12,12 +19,72 @@ public partial class ImageTests
...
             [Fact]
-            public void ConsumedBuffer_IsMemoryOwner_ReturnsFalse()
+            public void WrapSystemDrawingBitmap()
             {
-                var memory = new Memory<int>(new int[55]);
-                var buffer = new ConsumedBuffer<int>(memory);
-                Assert.False(buffer.IsMemoryOwner);
+                using (var bmp = new Bitmap(51, 23))
+                {
+                    using (var memoryManager = new BitmapMemoryManager(bmp))
```
tocsoft: 👍 this is awesome..

Would love to be able to have a way for not needing to have the 2 `using`s here it would be nice if when constructing a wrapped `Image<TPixel>` we can tell it to own the memory manager directly and thus dispose of it when its disposed. That would then allow us to more easily develop an extension method directly against a `S.D.Image`.

antonfirsov: We need an `Image.WrapMemory<T>(IMemoryOwner<T> memoryOwner)` overload then. The [guideline tells](https://gist.github.com/GrabYourPitchforks/4c3e1935fd4d9fa2831dbfcab35dffc6#rule-8-if-you-have-an-imemoryownert-parameter-in-your-api-surface-you-are-accepting-ownership-of-that-instance) that the memory ownership will be transferred to `Image<T>` in that case, so it will be responsible for disposing the image.

Need to rethink the logic inside `BufferManager<T>` to correctly support this case in [`SwapOrCopyContent`](https://github.com/SixLabors/ImageSharp/blob/f9dbe41949921bd36fbb958dc7fb6a8c39f24280/src/ImageSharp/Memory/BufferManager.cs#L47).

antonfirsov:
@tocsoft just implemented your suggestion, check out [the new sample test case](https://github.com/SixLabors/ImageSharp/blob/390691f20905ba34d7feb7d32c43ad180e9f78d2/tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs#L131).

This should allow implementing a powerful interop library.
@JimBobSquarePants @tocsoft be aware that we are [throwing `InvalidOperationException`](https://github.com/SixLabors/ImageSharp/blob/390691f20905ba34d7feb7d32c43ad180e9f78d2/src/ImageSharp/Memory/MemorySource.cs#L76) when someone tries to resize an image instance that wraps an existing memory. (Which is kinda fine IMO)

PR #664, opened 2018-07-27
WIP Colorspace Transforms

JimBobSquarePants:
### Prerequisites

### Description
I've been slowly but surely refactoring the colorspace transformation code to make it usable and performant. This PR reintroduces the transform code into the codebase.

- Streamlined struct layout to better fit memory and removed boxing.
- Made structs readonly
- Added `Span<T>` based bulk transforms.
- General style cleanup.
- Added better tests.


JimBobSquarePants:
Still tons of work to do here to improve tests coverage. API itself is sound though I believe.

JimBobSquarePants:
>1. IColorConversion<TFrom, TResult> and implementations
No color converter is used through the IColorConversion<TFrom, TResult> interface, which raises the question whether it's a meaningful abstraction. Can we imagine an actual use case, where a user actually needs an abstract TFrom -> TResult colorspace converter?
>- If not: we should remove the interface, and turn all converters into one of the following: 
Static classes taking the whitepoint as a parameter
>- Structs taking whitepoint at construction time
If yes: Bulk conversion should be part of the API to avoid per-pixel virtual calls

`IColorConversion<TFrom, TResult>`  is used to enforce the conversion rule only. It's never called directly so there's no virtual calls to avoid. We could simply remove it with the individual classes static however they are currently only initialized once as the converter is initialized.

The bulk conversion calls are currently part of the `ColorSpaceConverter` class API since some of the per-pixel methods require adaptation. 

>2. IChromaticAdaptation and VonKriesChromaticAdaptation
Same as 1.: Do we need the abstraction? If yes, we need to move bulk operations here.

Yes, we need the abstraction, there's potentially more adaptation classes to add. I don't see how we could add the bulk transforms to that interface since it is called as part of a per-pixel conversion process. 

>3. The code is full of on-the-fly per-pixel converter class allocations now.
We should avoid (re)instantiation of converters when doing bulk conversion (see comments in 1.)
As far as I'm aware everything is only allocated once, when `ColorSpaceConverter` is initialized. 

I'm considering making `ColorSpaceConverter` immutable and passing an options class to the constructor. That would remove all of the checking logic that takes place internally and simplify things. 

I'm not sure about doing anything more dramatic than that (adding the adapter or whitepoint as a parameter to conversion etc).

Bear in mind we're already 4x faster than the Colorful library this API is based on.

JimBobSquarePants:
@antonfirsov  Current state of play - CieXyz To CieLab following latest immutability changes.

```
// * Summary *

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i7-6600U CPU 2.60GHz (Skylake), 1 CPU, 4 logical and 2 physical cores
Frequency=2742189 Hz, Resolution=364.6722 ns, Timer=TSC
.NET Core SDK=2.1.300
  [Host]     : .NET Core 2.0.9 (CoreCLR 4.6.26614.01, CoreFX 4.6.26614.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.0.9 (CoreCLR 4.6.26614.01, CoreFX 4.6.26614.01), 64bit RyuJIT


               Method |       Mean |     Error |   StdDev | Scaled | ScaledSD |
--------------------- |-----------:|----------:|---------:|-------:|---------:|
  'Colourful Convert' | 1,090.0 ns | 24.385 ns | 71.52 ns |   1.00 |     0.00 |
 'ImageSharp Convert' |   199.5 ns |  4.735 ns | 13.81 ns |   0.18 |     0.02 |

// * Hints *
Outliers
  ColorspaceCieXyzToCieLabConvert.'Colourful Convert': Default  -> 1 outlier  was  removed
  ColorspaceCieXyzToCieLabConvert.'ImageSharp Convert': Default -> 2 outliers were removed

// * Legends *
  Mean     : Arithmetic mean of all measurements
  Error    : Half of 99.9% confidence interval
  StdDev   : Standard deviation of all measurements
  Scaled   : Mean(CurrentBenchmark) / Mean(BaselineBenchmark)
  ScaledSD : Standard deviation of ratio of distribution of [CurrentBenchmark] and [BaselineBenchmark]
  1 ns     : 1 Nanosecond (0.000000001 sec)

// ***** BenchmarkRunner: End *****
// * Artifacts cleanup *
```

antonfirsov:
@JimBobSquarePants it's unfortunate that I can only comment on your changes here. If we want to publish the API, we need a full review here including the work of JBildstein.

Gonna try to be more specific:
- [This allocation](https://github.com/SixLabors/ImageSharp/blob/0d2788752c137e9880d697fc3c36736f8ad1d2a0/src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.CieLab.cs#L173) is performed for each pixel, when called through [the bulk conversion method](https://github.com/SixLabors/ImageSharp/blob/0d2788752c137e9880d697fc3c36736f8ad1d2a0/src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.CieLab.cs#L194). We can (and should) avoid this.
- It's possible to execute [these comparisons](https://github.com/SixLabors/ImageSharp/blob/0d2788752c137e9880d697fc3c36736f8ad1d2a0/src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.CieLab.cs#L168) only once per bulk-conversion
- The [validation](https://github.com/SixLabors/ImageSharp/blob/0d2788752c137e9880d697fc3c36736f8ad1d2a0/src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.Adapt.cs#L106) and the white point equality comparison should be also performed only once, when we are doing bulk conversion.
- ↑↑↑ Same for all color spaces!
- If we do the previous refactors, it should enable `Adapt(span, span)` and `ChromaticAdaptation.Transform(span, span)` overloads

Generally speaking, what we need is to make sure that the bulk conversion API-s are real optimization enablers by:
- Making sure certain operations are performed only once if we do bulk conversion (or such an optimization is possible in the future without API change)
- Making sure the API and the design is prepared for adapting SIMD optimizations in the future (eg. converting multiple colors in a single step with AVX2 etc.)

I still have the feeling that an unused abstraction (`IColorConversion<TFrom, TResult>`) is a code smell, and we should get rid of it, while real abstractions should include bulk variants. (Or at least we should validate, it's possible to introduce them without public API changes.)

JimBobSquarePants:
Good spot regarding the allocation. I'll remove that. 

My concern with moving the comparison and validation code out of the loop is that there is then no way to enforce that the output is the expected one when a span of CieXyz structs with varying whitepoint values is passed to the bulk method. 

I'd love to remove the branching and allow SIMD optimization but I can't, hand-on-heart state that we won't come up against that scenario.

Regarding the validation, I'm wondering whether we should throw at all, we could simple return the input color.

I'm gonna remove the `IColorConversion<TFrom, TResult>` interface. You're right, it's smell.

antonfirsov:
> My concern with moving the comparison and validation code out of the loop is that there is then no way to enforce that the output is the expected one when a span of CieXyz structs with varying whitepoint values is passed to the bulk method.

Correct me if I'm missing something, but it looks to me that it's not possible to have varying white points here. The result of the expression `!this.whitePoint.Equals(this.targetLabWhitePoint) && this.performChromaticAdaptation` only depends on parameters known at construction time. We can even cache this bool inside the constructor!

JimBobSquarePants:
Nope, you’re absolutely correct. Wood for the trees...

antonfirsov:
@JimBobSquarePants Now I can't see any technical issue about adding the following bulk method to `IChromaticAdaptation`:

```C#
void Transform(Span<CieXyz> sourceColors, Span<CieXyz> destColors, CieXyz sourceWhitePoint, in CieXyz targetWhitePoint);
```

This way we can enable optimized bulk-conversion, but this may lead to API changes (just as I expected), because we need a `MemoryAllocator<T>` for the creation of temporary buffers. We can pass a `Configuration` instance to `ColorSpaceConverter` for this.

Do you want me to push these changes?

JimBobSquarePants:
Yeah, if you have time rock on!

antonfirsov:
Accidentally pushed Guard / DebugGuard changes here :\

JimBobSquarePants:
@antonfirsov I've been revisiting this PR and the method you suggested and I cannot see how adding it will give use any advantage. 

As it stands we loop through the span and perform conversion + adaptation in a single loop. If we introduce that method and attempt to integrate it into a bulk conversion we have to loop through and create the unadapted destination span, then loop through again to create the adapted version.

src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.cs
```diff
@@ -2,100 +2,59 @@
...
-        /// The default whitepoint used for converting to CieLab
-        /// </summary>
-        public static readonly CieXyz DefaultWhitePoint = Illuminants.D65;
-
-        private Matrix4x4 transformationMatrix;
-
-        private CieXyzAndLmsConverter cachedCieXyzAndLmsConverter;
+        // Options.
+        private Matrix4x4 lmsAdaptationMatrix;
+        private CieXyz whitePoint;
```
antonfirsov: All white point members could be readonly!

JimBobSquarePants: Thought I'd got these... Will update.

JimBobSquarePants:
>I think I fail to understand this. But if you mean that it's an issue to create a temporary buffer to call the bulk variant of IChromaticAdaptation.Adapt(...) before calling bulk conversion on the adapted data -it shouldn't be. A MemoryAllocator could quickly allocate it, and with the bulk optimization, the execution will be faster.

@antonfirsov What I mean it that for many, if not most conversions, the Adapt method comes after the initial conversion which takes place inside the loop. 

In pseudo code.
``` c#
(for var i = 0 etc){
 var d = ToDWithAdapt(source[i]);
}
```

In order to do the actual optimization we'd have to stop calling the single methods and inline their code. That dramatically decreases our code reuse and increases the tests we need to write. (I already have to write so, so many conversion tests). Possible of course but so much work...

Once inlined I don't think we'll need an allocater so I'll add the method to the interface now and we can add a chore optimization task to the tracker.

antonfirsov:
I guess we mean the same: Initial conversion could be done in a bulk way into a temporary buffer.

With `ColorSpaceConverterOptions` now we have the flexibility to add this later. I also consider it a time consuming, and low priority task, I just wanted the API to be future proof.

antonfirsov:
@JimBobSquarePants hope you are happy with the changes I just pushed.

src/ImageSharp/ColorSpaces/Conversion/Implementation/SRgbCompanding.cs
```diff
@@ -14,7 +14,7 @@ namespace SixLabors.ImageSharp.ColorSpaces.Conversion.Implementation.RgbColorSap
     /// <see href="http://www.brucelindbloom.com/index.html?Eqn_RGB_to_XYZ.html"/>
     /// <see href="http://www.brucelindbloom.com/index.html?Eqn_XYZ_to_RGB.html"/>
     /// </remarks>
-    internal class SRgbCompanding : ICompanding
+    public sealed class SRgbCompanding : ICompanding
```
antonfirsov: We have the same code in `Vector4Extensions`. Might make sense to have a unified & expressive common API.

src/ImageSharp/ColorSpaces/Conversion/ICompanding.cs
```diff
@@ -1,16 +1,14 @@
...
-namespace SixLabors.ImageSharp.ColorSpaces
+namespace SixLabors.ImageSharp.ColorSpaces.Conversion.Implementation
 {
     /// <summary>
     /// Pair of companding functions for <see cref="RgbWorkingSpace"/>.
     /// Used for conversion to <see cref="CieXyz"/> and backwards.
     /// See also: <seealso cref="RgbWorkingSpace.Companding"/>
     /// </summary>
-    internal interface ICompanding
+    public interface ICompanding
```
antonfirsov: I wonder why do we have interfaces which are not used through the abstraction. I think they shouldn't be there, and the methods should be moved to static utilities. It's actually **very** expensive to use virtual calls to convert only one `float` value.

JimBobSquarePants:
Right... let’s get this merged.

PR #665, opened 2018-08-04
Consume memory primitives from SixLabors.Core

### LINKED ISSUE #619: Migrate memory primitives to SixLabors.Core
A reminder that this should be done for 1.0.0-RC1!

TODO:
- [x] The non-ownership case should be refactored out of `IBuffer<T>`.
- [x] Consider replacing `IBuffer<T>` with `System.Buffers.IMemoryOwner<T>`
- [ ] Move the classes

antonfirsov:
### Prerequisites

### Description
- Updated `SixLabors.Core` package reference
- Removed duplicates of common memory primitives which are now defined in `SixLabors.Core`
- Move ImageSharp-specific memory primitives and utils to the `SixLabors.ImaSharp.Memory` namespace
- Update + cleanup tests and benchmarks

Fixes #619.


src/ImageSharp/Memory/BufferArea{T}.cs
```diff
@@ -120,7 +123,7 @@ public BufferArea<T> GetSubArea(int x, int y, int width, int height)
         public BufferArea<T> GetSubArea(Rectangle rectangle)
         {
             ImageSharp.DebugGuard.MustBeLessThanOrEqualTo(rectangle.Width, this.Rectangle.Width, nameof(rectangle));
-            DebugGuard.MustBeLessThanOrEqualTo(rectangle.Height, this.Rectangle.Height, nameof(rectangle));
+            SixLabors.DebugGuard.MustBeLessThanOrEqualTo(rectangle.Height, this.Rectangle.Height, nameof(rectangle));
 
             int x = this.Rectangle.X + rectangle.X;
```
JimBobSquarePants: We're mixing two `DebugGuard` methods now. We should ditch the ImageSharp one.

antonfirsov: Good catch! Having common internals is just too dangerous, we should probably drop `[InternalsVisibleTo(...)]` in SL.Core to avoid these issues.

JimBobSquarePants: Yeah, though I was wondering whether we should just make them public and avoid duplication? They're useful and well tested.

antonfirsov: One can accidentally include and use such internals with R# (been there). I would rather share them through submodules in the future. For now it's just safer to hide them IMO.

JimBobSquarePants: Hide them then. We can investigate private packages like they use for ASP Core in the future

PR #671, opened 2018-08-07
Use span directly for DetectFormat

JimBobSquarePants:
### Prerequisites

### Description
@davidfowl [spotted](https://github.com/SixLabors/ImageSharp/pull/618#pullrequestreview-143844759) an allocation we don't need in `Image.DetectFormat(Configuration config, ReadOnlySpan<byte> data)` so this fixes that.


JimBobSquarePants:
These code coverage results really don't count for much do they? Why would `ProjectiveTransformHelper.cs` coverage decrease?

PR #673, opened 2018-08-07
Feature: adaptive histogram equalization

brianpopow:
### Prerequisites

### Description

This PR adds a sliding window implementation of the adaptive histogram equalization (AHE). It also add clipping the histogram, which is supposed to reduce the effect of over-amplifying noise in relatively homogeneous regions (which happens with AHE).

Here is an example input: 

ahe_example_input

and the output:

ahe_example_output_jpg

(click on the image, at least for me the preview was not displaying the image correctly)

This algorithm is computational very demanding, especially with 65536 grey levels. 
To be honest, im a bit disappointed about the performance of this implementation with 16 bit greyscale images. It can take up to 30 seconds with an image of 1000x1000 pixels on my machine.
For 256 greylevels it would take ~500 ms.

I don't know if this is good enough for ImageSharp, so i was unsure if i should open the PR in the first place. AHE can produce some pretty nice result, so i though i give it a shot anyways and see what you guys think about it.


src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistEqualizationProcessor.cs
```diff
@@ -0,0 +1,230 @@
...
+            Span<TPixel> pixels = source.GetPixelSpan();
+
+            int pixelsInGrid = this.GridSize * this.GridSize;
+            int halfGridSize = this.GridSize / 2;
+            using (Buffer2D<TPixel> targetPixels = configuration.MemoryAllocator.Allocate2D<TPixel>(source.Width, source.Height))
+            {
+                ParallelFor.WithConfiguration(
+                    0,
+                    source.Width,
+                    Configuration.Default,
```
antonfirsov: Y0ou should use the method's `configuration` argument instead.

brianpopow: fixed

JimBobSquarePants:
>It can take up to 30 seconds with an image of 1000x1000 pixels on my machine.
For 256 greylevels it would take ~500 ms.

I'd like to see if we can do anything to improve this as the output is incredible!

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistEqualizationProcessor.cs
```diff
@@ -0,0 +1,230 @@
...
+            }
+
+            // Special cases for the left and the right border where GetPixelRowSpan can not be used
+            if (x < 0)
+            {
+                var rowPixels = new TPixel[gridSize];
+                int idx = 0;
+                for (int dx = x; dx < x + gridSize; dx++)
+                {
+                    rowPixels[idx] = source[Math.Abs(dx), y];
```
JimBobSquarePants: I feel like we can possibly do some offsetting and move the `Abs` out of the loop here.

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistEqualizationProcessor.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        /// <summary>
+        /// Adds a row of grey values to the histogram.
+        /// </summary>
+        /// <param name="greyValues">The grey values to add</param>
+        /// <param name="histogram">The histogram</param>
+        /// <param name="luminanceLevels">The number of different luminance levels.</param>
+        /// <returns>The maximum index where a value was changed.</returns>
+        private int AddPixelsTooHistogram(Span<TPixel> greyValues, Span<int> histogram, int luminanceLevels)
```
JimBobSquarePants: `AddPixelsToHistogram`

brianpopow:
> I'd like to see if we can do anything to improve this as the output is incredible!

@JimBobSquarePants: I am glad you like. I will try to improve it, but i think the main reason why this approach is slow, is that the distribution function needs to be calculated for each pixel and with 65k grey levels, thats a lot to chew on. 

There is a different approach to this: The image would be split into n tiles dependent on how big the gridsize is chosen. For each tile the cdf is pre-calculated. Now the final grey level would be calculated by interpolating between 4 adjacent tiles. This approach should be in theory much faster.
I will give this a try over the next week and see how it goes.

JimBobSquarePants:
Great! Looking forward to seeing what you come up with 👍

JimBobSquarePants:
That sounds much more promising!

It looks to me like the sum of your samples is off. When interpolating the sum of your samples should equal 1. This maintains the correct brightness. I’d choose clamping to deal with edge pixels also.

brianpopow:
@JimBobSquarePants just a quick update: i did not had much time last week to work on this. I hope i find some time this weekend to continue with it.

JimBobSquarePants:
@brianpopow no worries, looking forward to seeing this complete.

JimBobSquarePants:
@brianpopow You're getting bitten by Stylecop 👮 

```
Processing/Processors/Normalization/HistogramEqualizationProcessor.cs(23,26): error CS1572: XML comment has a param tag for 'tiles', but there is no parameter by that name [/home/travis/build/SixLabors/ImageSharp/src/ImageSharp/ImageSharp.csproj]
```

brianpopow:
@JimBobSquarePants:
i am sorry that it took me so long, i could not find much time to work on this. I think im done now and its ready for a review.

I kept the sliding window approach in even if its slow, because it can produce better quality results than the tile interpolation in some cases.
For people who do not care so much about speed, i think its a good option.

For the timings it looks now like this on my machine for a 1000x797 pixel image (using 10 tiles):

for 256 grey levels:
- ahe tile interpolation: 72 ms
- ahe sliding window: 661 ms

for 65536 grey levels:
- ahe tile interpolation: 84 ms
- ahe sliding window: 25756 ms

I could not come up with any useful unit tests so far. Let me know if you have any suggestions.

JimBobSquarePants:
@brianpopow Thanks for continuing to work on this. 

Would it be possible to share the output side-by-side for the two approaches. I'd like to see if the quality difference is due to a bug or whether is an algorithmic limitation.

brianpopow:
sure, here are the results from the different approaches. The input image is the one i posted first as a example image.: 

Sliding winow: 
ahe_sw_30_tiles

tile  interpolation:
ahe_tile_interpolation_30_tiles

Why i think its not a bug, but a limitation of the algorithm is that the output from opencv looks pretty much the same:

opencvclahe

I think the difference between the opencv result and mine it the clipping of the histogram. I think they do the clipping slightly different than i do. 

i used the following options: 
- LuminanceLevels = 65536,
- Tiles = 30,
- ClipHistogram = true

brianpopow:
I wanted to clarify, that this blocky effect will only be visible, if a large tile number is chosen. In the above case a tile number of 30 would lead to tiles with a pixel width of 84, which is very small. Those small tiles work well with the sliding window approach, but with the tile interpolation, it has this blocky effect.

A more reasonable choice would be 12 tiles for example. Here is how that will look for the tile interpolation mode:
ahe_12_tiles

JimBobSquarePants:
@brianpopow Thanks for this, the output looks great and your findings make sense. I'll do a deep dive review on this later on today. 

As far as tests go. Any algorithmic result should be tested and we can add reference images to the submodule repo to ensure there are no regression issues when applying any future optimization.

JimBobSquarePants:
I will review this, I promise!

brianpopow:
@JimBobSquarePants: dont worry, take as much time as you want.

JimBobSquarePants:
@brianpopow I've started reviewing this and have spotted a few places I should be able to ramp up performance. I'll push directly to your PR once finished.

JimBobSquarePants:
Hi @brianpopow 

As you'll be able to see from your commit logs I've been doing some performance work on the various processors. So far I've more than doubled the speed of both the global and tiled processors which is great. 😄 

I've got some questions on the sliding window implementation though if I may?

Is there a particular reason you go by column, then by row? By working the way it currently is we are working against the way the image is laid out in memory which will be slowing everything down, additionally this prevents us from using our normal optimized APIs for grabbing a row slice and operating on that. 

Would you be able to change that? I could then do further optimizations to improve the performance.

brianpopow:
@JimBobSquarePants: Thats awesome, great job!

I will change the sliding window to go by row instead of columns.

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistEqualizationSWProcessor.cs
```diff
@@ -51,8 +51,9 @@ protected override void OnFrameApply(ImageFrame<TPixel> source, Rectangle source
             Span<TPixel> pixels = source.GetPixelSpan();
 
             var parallelOptions = new ParallelOptions() { MaxDegreeOfParallelism = configuration.MaxDegreeOfParallelism };
-            int tileHeight = source.Height / this.Tiles;
-            int pixeInTile = tileHeight * tileHeight;
+            int tileWidth = source.Width / this.Tiles;
+            int tileHeight = tileWidth;
```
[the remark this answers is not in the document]
brianpopow: Yes they are in this implementation. The OpenCV implementation provides the option to choose the number of tiles in X direction and in Y-Direction, but i cannot see a real usecase where this could be beneficial.

JimBobSquarePants:
>I am tempted to remove this optimization, because im unsure if its working correctly. For 256 grey values this would not make much of a difference, but for 65536 it would add like 25% time on top.

I'd try that and see if it makes a difference. We can always look for optimization opportunities once we have the correct output. This all assumes of course that the original output before our refactor was correct?

brianpopow:
@JimBobSquarePants: I have removed this keeping track of the histogram position now.
We need to change the reference image, because it was created with this possible wrong optimization step.

I have compared the left to right vs. top to bottom approach without the keeping track of the histogram position step. The difference is still not 0, but much better then before: Total difference: 0,0976%.
In the corners of the image there are very small differences.

brianpopow:
just for reference here are the difference for bottom/top vs left/right:

output_left_right

output_top_bottom

JimBobSquarePants:
@brianpopow Finally getting round to having another look at this to get it finished. 

Two questions:

1. Is it a requirement of the algorithm to mirror edges? We can optimize things quite a lot if we can remove the reflection.
2. Would it be possible to gather the histogram values for each tile by row instead of per column? We could simply slice then copy.

brianpopow:
@JimBobSquarePants:

1. Mirroring the edges is the best way i know to deal with the borders of the image where there is no data. I do not understand exactly what you mean by reusing the edge pixels. Can you point me to an example source code line where this is done in th resizer?

2. We now move the window from left to right. This means when we move one pixel, we need to remove one column from the left and add another on the right. When i was moving the window from top to bottom i could read a row (with the size of the window) when moving one pixel down, but you said this is not a good idea.
Maybe i do not understand exactly what you mean or you see something i do not see here, but when moving the window from left to right we need to read a column when moving the window.

Im sorry that i could not be more helpful here.

JimBobSquarePants:
Hi @brianpopow 

>1\. Mirroring the edges is the best way i know to deal with the borders of the image where there is no data. I do not understand exactly what you mean by reusing the edge pixels. Can you point me to an example source code line where this is done in th resizer?

It's actually pretty simple, instead of mirroring you reuse the `0` and `source.Width` pixels when sampling. So in your current code you have:

```c#
if (x < 0)
{
    x = Math.Abs(x);
}
else if (x >= source.Width)
{
    int diff = x - source.Width;
    x = source.Width - diff - 1;
}
```

In the most simplistic terms you would have

``` c#
if (x < 0)
{
    x = 0;
}
else if (x >= source.Width)
{
    x = x - source.Width -1;
}
```
This yields a different result on the image edges but the algorithm still works. I'm fairly certain we can optimize that example further though if we were going per-row instead of per-column within the tile.

>2\. We now move the window from left to right. This means when we move one pixel, we need to remove one column from the left and add another on the right. When i was moving the window from top to bottom i could read a row (with the size of the window) when moving one pixel down, but you said this is not a good idea.
    Maybe i do not understand exactly what you mean or you see something i do not see here, but when moving the window from left to right we need to read a column when moving the window.

There must be something I am not understanding about the algorithm. 

When I first suggested going top down it was because you were using a parallel for `0 => source.Width` which meant you couldn't slice a row to operate on it. You're now going `0 => source.Height` on the outermost loop now which is good but you're now operating per column within the tile and I don't understand why?

https://github.com/SixLabors/ImageSharp/pull/673/commits/8f19e5edd23f13fd1ddf93b4e795f82e7f1334bb#diff-6ca0f7e3ceae4a5676d95d4de12f7b8bR62

> Im sorry that i could not be more helpful here.

Never say that chum, your work is amazing.

brianpopow:
@JimBobSquarePants:

1. ok i understand now what you mean, but we can not do that. That would over amplify the edge pixels in the border. The edge pixels would be added multiple times to the histogram, worst case is the corners of the image where it would lead adding the same value as often as the window width.

One other option is to just ignore the pixels outside of the image and do not add them. Im not sure, if that a good idea either. I have to try that and see how it looks.

Another option i have found is here: https://digitalcommons.unf.edu/cgi/viewcontent.cgi?referer=https://www.google.de/&httpsredir=1&article=1264&context=etd
See Section *Image Border*, Page 26.
They suggest to keep the window in place in the corners (what call window, they call *contextual region*)
I think that sounds like a good alternative, but im not sure at the moment, if this will be better in performance than the mirroring approach.

2. Moving the sliding window around the image is comparable to other filtering methods, where you apply one filter mask to a region around a pixel. Lets say like a Sobel filter, but here the mask / window / contextual region is a bit bigger (typically something like 64 by 64 pixels). This window needs to be moved over each pixel. It does not matter if its from left to right or from top to bottom, but it needs to be moved over all pixels. 
All pixels which are covered by the window needs to be added to the histogram. To not add all pixels under the window to the histogram again when you move one pixel, you only have to add one column to the right and remove one column from the left (if you move from left to right). 

Here is an example image, maybe that makes it clearer:

slidingwindow-2-figure1-1

If you we move the window from top to bottom, which was the case in the commit you pointed out, we would add one row on the bottom and remove one from the top.

I hope this makes it a bit clearer, let me know if you have still any questions. 

Unfortunately i think i will not find time to continue with this in this week. Maybe i can find some time next week.

JimBobSquarePants:
> ok i understand now what you mean, but we can not do that. That would over amplify the edge pixels in the border. The edge pixels would be added multiple times to the histogram, worst case is the corners of the image where it would lead adding the same value as often as the window width

Reading the linked article I understand now that sampling is more accurate away from the edge border so mirroring seems to be the valid approach.

>If you we move the window from top to bottom, which was the case in the commit you pointed out, we would add one row on the bottom and remove one from the top.

Definitely preferred over columns as we can do `span.Slice(x, range).Copy()` for anything other than edge pixels.

Gonna have a good read through the article. It explains the process very clearly. 👍

brianpopow:
@JimBobSquarePants: ok, so we should keep mirroring and switch back to moving the window from top to bottom? I will do that, if you agree to this.

JimBobSquarePants:
Sorry... on holiday. 

Yeah, grab per row but do a fast and slow path since we only need mirroring on edges.

brianpopow:
@JimBobSquarePants im sorry for no update on this one for so long, i really wanted to finish it, but i cannot find time recently to continue on this, because of too much work currently at my job. I will try to get back on this next weekend.

JimBobSquarePants:
@brianpopow No worries, thanks again for your help!

brianpopow:
@JimBobSquarePants: I finally found some time to continue on this. As discussed, the sliding window is moved from top to bottom again, so we can add a row to the histogram when it moves down one step.

I have changed the mirroring, so it will only be done on the edges now.

JimBobSquarePants:
@brianpopow Excellent! 

Looks like we need to update the reference image again as the output is now different. I'll review this asap. 👍

brianpopow:
@JimBobSquarePants: yes, but lets update the reference image after the review is done.

brianpopow:
@JimBobSquarePants: Great, thank you! This took quite some time, but i still think it was worth it.

PR #678, opened 2018-08-17
Improve CQ

iamcarbon:
### Prerequisites

### Description

- Improves a few error messages
- Replaces == null with is null (== is overloadable, and can be used on Spans)
- Throws NRE directly in trival contexts (avoiding the call overhead + expanded callstack on exception)
- Annotates a few additional fields as readonly

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
@@ -628,7 +628,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
 
             byte[] data = exifProfile?.ToByteArray();
 
-            if (data == null || data.Length == 0)
+            if (data is null || data.Length == 0)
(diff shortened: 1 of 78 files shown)


iamcarbon:
Ready for review.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -550,7 +550,8 @@ private void WriteApplicationHeader(ImageMetaData meta)
         private void WriteDefineHuffmanTables(int componentCount)
         {
             // Table identifiers.
-            byte[] headers = { 0x00, 0x10, 0x01, 0x11 };
+            Span<byte> headers = stackalloc byte[] { 0x00, 0x10, 0x01, 0x11 };
```
JimBobSquarePants: We should keep an eye out for more things like this but I suspect you already have it covered 👍

PR #679, opened 2018-08-20
Split PixelBlendMode into PixelColorBlendingMode and PixelAlphaCompositionMode

vpenades:
### Prerequisites

### Description

This continues the process started with #535 

As described in the issue, GraphicsOptions now has two separated modes for configuring Color Blending and Alpha Composition.

PixelAlphaCompositionMode has "SrcOver" as first value to ensure that a default GraphicsOptions structure will always be SrcOver-Normal which is the mode everybody expects by default.

Now there's a lot more possible combinations, so creating tests for all of them might take a while. Specially considering the test images are in a separate repository.

### DIFF
--- a/src/ImageSharp.Drawing/Processing/Processors/Drawing/FillProcessor.cs
+++ b/src/ImageSharp.Drawing/Processing/Processors/Drawing/FillProcessor.cs
@@ -104,12 +104,33 @@ namespace SixLabors.ImageSharp.Processing.Processors.Drawing
         {
             solidBrush = this.brush as SolidBrush<TPixel>;
 
-            return solidBrush != null
-                   && ((this.options.BlenderMode == PixelBlenderMode.Normal && this.options.BlendPercentage == 1f
-                                                                            && solidBrush.Color.ToVector4().W == 1f)
-                       || (this.options.BlenderMode == PixelBlenderMode.Over && this.options.BlendPercentage == 1f
-                                                                             && solidBrush.Color.ToVector4().W == 1f)
-                       || (this.options.BlenderMode == PixelBlenderMode.Src));
+            if (solidBrush == null)
+            {
+                return false;
+            }
+
+            if (this.options.ColorBlendingMode != PixelColorBlendingMode.Normal)
+            {
+                return false;
+            }
+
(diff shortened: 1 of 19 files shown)


src/ImageSharp.Drawing/Processing/Processors/Drawing/FillProcessor.cs
```diff
@@ -102,14 +102,35 @@ protected override void OnFrameApply(ImageFrame<TPixel> source, Rectangle source
...
-                                                                             && solidBrush.Color.ToVector4().W == 1f)
-                       || (this.options.BlenderMode == PixelBlenderMode.Src));
+            solidBrush = this.brush as SolidBrush<TPixel>;
+
+            if (solidBrush == null)
+            {
+                return false;
+            }
+
+            if (this.options.ColorBlendingMode != PixelColorBlendingMode.Normal)
```
JimBobSquarePants: Much more readable! 👍

src/ImageSharp.Drawing/Processing/TextGraphicsOptions.cs
```diff
@@ -32,7 +32,7 @@ public struct TextGraphicsOptions
 
         private float? dpiY;
 
-        private PixelBlenderMode blenderMode;
+        private PixelColorBlendingMode blenderMode;
```
tocsoft: we should add the alpha blending option in here too.

vpenades: @tocsoft I'll take a look

vpenades:
@JimBobSquarePants @tocsoft Some of the coverage tests are failing because the DrawImageProcessor methods I added are not being called from anywere because there were some missing methods in DrawImageExtensions.

And looking at [DrawImageExtensions](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp.Drawing/Processing/DrawImageExtensions.cs) I've noticed the arguments ordering is a bit of a mess....

Some methods have opacity before blending, and others the opposite way.

Should I rearrange the arguments so opacity is always the last?

- float opacity
- PixelColorBlendingMode color, float opacity
- PixelColorBlendingMode color,PixelAlphaCompositionMode alpha, float opacity

JimBobSquarePants:
Nice coverage!

PR #686, opened 2018-08-27
Ext.DrawImage generalization to support multiple formats. (Experimental)

vpenades:
With this pull request I wanted to achieve:

- Allow to draw images with a format different than the destination image (API breaking change)
   now you can do:   `image<Rgba>.Mutate(x => x.DrawImage( Image<argb> ) );`

About the implementation: Given that image drawing works with rows, and before drawing, both source and destination rows are converted to Vector4, the whole operation is done at no cost. This allows drawing images with different formats, without requiring a conversion.

- Faster rendering: for image drawing, the "amount" opacity was expanded to a row, which consumed memory, etc. I've made an alternative path that passes the amount directly (the old one is also there)

- Refactored PixelBlender class to move the bulk of the previously generated code to the main class. With this change a lot less code is generated.

### Prerequisites


src/ImageSharp/PixelFormats/PixelBlenders/DefaultPixelBlenders.Generated.tt
```diff
@@ -80,15 +80,15 @@ namespace SixLabors.ImageSharp.PixelFormats.PixelBlenders
...
+                    return PorterDuffFunctions.<#=blender_composer#>(background, source, amount.Clamp(0,1));
                 }
 
                 /// <inheritdoc />
                 protected override void BlendFunction(Span<Vector4> destination, ReadOnlySpan<Vector4> background, ReadOnlySpan<Vector4> source, float amount)                
                 {
                     for (int i = 0; i < destination.Length; i++)
                     {
-                        destination[i] = PorterDuffFunctions.<#=blender_composer#>(background[i], source[i], amount);
+                        destination[i] = PorterDuffFunctions.<#=blender_composer#>(background[i], source[i], amount.Clamp(0,1));
```
tocsoft: we should clamp outside the loop

vpenades: Indeed....

vpenades:
Coverage test fails again? WTF, it was fine yesterday

JimBobSquarePants:
@vpenades Coverage changes aren't failing, there's differences against the reference files.

JimBobSquarePants:
@vpenades Can you have a look at the generated output from these changes compared the master? 

We've got differences appearing on NET 4.7.1 and NET 4.6.2 in 64 bit mode.
https://ci.appveyor.com/project/six-labors/imagesharp/build/1.0.0-PullRequest00686001835

It's likely minor but best to cast your eye over to ensure nothing has gone wrong. If you're happy with the output please open a new PR against our reference images repository.

vpenades:
On it...

vpenades:
Finally!

src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor.cs
```diff
@@ -139,27 +139,42 @@ protected override void OnFrameApply(ImageFrame<TPixel> source, Rectangle source
...
                                     for (float x = scanStart; x < startX + 1; x += subpixelFraction)
                                     {
-                                        scanline[startX] += subpixelFractionPoint;
-                                        scanlineDirty = true;
-                                    }
+                                        scanline[startX] += subpixelFractionPoint;
+                                        scanlineDirty = true;
+                                    }
+
+                                    if (scanline[startX] > 1)
```
JimBobSquarePants: What do these new constraining checks do?

src/ImageSharp.Drawing/Processing/SolidBrush{TPixel}.cs
```diff
@@ -91,9 +91,18 @@ internal override void Apply(Span<float> scanline, int x, int y)
             {
                 Span<TPixel> destinationRow = this.Target.GetPixelRowSpan(y).Slice(x, scanline.Length);
 
-                MemoryAllocator memoryAllocator = this.Target.MemoryAllocator;
+                MemoryAllocator memoryAllocator = this.Target.MemoryAllocator;
+
+#if DEBUG
+                for (int i = 0; i < scanline.Length; i++)
+                {
+                    Guard.MustBeBetweenOrEqualTo(scanline[i], 0, 1, nameof(scanline));
```
JimBobSquarePants: Do we want to throw?

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.Generated.tt
```diff
@@ -27,44 +27,39 @@ namespace SixLabors.ImageSharp.PixelFormats.PixelBlenders
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static Vector4 <#=blender#>Src(Vector4 backdrop, Vector4 source, float opacity)
         {
-            opacity = opacity.Clamp(0, 1);
-            source.W *= opacity;          
+            source.W *= opacity.Clamp(0,1);
```
JimBobSquarePants: Are these called by per-row operations or are they purely for per-pixel operations?

src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawImageProcessor.cs
```diff
@@ -76,23 +78,18 @@ protected override void OnFrameApply(ImageFrame<TPixel> source, Rectangle source
...
+
+            ParallelFor.WithConfiguration(
+                minY,
+                maxY,
+                configuration,
+                y =>
+                    {
+                        Span<TPixelDst> background = source.GetPixelRowSpan(y).Slice(minX, width);
+                        Span<TPixelSrc> foreground = targetImage.GetPixelRowSpan(y - locationY).Slice(targetX, width);
+                        blender.Blend<TPixelSrc>(memoryAllocator, background, background, foreground, this.Opacity);
```
JimBobSquarePants: I couldn't find the callee `Blend` method in the diff but it's imperative now that if we are mixing pixel formats and conversion to/from Vector4 needs to used the scaled conversion methods.

src/ImageSharp/PixelFormats/PixelBlenders/DefaultPixelBlenders.Generated.tt
```diff
@@ -78,32 +80,27 @@ namespace SixLabors.ImageSharp.PixelFormats.PixelBlenders
                 /// <inheritdoc />
                 public override TPixel Blend(TPixel background, TPixel source, float amount)
                 {
-                    return PorterDuffFunctions.<#=blender_composer#>(background, source, amount);
+                    TPixel dest = default;
+                    dest.PackFromVector4(PorterDuffFunctions.<#=blender_composer#>(background.ToVector4(),source.ToVector4(),amount.Clamp(0,1)));
```
JimBobSquarePants: I think this is what I was referring to before. Should be `ToScaledVector4()` with the reverse `PackFromScaledVector4()` on repacking.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.Generated.tt
JimBobSquarePants: If they are called from bulk we shouldn't be clamping per pixel.

src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor.cs
vpenades: When I had the DebugGuard set, FillRegionProcessor was producing some opacities of value 1.02xxx... the algorythms there probably need to be reviewed to ensure that opacity values are always in range. But since I don't fully understand what these functions do, I just added a clamp there.

Now that the clamps are back into the pixels, these checks could be removed...

src/ImageSharp.Drawing/Processing/SolidBrush{TPixel}.cs
vpenades: Ah, that one is no longer neccessary if the pixels do the clamp, I'll remove it now

src/ImageSharp/PixelFormats/PixelBlenders/DefaultPixelBlenders.Generated.tt
vpenades: Ahhh, okey, change it now

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.Generated.tt
vpenades: These are being clamped per pixel. that's why I wanted to remove them and replace them with DebugGuards.

But by using DebugGuards, several tests fail due to out of range values, which prevents the PR to pass the coverage tests,  etc.

So, for now, I preffer to go back to Clamp, and solve any potential out of range issues in future PRs

JimBobSquarePants: I think it'll be best to fix clamping here while we are working on it.

In the internal method `PorterDuffFunctions.NormalSrc(Vector4 backdrop, Vector4 source, float opacity)` the opacity is clamped before being used as a multiplier.

As far as I can see that method is only called by.

`PorterDuffFunctions.NormalSrc<TPixel>(TPixel backdrop, TPixel source, float opacity)`

`NormalSrc.Blend(TPixel background, TPixel source, float amount)`

`NormalSrc.BlendFunction(Span<Vector4> destination, ReadOnlySpan<Vector4> background, ReadOnlySpan<Vector4> source, float amount)`

`NormalSrc.BlendFunction(Span<Vector4> destination, ReadOnlySpan<Vector4> background, ReadOnlySpan<Vector4> source, ReadOnlySpan<float> amount)`

The clamping should take place in the caller methods not the callee, same for the other operations. Only the last method taking the `ReadOnlySpan<float> amount` requires per-pixel clamping of the values but that seems like something we could do with `Vector<float>` in the caller in supported architectures.

JimBobSquarePants: Thankfully all the changes required can be done in the T4 templates.

JimBobSquarePants: I'm happy to make the changes if you like?

vpenades: Sure, go for it!, it's just lunch time here! 😄

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.Generated.cs
```diff
@@ -16,93 +17,93 @@ internal static partial class PorterDuffFunctions
 
 
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
JimBobSquarePants: @vpenades @antonfirsov Confirmed that changing this fixes tests in 64bit on Net Framework 4.7.1 and 4.6.2

We'll need to figure out how to create a reduced sample to report.

vpenades: So that's the thing that's been driving me crazy...

JimBobSquarePants:
I'm happy to merge this now unless anyone objects.

PR #693, opened 2018-09-03
Add derived format info types and allow persistance of palette lengths

JimBobSquarePants:
### Prerequisites

### Description
This PR:

- Adds derived format info types for each of the current image formats.
- Adds `ColorPaletteLength` property to `ImageFrame<TPixel>`
- Enhanced `GifEncoderCore`  to preserve the current pallette lengths when encoding from source gifs.
- Fixes an issue in `OctreeFrameQuantizer<TPixel>` which caused the `GifEncoderCore` type to produce palettes 2x as long as required.
- Additional style cleanup on visited types.


src/ImageSharp/ImageInfo.cs
```diff
@@ -3,26 +3,26 @@
...
 using SixLabors.ImageSharp.MetaData;
+using SixLabors.Primitives;
 
 namespace SixLabors.ImageSharp
 {
     /// <summary>
     /// Contains information about the image including dimensions, pixel type information and additional metadata
     /// </summary>
-    internal sealed class ImageInfo : IImageInfo
+    public abstract class ImageInfo : IImageInfo
```
JimBobSquarePants: I'm tempted to remove this class and have each format Info implement the interface. The internal constructors on those derived classes allow us the freedom to alter the classes but this classes would now be fixed and exposed.

JimBobSquarePants:
@antonfirsov I’d love to see if we an dump the derived types and use an abstraction of the various format requirements if possible. It depends on how much we can effectively share.

JimBobSquarePants:
@antonfirsov I've updated the approach to use @dlemstra 's idea of dictionaries to store properties.

I think this is much better but I'm torn as to whether we should replace the `ImageProperty` struct with specific properties.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -82,6 +86,10 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
             Guard.NotNull(image, nameof(image));
             Guard.NotNull(stream, nameof(stream));
 
+            this.gifMetaData = image.MetaData.GetGifMetaData() ?? new GifMetaData();
```
antonfirsov: @JimBobSquarePants not sure if this was @dlemstra 's original idea. This way we can not share common properties between different formats.

@dlemstra isn't what you meant actually to have entries for individual properties? Eg. `image.MetaData.GetGifColorTableMode()`, `image.Metadata.GlobalColorTableLength` ?

I'm not saying this is the ideal way to go but we should consider pros/contras.

src/ImageSharp/MetaData/ImageFrameMetaData.cs
```diff
@@ -28,32 +31,57 @@ internal ImageFrameMetaData(ImageFrameMetaData other)
...
+        /// <param name="value">
+        /// When this method returns, contains the metadata value associated with the specified key,
+        /// if the key is found; otherwise, the default value for the type of the value parameter.
+        /// This parameter is passed uninitialized.
+        /// </param>
+        /// <returns>
+        /// true if the <see cref="ImageMetaData"/> contains an element with
+        /// the specified key; otherwise, false.
+        /// </returns>
+        public bool TryGetMetaData<T>(string key, out T value)
```
antonfirsov: Can't we find a more type-safe key than `string`?

If decide to stick with the one entry/format solution then I would suggest to use `IImageFormat` as a key (after making all formats singletons).

src/ImageSharp/Formats/Gif/GifEncoder.cs
```diff
@@ -33,7 +33,7 @@ public sealed class GifEncoder : IImageEncoder, IGifEncoderOptions
         /// <summary>
         /// Gets or sets the color table mode: Global or local.
         /// </summary>
-        public GifColorTableMode ColorTableMode { get; set; }
+        public GifColorTableMode? ColorTableMode { get; set; }
```
antonfirsov: This is a breaking API change. If we plan to introduce the same logic for all formats, we should introduce this pattern in all encoders now, even if the implementation is not yet here!

src/ImageSharp/Formats/Gif/GifMetaDataExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Gets the Gif format specific meta data from the image.
+        /// </summary>
+        /// <param name="meta">The image meta data.</param>
+        /// <returns>The <see cref="GifMetaData"/> or null.</returns>
+        public static GifMetaData GetGifMetaData(this ImageMetaData meta)
+        {
+            meta.TryGetMetaData(GifConstants.MetaDataKey, out GifMetaData value);
+            return value;
```
antonfirsov: We should consider returning `new GifMetaData()`, when there is no value present for the key. This would simplify the call site a lot.

JimBobSquarePants: I considered that actually. Probably wise.

src/ImageSharp/Formats/Gif/GifEncoder.cs
JimBobSquarePants: Agreed

src/ImageSharp/MetaData/ImageFrameMetaData.cs
JimBobSquarePants: I wanted to use `IImageFormat` but what instance do you use as a key and how do you access it from the extension methods?

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
JimBobSquarePants: I don't mind not sharing. I avoids potential ambiguity and ensures future format support can't break anything.

src/ImageSharp/Formats/ImageFormatBase{T}.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+
+namespace SixLabors.ImageSharp.Formats
+{
+    /// <summary>
+    /// The base class for all image formats.
+    /// Inheriting classes should implement the singleton pattern by creating a private constructor.
+    /// </summary>
+    /// <typeparam name="T">The type of image format.</typeparam>
+    public abstract class ImageFormatBase<T> : IImageFormat
```
antonfirsov: The drawback of this class is that it pollutes the public API by complicating the class hierarchy with implementation details, with no benefits IMO. The reason is that there is no need for lazy instance creation, because the constructor is trivial and brings no thread safety issues.

Defining a format without it is actually simpler:
```C#
    public sealed class GifFormat : IImageFormat
    {
        private GifFormat() { }

        public string Name => "GIF";

        public string DefaultMimeType => "image/gif";

        public IEnumerable<string> MimeTypes => GifConstants.MimeTypes;
        
        public IEnumerable<string> FileExtensions => GifConstants.FileExtensions;
        
        // The only extra line:
        public static GifFormat Instance { get; } = new GifFormat();
    }
```

src/ImageSharp/MetaData/ImageMetaData.cs
```diff
@@ -125,10 +132,46 @@ public double VerticalResolution
...
-        /// Gets or sets the number of times any animation is repeated.
-        /// <remarks>0 means to repeat indefinitely.</remarks>
+        /// Adds or updates the specified key and value to the <see cref="ImageMetaData"/>.
+        /// </summary>
+        /// <param name="key">The key of the metadata to add.</param>
+        /// <param name="value">The value of the element to add.</param>
+        /// <exception cref="ArgumentNullException">key is null.</exception>
+        /// <exception cref="ArgumentNullException">value is null.</exception>
+        /// <exception cref="ArgumentException">An element with the same key already exists in the <see cref="ImageMetaData"/>.</exception>
+        public void AddOrUpdateFormatMetaData(IImageFormat key, IImageFormatMetaData value)
```
antonfirsov: How about this pattern?
https://gist.github.com/antonfirsov/0eaf6ac9438655ea3affbf3212b33cc0

(Also posted on twitter)

src/ImageSharp/MetaData/ImageMetaData.cs
```diff
@@ -125,10 +132,46 @@ public double VerticalResolution
...
+        /// Gets the metadata value associated with the specified key.
         /// </summary>
-        public ushort RepeatCount { get; set; }
+        /// <typeparam name="T">The type of metadata.</typeparam>
+        /// <param name="key">The key of the value to get.</param>
+        /// <returns>
+        /// The <typeparamref name="T"/>.
+        /// </returns>
+        public T GetOrAddFormatMetaData<T>(IImageFormat key)
+            where T : IImageFormatMetaData, new()
```
antonfirsov: Can't decide whether I like the `new()` constraint or not. Seems reasonable here, but I've read pretty bad things about it:
https://blogs.msdn.microsoft.com/seteplia/2017/02/01/dissecting-the-new-constraint-in-c-a-perfect-example-of-a-leaky-abstraction/

(The performance concern does not apply here, for sure.)

antonfirsov: Actually, the constraint can be removed, if we follow the proposal in my gist!

We just need a factory method in `IImageFormat<TFormatMetaData>`:

```C#
public interface IImageFormat<TFormatMetaData> : IImageFormat
{
	TFormatMetaData CreateDefaultMetaData();
}
```

JimBobSquarePants: I like this!!!

src/ImageSharp/Formats/ImageFormatBase{T}.cs
JimBobSquarePants: Yeah, what I wanted though was a means of enforcing the standard. I'm still not sure.

JimBobSquarePants:
@antonfirsov Thanks for pushing me on this. It feels really clean now with your design.

JimBobSquarePants:
@antonfirsov Adding the missing meta management isn't that difficult. I've just done Png and will add the others later.

JimBobSquarePants:
@antonfirsov I've added proper metadata and encoder settings management for gif, png, and bmp for now. There's not really anything in jpeg I can add currently. 

It might be feasible, if it's not too much of a performance hog, to estimate the input quality of jpeg files in the manner ImageMagick does and add that to the JpegMetaData class.

https://github.com/ImageMagick/ImageMagick/blob/f362c02083d27211b913c6e44794f0ac6edaf2bd/coders/jpeg.c#L855

src/ImageSharp/Common/Helpers/ImageMaths.cs
```diff
@@ -36,10 +36,15 @@ public static int FastAbs(int x)
...
-        }
+        public static int GetBitsNeededForColorDepth(int colors) => Math.Max(1, (int)Math.Ceiling(Math.Log(colors, 2)));
+
+        /// <summary>
+        /// Returns how many colors will be created by the specified number of bits.
+        /// </summary>
+        /// <param name="bitDepth">The bit depth.</param>
+        /// <returns>The <see cref="int"/></returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int GetColorCountForBitDepth(int bitDepth) => (int)Math.Pow(2, bitDepth);
```
antonfirsov: you can write it as: `1 << bitDepth`

src/ImageSharp/MetaData/ImageMetaData.cs
```diff
@@ -125,10 +132,48 @@ public double VerticalResolution
...
-        /// <remarks>0 means to repeat indefinitely.</remarks>
+        /// Adds or updates the specified key and value to the <see cref="ImageMetaData"/>.
+        /// </summary>
+        /// <typeparam name="TFormatMetaData">The type of format metadata.</typeparam>
+        /// <param name="key">The key of the metadata to add.</param>
+        /// <param name="value">The value of the element to add.</param>
+        /// <exception cref="ArgumentNullException">key is null.</exception>
+        /// <exception cref="ArgumentNullException">value is null.</exception>
+        /// <exception cref="ArgumentException">An element with the same key already exists in the <see cref="ImageMetaData"/>.</exception>
+        public void AddOrUpdateFormatMetaData<TFormatMetaData>(IImageFormat<TFormatMetaData> key, TFormatMetaData value)
```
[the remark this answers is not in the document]
JimBobSquarePants: Absolutely agree on both methods. We've already constrained the type to a reference type so no need for the complexity.

I'd actually consider doing similar to `ImageMetaData` and `ImageFrameMetaData`. We check and create the metadata in the decoders but it really should always be there.

src/ImageSharp/Formats/Jpeg/Components/Decoder/QualityEvaluator.cs
```diff
@@ -0,0 +1,140 @@
...
+            667,   592,   518,   441,   369,   292,   221,   151,    86,
+            64,     0
+        };
+
+        /// <summary>
+        /// Returns an estimated quality of the image based on the quantization tables.
+        /// </summary>
+        /// <param name="quantizationTables">The quantization tables.</param>
+        /// <returns>The <see cref="int"/>.</returns>
+        public static int EstimateQuality(Block8x8F[] quantizationTables)
```
antonfirsov: 😮 this is real magick!

antonfirsov:
@JimBobSquarePants good job, I think we can merge this. Thanks to your patience, it's way much better now than the original proposal!

JimBobSquarePants:
I'm glad you pushed me!

Issue #694, opened 2018-09-06
ArgumentOutOfRangeException (Parameter name: start) when loading image


mabead:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Loading image throws exception.

### Steps to Reproduce
With 1.0.0-beta005, calling Image.Load on the attached file throws the following exception:

```
System.ArgumentOutOfRangeException
  HResult=0x80131502
  Message=Specified argument was out of the range of valid values.
  Source=System.Memory
  StackTrace:
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument)
   at SixLabors.ImageSharp.MetaData.Profiles.Exif.ExifReader.TryReadSpan(Int32 length, ReadOnlySpan`1& span)
   at SixLabors.ImageSharp.MetaData.Profiles.Exif.ExifReader.ReadUInt16()
   at SixLabors.ImageSharp.MetaData.Profiles.Exif.ExifReader.AddValues(List`1 values, Int32 index)
   at SixLabors.ImageSharp.MetaData.Profiles.Exif.ExifReader.GetThumbnail(Int32 offset)
   at SixLabors.ImageSharp.MetaData.Profiles.Exif.ExifReader.ReadValues()
   at SixLabors.ImageSharp.MetaData.Profiles.Exif.ExifProfile.InitializeValues()
   at SixLabors.ImageSharp.MetaData.Profiles.Exif.ExifProfile.TryGetValue(ExifTag tag, ExifValue& value)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.InitDerivedMetaDataProperties()
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.Decode[TPixel](Stream stream)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoder.Decode[TPixel](Configuration configuration, Stream stream)
   at SixLabors.ImageSharp.Image.Decode[TPixel](Stream stream, Configuration config)
[...]

Issue #696, opened 2018-09-06
ImageProcessingException/ArgumentOutOfRangeException when resizing image


mabead:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When resizing the attached image, the following exception is thrown:

```
SixLabors.ImageSharp.ImageProcessingException
  HResult=0x80131500
  Message=An error occurred when processing the image using ResizeProcessor`1. See the inner exception for more detail.
  Source=SixLabors.ImageSharp
  StackTrace:
   at SixLabors.ImageSharp.Processing.Processors.CloningImageProcessor`1.CloneAndApply(Image`1 source, Rectangle sourceRectangle)
   at SixLabors.ImageSharp.Processing.Processors.CloningImageProcessor`1.Apply(Image`1 source, Rectangle sourceRectangle)
   at SixLabors.ImageSharp.Processing.DefaultInternalImageProcessorContext`1.ApplyProcessor(IImageProcessor`1 processor, Rectangle rectangle)
   at SixLabors.ImageSharp.Processing.DefaultInternalImageProcessorContext`1.ApplyProcessor(IImageProcessor`1 processor)
   at SixLabors.ImageSharp.Processing.ResizeExtensions.Resize[TPixel](IImageProcessingContext`1 source, Int32 width, Int32 height)
   at LANDR.Distribution.ImageProcessor.PictureProvider.<>c__DisplayClass4_0.<Resize>b__0(IImageProcessingContext`1 ctx) in C:\git\Distribution\src\LANDR.Distribution.ImageProcessor\PictureProvider.cs:line 84

Inner Exception 1:
ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
```

Here's the image:
[Sans titre 1.zip](https://github.com/SixLabors/ImageSharp/files/2357397/Sans.titre.1.zip)

### Steps to Reproduce
```csharp
using (var imageStream = File.OpenRead(imageName))
using (var image = Image.Load(imageStream))
{
   image.Mutate(ctx => ctx.Resize(640, 640));
}
```
### System Configuration

[...]

Issue #695, opened 2018-09-06
NullReferenceException when loading jpeg image


mabead:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Loading image throws exception.

### Steps to Reproduce
With 1.0.0-beta005, calling Image.Load on the attached file throws the following exception:

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=SixLabors.ImageSharp
  StackTrace:
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder.IRawJpegData.get_Components()
   at SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder.JpegImagePostProcessor..ctor(MemoryAllocator memoryAllocator, IRawJpegData rawJpeg)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.PostProcessIntoImage[TPixel]()
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoder.Decode[TPixel](Configuration configuration, Stream stream)
   at SixLabors.ImageSharp.Image.Decode[TPixel](Stream stream, Configuration config)
   at SixLabors.ImageSharp.Image.WithSeekableStream[T](Configuration config, Stream stream, Func`2 action)
   at SixLabors.ImageSharp.Image.Load[TPixel](Configuration config, Stream stream, IImageFormat& format)
   at SixLabors.ImageSharp.Image.Load(Stream stream)
   at LANDR.Distribution.ImageProcessor.PictureProvider.Resize(Stream sourceStream, String target, Int32 newWidth, Int32 newHeight) in C:\git\Distribution\src\LANDR.Distribution.ImageProcessor\PictureProvider.cs:line 81
```

Here's the file: 
[i think im in love art 3000pg.zip](https://github.com/SixLabors/ImageSharp/files/2357364/i.think.im.in.love.art.3000pg.zip)

### System Configuration

- ImageSharp version: 1.0.0-beta005
[...]

PR #697, opened 2018-09-06
Fixed overflow reported in #696.

dlemstra:
### Prerequisites

### Description
This fixes an integer overflow inside the `AddValues` method of the ExifProfile that will set the position to a negative value and cause an exception when reading the data. This fixes the bug reported in #696.

### DIFF
--- a/src/ImageSharp/MetaData/Profiles/Exif/ExifReader.cs
+++ b/src/ImageSharp/MetaData/Profiles/Exif/ExifReader.cs
@@ -88,19 +88,19 @@ namespace SixLabors.ImageSharp.MetaData.Profiles.Exif
             }
 
             uint ifdOffset = this.ReadUInt32();
-            this.AddValues(values, (int)ifdOffset);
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Ah... I've actually fixed #694, #695, and #696 locally and was gonna submit a single PR. Just need to grab the latest reference images for the tests.

dlemstra:
Want me to close this one?  And isn't #694 a different issue? Also feel free to add some tests to this PR.

JimBobSquarePants:
It's actually the same issue, overflow when grabbing the values. I think it must have had something to do with the order of them during decoding.

If you could close this one please yes, I'll push the other PR soon and get you to review.

dlemstra:
I mean #695 where I said 694. Could we get this in a separate PR? 

The problem here is that the position has been changed from an `uint` to an `int` to avoid casting when reading from the span. Will close this one. 

p.s. Pinged you on twitter.

PR #707, opened 2018-09-18
Ensure frame metadata is a deep copy.

JimBobSquarePants:
### Prerequisites

### Description
I've introduced a new pair of interfaces here `IDeepCloneable` and `IDeepCloneable<T>` to ensure we force deep cloning of metadata objects in the library. This fixes an issue we introduced in #693 

I'd like to add the interface in a few more places actually to ensure that any cloning intent is unambiguous. There's a few places where it's not clear. I think adding a `IShallowCloneable` interface might be wise also since `ICloneable` is not available across all our target frameworks.

### DIFF
--- a/src/ImageSharp/Configuration.cs
+++ b/src/ImageSharp/Configuration.cs
@@ -125,7 +125,7 @@ namespace SixLabors.ImageSharp
         /// Creates a shallow copy of the <see cref="Configuration"/>
         /// </summary>
         /// <returns>A new configuration instance</returns>
-        public Configuration ShallowCopy()
+        public Configuration Clone()
         {
             return new Configuration
             {
--- a/src/ImageSharp/IDeepCloneable.cs
+++ b/src/ImageSharp/IDeepCloneable.cs
@@ -14,7 +14,7 @@ namespace SixLabors.ImageSharp
         /// Creates a new <typeparamref name="T"/> that is a deep copy of the current instance.
         /// </summary>
         /// <returns>The <typeparamref name="T"/>.</returns>
-        T DeepClone();
+        T Clone();
     }
 
     /// <summary>
@@ -26,6 +26,6 @@ namespace SixLabors.ImageSharp
         /// Creates a new object that is a deep copy of the current instance.
         /// </summary>
         /// <returns>The <see cref="IDeepCloneable"/>.</returns>
-        IDeepCloneable DeepClone();
+        IDeepCloneable Clone();
     }
 }
--- a/src/ImageSharp/IDeepCloneable.cs
+++ b/src/ImageSharp/IDeepCloneable.cs
@@ -7,8 +7,8 @@ namespace SixLabors.ImageSharp
     /// A generic interface for a deeply cloneable type.
     /// </summary>
     /// <typeparam name="T">The type of object to clone.</typeparam>
-    public interface IDeepCloneable<T>
-        where T : class, IDeepCloneable<T>
+    public interface IDeepCloneable<out T>
+        where T : class
     {
         /// <summary>
         /// Creates a new <typeparamref name="T"/> that is a deep copy of the current instance.
--- a/src/ImageSharp/IDeepCloneable.cs
+++ b/src/ImageSharp/IDeepCloneable.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) Six Labors and contributors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp
+{
+    /// <summary>
+    /// A generic interface for a deeply cloneable type.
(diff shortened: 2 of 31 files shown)


JimBobSquarePants:
@antonfirsov I've added some more deep cloning and will add tests now to ensure they all do what they should.

JimBobSquarePants:
Looking through the code it appears when we clone an image everything is deep cloned except for the `Configuration`. We can't currently perform a deep clone of that type. Do you think it's something we would require?

antonfirsov:
@JimBobSquarePants I don't think `MemoryAllocator` should be a clonable object. It's a good question whether we should clone `Configuration` when cloning an image. When a user is working with `Configuration.Default`, having several instances ` != Configuration.Default` after doing a `DeepClone` would be an unexpected side effect in the majority of the cases.

So in the end we have this complicated issue:
- `Configuration` should be Deep Clonable
- When deep copying an `Image<T>`, majority of the users probably do not want to copy `Configuration` (and they for sure do not want to deep clone `MemoryAllocator`!), but maybe some users intentionally need it.

It's a general issue with the concept of cloning. How deep should it be? It always depends on the context & the use case. I used to solve this with rather a complex pattern:

```C#
public interface IDuplicationContext
{
    // can hook any logic here
    // eg. track complex object graphs with loops, or do shallow copy at certain points
    T Duplicate<T>(T source); 
}

public interface IDuplicatable
{
    object Duplicate(IDuplicationContext ctx);
}

public class SimplestPossibleDuplicationContext : IDuplicationContext
{
    public T Duplicate<T>(T source)
    {
        if (T is IDuplicatable dup) return (T)source.Duplicate(this);
        return source;
    }
}

public class Head : IDuplicatable
{
    object Duplicate(IDuplicationContext ctx) { .... }
}

public class Cat : IDuplicatable
{
    public Head Head { get; private set; }
    public int Legs { get; private set; }

    public Duplicate(IDuplicationContext ctx)
    {
        Cat clone = new Cat();
        clone.Head = ctx.Duplicate(this.Head);
        clone.Legs = this.Legs;
        return clone;
    }
}
```

Good question if we can keep it simple, avoiding stuff like this? Or can we somehow simplify the concept of deep cloning without users experiencing unwanted side effects?

JimBobSquarePants:
That's all good points and an interesting code sample. 

I think we should keep it simple. We're explicitly creating a singleton for the default configuration and to clone that configuration when creating a deep clone of an image, in my mind, seems silly. We don't actually use the current `ShallowCopy()` method anywhere in the code, only the tests.

If someone wants a clone of an image with a new configuration they can already do this.

``` c#
var configuration = new Configuration();
IEnumerable<ImageFrame<Rgba32>> clonedFrames = image.Frames.Select(x => x.Clone());
var clone = new Image<Rgba32>(configuration, image.MetaData.DeepClone(), clonedFrames);
```

We could make it easier to do this by creating overloads for `Clone()` and `CloneAs()` which take a `Configuration` parameter, then all they have to do is create a new one. 


Anything else.... Probably overkill.

antonfirsov:
Yeah, looks like the solution is that `Image<T>` should not be `IDeepCloneable`. `image.Clone()` should do a deep copy of all the metadata however. (Which is already implemented based on the code changes.)

JimBobSquarePants:
I'll do some minor cleanup then and get this merged.

src/ImageSharp/Configuration.cs
```diff
@@ -125,7 +125,7 @@ public void Configure(IConfigurationModule configuration)
         /// Creates a shallow copy of the <see cref="Configuration"/>
         /// </summary>
         /// <returns>A new configuration instance</returns>
-        public Configuration ShallowCopy()
+        public Configuration Clone()
```
antonfirsov: I think this is misleading now. `IDeepCloneable.Clone()` does a deep copy + we also need to make `Configuration` deep clonable.

src/ImageSharp/IDeepCloneable.cs
```diff
@@ -14,7 +14,7 @@ public interface IDeepCloneable<T>
         /// Creates a new <typeparamref name="T"/> that is a deep copy of the current instance.
         /// </summary>
         /// <returns>The <typeparamref name="T"/>.</returns>
-        T DeepClone();
+        T Clone();
```
antonfirsov: I think more a verbose name is better in this specific case. What if an object is both deep an shallow cloneable? (Like `Configuration`)

JimBobSquarePants: It just looked weird that's all when we had a mix of both `Clone()` and `DeepClone()` together. I'll revert it though.

JimBobSquarePants:
@antonfirsov Good idea!

PR #708, opened 2018-09-20
Fix issue on mono <= 5.14 when reading multiple png data chunks.

### LINKED ISSUE #598: PNG gets decoded incorrectly on Mono
### Prerequisites

### Description
The image below gets decoded incorrectly on Mono.
second

It gets loaded correctly on Windows, but on Mono the bottom is transparent.
output

This bug seems to be similar to #223.

I have confirmed this bug on Android 8 and on Ubuntu 16.04 using the Linux Subsystem for Windows.

### Steps to Reproduce
Load the first image in a Mono environment.

### System Configuration
Android 8.0.0 running Mono
Ubuntu 16.04 running Mono, using the Linux Subsystem for Windows

- ImageSharp version: v1.0.0-beta0004
- Environment (Operating system, version and so on): Android 8.0.0, Ubuntu 16.04
- .NET Framework version: Mono
- Additional information:

cwensley:
### Prerequisites

### Description
This fixes #598 when decoding PNG data chunks in mono.  

Some people might appreciate this fix, as: 
- it will take a while for the [fix in mono](https://github.com/mono/mono/pull/10329) to make it to a stable release (looks like it is back ported to 5.18 at least)
- sometimes you are developing for software that embeds an old version of mono (like we do)
- or, you might not have control over what version of mono you get to target

What this does is changes `PngDecoderCore` and `ZlibInflateStream` so that when it is finished with the current chunk, it attempts to read the next data chunk before returning less than the buffer count from its `Read()` method.  Doing so would stop the `DeflateStream` from continuing, even on Windows.  It will only return less than the buffer count after it reads a non-data chunk, then forwards that chunk to the next call of `TryReadChunk`.

The [PNG spec](http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html) specifies that all data chunks must be consecutive, which this code takes advantage of. Specifically:

>There can be multiple IDAT chunks; if so, they must appear consecutively with no other intervening chunks

I've tried to follow the same coding conventions (no warnings in VS), and tried to make the changes as minimal as possible.  I also ran all unit tests to ensure they pass.  I believe a few of the existing unit tests will fail on mono, so I didn't add any new unit tests (plus, I can't seem to figure out how to run the xunit tests on macOS, they all fail with `System.TypeLoadException : Could not resolve the signature of a virtual method`).

Please let me know if there's anything that isn't quite right, and thanks for your time!

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -188,6 +188,11 @@ namespace SixLabors.ImageSharp.Formats.Png
         private bool hasTrans;
 
         /// <summary>
+        /// The next chunk of data to return
+        /// </summary>
+        private PngChunk? nextChunk;
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="PngDecoderCore"/> class.
         /// </summary>
         /// <param name="configuration">The configuration.</param>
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Oh well done!!!!!! Thank you!

cwensley:
Awesome, thanks for merging it in!  It will be very helpful to be able to use ImageSharp with mono.

PR #710, opened 2018-09-23
Optimized parallel pixel processing

### LINKED ISSUE #700: Performance overhead in Image<> constructor's Clear()
### Prerequisites

### Description

In use cases where many `Image<>`s are being constructed and expected to be produced at low latency, an overhead can currently be experienced due to the presence of `Parallel.For` in the [pixel clearing code](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/ImageFrame%7BTPixel%7D.cs#L295).

image

Even in the case where task inlining occurs, an overhead of around 65% can be observed. The case in this profiling involves 1x100 image created and populated [each game frame](https://github.com/ppy/osu-framework/pull/1856/files#diff-9283ee41d1dcd3a45f04350f89e3f254R350).

Two points here:
- `Parallel.For` should probably not be used under a certain threshold of image pixels as the overhead is larger than any benefit.
- In the example cited, as we are manually populating the image, the clear is completely redundant. It may be nice to have the ability to skip the `Clear` in such cases (although I could understand if this is not accepted, as it could potentially be a security concern in a multi-tenant environment).

On advice, I did try converting my image to 100x1 to reduce parallelism (since the `Parallel.For` is done [per row](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/ImageFrame%7BTPixel%7D.cs#L303)) but this has no effect. The overhead was already occurred from the `Parallel.For`'s task paradigm.

This is a low priority proposal as I can restore my local pooling of `Image<>`s rather than constructing a new one each time (probably necessary anyway to avoid object alloc overheads anyway).

### Steps to Reproduce

Profile CPU usage for a loop like:

```csharp
for (int i = 0; i < 10000; i++)
	using (var test = new Image<Rgba32>(100,1);
```

### System Configuration

.net core 2.1, latest ImageSharp 1.0.0-beta0005

antonfirsov:
### Prerequisites

### Description
This PR minimizes the overhead of TPL by processing multiple consecutive rows by a single task:

```csharp
var workingRect = Rectangle.FromLTRB(minX, minY, maxX, maxY);

ParallelHelper.IterateRows(
            workingRect,
            configuration,
            rows =>
            {
                for (int y = rows.Min; y < rows.Max; y++)
                {
                    // Process pixels here
                }
            });
```

- Per-row delegate invocations are eliminated
- There is a minimum number of pixels (`ParallelExecutionSettings.MinimumPixelsProcessedPerTask`) to be processed within a single row interval. This way we avoid creating tasks for small images.
- If the whole image fits into this minimum, or `Configuration.MaxDegreeOfParallelism == 1` TPL is not used at all, the whole image is processed within a single row interval.
- In some cases (like `ImageFrame<T>.Clear(...)`) parallelization is entirely dropped.

Fixes #700 and similar issues.

### Test improvements
Reference images + assertions have been added to many processor tests.


src/ImageSharp/Common/ParallelUtils/ParallelHelper.cs
```diff
@@ -0,0 +1,151 @@
...
+            IterateRowsWithTempBuffer(rectangle, configuration.GetParallelSettings(), body);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static int DivideCeil(int dividend, int divisor)
+        {
+            // TODO: Is there a more efficient way to calculate this?
+            int result = dividend / divisor;
+            return dividend % divisor == 0 ? result : result + 1;
+        }
```
JimBobSquarePants: `(dividend > 0) ? 1 + ((dividend - 1) / divisor) : (dividend / divisor)` Maybe? Seems to work in my local tests.

src/ImageSharp/Processing/Processors/Transforms/CropProcessor.cs
```diff
@@ -53,21 +53,23 @@ protected override void OnFrameApply(ImageFrame<TPixel> source, ImageFrame<TPixe
...
-                configuration,
-                y =>
-                {
-                    Span<TPixel> sourceRow = source.GetPixelRowSpan(y).Slice(minX);
-                    Span<TPixel> targetRow = destination.GetPixelRowSpan(y - minY);
-                    sourceRow.Slice(0, maxX - minX).CopyTo(targetRow);
-                });
+            // Copying is cheap, we should process more pixels per task:
+            ParallelExecutionSettings parallelSettings = configuration.GetParallelSettings().MultiplyMinimumPixelsPerTask(4);
+
```
JimBobSquarePants: Agreed 👍

tests/ImageSharp.Tests/Helpers/ParallelHelperTests.cs
```diff
@@ -0,0 +1,338 @@
...
+using SixLabors.ImageSharp.ParallelUtils;
+using SixLabors.Memory;
+using SixLabors.Primitives;
+
+using Xunit;
+using Xunit.Abstractions;
+
+namespace SixLabors.ImageSharp.Tests.Helpers
+{
+    public class ParallelHelperTests
```
JimBobSquarePants: Good tests! 👍

src/ImageSharp/Common/ParallelUtils/ParallelHelper.cs
```diff
@@ -0,0 +1,151 @@
...
+using SixLabors.Primitives;
+
+namespace SixLabors.ImageSharp.ParallelUtils
+{
+    /// <summary>
+    /// Utility methods for batched processing of pixel row intervals.
+    /// Parallel execution is optimized for image processing.
+    /// Use this instead of direct <see cref="Parallel"/> calls!
+    /// </summary>
+    internal static class ParallelHelper
```
JimBobSquarePants: Loving this, very clever! 👍

src/ImageSharp/Common/ParallelUtils/ParallelExecutionSettings.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    /// <summary>
+    /// Defines execution settings for methods in <see cref="ParallelHelper"/>.
+    /// </summary>
+    internal struct ParallelExecutionSettings
+    {
+        /// <summary>
+        /// Default value for <see cref="MinimumPixelsProcessedPerTask"/>.
+        /// </summary>
+        public const int DefaultMinimumPixelsProcessedPerTask = 4096;
```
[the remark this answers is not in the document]
antonfirsov: No theory just intutively feel this should be a relatively high value (probably even higher than the current one). It's not that important to find the best value TBH and it's opreation specific. Needs lots of benchmarking.

antonfirsov:
@JimBobSquarePants all fixed!

JimBobSquarePants:
Is everyone happy to have this merged?

PR #712, opened 2018-09-26
Add 1, 2, and 4 bit Png Encoding

JimBobSquarePants:
### Prerequisites

### Description
This PR will add 1,2, and 4 bit encoding to png. 

- [x] Add encoding for paletted images
- [x] Add encoding for grayscale images
- [x] Add tests and optimize performance and readability

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -93,6 +93,11 @@ namespace SixLabors.ImageSharp.Formats.Png
         private readonly bool ignoreMetadata;
 
         /// <summary>
+        /// Used the manage memory allocations.
+        /// </summary>
+        private readonly MemoryAllocator memoryAllocator;
+
+        /// <summary>
         /// The stream to decode from.
         /// </summary>
         private Stream currentStream;
@@ -200,11 +205,10 @@ namespace SixLabors.ImageSharp.Formats.Png
         public PngDecoderCore(Configuration configuration, IPngDecoderOptions options)
         {
             this.configuration = configuration ?? Configuration.Default;
+            this.memoryAllocator = this.configuration.MemoryAllocator;
             this.textEncoding = options.TextEncoding ?? PngConstants.DefaultEncoding;
             this.ignoreMetadata = options.IgnoreMetadata;
         }
-
-        private MemoryAllocator MemoryAllocator => this.configuration.MemoryAllocator;
 
         /// <summary>
         /// Decodes the stream to the image.
@@ -391,12 +395,12 @@ namespace SixLabors.ImageSharp.Formats.Png
                 return false;
             }
 
-            buffer = this.MemoryAllocator.AllocateManagedByteBuffer(bytesPerScanline * 8 / bits, AllocationOptions.Clean);
+            buffer = this.memoryAllocator.AllocateManagedByteBuffer(bytesPerScanline * 8 / bits, AllocationOptions.Clean);
             byte[] result = buffer.Array;
             int mask = 0xFF >> (8 - bits);
             int resultOffset = 0;
 
-            for (int i = 0; i < bytesPerScanline - 1; i++)
+            for (int i = 0; i < bytesPerScanline; i++)
             {
                 byte b = source[i];
                 for (int shift = 0; shift < 8; shift += bits)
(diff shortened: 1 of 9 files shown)


src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -830,6 +841,42 @@ private void WriteChunk(Stream stream, PngChunkType type, byte[] data, int offse
             stream.Write(this.buffer, 0, 4); // write the crc
         }
 
+        private void ScaleDownFrom8BitArray(ReadOnlySpan<byte> source, Span<byte> result, int bits)
```
JimBobSquarePants: There's something not right here. We're getting pixels from the end of a row overwriting pixels at the start.

bpp1

bpp1

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -830,6 +841,42 @@ private void WriteChunk(Stream stream, PngChunkType type, byte[] data, int offse
...
+                }
+                else
+                {
+                    shift -= bits;
+                }
+            }
+
+            if (shift != shift0)
+            {
+                result[0] = (byte)v;
```
dlemstra: Should you not move the `i` outside the for loop and use it here?

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -830,6 +841,42 @@ private void WriteChunk(Stream stream, PngChunkType type, byte[] data, int offse
             stream.Write(this.buffer, 0, 4); // write the crc
         }
 
+        private void ScaleDownFrom8BitArray(ReadOnlySpan<byte> source, Span<byte> result, int bits)
+        {
+            if (bits >= 8)
```
dlemstra: Do we also need checks for values like `3,5,7`? But do we really need this check, we cannot enter this method with a higher value?

src/ImageSharp/Formats/Png/PngEncoderCore.cs
JimBobSquarePants: Fixed. Thanks @dlemstra for spotting my daft mistakes.

JimBobSquarePants:
Ready for review

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -391,19 +395,19 @@ private bool TryScaleUpTo8BitArray(ReadOnlySpan<byte> source, int bytesPerScanli
...
+            ref byte sourceRef = ref MemoryMarshal.GetReference(source);
+            ref byte resultRef = ref buffer.Array[0];
             int mask = 0xFF >> (8 - bits);
             int resultOffset = 0;
 
-            for (int i = 0; i < bytesPerScanline - 1; i++)
+            for (int i = 0; i < bytesPerScanline; i++)
             {
-                byte b = source[i];
+                byte b = Unsafe.Add(ref sourceRef, i);
```
dlemstra: I wonder if it would not be more efficient to just increment by 1 instead?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -718,211 +722,74 @@ private void DecodeInterlacedPixelData<TPixel>(Stream compressedStream, ImageFra
...
         {
-            TPixel pixel = default;
             Span<TPixel> rowSpan = pixels.GetPixelRowSpan(this.currentRow);
 
             // Trim the first marker byte from the buffer
             ReadOnlySpan<byte> trimmed = defilteredScanline.Slice(1, defilteredScanline.Length - 1);
 
             // Convert 1, 2, and 4 bit pixel data into the 8 bit equivalent.
-            ReadOnlySpan<byte> scanlineSpan = this.TryScaleUpTo8BitArray(trimmed, this.bytesPerScanline, this.header.BitDepth, out IManagedByteBuffer buffer)
+            ReadOnlySpan<byte> scanlineSpan = this.TryScaleUpTo8BitArray(trimmed, this.bytesPerScanline - 1, this.header.BitDepth, out IManagedByteBuffer buffer)
```
dlemstra: Better to do the -1 here 👍

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: We avoid index checks using Unsafe

PR #728, opened 2018-10-06
Add ToString implementations to Bgr24 and Bgra32

jongleur1983:
### Prerequisites

### Description
Bgr24 and Bgra32 did not yet override ToString(). Did that now, following the same pattern as other Pixel formats (printing the components)


JimBobSquarePants:
Good catch! 👍

Issue #718, opened 2018-09-30
Add Gray8 and Gray16 Pixel Formats


JimBobSquarePants:
Currently if someone wants to read/write opaque grayscale png files they need to convert the image to and from  `Rgb24` or `Rgba32` which requires the user to consume more memory than necessary. 

https://stackoverflow.com/questions/52407878/load-and-save-opaque-8-bit-png-files-using-imagesharp?rq=1

I propose we do the following:

Create the following `IPixel` implementations `Gray8` and `Gray16`. 

**Gray8**
``` c#
/// <summary>
/// Packed pixel type containing one 8-bit unsigned normalized value ranging from 0 to 255.
// <para>
/// Ranges from [0, 0, 0, 1] to [1, 1, 1, 1] in vector form.
/// </para>
/// </summary>
public struct Gray8 : IPixel<Gray8>
{
    /// <summary>
    /// Gets or sets the luminance component.
    /// </summary>
    public ushort L;

   // etc...
}
``` 

**Gray16**
``` c#
/// <summary>
/// Packed pixel type containing one 16-bit unsigned normalized value ranging from 0 to 635535.
/// <para>
/// Ranges from [0, 0, 0, 1] to [1, 1, 1, 1] in vector form.
/// </para>
/// </summary>
public struct Gray16 : IPixel<Gray16>
{
    /// <summary>
    /// Gets or sets the luminance component.
    /// </summary>
    public ushort L;

   // etc...
}
```

Add the additional methods to `IPixel`

``` c#
/// <summary>
/// Packs the pixel from an <see cref="Gray8"/> value.
/// </summary>
/// <param name="source">The <see cref="Gray8"/> value.</param>
void PackFromGray8(Gray8 source);

/// <summary>
/// Converts the pixel to <see cref="Gray8"/> format.
/// </summary>
/// <param name="dest">The destination pixel to write to.</param>
void ToGray8(ref Gray8 dest)

/// <summary>
/// Packs the pixel from an <see cref="Gray16"/> value.
/// </summary>
/// <param name="source">The <see cref="Gray16"/> value.</param>
void PackFromGray16(Gray16 source);

/// <summary>
/// Converts the pixel to <see cref="Gray16"/> format.
/// </summary>
/// <param name="dest">The destination pixel to write to.</param>
void ToGray16(ref Gray16 dest)
```
Add the following methods to `PixelOperations` (These can be added to the T4 template).

``` c#
/// <summary>
[...]

PR #729, opened 2018-10-06
#718 #730: Add Gray8 and Gray16 Pixel Formats and clean up IPixel

jongleur1983:
### Prerequisites

### Description
Implements Pixel formats Gray8 and Gray16.
Where I knew how to do it I provided optimized conversion methods, where not I used the "fallback" strategy to convert from and to Vector4, as it's done for many other implementations as well.

Unit tests are based on some of the other Pixel formats tests. Not sure this is seen as sufficient though, please tell me if not.

I guess I have to squash commits again, but for review the individual commits may help so I'm happy to do so after a first review.


src/ImageSharp/PixelFormats/Argb32.cs
```diff
@@ -289,6 +290,35 @@ public void ToBgra32(ref Bgra32 dest)
...
+        {
+            this.A = 255;
+            this.R = source.PackedValue;
+            this.G = source.PackedValue;
+            this.B = source.PackedValue;
+        }
+
+        /// <inheritdoc/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void ToGray8(ref Gray8 dest) => dest.PackFromScaledVector4(this.ToScaledVector4());
```
JimBobSquarePants: This can be optimized.

src/ImageSharp/PixelFormats/Gray8.cs
```diff
@@ -0,0 +1,266 @@
...
+        public Vector4 ToVector4()
+        {
+            return new Vector4(this.PackedValue, this.PackedValue, this.PackedValue, 1.0f);
+        }
+
+        /// <inheritdoc />
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void PackFromRgba32(Rgba32 source)
+        {
+            this.PackedValue = Pack(source.R, source.G, source.B);
```
JimBobSquarePants: This and others is incorrect. Will lead to a packed value of 255.

src/ImageSharp/PixelFormats/Gray8.cs
```diff
@@ -0,0 +1,266 @@
...
+        /// <inheritdoc/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void PackFromRgb48(Rgb48 source) =>
+            this.PackedValue = Pack(source.R, source.G, source.B);
+
+        /// <inheritdoc/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void ToRgb48(ref Rgb48 dest)
+        {
+            ushort gray = (ushort)(this.PackedValue * 255);
```
JimBobSquarePants: `* 257`

src/ImageSharp/PixelFormats/Gray8.cs
```diff
@@ -0,0 +1,266 @@
...
+
+        /// <summary>
+        /// Packs a <see cref="float"/> into a byte.
+        /// </summary>
+        /// <param name="r">Red value of the color to pack.</param>
+        /// <param name="g">Green value of the color to pack.</param>
+        /// <param name="b">Blue value of the color to pack.</param>
+        /// <returns>The <see cref="byte"/> containing the packed value.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static byte Pack(float r, float g, float b)
```
JimBobSquarePants: This only works with `Vector4` but is being used for `Rgba32` etc.

JimBobSquarePants:
I'm gonna do #730 at the same time otherwise we're writing code we'll immediately delete.

JimBobSquarePants:
@jongleur1983 Ok. So I've fixed up all the pixels formats and commented out all the old tests. 

I need to do two things now to complete this PR:

1. Add a new suite of tests to cover IPixel functionality
2. Add optimized overloads for the PixelOperations methods.

JimBobSquarePants:
Got four tests failing due to `ToString` change in `Rgba32` as the reference image path depended on original name. We should update that. Perhaps hex?

https://ci.appveyor.com/project/six-labors/imagesharp/builds/19386791/job/x9ru3njufhn9ec6j/tests

antonfirsov:
Yeah hex is better for this purpose

JimBobSquarePants:
Ready for review.

Code coverage is a bit less than I would like but I'm done now. It's been 10 days writing this.

src/ImageSharp/PixelFormats/Generated/PixelOperations{TPixel}.Generated.cs
```diff
@@ -10,320 +10,379 @@ namespace SixLabors.ImageSharp.PixelFormats
...
+            ref Argb32 destBaseRef = ref MemoryMarshal.GetReference(destPixels);
 
             for (int i = 0; i < count; i++)
             {
                 ref TPixel sp = ref Unsafe.Add(ref sourceBaseRef, i);
-                ref Rgba64 dp = ref Unsafe.Add(ref destBaseRef, i);
-                sp.ToRgba64(ref dp);
+                ref Argb32 dp = ref Unsafe.Add(ref destBaseRef, i);
+
+                dp.PackFromScaledVector4(sp.ToScaledVector4());
```
antonfirsov: This is again a significant performance regression. Many pixel formats had efficient implementations for the conversion of basic formats like (`Rgba32` <-> `Bgra32` <-> `Argb32` <-> `Rgb24` etc, implemented with byte-shuffling.

We should not replace those with expensive internal `Vector4` conversions!

antonfirsov:
Ok, actually this *might* be good, I haven't noticed the new optimized bulk implementations in my review!

src/ImageSharp/PixelFormats/IPixel.cs
```diff
@@ -48,83 +42,77 @@ public interface IPixel
...
+        void PackFromRgba32(Rgba32 source);
 
         /// <summary>
-        /// Converts the pixel to <see cref="Argb32"/> format.
+        /// Expands the packed representation into an <see cref="Rgba32"/>.
         /// </summary>
-        /// <param name="dest">The destination pixel to write to</param>
-        void ToArgb32(ref Argb32 dest);
+        /// <returns>The <see cref="Rgba32"/>.</returns>
+        Rgba32 ToRgba32();
```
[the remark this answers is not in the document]
antonfirsov: The good thing is that this method is rarely used. We can probably address this later together with the review on `ToVector4()` / `ToScaledVector4()`.

JimBobSquarePants:
@antonfirsov Yeah, all the `IPixel` conversions have optimized bulk implementations so there should be no performance loss. It should actually be faster in some cases than before.

src/ImageSharp/PixelFormats/IPixel.cs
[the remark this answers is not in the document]
JimBobSquarePants: I don't follow. Are you saying that `ToRgba()` is 9x slower than `ToRgba(ref Rgba)`? I would have thought the cost of the pointer would outweigh any benefit especially in 64bit mode.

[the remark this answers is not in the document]
antonfirsov: Depends on the use case.
The case in the benchmark is basically the `Span<TPixel>` -> `Span<Rgba32>` conversion. Writing right into the reference of the destination buffer removes one extra copy, compared to the case when you are assigning the destiniation buffer item through a return value.

I suspect, there are probably some other unnecessary operations JIT fails to optimize.

We should also benchmark an other case:
When the result of `ToRgba()` is being copied into a temporary variable for a small modification before writing it to the destination buffer. This is what we do in the PNG codec.

We should implement these benchmarks for `ToVector4()` as well. Gonna work on these points tonight!

JimBobSquarePants:
@antonfirsov Those results are extraordinary, I'm very surprised!

Let's do it here now. A new PR will be another very large one again probably too large to digest. We know where we are here and can review knowing exactly what we are looking at having recently worked on the files.

antonfirsov:
One possible explanation is that JIT is not capable for [Return value optimization](https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization) in most cases.

Do you mean also finishing up the `Vector4` stuff here? Should/can we split the work somehow?

JimBobSquarePants:
I was tempted to do it all here but for sanities sake lets do `Rgba32` here and immediately raise an issue for next priority for `Vector4`. We don't want to be blocking any open PR's

antonfirsov:
@JimBobSquarePants my feeling is that it would be much simpler to merge this as is, right after #742 and carry out all my proposed API changes together in a new PR.

JimBobSquarePants:
@antonfirsov works for me! 👍

antonfirsov:
@JimBobSquarePants any idea how to ping the CLA assistant? Pressing "Recheck PR-s" on it's UI did not help.

JimBobSquarePants:
@antonfirsov Github's having all sorts of issues just now.

PR #734, opened 2018-10-11
Refactor Companding

JimBobSquarePants:
### Prerequisites

### Description
Move all the sRGB Companding code to the same class. Also replaces extension method with static method call. 

Additional refactorings to consuming algorithms.


src/ImageSharp/Common/Helpers/Vector4Utils.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            // TODO: This method can be AVX2 optimized using Vector<float>
+            ref Vector4 baseRef = ref MemoryMarshal.GetReference(vectors);
+
+            for (int i = 0; i < vectors.Length; i++)
+            {
+                ref Vector4 v = ref Unsafe.Add(ref baseRef, i);
+                var s = new Vector4(v.W)
+                {
+                    W = 1
```
antonfirsov: I think there is an extra copy with the property assignment:
1. Create `temp = new Vector4(v.W)`
2. Assign `temp.W=1`
3. Assign the whole stuff to `s`: `s = temp`

Both the C# compiler and JIT fails to optimize this to my orginal code:
1. `s = new Vector4(v.W)`
2. `s.W = 1`

src/ImageSharp/Common/Helpers/Vector4Utils.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            // TODO: This method can be AVX2 optimized using Vector<float>
+            ref Vector4 baseRef = ref MemoryMarshal.GetReference(vectors);
+
+            for (int i = 0; i < vectors.Length; i++)
+            {
+                ref Vector4 v = ref Unsafe.Add(ref baseRef, i);
+                var s = new Vector4(1 / v.W)
+                {
+                    W = 1
```
antonfirsov: Same here

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -0,0 +1,131 @@
...
+            for (int y = 0; y < matrixHeight; y++)
+            {
+                int offsetY = (row + y - radiusY).Clamp(0, maxRow);
+                Span<TPixel> sourceRowSpan = sourcePixels.GetRowSpan(offsetY);
+
+                for (int x = 0; x < matrixWidth; x++)
+                {
+                    int offsetX = (column + offsetColumn + x - radiusX).Clamp(offsetColumn, maxColumn);
+                    var currentColor = sourceRowSpan[offsetX].ToVector4();
+                    Vector4Utils.Premultiply(ref currentColor);
```
antonfirsov: I assume that applying bulk premultiplication outside the convolution might be better, because simpler bulk methods are usually faster. (And it's definitely a better separation of concerns!). 

If any concerns, we should benchmark this!

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -0,0 +1,131 @@
...
+                    currentColor *= matrix[y, x];
```
antonfirsov: Why don't you take the pixel span (or the reference to the first row item) outside the loop?

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -0,0 +1,131 @@
...
+                int offsetY = (row + y - radiusY).Clamp(0, maxRow);
+                Span<TPixel> sourceRowSpan = sourcePixels.GetRowSpan(offsetY);
+
+                for (int x = 0; x < matrixXWidth; x++)
+                {
+                    int offsetX = (column + offsetColumn + x - radiusX).Clamp(offsetColumn, maxColumn);
+                    var currentColor = sourceRowSpan[offsetX].ToVector4();
+                    Vector4Utils.Premultiply(ref currentColor);
+
+                    if (y < matrixXHeight)
```
antonfirsov: These if-s are perf killers, because they mess up CPU branch prediction. Can't we iterate with the `y < Math.Min(matrixYHeight, matrixXHeight)` instead?

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -0,0 +1,131 @@
...
+                    {
+                        vectorX += matrixX[y, x] * currentColor;
```
antonfirsov: Slow indexer!

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -0,0 +1,131 @@
...
+                    }
+
+                    if (x < matrixYWidth)
```
antonfirsov: Same as for the other condition.

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -0,0 +1,131 @@
...
+            for (int y = 0; y < matrixYHeight; y++)
+            {
+                int offsetY = (row + y - radiusY).Clamp(0, maxRow);
+                Span<TPixel> sourceRowSpan = sourcePixels.GetRowSpan(offsetY);
+
+                for (int x = 0; x < matrixXWidth; x++)
+                {
+                    int offsetX = (column + offsetColumn + x - radiusX).Clamp(offsetColumn, maxColumn);
+                    var currentColor = sourceRowSpan[offsetX].ToVector4();
+                    Vector4Utils.Premultiply(ref currentColor);
```
antonfirsov: Again, I suppose it's better to do the premultiplication/unpremultiplication outside the convolution method.

src/ImageSharp/Processing/Processors/Convolution/Convolution2DProcessor.cs
```diff
@@ -47,89 +45,43 @@ public Convolution2DProcessor(DenseMatrix<float> kernelX, DenseMatrix<float> ker
...
                         {
+                            Span<Vector4> vectorSpan = vectorBuffer.Span;
+                            int length = vectorSpan.Length;
+
                             for (int y = rows.Min; y < rows.Max; y++)
                             {
-                                Span<TPixel> sourceRow = source.GetPixelRowSpan(y);
-                                Span<TPixel> targetRow = targetPixels.GetRowSpan(y);
+                                Span<TPixel> targetRowSpan = targetPixels.GetRowSpan(y).Slice(startX);
+                                PixelOperations<TPixel>.Instance.ToVector4(targetRowSpan, vectorSpan, length);
```
antonfirsov: 👍

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
JimBobSquarePants: Yeah, that'd be better.

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
JimBobSquarePants: We're only dealing with small numbers but maybe.... I gave it a go outside the loop but soon got lost.

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
JimBobSquarePants: I'll have to add a `GetRowSpan()` method to the matrix.

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
JimBobSquarePants: These checks were there originally because this would have been two loops and there was no guarantee that the matrices were the same dimensions. However I just checked and all the matrices using the method are so I'll clean this up.

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
antonfirsov: I see. Then it would be better to introduce a `.GetReferenceToFirstItem()` or something, and do everything with `Unsafe` arithmetics than bothering with spans!

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
antonfirsov: Actually forgot that it's all about tiny matrices. They really should be value types (#142), and other optimization attempts are maybe unworthy.

JimBobSquarePants: `DenseMatrix<T>` is a readonly struct. The indexer is benchmarked to be as fast as raw 1D array indexing.

antonfirsov: Indexing a 2D struct contains a multiplication which is unnecessary within the inner loop if we have a pointer to the first item of the row.

A note on benchmarking:
In this case, a proper benchmark is a one that iterates through all elements. Using constant index values could be misleading.

But nevermind the whole thing, this is of low importance at this point. I'd rather review the design, than make small improvements on the implementation.

JimBobSquarePants: Agreed. I'll merge this for now  and we can bash out some designs in the future. I'd like to get this right before we make the processors public.

FYI. Re the benchmark. It benchmarks iterating across entire arrays of different sizes.

https://github.com/SixLabors/ImageSharp/blob/13850d1fec2af49176d08a3576c5e07b7430e476/tests/ImageSharp.Benchmarks/General/Array2D.cs#L113

antonfirsov: But it has [the unnecessary `count * i` multiplication](https://github.com/SixLabors/ImageSharp/blob/13850d1fec2af49176d08a3576c5e07b7430e476/tests/ImageSharp.Benchmarks/General/Array2D.cs#L70) within the inner loop! 

Also: Going with `Unsafe` and remove the array indexers would strip the benchmarks from the array bounds checking, which is probably the most expensive stuff here, making the results very similar.

PR #736, opened 2018-10-15
Write float values using the invariant culture.

j0rn:
### Prerequisites

### Description
Float values a written using the current users culture.  Tests fails for "non-english" culture.

    StringRepresentationsAreCorrect(color: CieLab(42,4, 94,5, 83,4), text: "CieLab(42.4, 94.5, 83.4)") [FAIL]
      Assert.Equal() Failure
                          (pos 9)
      Expected: CieLab(42.4, 94.5, 83.4)
      Actual:   CieLab(42,4, 94,5, 83,4)
                          (pos 9)


src/ImageSharp/ColorSpaces/CieLab.cs
```diff
@@ -127,7 +128,7 @@ public override int GetHashCode()
         }
 
         /// <inheritdoc/>
-        public override string ToString() => $"CieLab({this.L:#0.##}, {this.A:#0.##}, {this.B:#0.##})";
+        public override string ToString() => $"CieLab({FloatToString(this.L, this.A, this.B)})";
```
tocsoft: inset of adding `FloatToStringUtil` (and the extra allocation that involves) you should use this syntax instead:

```
public override string ToString() => FormattableString.Invariant($"CieLab({this.L:#0.##}, {this.A:#0.##}, {this.B:#0.##})");
```

j0rn: image

ColorSpaces\CieXyz.cs(97,46): error CS0103: The name 'FormattableString' does not exist in the current context [ImageSharp.csproj]

dlemstra: We probably should

JimBobSquarePants: Yup Windows phone is dead now and the only reason I clung on.

https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support

iamcarbon: I submitted PRs across the repos to increase the NETSTANDARD target to v1.3.

JimBobSquarePants: Merged, and I've updated this PR from master so should be good to use now.

j0rn: Comitted with FormattableString.Invariant.

tocsoft:
@j0rn thanks for the fix 👍

PR #744, opened 2018-10-23
Refactor & simplify conversion API's

antonfirsov:
### Prerequisites

### 1. `IPixel`
  - Better terminology/naming: `PackFrom***` ==> `From***`
  - `Rgba32 ToRgba32()` ==>  `void ToRgba32(ref dest)` for perf
  - `ToVector4` and `ToScaledVector4` were not touched: after introducing scaling to my benchmarks it turned out that the current retval form is OK
### 2. `PixelOperations<T>`
  - `PackFrom***` -> `From***`
  - The last `count` parameter has been removed, we now follow `Span.CopyTo(...)` semantics, copying `source.Length` elements
### 3. Bulk conversion methods on `ColorSpaceConverter`
  - For consistency with `PixelOperations<T>`: The last `count` parameter has been removed, we now follow `Span.CopyTo(...)` semantics, copying `source.Length` elements

## Internal refactor
- All `IPixel` structs and specific `PixelOperations<T>` implementations were moved to a (non-namespace-provider) subfolder `PixelImplementations`
- DRY out all pixel specific generator templates (`***.PixelOperations.Generated.tt`) by using shared code in `_Common.ttinclude`
- New methods and `MethodImpl` tricks in `Guard`


src/ImageSharp/PixelFormats/IPixel.cs
```diff
@@ -24,93 +24,93 @@ public interface IPixel<TSelf> : IPixel, IEquatable<TSelf>
     }
 
     /// <summary>
-    /// An interface that represents a pixel type.
+    /// A base interface for all pixels, defining the mandatory operations to be implemented by a pixel type.
```
JimBobSquarePants: Perfect!

PR #746, opened 2018-10-24
Cross target NET472 and enable extended intrinsics behind symbol

iamcarbon:
### Prerequisites

### Description

This PR cross targets NET472 and introduces a SUPPORTS_EXTENDED_INTRINSICS symbol to enable the new extended intrinsics support on modern runtimes (NET472+ & NETCOREAPP2.1+).

This PR also retarget the NET471 tests to NET472. Extended tests continue to run against NET471.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ environment:
     - target_framework: netcoreapp2.1
       is_32bit: True
 
-    - target_framework: net471
+    - target_framework: net472
       is_32bit: False
 
     - target_framework: net472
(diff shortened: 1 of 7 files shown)


antonfirsov:
@iamcarbon in my original proposal I supposed that the spannified stream API's are also available on .NET FW 4.7.2, but looks like the full unification is out of scope until .NET Standard 2.1 :(

iamcarbon:
@antonfirsov -- Yeah. :( 

Even worse, their latest plan was for .NETCOREAPP3.0 to be the first runtime to implement .NETSTANDARD2.1. And it doesn't look like the desktop framework will ever implement it.

There may never be a full unification.

PR #754, opened 2018-10-26
Imrpoved Gray8 accuracy

antonfirsov:
### Prerequisites

### Description
- Adding `0.5f` in `Get8BitBT709Luminance()` to get a rounded result
    - This makes the `Gray8` -> `Rgba32` -> `Gray8` conversion return the original value to itself for all inputs (31 was an outlier for example)
    - Please let me know if I misunderstood how should this work, but for me it seems much reasonable with the rounding!
- More efficient & more accurate `FromVector4` implementation
- More tests


JimBobSquarePants:
That makes sense me. We're not losing when rounding then.

antonfirsov:
@JimBobSquarePants should we merge this then?

PR #755, opened 2018-10-27
Prevent text wandering from baseline

### LINKED ISSUE #685: Wrapped text "wandering" from baseline on successive lines.
### Prerequisites

### Description
Wrapped text appears to "wander" on successive "wraps"

### Steps to Reproduce
Following code:

``` c#
public static void DrawWrappedText()
{
    Console.WriteLine($"Executing {System.Reflection.MethodBase.GetCurrentMethod().Name}");

    // White filled "4 x 6" (if measured as 200 DPI).
    Image<Rgba32> TestImage = new Image<Rgba32>(800, 1200);
    TestImage.Mutate(x => x.Fill(Rgba32.White));

    // Font F = SystemFonts.CreateFont("Arial", 30, FontStyle.Regular);
    Font F = SystemFonts.CreateFont("Baskerville Old Face", 30, FontStyle.Regular);

    string T = "’Twas brillig, and the slithy toves Did gyre and gimble in the wabe: All mimsy were the borogoves, And the mome raths outgrabe.\"Beware the Jabberwock, my son!  The jaws that bite, the claws that catch!  Beware the Jubjub bird, and shun The frumious Bandersnatch!\"";

    TextGraphicsOptions TGO = new TextGraphicsOptions(true)
    {
        WrapTextWidth = TestImage.Width
    };

    TestImage.Mutate(x => x.DrawText(TGO, T, F, Rgba32.Black, new PointF(10, 50)));

    TestImage.Save("WrappedText.jpg");


}
```
Produced the following:

wandering-text

Zoom into the 4th line of text.  The characters are "wandering" vertically off the baseline.

1st line seems OK, and each successive line seems to get "worse".

Other fonts tried (Arial), but the Baskerville shows issue clearly.,

### System Configuration

- ImageSharp version: 1.0.0-beta0005
- Other ImageSharp packages and versions: SixLabors.ImageSharp; SixLabors.ImageSharp.Drawing
- Environment (Operating system, version and so on): Win10; Visual Studio 2017 v 15.8.0
- .NET Framework version: .Net Core 2.1
- Additional information:

tocsoft:
### Prerequisites

### Description

this fixes #685 

The issue was the fact we where being to aggressive when caching rendered/repeated glyphs.

We now should be allowing approx 4 slightly offset variants of a single glyph (+/- 1 pixel on both axis).

### before
before image with wandering text

### after
after with baseline fixed

### DIFF
--- a/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor.cs
+++ b/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor.cs
@@ -97,7 +97,8 @@ namespace SixLabors.ImageSharp.Processing.Processors.Text
 
             this.textRenderer = new CachingGlyphRenderer(source.GetMemoryAllocator(), this.Text.Length, this.Pen, this.Brush != null);
             this.textRenderer.Options = (GraphicsOptions)this.Options;
-            TextRenderer.RenderTextTo(this.textRenderer, this.Text, style);
+            var renderer = new TextRenderer(this.textRenderer);
+            renderer.RenderText(this.Text, style);
         }
 
         protected override void AfterImageApply(Image<TPixel> source, Rectangle sourceRectangle)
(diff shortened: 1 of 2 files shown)


tocsoft:
@antonfirsov feel free to tweek the tolerances... I never get the level them right.

The reason I made the tolerances exact was because the sample FontShapesAreRenderedCorrectly_LargeText.png did show the issue (its just more prominent with the other font)... 'tis why I needed to update the reference image too now I had resolved the problem.

JimBobSquarePants:
> @antonfirsov feel free to tweek the tolerances... I never get the level them right.

I commented about this in Gitter the other week. I find the tolerance comparer confusing. 

Current the default threshold is `3.921569E-07` calculated as `((1 / 255F) * 65535) / (100 * 100 * 65535)` which is defined as

 _1% of all pixels in a 100*100 pixel area are allowed to have a difference of 1 unit_ 

That's amazing technically that we can calculate this but far too complicated to use. 

I want to simply pass something like 2% and if the image is more than 2% different then it fails.

JimBobSquarePants:
@antonfirsov Ok that makes more sense!

antonfirsov:
@JimBobSquarePants actually, we can simplify the comparison and the way we parameterize it by rebasing it on scaled `Vector4` pixel values.

tocsoft:
Differences are fine and expected... It's differences that effect the visual baseline that are the problem... Which is something that is nigh on impossible to quantify and test for.

antonfirsov:
@tocsoft if we allow higher differences with `AccuracyMultiple = 2` the "wandering" is visually well visible on the output images (check the images in my comment). 

The smaller the reported difference is, the more correct the output feels, so I think the difference ratio is a good quantitative value to benchmark output correctness.

antonfirsov:
Or maybe I'm missing the point of your comment? 😕

tocsoft:
That's strange because with a 2 it wasn't wandering on my machine... Odd  yeah 8 is fine then... I believe it's going to be one of those tweek the settings as/if people complain that it's still happening.

JimBobSquarePants:
Is the error a point of origin calculation issue?

antonfirsov:
Anything against merging this now?

Issue #763, opened 2018-10-30
Rgb image with transparency color will have transparency removed when being saved


devedse:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

When simply opening and saving the image below all transparency is removed.

Input:
![snake](https://user-images.githubusercontent.com/2350015/47756148-3ca39900-dca1-11e8-88a1-c168db9f1dc7.png)

Output:
![snake_output](https://user-images.githubusercontent.com/2350015/47756174-53e28680-dca1-11e8-8570-971242eb4e50.png)

### Steps to Reproduce

The following code reproduces the problem:
```
using (var img = Image.Load(image1path))
{
    using (var fs = new FileStream(outputImage, FileMode.Create))
    {
        img.SaveAsPng(fs);
    }
}
``` 

### Investigation into issue

To help you guys out a bit I did some investigation myself to see where the issue comes from. First of all I did see that my image was detected as an Rgb image but it still had a transparent title bar in it.

PNG Analyzer shows a Transparency header though:
![image](https://user-images.githubusercontent.com/2350015/47756281-bdfb2b80-dca1-11e8-9099-089f9956b89d.png)

Doing some debugging I finally found out what was causing the problem. When the image is loaded a bool: 'hasTrans' is set to true and the transparent color is set to 211, 211, 211. This basically means that everywhere in the image where the color 211, 211, 211 is used it will be rendered as transparent.

However the issue occurs when we save the image again. The hasTrans flag is not taken into account when saving the image and thus it will simply be stored as Rgb without any data about transparency.

### Possible resolutions

As far as I can think of right now there's 2 possible resolutions:
1. Store transparency data on the image metadata and use this when writing the image again
2. If the data of an Rgb image contains a transparency value set to 0, save the image as Rgba32

I guess that both solution still have quite a bit of work to do though.

### System Configuration

- ImageSharp version: Github version as of 31-10-2018 and also Myget 1.0.0-dev002131
- Other ImageSharp packages and versions:
- Environment (Operating system, version and so on): Win10
- .NET Framework version: .NET Core 2.1
[...]

PR #764, opened 2018-10-30
[WIP] Preserving isTrans data

devedse:
### Prerequisites

### Description

Currently still WIP but I wanted to contribute something to the project :smile:. So I'm working on fixing my own issue:
#763


devedse:
I'm not sure if we want to implement it this way or have a look at the already existing Palette code which should be extended to always write alpha if the hasAlpha flag is true. But please let me know :)

JimBobSquarePants:
Hey @devedse this all looks great! Thanks for making the effort to fix this for us. 

I'll pull it down for a final review but from looking on GitHub I can't see anything amiss. 

In the interim could you please sign the CLA? Cheers! 👍

devedse:
I think the bitconversion could use some improvements

src/ImageSharp/Formats/Png/PngMetaData.cs
```diff
@@ -42,6 +44,31 @@ private PngMetaData(PngMetaData other)
         /// </summary>
         public float Gamma { get; set; }
 
+        /// <summary>
+        /// Gets or sets the Rgb 24 transparent color. This represents any color in an 8 bit Rgb24 encoded png that should be transparent
+        /// </summary>
+        public Rgb24? Rgb24Trans { get; set; }
```
antonfirsov: I think we should not use abbrevations in public API-s. `TransparentRgb24`? @JimBobSquarePants any better idea?

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -300,6 +305,52 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
...
+        /// <summary>
+        /// Writes the transparency markers to the stream
+        /// </summary>
+        /// <param name="stream">The <see cref="Stream"/> containing image data.</param>
+        /// <param name="pngMetaData">The image meta data.</param>
+        private void WriteTransparencyMarkers(Stream stream, PngMetaData pngMetaData)
+        {
+            if (pngMetaData.ColorType == PngColorType.Rgb)
+            {
+                if (pngMetaData.Rgb48Trans != null)
```
JimBobSquarePants: We should be checking the bit depth here also.

src/ImageSharp/MetaData/ImageMetaData.cs
```diff
@@ -5,6 +5,7 @@
 using SixLabors.ImageSharp.Formats;
 using SixLabors.ImageSharp.MetaData.Profiles.Exif;
 using SixLabors.ImageSharp.MetaData.Profiles.Icc;
+using SixLabors.ImageSharp.PixelFormats;
```
JimBobSquarePants: I don't think we need this.

JimBobSquarePants:
@antonfirsov To answer your questions regarding the tRNS chunk.

When decoding either grayscale,  24bit or 48bit color types, if the chunk is encountered we decode to `Rgba32` and `Rgba64` pixel formats (we don't have a GrayAlpha pixel format) and set the alpha component to fully opaque for any pixel that does not match the value contained within the chunk.

When encoding we use the `Rgb24`, `Rgb48`, `Gray8`, or `Gray16` pixel format layouts and assign a color that should be regarded as transparent to the chunk.

Does that make sense?

src/ImageSharp/MetaData/ImageMetaData.cs
devedse: Fixed

devedse: Fixed

src/ImageSharp/Formats/Png/PngEncoderCore.cs
devedse: I'm not exactly sure how to do that, might be better if you implement this.

src/ImageSharp/Formats/Png/PngMetaData.cs
devedse: Fixed

src/ImageSharp/Formats/Png/PngMetaData.cs
```diff
@@ -42,6 +44,31 @@ private PngMetaData(PngMetaData other)
...
+
+        /// <summary>
+        /// Gets or sets the 8 bit grayscale transparent color. This represents any color in an 8 bit grayscale encoded png that should be transparent
+        /// </summary>
+        public byte? LuminanceTrans { get; set; }
+
+        /// <summary>
+        /// Gets or sets the 16 bit grayscale transparent color. This represents any color in a 16 bit grayscale encoded png that should be transparent
+        /// </summary>
+        public ushort? Luminance16Trans { get; set; }
```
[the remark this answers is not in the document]
devedse: Fixed

src/ImageSharp/Formats/Png/PngMetaData.cs
```diff
@@ -42,6 +44,31 @@ private PngMetaData(PngMetaData other)
...
+
+        /// <summary>
+        /// Gets or sets the Rgb 48 transparent color. This represents any color in a 16 bit Rgb24 encoded png that should be transparent
+        /// </summary>
+        public Rgb48? Rgb48Trans { get; set; }
+
+        /// <summary>
+        /// Gets or sets the 8 bit grayscale transparent color. This represents any color in an 8 bit grayscale encoded png that should be transparent
+        /// </summary>
+        public byte? LuminanceTrans { get; set; }
```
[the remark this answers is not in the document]
devedse: Fixed

devedse:
I've implemented almost all findings, please redo code review :)

antonfirsov:
> and set the alpha component to fully opaque for any pixel that does not match the value contained within the chunk

But what happens when we encounter pixels matching the value? Do we decode fully transparent pixels? Do we alter transparent pixels to chunk-defined pixels when encoding?

JimBobSquarePants:
@antonfirsov 

>But what happens when we encounter pixels matching the value? Do we decode fully transparent pixels in those cases?

We set the alpha component to 0 leaving the other component data intact.

>Do we alter transparent pixels to chunk-defined pixels when encoding?

No the tRNS chunk is simply a marker. Pixels are encoded as normal. Delete the tRNS chunk and you have an opaque image.

@devedse 

We have some test images already that might cover your needs but there's an official suite here.

http://www.schaik.com/pngsuite/

devedse:
> @devedse
> 
> We have some test images already that might cover your needs but there's an official suite here.
> 
> http://www.schaik.com/pngsuite/

What do you mean by this? For writing test cases?

antonfirsov:
@devedse input images for decoder tests.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
JimBobSquarePants: Actually I think it's ok as we already throw if the combination is invalid.

JimBobSquarePants:
Tests added and correct behaviour confirmed. I think this is good to go now.

PR #765, opened 2018-11-01
Optimize filling a region with a solid brush when antialias is off

cwensley:
### Prerequisites

### Description

This optimizes filling a solid region when not using antialiasing, which is similar to the optimization done for the FillProcessor [here](https://github.com/SixLabors/ImageSharp/blob/2979267c055568ccdc790327a60ba340858f676a/src/ImageSharp.Drawing/Processing/Processors/Drawing/FillProcessor.cs#L59-L73).  The following sample performs 4.2x faster with these changes by avoiding the use of the applicator:

```c#
Image<Rgba32> image = new Image<Rgba32>(Configuration.Default, 10000, 10000, new Rgba32(30, 35, 36));

// wouldn't normally fill the entire image, but it exacerbates the issue
image.Mutate(x => x.Fill(new GraphicsOptions(false), Rgba32.White, new RectangleF(0, 0, image.Width, image.Height)));
```

Please let me know if you want any additional tests or changes.  I think there may be other ways to optimize the code that fills the scanline when not using antialiasing, but this is hopefully a great start.

Some benchmark results:

|               Method |        Mean |       Error |     StdDev | Scaled | ScaledSD |
|--------------------- |------------:|------------:|-----------:|-------:|---------:|
|                 Fill |    44.57 ms |    15.53 ms |  0.8772 ms |   1.00 |     0.00 |
| FillWithClip(before) | 4,701.17 ms | 1,113.41 ms | 62.9096 ms | 105.52 |     2.03 | 
| FillWithClip(after)  | 1,073.93 ms | 258.06 ms   | 14.5806 ms |  23.61 |     0.46 |


src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor.cs
```diff
@@ -168,17 +168,31 @@ protected override void OnFrameApply(ImageFrame<TPixel> source, Rectangle source
...
+                                        hasOnes = true;
                                     }
                                     else
                                     {
                                         scanline[x] = 0;
+                                        hasZeros = true;
                                     }
                                 }
+
+                                if (hasOnes != hasZeros && this.IsSolidBrushWithoutBlending(out SolidBrush<TPixel> solidBrush))
```
tocsoft: `this.IsSolidBrushWithoutBlending(out SolidBrush<TPixel> solidBrush)` should be only done once outside the loops as it can't change between lines

cwensley: Good point, thanks!  I'll update it.

cwensley: Sorry it took a while (I don't work on this project often), the requested changes are now done!

PR #770, opened 2018-11-06
Fix infinite loop when a GIF prematurely terminates

iamcarbon:
### Prerequisites

### Description

When the last frame of a GIF isn't properly terminated, Skip loops forever.

This ensures we break at the EOF.

This PR also renames Skip to SkipBlock to better describe it's behavior.


iamcarbon:
Single unrelated GDI error on 32bit. Otherwise, all tests pass.

iamcarbon:
Ready for review.

iamcarbon:
I created a minimal test case that hits this path (without adding the original 10MB+ gifs that trigger this case).

PR #771, opened 2018-11-06
Avoid doing the same thing multiple times

jongleur1983:
Small code optimizations, all reducing the operations to be done less often by reusing intermediate results instead of doing the same calculations multiple times.

### DIFF
--- a/src/ImageSharp/ColorSpaces/Conversion/Implementation/Converters/HunterLabToCieXyzConverter.cs
+++ b/src/ImageSharp/ColorSpaces/Conversion/Implementation/Converters/HunterLabToCieXyzConverter.cs
@@ -26,9 +26,12 @@ namespace SixLabors.ImageSharp.ColorSpaces.Conversion.Implementation
             float ka = ComputeKa(input.WhitePoint);
             float kb = ComputeKb(input.WhitePoint);
 
-            float y = ImageMaths.Pow2(l / 100F) * yn;
-            float x = (((a / ka) * MathF.Sqrt(y / yn)) + (y / yn)) * xn;
-            float z = (((b / kb) * MathF.Sqrt(y / yn)) - (y / yn)) * (-zn);
+            float pow = ImageMaths.Pow2(l / 100F);
+            float sqrtPow = MathF.Sqrt(pow);
+            float y = pow * yn;
+
+            float x = (((a / ka) * sqrtPow) + pow) * xn;
+            float z = (((b / kb) * sqrtPow) - pow) * (-zn);
 
             return new CieXyz(x, y, z);
         }
--- a/src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
+++ b/src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
@@ -310,8 +310,8 @@ namespace SixLabors.ImageSharp.Formats.Bmp
                 }
                 else
                 {
-                    int max = cmd[0] + count;
-                    byte cmd1 = cmd[1];
+                    int max = count + cmd[0]; // as we start at the current count in the following loop, max is count + cmd[0]
+                    byte cmd1 = cmd[1]; // store the value to avoid the repeated indexer access inside the loop
 
                     for (; count < max; count++)
                     {
--- a/src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
+++ b/src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
@@ -310,9 +310,12 @@ namespace SixLabors.ImageSharp.Formats.Bmp
                 }
                 else
                 {
-                    for (int i = 0; i < cmd[0]; i++)
+                    int max = cmd[0] + count;
+                    byte cmd1 = cmd[1];
+
+                    for (; count < max; count++)
                     {
-                        buffer[count++] = cmd[1];
+                        buffer[count] = cmd1;
                     }
                 }
             }
--- a/src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
@@ -543,14 +543,13 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
         {
             var sb = new StringBuilder();
             sb.Append('[');
-            for (int i = 0; i < Size; i++)
+            for (int i = 0; i < Size - 1; i++)
             {
                 sb.Append(this[i]);
-                if (i < Size - 1)
-                {
-                    sb.Append(',');
-                }
+                sb.Append(',');
             }
+
+            sb.Append(this[Size - 1]);
 
             sb.Append(']');
             return sb.ToString();
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -856,10 +856,9 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
         private void ProcessStartOfScanMarker()
         {
             int selectorsCount = this.InputStream.ReadByte();
-            int componentIndex = -1;
             for (int i = 0; i < selectorsCount; i++)
             {
-                componentIndex = -1;
+                int componentIndex = -1;
                 int selector = this.InputStream.ReadByte();
 
                 for (int j = 0; j < this.Frame.ComponentIds.Length; j++)
--- a/src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
@@ -822,11 +822,12 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
             {
                 for (int i = 0; i < componentCount; i++)
                 {
-                    this.buffer[(3 * i) + 6] = (byte)(i + 1);
+                    int i3 = 3 * i;
+                    this.buffer[i3 + 6] = (byte)(i + 1);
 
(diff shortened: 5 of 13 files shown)


src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -310,9 +310,12 @@ private void UncompressRle8(int w, Span<byte> buffer)
...
                 else
                 {
-                    for (int i = 0; i < cmd[0]; i++)
+                    int max = cmd[0] + count;
+                    byte cmd1 = cmd[1];
+
+                    for (; count < max; count++)
                     {
-                        buffer[count++] = cmd[1];
+                        buffer[count] = cmd1;
```
iamcarbon: Use Span.Fill?

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -310,9 +310,12 @@ private void UncompressRle8(int w, Span<byte> buffer)
                 }
                 else
                 {
-                    for (int i = 0; i < cmd[0]; i++)
+                    int max = cmd[0] + count;
```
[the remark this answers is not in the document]
jongleur1983: the loop went from 0 to <cmd[0] before, but inside the loop count was incremented as well.
instead, we co from count (not initializing, thus keeping the old value) to <(cmd[0]-{initialCount}).

adding count to the upper bound of the loop keeps the number of iterations the same as before but allows to use count as the only variable we have to increment.
Not adding count would iterate from count=x (where x is determined before in the algorithm) to cmd[0] which is not what should happen here.

src/ImageSharp/ColorSpaces/Conversion/Implementation/Converters/HunterLabToCieXyzConverter.cs
```diff
@@ -26,9 +26,12 @@ public CieXyz Convert(in HunterLab input)
...
             float kb = ComputeKb(input.WhitePoint);
 
-            float y = ImageMaths.Pow2(l / 100F) * yn;
-            float x = (((a / ka) * MathF.Sqrt(y / yn)) + (y / yn)) * xn;
-            float z = (((b / kb) * MathF.Sqrt(y / yn)) - (y / yn)) * (-zn);
+            float pow = ImageMaths.Pow2(l / 100F);
+            float sqrtPow = MathF.Sqrt(pow);
+            float y = pow * yn;
+
+            float x = (((a / ka) * sqrtPow) + pow) * xn;
```
dlemstra: Are you fixing a bug here or creating one? `sqrtPow != MathF.Sqrt(y / yn)` and `pow != y / yn`.

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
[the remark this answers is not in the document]
dlemstra: Maybe you could reverse the order to make it more clear that you are adding `cmd[0]` to `count`?

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
```diff
@@ -543,15 +543,14 @@ public override string ToString()
         {
             var sb = new StringBuilder();
             sb.Append('[');
-            for (int i = 0; i < Size; i++)
+            for (int i = 0; i < Size - 1; i++)
```
dlemstra: Maybe you could also 'calculate' `Size-1` once?

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -856,10 +856,9 @@ private void ProcessDefineRestartIntervalMarker(int remaining)
         private void ProcessStartOfScanMarker()
         {
             int selectorsCount = this.InputStream.ReadByte();
-            int componentIndex = -1;
             for (int i = 0; i < selectorsCount; i++)
             {
-                componentIndex = -1;
+                int componentIndex = -1;
```
dlemstra: Good catch 👍

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -822,11 +822,12 @@ private void WriteStartOfFrame(int width, int height, int componentCount)
             {
                 for (int i = 0; i < componentCount; i++)
                 {
-                    this.buffer[(3 * i) + 6] = (byte)(i + 1);
+                    int i3 = 3 * i;
```
dlemstra: Maybe you could just increment `i3` instead of doing the multiplication? You then also don't need the i3+6/7/8?

src/ImageSharp/Processing/ResizeHelper.cs
```diff
@@ -337,21 +337,19 @@ private static (Size, Rectangle) CalculateMinRectangle(Size source, ResizeOption
...
             {
-                destinationHeight = (int)MathF.Round(width * sourceRatio);
+                destinationHeight = (int)MathF.Round(width * ((float)sourceHeight / sourceWidth));
                 height = destinationHeight;
                 destinationWidth = width;
             }
             else if (widthDiff > heightDiff)
             {
-                destinationWidth = (int)MathF.Round(height / sourceRatio);
+                destinationWidth = (int)MathF.Round(height * ((float)sourceWidth / sourceHeight));
```
dlemstra: I think I prefer the old method. What are you fixing here?

src/ImageSharp/ColorSpaces/Conversion/Implementation/Converters/HunterLabToCieXyzConverter.cs
jongleur1983: `pow = ImageMaths.Pow2(l/100F)` // new intermediate variable
y was `ImageMaths.Pow2(l/100F) * yn`. Replace the first term by pow and you get
`y = pow*yn` // new assignment to y, line 31.

for x:
original: `x = (((a/ka) * Sqrt(y/yn)) + (y/yn)) * xn`
as pow is the old `y/yn` (that's why for the new y we multiply `pow * yn` now), we can replace this to
tmp1: `x = (((a/ka) * Sqrt(pow)) + pow) * xn`
as Sqrt(pow) is needed again, we replace that as well, so we end up with what I defined.

for z the same pattern as for x applies.

Or in Short according to your remark:
```
sqrtPow  
= MathF.Sqrt(pow) 
= MathF.Sqrt(ImagePath.Pow2(l/100)) 
= MathF.Sqrt(ImageMath.Pow2(l/100)*yn / yn)) 
= MathF.Sqrt(y/yn)
```

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
jongleur1983: I could, but this is done by the compiler: Size is constant, so Size-1 is constant, and in fact the compiler simplifies the term.
Thus it's not only calculated once, it's calculated at compile time and the intermediate variable, used as VARIABLE would even harm performance instead.

jongleur1983: minimized example on what the compiler does (even in debug mode, but release is sligthly more concise): https://sharplab.io/#v2:EYLgHgbALANALiAlgGwD4AEDMACdBGCXAJmwGEBYAKAG8rt7sBjAewDsBnObRVrgZUQAvAKbYAvNmgBuOgyy4CuKNgCyAQx4AKAA5qATmoC27BQAYA2gF1s+gObsAlLPq1KDd9gBmzPZp5dEMVMpbgAeAREAWjwQxABqOIdsVw8GAF9nbAzKNKA=

you see the constant value 63 in IL_0009 being loaded as ldc.i4.s 63

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
jongleur1983: I thought about that first, but i is used itself in liness 829 and 830 again.
We could store the iterations initial i as iPrime or whatever, or store the subsamples[i] and chroma[i] first, but it doesn't make it better,
so I decided to keep it as less invasive as possible and would like not to do more in this PR alone.
Feel free to add further improvements.

src/ImageSharp/Processing/ResizeHelper.cs
jongleur1983: I agree this is a very minor performance improvement on probably less readability.
sourceRatio is calculated always, but only used when widthDiff != heightDiff (the third case doesn't use sourceRatio).
Inlining improves that special case, while the other cases are not affected negatively.

If it's not wanted I'm fine with undoing this again ;)

src/ImageSharp/ColorSpaces/Conversion/Implementation/Converters/HunterLabToCieXyzConverter.cs
dlemstra: That is a good catch. And thanks for clearing it up. Missed that you optimized the division away.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
dlemstra: Thanks for the explanation.

src/ImageSharp/Processing/ResizeHelper.cs
dlemstra: It is indeed a tiny optimization. Maybe put the division in a variable inside to make it more clear that this is the aspect ration? You can also move `percentHeight` and `percentWidth`. These are only used at one spot.

jongleur1983: done.

PR #772, opened 2018-11-06
Fix naming of variable and class

jongleur1983:
fixes typos in class name (and usages of that class) as well as for a local variable in code.


jongleur1983:
what's travis doing here? don't understand the error and/or don't think it's related in any way to these code changes.

dlemstra:
Thanks for the fixes 👍

PR #773, opened 2018-11-06
simplify BoxBlurProcessor kernel initialization

jongleur1983:
old code seemed to be a relict from the GaussianBlurProcessor where this pattern made sense.
Here each value of the matrix is just 1.0/size as a result, and we can leverage DenseMatrix.Fill() for that.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Convolution/BoxBlurProcessor.cs
+++ b/src/ImageSharp/Processing/Processors/Convolution/BoxBlurProcessor.cs
@@ -66,36 +66,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Convolution
                 ? new DenseMatrix<float>(size, 1)
                 : new DenseMatrix<float>(1, size);
 
-            float sum = 0F;
-            for (int i = 0; i < size; i++)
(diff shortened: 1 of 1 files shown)


src/ImageSharp/Processing/Processors/Convolution/BoxBlurProcessor.cs
```diff
@@ -66,36 +66,7 @@ private DenseMatrix<float> CreateBoxKernel(bool horizontal)
...
-                }
-            }
-            else
-            {
-                for (int i = 0; i < size; i++)
-                {
-                    kernel[i, 0] = kernel[i, 0] / sum;
-                }
-            }
+            kernel.Fill(1.0F / size);
```
JimBobSquarePants: Very silly of me eh? 😄

jongleur1983: nobody is perfect - and when changing existing code it's often hard to see when something get's simple enough to collapse in complexity.
Let's just call your version here "legacy code" and everything is forgiven ;)

PR #784, opened 2018-12-06
Jpeg 12 bit support

feliwir:
### Description
- Adds 12-bit jpeg support as implemented in `libjpeg` and `libjpeg-turbo`
- Adds a 12-bit jpeg testfile
- Make jpeg decoder bitdepth invariant


src/ImageSharp/Formats/Jpeg/Components/Block8x8F.Generated.tt
```diff
@@ -59,10 +56,14 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
         }
 
 		/// <summary>
-        /// Level shift by +128, clip to [0, 255]
+        /// Level shift by +maximum/2, clip to [0, maximum]
         /// </summary>
-        public void NormalizeColorsInplace()
+        public void NormalizeColorsInplace(float maximum)
         {
+            Vector4 CMin4 = new Vector4(0F);
```
JimBobSquarePants: These values are being calculated more than once in the same partial struct.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.cs
```diff
@@ -22,28 +22,58 @@ internal abstract partial class JpegColorConverter
...
+        /// </summary>
+        public int Precision { get; }
+
+        /// <summary>
+        /// Gets the maximum value of a sample
+        /// </summary>
+        private float MaximumValue { get; }
+
+        /// <summary>
+        /// Gets the maximum value of a sample
```
JimBobSquarePants: Repeated description

JimBobSquarePants:
It occurs to me that this adds decoding support not encoding so the image data will be lost. We should get the work done to allow supporting encoding also.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.Generated.tt
antonfirsov: What we can do is to refactor the normalization logic into it's own stateful type (preferably a struct), where minimum/maximum/half values are nonstatic members initialized in the constructor.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromCmyk.cs
```diff
@@ -10,8 +10,8 @@ internal abstract partial class JpegColorConverter
     {
         internal class FromCmyk : JpegColorConverter
         {
-            public FromCmyk()
-                : base(JpegColorSpace.Cmyk)
+            public FromCmyk(int precision)
```
antonfirsov: I prefer to have more type safety in the codebase. (`-1`, `17` and `222` are invalid values for precision ;) ). I suggest an enum having the following form, so you can easily cast it to `int` when necessary.

```C#
enum JpegPrecision
{
   EightBits = 8,
   TwelveBits = 12
}
```

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
```diff
@@ -44,7 +44,7 @@ public JpegColorConverterTests(ITestOutputHelper output)
         public void ConvertFromYCbCrBasic(int inputBufferLength, int resultBufferLength, int seed)
         {
             ValidateRgbToYCbCrConversion(
-                new JpegColorConverter.FromYCbCrBasic(),
+                new JpegColorConverter.FromYCbCrBasic(8),
```
antonfirsov: We should add tests for 12 bit precision case.

JimBobSquarePants:
@antonfirsov 
> Is it possible to have 12 bit precision jpegs for colorspaces other than YCbCr? (eg. CMYK, YCCK, RGB)

Yes, when 12bit support was added to the jpeg specification it covered all color spaces. On that note it occurs to me that we cannot preserve that color data when saving the jpeg. We'll need to add the equivalent colorspace transformation code to the encoder with any additional requirements.

antonfirsov:
@JimBobSquarePants then we should find or generate 12 bit test jpegs for all the colorspaces, the feature is only partially tested otherwise!
We can decode the reference output with the ImageMagick reference decoder. The encoder could be implemented/tested separately.

JimBobSquarePants:
@antonfirsov Just spent the last 20 mins scouring the internet for an official source of test images and I've come up with nothing. libjpeg-turbo's repo contains 1 12bit file that doesn't appear to be supported by anything I have on my machine. https://github.com/libjpeg-turbo/libjpeg-turbo/tree/master/testimages

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromCmyk.cs
JimBobSquarePants: Hmmm... Read section 4.7 here. Lossless jpeg supports additional precision so we should bear that in mind when naming our enum. https://www.w3.org/Graphics/JPEG/itu-t81.pdf

> For DCT-based processes, two alternative sample precisions are specified: either 8 bits or 12 bits per sample. Applications which use samples with other precisions can use either 8-bit or 12-bit precision by shifting their source image samples appropriately. The baseline process uses only 8-bit precision. DCT-based implementations which handle 12-bit source image samples are likely to need greater computational resources than those which handle only 8-bit source images. Consequently in this Specification separate normative requirements are defined for 8-bit and 12-bit DCT-based processes.
> 
> For lossless processes the sample precision is specified to be from 2 to 16 bits

antonfirsov: Oops. The options I see:
1. Keep the type as is (`int`), and introduce some `DebugGuard` checks in the code
2. Introduce an enum defining all entries for 2-16
3. Since we do not support lossless jpeg, define only `EightBits` and `TwelveBits` for now, and YAGNI out the rest, letting our future selves to find out how to manage the lossless case.

I'm pretty much for option 3.

antonfirsov:
Ok, this means that those aren't real use cases on pure statistical basis, so we can leave them untested.

Standard meets reality.

JimBobSquarePants:
@feliwir @antonfirsov I had some time so I updated everything in the branch, removed some of the Linq stuff from the color converter and merged the reference image. 

We're getting very close to having this ready to merge 🎉 

Question though. Shall we had encoding 12bit jpegs to this PR or open a new Issue/PR. We'll need to create optimized color conversion methods to handle this.

feliwir:
We could add encoding, but not sure if i could help with that too much :D

JimBobSquarePants:
@antonfirsov When you have a chance can you please conduct a final review?

JimBobSquarePants:
Merged. Thanks @feliwir !

Issue #776, opened 2018-11-14
Load a png on IOS causes an "Attempting to JIT compile method while running in aot-only mode"


vicfergar:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When trying to load a PNG image on iOS 12 an unhandled exception is raised.

### Steps to Reproduce
On an Xamarin.IOS project try to invoke:
`var image = SixLabors.ImageSharp.Image.Load(stream);`

The next exception will be raised:
```
System.ExecutionEngineException: Attempting to JIT compile method 'SixLabors.ImageSharp.Formats.Png.PngScanlineProcessor:ProcessRgbScanline<SixLabors.ImageSharp.PixelFormats.Rgba32> (SixLabors.ImageSharp.Configuration,SixLabors.ImageSharp.Formats.Png.PngHeader&,System.ReadOnlySpan`1<byte>,System.Span`1<SixLabors.ImageSharp.PixelFormats.Rgba32>,int,int,bool,SixLabors.ImageSharp.PixelFormats.Rgb48,SixLabors.ImageSharp.PixelFormats.Rgb24)' while running in aot-only mode. See https://developer.xamarin.com/guides/ios/advanced_topics/limitations/ for more information.

  at SixLabors.ImageSharp.Formats.Png.PngDecoderCore.ProcessDefilteredScanline[TPixel] (System.ReadOnlySpan`1[T] defilteredScanline, SixLabors.ImageSharp.ImageFrame`1[TPixel] pixels) [0x000db] in <eb47c90af2ba451c9c07fe98c40e7810>:0 
  at (wrapper unknown) System.Object.gsharedvt_in()
  at (wrapper unknown) System.Object.gsharedvt_out()
  at SixLabors.ImageSharp.Formats.Png.
PngDecoderCore.DecodePixelData[TPixel] (System.IO.Stream compressedStream, SixLabors.ImageSharp.ImageFrame`1[TPixel] image) [0x000e5] in <eb47c90af2ba451c9c07fe98c40e7810>:0 
  at (wrapper unknown) System.Object.gsharedvt_in()
[...]

PR #785, opened 2018-12-08
Add more seeding to the AoT compiler for iOS

swoolcock:
### Prerequisites

### Description

Adds some calls to `Unsafe.SizeOf` for various primitives and structs to prevent JIT errors on iOS.
Also seeds the decoders for a given pixel format, which should solve #776.
Note that although JPEG will now also load without crashing, it is incredibly slow on Xamarin.iOS.  A 1366x768 JPEG takes about 5 seconds to load on an iPad Pro.

As per the previous PR #767, this is a difficult one to write tests for.

### DIFF
--- a/src/ImageSharp/Advanced/AotCompilerTools.cs
+++ b/src/ImageSharp/Advanced/AotCompilerTools.cs
@@ -43,10 +43,10 @@ namespace SixLabors.ImageSharp.Advanced
 
             System.Runtime.CompilerServices.Unsafe.SizeOf<TPixel>();
 
-            AotDecoder<TPixel>(new Formats.Png.PngDecoder());
-            AotDecoder<TPixel>(new Formats.Bmp.BmpDecoder());
-            AotDecoder<TPixel>(new Formats.Gif.GifDecoder());
-            AotDecoder<TPixel>(new Formats.Jpeg.JpegDecoder());
+            AotCodec<TPixel>(new Formats.Png.PngDecoder(), new Formats.Png.PngEncoder());
+            AotCodec<TPixel>(new Formats.Bmp.BmpDecoder(), new Formats.Bmp.BmpEncoder());
+            AotCodec<TPixel>(new Formats.Gif.GifDecoder(), new Formats.Gif.GifEncoder());
+            AotCodec<TPixel>(new Formats.Jpeg.JpegDecoder(), new Formats.Jpeg.JpegEncoder());
 
             // TODO: Do the discovery work to figure out what works and what doesn't.
         }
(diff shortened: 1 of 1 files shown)


src/ImageSharp/Advanced/AotCompilerTools.cs
```diff
@@ -99,5 +123,22 @@ private static void AotCompileDithering<TPixel>()
...
+        /// <param name="decoder">The image decoder to seed..</param>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        private static void AotDecoder<TPixel>(IImageDecoder decoder)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            try
+            {
+                decoder.Decode<TPixel>(Configuration.Default, null);
+            }
+            catch
```
antonfirsov: This isn't nice, but can't think of a better solution.

src/ImageSharp/Advanced/AotCompilerTools.cs
```diff
@@ -15,6 +18,20 @@ namespace SixLabors.ImageSharp.Advanced
     /// </summary>
     public static class AotCompilerTools
     {
+        /// <summary>
+        /// Seeds the <see cref="System.Runtime.CompilerServices.Unsafe"/> calls.
+        /// </summary>
+        public static void SeedUnsafe()
```
antonfirsov: Looks like an implementation detail from the outside.

Can't we hide this from the outside invoking it internally in a lazy way when `Seed<TPixel>` (and other variants) are invoked?

src/ImageSharp/Advanced/AotCompilerTools.cs
swoolcock: Yeah I agree.  Would be nice if there were a way to make a dummy call to `Decode` without generating exceptions.

src/ImageSharp/Advanced/AotCompilerTools.cs
swoolcock: I'll lazy call this (once) from `Seed<TPixel>` and push.

swoolcock: Calling it from the static constructor instead, since that's guaranteed to only execute once.

src/ImageSharp/Advanced/AotCompilerTools.cs
```diff
@@ -99,5 +120,22 @@ private static void AotCompileDithering<TPixel>()
             TPixel pixel = default;
             test.Dither<TPixel>(new ImageFrame<TPixel>(Configuration.Default, 1, 1), pixel, pixel, 0, 0, 0, 0, 0, 0);
         }
+
+        /// <summary>
+        /// This method pre-seeds the decoder for a given pixel format in the AoT compiler for iOS.
+        /// </summary>
+        /// <param name="decoder">The image decoder to seed..</param>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        private static void AotDecoder<TPixel>(IImageDecoder decoder)
```
antonfirsov: Isn't it worth to also seed both decoder and encoder instead?

```C#
private static void SeedCodecs<TPixel>(IImageDecoder decoder, IImageEncoder encoder)
{
   // invoke both decoder and encoder in separate try-catch blocks
}
```

swoolcock:
I'm not sure what the deal is with that CI error, I can't see how any of my changes could have triggered it.

antonfirsov:
It was a typical intermittent test failure, related to test-only code. Restarting the build helped.

JimBobSquarePants:
@antonfirsov Nope. I'll update from master and get it merged. @swoolcock Thanks for your help!

PR #789, opened 2018-12-12
Use correct Kernel dimensions and optimize. Fix #786

### LINKED ISSUE #786: Gaussian Blur is not actually Gaussian
```csharp
using (var image = new Image<Rgba32>(128, 128))
{
	image.Mutate(ctx => ctx
		.Fill(Rgba32.Black)
		.Fill(Rgba32.White, new EllipsePolygon(64, 64, 64, 64)));

	image.Mutate(ctx => ctx.GaussianBlur(12));

	using (var stream = new FileStream("image.png", FileMode.Create))
	{
		image.SaveAsPng(stream);
	}
}
```

Mutating an image with a Gaussian blur yields a blur that isn't quite Gaussian, much more closely resembling a box blur. I implemented a mathematically accurate Gaussian blur that factors in every pixel in the image (obviously overkill and inefficient in a 0-255 color space, but completely correct), and the results are as follows:

image

Note the similarity between ImageSharp's Gaussian blur and its box blur.

I would fix this myself and submit a PR but I genuinely have no clue why this happens. All of the processing regarding the generation of the kernel and the underlying Gaussian function seems to be correct and yields the same values as my own implementation.

Notes: Using ImageSharp v1.0.0-beta0005 on .NET Framework 4.7.1 on a Windows 10 64-bit OS.

JimBobSquarePants:
### Prerequisites

### Description
Our current gaussian kernels are too small and fail to capture the full curve leading to boxed output. I've updated the algorithm based on the information found at http://chemaguerra.com/gaussian-filter-radius/ to correct the output.

### DIFF
--- a/src/ImageSharp/Primitives/DenseMatrix{T}.cs
+++ b/src/ImageSharp/Primitives/DenseMatrix{T}.cs
@@ -110,7 +110,7 @@ namespace SixLabors.ImageSharp.Primitives
         /// <returns>The <see typeparam="T"/> at the specified position.</returns>
         public ref T this[int row, int column]
         {
-            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            [MethodImpl(InliningOptions.ShortMethod)]
             get
             {
                 this.CheckCoordinates(row, column);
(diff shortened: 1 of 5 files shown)


JimBobSquarePants:
> We should probably add at least one reference-image based test case to ensure correctness, with a reference image generated by CSS.

@antonfirsov  I would but CSS blur [over-samples](https://github.com/SixLabors/ImageSharp/issues/786#issuecomment-446451749) past the edge of the image so we can't compare. 

Additionally we use a constrained kernel when sampling, capturing 97% of the Gaussian integral but ignoring changes generally invisible to the human eye when compared to an infinite kernel.

PR #791, opened 2018-12-15
WIP: Adds support for Decoding Bitmaps with 16 bytes headers

brianpopow:
### Prerequisites

### Description

This MR adds the support for decoding Bitmaps with a 16 bytes header, a short variant of the OS22XBITMAPHEADER. It is identical to the BITMAPCOREHEADER, except that the width and height are 4 bytes instead of 2.


JimBobSquarePants:
@brianpopow None of our reference decoders can correctly decode the bmp so I've disabled the comparison for now.

JimBobSquarePants:
Whoop! Thanks for your help!

PR #792, opened 2018-12-16
WIP: Fix for Windows 2.0 or OS/2 1.x bitmaps only use 3 bytes per color palette

brianpopow:
### Prerequisites

### Description

The following bitmap is not decoded correctly:

[pal8os2.zip](https://github.com/SixLabors/ImageSharp/files/2683845/pal8os2.zip)

The output is:

bmpv2_output

The expected output is:

bmpv2_input

Windows 2.0 or OS/2 1.x bitmaps use 3 bytes per color palette instead of 4, as described here: 

http://www.fileformat.info/format/bmp/egff.htm

This PR provides a fix for that.

### DIFF
--- a/src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
+++ b/src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
@@ -108,7 +108,7 @@ namespace SixLabors.ImageSharp.Formats.Bmp
         {
             try
             {
-                this.ReadImageHeaders(stream, out bool inverted, out byte[] palette);
+                int bytesPerColorMapEntry = this.ReadImageHeaders(stream, out bool inverted, out byte[] palette);
 
                 var image = new Image<TPixel>(this.configuration, this.infoHeader.Width, this.infoHeader.Height, this.metaData);
 
@@ -137,6 +137,7 @@ namespace SixLabors.ImageSharp.Formats.Bmp
                                 this.infoHeader.Width,
                                 this.infoHeader.Height,
                                 this.infoHeader.BitsPerPixel,
+                                bytesPerColorMapEntry,
                                 inverted);
                         }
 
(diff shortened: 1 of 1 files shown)


brianpopow:
@JimBobSquarePants: maybe you can help me here, i think i am missing something obvious. 

I wanted to create a unit test which verifies the Decoded bitmap with reference bitmap, which was added yesterday. 

```
[Theory]
[WithFile(WinBmpv2, PixelTypes.Rgba32)]
public void BmpDecoder_CanDecodeBmpv2<TPixel>(TestImageProvider<TPixel> provider)
{
        using (Image<TPixel> image = provider.GetImage(new BmpDecoder()))
	{
		// TODO: how to compare to the reference?
	}
} 
```

I think i need to use CompareToReferenceOutput from the TestImageExtension, but i am unsure howto do it correctly.

Can you point me to an example where i can see how i should use it.

JimBobSquarePants:
@brianpopow No worries. When comparing a decoder we save the output (defaults to 32bit png) so we can view it manually then we call the `CompareToOriginal` extension method which compares our decoded result against a reference decoder (System.Drawing for bitmaps). 

https://github.com/SixLabors/ImageSharp/blob/c3a6040cdf066400cee0f377b53f6140f52a4b41/tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs#L54

That reference you added yesterday would be useful only when testing encoding once we can encode that bitmap type. You would normally just add the image to the input folder for testing the decoder.

brianpopow:
@JimBobSquarePants thank you for the clarification! 

The image i have added to the submodul was a bitmap with 40 bytes header which imagesharp can already decode. I wanted to use that as a reference. 

Saving the image as png and then comparing to System.Drawing makes much more sense, then the stunt i wanted to do :-) . So the image i have added to the submodul can be safely deleted. Sorry for this misunderstanding.

I have added now two tests. One for decoding a windows BMPv2 and to make sure decoding images with a palette of 4 bytes per entry still works, another one for that case.

brianpopow:
unfortunately that does not work either. System.Drawing can not decode this bitmap.

brianpopow:
@JimBobSquarePants: thank you! I am happy, that i could help a bit.

PR #796, opened 2018-12-22
Decoding Bitmaps with BITFIELDS masks

### LINKED ISSUE #732: Resized 32-bit BI_RGB BMP is completely black
### Prerequisites

### Description
The file "Test-original.bmp" is loaded, resized to half width and height, and then saved as "Test-resized.bmp".  The resulting image is the correct size, but every pixel is black.

"Test-original.bmp" is a 32-bit BI_RGB image where the alpha channel for each pixel is 0.  According to MSDN, the alpha channel byte is not used, so I don't _think_ it should be interpreted during the resize if that's what's causing resulting image to be black.

https://msdn.microsoft.com/en-us/library/dd183376%28VS.85%29.aspx
_"The value for blue is in the least significant 8 bits, followed by 8 bits each for green and red. **The high byte in each DWORD is not used**"_

### Steps to Reproduce
```
using SixLabors.ImageSharp;
using System.IO;

namespace TestImageSharp
{
    class Program
    {
        {
            Image<Rgba32> image = Image.Load("Test-original.bmp");
            image.Mutate(img => img.Resize(image.Width / 2, image.Height / 2));
            using (FileStream output = File.OpenWrite("Test-resized.bmp"))
            {
                image.SaveAsBmp(output);
            }
        }
    }
}
```

Test images attached:
[TestImages.zip](https://github.com/SixLabors/ImageSharp/files/2462200/TestImages.zip)

### System Configuration

- ImageSharp version:
-- SixLabors.ImageSharp.1.0.0-beta0005
-- SixLabors.Core.1.0.0-beta0006
-- SixLabors.Fonts.1.0.0-beta0007
-- SixLabors.ImageSharp.Drawing.1.0.0-beta0005
-- SixLabors.Shapes.1.0.0-beta0007
-- Windows 10 Pro 64-bit x64
-- Microsoft Visual Studio Community 2017 Preview Version 15.9.0 Preview 3.0
- .NET Framework version:
-- 4.6.1
- Additional information:
-- None

### LINKED ISSUE #735: ImageSharp "Does not support this kind of bitmap files"
### Prerequisites

### Description
Trying to open a specific bitmap file created with `Magick.NET` fails with an exception. Other files created in the same way don't.

Selecting `ImageMagick.MagickFormat.Bmp2` or `ImageMagick.MagickFormat.Bmp3` resolves the issue. The issue is with `ImageMagick.MagickFormat.Bmp` specifically, which uses, according to the docs, version 4 of the format.

The image in question _does_ open without issues in other Windows programs (Paint, Photos, Photoshop) and Dropbox.

### Steps to Reproduce
Download the [file](https://www.dropbox.com/s/i3git6ovot9mp3a/bitmap.bmp?dl=0).

Try to open it:

```csharp
using (var image = Image.Load("bitmap.bmp"))
{}
```

generates

```
Exception has occurred: CLR/System.NotSupportedException
An unhandled exception of type 'System.NotSupportedException' occurred in SixLabors.ImageSharp.dll: 'Does not support this kind of bitmap files.'
   at SixLabors.ImageSharp.Formats.Bmp.BmpDecoderCore.Decode[TPixel](Stream stream)
   at SixLabors.ImageSharp.Image.Decode[TPixel](Stream stream, Configuration config)
   at SixLabors.ImageSharp.Image.WithSeekableStream[T](Configuration config, Stream stream, Func`2 action)
   at SixLabors.ImageSharp.Image.Load[TPixel](Configuration config, Stream stream, IImageFormat& format)
   at SixLabors.ImageSharp.Image.Load[TPixel](Configuration config, String path)
```

### System Configuration

- ImageSharp version: 1.0.0-beta0005
- Environment (Operating system, version and so on): Windows 10 x64 1803
- .NET Framework version: 4.7.2 / .NET Core 2.1.4

brianpopow:
### Prerequisites

### Description

This PR adds to the BMP decoder the ability to decode Bitmaps with the Compression type BITFIELDS. For bitmap V3 (40 bytes), color masks are followed after the BitmapInfo header.

For bitmap >= V4, this information is stored inside the BitmapInfo header. To be able to get this info, i added parsing V4 information headers. 

This fixes #735 and also fixes #732


brianpopow:
I think this one is ready for review now. 

One note about the 16 Bits per pixel case: only bit masks with 5 or 6 bits are supported, because according to: http://www.fileformat.info/format/bmp/egff.htm. Windows 95 only supports the RGB555 and RGB565 pixel types. Bitmaps with other bitmasks may be theoretically possible, but i do not really think they can be found in the wild.

src/ImageSharp/Formats/Bmp/BmpConstants.cs
```diff
@@ -19,5 +19,41 @@ internal static class BmpConstants
...
+
+        /// <summary>
+        /// Valid magic bytes markers identifying a Bitmap file.
+        /// </summary>
+        internal static class TypeMarkers
+        {
+            /// <summary>
+            /// Single-image BMP file that may have been created under Windows or OS/2.
+            /// </summary>
+            public const int Bitmap = 19778;
```
JimBobSquarePants: Can these be hex so it's easier to match them in an editor.

tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
```diff
@@ -44,25 +49,100 @@ public void DecodeBmp<TPixel>(TestImageProvider<TPixel> provider)
...
+        [Theory]
+        [WithFile(Rgba321010102, PixelTypes.Rgba32)]
+        public void BmpDecoder_CanDecodeBitfields_WithUnusualBitmasks<TPixel>(TestImageProvider<TPixel> provider)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            using (Image<TPixel> image = provider.GetImage(new BmpDecoder()))
+            {
+                image.DebugSave(provider);
+
+                // Choosing large tolerance of 6.1 here, because for some reason with the MagickReferenceDecoder the alpha channel is always 191 instead of 255.
```
JimBobSquarePants: Youthink there's a bug in the reference decoder then?

src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
```diff
@@ -105,23 +106,23 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
...
 #else
-            byte[] buffer = new byte[40];
+            byte[] buffer = new byte[infoHeaderSize];
 #endif
             fileHeader.WriteTo(buffer);
 
             stream.Write(buffer, 0, BmpFileHeader.Size);
 
-            infoHeader.WriteTo(buffer);
+            infoHeader.WriteV3Header(buffer);
```
JimBobSquarePants: How does this affect alpha values. Will decoders now have to check each byte to see if we are the V3 ICO variant? 

It would be really nice if we stored the header type as an enum **Name - Length** in the metadata and gradually build up our ability to write each type starting with v3 and v4.

tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
brianpopow: I have loaded this image with GIMP, exported it as PNG and compared to my results. It produces the same results as i have, including the alpha channel, which gives me more confidence, that it is correct. I could add this to the ReferenceOutput and compare against it, if you like.

JimBobSquarePants: No, don't bother, just add to the comment saying you have done this.

src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
JimBobSquarePants: @brianpopow Ah so that was just a method name change. I thought we were changing the header itself!

We already have a method to encode 32bit bitmaps, I'm assuming we're doing that wrong then.

https://github.com/SixLabors/ImageSharp/blob/db0fc0dd7df4b23236120312419f21916b6c6d8a/src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs#L163

By V3 ICO variant I mean this. https://github.com/SixLabors/ImageSharp/issues/732#issuecomment-447151183

Yeah, we are not using the metadata, we certainly should be.

brianpopow: yes only the names changed, i wanted to make it clear, that a V3 header is written.

This V3 ICO variant was new to me. I could not find much about it, except the mozilla comment you mentioned. 

My understanding is, that a V3 header implies, no support for alpha channel. As you mentioned in the other ticket, what you called Option 2 could be a way to solve this, but it feels a bit hacky to me.

I think a better way would be to write a V4 header with BITFIELDS compression to indicate which bits are really the alpha channel.

JimBobSquarePants: I've seen the similar handling code in several decoders now. Chrome, Mozilla, ImageMagick, Gimp, all do this when decoding V3 bitmaps. It is hacky IMO also but a necessary evil. 

I'm all for writing the V4 header! 👍

brianpopow: The encoder now writes V4 header. 

I wanted to add rgb32.bmp to the list of testimages for the decoder, but this failed, because of the issue described in #732 that alpha channel is not correctly decoded. This image is a bitmap with a V3 header, but no alpha channel. In the info header it is marked as a 32 bpp image, but the alpha values are all 0, meaning no alpha channel is there. As a consequence the encoder would encode images with v4 header wrong, because it expects there an alpha channel for 32 bpp images.

As a temporary workaround i have ignored the alpha channel during encoding, when its a V3 header. This should be removed once #732 is fixed properly. I have also disabled rgb32.bmp from the decoder tests, until #732 is fixed.

JimBobSquarePants:
@brianpopow Great work! This all looks good to me so once it's built I'llmerge it. 

I took the liberty to fix #732 here also since it was blocking your tests.

brianpopow:
@JimBobSquarePants: Thank you! 

We are getting close to be able to decode all bitmaps from the "good" category from: http://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html

Only RLE4 is left now. I will do that next.

PR #801, opened 2019-01-04
#797 throw ImageFormatException when no StartOfFrame marker is found

### LINKED ISSUE #797: Image.Load throws NullReferenceException
**Image.Load** throws a **NullReferenceException** when loading the image from the [attached archive]. You can run the following code to reproduce it (the **path** variable should contain the path to the extracted file):

```csharp
Image.Load(path);
```

Found via [SharpFuzz].

[attached archive]: https://github.com/SixLabors/ImageSharp/files/2707118/NullReferenceException.zip
[SharpFuzz]: https://github.com/Metalnem/sharpfuzz

jongleur1983:
### Prerequisites

### Description
Fixes #797:

For jpg images where we can't find a valid SOF (start of frame) marker we now throw an ImageFormatException reporting that fact instead of a NullReferenceException.

Added the failing image from the issue for a test case.


src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -292,6 +292,11 @@ public void ParseStream(Stream stream, bool metadataOnly = false)
                         case JpegConstants.Markers.SOS:
                             if (!metadataOnly)
                             {
+                                if (this.Frame is null)
+                                {
+                                    throw new ImageFormatException("no readable SOF marker found.");
```
JimBobSquarePants: Can you change this to say "No readable SOFn (Start Of Frame) marker found" and move the check to `ProcessStartOfScanMarker()`

jongleur1983: sure, done.

Issue #802, opened 2019-01-06
HueProcessor doesn't behave as expected


Nacimota:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When using `HueExtensions.Hue()`, the resulting image does not look the way I would expect it to, based on my understanding of what hue is.

Given this example:

![example](https://user-images.githubusercontent.com/651955/50741780-baf65b80-124d-11e9-9165-1347654ab386.png)

I would expect a hue shift of 180 degrees to produce something like this:

![example-photoshop-180](https://user-images.githubusercontent.com/651955/50741787-e6794600-124d-11e9-93fc-2e8e470464d3.png)

However, ImageSharp produces this:

![example-180](https://user-images.githubusercontent.com/651955/50741789-f729bc00-124d-11e9-8e46-4f4d741017dd.png)

Have I misunderstood what this method is supposed to do? Basically I'm expecting it to behave like the Hue/Saturation dialog in Photoshop does, when you change the hue and `Colorize` is unchecked.

### Steps to Reproduce
1. Download and save the first image from above as `example.png`.
2. Run this code:

```csharp
using (var img = Image.Load("example.png"))
{
    img.Mutate(c => c.Hue(180));

    using (var stream = File.Create("example-180.png"))
    {
        img.SaveAsPng(stream);
    }
}
```

### System Configuration

- ImageSharp version: `1.0.0-beta0005`
- Other ImageSharp packages and versions: `none`
- Environment (Operating system, version and so on): 
  - `Windows 10 Pro 1809 x64 v10.0.17763.194`
  - `Ubuntu 18.04.1 LTS (GNU/Linux 4.15.0-43-generic x86_64)`
- .NET Framework version: `.NET Core 2.1`

Nacimota:
Okay I think I understand what's going on here. I had a dig around in the source and it pretty explicitly states that lum is preserved in the filter:

[...]

PR #806, opened 2019-01-09
Fix Color Filters

JimBobSquarePants:
### Prerequisites

### Description
This PR replaces the incorrect `Matrix4x4` filter implementation with a new `ColorMatrix` transform.

All color matrices normally have a minimum of [5x4](https://developer.android.com/reference/android/graphics/ColorMatrix) indices which are used to transform the vector. While our implementation gave accurate output in fully opaque images, images with an alpha component were incorrect. In addition, the Hue filter was using an incorrect operator in its calculation which led to output error.

This PR.
- Introduces a new `ColorMatrix` struct modelled after the `Matrix4x4` struct. This contains all the properties required to implement color transforms.
- Updates the `KnownFilterMatrices` generation code to reflect the new struct layout and fix bugs. #802
- Optimizes the filter process using bulk operations.

### DIFF
--- a/src/ImageSharp/Primitives/Matrix5x4.cs
+++ b/src/ImageSharp/Primitives/Matrix5x4.cs
@@ -1,0 +1,459 @@
+﻿// Copyright (c) Six Labors and contributors.
+// Licensed under the Apache License, Version 2.0.
+
+#pragma warning disable SA1117 // Parameters should be on same line or separate lines
+using System;
+using System.Globalization;
+using System.Runtime.InteropServices;
+
+namespace SixLabors.ImageSharp.Primitives
+{
+    /// <summary>
+    /// A structure encapsulating a 5x4 matrix.
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public struct Matrix5x4 : IEquatable<Matrix5x4>
+    {
+        /// <summary>
+        /// Value at row 1, column 1 of the matrix.
+        /// </summary>
+        public float M11;
+
+        /// <summary>
+        /// Value at row 1, column 2 of the matrix.
+        /// </summary>
+        public float M12;
+
+        /// <summary>
+        /// Value at row 1, column 3 of the matrix.
+        /// </summary>
+        public float M13;
+
+        /// <summary>
+        /// Value at row 1, column 4 of the matrix.
+        /// </summary>
+        public float M14;
+
+        /// <summary>
+        /// Value at row 2, column 1 of the matrix.
+        /// </summary>
+        public float M21;
+
+        /// <summary>
+        /// Value at row 2, column 2 of the matrix.
+        /// </summary>
+        public float M22;
+
+        /// <summary>
+        /// Value at row 2, column 3 of the matrix.
+        /// </summary>
+        public float M23;
+
+        /// <summary>
+        /// Value at row 2, column 4 of the matrix.
+        /// </summary>
+        public float M24;
+
+        /// <summary>
+        /// Value at row 3, column 1 of the matrix.
+        /// </summary>
+        public float M31;
+
+        /// <summary>
+        /// Value at row 3, column 2 of the matrix.
+        /// </summary>
+        public float M32;
+
+        /// <summary>
+        /// Value at row 3, column 3 of the matrix.
+        /// </summary>
+        public float M33;
+
+        /// <summary>
+        /// Value at row 3, column 4 of the matrix.
+        /// </summary>
+        public float M34;
+
+        /// <summary>
+        /// Value at row 4, column 1 of the matrix.
+        /// </summary>
+        public float M41;
+
+        /// <summary>
+        /// Value at row 4, column 2 of the matrix.
+        /// </summary>
+        public float M42;
+
+        /// <summary>
+        /// Value at row 4, column 3 of the matrix.
+        /// </summary>
+        public float M43;
+
+        /// <summary>
+        /// Value at row 4, column 4 of the matrix.
+        /// </summary>
+        public float M44;
+
+        /// <summary>
+        /// Value at row 5, column 1 of the matrix.
+        /// </summary>
+        public float M51;
+
+        /// <summary>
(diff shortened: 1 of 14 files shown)


JimBobSquarePants:
Starting to get really annoyed by tests randomly failing in different environments. The NET Core tests pass on my machine.

src/ImageSharp/Primitives/Matrix5x4.cs
```diff
@@ -0,0 +1,459 @@
...
+using System.Globalization;
+using System.Runtime.InteropServices;
+
+namespace SixLabors.ImageSharp.Primitives
+{
+    /// <summary>
+    /// A structure encapsulating a 5x4 matrix.
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public struct Matrix5x4 : IEquatable<Matrix5x4>
```
antonfirsov: Shouldn't we move this to SixLabors.Core?

src/ImageSharp/Primitives/Matrix5x4.cs
```diff
@@ -0,0 +1,459 @@
...
+            this.M51 = m51;
+            this.M52 = m52;
+            this.M53 = m53;
+            this.M54 = m54;
+        }
+
+        /// <summary>
+        /// Gets the multiplicative identity matrix.
+        /// </summary>
+        public static Matrix5x4 Identity { get; } =
```
antonfirsov: The term identity matrix [is only valid for square matrices](https://en.wikipedia.org/wiki/Identity_matrix)! We should drop this.

tests/ImageSharp.Tests/Primitives/Matrix5x4Tests.cs
```diff
@@ -0,0 +1,270 @@
...
+            Matrix5x4 m = this.CreateAllOnes() + this.CreateAllTwos();
+            Assert.Equal(this.CreateAllThrees(), m);
+        }
+
+        [Fact]
+        public void Matrix5x4HashCode()
+        {
+#if NETCOREAPP2_1
+            Matrix5x4 m = KnownFilterMatrices.CreateBrightnessFilter(.5F);
+            HashCode hash = default;
```
antonfirsov: I would not add such a strict test case for `GetHashCode()`. The only thing we should ensure is that for equal objects hash code is also equal. (The opposite is not true!)

src/ImageSharp/Primitives/Matrix5x4.cs
```diff
@@ -0,0 +1,459 @@
...
+            return m;
+        }
+
+        /// <summary>
+        /// Multiplies a matrix by another matrix.
+        /// </summary>
+        /// <param name="value1">The first source matrix.</param>
+        /// <param name="value2">The second source matrix.</param>
+        /// <returns>The result of the multiplication.</returns>
+        public static Matrix5x4 operator *(Matrix5x4 value1, Matrix5x4 value2)
```
antonfirsov: The result of this multiplication should be [either a `5x5` or a `4x4` matrix](https://en.wikipedia.org/wiki/Matrix_multiplication), depending on how we define the operation.

As for non-square fixed size matrices the multiplication has usually no practical use, I suggest to drop this operation.

antonfirsov:
Re `Matrix5x4`:
- Multiplication and identity actually makes sense if `Matrix5x4` is an optimized form for a `Matrix5x5` having an invariant `(0, 0, 0, 0, 1)˙ row/column the same way as `Matrix3x2` is really just an optimized `Matrix3x3`.
- We may consider renaming it to `ColorMatrix` and keep it inside the bounds of ImageSharp

tests/ImageSharp.Tests/Primitives/Matrix5x4Tests.cs
JimBobSquarePants: I just copied what they do with Numerics tests. I think it still adds value as it ensures we touch every field when calculating the hashcode

src/ImageSharp/Primitives/Matrix5x4.cs
JimBobSquarePants: See below

src/ImageSharp/Primitives/Matrix5x4.cs
JimBobSquarePants: I don't think it has any practical application outside of ImageSharp.

src/ImageSharp/Primitives/Matrix5x4.cs
JimBobSquarePants: As you touched upon below this does make sense if our struct is an optimized form of a `Matrix5x5` in the manner that `Matrix3x2` is an optimized `Matrix3x3` which it is. 

Most importantly it works!
https://github.com/SixLabors/ImageSharp/blob/f10a96c2d9b288019e7a4b3bcca04cd7ce87179f/tests/ImageSharp.Tests/Processing/Processors/Filters/FilterTest.cs#L45

JimBobSquarePants:
> We should probably add tests based on images in #802. For my naive eyes they look like important special cases.

I did consider that but it only really adds regression protection which we already have covered with our sample files. I could replace the samples though.

> We may consider renaming it to ColorMatrix and keep it inside the bounds of ImageSharp

This did cross my mind also, I just like the conformity of naming it currently has with Numerics. It's a coin toss.

tests/ImageSharp.Tests/Primitives/Matrix5x4Tests.cs
antonfirsov: Ok, let's keep it as is for now!

antonfirsov:
> I don't think it has any practical application outside of ImageSharp.

If we keep the class within the bounds of ImageSharp, and only use it only for colorspace transformations, it makes the most sense to name it as `ColorMatrix`. Fight me! 😄

(The rest of my findings are obsolete.)

JimBobSquarePants:
Haha... I’ll rename it tomorrow then.

PR #813, opened 2019-01-20
Adds support for OS/2 version 2 bitmaps

brianpopow:
### Prerequisites

### Description

This PR adds support for decoding OS/2 version 2 bitmaps (at least partially). OS/2 v2 introduces additional compression type of RLE24 and 1D Huffman, which is not supported.

I had some trouble finding actual example images for this. I have only one with no compression. RLE4 and RLE8 should work, too, but i could not find those.


JimBobSquarePants:
@brianpopow Those new exceptions are going to start to hurt performance and coverage. Would it be possible for you to refactor them out into a static BmpThrowHelper class like the ThrowHelper class in the jpeg decoder?

JimBobSquarePants:
Merged. Thanks again, couldn't do this without you!

PR #840, opened 2019-02-20
Draft PR for the bokeh blur effect (updated)

Sergio0694:
### Prerequisites

### Description

Follow up from #837 - with some minor updates:
• I've added the `Complex64` type, which uses the `float` for the real and imaginary components, to save memory and processing time.
• Following advice from @antonfirsov, I've added a caching system in the `BokehBlurProcessor` class, so that the kernel parameters and values can be cached and reused across multiple instances using the same initialization parameters.

I need some guidance on how to proceed with the actual convolution pass (I mean, on how to properly structure it following your repo guidelines and coding style), and on eventual optimization steps I need to be aware of while writing this down (especially with the various pixel operations). Unlike with the other blur effects, both convolution passes here produce a `Complex64` result, so some changes are needed in both the custom convolution 2 pass processor we'll need, as well as the actual 1D convolution code to be added to the helper class.

### DIFF
--- a/src/ImageSharp/Primitives/Complex64.cs
+++ b/src/ImageSharp/Primitives/Complex64.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Six Labors and contributors.
 // Licensed under the Apache License, Version 2.0.
 
+using System.Runtime.CompilerServices;
+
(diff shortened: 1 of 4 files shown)


Sergio0694:
Continues in #842 with a work in progress implementation.

PR #842, opened 2019-02-21
Bokeh blur implementation

Sergio0694:
### Prerequisites

### Description
A bokeh blur processor that approximates a 2D bokeh blur convolution effect by combining a series of 2D linearly separable convolutions (2-pass 1D convolutions) in complex space. This reduces the cost from `O(n^2)` to `O(n)` for each computed pixel.

#### Example image:
cyberpunk
cyberpunk_bokeh


Sergio0694:
@antonfirsov It looks like the latest commit failed the AppVeyor build with an `OutOfMemoryException` in `PaletteColorType_WuQuantizer<Rgba32>`, which is completely unrelated to this PR.
This only happened for the 32 bit .NET 4.62 target, the others worked just fine.
Is this just a random error thrown by the CI by mistake? 🤔

antonfirsov:
@Sergio0694 unfortunately, certain sporadic errors are common, this has nothing to do with your commit. Just triggered the failing build again.

Sergio0694:
This PR is almost ready for review, the bokeh blur effect is working and I've optimized the code as much as I could (this doesn't mean it couldn't be improved even further, of course).
There's just a small bug that remains to be fixed, and it only happens when applying the blur to a specific region (not on the whole image), and when the `gamma` parameter is greater than 1.
Here's the result, from one of the test images: 

image

As you can see, there's a slight light margin at the top of the blurred rectangle. A few notes:

- I'm sure this is somehow related to the gamma exposure, as leaving the `gamma` parameter to `1` hides the issue.
- This doesn't seem to happen when applying the blur on the whole image, even with the same settings (including the gamma exposure). In this case, the top of the image is perfectly fine instead.
- This only seems to happen on the top edge of the target rectangle, the other edges are fine.

I spent some time to investigate this but I couldn't figure out what was causing it. I might very well be missing something obvious here 🤔

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -86,6 +87,7 @@ internal static class DenseMatrixUtils
             Span<ComplexVector4> targetRow,
             int row,
             int column,
+            int minRow,
```
dlemstra: Maybe use `offsetRow` instead and reorder the arguments to gave a more clear signature?

Sergio0694: Isn't `minRow` more coherent though?
There's already a `maxRow` parameter that does exactly the same, just on the opposite side of the target rectangle, and that is used in the `Clamp` method.
This way it is called with `Clamp(minRow, maxRow)` which makes it extremely clear.
If I used `offsetRow` it'll be `Clamp(offsetRow, maxRow)`, which to me is much less intuitive, as each parameter has a different name there for no apparent reason.

Sergio0694:
I've fixed that last remaining issue, and this PR is now ready for review 👍
I don't see the button to mark the PR as ready for review though, so I don't know how to get out of the draft status. If any of you admins could do that for me, that'd be much appreciated 😄

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
dlemstra: Using `minRow` and `offsetColumn` just looked a bit confusing. Could you not use `min` or `offset` for both?

Sergio0694: I see what you mean now, I hadn't noticed the name mismatch in the column paarameters since those were the same from the already existing convolution methods, I'll go ahead and make them all use `min`/`max` 👍

Also, can you mark this PR as ready for review? I don't see the option to change the PR state on my end, and waas wondering if you mods could do that for me.

Sergio0694:
@antonfirsov Looks like the CI tests on 32 bit .NET 4.7.2 failed with the last commit, but I guess that could just be the VM acting up? Could you trigger the build again to see if that was just a temporary quirk?

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -99,7 +101,7 @@ internal static class DenseMatrixUtils
 
             for (int i = 0; i < kernelLength; i++)
             {
-                int offsetY = (row + i - radiusY).Clamp(0, maxRow);
+                int offsetY = (row + i - radiusY).Clamp(minRow, maxRow);
```
JimBobSquarePants: Ah! Bad James!

Sergio0694: I actually realized this needed to be fixed in the second complex 1D convolution pass as well, done in https://github.com/SixLabors/ImageSharp/pull/842/commits/15044d740319ca0ba702191d5efdb8367299d159.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
```diff
@@ -0,0 +1,534 @@
...
+
+        /// <summary>
+        /// Gets the gamma highlight factor to use when applying the effect
+        /// </summary>
+        public float Gamma { get; }
+
+        /// <summary>
+        /// Gets the kernel scales to adjust the component values in each kernel
+        /// </summary>
+        private static IReadOnlyList<float> KernelScales { get; } = new[] { 1.4f, 1.2f, 1.2f, 1.2f, 1.2f, 1.2f };
```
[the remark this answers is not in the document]
Sergio0694: I'm confused, that's... Already readonly? 😗
It's a property, not a field.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
```diff
@@ -0,0 +1,534 @@
...
+
+        /// <summary>
+        /// Gets the available bokeh blur kernel parameters
+        /// </summary>
+        private static IReadOnlyList<Vector4[]> KernelComponents { get; } = new[]
```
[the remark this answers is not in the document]
Sergio0694: Same as before, this is already readonly. I can make both of these fields instead of properties if you prefer though. The access time wouldn't be a problem either way, as both of these are only accessed once in the constructor, if the kernels aren't present in the cache.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
```diff
@@ -0,0 +1,534 @@
...
+        /// <param name="gamma">
+        /// The gamma highlight factor to use to further process the image.
+        /// </param>
+        public BokehBlurProcessor(int radius = 32, int components = 2, float gamma = 3)
+        {
+            this.Radius = radius;
+            this.kernelSize = (radius * 2) + 1;
+            this.componentsCount = components;
+
+            SixLabors.Guard.MustBeGreaterThanOrEqualTo(gamma, 1, nameof(gamma));
```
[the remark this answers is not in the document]
Sergio0694: Fixed in https://github.com/SixLabors/ImageSharp/pull/842/commits/67556c85ffdcb39bfc42a29495e416300f277681.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
```diff
@@ -0,0 +1,534 @@
...
+            this.Gamma = gamma;
+
+            // Reuse the initialized values from the cache, if possible
+            if (Cache.TryGetValue((radius, components), out (Vector4[], float, Complex64[][]) info))
```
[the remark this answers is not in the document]
Sergio0694: I'm not sure I understand the issue with the second point. I mean, a `TryGetValue` call is always `O(1)` anyway, and I don't get why would anyone try to exhaust the available RAM by just spamming the bokeh extension with random parameters.
Instead, say someone is really using some parameters that are outside the range we define, wouldn't they always need to recalculate the kernels every time then?
(honest question, I might very well be missing something obvious here)

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
```diff
@@ -0,0 +1,534 @@
...
+        /// <param name="radius">
+        /// The 'radius' value representing the size of the area to sample.
+        /// </param>
+        /// <param name="components">
+        /// The number of components to use to approximate the original 2D bokeh blur convolution kernel.
+        /// </param>
+        /// <param name="gamma">
+        /// The gamma highlight factor to use to further process the image.
+        /// </param>
+        public BokehBlurProcessor(int radius = 32, int components = 2, float gamma = 3)
```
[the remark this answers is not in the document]
Sergio0694: I've used default parameters there because especially for the `components` and `gamma` parameters, I wouldn't expect most users to be able to pick the right values on their own, as they require at least some knowledge of how the effect works to be undestoood at all.
`2` components and a gamma of `3` provide a good compromise for quality and speed of execution, that should work fine in most cases.
What if I only made those two optional, and left the `radius` parameter with no default value?

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
```diff
@@ -0,0 +1,534 @@
...
+
+        /// <summary>
+        /// The scaling factor for kernel values
+        /// </summary>
+        private readonly float kernelsScale;
+
+        /// <summary>
+        /// The mapping of initialized complex kernels and parameters, to speed up the initialization of new <see cref="BokehBlurProcessor{TPixel}"/> instances
+        /// </summary>
+        private static readonly Dictionary<(int, int), (Vector4[], float, Complex64[][])> Cache = new Dictionary<(int, int), (Vector4[], float, Complex64[][])>();
```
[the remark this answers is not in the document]
Sergio0694: Fixed in https://github.com/SixLabors/ImageSharp/pull/842/commits/f8d7964766b8f63e8b30fab57a81263e83e6f92a.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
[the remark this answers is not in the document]
antonfirsov: Ooops, its a property yeah. Nevermind.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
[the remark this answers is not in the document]
antonfirsov: If you expose your parameters through a web API, it might be an attack surface.

Actually, pretty unlikely for a such a fancy processor, so feel free to ignore me, but for more common stuff, like Resize, I would definitely consider this.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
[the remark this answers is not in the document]
Sergio0694: Ahahahah perfect 😄

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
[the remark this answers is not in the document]
Sergio0694: Mmmh, if you really like the idea of having a limited number of items in the cache, what about capping the cache size at a given size (say, 100, or some other arbitrary value), and using a list that stores the insertion time for each key? This way when the cap is reached, we could throw the oldest key away to make room for the new one. This would tend to keep the most used keys always available.
This simple idea would cost `O(n)` in theory, but if the size is capped that'd technically still be `O(1)`, just with a slightly higher constant.
We could wrap this into a separate type to keep things easy and possibly reusable.
Something like a `PriorityDictionary<TKey,TValue>`, for instance.
Otherwise I can just leave this part as it is in the bokeh processor.
What do you think?

antonfirsov:
@Sergio0694 strange, github UI tells me that the branch is up-to-date, while there are compilation errors on CI because of API changes.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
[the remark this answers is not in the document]
antonfirsov: Let's keep it simple for now, and worry about this later!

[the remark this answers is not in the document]
Sergio0694: Got it, perfect 👍

Sergio0694:
@antonfirsov I've tried to build locally and I see two errors where the gamma steps in the bokeh blur processor are using the `PixelOperations<TPixel>.Instance.FromVector4` API, which is now gone.
Not sure why that one was missed from the merge, as I assumed you guys have already properly adjusted all the other occurrences everywhere else.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
```diff
@@ -0,0 +1,543 @@
...
+                (rows, vectorBuffer) =>
+                    {
+                        Span<Vector4> vectorSpan = vectorBuffer.Span;
+                        int length = vectorSpan.Length;
+
+                        for (int y = rows.Min; y < rows.Max; y++)
+                        {
+                            Span<TPixel> targetRowSpan = targetPixels.GetRowSpan(y).Slice(startX);
+                            PixelOperations<TPixel>.Instance.ToVector4(configuration, targetRowSpan.Slice(0, length), vectorSpan);
+                            Vector4Utils.Premultiply(vectorSpan);
```
antonfirsov: Instead of explicit premultiplication, you should pass `PixelConversionModifiers.Premultiply` to the methods `.ToVector4(...)` and `.FromVector4(...)`.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
```diff
@@ -0,0 +1,543 @@
...
+
+                            for (int x = 0; x < width; x++)
+                            {
+                                ref Vector4 v = ref Unsafe.Add(ref baseRef, x);
+                                v.X = (float)Math.Pow(v.X, gamma);
+                                v.Y = (float)Math.Pow(v.Y, gamma);
+                                v.Z = (float)Math.Pow(v.Z, gamma);
+                            }
+
+                            PixelOperations<TPixel>.Instance.FromVector4Destructive(configuration, vectorSpan.Slice(0, length), targetRowSpan);
```
antonfirsov: Use `PixelConversionModifiers` here as well!

Sergio0694: Fixed in https://github.com/SixLabors/ImageSharp/pull/842/commits/fe6c5026ac62b2971a74845263f179bd4ae06efe.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
Sergio0694: Fixed in https://github.com/SixLabors/ImageSharp/pull/842/commits/fe6c5026ac62b2971a74845263f179bd4ae06efe.

Sergio0694: Actually, I think I'll need to leave out the modifier in this `FromVector4` call, as the processor is expecting premultiplied data to work on.

Sergio0694:
@antonfirsov The tests are still failed, I'm a bit confused as to why 🤔

To recap, with the previous APIs:
- In `ApplyGammaExposure`:
  - `PixelOperations<TPixel>.Instance.ToVector4`
  - `Vector4Utils.Premultiply`
  - `PixelOperations<TPixel>.Instance.FromVector4` (with no unpremultiplication here)
- In `ApplyInverseGammaExposure`:
  - `Vector4Utils.UnPremultiply`
  - `PixelOperations<TPixel>.Instance.FromVector4`

With the updated APIs:
- In `ApplyGammaExposure`:
  - `PixelOperations<TPixel>.Instance.ToVector4Destructive(PixelConversionModifiers.Premultiply)`
  - `PixelOperations<TPixel>.Instance.FromVector4Destructive` (with no unpremultiplication here)
- In `ApplyInverseGammaExposure`:
  - `PixelOperations<TPixel>.Instance.FromVector4(PixelConversionModifiers.Premultiply)`

Though since the tests are falling I assume the changes I made don't exactly reproduce the behavior that was there with the earier version of those APIs.
Any ideas? 😅

antonfirsov:
@Sergio0694 many tests were failing on appveyor because the submodule change also has to be commited + pushed. I did it now.

Some of the tests do fail locally for me:

**Example:** `BokehBlurFilterProcessor_TestPattern200x100_R16_C1_G3.png`
Expected (reference) output:
image

Actual output:
image

With my late evening brain, I just can't follow this premultiplication/unpremultiplication thing. (Why do we have to go with premultiplication sometimes while omiting premultiplication.) So I let you resolve this the way you suggest, and merge the PR as soon as the failing tests are fixed.

Sergio0694:
@antonfirsov Mmmmh I have to say that reference output image doesn't 100% convince me either 🤔
You know what, I'll try to rollback to before the merge from `master` with the new `ToVector4` and `FromVector4` APIs, and see if everything was indeed fine before then (it should've been).
Then I'll make some comparisons again with the Python script as well just to double check.

I'm 99% sure though that the bokeh blur waas working fine before the refactoring to those `To/FromVector4` APIs, ooof 😓

antonfirsov:
This is how the original 200x100 test pattern looks like:
image

It has transparency. Maybe you can check it against your python scripts.

antonfirsov:
Sorry, the first image I uploaded was wrong, now edited!

antonfirsov:
Also the second one 😆 Now it's **probably** OK. It will better to everyone if I just go to bed now 🛌

Sergio0694:
@antonfirsov Ahahah thanks for your help, and yeah I'll definitely check with the Python scripts too.
Have a good night man! 😊

src/ImageSharp/Primitives/ComplexVector4.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System.Numerics;
+using System.Runtime.CompilerServices;
+
+namespace SixLabors.ImageSharp.Primitives
+{
+    /// <summary>
+    /// A vector with 4 values of type <see cref="Complex64"/>.
+    /// </summary>
+    internal struct ComplexVector4
```
JimBobSquarePants: Implement `IEquatable<T>` etc  for best practice

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -247,6 +248,91 @@ internal static class DenseMatrixUtils
...
+        /// <param name="kernel">The 1D convolution kernel.</param>
+        /// <param name="sourceValues">The source frame.</param>
+        /// <param name="targetRow">The target row.</param>
+        /// <param name="row">The current row.</param>
+        /// <param name="column">The current column.</param>
+        /// <param name="minRow">The minimum working area row.</param>
+        /// <param name="maxRow">The maximum working area row.</param>
+        /// <param name="minColumn">The minimum working area column.</param>
+        /// <param name="maxColumn">The maximum working area column.</param>
+        public static void Convolve1D(
```
JimBobSquarePants: I'd like to remove these two methods from here as they do not use the `DenseMatrix<T>` struct. Also consider renaming to `Convolve4` to indicate that all four components are convolved.

src/ImageSharp/Primitives/BokehBlurParameters.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Six Labors and contributors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Primitives
+{
+    /// <summary>
+    /// A <see langword="struct"/> that contains parameters to apply a bokeh blur filter
+    /// </summary>
+    internal readonly struct BokehBlurParameters
```
JimBobSquarePants: Implement `IEquatable<T>` etc  for best practice

JimBobSquarePants: I don't think this an any other BokehBlur specific primitives should live under this namespace. `ComplexVector` and `Complex64` are an exception.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
```diff
@@ -0,0 +1,531 @@
...
+using SixLabors.Memory;
+using SixLabors.Primitives;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Convolution
+{
+    /// <summary>
+    /// Applies bokeh blur processing to the image.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    internal class BokehBlurProcessor<TPixel> : ImageProcessor<TPixel>
```
JimBobSquarePants: Let's put a link here to the original source you based this from.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
```diff
@@ -0,0 +1,531 @@
...
+                        for (int y = rows.Min; y < rows.Max; y++)
+                        {
+                            Span<Vector4> targetRowSpan = sourceValues.GetRowSpan(y).Slice(startX);
+                            Span<TPixel> targetPixelSpan = targetPixels.GetRowSpan(y).Slice(startX);
+                            ref Vector4 baseRef = ref MemoryMarshal.GetReference(targetRowSpan);
+
+                            for (int x = 0; x < width; x++)
+                            {
+                                ref Vector4 v = ref Unsafe.Add(ref baseRef, x);
+                                v.X = (float)Math.Pow(v.X.Clamp(0, float.PositiveInfinity), expGamma);
```
JimBobSquarePants: `Vector4.Clamp` with Hi/Low defined outside the loop.
Use `MathF.Pow`

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
```diff
@@ -0,0 +1,531 @@
...
+                        for (int y = rows.Min; y < rows.Max; y++)
+                        {
+                            Span<TPixel> targetRowSpan = targetPixels.GetRowSpan(y).Slice(startX);
+                            PixelOperations<TPixel>.Instance.ToVector4(configuration, targetRowSpan.Slice(0, length), vectorSpan, PixelConversionModifiers.Premultiply);
+                            ref Vector4 baseRef = ref MemoryMarshal.GetReference(vectorSpan);
+
+                            for (int x = 0; x < width; x++)
+                            {
+                                ref Vector4 v = ref Unsafe.Add(ref baseRef, x);
+                                v.X = (float)Math.Pow(v.X, gamma);
```
JimBobSquarePants: `MathF.Pow`

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
```diff
@@ -247,6 +248,91 @@ internal static class DenseMatrixUtils
...
+            int radiusX = kernelLength >> 1;
+            int sourceOffsetColumnBase = column + minColumn;
+
+            int offsetY = row.Clamp(minRow, maxRow);
+            ref ComplexVector4 sourceRef = ref MemoryMarshal.GetReference(sourceValues.GetRowSpan(offsetY));
+            ref Complex64 baseRef = ref MemoryMarshal.GetReference(kernel);
+
+            for (int x = 0; x < kernelLength; x++)
+            {
+                int offsetX = (sourceOffsetColumnBase + x - radiusX).Clamp(minColumn, maxColumn);
```
JimBobSquarePants: How does this convolution work with semi-opaque pixels? We use alpha premultiplication for other blur algorithms to ensure color output is correct.

src/ImageSharp/Primitives/ComplexVector4.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/842/commits/9491262b4b4214c273044910107f24b99ee65cd8.

src/ImageSharp/Primitives/BokehBlurParameters.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/842/commits/9edf46b511c363839c3241ff38bb401c505ae8fc and https://github.com/SixLabors/ImageSharp/pull/842/commits/db581635595e6db4dff75fbe726bfc67f51b8188.
I've moved the two types into a `Parameters` folder in the `Convolution` processors subfolders. We can chage this later on if you want.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/842/commits/24cc75ab7037254f7ec5f15d5de212e5b67fb8a0.

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/842/commits/0a77f69ebf043b7d2811360718c280a94b7746f8.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/842/commits/04984cd65006b8794fcef38564ef1414654eecb5.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/842/commits/04984cd65006b8794fcef38564ef1414654eecb5 and https://github.com/SixLabors/ImageSharp/pull/842/commits/a70518bcdad126be73aa97b5a1398b4af4c52bce.

src/ImageSharp/Processing/Processors/Convolution/Parameters/BokehBlurExecutionMode.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Six Labors and contributors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Processing.Processors.Convolution.Parameters
```
JimBobSquarePants: Move this to `SixLabors.ImageSharp.Processing` please so it's easier for devs to consume. We lump all options there.

JimBobSquarePants:
We've got a reference comparison failing here against `BokehBlurFilterProcessor_WorksWithAllPixelTypes<Gray8>(provider: TestPattern200x200[Gray8])`

src/ImageSharp/Processing/Processors/Convolution/Parameters/BokehBlurExecutionMode.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/842/commits/9aa75ccf4eae64a8be5be0f3a1cdb6d625a46da8.

src/ImageSharp/Common/Helpers/DenseMatrixUtils.cs
Sergio0694: So, just to follow up on our conversation on gitter.
In case you want to actually check the code, the premultiplication pass is done right before doing the initial gamma exposure correction, here: https://github.com/Sergio0694/ImageSharp/blob/8e9986e058b07d44a5ddf9990931ad335ced35a9/src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor%7BTPixel%7D.cs#L470.

Then the various steps to apply the effect are performed, and the pixels are finally unpremultiplied after the inverse gamma exposure step, here: https://github.com/Sergio0694/ImageSharp/blob/8e9986e058b07d44a5ddf9990931ad335ced35a9/src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor%7BTPixel%7D.cs#L533.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor.cs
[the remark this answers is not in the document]
JimBobSquarePants: I changed this during the rebase

JimBobSquarePants:
And merged! Thanks so much for adding this feature! :+1:

Sergio0694:
Right on, awesome! Thank you! 😄🍻

Issue #854, opened 2019-03-21
Modify static readonly byte arrays to take advantage of compiler functionality.


tannergooding:
The C# compiler added some optimizations around static readonly arrays that will emit them as compile time constants embedded in the PE file.

For most primitive types something like `static readonly uint[] x = new uint[] { 1, 2, 3 };` will be compiled down to the following code in the static ctor for the containing type:
```cs
uint[] array = new uint[3];
RuntimeHelpers.InitializeArray(array, &global::<PrivateImplementationDetails>.7037807198C22A7D2B0807371D763779A84FDFCF);
x = array;
```

The `RuntimeHelpers.InitializeArray` function effectively performs a memcpy of the constant data into the array. This function also handles the endian conversion on non little-endian machines.

For `byte` and `sbyte` in particular, there is no endian conversion required and an additional optimization is available. Namely, you can change the signature from `static readonly byte[] x = new byte[] { 1, 2, 3 };` to `static ReadOnlySpan<byte> x => new byte[] { 1, 2, 3 };`.  This property pattern will get optimized down to:
```
static ReadOnlySpan<byte> x => new ReadOnlySpan<byte>(&global::<PrivateImplementationDetails>.7037807198C22A7D2B0807371D763779A84FDFCF, 3);
```

In this case, the `ReadOnlySpan` is just initialized with a direct pointer to the compile time constant data.

ImageSharp should review their various static readonly arrays and determine when it would be applicable to take advantage of this optimization. In cases where it is not returning a `byte[]` but where all elements can fit within a `byte` or `sbyte`, it should additionally be considered to change the backing storage to be `byte` or `sbyte`.

More details can be found: https://github.com/dotnet/roslyn/pull/24621

JimBobSquarePants:
I think we've pretty much covered most cases in the code base now. Anything further can be dealt with on a case by case basis.

PR #855, opened 2019-03-21
Updating some readonly static data in JpegEncoderCore to take advantage of compiler functionality.

tannergooding:
### Prerequisites

### Description

This is a partial fix of https://github.com/SixLabors/ImageSharp/issues/854 and shows how to take advantage of the underlying compiler functionality.

### DIFF
--- a/src/ImageSharp/Common/Extensions/StreamExtensions.cs
+++ b/src/ImageSharp/Common/Extensions/StreamExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0.
 
 using System;
+using System.Buffers;
 using System.IO;
 
 using SixLabors.ImageSharp.Memory;
@@ -82,5 +83,25 @@ namespace SixLabors.ImageSharp
         {
             stream.Write(buffer.Array, 0, buffer.Length());
         }
+
+#if NET472 || NETSTANDARD1_3 || NETSTANDARD2_0
+        // This is a port of the CoreFX implementation and is MIT Licensed: https://github.com/dotnet/coreclr/blob/c4dca1072d15bdda64c754ad1ea474b1580fa554/src/System.Private.CoreLib/shared/System/IO/Stream.cs#L770
+        public static void Write(this Stream stream, ReadOnlySpan<byte> buffer)
+        {
+            // This uses ArrayPool<byte>.Shared, rather than taking a MemoryAllocator,
+            // in order to match the signature of the framework method that exists in
+            // .NET Core.
+            byte[] sharedBuffer = ArrayPool<byte>.Shared.Rent(buffer.Length);
+            try
+            {
+                buffer.CopyTo(sharedBuffer);
+                stream.Write(sharedBuffer, 0, buffer.Length);
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(sharedBuffer);
+            }
+        }
+#endif
     }
 }
--- a/src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
@@ -28,85 +28,6 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
         private const int QuantizationTableCount = 2;
 
         /// <summary>
-        /// Counts the number of bits needed to hold an integer.
-        /// </summary>
-        private static readonly uint[] BitCountLut =
-            {
-                0, 1, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4, 4, 4, 4, 4, 5, 5, 5, 5, 5, 5,
-                5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,
-                6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,
-                7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7,
-                7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7,
-                7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7,
-                7, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
-                8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
-                8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
-                8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
-                8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
-                8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
-                8, 8, 8,
-            };
-
-        /// <summary>
-        /// The SOS (Start Of Scan) marker "\xff\xda" followed by 12 bytes:
-        /// - the marker length "\x00\x0c",
-        /// - the number of components "\x03",
-        /// - component 1 uses DC table 0 and AC table 0 "\x01\x00",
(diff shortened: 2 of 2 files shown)


src/ImageSharp/Common/Extensions/StreamExtensions.cs
```diff
@@ -82,5 +83,22 @@ public static void Write(this Stream stream, IManagedByteBuffer buffer)
         {
             stream.Write(buffer.Array, 0, buffer.Length());
         }
+
+#if NET472 || NETSTANDARD1_3 || NETSTANDARD2_0
+        // This is a port of the CoreFX implementation: https://github.com/dotnet/coreclr/blob/c4dca1072d15bdda64c754ad1ea474b1580fa554/src/System.Private.CoreLib/shared/System/IO/Stream.cs#L770
```
tannergooding: I didn't see a 3rd party license file in the repo root. Do you normally just list that inline (given this is a copy from CoreFX, it is technically MIT).

JimBobSquarePants: Inline is fine for now but we should certainly add a file for future work.

tannergooding: Updated the comment to explicitly call out the MIT license and logged https://github.com/SixLabors/ImageSharp/issues/856 to track adding a 3rd party notices file.

tannergooding:
```
BenchmarkDotNet=v0.11.3, OS=Windows 10.0.17763.379 (1809/October2018Update/Redstone5)
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100-preview4-010901
  [Host]     : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT
```

Before:

Method |   TestImage |     Mean |     Error |    StdDev |   Median | Ratio | RatioSD |
---------------------- |------------ |---------:|----------:|----------:|---------:|------:|--------:|
'System.Drawing Jpeg' | Bmp/Car.bmp | 4.817 ms | 0.1838 ms | 0.5419 ms | 4.985 ms |  1.00 |    0.00 |
'ImageSharp Jpeg' | Bmp/Car.bmp | 7.079 ms | 0.4893 ms | 1.4428 ms | 6.105 ms |  1.51 |    0.44 |

After:

Method |   TestImage |     Mean |     Error |    StdDev |   Median | Ratio | RatioSD |
---------------------- |------------ |---------:|----------:|----------:|---------:|------:|--------:|
'System.Drawing Jpeg' | Bmp/Car.bmp | 4.857 ms | 0.1933 ms | 0.5698 ms | 5.036 ms |  1.00 |    0.00 |
'ImageSharp Jpeg' | Bmp/Car.bmp | 5.716 ms | 0.0762 ms | 0.0595 ms | 5.716 ms |  1.29 |    0.15 |

src/ImageSharp/Common/Extensions/StreamExtensions.cs
```diff
@@ -82,5 +83,22 @@ public static void Write(this Stream stream, IManagedByteBuffer buffer)
         {
             stream.Write(buffer.Array, 0, buffer.Length());
         }
+
+#if NET472 || NETSTANDARD1_3 || NETSTANDARD2_0
+        // This is a port of the CoreFX implementation and is MIT Licensed: https://github.com/dotnet/coreclr/blob/c4dca1072d15bdda64c754ad1ea474b1580fa554/src/System.Private.CoreLib/shared/System/IO/Stream.cs#L770
+        public static void Write(this Stream stream, ReadOnlySpan<byte> buffer)
+        {
+            byte[] sharedBuffer = ArrayPool<byte>.Shared.Rent(buffer.Length);
```
antonfirsov: We usually prefer using our own memory management primitives like [`MemoryAllocator.AllocateManagedByteBuffer()`](https://github.com/SixLabors/Core/blob/master/src/SixLabors.Core/Memory/MemoryAllocator.cs#L29) instead of `ArrayPool.Shared`.

src/ImageSharp/Common/Extensions/StreamExtensions.cs
```diff
@@ -82,5 +83,22 @@ public static void Write(this Stream stream, IManagedByteBuffer buffer)
...
+            try
+            {
+                buffer.CopyTo(sharedBuffer);
```
antonfirsov: To me it looks like we are loosing all the benefits of the copy-free initialization at this line...

antonfirsov: But the benchmarks show different results ... I wonder why?

src/ImageSharp/Common/Extensions/StreamExtensions.cs
tannergooding: This is matching a signature and implementation available in .NET Core. `ArrayPool<byte>.Shared` should generally be allocation-less for something like this (not necessarily for other `T`) since most of the framework is fairly dependent on it.

I'll run benchmarks on net472 as well though, in order to see what the overhead here is, if any, since it isn't applicable to netcoreapp2.1 (where the above benchmark was run).

src/ImageSharp/Common/Extensions/StreamExtensions.cs
tannergooding: In the case of netcoreapp, the method is actually virtual and various stream types can override it to avoid the copy. I'll double check the numbers on full framework to see if it hurts anything (and if so, will look at what can be done).

tannergooding:
```
BenchmarkDotNet=v0.11.3, OS=Windows 10.0.17763.379 (1809/October2018Update/Redstone5)
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100-preview4-010901
  [Host] : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT
  Clr    : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3362.0
```

Before:

Method |   TestImage |     Mean |     Error |    StdDev |   Median | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
---------------------- |------------ |---------:|----------:|----------:|---------:|------:|--------:|------------:|------------:|------------:|--------------------:|
'System.Drawing Jpeg' | Bmp/Car.bmp | 4.856 ms | 0.1641 ms | 0.4838 ms | 5.064 ms |  1.00 |    0.00 |    210.9375 |           - |           - |           873.48 KB |
'ImageSharp Jpeg' | Bmp/Car.bmp | 6.041 ms | 0.0559 ms | 0.0467 ms | 6.022 ms |  1.24 |    0.15 |     23.4375 |           - |           - |           129.19 KB |

After:
 
Method |   TestImage |     Mean |     Error |    StdDev |   Median | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
---------------------- |------------ |---------:|----------:|----------:|---------:|------:|--------:|------------:|------------:|------------:|--------------------:|
'System.Drawing Jpeg' | Bmp/Car.bmp | 4.909 ms | 0.1418 ms | 0.4180 ms | 5.046 ms |  1.00 |    0.00 |    210.9375 |           - |           - |           873.48 KB |
'ImageSharp Jpeg' | Bmp/Car.bmp | 6.071 ms | 0.1181 ms | 0.1406 ms | 5.995 ms |  1.28 |    0.11 |     23.4375 |           - |           - |           129.19 KB |

src/ImageSharp/Common/Extensions/StreamExtensions.cs
antonfirsov: Our "allocator" is typically also an allocation free pool, but we definitely have an overhead because of nested virtual calls, and other infrastructure stuff.
Everything depends on the size of the array. It's not uncommon to have very large (> 1 MB) buffers in image processing, which works better with our allocator/pool in my experience, but for those buffers we should avoid invoking this extension method anyways.

tannergooding: Definitely and I don't recall if benchmark.net isolates each iteration or not (I know it does for separate tests), so it could be that the benchmark is "lying" for full framework and it is hiding the allocation cost (so it might not be representative of real world scenarios).

It would be possible to take a `MemoryAllocator` as an optional parameter or to not have `SosHeaderYCbCr` take advantage of this compiler optimization (the single copy on class instantiation isn't terrible, but losing out on permanent pinning is a bit unfortunate) or to only not use the optimization on full framework if there are concerns about this regressing full framework.

IIRC, the `ArrayPool<byte>.Shared` is optimized for up to 2MB, since we use it for things like the JSON and XML readers.

JimBobSquarePants: Let's add a comment explaining why we use that pool (signature matching) and move on.

tannergooding: Fixed. Also rebased onto the current HEAD.

JimBobSquarePants:
Merging this in. Thanks @tannergooding

PR #870, opened 2019-04-01
Nits

iamcarbon:
### Prerequisites

### Description

A few small nits.

- Remove unused namespaces
- Add missing periods in documentation
- Replace Endianness enum with bool
- Use => to format a few shorter method bodies
- Use MathF instead of casting


iamcarbon:
Ready for review.

src/ImageSharp/ColorSpaces/CieLch.cs
```diff
@@ -13,15 +13,15 @@ namespace SixLabors.ImageSharp.ColorSpaces
...
-        private static readonly Vector3 Min = new Vector3(0, -200, 0);
-        private static readonly Vector3 Max = new Vector3(100, 200, 360);
-
         /// <summary>
         /// D50 standard illuminant.
         /// Used when reference white is not specified explicitly.
         /// </summary>
         public static readonly CieXyz DefaultWhitePoint = Illuminants.D50;
 
+        private static readonly Vector3 Min = new Vector3(0, -200, 0);
```
antonfirsov: Public fields should preceed private fields, right? I always get confused by these rules.

iamcarbon: Yep (and why I reordered this one).

src/ImageSharp/IO/Endianness.cs
```diff
@@ -1,21 +0,0 @@
-﻿// Copyright (c) Six Labors and contributors.
-// Licensed under the Apache License, Version 2.0.
-
-namespace SixLabors.ImageSharp.IO
-{
-    /// <summary>
-    /// Endianness of a converter
-    /// </summary>
-    internal enum Endianness
```
antonfirsov: Debate on taste 3 .. 2 .. 1 ...

For me - fine. @JimBobSquarePants ?

src/ImageSharp/ColorSpaces/CieLch.cs
JimBobSquarePants: I'm surprised StyleCop hadn't already picked up on this one.

src/ImageSharp/IO/Endianness.cs
JimBobSquarePants: Fine by me, simplicity always wins.

JimBobSquarePants:
> Add missing periods in documentation

@iamcarbon Did I ever tell you how much I love you 😍

src/ImageSharp/IO/DoubleBufferedStreamReader.cs
```diff
@@ -13,10 +13,10 @@ namespace SixLabors.ImageSharp.IO
     /// A stream reader that add a secondary level buffer in addition to native stream buffered reading
     /// to reduce the overhead of small incremental reads.
     /// </summary>
-    internal class DoubleBufferedStreamReader : IDisposable
+    internal sealed class DoubleBufferedStreamReader : IDisposable
```
JimBobSquarePants: Ah yes.... We should be sealing everything we can actually. There's definite advantages in the new compilers.

iamcarbon: Totally. There are some cases where it can avoid virtual calls when sealing the class. I think we've already got most of them.

iamcarbon:
I think this is a good chunk of stuff for this round of nits. :)

JimBobSquarePants:
@iamcarbon Just a heads up, those failed builds... Each pushed commit triggers a new build so I'm keeping an eye on things and cancelling all but the latest build each time you push to save build time.

iamcarbon:
@JimBobSquarePants thanks! not sure how i missed that stylecop violation.

iamcarbon:
# success

PR #872, opened 2019-04-03
Update SDKs and use C# 8.0

iamcarbon:
### Prerequisites


iamcarbon:
Waiting for the VS2019 images to land on CI.

src/ImageSharp/ImageSharp.csproj
```diff
@@ -25,7 +25,7 @@
     <DebugType Condition="$(codecov) == ''">portable</DebugType>
     <DebugSymbols>True</DebugSymbols>
     <Features>IOperation</Features>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>preview</LangVersion>
```
JimBobSquarePants: This could be an issue.

iamcarbon: This works with the v2.2.202 of the dotnet SDK that just got published. And will also work when the Visual Studio 2019 images go live.

antonfirsov:
I pressed "ready for review" accidentally.

src/ImageSharp/ImageSharp.csproj
antonfirsov: @dlemstra noticed an issue with the "wildcard" settings. Can't remember what it was, but we decided to go explicit. Even without understanding it, it seems safer to me if we want interop across different IDE-s / SDK-s.

appveyor.yml
```diff
@@ -1,5 +1,5 @@
 version: 1.0.0.{build}
-image: Visual Studio 2017
+image: Visual Studio 2019 Preview
```
antonfirsov: This is just an enabler for C# 8.0, leaving `target_framework` compatibility untouched. Is my understanding correct?

iamcarbon: Yep. We don't need to touch the target frameworks. I have things compiling locally -- just waiting on AppVeyor to update now.

iamcarbon: Things look good on Travis.

src/ImageSharp/Formats/Bmp/BmpInfoHeader.cs
```diff
@@ -372,21 +372,13 @@ public static BmpInfoHeader ParseOs2Version2(ReadOnlySpan<byte> data)
 
             // The compression value in OS/2 bitmap has a different meaning than in windows bitmaps.
             // Map the OS/2 value to the windows values.
-            switch (compression)
+            infoHeader.Compression = compression switch
```
JimBobSquarePants: Damn that's nice!

iamcarbon:
And we've landed!

iamcarbon:
And now travis is broke. Hahah.

iamcarbon:
Success! Ready for review. CC @JimBobSquarePants @antonfirsov 

This enables use to use static local functions, switch expressions, nullable reference types, and disposable ref structs.

Some additional features are enabled using the preview lang-version but I think these are still being worked on and will be enabled in an upcoming SDK.

vpenades:
@antonfirsov I'm still using 2017.  At home I'll probably switch to 2019 next week. At work it'll take a couple months. But it's fine since I'm only doing things with the source code of imagesharp at home.

I know some people that is waiting to switch to 2019 when NetCore 3 is released. This makes some sense since 2019 is required for NetCore 3, and it's what people actually want to target.

Another matter is if you're exposing C# 8 characteristics in the public API. In my experience, I've been using C# 7.3 "in" and "unmanaged" keywords. Nuget packages that target C# 7.3 can be consumed by c# 7.0 projects as long as you don't call any method using higher level features, in that case, it will throw an exception.

Sergio0694:
I already have VS2019 + .NET Core 3.0 preview on my PC, so this wouldn't be a breaking change for me personally. But:
- I agree with those who said that VS2019 is not that widespread yet, as many users are waiting for the final .NET Core 3.0 version to be released before upgrading. This would break the repo for them until they switch.
- I don't need any particular C# 8.0 features in my bokeh processor, so regarding my open PR in particular, this is fine either way
- Considering that ImageSharp can't even build on UWP right now, I'm a bit scared switching to a language preview version could mess things up even more there. 🙈
By that I mean, while some C# 8.0 features are basically just syntactic sugar (eg the `switch` statement), other actually require some runtime support, and I'm not sure how that'd play out for frameworks that right now don't officially support those (eg. UWP, or .NET Framework).

Just thinking out loud here 😄

dlemstra:
I am one of the people that will switch to 2019 after NetCore 3 has been released. But if there are more people who will want to switch already then feel free to do so. If the switch is made earlier I can always spin up a VM in Azure and install VS2019 on there and work from there.

JimBobSquarePants:
I'm reading a little confusion here...

.NET Core 3.0 and C#8 should be considered independantly. VS2019 does not mean we require .NET Core 3.0 nor should it be a consideration when choosing to move from 2017. Bug fixes, faster compilation, better refactorings etc should be your consideration here.

We also should not be concerned by certain C#8 features not being availabe on .NET Framework. We specifically target the framework so if something doesn't work the build will simply fail. 

That said, I agree with @antonfirsov here. I think we should split this PR into two. 

1. Build configuration changes updating the image to VS2019. This fixes issues we are having with Travis and also speeds up our build/test times on Appveyor. **We should merge those changes as soon as possible as our builds are blocked!**
2. Update the language version. We should be a lot more careful here and carefully consider where we make changes (no public APIs). I say hold off on this for now until we have a clearer picture of any possible side effects and a plan to update the whole suite of libraries.

@tannergooding You're our man on the inside. Does this make sense to you?

tannergooding:
@JimBobSquarePants, that makes sense to me.

VS2019 is released and supported, you should feel free to move to it as you would any new version. Community edition can also be used for contributing to open source projects, so I don't believe moving to it will block anyone.

C# 8 and .NET Core 3 are both still in preview and being worked on. If you want to support these, there are ways you can configure the infrastructure to require a minimum SDK version already be on box or for your `build.{cmd|sh}` script to pull it down locally from the internet (similarly to CoreFX, CoreCLR, Roslyn, and other projects). There will likely be more information released at Microsoft Build, and it may be good to wait until then before making a decision here.

jongleur1983:
I still don't use ImageSharp in any projects, just contributing for fun. I'm on VS2019 at home and at work already, but did not yet switch to C#8 anywhere (but I'm fine with doing so).

Sergio0694:
@JimBobSquarePants In case that "I see some confusion here" also referred to my answer, just to clarify, what I meant is that while I'm aware the VS2019 is completely independent from .NET Core 3.0 and C# 8.0, from what I saw lots of users are just waiting for .NET Core 3.0 to be publicly released before bothering to upgrade from VS2017.
And in that case, those users would no longer be able to contribute to the repo if we start to integrate C# 8.0 features, as those aren't available on VS2017.

Then of course, you could still go ahead with the C# 8.0 integration and eventually just force those users to upgrade to VS2019 now, I'm not saying that'd be a bad thing per se 😄

As far as I'm concerned, my only fear regarding C# 8.0 is how it'll play out with the .NET Native compiler on UWP, which as you well know, already just fails to compile ImageSharp.

brianpopow:
I am using VS2019 at home and at work, so i am fine with that. I also have no objection in switching to C# 8.0

antonfirsov:
@JimBobSquarePants wow, actually I wanted to accept the changes based on the positive responses from the community, but let's do the split then! :smile:

I will open a separate PR copying the SDK changes from this one very soon.

iamcarbon:
I would still push for us to adopt C# 8.0 as soon as possible to keep the project (and it's contributors) on the .NET wave culminating in .NETStandard 2.1/ .NETCOREAPP 3.0. While it's just syntax currently, it's important that our contributors adopt Visual Studio 2019 so there's no surprises when the library cross targets .NETSTANDARD 2.1 / .NETCOREAPP3.0.

iamcarbon:
Let's revisit this again soon.

JimBobSquarePants:
Thanks all for your contribution to the discussion. It's really great to read your input!

@iamcarbon I'm very keen on moving to C# 8. There's some stuff in there that I really like the look of that could be very useful to us. I just want to be cautious and work with a proper plan. ImageSharp is the end of a dependency chain and I think we should begin any migration efforts at Core so we can dogfood our changes as they propegate through that chain.

PR #877, opened 2019-04-07
Now throws a better excpetion DrawImage source does not overlap target

### LINKED ISSUE #875: Negative height rectangles cause ArgumentOutOfRangeException for property MaxDegreeOfParallelism in IterateRows
### Prerequisites

### Description
I made a mistake when cloning and cropping an image.

The problem is that the exception thrown was quite obscure. The error was related to `MaxDegreeOfParallelism` being invalid. Here is the stack trace:

```
SixLabors.ImageSharp.ImageProcessingException: An error occurred when processing the image using DrawImageProcessor`2. See the inner exception for more detail. ---> SixLabors.ImageSharp.ImageProcessingException: An error occurred when processing the image using DrawImageProcessor`2. See the inner exception for more detail. ---> System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: MaxDegreeOfParallelism
   at System.Threading.Tasks.ParallelOptions.set_MaxDegreeOfParallelism(Int32 value)
   at SixLabors.ImageSharp.ParallelUtils.ParallelHelper.IterateRows(Rectangle rectangle, ParallelExecutionSettings& parallelSettings, Action`1 body)
   at SixLabors.ImageSharp.Processing.Processors.Drawing.DrawImageProcessor`2.OnFrameApply(ImageFrame`1 source, Rectangle sourceRectangle, Configuration configuration)
   at SixLabors.ImageSharp.Processing.Processors.ImageProcessor`1.Apply(ImageFrame`1 source, Rectangle sourceRectangle, Configuration configuration)
   --- End of inner exception stack trace ---
   at SixLabors.ImageSharp.Processing.Processors.ImageProcessor`1.Apply(ImageFrame`1 source, Rectangle sourceRectangle, Configuration configuration)
   at SixLabors.ImageSharp.Processing.Processors.ImageProcessor`1.Apply(Image`1 source, Rectangle sourceRectangle)
   --- End of inner exception stack trace ---
   at SixLabors.ImageSharp.Processing.Processors.ImageProcessor`1.Apply(Image`1 source, Rectangle sourceRectangle)
   at SixLabors.ImageSharp.Processing.DefaultInternalImageProcessorContext`1.ApplyProcessor(IImageProcessor`1 processor, Rectangle rectangle)
   at SixLabors.ImageSharp.Processing.DefaultInternalImageProcessorContext`1.ApplyProcessor(IImageProcessor`1 processor)
   at SixLabors.ImageSharp.Processing.DrawImageExtensions.DrawImage[TPixelDst,TPixelSrc](IImageProcessingContext`1 source, Image`1 image, Point location, GraphicsOptions options)
```

I've tracked down the root cause to:
https://github.com/SixLabors/ImageSharp/blob/2849fd9c1dc19a7338235f6c33f4970e5a12dee3/src/ImageSharp/Common/ParallelUtils/ParallelHelper.cs#L50-L64

If the rectangle height is negative, then `maxSteps` will be negative, which will then set the `MaxDegreeOfParallelism` to a negative value.

### Steps to Reproduce

It took me ages to work out what happened! I got my `top` and `left` variables back to front.

My crop was: 
```
image.Mutate(x => x.DrawImage(source, new Point(top, left), GraphicsOptions.Default));
```

And should have been:

```
image.Mutate(x => x.DrawImage(source, new Point(left, top), GraphicsOptions.Default));
```

### System Configuration

- ImageSharp version: (local build) 2849fd9c1dc19a7338235f6c33f4970e5a12dee3
- Other ImageSharp packages and versions: (local build) 2849fd9c1dc19a7338235f6c33f4970e5a12dee3
- Environment (Operating system, version and so on): Win 10, 1809
- .NET Framework version: ".NETFramework,Version=v4.6.2"
- Additional information:

I think the ideal solution here is a couple more guards to provide a better error message when someone (like me) does something silly and messes up their coordinates.

atruskie:
### Prerequisites

### Description
Previously, when DrawImage was used to overlay an image, in cases where the source image did not overlap the target image, a very confusing error was reported: "System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: MaxDegreeOfParallelism"

Now, when this case happens, the DrawImage method will simply not affect the target image, which is the same way FillRegionProcessor handles such cases.

ParallelHelper.IterRows also now does more validation of the input rectangle so that any further cases of this kind of problem throw a more relevant exception. Note I switched from DebugGuard to Guard because IterRows is a public API and thus should always validate its inputs.

Fixes #875

Other notes: I had trouble getting the full test suite to pass on my laptop. My individual tests seem to be running fine. So I've created this PR to see what the CI says.


src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawImageProcessor.cs
```diff
@@ -81,6 +81,12 @@ protected override void OnFrameApply(ImageFrame<TPixelDst> source, Rectangle sou
 
             var workingRect = Rectangle.FromLTRB(minX, minY, maxX, maxY);
 
+            if (workingRect.Width < 0 || workingRect.Height < 0)
```
dlemstra: Wondering if this should be `<= 0` instead?

src/ImageSharp/Common/ParallelUtils/ParallelHelper.cs
```diff
@@ -45,7 +45,8 @@ public static void IterateRows(Rectangle rectangle, Configuration configuration,
             in ParallelExecutionSettings parallelSettings,
             Action<RowInterval> body)
         {
-            DebugGuard.MustBeGreaterThan(rectangle.Width, 0, nameof(rectangle));
+            Guard.MustBeGreaterThan(rectangle.Width, 0, nameof(rectangle));
```
dlemstra: Maybe something like `$"{nameof(rectangle)}.{nameof(rectangle.Width)}"` or `nameof(rectangle.Width)` instead to make it clear which value is incorrect?

src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawImageProcessor.cs
atruskie: It should be yes. Changed in 4411ddd

src/ImageSharp/Common/ParallelUtils/ParallelHelper.cs
atruskie: Done

atruskie:
@JimBobSquarePants I'm sorry but I don't understand your comment.

This https://github.com/SixLabors/ImageSharp/pull/877/files#diff-c079dda36daaace2698dece840baec21R48 change does cause parallel helper to throw if given invalid input.

Additionally, new behavior is that DrawImage is now a no-op when the image does not overlap. Personally I'd prefer DrawImage to throw in this case but when implementing this fix, I tried to make the behavior as close as I could to existing behavior. I thought the code in FillImageProcessor was a good example of intended behavior: 

https://github.com/SixLabors/ImageSharp/blob/0dfa2f9f98fb93cf1fc9d082df5d08affc4be5a8/src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor.cs#L69-L77

Would you prefer DrawImage to throw on non-overlapping cases? 
(I apologize: I realized after I pushed changes that I should have asked for intended behavior before writing a fix)

JimBobSquarePants:
No need to apologise! I’m grateful for any assistance. 

I don’t think a no-op is wise as that leads to unexpected behaviour. The change to guard in the helper is good I think we should improve the error message from the default.

atruskie:
I've addressed the feedback, however, I'm pretty unhappy with the gymnastics required to expect an exception from an ImageProcessor. Any suggestions?

src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawImageProcessor.cs
```diff
@@ -81,6 +81,12 @@ protected override void OnFrameApply(ImageFrame<TPixelDst> source, Rectangle sou
 
             var workingRect = Rectangle.FromLTRB(minX, minY, maxX, maxY);
 
+            // not a valid operation because rectangle does not overlap with this image.
+            Guard.IsFalse(
```
antonfirsov: In my opinion, we only use `Guard` for trivial argument validation at the beginning of the methods. 

I think this is what we should do:
- Throw an `ImageProcessingException` if the condition is not matched
- In `ImageProcessor<T>.Apply(...)` we should detect (catch and rethrow) `ImageProcessingException`

@dlemstra @JimBobSquarePants thoughts?

src/ImageSharp/Common/ParallelUtils/ParallelHelper.cs
```diff
@@ -142,5 +144,18 @@ public static void IterateRows(Rectangle rectangle, Configuration configuration,
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static int DivideCeil(int dividend, int divisor) => 1 + ((dividend - 1) / divisor);
+
+        private static void ValidateRectangle(Rectangle rectangle)
+        {
+            Guard.IsTrue(
+                rectangle.Width > 0,
+                $"{nameof(rectangle)}.{nameof(rectangle.Width)}",
+                "Can't iterate rows on a with a rectangle that has a width less than or equal to 0");
```
JimBobSquarePants: We have `Guard.MustBeGreaterThan` which will throw the correct `ArgumentOutOfRangeException`. I don't think we need a custom message then.

src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawImageProcessor.cs
JimBobSquarePants: I think this check is superfluous and the errors in `ParallelHelper.ValidateRectangle` should be `ArgumentOutOfRangeException` via `Guard.MustBeGreaterThan`.

We already capture the inner exception and throw an `ImageProcessingException` in release mode. 

@atruskie sorry to be messing you around here.

antonfirsov: My comment was actually addressing a concern from @atruskie:
>I 'm pretty unhappy with the gymnastics required to expect an exception from an ImageProcessor. Any suggestions?

However, the more I think about fixing this, the further it leads, so let's not waste time on it in this PR.

src/ImageSharp/Common/ParallelUtils/ParallelHelper.cs
atruskie: Ok I'll revert this to a guard

src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawImageProcessor.cs
atruskie: @antonfirsov I've followed your advice here already, and it leads to a nicer test.

@JimBobSquarePants I'd like to leave both errors in because they represent different types of error. One is descriptive for a user mistake. The other, in IterRows, really is exceptional and represents that that code should never have been called with the given argument.

JimBobSquarePants: I’m trying to avoid as much as possible, per-processor error handling. A negative rectangle would cause an error in many places and multiple handlers would have to be added. 

I do wonder whether we should be using Recangle.Intersect to avoid the possibility as we do in several places?

atruskie: Sorry, I didn't see this last message before I pushed.

I feel like I'm not doing a good job at understanding what you'd like. The options:

1. Throw in DrawImage if there is not an overlap with a descriptive error message (my preference)
2. Have no-effect in DrawImage if there is not an overlap (my original version of this fix, possibly what you mean by doing Rectangle.Intersect checks)?
3. Rely on the exception in IterRows when a non-overlap occurs (I think this is only marginally better than the original confusing error message this PR is seeking to address)
4. Some other option?

I'm happy to keep making changes until I get this right :-) , it just feels a little like walking in circles.

JimBobSquarePants: In retrospect I think you're doing the correct thing for now with the API we have. I'm going to update and merge this. Thanks for your help! 😄

Issue #886, opened 2019-04-16
bmp written is not supported by Chrome or Firefox


pammann-work:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When saving image as bmp it can not be seen with Chrome or Firefox, however it can be read by MS Paint.

This started to happen with 1.0.0-beta0006 when i compare the produced image with one from 1.0.0-beta0005 i can see that the bmp header is different....

[images.zip](https://github.com/SixLabors/ImageSharp/files/3085593/images.zip)

### Steps to Reproduce
``` c#
using (var image = Image.Load<Rgb24>(mem))
{
    // issue here
    image.SaveAsBmp(fullMem);
}
```

### System Configuration
Seen under Windows and Linux

- ImageSharp version: 1.0.0-beta0006
- Other ImageSharp packages and versions:
- Environment (Operating system, version and so on):
- .NET Framework version:
- Additional information:

pammann-work:
![header_compare](https://user-images.githubusercontent.com/15227462/56222451-5af12b80-606c-11e9-8f2b-1c2d2b9cc126.png)

JimBobSquarePants:
Odd. V4 bitmaps should be supported. @brianpopow do you have any insight here?

http://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html

brianpopow:
we have indeed changed the bitmap header in beta0006 to V4 to support bitmaps with transparency. As @JimBobSquarePants said, chrome and firefox should support those.

I checked the imagesharp bitmap encoder and at first glance i can not spot an error. 
I was thinking maybe we should write V3 bitmap header as default and keep writing V4 headers as optional. I will look further into this the next days.

JimBobSquarePants:
Thanks @brianpopow appreciate it! That sounds like a sensible plan.

pammann-work:
update: gimp can also not read the image produced with beta006, it complains with:
"Unrecognized or invalid BMP compression format."

pammann-work:
[...]

PR #889, opened 2019-04-20
Bitmap encoder writes V3 header as default

brianpopow:
### Prerequisites

### Description

Some bitmap decoders seem to have problems with Bitmap V4 header in combination with 24 bits per pixel and the BITFIELDS compression type (see: #886).

This PR changes the Bitmap encoder to write a V3 header as default. This should be supported by any decoders.
Introduces a new encoder option `SupportTransparency`: With this option a V4 header will be written with BITFIELDS compression.


tests/ImageSharp.Tests/Formats/Bmp/BmpEncoderTests.cs
```diff
@@ -102,15 +104,43 @@ public void Encode_IsNotBoundToSinglePixelType<TPixel>(TestImageProvider<TPixel>
...
             where TPixel : struct, IPixel<TPixel> => TestBmpEncoderCore(provider, bitsPerPixel);
 
-        private static void TestBmpEncoderCore<TPixel>(TestImageProvider<TPixel> provider, BmpBitsPerPixel bitsPerPixel)
+        [Theory]
+        [WithFile(Bit32Rgb, PixelTypes.Rgba32 | PixelTypes.Rgb24, BmpBitsPerPixel.Pixel32)]
+        [WithFile(Bit32Rgba, PixelTypes.Rgba32 | PixelTypes.Rgb24, BmpBitsPerPixel.Pixel32)]
+        [WithFile(WinBmpv4, PixelTypes.Rgba32 | PixelTypes.Rgb24, BmpBitsPerPixel.Pixel32)]
+        [WithFile(WinBmpv5, PixelTypes.Rgba32 | PixelTypes.Rgb24, BmpBitsPerPixel.Pixel32)]
+        // WinBmpv3 is a 24 bits per pixel image
+        [WithFile(WinBmpv3, PixelTypes.Rgb24, BmpBitsPerPixel.Pixel24)]
```
antonfirsov: 👍 these tests are great! 💯

brianpopow:
I just wanted to clarify that rejecting bitmaps with BITFIELDS and 24 bpp is a valid thing to do. While re-reading the bitmap documentation again while i was making the changes for this PR, BITFIELDS should only be used with 16 and 32 bits per pixel. 
It was a mistake, that i have allowed encoding 24 bpp with BITFIELDS.

JimBobSquarePants:
I'll get this reviewed asap

src/ImageSharp/Formats/Bmp/BmpEncoder.cs
```diff
@@ -18,6 +18,11 @@ public sealed class BmpEncoder : IImageEncoder, IBmpEncoderOptions
         /// </summary>
         public BmpBitsPerPixel? BitsPerPixel { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the encoder should support transparency.
+        /// </summary>
+        public bool SupportTransparency { get; set; }
```
JimBobSquarePants: Wouldn't `BmpBitsPerPixel.Pixel32` implicitly mean this?

brianpopow:
@JimBobSquarePants 
> Wouldn't BmpBitsPerPixel.Pixel32 implicitly mean this?

It is complicated. With a V3 header and 32 bits per pixel (and compression type `no compression`), the 4th byte is just considered a padding byte not actually a alpha channel. This is how a V3 palette entry looks like:

```
typedef struct _Win3xPaletteElement
{
	BYTE Blue;      /* Blue component */
	BYTE Green;     /* Green component */
	BYTE Red;       /* Red component */
	BYTE Reserved;  /* Padding (always 0) */
} WIN3XPALETTEELEMENT;
````

My usual reference source is: [fileformat.info](http://www.fileformat.info/format/bmp/egff.htm)
Some decoders (imagesharp and mozilla included) try to guess if there is an alpha channel present: If there is a none zero 4th byte present, it has an alpha component (see issue #732).

I think the correct way to handle transparency is with the BITFIELDS compression type. 

If we would follow the documentation strictly, we should make the image opaque when we write a V3 header. This is currently not the case. It would not hurt any decoder, because most of them probably ignore the 4th byte rightfully anyway. What is your opinion on this, should we make it opaque with V3 headers?

I could try to expand the comment on `SupportTransparency` to make this more clear. 

My first thought when i was implementing this was add an option which header to write (V3 or V4) and
another option compression type: (none or BITFIELDS), but which end user would actually understand the implication of those anyway. Also this adds another level of complexity, for example V3 + BITFIELDS does not mean transparency is supported, that would be the compression type: BI_ALPHABITFIELDS (and BI_ALPHABITFIELDS does not make much sense in combination with V4 header).

**Comment by JimBobSquarePants** [comment]
@brianpopow Thanks for the thorough explanation. That all makes sense. 

Let's flesh out the comment on the `SupportTransparency` property then dictating exactly what setting are implemented when the property is set.

**Comment by brianpopow** [comment]
I have changed the comment on the `SupportTransparency` option with the implication to the encoder.

Im still on the fence whether or not we should set the alpha / reserved byte to 0, if we write a V3 header with 32 bbp. I have not changed the behavior of the encoder behavior in that regard. If for example a png with an alpha channel is loaded and the again saved as bmp with a V3 header, the encoder would write the alpha channel loaded from the original image. 
All decoders i know of can decode those images, but only some (e.g. gimp, imagemagick, imagesharp) will look for none zero values in the alpha channel to determine if there is transparency present. Others will decode it as full opaque, which is also correct behavior according to the documentation for V3 header bitmaps.

File: src/ImageSharp/Formats/Bmp/BmpEncoder.cs [hunk 3]
```diff
@@ -18,6 +18,14 @@ public sealed class BmpEncoder : IImageEncoder, IBmpEncoderOptions
         /// </summary>
         public BmpBitsPerPixel? BitsPerPixel { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the encoder should support transparency.
+        /// Note: Transparency support only works together with 32 bits per pixel. This option will
+        /// change the default behavior of the encoder of writing a bitmap version 3 info header with no compression.
+        /// Instead a bitmap version 4 info header will be written with the BITFIELDS compression.
```
JimBobSquarePants: Perfect!

PR #897, opened 2019-05-01
API proposal: Pixel-agnostic Image base class

antonfirsov:
This is an API proposal (+POC), communicated in the form of a pull request, not intended for merging. The target branch contains stripped ImageSharp source code, so we can focus on the important changes.

Please, feel free to review it, any feedback is welcome!

## The problem
This proposal aims to solve the following issues with the current API:
- The majority of our users (eg. thumbnail makers) are not interested in the pixel type backing the `Image<TPixel>` instance. They just want to apply a processor, and resave the image. `TPixel`, `Rgba32` is only a noise for them.
- `Image.Load(...)` is decoding images into into `Image<Rgba32>` in an arbitrary manner. This is sub-optimal in many cases. Eg. jpegs do not have an alpha-channel, therefore it's better to decode them into `Image<Rgb24>`. It would be benefitical to solve this in a way that's transparent to the user.
- It's not possible to make the current ImageProcessor implementations public without exposing a bunch of implementation details. This means that any refactor and optimization could introduce breaking changes in the future.
- From API perspective, it's not necessary for ImageProcessors and extension methods to be pixel-type specific.

## The solution
- Making `Image` an abstract, pixel-agnostic base class for `Image<TPixel>`
- `Image.Load(....)` overloads return an `Image` instead of `Image<Rgba32>`
- `Image.Load<TPixel>(....)` overloads keep returning `Image<TPixel>`
- Introcuce a pixel-agnostic `IImageProcessor` and `IImageProcessingContext` interfaces
- Separate API from implementation in Image Processors. The "definition" class implements `IImageProcessor`, it's only responsible to encapsulate the parameters, and create the pixel specific `IImageProcessor<TPixel>` implementation.
- The majority of the extension methods could operate on the pixel-agnostic `Image` and `IImageProcessingContext` (See `ResizeExtensions` in the PR changes).
- For stuff like drawing, users should be aware of the pixel type.
- The "trick" is done by [Double Dispatch](https://en.wikipedia.org/wiki/Double_dispatch), but it makes sense to follow the terminology defined by by the [Visitor Pattern](https://en.wikipedia.org/wiki/Visitor_pattern).

## Impact
This is a breaking change, however it's scope is very limited, most of the users should be able to adapt the new API with minimal or no code changes:

- All users should recompile their project after package update
- Code using specific pixel types when loading an image should be unaffected:
```C#
using (Image<Rgb24> image = Image.Load<Rgb24>("foo.jpg"))
{
    image.Mutate(x => x
         .Resize(image.Width / 2, image.Height / 2);
    image.Save("bar.jpg");
}
```

- Processing code using `var` should be unaffected:
```C#
using (var image = Image.Load("foo.jpg"))
{
    image.Mutate(x => x
         .Resize(image.Width / 2, image.Height / 2);
    image.Save("bar.jpg");
}
```

- Code which used to specify the `Image<Rgba32>` class when using `Image.Load(...)` could be easily adapted by replacing `Image<Rgba32>` with `Image`:
```C#
{
    image.Mutate(x => x
         .Resize(image.Width / 2, image.Height / 2);
    image.Save("bar.jpg");
}
```

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoder.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoder.cs
@@ -38,5 +38,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
                 return decoder.Identify(stream);
             }
         }
+        
+        public Image Decode(Configuration configuration, Stream stream) => this.Decode<Rgba32>(configuration, stream);
     }
 }
--- a/src/ImageSharp/Image.cs
+++ b/src/ImageSharp/Image.cs
@@ -1,0 +1,29 @@
+// Copyright (c) Six Labors and contributors.
+// Licensed under the Apache License, Version 2.0.
+
+using SixLabors.ImageSharp.Formats;
+using SixLabors.ImageSharp.Metadata;
+
+namespace SixLabors.ImageSharp
+{
+    public abstract partial class Image : IImage
+    {
+        /// <inheritdoc/>
+        public PixelTypeInfo PixelType { get; }
+
+        public abstract int Width { get; }
+        
+        public abstract int Height { get; }
+        
+        /// <inheritdoc/>
+        public ImageMetadata Metadata { get; }
+
+        protected Image(PixelTypeInfo pixelType, ImageMetadata metadata)
+        {
+            this.PixelType = pixelType;
+            this.Metadata = metadata ?? new ImageMetadata();
+        }
+
+        public abstract void Dispose();
+    }
+}
--- a/src/ImageSharp/Image.cs
+++ b/src/ImageSharp/Image.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Six Labors and contributors.
 // Licensed under the Apache License, Version 2.0.
 
+using System.IO;
+
 using SixLabors.ImageSharp.Advanced;
 using SixLabors.ImageSharp.Formats;
 using SixLabors.ImageSharp.Metadata;
@@ -45,5 +47,39 @@ namespace SixLabors.ImageSharp
         public abstract void Dispose();
 
         internal abstract void AcceptVisitor(IImageVisitor visitor);
+
+        /// <summary>
+        /// Saves the image to the given stream using the given image encoder.
+        /// </summary>
+        /// <param name="stream">The stream to save the image to.</param>
+        /// <param name="encoder">The encoder to save the image with.</param>
+        /// <exception cref="System.ArgumentNullException">Thrown if the stream or encoder is null.</exception>
+        public void Save(Stream stream, IImageEncoder encoder)
+        {
+            Guard.NotNull(stream, nameof(stream));
+            Guard.NotNull(encoder, nameof(encoder));
+
+            EncodeVisitor visitor = new EncodeVisitor(encoder, stream);
+            this.AcceptVisitor(visitor);
+        }
+
+        class EncodeVisitor : IImageVisitor
+        {
+            private readonly IImageEncoder encoder;
+
+            private readonly Stream stream;
+
+            public EncodeVisitor(IImageEncoder encoder, Stream stream)
+            {
+                this.encoder = encoder;
+                this.stream = stream;
+            }
+
+            public void Visit<TPixel>(Image<TPixel> image)
+                where TPixel : struct, IPixel<TPixel>
+            {
+                this.encoder.Encode(image, this.stream);
+            }
+        }
     }
 }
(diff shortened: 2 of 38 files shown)


src/ImageSharp/Image.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    internal interface IImageVisitor
+    {
+        void Visit<TPixel>(Image<TPixel> image)
+            where TPixel : struct, IPixel<TPixel>;
+    }
+    
+    public abstract partial class Image : IImage, IConfigurable
+    {
+        protected readonly Configuration configuration;
```
JimBobSquarePants: `private`

src/ImageSharp/Image.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.IO;
+
+using SixLabors.ImageSharp.Advanced;
+using SixLabors.ImageSharp.Formats;
+using SixLabors.ImageSharp.Metadata;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp
+{
+    internal interface IImageVisitor
```
JimBobSquarePants: Let's move this.

src/ImageSharp/Formats/Jpeg/JpegDecoder.cs
```diff
@@ -38,5 +38,7 @@ public IImageInfo Identify(Configuration configuration, Stream stream)
                 return decoder.Identify(stream);
             }
         }
+        
+        public Image Decode(Configuration configuration, Stream stream) => this.Decode<Rgba32>(configuration, stream);
```
JimBobSquarePants: This will become `Rgb24` yes?

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeProcessorImplementation.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        /// <summary>
+        /// Gets the sampler to perform the resize operation.
+        /// </summary>
+        public IResampler Sampler => this.parameterSource.Sampler;
+
+        /// <summary>
+        /// Gets the target width.
+        /// </summary>
+        public int Width => this.parameterSource.Width;
```
JimBobSquarePants: These properties can all be private fields now unless you're expecting inheritance.

src/ImageSharp/Formats/Jpeg/JpegDecoder.cs
antonfirsov: It will delegate into a method of `JpegDecoderCore` which can then decide what pixel type to use in the resulting `Image<T>`. For grayscale jpegs we can return `Image<Gray8>` (if worth).

Designing the code sharing between the specialized (generic) and "native pixel type" (non-generic) methods is an interesting question. Probably even more challenging for `PngDecoder`.

antonfirsov:
Cool, thanks for checking! This code is not production ready in any way, I just made the fastest possible route to make the POC work. I will start from scratch and/or rebase.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeProcessorImplementation.cs
```diff
@@ -0,0 +1,184 @@
...
+using SixLabors.Primitives;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Transforms
+{
+    /// <summary>
+    /// Provides methods that allow the resizing of images using various algorithms.
+    /// Adapted from <see href="http://www.realtimerendering.com/resources/GraphicsGems/gemsiii/filter_rcg.c"/>
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    internal class ResizeProcessorImplementation<TPixel> : TransformProcessorBase<TPixel>
```
antonfirsov: @JimBobSquarePants I wonder whether I should keep the name `ResizeProcessor<TPixel>` instead. For certain processors there will be no pixel-agnostic variant, therefore it might be better to use a uniform naming.

JimBobSquarePants: I prefer that over Implementation. File name can be ResizeProcessorTPixel.cs I think and pass stylecop naming rules.

src/ImageSharp/Formats/Jpeg/JpegDecoder.cs
JimBobSquarePants: That makes sense. Thinking about it shouldn’t be too bad for png either.

Issue #621, opened 2018-06-17
Make ImageProcessor-s public


antonfirsov:
Currently all processors are internal. We agreed that we should open up their API for 1.0.0-RC1, after an extensive review.

DanStout:
Yes, please! I wanted to split the Sobel edge detection processor into separate X and Y processors, but found that all the required classes and their dependencies are all marked internal.

JimBobSquarePants:
Yeah we’ve been keeping them internal to work on the API. Is there something missing from the current edge detection processors? We could work together to fix that?

DanStout:
The one in particular I was looking for was to be able to use the Sobel processor on X and Y independently. My plan was at first just to create two subclasses of EdgeDetectorProcessor - SobelXProcessor and SobelYProcessor

JimBobSquarePants:
Perhaps we should make that operation an Enum?

DanStout:
As in with values X, Y, and XY? That would make sense, I think.

JimBobSquarePants:
I'll have to have a look to see if we can use separableness without breaking anything. Feel free to have a dig around yourself.

vpenades:
I would consider opening also the pixel blender API, at least the method that lets you process a single pixel.

Maybe its not even required to expose the whole PixelBlender<TPixel> class.... Personally, I would consider adding a method that returns a delegate, something like this:

```
delegate TPixel PixelFunction<TPixel>(TPixel backdrop, TPixel source, float opacity);

PixelFunction<TPixel>  <somewhere>.GetPixelBlendingFunction(PixelAlphaComposition composition, PixelColorBlending blending);
```

Where <somewhere> would be TPixel itself.

antonfirsov:
@vpenades I'm strongly against per-pixel delegates, because it's a performance-killer.

vpenades:
@antonfirsov I know, I am aware of that, and I guess you want to avoid exposing them to prevent end users implementing linear operations using naive loops.

but there's some scenarios in which they can be useful, specially for random pixel access.

Let's say I want to render a star field, drawing random pixels ?  

There's lots of procedural algorythms that require non linear access to the pixels, and they cannot benefit from the pixel blenders.

The only alternative I could find around this is to add a .Mutate( c->DrawPointList(...) );  which could be parallelized.... but most probably users would call it using one pixel at a time.  But i'm sure it would still be much faster than drawing 1 pixel sized rectangles which is what I'm using right now.

[...]

PR #904, opened 2019-05-05
Introduce a non-generic Image base class

antonfirsov:
### Prerequisites

### Description
- Implements the proposal demonstrated in #897.
- Change majority of the extension methods working on `Image<TPixel>` to work on `Image` instead:
```C#
// The non-generic Image.Load overloads no longer return Image<Rgba32>
{
    image.Mutate(x => x.Resize(100, 100);
    image.SaveAsPng(someStream);
}
```

- Touches #621 by making most of the non-generic processors public. (Exception: Histogram processors, because I'm not entirely happy with all namings.)
- Fix a few doc issues while doing the refactor.
- While implementing better test coverage for `Image.Load` variants, discovered and added a few missing overloads

Hope I haven't missed any of the important processors and / or extensions!

To enable extensions using color values as parameters on `Image`, we need to introduce a pixel-agnostic color type. I have [a proposal](https://gist.github.com/antonfirsov/ba3b3fc05e7aba19458d6d61c8b38e2c) for this, planning to PR it within the next week.


JimBobSquarePants:
I'm very happy for the most part with this thus far. I'd like to know more about your concerns with the Histogram naming and would like to help there.

antonfirsov:
`AdaptiveHistEqualizationSWProcessor` - too many abbrevations.

On the other hand, `AdaptiveHistogramEqualizationSlidingWindowProcessor` would be way too long. I don't know what is the good solution.

JimBobSquarePants:
I think we’ve still got a long way to go before we hit the dizzy heights of https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/BoundTree/BoundTreeRewriter.cs#L100

So I’m ok with the longer form

antonfirsov:
Ok, went with the longer names. All processors without color parameters should be `public (sealed)` now.

src/ImageSharp/Image{TPixel}.cs
```diff
@@ -3,26 +3,24 @@
...
 namespace SixLabors.ImageSharp
 {
     /// <summary>
     /// Encapsulates an image, which consists of the pixel data for a graphics image and its attributes.
+    /// For generic <see cref="Image{TPixel}"/>-s the pixel type is known at compile time.
     /// </summary>
     /// <typeparam name="TPixel">The pixel format.</typeparam>
-    public sealed class Image<TPixel> : IImage, IConfigurable
+    public sealed class Image<TPixel> : Image
         where TPixel : struct, IPixel<TPixel>
```
iamcarbon: Consider unmanaged constraint?

antonfirsov:
I'm merging this to go on with the rest of #907. If I made a mistake, I will be able to fix it in the follow up PR-s.

Issue #907, opened 2019-05-09
Epic: introduce pixel-agnostic API across the whole library


antonfirsov:
I'm opening this to track the API changes I'm planning to introduce in multiple PR-s:
- [x] Adding the non-generic `Image` base class, refactor and publish processors without color parameters. (#904)
- [x] Introduce pixel [independant color type](https://gist.github.com/antonfirsov/ba3b3fc05e7aba19458d6d61c8b38e2c), refactor and publish processors having color values as parameters.
- [x] Make **ImageSharp.Drawing** API non-generic.
- [x] Do a sanity check on the resulting API set. Drop (or consider dropping) generic classes made obsolete by the new API (`ColorBuilder<T>`, `NamedColors<T>` and `ImageProcessingContext<T>`)
- [x] Define (or at least considerdefining) a non-generic variant of `ImageFrameCollection`
- [x] Update SixLabors/Samples (also taking [other aspects](https://github.com/SixLabors/ImageSharp/issues/931#issuecomment-501575653) into account)
- [x] Anaylize impact on AOT, extend `AotCompilerTools`. **EDIT:** Won't do it for RC1. Moved to separate issue: #946.

After finishing the changes, I'm planning to publish a blog post to explain the new API and help existing users with a migration guide.

antonfirsov:
All points are finished here, including the update of the [Samples](https://github.com/SixLabors/Samples), closing!

PR #908, opened 2019-05-11
Introduce representation-agnostic Color type

antonfirsov:
### Prerequisites

## Description
- Next step in #907
- Adding a general `Color` struct with a hidden internal hi-res representation, that is convertible to all pixel types
    - Add implicit conversion methods to/from RGB(A)-like pixel types.
    - Add all color constants to this type. (Like `Color.Black` or `Color.WebSafePalette`)
- Make all remaining processors and processing extensions in the ImageSharp package pixel-agnostic. Color parameters (including palette) are using the new `Color` type now.
    - (ImageSharp.Drawing has not changed yet: see #907)
- Refactor quantization:
    - Use `Color` on all public API-s (Consequence: `PaletteQuantizer<TPixel>` has been dropped.)
    - Use `ReadOnlyMemory<T>` on public API-s instead of `T[]`
    - `QuantizedFrame<T>.GetPixelSpan()` returns `ReadOnlySpan<T>`. Populating the span is an implementation detail. Introduce `IQuantizedFrame<T>` to maintain clean abstractions here.
- Miscellaneous chore work:
    - Rename `ColorBuilder<T>.FromRGBA()` to `FromRgba`
    - Move processing extension classes to a (non-namespace-providing) subfolder

### Consequence on our terminlogy
**Color ≠ Pixel** any longer. Color is a general term, pixel is a representation drawable on an `Image<TPixel>`.


src/ImageSharp/Color/Color.NamedColors.cs
```diff
@@ -0,0 +1,721 @@
...
+{
+    /// <content>
+    /// Contains static named color values.
+    /// </content>
+    public readonly partial struct Color
+    {
+        /// <summary>
+        /// Represents a <see paramref="Color"/> matching the W3C definition that has an hex value of #F0F8FF.
+        /// </summary>
+        public static readonly Color AliceBlue = FromRgba(240, 248, 255, 255);
```
JimBobSquarePants: `public static readonly Color AliceBlue = Rgba32.AliceBlue;` etc would be better here. That way we're only declaring component values in one place.

src/ImageSharp/Color/Color.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        /// <summary>
+        /// Creates a new <see cref="Color"/> instance from the string representing a color in hexadecimal form.
+        /// </summary>
+        /// <param name="hex">
+        /// The hexadecimal representation of the combined color components arranged
+        /// in rgb, rgba, rrggbb, or rrggbbaa format to match web syntax.
+        /// </param>
+        /// <returns>Returns a <see cref="Color"/> that represents the color defined by the provided RGBA hex string.</returns>
+        public static Color FromHex(string hex)
```
JimBobSquarePants: Reuse `Rgba32.FromHex(string hex)` here.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -160,16 +158,19 @@ private void EncodeGlobal<TPixel>(Image<TPixel> image, QuantizedFrame<TPixel> qu
                 }
                 else
                 {
-                    using (IFrameQuantizer<TPixel> palleteFrameQuantizer = palleteQuantizer.CreateFrameQuantizer(image.GetConfiguration()))
-                    using (QuantizedFrame<TPixel> paletteQuantized = palleteFrameQuantizer.QuantizeFrame(frame))
+                    using (IFrameQuantizer<TPixel> palleteFrameQuantizer =
+                        new PaletteFrameQuantizer<TPixel>(this.quantizer.Diffuser, quantized.Palette))
```
JimBobSquarePants: This is all so much nicer! 😍

src/ImageSharp/Formats/Gif/LzwEncoder.cs
```diff
@@ -180,7 +180,7 @@ public LzwEncoder(MemoryAllocator memoryAllocator, int colorDepth)
         /// </summary>
         /// <param name="indexedPixels">The span of indexed pixels.</param>
         /// <param name="stream">The stream to write to.</param>
-        public void Encode(Span<byte> indexedPixels, Stream stream)
+        public void Encode(ReadOnlySpan<byte> indexedPixels, Stream stream)
```
JimBobSquarePants: Good!

src/ImageSharp/Color/Color.cs
```diff
@@ -0,0 +1,193 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The internal representation and layout of this structure is hidden by intention.
+    /// It's not serializable, and it should not be considered as part of a contract.
+    /// Unlike System.Drawing.Color, <see cref="Color"/> has to be converted to a specific pixel value
+    /// to query the color components.
+    /// </remarks>
+    public readonly partial struct Color : IEquatable<Color>
+    {
+        private readonly Rgba64 data;
```
JimBobSquarePants: Are we not going to lose accuracy when converting to/from `RgbaVector` here due to rounding. Do we even need the vector type?

src/ImageSharp/Color/Color.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <inheritdoc />
+        public bool Equals(Color other)
+        {
+            return this.data.PackedValue == other.data.PackedValue;
+        }
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj))
```
JimBobSquarePants: `if (obj is null)`

src/ImageSharp/Color/Color.WernerPalette.cs
```diff
@@ -0,0 +1,135 @@
...
+    {
+        private static readonly Lazy<Color[]> WernerPaletteLazy = new Lazy<Color[]>(CreateWernerPalette, true);
+
+        /// <summary>
+        /// Gets a collection of colors as defined in the original second edition of Werner’s Nomenclature of Colours 1821.
+        /// The hex codes were collected and defined by Nicholas Rougeux <see href="https://www.c82.net/werner"/>.
+        /// </summary>
+        public static ReadOnlyMemory<Color> WernerPalette => WernerPaletteLazy.Value;
+
+        private static Color[] CreateWernerPalette() => new[]
```
JimBobSquarePants: Should we cast from `NamedColors` or remove the separate implementation entirely?

src/ImageSharp/Color/Color.cs
antonfirsov: I also keep thinking on this.
We never actually convert a `Color` from a pixel of an `Image<T>`, so there is no practical need for that scenario probably.

src/ImageSharp/Color/Color.NamedColors.cs
antonfirsov: `Rgba32.AliceBlue` is created using `NamedColors<T>`. I'm not sure about the fate of the latter, it might become obsolete with #907.

JimBobSquarePants: Ah you’re exactly right! I’m thinking the wrong way entirely. There’ll be no need for them at all so I think they can be removed (I don't want to obsolete classes at the moment as people ignore warnings). 

I’m gonna cancel my current review and approve it.

src/ImageSharp/Color/Color.Conversions.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            this.data = new Rgba64(vector);
+        }
+
+        /// <summary>
+        /// Converts an <see cref="Rgba64"/> to <see cref="Color"/>.
+        /// </summary>
+        /// <param name="source">The <see cref="Rgba64"/>.</param>
+        /// <returns>The <see cref="Color"/>.</returns>
+        public static implicit operator Color(Rgba64 source) => new Color(source);
```
tocsoft: shouldn't these implicit operators be on the Pixel type instead of Color? there are alot more Pixel types after all.

antonfirsov: Good point! I think I will change this.

src/ImageSharp/Color/Color.cs
antonfirsov: All credit goes to JetBrains! 😄

Issue #10, opened 2016-10-30
Codecs


JimBobSquarePants:
_From @JimBobSquarePants on August 4, 2016 6:15_

This issue serves to collect various issues with the image codecs. Additional issues will be added upon identification.

Png
- [x] Decoder doesn't decode progressive files #379 
- [x]  Encoder does not encode progressive files

Jpeg
- [x] Decoder does not decode CMYK files #444 
- [ ] Encoder does not encode CMYK files
- [ ] Encoder does not encode progressive files

Tiff
- [ ] No Tiff support #464 

_Copied from original issue: JimBobSquarePants/ImageProcessor#446_

PR #909, opened 2019-05-15
Not using quantization for Grey8 images when encoding 8Bit Bitmaps

brianpopow:
### Prerequisites

### Description

The title already says it: When its an Gray8 image, color quantization is actually not needed.


brianpopow:
I would like to add two additional test cases to test the quantization of a 32 bit color image to a 8 bit color image [see:#10](https://github.com/SixLabors/Imagesharp.Tests.Images/pull/10). This may require again comparing with a tolerance, because quantization could produce different results on .net462 vs netcoreapp2.1, but i think thats still better than not testing it.

Or is it maybe enough that the quantization processors already are tested?

JimBobSquarePants:
@brianpopow I've merged those images for you 👍

antonfirsov:
@brianpopow `0.6686%` difference is being reported on 32bit .NET 4.7. Can you reproduce this locally by enforcing 32 bit execution?

brianpopow:
@antonfirsov yes this time i can actually reproduce it locally. I will try to see if i can find out why this is happening.

brianpopow:
@antonfirsov it is, what you already said previously, the quantization produces slightly different results on net472 than on netcore. It looks to me as if the quantized colors are in many case  (not all) identical in net472 and netcore, but they appear in a slightly different position in the palette. For example:

```
net472 : idx: 112, R: 237, G: 144, B: 144
netcore: idx: 112, R: 128, G: 55, B: 128
---------------------------------------------
net472 : idx: 113, R: 128, G: 55, B: 128
netcore: idx: 113, R: 237, G: 144, B: 144
---------------------------------------------
net472 : idx: 115, R: 10, G: 99, B: 99
netcore: idx: 115, R: 221, G: 99, B: 99
---------------------------------------------
net472 : idx: 116, R: 221, G: 99, B: 99
netcore: idx: 116, R: 10, G: 99, B: 99
---------------------------------------------
```

Here is a zip file containing the quantized palette of net472 and netcore21. Also a comparison of which pixels differ. 
[quantize_issue_net472.zip] https://github.com/SixLabors/ImageSharp/files/3194282/quantize_issue_net472.zip)

I will increase the tolerance of the bitmap color encoding test further.

JimBobSquarePants:
There’s a lot of floating point calculation going on in Wu which will be causing the issue. If the output looks ok I’m good with dropping the tests.

antonfirsov:
My tecnhique is to check `TestEnvironment.Is32BitProcess` at the beginning of the test. This way we can keep the coverage for conforming platforms.

brianpopow:
@antonfirsov ok that makes sense, i will do that.

src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
```diff
@@ -340,5 +363,51 @@ private void Write8Bit<TPixel>(Stream stream, ImageFrame<TPixel> image)
...
+
+            stream.Write(colorPalette);
+
+            using (IMemoryOwner<byte> rowSpanBuffer = this.memoryAllocator.AllocateManagedByteBuffer(image.Width, AllocationOptions.Clean))
+            {
+                Span<byte> outputPixelRow = rowSpanBuffer.GetSpan();
+                for (int y = image.Height - 1; y >= 0; y--)
+                {
+                    Span<TPixel> inputPixelRow = image.GetPixelRowSpan(y);
+                    PixelOperations<TPixel>.Instance.ToGray8Bytes(
```
antonfirsov: Can't we replace this with `inputPixelRow.AsBytes().CopyTo(outputPixelRow)` here? Seems more explanatory about our intention.

brianpopow: I can not find AsBytes(), is that an extension method im missing?

antonfirsov: Apologies, it's `MemoryMarshal.AsBytes(spanOfThings)` 😄

brianpopow:
@antonfirsov i was unsure about DrawImageTests.cs while resolving the merge conflict: should this tests be removed or should we keep it?

src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
brianpopow: i have changed that, please check it again.

antonfirsov:
DrawImageTests should be kept in it's original form, strange you had a conflict here. Can you easily revert all changes on that class? Or just replace the code in your file with the code on current master?

src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
antonfirsov: Looks very good, thanks!

brianpopow:
after i integrated the master branch the DrawImageTests.cs was named: DrawImageTest.cs (without an s), so the tests failed because the reference images could not be found. I just changed that to fix the failing tests.
I will take the code from master

antonfirsov:
All good now! Thanks a lot for taking the time, and sorry for the conflicts!

Issue #867, opened 2019-03-29
Expand the build infrastructure support


tannergooding:
The build infrastructure for ImageSharp is rather limited today and doesn't always follow best practices.

It would be ideal if this could be updated to have cross-platform build scripts, rely on things like Directory.Build.props/targets for centralizing build metadata, and generally be updated to do things like work with long paths, with paths that have spaces in the name, and where the working directory is not the repo root.

tannergooding:
@JimBobSquarePants, I know I had discussed this with you briefly at the MVP summit; but if I were to provide some fixes here; would you prefer it incrementally or in one big PR?

tannergooding:
Would you also prefer it in a PR or with a layout of changes discussed here first?

JimBobSquarePants:
Hey @tannergooding 

Yeah this would be great! Looking forward to seeing what will change.

I think the best bet would be a layout of changes here and everything in a big fat PR. That way we can avoid any interim complexity in the build process.

tannergooding:
I think there a couple potential options here:
1. Take a dependency on a "build toolchain" like `dotnet/arcade`. This is what Roslyn, CoreFX, and other repositories are currently using. CoreCLR and other dotnet repositories are planning on moving to it as well.
2. You could go with a smaller more customized approach and rewrite your own infrastructure to how you need.

The process overall would be fairly similar for both, mostly just differing in whether you have code props/targets locally or if you are directing those same props/targets to the `dotnet/arcade` NuGet package.

The primary recommendations I have are:
1. Use `Directory.Build.props` and `Directory.Build.targets` to share common project properties/targets/etc. This also makes it easy to bring down shared toolset packages or analyzers.
2. Rewrite the build scripts to have both `sh` and `cmd` versions (the `cmd` version forwards to `ps1`). This also needs to handle common issues like paths that have a space in their name, pulling down build dependencies, or triggering the scripts from any folder (not just the repo root). It should also provide a consistent interface across all platforms/architectures (so you can easily move between Linux, Windows, OSX, etc).

I've been going with the second option (customized approach) in my own projects as I find it easier to understand/reason about and find it I can't be caught in surprise by changes that arcade makes for dotnet which might not be applicable to me. The drawback, is I'm responsible for my own infrastructure and bringing functionality on-board as needed.

[...]

PR #920, opened 2019-05-28
Updating the repo to use Directory.Build.props/targets files

tannergooding:
### Prerequisites

### Description

This completes the first half of https://github.com/SixLabors/ImageSharp/issues/867 by moving ImageSharp to use centralized Directory.Build.props and Directory.Build.targets files for managing project and repo-wide configuration settings.

This also sets everything up to support strong-name signing, but is first dependent on https://github.com/SixLabors/Standards/pull/1 being merged and then the various dependencies being updated to consume that strong name key.

After this is merged, work can begin on a follow up PR which updates the build scripts for the repository.


.editorconfig
```diff
@@ -1,20 +1,371 @@
-# top-most EditorConfig file
+###############################################################################
```
tannergooding: Configuration options from the previous are kept and the remaining defaults are explicitly listed. This ensures that any user-specific VS configuration doesn't take precedence instead.

.editorconfig
```diff
@@ -1,20 +1,371 @@
...
+# Set default behavior to:
+#   a UTF-8 encoding,
+#   Unix-style line endings,
+#   a newline ending the file,
+#   4 space indentation, and
+#   trimming of trailing whitespace
+###############################################################################
+[*]
+charset = utf-8
+end_of_line = lf
```
tannergooding: The `end_of_line` default was changed to `lf` since that is what git will check it in as (all platforms/by default) and is what git will check it out as on Unix. It makes for better cross-platform consistency.

.editorconfig
```diff
@@ -1,20 +1,371 @@
...
+#   indent case contents,
+#   not indent case contents when block,
+#   indent labels one less than the current, and
+#   indent switch labels
+###############################################################################
+[*.cs]
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = false
```
tannergooding: This is the only new option that differs from the VS default. The VS default is `true` for compatibility reasons. This option is used in combination with `csharp_indent_case_contents` and you have effectively the following combinations:

(VS Default) `csharp_indent_case_contents=true`, `csharp_indent_case_contents_when_block=true`:
```cs
switch (value)
{
    case 1:
        break;

    case 2:
        {
            break;
        }
}
```

`csharp_indent_case_contents=false`, `csharp_indent_case_contents_when_block=true`:
```cs
switch (value)
{
    case 1:
    break;

    case 2:
        {
            break;
        }
}
```

`csharp_indent_case_contents=false`, `csharp_indent_case_contents_when_block=false`:
```cs
switch (value)
{
    case 1:
    break;

    case 2:
    {
        break;
    }
}
```

(The one I chose) `csharp_indent_case_contents=true`, `csharp_indent_case_contents_when_block=false`:
```cs
switch (value)
{
    case 1:
        break;

    case 2:
    {
        break;
    }
}
```

Directory.Build.props
```diff
@@ -0,0 +1,102 @@
...
+
+  <!-- Default settings that explicitly differ from the Sdk.props defaults  -->
+  <PropertyGroup>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <BaseIntermediateOutputPath>$(BaseArtifactsPath)obj/$(BaseArtifactsPathSuffix)/</BaseIntermediateOutputPath>
+    <DebugType>portable</DebugType>
+    <DebugType Condition="'$(codecov)' != ''">full</DebugType>
+    <NullableContextOptions>disable</NullableContextOptions>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <SignAssembly>false</SignAssembly>
```
tannergooding: Once strong-name signing is enabled, it should just involve:
* Updating the `standards` submodule
* Updating the SixLabors dependencies to the strong-name signed versions
* Flipping this to `true`
* Adding the `PublicKey` metdata to the `InternalsVisibleTo` items

Directory.Build.props
```diff
@@ -0,0 +1,102 @@
...
+      https://www.myget.org/F/sixlabors/api/v3/index.json;
+      https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
+      https://api.nuget.org/v3/index.json;
+    </RestoreSources>
+    <SixLaborsPublicKey>002400000c8000009400000006020000002400005253413100040000010001000147e6fe6766715eec6cfed61f1e7dcdbf69748a3e355c67e9d8dfd953acab1d5e012ba34b23308166fdc61ee1d0390d5f36d814a6091dd4b5ed9eda5a26afced924c683b4bfb4b3d64b0586a57eff9f02b1f84e3cb0ddd518bd1697f2c84dcbb97eb8bb5c7801be12112ed0ec86db934b0e9a5171e6bb1384b6d2f7d54dfa97</SixLaborsPublicKey>
+    <UseSharedCompilation>true</UseSharedCompilation>
+  </PropertyGroup>
+
+  <!-- Package versions for package references across all projects -->
+  <PropertyGroup>
```
tannergooding: Package versions are all listed/managed here. VS won't update these currently (but likely will eventually).

This allows you to ensure that you can upgrade all projects at the same time, easily.

Directory.Build.props
```diff
@@ -0,0 +1,102 @@
...
+    <SystemThreadingTasksParallelPackageVersion>4.3.0</SystemThreadingTasksParallelPackageVersion>
+    <SystemValueTuplePackageVersion>4.5.0</SystemValueTuplePackageVersion>
+    <XUnitPackageVersion>2.3.1</XUnitPackageVersion>
+    <XUnitRunnerConsolePackageVersion>2.3.1</XUnitRunnerConsolePackageVersion>
+    <XUnitRunnerVisualStudioPackageVersion>2.3.1</XUnitRunnerVisualStudioPackageVersion>
+  </PropertyGroup>
+
+  <!-- Package references which are consumed by all projects -->
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="$(MicrosoftNetCompilersToolsetPackageVersion)" IsImplicitlyDefined="true" />
```
tannergooding: This is the Roslyn compiler toolset and ensures all machines build with the same compiler version, regardless of what is on the box.

`IsImplicitlyDefined=true` just means users can't remove this from the list of NuGet packages in the UX.

NuGet.config
```diff
@@ -1,9 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<configuration>
```
tannergooding: All of these are managed via the `RestoreSources` property in the `Directory.Build.props`

src/Directory.Build.props
```diff
@@ -0,0 +1,40 @@
...
+  <PropertyGroup>
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)..\standards\SixLabors.ruleset</CodeAnalysisRuleSet>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <AdditionalFiles Include="$(MSBuildThisFileDirectory)..\standards\stylecop.json" />
+  </ItemGroup>
+
+  <ItemGroup>
```
tannergooding: The `Directory.Build.targets` allows you to manage `InternalsVisibleTo` via these item groups.

src/Directory.Build.targets
```diff
@@ -0,0 +1,55 @@
...
+
+  <Import Project="$(MSBuildThisFileDirectory)..\Directory.Build.targets" />
+
+  <PropertyGroup>
+    <GeneratedInternalsVisibleToFile Condition="'$(GeneratedInternalsVisibleToFile)' == ''">$(IntermediateOutputPath)$(MSBuildProjectName).InternalsVisibleTo$(DefaultLanguageSourceExtension)</GeneratedInternalsVisibleToFile>
+  </PropertyGroup>
+
+  <ItemDefinitionGroup>
+    <InternalsVisibleTo>
+      <Visible>false</Visible>
```
tannergooding: This just keeps the `InternalsVisibleTo` from showing up in the UX as "files". The `ItemDefinitionGroup` allows you to add this metadata to all `InternalsVisibleTo` items centrally.

src/Directory.Build.targets
```diff
@@ -0,0 +1,55 @@
...
+    </InternalsVisibleTo>
+  </ItemDefinitionGroup>
+
+  <Target Name="GenerateInternalsVisibleTo"
```
tannergooding: A simple target to dynamically create the `InternalsVisibleTo` file, similarly to how `AssemblyInfo` is generated in SDK style projects.

src/ImageSharp.Drawing/ImageSharp.Drawing.csproj
```diff
@@ -1,55 +1,23 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
+    <AssemblyName>SixLabors.ImageSharp.Drawing</AssemblyName>
     <AssemblyTitle>SixLabors.ImageSharp.Drawing</AssemblyTitle>
-    <Authors>SixLabors and contributors</Authors>
```
tannergooding: All of this removed stuff is now centrally managed.

src/Shared/AssemblyInfo.Common.cs
```diff
@@ -1,15 +0,0 @@
...
-using System.Runtime.CompilerServices;
-
-// Ensure the internals can be built and tested.
-[assembly: InternalsVisibleTo("SixLabors.ImageSharp.Drawing")]
-[assembly: InternalsVisibleTo("ImageSharp.Benchmarks")]
-[assembly: InternalsVisibleTo("SixLabors.ImageSharp.Tests")]
-[assembly: InternalsVisibleTo("SixLabors.ImageSharp.Sandbox46")]
-
-[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
-[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKeyToken=null")]
```
tannergooding: Need to have an `InternalsVisibleTo` for this, will figure out how to get that working...

tannergooding: Should be fixed.

src/ImageSharp/ImageSharp.csproj
```diff
@@ -203,6 +161,87 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
+    <InternalsVisibleTo Include="SixLabors.ImageSharp.Drawing" />
   </ItemGroup>
-</Project>
\ No newline at end of file
+
+  <ItemGroup>
```
tannergooding: Will fix this, mismerge made this think it is new entries....

src/Shared/AssemblyInfo.Common.cs
```diff
@@ -1,15 +0,0 @@
-﻿// Copyright (c) Six Labors and contributors.
-// Licensed under the Apache License, Version 2.0.
-
-using System.Reflection;
-using System.Resources;
-using System.Runtime.CompilerServices;
-
-// Ensure the internals can be built and tested.
-[assembly: InternalsVisibleTo("SixLabors.ImageSharp.Drawing")]
```
tannergooding: These are now project metadata in the form of `InternalsVisibleTo` items

tests/ImageSharp.Benchmarks/ImageSharp.Benchmarks.csproj
```diff
@@ -1,29 +1,37 @@
...
     <Compile Include="..\ImageSharp.Tests\TestImages.cs" Link="Tests\TestImages.cs" />
     <Compile Include="..\ImageSharp.Tests\TestUtilities\TestEnvironment.cs" Link="Tests\TestEnvironment.cs" />
   </ItemGroup>
-  <ItemGroup>    
-    <PackageReference Include="BenchmarkDotNet" Version="0.11.5" />
-    <PackageReference Include="Colourful" Version="2.0.2" />
-    <PackageReference Include="SixLabors.Shapes.Text" Version="1.0.0-beta0007" />
-    <PackageReference Include="System.Drawing.Common" Version="4.5.1" />
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
```
tannergooding: This appeared to be the only think preventing `TargetFrameworks` from working correctly...

tests/ImageSharp.Sandbox46/ImageSharp.Sandbox46.csproj
```diff
@@ -1,30 +1,28 @@
...
   </PropertyGroup>
+
   <ItemGroup>
-    <ProjectReference Include="..\..\src\ImageSharp.Drawing\ImageSharp.Drawing.csproj" />
-    <ProjectReference Include="..\..\src\ImageSharp\ImageSharp.csproj" />
+    <Compile Include="..\ImageSharp.Tests\**\*.cs" Link="Tests\%(RecursiveDir)%(Filename)%(Extension)" />
   </ItemGroup>
+
   <ItemGroup>
-    <PackageReference Include="Magick.NET-Q16-AnyCPU" Version="7.9.0.1" />
```
tannergooding: This was bumped to 7.12.0 as part of centralizing the package versions.

src/ImageSharp/ImageSharp.csproj
tannergooding: Fixed.

tannergooding:
I think the codecov "regression" may be due to a dead-code optimization done by the compiler as the code is currently annotated and is indeed not used:
```cs
// TODO: This is not used?
byte mb = (byte)((minor << 4) | bugfix);
```

tannergooding:
I think this is all good now. CC. @JimBobSquarePants, @antonfirsov

.editorconfig
JimBobSquarePants: I wasn't aware this was still in the repo. Ideally we would use the one from Standards

https://github.com/SixLabors/Standards/blob/8b085c0ec4fb64797b9965741f7138f8f66a6b44/.editorconfig

.editorconfig
JimBobSquarePants: You chose..... Wisely.

.editorconfig
JimBobSquarePants: Super happy to update the Standards one with these defaults.

Directory.Build.props
JimBobSquarePants: It's a damn shame the tooling is behind here. What happens when you update a package through the manager?

Directory.Build.props
JimBobSquarePants: TIL

ImageSharp.sln
```diff
@@ -1,47 +1,358 @@
...
+		build.cmd = build.cmd
+		build.ps1 = build.ps1
 		codecov.yml = codecov.yml
 		CodeCoverage.runsettings = CodeCoverage.runsettings
-		.github\CONTRIBUTING.md = .github\CONTRIBUTING.md
-		.github\ISSUE_TEMPLATE\feature-request.md = .github\ISSUE_TEMPLATE\feature-request.md
-		features.md = features.md
+		Directory.Build.props = Directory.Build.props
+		Directory.Build.targets = Directory.Build.targets
+		ImageSharp.ruleset = ImageSharp.ruleset
```
JimBobSquarePants: We shouldn't be using this old ruleset. I thought it was deleted but somehow not.

ImageSharp.sln
```diff
@@ -1,47 +1,358 @@
...
+		tests\Images\Input\Gif\cheers.gif = tests\Images\Input\Gif\cheers.gif
+		tests\Images\Input\Gif\giphy.gif = tests\Images\Input\Gif\giphy.gif
+		tests\Images\Input\Gif\kumin.gif = tests\Images\Input\Gif\kumin.gif
+		tests\Images\Input\Gif\leo.gif = tests\Images\Input\Gif\leo.gif
+		tests\Images\Input\Gif\rings.gif = tests\Images\Input\Gif\rings.gif
+		tests\Images\Input\Gif\trans.gif = tests\Images\Input\Gif\trans.gif
+	EndProjectSection
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "issues", "issues", "{BF8DFDC1-CEE5-4A37-B216-D3085360C776}"
+	ProjectSection(SolutionItems) = preProject
```
JimBobSquarePants: What's all this stuff? Feels like a relic from the past.

tests/ImageSharp.Benchmarks/ImageSharp.Benchmarks.csproj
JimBobSquarePants: What's happening?

.editorconfig
tannergooding: This needs to live in the appropriate repo root either way, as it is hierarchical in nature and wouldn't be impactful from the standards directory.

Directory.Build.props
tannergooding: It will change the version in the project file directly.

tests/ImageSharp.Benchmarks/ImageSharp.Benchmarks.csproj
tannergooding: In master right now, this project is only targeting a single tfm with `TargetFrameworks` being commented out.

It looks like the double inclusion of an entry point was what was preventing the other teams from working and that filtering out Program.cs for the relevant tfms allows everything to work.

I explicitly called it out in case there was some other reason this should stay targeting a single tfm.

ImageSharp.sln
tannergooding: I can remove it as part of this PR.

ImageSharp.sln
tannergooding: These are all files that you currently have checked directly into the repository, namely under `tests\Images`.

.editorconfig
JimBobSquarePants: I think we can add it from anywhere as long as it's added to the solution and marked as root.

https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019#add-and-remove-editorconfig-files

ImageSharp.sln
JimBobSquarePants: Can we wildcard this?

.editorconfig
tannergooding: `editorconfig` isn't some standard that Microsoft/Visual Studio created, it is an external standard that we adopted because of its wide-spread use.

The way it is meant to be used (and the way it works) is described both here: https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019#file-hierarchy-and-precedence and at the source of truth here: https://editorconfig.org/#file-location.

Things won't function (in VS and in other editors) if it isn't located in the current directory or a parent of the current directory.

ImageSharp.sln
tannergooding: I don't believe solution files support that, and from a quick test locally it doesn't work.

I also don't feel particularly strongly about this being in here, I've just found that having all files accessible from the solution is convenient.

.editorconfig
Paxxi: Won't this make a mess of files if a user has set autocrlf=true in their git global settings?
Might want to add a gitattributes file to make sure they're checked out as lf https://www.git-scm.com/docs/gitattributes

tannergooding: Good point. I hadn't looked at the `.gitattributes` file as part of this PR, will update.

ImageSharp.sln
tannergooding: Removed.

Directory.Build.props
tannergooding: Updated to use `PackageReference Update="package-name" Version=""` rather than the `*PackageVersion` properties, as suggested on the twitter thread.

It, unfortunately, doesn't work with the UX either (project file will still be directly changed), but it seems a lot "cleaner" way of doing this.

.editorconfig
tannergooding: `.gitattributes` file was updated to explicitly call out `eol=lf` as the default and to explicitly specify various attributes for existing file types to ensure that it is handled correctly.

Paxxi: I think you need to set bat/cmd to always be crlf or they won't run. Not at computer so can't test it, going on memory

tannergooding: bat/cmd definitely work with lf line endings. I have it configured that way in all of my projects and did just test locally (and I don't believe it is Windows 10 specific).

Paxxi: Perfect 😀

.editorconfig
JimBobSquarePants: Man that's such an incredibly frustrating design. I'm going o have to rethink the entire way I enforce standards now. (╯°□°）╯︵ ┻━┻

PR #921, opened 2019-05-30
Fix missing "using PixelFormats" line in Readme example

tonygameplan:
### Prerequisites

- (N/A) I have provided test coverage for my change (where applicable)

### Description

"<Rgba32>" does not appear to be defined without the "using SixLabors.ImageSharp.PixelFormats;" line and causes a "The type or namespace name 'Rgba32' could not be found (are you missing a using directive or an assembly reference?) (CS0246)" error to occur. This change fixes the issue by adding the PixelFormats line to the initial dependencies section.


JimBobSquarePants:
Thanks for this. We actually need to change the sample slightly though since we introduced the new non generic syntax in #904.

The sample should be:

```
using SixLabors.ImageSharp;

// Image.Load(string path) is a shortcut for the default type. 
// Other pixel formats use Image.Load<TPixel>(string path))
{
    image.Mutate(x => x
         .Resize(image.Width / 2, image.Height / 2)
         .Grayscale());
    image.Save("bar.jpg"); // Automatic encoder selected based on extension.
}
```

antonfirsov:
It could be a source of confusion to use the  `Image image = ...` declaration here, since the class is only present in the nightlies at the moment. Maybe we should replace it with `var` for now.

JimBobSquarePants:
@antonfirsov You're making me twitchy reading that 😄 `var` without an explicit declaration on the right hand side. Naughty!

antonfirsov:
> @antonfirsov You're making me twitchy reading that 😄 var without an explicit declaration on the right hand side. Naughty!

Not if you look at it as a sample code for Nightly / RC1 versions 😄 It might be also an option to merge this PR as is, and fix the samples before releasing RC1.

JimBobSquarePants:
Let's do that then. @tonygameplan please ignore all our chatter; I'll get this updated and merged now. Cheers! 👍

PR #927, opened 2019-06-05
Feature: Bitmap RLE undefined pixel handling

brianpopow:
### Prerequisites

### Description

With RLE encoding it is possible to skip over some pixels, which leaves them undefined. 
This occures when *delta*, *End of line* or *End of Bitmap.* is used. All those commands would skip pixels.
See [RLE documentation](https://docs.microsoft.com/en-us/windows/desktop/gdi/bitmap-compression)
For an example see [Bitmap Testsuite](http://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html) q/pal4rletrns.bmp.

There are three possible options to handle this, each equally plausible:
- Undefined pixels get the first color of the palette (this is what ImageSharp currently does)
- Undefined pixels are assign black color (System drawing does that)
- Undefined pixels will be transparent

This PR adds an additional decoder option, so the user can choose how to deal with those skipped pixels.

Additional changes in this PR:
- Allow decoder to decode bitmaps with oversized palette.
- Refactor unit tests in smaller tests units, instead of one big "test all" testcase.


tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
```diff
@@ -60,6 +60,158 @@ public void BmpDecoder_CanDecodeBitfields<TPixel>(TestImageProvider<TPixel> prov
...
+        [Theory]
+        [WithFile(Bit1, PixelTypes.Rgba32)]
+        [WithFile(Bit1Pal1, PixelTypes.Rgba32)]
+        public void BmpDecoder_CanDecode_1Bit<TPixel>(TestImageProvider<TPixel> provider)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            using (Image<TPixel> image = provider.GetImage(new BmpDecoder()))
+            {
+                image.DebugSave(provider);
+                image.CompareToOriginal(provider, new SystemDrawingReferenceDecoder());
```
antonfirsov: I wonder if we understand on what images does ImageMagick fail? Any rule here?

tests/ImageSharp.Tests/TestImages.cs
```diff
@@ -278,25 +289,32 @@ public static class Bmp
...
+
+            public static readonly string[] Benchmark
+                = {
+                          Car,
+                          F,
+                          NegHeight,
+                          CoreHeader,
+                          V5Header,
+                          RLE4,
+                          RLE8,
```
JimBobSquarePants: Make sure any new images are added to the solution folder for the project also

JimBobSquarePants:
@brianpopow Hey mate, just added a few comments to the PR. 

Looking good but just had some questions 👍

tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
brianpopow: yes, it fails with 4-bit images

brianpopow: just to clarify: i decided to execute the RLE tests for 8 bit only on windows, because System.drawing choses a different approach than ImageMagick to deal with the skipped pixels (System.Drawing == Black, Magick == First color of palette). The 8 bit tests were failing on linux.

JimBobSquarePants: Let's put some comments in the source to explain that then please.

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -291,7 +304,29 @@ private void ReadRle<TPixel>(BmpCompression compression, Buffer2D<TPixel> pixels
 
                     for (int x = 0; x < width; x++)
                     {
-                        color.FromBgr24(Unsafe.As<byte, Bgr24>(ref colors[bufferRow[x] * 4]));
+                        byte colorIdx = bufferRow[x];
+                        if (undefinedPixels[x, y])
+                        {
+                            switch (this.options.RleSkippedPixelHandling)
```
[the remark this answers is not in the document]
brianpopow: I think i have an idea how this could be done. I could keep track which rows have skipped pixels and then branching per row should be possible.

tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
brianpopow: i have added comments for the 4 bit tests. I have added an additional test for RLE 8 Bit with magick reference decoder

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -291,7 +304,29 @@ private void ReadRle<TPixel>(BmpCompression compression, Buffer2D<TPixel> pixels
...
+                            {
+                                case RleSkippedPixelHandling.FirstColorOfPalette:
+                                    color.FromBgr24(Unsafe.As<byte, Bgr24>(ref colors[colorIdx * 4]));
+                                    break;
+                                case RleSkippedPixelHandling.Transparent:
+                                    color.FromVector4(new Vector4(0.0f, 0.0f, 0.0f, 0.0f));
```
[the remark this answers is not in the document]
brianpopow: i have changed that

PR #930, opened 2019-06-10
Bitmap decoder now can decode bitmap arrays

brianpopow:
### Prerequisites

### Description

This PR adds to the bitmap decoder the ability to decode Bitmap Arrays. Bitmap array files are typically used to hold multiple resolution bitmaps of the same image. The first image in the array will be decoded.
For more info see: [Bitmap Array](http://www.fileformat.info/format/bmp/egff.htm)


JimBobSquarePants:
@brianpopow Wow. You've been busy! 

Is there an application that supports this variant that you know of I can use to compare the output with?

brianpopow:
@JimBobSquarePants yes i had some spare time and i thought i could follow my quest in supporting all bitmap formats :smile:. We are getting close. 

The only program i know is gimp which can decode those images.

JimBobSquarePants:
@brianpopow  I checked everything against Gimp and it all looks great! Thanks again for this! 👍

brianpopow:
@JimBobSquarePants you are welcome!

I think i understand now how ColorIcon, ColorPointer, Icon and Pointer bitmaps can be decoded as well. They are a bit weird though, especially ColorIcon, ColorPointer. They contain two Bitmap info headers and therefore two bitmaps which need to be combined into one.

I thought it would be nice to support all valid bitmaps, even if they are ancient and not common anymore. Before i continue with that i wanted to ask you what you think about this. Does it make sense to support those old formats? I do not want to force something on you, which you may not want. Its totally fine with me, if you say that this old stuff is too weird, to be supported.

JimBobSquarePants:
If you're happy to do the work then I'm happy to accept the enhancements. 😄

PR #932, opened 2019-06-13
Throw UnkownFormatException on Image.Load 

fredeil:
### Description

Image.Load throws an `ImageFormatException` instead of `NotSupportedException` when no available decoders are available for the source/image.

The exception documentation is still annotated with `throw NotSupportedException` since the `WithSeekableStream` throws that still.

Did not add any unit tests for this since there were no Assert.Throws for image load. Let me know if you prefer this to be added 👍


JimBobSquarePants:
@antonfirsov I like the idea of `UnknownImageFormatException`

@fredeil if you could create a dummy stream filled with junk and create some tests around that it would be really great!

fredeil:
Added `UnkownFormatException` and added unit tests to the two "base" Image.Load methods

PR #941, opened 2019-06-23
Introduce non-generic ImageFrameCollection

antonfirsov:
### Prerequisites

## Description
Next point of the #907 checklist.
After merging #904, `image.Frames` disappeared from the "main" `Image` class which is kind-of unexpected. This PR fixes it by introducing non-generic base classes `ImageFrame` and `ImageFrameCollection`. 

The non generic frame collection has the same methods as `ImageFrameCollection<TPixel>`, which are hidden using the `new` keyword in the derived implementation class. To manage the name collision, `protected virtual` implementation methods have been introduced.

#### Converting ImageFrame-s
It's allowed now to feed `ImageFrameCollection` using images of different pixel type. (I had to refactor `PixelOperations<T1>.To<T2>(...)` to make it a reliable and efficient operation). However, something must be wrong with [my test code](https://github.com/SixLabors/ImageSharp/blob/af/nongeneric-framecollection/tests/ImageSharp.Tests/Image/ImageFrameCollectionTests.NonGeneric.cs#L272), because it's producing strange frames:
image

@JimBobSquarePants can you help me with this?


.gitattributes
```diff
@@ -50,7 +50,7 @@
 *.svg               text        eol=lf
 *.targets           text        eol=lf
 *.tt                text        eol=lf
-*.ttinclude         text        eol=lf
+*.ttinclude         text        eol=crlf
```
antonfirsov: for some reason `ttinclude` files stop working with `lf` file endings.

JimBobSquarePants:
@antonfirsov I'll have a look tonight. Hopefully shouldn't be too difficult to figure out.

JimBobSquarePants:
@antonfirsov 

Your bug is here.
https://github.com/SixLabors/ImageSharp/blob/f3c31f56f43cc4fc2e5afb188c0ec7baffaf6dca/src/ImageSharp/ImageFrameCollection%7BTPixel%7D.cs#L264

When the source and result are not compatible you're calling `CreateFrame` which in turn uses the `ImageFrameCollection` that creates a new `ImageFrameMetadata` rather than cloning.

There's a couple of ways to fix that. Either, add an internal setter to `Metadata` or update the internal method calls to use the correct constructor.  

However, there's another issue with the code I can't figure out at the moment. The root frame is the input image size following any transforms despite the Image itself reporting the correct size.

antonfirsov:
@JimBobSquarePants thanks for your help! I think I fixed all the bugs, so the PR is ready for review.

.gitattributes
JimBobSquarePants: Yeah, did some reading. No workaround I can find.

Issue #946, opened 2019-07-14
Maintain Xamarin iOS support by keeping the library "AOT-proof"


antonfirsov:
## Description
Xamarin iOS users need to be able to "seed" open-generic library API-s with the help of [AotCompilerTools](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Advanced/AotCompilerTools.cs), otherwise it will lead to [runtime exceptions](https://github.com/SixLabors/ImageSharp/issues/776#issuecomment-510424553) on that platform.

**Related:**  #767, #785, #776, #800

## Problems
Currently we are struggling to maintain this feature because of the following reasons:
- There is no one in the core team having experience with Xamarin iOS or any other AOT platform.
- We don't have any regression tests for this, we never know when a change breaks the "AOT-proofness" of an existing API
- We have introduced a massive change with #904, which will most likely break AOT & Xamarin iOS for many existing API-s in RC1. I have defined a point in epic #907 to investigate this, however in practice it turned out that I hardly have the resources to do so during this summer. (Neither I think @JimBobSquarePants can make it.) Therefore it will most likely not happen for RC1.

## Conclusion
With our current prioritites and resources, the only way to maintain this feature is community support. If you can provide any help (add missing AOT seeds, or fix regressions), feel free to open a Pull Request!

#### Long term solution
If someone could define a mono AOT test configuration for appveyor, that would be ace.

/CC @vicfergar  @davilovick  @gameleon-dev @Lapinou42 @dmanning23 @swoolcock

**EDIT**
**PS:** Could someone do a favor and chack the latest MyGet dev builds from on Xamarin iOS? (JPEG, PNG, Resize, Quantization etc.)

dmanning23:
Cool, super appreciate the work yall have been putting into ImageSharp. I have some downtime this week and can take a look at the iOS build.

JimBobSquarePants:
@dmanning23 That would be absolutely amazing if you can, thank you very much!

dmanning23:
Opened a pull-request with a few changes to fix ImageSharp on iOS. I've verified with these changes that it works for jpg, gif, and image resizing.
I also rolled up some changes I've had in mind for the https://github.com/SixLabors/ImageSharp/issues/800 ticket. I've verified that the AoTCompilerTools.Seed methods never actually have to be called, that code just needs to be present for the AoT compiler to pick it up.

[...]

Issue #800, opened 2018-12-27
Refactor AOTCompilerTools to hide implementation


dmanning23:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
This is related to a lot of the AOT -> JIT bugs that have been written lately... I've been looking into iOS AOT errors more, and found some documentation of how to solve them (in the Unity docs of all places):

https://docs.unity3d.com/Manual/ScriptingRestrictions.html

ImageSharp has the `AOTCompilerTools.Seed<TPixel>()` method that can be called to pre-seed the AOT compiler, however that method never actually has to be called. I've verified that a call to that method **only has to exist** in the codebase with the correct generic type, and the AOT compiler will pick it up:

```
private static void NeverCallThisMethod()
        {
            AotCompilerTools.Seed<Rgba32>();
            throw new Exception("Don't call AotCompilerTools.NeverCallThisMethod(), it only exists to preseed the AOT compiler.");
        }
```

^ This totally works, and wouldn't force ImageSharp users to use troubleshoot confusing AOT compiler errors. I'd be happy to clean up this code and create a PR for it.

### System Configuration

- ImageSharp version: This is on the tip of master
- Other ImageSharp packages and versions:
- Environment (Operating system, version and so on): iOS
- .NET Framework version:
- Additional information:

antonfirsov:
@dmanning23 thanks for investigating this!

I think we will drop those methods from the public API before 1.0. 

/cc @swoolcock

PR #949, opened 2019-07-19
#946: AoT compiler fixes and cleaned up AoTCompilerTools

dmanning23:
### Prerequisites

### Description

These changes update the AoTCompilerTools to fix a few issues with the iOS build.
I've verified that jpeg, gif, and resizing work on iOS with these changes.

- Added several new virtual generic methods to the AoT compiler tools.
- Hid the Seed methods. They never actually need to be called, the code just needs to present at compile time. This will be a breaking change for anyone currently using ImageSharp on iOS, but it will be a much better experience in the long run. This is the change proposed in https://github.com/SixLabors/ImageSharp/issues/800
- Removed AoT unit tests. This code should never be called!


src/ImageSharp/Advanced/AotCompilerTools.cs
```diff
@@ -146,5 +161,28 @@ private static void AotCodec<TPixel>(IImageDecoder decoder, IImageEncoder encode
...
+        }
+
+        /// <summary>
+        /// This method pre-seeds the ResizeProcessor for the AoT compiler on iOS.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        private static void AotCompileResizeOperations<TPixel>()
+            where TPixel : struct, IPixel<TPixel>
+        {
+            var resizeProcessor = new ResizeProcessor(new ResizeOptions(), new Size());
```
antonfirsov: `new ResizeProcessor(new ResizeOptions(), default)` should do the job.

antonfirsov:
I'm merging this, we can do the rest of #946 in follow-up PR-s.

JimBobSquarePants:
Thanks so much for your help here @dmanning23 !!

PR #951, opened 2019-07-24
Add support for read and write tEXt, iTXt and zTXt chunks

brianpopow:
### Prerequisites

### Description

This PR adds to the PNG encoder/decoder the ability to read different text chunks. It adds support for:

- **tEXt**: Contains a keyword and a text string, both with latin encoding.
- **zTXt**: Same as **tEXt**, but compressed.
- **iTXt**: For international strings where the text string is UTF-8 encoded. Can optionally be compressed.
**note: for iTXt**: This chunk can optionally contain a language tag and a translated keyword. This is ignored in the current implementation.

Also this PR adds another option to the encoder: CompressTextThreshold. If a string exceeds this length, the text will be compressed.
The option from the textEncoding Option from the decoder was removed, because should not be a concern of the user. This is dictated by the specification: https://www.w3.org/TR/PNG/#11zTXt
As described above tEXt should be Latin-1 character set [ISO-8859-1]. iTXt should use UTF-8.


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -876,12 +883,151 @@ private void ReadTextChunk(ImageMetadata metadata, ReadOnlySpan<byte> data)
 
             int zeroIndex = data.IndexOf((byte)0);
 
-            string name = this.textEncoding.GetString(data.Slice(0, zeroIndex));
-            string value = this.textEncoding.GetString(data.Slice(zeroIndex + 1));
+            // Keywords are restricted to 1 to 79 bytes in length.
+            if (zeroIndex <= 0 || zeroIndex > 79)
```
JimBobSquarePants: Maybe move these to consts

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -876,12 +883,151 @@ private void ReadTextChunk(ImageMetadata metadata, ReadOnlySpan<byte> data)
...
+                inflateStream.AllocateNewBytes(compressedData.Length);
+                var uncompressedBytes = new List<byte>();
+                int bytesRead = inflateStream.CompressedStream.Read(this.buffer, 0, this.buffer.Length);
+                while (bytesRead != 0)
+                {
+                    uncompressedBytes.AddRange(this.buffer.AsSpan().Slice(0, bytesRead).ToArray());
+                    bytesRead = inflateStream.CompressedStream.Read(this.buffer, 0, this.buffer.Length);
+                }
+
+                return encoding.GetString(uncompressedBytes.ToArray());
```
JimBobSquarePants: Are there span overloads for this now?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1064,11 +1210,39 @@ private bool TryReadChunkLength(out int result)
...
+        /// </summary>
+        /// <param name="keywordBytes">The keyword bytes.</param>
+        /// <param name="name">The name.</param>
+        /// <returns>True, if the keyword could be read and is valid.</returns>
+        private bool TryReadTextKeyword(ReadOnlySpan<byte> keywordBytes, out string name)
+        {
+            name = string.Empty;
+
+            // Keywords shall contain only printable Latin-1.
+            if (keywordBytes.ToArray().Any(c => !((c >= 32 && c <= 126) || (c >= 161 && c <= 255))))
```
JimBobSquarePants: Let's not use Linq here and avoid the `ToArray()` call.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -172,6 +180,11 @@ internal sealed class PngEncoderCore : IDisposable
         /// </summary>
         private IManagedByteBuffer paeth;
 
+        /// <summary>
+        /// Latin encoding is used for text chunks.
+        /// </summary>
+        private Encoding latinEncoding = Encoding.GetEncoding("ISO-8859-1");
```
JimBobSquarePants: static?

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -738,6 +753,78 @@ private void WriteExifChunk(Stream stream, ImageMetadata meta)
...
+            {
+                bool hasUnicodeCharacters = imageProperty.Value.Any(c => c > MaxLatinCode);
+                if (hasUnicodeCharacters)
+                {
+                    // Write iTXt chunk.
+                    byte[] keywordBytes = this.latinEncoding.GetBytes(imageProperty.Name);
+                    byte[] textBytes = imageProperty.Value.Length > this.compressTextThreshold ? this.GetCompressedTextBytes(Encoding.UTF8.GetBytes(imageProperty.Value)) : Encoding.UTF8.GetBytes(imageProperty.Value);
+
+                    // Note: The optional language tag and the translated keyword will be omitted.
+                    Span<byte> outputBytes = new byte[keywordBytes.Length + textBytes.Length + 5];
```
JimBobSquarePants: This is why I was concerned about using the `ImageProperty` type. I think these properties should all be specific to Png Metadata only. 

In fact, I think we should dump `ImageProperty`!

src/ImageSharp/Formats/Png/PngDecoderCore.cs
brianpopow: yes there is a overload of `encoding.GetString` which takes a span, but i still need to get the byte  s from the list first, so i dont have a span in the first place.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
brianpopow: ok i will create a png specific TextProperty then, with a language tag and a translated keyword.

But i think we can not dump ImageProperty because its used in GifDecoder, too.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -876,12 +883,151 @@ private void ReadTextChunk(ImageMetadata metadata, ReadOnlySpan<byte> data)
...
+            using (var memoryStream = new MemoryStream(compressedData.ToArray()))
+            using (var inflateStream = new ZlibInflateStream(memoryStream, () => 0))
+            {
+                inflateStream.AllocateNewBytes(compressedData.Length);
+                var uncompressedBytes = new List<byte>();
+                int bytesRead = inflateStream.CompressedStream.Read(this.buffer, 0, this.buffer.Length);
+                while (bytesRead != 0)
+                {
+                    uncompressedBytes.AddRange(this.buffer.AsSpan().Slice(0, bytesRead).ToArray());
+                    bytesRead = inflateStream.CompressedStream.Read(this.buffer, 0, this.buffer.Length);
```
brianpopow: @JimBobSquarePants: i was a bit unsure about this part. Im re-using here the buffer from the PngDecoderCore to uncompress, but its only 4 bytes long. Does it make sense here for performance reasons to allocate a larger buffer? Or is it ok like it is?

src/ImageSharp/Formats/Png/PngEncoderCore.cs
JimBobSquarePants: If you can replace the property in the Gif meta with a specific one as part of this PR, id' be happy. It should only be a minor change, but if not, we can wait and do it separately.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: Oh yeah, I see what you mean. I'm not too concerned as I don't think these tags are commonly used. Maybe we can just add a note for now.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
brianpopow: i have added a png specific TextDataProperty and the old `ImageProperty` is now the new Gif specific TextProperty. I dont know about any gif textdata (are there any restrictions?), i assume its similar to the PNG textdata and can hold  something like Author, Comment, Description etc.

brianpopow:
@JimBobSquarePants i think i have convert all your points, let me know if you see something else i should change.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
JimBobSquarePants: So it looks like we've been doing Gif comment blocks incorrectly. We should be checking for sub blocks within the extension block. 

https://tronche.com/computer-graphics/gif/gif89a.html#comment-extension

brianpopow: i have just tried to add a comment to an image with imagemagick and read it with ImageSharp.
I think the decoder handles it correctly:
- Checks first for ExtensionIntroducer
- Then length of the comment
- The actual string
- A final 0 byte to terminate

What i noticed reading the gif specs is, that gif comments does not have a Name for the Propertys like PNG. Maybe we should remove the Name field of the GifTextData then?

JimBobSquarePants: I'm not sure it is right. The code in the master branch appears to treat each comment sub block as a distinct comment which is incorrect and also check against and arbitrary maximum length.

My interpretation of the [spec ](https://www.w3.org/Graphics/GIF/spec-gif89a.txt) section highlighted below is as follows:

- There can be multiple Comments in a gif.
- Each Comment is made up of multiple sub blocks stored in blocks ranging from 1-255 bytes in length.
- There is no maximum size of a Comment

So by my reckoning we should have a `Comments` property within the Gif metadata which is a `List<string>`.

What do you think? Apologies btw for throwing all this at you!

```
24. Comment Extension.

      a. Description. The Comment Extension contains textual information which
      is not part of the actual graphics in the GIF Data Stream. It is suitable
      for including comments about the graphics, credits, descriptions or any
      other type of non-control and non-graphic data.  The Comment Extension
      may be ignored by the decoder, or it may be saved for later processing;
      under no circumstances should a Comment Extension disrupt or interfere
      with the processing of the Data Stream.

      This block is OPTIONAL; any number of them may appear in the Data Stream.

      b. Required Version.  89a.

      c. Syntax.

      7 6 5 4 3 2 1 0        Field Name                    Type
     +---------------+
  0  |               |       Extension Introducer          Byte
     +---------------+
  1  |               |       Comment Label                 Byte
     +---------------+

     +===============+
     |               |
  N  |               |       Comment Data                  Data Sub-blocks
     |               |
     +===============+

     +---------------+
  0  |               |       Block Terminator              Byte
     +---------------+

            i) Extension Introducer - Identifies the beginning of an extension
            block. This field contains the fixed value 0x21.

            ii) Comment Label - Identifies the block as a Comment Extension.
            This field contains the fixed value 0xFE.

            iii) Comment Data - Sequence of sub-blocks, each of size at most
            255 bytes and at least 1 byte, with the size in a byte preceding
            the data.  The end of the sequence is marked by the Block
            Terminator.

            iv) Block Terminator - This zero-length data block marks the end of
            the Comment Extension.

      d. Extensions and Scope. This block does not have scope. This block
      cannot be modified by any extension.

      e. Recommendations.

            i) Data - This block is intended for humans.  It should contain
            text using the 7-bit ASCII character set. This block should
            not be used to store control information for custom processing.

            ii) Position - This block may appear at any point in the Data
            Stream at which a block can begin; however, it is recommended that
            Comment Extensions do not interfere with Control or Data blocks;
            they should be located at the beginning or at the end of the Data
            Stream to the extent possible.
```

brianpopow: I have read the spec again and i think you are right. I believe i can change the gif encoder / decoder to follow the spec.

JimBobSquarePants: You rock! 🤘

brianpopow: ok there is now a first version ready for review for the changes to the gif encoder / decoder for the comments. Let me know what you think.

JimBobSquarePants:
@brianpopow Rather than bother you I've just pushed some cleanup to the code to fix a few issues I spotted.

- Moved format specific properties to their relative metadata classes.
- Introduced constants for the different encoding types to the format constants classes.

I think something is going funky with Github just now as it's not reporting statuses properly but as soon as the build completes, I'm happy to merge it. 

Thanks again for your help, really excellent work! 👍

brianpopow:
> I think something is going funky with the CLA Assistant site just now as it's not reporting statuses properly but as soon as the build completes, I'm happy to merge it.

awesome, thank you!

PR #957, opened 2019-08-07
fix some spelling

SimonCropp:


tests/ImageSharp.Tests/Drawing/DrawPathTests.cs
```diff
@@ -30,17 +30,17 @@ public class DrawPathTests
         public void DrawPath<TPixel>(TestImageProvider<TPixel> provider, string colorName, byte alpha, float thickness)
             where TPixel : struct, IPixel<TPixel>
         {
-            var linerSegemnt = new LinearLineSegment(
+            var linerSegment = new LinearLineSegment(
```
dlemstra: Looks like this should be `linearSegment` instead.

SimonCropp: fixed

PR #959, opened 2019-08-07
fix "as" usage

SimonCropp:
`as` implies that "we expect the instance might not be of that type" and hence there should (almost) always be a null check before using the result of the operation. if no check is done then a null ref might occur.


JimBobSquarePants:
@SimonCropp There's a certain joy in my heart that is brought about when I see someone having a really good dig around the codebase. Let me know what you think, I'm very interested in hearing your feedback. :+1:

SimonCropp:
@JimBobSquarePants 

> Let me know what you think

so far very impressed. as u have seen, i have only been able to find cosmetic issues so far.

PR #963, opened 2019-08-10
remove unused variables and methods

SimonCropp:
not sure about the history here? should any of these remain?


JimBobSquarePants:
> not sure about the history here? should any of these remain?

No, you're correct to remove anything unused. There's been a lot of churn in the repo and it's a complex library so there's definitely been things missed. Your PR also let me spot two areas in the code we could optimize the behavior so I'm very glad you gave it a once-over.

SimonCropp:
@JimBobSquarePants yeah i wasnt sure on some of them. happy for you to take the parts you want and discard the PR

JimBobSquarePants:
It's all good, taking as-is. Thanks! 👍

Issue #928, opened 2019-06-07
FillRegionProcessor.OnFrameApply IndexOutOfRangeException when calling DrawLines with specific parameters


woutware:
### Description
Bug

### Steps to Reproduce
```
using System;
using System.Collections.Generic;
using System.IO;
using System.Linq;
using System.Numerics;
using System.Text;
using SixLabors.Fonts;
using SixLabors.ImageSharp;
using SixLabors.ImageSharp.PixelFormats;
using SixLabors.ImageSharp.Processing;
using SixLabors.ImageSharp.Processing.Drawing;
using SixLabors.ImageSharp.Processing.Overlays;
using SixLabors.ImageSharp.Processing.Text;
using SixLabors.Primitives;
using SixLabors.Shapes;

namespace ConsoleApp1
{
    class Program
    {
        static void Main(string[] args)
        {
            DrawLinesBug(@"D:\support\ImageSharp\DrawLinesBug.png", new RectangleF(0, 0, 2000, 2000));
            
            Console.WriteLine("Hello World!");
            Console.ReadLine();
        }

        public static void DrawLinesBug(string savePath, RectangleF rectText) {
            using (Image<Rgba32> img = new Image<Rgba32>((int)rectText.Width, (int)rectText.Height)) {
                img.Mutate(x => x.Fill(Rgba32.Transparent));

                img.Mutate(ctx => {
                    ctx.DrawLines(
                        Rgba32.Red,
                        0.984252f,
                        new PointF(104.762581f, 1074.99365f),
                        new PointF(104.758667f, 1075.01721f),
                        new PointF(104.757675f, 1075.04114f),
                        new PointF(104.759628f, 1075.065f),
                        new PointF(104.764488f, 1075.08838f),
                        new PointF(104.772186f, 1075.111f),
                        new PointF(104.782608f, 1075.13245f),
                        new PointF(104.782608f, 1075.13245f)
                        );
                    }
                );

                SixLabors.ImageSharp.Formats.Png.PngEncoder pngEncoder = new SixLabors.ImageSharp.Formats.Png.PngEncoder();
                pngEncoder.CompressionLevel = 0;

                FileStream fileStream = new FileStream(savePath, FileMode.OpenOrCreate);
                img.SaveAsPng(fileStream, pngEncoder);
            }
        }
    }
}

```
### System Configuration
Win 10, .NET Core 2.0

- ImageSharp version: 1.0.0-beta0006
- Other ImageSharp packages and versions:
[...]

PR #964, opened 2019-08-10
Fix IndexOutOfRangeException in FillRegionProcessor

101100:
### Prerequisites

### Description

This is a fix for an issue that I have discovered as well as the reported issue #928.  I have added two tests to illustrate the issue that fail on `master` as well as a possible fix.

The issue arises due to this code:

https://github.com/SixLabors/ImageSharp/blob/aac8eae64aab06ecd8aa0210e49fb0ba67e9a4ea/src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor%7BTPixel%7D.cs#L99-L112

If `pointsFound` is odd (such as 3) and equal to the size of the buffer, the loop will go up to `buffer.Length - 1`, but that means that `buffer[point + 1]` will be out of bounds and cause an exception.

My solution was to add a second condition to the loop: `point < buffer.Length - 1`  Unfortunately, I don't know enough about the algorithm to know if this will result in an incorrect result.

### DIFF
--- a/src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor{TPixel}.cs
+++ b/src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor{TPixel}.cs
@@ -105,7 +105,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Drawing
 
                             QuickSort.Sort(buffer.Slice(0, pointsFound));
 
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Thanks! @101100 !!

We'll have a good look at that asap. 👍

101100:
Thanks for the quick merge!  I tested this in my project with the nightly build `1.0.0-dev002846` and it worked without an issue.

PR #971, opened 2019-08-12
remove some redundant variables and type params

SimonCropp:
u may want to squash this one since i decided to revert the sum commit

### DIFF
--- a/tests/ImageSharp.Tests/Formats/Gif/GifEncoderTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Gif/GifEncoderTests.cs
@@ -150,7 +150,7 @@ namespace SixLabors.ImageSharp.Tests.Formats.Gif
                 outStream.Position = 0;
                 var clone = Image.Load<Rgba32>(outStream);
 
-                GifMetadata cloneMetaData = clone.Metadata.GetFormatMetadata<GifMetadata>(GifFormat.Instance);
+                GifMetadata cloneMetaData = clone.Metadata.GetFormatMetadata(GifFormat.Instance);
                 Assert.Equal(metaData.ColorTableMode, cloneMetaData.ColorTableMode);
 
                 // Gifiddle and Cyotek GifInfo say this image has 64 colors.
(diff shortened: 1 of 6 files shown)


tests/ImageSharp.Tests/Formats/Gif/GifEncoderTests.cs
```diff
@@ -150,7 +150,7 @@ public void NonMutatingEncodePreservesPaletteCount()
                 outStream.Position = 0;
                 var clone = Image.Load<Rgba32>(outStream);
 
-                GifMetadata cloneMetaData = clone.Metadata.GetFormatMetadata<GifMetadata>(GifFormat.Instance);
+                GifMetadata cloneMetaData = clone.Metadata.GetFormatMetadata(GifFormat.Instance);
```
dlemstra: Maybe better to use `var cloneMetaData` here instead?

SimonCropp: do u want another pr that makes the use of `var` consistent?

dlemstra: That would be nice, but you can probably sneak this one in?

antonfirsov: I think, we should merge it as-is now to keep the ball rolling.

JimBobSquarePants:
I’m gonna merge this. I expect a lot of test code cleanup over the next few months anyway.

tests/ImageSharp.Tests/Formats/Gif/GifEncoderTests.cs
SimonCropp: @dlemstra @antonfirsov @JimBobSquarePants so re var. whats your preferences? "just var everywhere"?

JimBobSquarePants: The ide should hint via editorconfig. Var only when the type is apparent.

SimonCropp: weird. i dont get a hint?

JimBobSquarePants: Very weird, what’s your dev setup? Any plugins etc?

SimonCropp: I will troubleshoot when i get home, and get back to you

PR #982, opened 2019-08-15
cache max in ConvertToRgba

SimonCropp:
not sure if this is a fast path?


src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromGrayScale.cs
```diff
@@ -19,10 +19,11 @@ public FromGrayscale(int precision)
 
             public override void ConvertToRgba(in ComponentValues values, Span<Vector4> result)
             {
+                var maximum = 1 / this.MaximumValue;
                 var scale = new Vector4(
-                                1 / this.MaximumValue,
-                                1 / this.MaximumValue,
-                                1 / this.MaximumValue,
+                                maximum,
```
dlemstra: Not sure if we still need this on separate lines?

SimonCropp: fixed

PR #984, opened 2019-08-20
Fix adaptive histogram out of bounds bug

brianpopow:
### Prerequisites

### Description

This PR fixes an ArrayOutOfBound Exception issue which occurs in some cases with the adaptive histogram interpolation. It happens because in some cases the tile start position array is calculated one entry too large.

This happens for example when the image width is 990 and height is 750. With 10 tiles this makes a ```tileHeight``` of 99. Half tile width would be 49 then, which leads to a wrong termination condition of the for loop which calculates the start positions.


antonfirsov:
I wonder if we can provide a regression test with input smaller than `990 x 750`?

brianpopow:
> I wonder if we can provide a regression test with input smaller than `990 x 750`?

@antonfirsov  I could make a test which resizes an existing image to some image sizes were i know this issue happens and then run the histogram equalization and assert in the end that no exception is thrown. Would that be sufficient?

An alternative could be that i add an additional image to the ReferenceOutput which has a image size which triggers this issue and make an additional test similar to the already existing histogram tests: [HistogramTests](https://github.com/SixLabors/ImageSharp/blob/master/tests/ImageSharp.Tests/Processing/Normalization/HistogramEqualizationTests.cs)

antonfirsov:
[`[WithBasicTestPatternImages]`](https://github.com/SixLabors/ImageSharp/blob/aac8eae64aab06ecd8aa0210e49fb0ba67e9a4ea/tests/ImageSharp.Tests/Processing/Processors/Transforms/ResizeTests.cs#L85) is what you need in such cases.

antonfirsov:
@brianpopow I think it's also useful to validate the output, not just the fact that exception is not thrown. As a first step, I pushed a change `DebugSave()`-ing the image.

For 110x110 it looks like this, which doesn't seem correct to me:
Issue984_Rgba32_TestPattern110x110

antonfirsov:
I'm merging this as-is, so we can release beta7.

brianpopow:
@antonfirsov i will look into this and provide a better test, as soon as possible.

PR #989, opened 2019-08-25
Implement gradient brush similar to PathGradientBrush (#969)

### LINKED ISSUE #969: Implement equivalent of System.Drawing.Drawing2D.PathGradientBrush
Currently, ImageSharp allows you to define various types of gradient brushes none of them provides an ability to make a gradient that has multiple color stops that directly correspond to user specified x, y coordinates like `System.Drawing.Drawing2D.PathGradientBrush` does.

For example, you may want to fill a triangle with each point corresponding to red, green, blue color respectively which, if I'm not mistaken, is not supported by ImageSharp currently.

As a side note, the Mono's implementation of `PathGradientBrush` is currently broken on non-Windows platforms, so there seems to be no viable pure C# option to achieve this at the moment, which I found when I tried to write a texture generating module for a Godot game.

I tried to implement a custom brush as it seemed not so difficult a task, but I found that it's not possible to do that either due to how relevant APIs are declared as 'internal' (will open a separate issue for this problem).

mysticfall:
###  Prerequisites

### Description

A gradient brush implementation that works similar to System.Drawing.Drawing2D.PathGradientBrush.

This fixes #969, but only convex paths are supported for now.


mysticfall:
To maintainers: I think I need some help about this PR. I wrote test cases for this feature, but I don't know what to do with the text fixture (reference output images).

Should I fork SixLabors/Imagesharp.Tests.Images also and make a separate PR? Or maybe I should just attach those images here (which I'll do it just in case)?

Sorry if it's something obvious, but it's the first time I've ever made a PR that spans across multiple Git submodules.

As to the code itself, I think it works as intended for simple convex shapes but it probably won't work for convex ones, and I haven't compared its output with that of the original `System.Drawing.Drawing2D.PathGradientBrush`.

There's an implicit restriction to the `ILineSegment[] lines` constructor argument that it must constitute a closed path (polygon). I thought about adding a validation check, but realized other brush implementations don't seem to do any input validation, so I just leaved it as it is for consistency.

Still, I wanted to ensure that the `lines` argument would represent a closed path (so that users can omit the last segment). I thought invoking `new Polygon(lines)` would automatically make it a closed path, as the implementation of `Path.AsClosedPath` suggests. But it looks like it doesn't work that way, so I also leaved it as it is.

FillRectangleWithDifferentColors_Rgba32_Blank10x10
FillTriangleWithDifferentColors_Rgba32_Blank10x10
FillWithCustomCenterColor_Rgba32_Blank10x10
ShouldRotateTheColorsWhenThereAreMorePoints_Rgba32_Blank10x10

JimBobSquarePants:
@mysticfall If you fork the` Imagesharp.Tests.Images.Repo` and make a separate PR there we can merge that and update the submodule reference in this PR accordingly.

We're going to simplify all this for the future, it's far too complicated and disjointed a process currently.

mysticfall:
@JimBobSquarePants Sure, I can do that. Thank for the instruction :)

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
```diff
@@ -0,0 +1,219 @@
...
+                    if (!this.path.Contains(point))
+                    {
+                        return Color.Transparent.ToPixel<TPixel>();
+                    }
+
+                    Vector2 direction = Vector2.Normalize(point - this.center);
+
+                    PointF end = point + (PointF)(direction * this.maxDistance);
+
+                    (Edge edge, PointF intersection) = this.edges
```
JimBobSquarePants: Linq here will negatively affect performance.

antonfirsov:
@mysticfall merged your PR. If you push a submodule update, your tests should pass.

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
antonfirsov: The performance impacts are coming from stuff like extra allocations, virtual calls, deeper stack etc.

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
```diff
@@ -0,0 +1,219 @@
...
+            GraphicsOptions options)
+            where TPixel : struct, IPixel<TPixel>
+        {
+            return new PathGradientBrushApplicator<TPixel>(source, this.path, this.edges, this.centerColor, options);
+        }
+
+        private static Color CalculateCenterColor(Color[] colors) =>
+            new Color(colors.Select(c => c.ToVector4()).Aggregate((p1, p2) => p1 + p2) / colors.Length);
+
+        private static float DistanceBetween(PointF p1, PointF p2) => ((Vector2)(p2 - p1)).Length();
```
antonfirsov: Squared distance is enough for your purpose.  (`dx^2 + dy^2` instead of  `sqrt( dx^2 + dy^2 )`)

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
```diff
@@ -0,0 +1,219 @@
...
+
+                if (intersections == 0)
+                {
+                    return null;
+                }
+
+                return this.buffer
+                    .Take(intersections)
+                    .Select(p => (point: p, distance: ((Vector2)(p - start)).Length()))
+                    .OrderBy(v => v.distance)
```
antonfirsov: Another optimization hint:
Finding a closest intersection might be enough you don't need to order the whole ordered list.

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
antonfirsov: https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector4.lengthsquared?view=netframework-4.8

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
mysticfall: @antonfirsov Thanks for the clarification. I just want to know exactly how and when using LINQ would affect the performance. I've made this PR to be used for my game project, so the knowledge would benefit me in that aspect too.

I'll try to dig deeper into that LINQ chain to see if there's significant overhead before rewriting it to simple for loops.

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
mysticfall: Removed unnecessary `OrderBy` invocations. Thanks!

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
mysticfall: Upon inspection, `DistanceBetween` cannot be modified that way without affecting the output, since the brush relies on a ratio of actual distances.

I replaced `Length()` used in `FindIntersection` method as suggested, however, as it's only used to find the closest edge.

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
mysticfall: I'm still not sure if the original code would incur significant overhead, but I just replace it with a plain foreach-loop just in case.

I'm not a seasoned C# developer, so if anyone could help me understand where the performance caveat lies, or how to find out such a problem, it would be much appreciated.

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
```diff
@@ -0,0 +1,247 @@
...
+        private readonly Color centerColor;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PathGradientBrush"/> class.
+        /// </summary>
+        /// <param name="lines">Line segments of a polygon that represents the gradient area.</param>
+        /// <param name="colors">Array of colors that correspond to each point in the polygon.</param>
+        /// <param name="centerColor">Color at the center of the gradient area to which the other colors converge.</param>
+        public PathGradientBrush(ILineSegment[] lines, Color[] colors, Color centerColor)
+        {
```
antonfirsov: We need arg. check here:
`lines`, `colors`: not null, and counts are matching

mysticfall: Sure, I considered adding them but thought it wasn't a convention after looking at other brush implementations (see my first comment).

Would it be ok if I just throw `ArgumentNullException` / `ArgumentOutOfRangeException` as I see fit? And, should I also add associated tests?

P. S.: I'm a Godot developer by the way, so yeah I really should be aware of such things :)

mysticfall: It's done. Please let me know if you want me to change anything else.

Issue #967, opened 2019-08-11
Expose ImageProcessor internals to unblock externalizing ImageSharp.Drawing


antonfirsov:
## Plans to externalize SixLabors.ImageSharp.Drawing

The main library, SixLabors.ImageSharp almost reached Release Candidate quality, however we realized, that it will take much more time to bring **SixLabors.ImageSharp.Drawing** and it's base libraries  **SixLabors.Shapes** and **SixLabors.Fonts** to the same level.

In order to unblock the release of ImageSharp, we decided to externalize **SixLabors.ImageSharp.Drawing**. It shall be:
- Versioned independently
- Moved to a separate GitHub repository

We are afraid that Drawing, Shapes, Fonts will reach RC status only after the 1.0 release of the main library.

## Blockers
Unfortunately, externalizing ImageSharp.Drawing is not a trivial job, since it depends on certain internals of the main library by using `[InternalsVisibleTo(...)]`, namely:
- `Buffer2D<T>`
- `ImageProcessor<T>`
- `PixelBlender<T>` and `PixelOperations<T>.GetPixelBlender(...)`
- `IImageVisitor` and `Image.AcceptVisitor(...)`
- `ParallelExecutionSettings`, `configuration.GetParallelSettings()` and `ParallelHelper`

These issues are actually well-known to anyone trying to implement Image Processors, and complaining about classes being internal. (Hi there @Sergio0694 😄!) This turns the issue of ImageSharp.Drawing into an **API-cleanup** problem. 

## Suggested solution

- [x] Expose `Buffer2D<T>` with a minimal subset of public methods. Keep `MemorySource<T>` hidden. Consider reintroducing `IBuffer2D<T>`.
- [x] Expose `ImageProcessor<T>` but consider refactoring the "processor application" logic. See follow-up explanation.
- [x] Expose `PixelBlender<T>` and `PixelOperations<T>.GetPixelBlender(...)`
- [x] Move `IImageVisitor` to the `Advanced` subnamespace, expose `IImageVisitor` and `Image.AcceptVisitor(...)`
- [ ] Find out what to do with the rest 😄. Personally, I'm not sure if we should encourage parallel pixel processing. It might be better+easier to keep that logic hidden, and replicate it in the Drawing library.

### Proposal to refactor current ImageProcessor logic
`IQuantizer` is a good example for the interaction and lifecycle rules of non-generic (pixel agnostic) objects and their generic (pixel-specific) counterparts:
[...]

PR #990, opened 2019-08-25
Implement IDisposable in IImageProcessor<TPixel> instances.

JimBobSquarePants:
### Prerequisites

### Description
Touches #967. Makes `IImageProcessor<TPixel>` inherit `IDIsposable` and ensures all implementations are correctly disposed.

Also ensures any overriding methods are calling their base implementation.


src/ImageSharp/Processing/DefaultImageProcessorContext{TPixel}.cs
```diff
@@ -67,16 +67,23 @@ public IImageProcessingContext ApplyProcessor(IImageProcessor processor, Rectang
...
-                if (processor.CreatePixelSpecificProcessor(this.source, rectangle) is ICloningImageProcessor<TPixel> cloningImageProcessor)
+                using (IImageProcessor<TPixel> specificProcessor = processor.CreatePixelSpecificProcessor(this.source, rectangle))
                 {
-                    this.destination = cloningImageProcessor.CloneAndApply();
-                    return this;
+                    if (specificProcessor is ICloningImageProcessor<TPixel> cloningImageProcessor)
+                    {
+                        this.destination = cloningImageProcessor.CloneAndApply();
+                        return this;
+                    }
```
antonfirsov: If the processor is not a `ICloningImageProcessor<TPixel>` we are disposing and re-creating it.

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
```diff
@@ -144,5 +153,17 @@ protected virtual void AfterFrameApply(ImageFrame<TPixel> source, ImageFrame<TPi
         protected virtual void AfterImageApply(Image<TPixel> destination)
         {
         }
+
+        /// <summary>
+        /// Disposes the object and frees resources for the Garbage Collector.
+        /// </summary>
+        /// <param name="disposing">Whether to dispose managed and unmanaged objects.</param>
+        protected virtual void Dispose(bool disposing)
```
antonfirsov: If we expect unmanaged objects (at least in theory) we should consider adding a finalizer.

In `Image`, I have distinnguished the dispose implementation by naming our method [`DisposeImpl`](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Image.cs#L149). Still wondering if it's a good approach or we should rather `Dispose(bool disposing)`, even if the `bool` parameter is dummy.

JimBobSquarePants: I'd say stick with convention and use `Dispose(bool disposing)`

src/ImageSharp/Processing/DefaultImageProcessorContext{TPixel}.cs
JimBobSquarePants: Good catch, will fix. 👍

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
JimBobSquarePants: I didn't use a finalizer since I didn't want to use (or condone) unmanaged object use.

tests/ImageSharp.Tests/Processing/ImageProcessingContextTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (throwException)
+            {
+                Assert.Throws<ImageProcessingException>(() => this.MutateApply(useBounds));
+            }
+            else
+            {
+                this.MutateApply(useBounds);
+            }
+
+            this.cloningProcessorImpl.Verify(p => p.CloneAndApply(), Times.Once());
```
antonfirsov: @JimBobSquarePants oops. We should expect `this.cloningProcessorImpl.Verify(p => p.Apply(), Times.Once());` here instead. To avoid conflicts, I'll leave it to you.

src/ImageSharp/Processing/DefaultImageProcessorContext{TPixel}.cs
JimBobSquarePants: Maybe I'm just too tired but I cannot see a workaround here. The processors are created from different images.

antonfirsov: Indeed, haven't noticed that!
Makes it harder, I will also keep  thinking tomorrow.

antonfirsov:
We can skip the failing test and merge this as is for now. I may have an idea for unifying the regular and the cloning processor logic, and I can add it in a follow up PR.

PR #1005, opened 2019-09-08
Fix #1004

### LINKED ISSUE #1004: PngDecoder should not throw when encountering invalid Zlib data for non critical chunks.
### Prerequisites

### Description
The following image will cause our decoder to throw due to an invalid Zlib data contained within the zTxt chunk.  While correct, this chunk is non-critical and will be ignored during decoding by libpng and System.Drawing.

versioning-1_1

### System Configuration

- ImageSharp version:
- Other ImageSharp packages and versions: 1.0.0-beta0007
- Environment (Operating system, version and so on): All
- .NET Framework version: All
- Additional information:

JimBobSquarePants:
### Prerequisites

### Description
Fixes #1004  Changes the behavior of the `ZlibInflateStream` to throw only when operating on a critical chunk.


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -175,11 +175,18 @@ public Image<TPixel> Decode<TPixel>(Stream stream)
                                     this.InitializeImage(metadata, out image);
                                 }
 
-                                using (var deframeStream = new ZlibInflateStream(this.currentStream, this.ReadNextDataChunk))
+                                var deframeStream = new ZlibInflateStream(this.currentStream, this.ReadNextDataChunk);
```
antonfirsov: I think keeping the `using` block would be cleaner here. Or am I missing something?

JimBobSquarePants: Yeah, thought it was throwing after we assign the stream but it isn't so have refactored.

Issue #688, opened 2018-08-30
DrawText throws ArgumentException if text paint exceeds bounds of image


agrath:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When painting text if the text exceeds the boundary of the image horizontally due to font size with no wrapping
The DrawText method throws an argumentexception

```
ArgumentOutOfRangeException: Specified argument was out of the range of valid values.``
 Parameter name: start
System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument)
SixLabors.ImageSharp.Processing.SolidBrush+SolidBrushApplicator.Apply(Span<float> scanline, int x, int y)
SixLabors.ImageSharp.Processing.Processors.Text.DrawTextProcessor.<OnFrameApply>g__Draw|22_0(List<DrawingOperation> operations, IBrush<TPixel> brush, ref <>c__DisplayClass22_0 )
SixLabors.ImageSharp.Processing.Processors.Text.DrawTextProcessor.OnFrameApply(ImageFrame<TPixel> source, Rectangle sourceRectangle, Configuration configuration)
SixLabors.ImageSharp.Processing.Processors.ImageProcessor.Apply(ImageFrame<TPixel> source, Rectangle sourceRectangle, Configuration configuration)
```

### Steps to Reproduce

```
sing (Image<Rgba32> image = Image.Load("Assets/share-bg.png"))
            {
                FontCollection fontCollection = new FontCollection();
                Font regularFont = fontCollection.Install("Assets/TitilliumWeb-SemiBold.ttf").CreateFont(24, FontStyle.Regular);
                Font italicFont = fontCollection.Install("Assets/TitilliumWeb-BoldItalic.ttf").CreateFont(24, FontStyle.Italic);

                image.Mutate(x => x.DrawText(text, regularFont, Rgba32.White, new PointF(100,100)));

                MemoryStream stream = new MemoryStream();
                image.Save(stream, new PngEncoder());
                stream.Position = 0;
                return stream;
            }
```

[...]

PR #1008, opened 2019-09-12
Fix horizontally out-of-bounds error when drawing text

Turnerj:
### Prerequisites

### Description

This addresses the horizontal out-of-bounds error raised in #688. The error was confirmed using [the code sample provided in one of the comments](https://github.com/SixLabors/ImageSharp/issues/688#issuecomment-451537085).

### DIFF
--- a/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
+++ b/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
@@ -90,26 +90,31 @@ namespace SixLabors.ImageSharp.Processing.Processors.Text
                             Buffer2D<float> buffer = operation.Map;
                             int startY = operation.Location.Y;
                             int startX = operation.Location.X;
-                            int offSetSpan = 0;
+                            int offsetSpan = 0;
                             if (startX < 0)
                             {
-                                offSetSpan = -startX;
+                                offsetSpan = -startX;
                                 startX = 0;
                             }
 
-                            int fistRow = 0;
+                            if (startX >= source.Width)
+                            {
+                                continue;
+                            }
+
+                            int firstRow = 0;
                             if (startY < 0)
                             {
-                                fistRow = -startY;
+                                firstRow = -startY;
                             }
 
                             int maxHeight = source.Height - startY;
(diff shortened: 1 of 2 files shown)


src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
```diff
@@ -90,26 +90,31 @@ void Draw(List<DrawingOperation> operations, IBrush brush)
...
                             {
-                                offSetSpan = -startX;
+                                offsetSpan = -startX;
                                 startX = 0;
                             }
 
-                            int fistRow = 0;
+                            if (startX >= source.Width)
+                            {
+                                continue;
```
Turnerj: I'm not sure if `continue` here is actually the best idea and not just short-circuiting with a `break`. I might look into this further just so the code doesn't need to loop additionally for no reason.

tests/ImageSharp.Tests/Drawing/Text/DrawTextOnImageTests.cs
```diff
@@ -65,6 +65,24 @@ public void DoesntThrowExceptionWhenOverlappingRightEdge_Issue688<TPixel>(TestIm
             }
         }
 
+        [Theory]
+        [WithSolidFilledImages(1500, 500, "White", PixelTypes.Rgba32)]
+        public void DoesntThrowExceptionWhenOverlappingRightEdge_Issue688_2<TPixel>(TestImageProvider<TPixel> provider)
```
Turnerj: There is a test above this which is very similar and was targeted to hit the same bug but based on the comments of #688, it doesn't actually hit the issue (or at least not the same way). With this in mind, the names of the tests are extremely similar (mine has "_2") so I'm not sure if I should just replace that test with this one or give mine a different name.

src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
Turnerj: After a bit more looking into this, it seems like it could break (no pun intended) certain brushes depending what direction they were drawing (eg. backward diagonal) if I were to `break` instead of `continue`. I could write a short circuit for it depending on the type of brush but that feels a little overkill.

PR #1009, opened 2019-09-17
Refactor IDisposable pattern

JimBobSquarePants:
### Prerequisites

### Description
Touches #967 

- Refactor of `IDisposable` pattern for images, frames, and processors to work better with inheritance and ensure finalizer suppression.
- `Image<TPixel>` is no longer sealed as that prevents extending for potential types like `IntegralImage<TPixel>`. `ImageFrame<TPixel>` is left sealed as inheritance should not be required.

### DIFF
--- a/src/ImageSharp/Image{TPixel}.cs
+++ b/src/ImageSharp/Image{TPixel}.cs
@@ -18,7 +18,7 @@ namespace SixLabors.ImageSharp
     /// For generic <see cref="Image{TPixel}"/>-s the pixel type is known at compile time.
     /// </summary>
     /// <typeparam name="TPixel">The pixel format.</typeparam>
-    public class Image<TPixel> : Image
+    public sealed class Image<TPixel> : Image
         where TPixel : struct, IPixel<TPixel>
     {
         private bool isDisposed;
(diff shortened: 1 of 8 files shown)


src/ImageSharp/Image{TPixel}.cs
```diff
@@ -18,9 +18,11 @@ namespace SixLabors.ImageSharp
     /// For generic <see cref="Image{TPixel}"/>-s the pixel type is known at compile time.
     /// </summary>
     /// <typeparam name="TPixel">The pixel format.</typeparam>
-    public sealed class Image<TPixel> : Image
+    public class Image<TPixel> : Image
```
antonfirsov: Without prototyping an actual subclassing use-case, I find this dangerous and unnecessary. We can do the sealed -> non-sealed flip any time, but not the opposite.

JimBobSquarePants: No worries, will re seal.

JimBobSquarePants:
Gonna merge this without waiting for the build. I don't want to wait 40 minutes and we know it builds/passes tests.

PR #1010, opened 2019-09-17
Expose visitor through advanced namespace.

JimBobSquarePants:
### Prerequisites

### Description
- Adds new `AcceptVisitor` extension method to the `Advanced` namespace and redirects all calls to that method.

### DIFF
--- a/src/ImageSharp/Advanced/AdvancedImageExtensions.cs
+++ b/src/ImageSharp/Advanced/AdvancedImageExtensions.cs
@@ -16,6 +16,15 @@ namespace SixLabors.ImageSharp.Advanced
     public static class AdvancedImageExtensions
     {
         /// <summary>
+        /// Accepts a <see cref="IImageVisitor"/> to implement a double-dispatch pattern in order to
+        /// apply pixel-specific operations on non-generic <see cref="Image"/> instances
+        /// </summary>
+        /// <param name="source">The source.</param>
+        /// <param name="visitor">The visitor.</param>
+        public static void AcceptVisitor(this Image source, IImageVisitor visitor)
+            => source.Accept(visitor);
+
+        /// <summary>
         /// Gets the configuration for the image.
         /// </summary>
         /// <param name="source">The source image.</param>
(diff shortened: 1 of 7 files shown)


src/ImageSharp/Image.cs
```diff
@@ -131,11 +130,12 @@ public Image<TPixel2> CloneAs<TPixel2>()
...
 
         /// <summary>
-        /// Accept a <see cref="IImageVisitor"/>.
+        /// Accepts a <see cref="IImageVisitor"/>.
         /// Implemented by <see cref="Image{TPixel}"/> invoking <see cref="IImageVisitor.Visit{TPixel}"/>
         /// with the pixel type of the image.
         /// </summary>
-        internal abstract void AcceptVisitor(IImageVisitor visitor);
+        /// <param name="visitor">The visitor.</param>
+        protected internal abstract void Accept(IImageVisitor visitor);
```
antonfirsov: It's unlikely that anyone will try to subclass `Image`, but in theory, we expose an extension point with `protected internal`, so `internal` is probably better.

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
```diff
@@ -15,6 +15,15 @@ namespace SixLabors.ImageSharp.Advanced
     /// </summary>
     public static class AdvancedImageExtensions
     {
+        /// <summary>
+        /// Accepts a <see cref="IImageVisitor"/> to implement a double-dispatch pattern in order to
+        /// apply pixel-specific operations on non-generic <see cref="Image"/> instances
+        /// </summary>
+        /// <param name="source">The source.</param>
+        /// <param name="visitor">The visitor.</param>
+        public static void AcceptVisitor(this Image source, IImageVisitor visitor)
```
antonfirsov: 👍

src/ImageSharp/Image.cs
antonfirsov: Ok, I've just seen #1011. See my comments there!

JimBobSquarePants: Cool. Will update once I merge #1009

PR #1011, opened 2019-09-17
Make processors public, refactor cloning.

JimBobSquarePants:
### Prerequisites

### Description
Touches. #967 

- Refactors the image processors to make the interfaces and base classes public. 
- Removes double pixel processor creation.
- Simplified cloned target image creation.

<strike>I've marked this as draft as I want to do something more extreme and remove `ICloningImageProcessor<TPixel>` moving the interface method into the base class. This makes additional methods impossible to test against though via mocking. </strike>

Changed my mind and figured out a better way to create contracts for cloned/regular processors.

Thoughts please!

### DIFF
--- a/src/ImageSharp/Advanced/AotCompilerTools.cs
+++ b/src/ImageSharp/Advanced/AotCompilerTools.cs
@@ -2,13 +2,12 @@
 // Licensed under the Apache License, Version 2.0.
 
 using System.Numerics;
+using System.Runtime.CompilerServices;
 using SixLabors.ImageSharp.Formats;
 using SixLabors.ImageSharp.Formats.Jpeg.Components;
 using SixLabors.ImageSharp.PixelFormats;
-using SixLabors.ImageSharp.Processing;
 using SixLabors.ImageSharp.Processing.Processors.Dithering;
 using SixLabors.ImageSharp.Processing.Processors.Quantization;
-using SixLabors.ImageSharp.Processing.Processors.Transforms;
 
 namespace SixLabors.ImageSharp.Advanced
 {
@@ -81,9 +80,8 @@ namespace SixLabors.ImageSharp.Advanced
             AotCompileWuQuantizer<TPixel>();
             AotCompileDithering<TPixel>();
             AotCompilePixelOperations<TPixel>();
-            AotCompileResizeOperations<TPixel>();
 
-            System.Runtime.CompilerServices.Unsafe.SizeOf<TPixel>();
+            Unsafe.SizeOf<TPixel>();
 
             AotCodec<TPixel>(new Formats.Png.PngDecoder(), new Formats.Png.PngEncoder());
             AotCodec<TPixel>(new Formats.Bmp.BmpDecoder(), new Formats.Bmp.BmpEncoder());
@@ -107,8 +105,10 @@ namespace SixLabors.ImageSharp.Advanced
         private static void AotCompileOctreeQuantizer<TPixel>()
             where TPixel : struct, IPixel<TPixel>
         {
-            var test = new OctreeFrameQuantizer<TPixel>(new OctreeQuantizer(false));
-            test.AotGetPalette();
+            using (var test = new OctreeFrameQuantizer<TPixel>(new OctreeQuantizer(false)))
+            {
+                test.AotGetPalette();
+            }
         }
 
         /// <summary>
@@ -118,9 +118,11 @@ namespace SixLabors.ImageSharp.Advanced
         private static void AotCompileWuQuantizer<TPixel>()
             where TPixel : struct, IPixel<TPixel>
         {
-            var test = new WuFrameQuantizer<TPixel>(Configuration.Default.MemoryAllocator, new WuQuantizer(false));
-            test.QuantizeFrame(new ImageFrame<TPixel>(Configuration.Default, 1, 1));
-            test.AotGetPalette();
+            using (var test = new WuFrameQuantizer<TPixel>(Configuration.Default.MemoryAllocator, new WuQuantizer(false)))
+            {
+                test.QuantizeFrame(new ImageFrame<TPixel>(Configuration.Default, 1, 1));
+                test.AotGetPalette();
+            }
         }
 
         /// <summary>
@@ -132,7 +134,10 @@ namespace SixLabors.ImageSharp.Advanced
         {
             var test = new FloydSteinbergDiffuser();
             TPixel pixel = default;
-            test.Dither(new ImageFrame<TPixel>(Configuration.Default, 1, 1), pixel, pixel, 0, 0, 0, 0, 0, 0);
+            using (var image = new ImageFrame<TPixel>(Configuration.Default, 1, 1))
+            {
+                test.Dither(image, pixel, pixel, 0, 0, 0, 0, 0, 0);
+            }
         }
 
         /// <summary>
@@ -170,17 +175,6 @@ namespace SixLabors.ImageSharp.Advanced
         {
             var pixelOp = new PixelOperations<TPixel>();
             pixelOp.GetPixelBlender(PixelColorBlendingMode.Normal, PixelAlphaCompositionMode.Clear);
-        }
-
-        /// <summary>
-        /// This method pre-seeds the ResizeProcessor for the AoT compiler on iOS.
-        /// </summary>
-        /// <typeparam name="TPixel">The pixel format.</typeparam>
-        private static void AotCompileResizeOperations<TPixel>()
-            where TPixel : struct, IPixel<TPixel>
-        {
-            var genericResizeProcessor = (ResizeProcessor<TPixel>)new ResizeProcessor(new ResizeOptions(), default).CreatePixelSpecificProcessor(new Image<TPixel>(0, 0), default);
-            genericResizeProcessor.AotCreateDestination();
         }
     }
 }
--- a/src/ImageSharp/Processing/DefaultImageProcessorContext{TPixel}.cs
+++ b/src/ImageSharp/Processing/DefaultImageProcessorContext{TPixel}.cs
@@ -69,9 +69,9 @@ namespace SixLabors.ImageSharp.Processing
             {
                 // When cloning an image we can optimize the processing pipeline by avoiding an unnecessary
                 // interim clone if the first processor in the pipeline is a cloning processor.
-                if (processor is CloningImageProcessor cloningImageProcessor)
+                if (processor is ICloningImageProcessor cloningImageProcessor)
                 {
-                    using (ICloningImageProcessor<TPixel> pixelProcessor = cloningImageProcessor.CreatePixelSpecificProcessor(this.source, rectangle))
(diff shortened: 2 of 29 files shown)


antonfirsov:
I had an idea a few weeks ago for the same point. Actually, I had two options in my mind, and wanted to ask for input to decide which one is better, but I haven't got the the chance to present them so far.

This solution is a third option actually 😄 I will have a deeper look at it now, but I can't promise I can summarize my thoughts properly tonight.

src/ImageSharp/Processing/DefaultImageProcessorContext{TPixel}.cs
```diff
@@ -64,26 +67,25 @@ public IImageProcessingContext ApplyProcessor(IImageProcessor processor, Rectang
         {
             if (!this.mutate && this.destination is null)
             {
-                // This will only work if the first processor applied is the cloning one thus
-                // realistically for this optimization to work the resize must the first processor
-                // applied any only up processors will take the double data path.
-                using (IImageProcessor<TPixel> specificProcessor = processor.CreatePixelSpecificProcessor(this.source, rectangle))
+                // When cloning an image we can optimize the processing pipeline by avoiding an unnecessary
+                // interim clone if the first processor in the pipeline is a cloning processor.
+                if (processor is CloningImageProcessor cloningImageProcessor)
```
antonfirsov: Feels like we should not use a specific implementation here.

src/ImageSharp/Processing/Processors/CloningImageProcessor.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <param name="source">The source image. Cannot be null.</param>
+        /// <param name="sourceRectangle">
+        /// The <see cref="Rectangle"/> structure that specifies the portion of the image object to draw.
+        /// </param>
+        /// <returns>The <see cref="ICloningImageProcessor{TPixel}"/></returns>
+        public abstract ICloningImageProcessor<TPixel> CreatePixelSpecificProcessor<TPixel>(Image<TPixel> source, Rectangle sourceRectangle)
+            where TPixel : struct, IPixel<TPixel>;
+
+        /// <inheritdoc/>
+        IImageProcessor<TPixel> IImageProcessor.CreatePixelSpecificProcessor<TPixel>(Image<TPixel> source, Rectangle sourceRectangle)
```
antonfirsov: This method is actually never called.

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
```diff
@@ -45,16 +47,12 @@ protected CloningImageProcessor(Image<TPixel> source, Rectangle sourceRectangle)
         protected Configuration Configuration { get; }
 
         /// <inheritdoc/>
-        public Image<TPixel> CloneAndApply()
+        public Image<TPixel> CloneAndExecute()
```
antonfirsov: Suspicious: two verbs in the method name.

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
```diff
@@ -45,16 +47,12 @@ protected CloningImageProcessor(Image<TPixel> source, Rectangle sourceRectangle)
...
         {
             try
             {
                 Image<TPixel> clone = this.CreateDestination();
```
antonfirsov: All `CreateDestination` implementations share the same implementation actually, doing deep clone in fact.

JimBobSquarePants:
> This solution is a third option actually 😄 I will have a deeper look at it now, but I can't promise I can summarize my thoughts properly tonight.

Excellent. Looking forward to seeing what you come up with. I'm going to replace the Mocks with dummy pass-through instances so I can demonstrate my intent in the interim and continue to provide coverage.

src/ImageSharp/Processing/Processors/CloningImageProcessor.cs
JimBobSquarePants: It is, if the destination image is not null in the default processing context.

antonfirsov: Yeah, correct!

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
JimBobSquarePants: Yeah, pretty much. I want to do some refactoring to allow us to do a deep but empty clone without having to write so much code.

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
JimBobSquarePants: Naming is hard. I am up for suggestions.

src/ImageSharp/Processing/DefaultImageProcessorContext{TPixel}.cs
JimBobSquarePants: I did this as I think we're going too heavy on the interfaces and I want to allow us flexibility to change implementation hence my idea to remove `ICloningImageProcessor<TPixel>`. By using the concrete instance and proxying the `CreatePixelSpecificProcessor` I think that better lets us do that.

This optimization behavior is specific to `DefaultImageProcessorContext` anyway.

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
antonfirsov: Or just query the size from the cloning processor, and do the rest in a helper method ;)

JimBobSquarePants: Made the method private and used the size as an abstract method.

src/ImageSharp/Processing/DefaultImageProcessorContext{TPixel}.cs
JimBobSquarePants: Introduced `ICloningImageProcessor` and appropriate methods. Much cleaner now.

src/ImageSharp/Advanced/AotCompilerTools.cs
```diff
@@ -107,8 +105,10 @@ private static void Seed<TPixel>()
         private static void AotCompileOctreeQuantizer<TPixel>()
             where TPixel : struct, IPixel<TPixel>
         {
-            var test = new OctreeFrameQuantizer<TPixel>(new OctreeQuantizer(false));
-            test.AotGetPalette();
+            using (var test = new OctreeFrameQuantizer<TPixel>(new OctreeQuantizer(false)))
```
JimBobSquarePants: Spotted these when removing the AOT CreateDestination method. No longer required as not virtual.

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
```diff
@@ -38,21 +42,17 @@ protected CloningImageProcessor(Image<TPixel> source, Rectangle sourceRectangle)
...
 
         /// <summary>
-        /// Gets the <see cref="ImageSharp.Configuration"/> instance to use when performing operations.
+        /// Gets the <see cref="Configuration"/> instance to use when performing operations.
         /// </summary>
         protected Configuration Configuration { get; }
 
         /// <inheritdoc/>
-        public Image<TPixel> CloneAndApply()
+        Image<TPixel> ICloningImageProcessor<TPixel>.CloneAndExecute()
```
JimBobSquarePants: Hidden so not to clutter child classes.

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
```diff
@@ -160,5 +167,30 @@ protected virtual void AfterImageApply(Image<TPixel> destination)
         protected virtual void Dispose(bool disposing)
         {
         }
+
+        private Image<TPixel> CreateTarget()
```
JimBobSquarePants: This method was identical in every implementation. The only different was where the target size came from.

src/ImageSharp/Processing/Processors/Transforms/AffineTransformProcessor{TPixel}.cs
```diff
@@ -29,50 +30,37 @@ internal class AffineTransformProcessor<TPixel> : TransformProcessor<TPixel>
...
 
-            IResampler sampler = this.Definition.Sampler;
-
-            if (sampler is NearestNeighborResampler)
+            if (this.resampler is NearestNeighborResampler)
             {
                 ParallelHelper.IterateRows(
                     targetBounds,
-                    this.Configuration,
+                    configuration,
```
JimBobSquarePants: Spotted we were capturing the whole class in the lambda

antonfirsov: I don't get the difference. `this.Configuration` also returns a `Configuration` instance. 

Also: capturing the whole class is not necessarily evil. Or is it? Am I missing something?

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
antonfirsov: It's not the naming actually. Might be a sign, that we can improve the design here.

src/ImageSharp/Processing/Processors/Transforms/AffineTransformProcessor{TPixel}.cs
JimBobSquarePants: Bad practice as I understand that can lead to negative performance and issues since capturing variables affects the lifetime of those captures.

src/ImageSharp/Processing/Processors/CloningImageProcessor{TPixel}.cs
```diff
@@ -38,21 +42,17 @@ protected CloningImageProcessor(Image<TPixel> source, Rectangle sourceRectangle)
...
             try
             {
-                Image<TPixel> clone = this.CreateDestination();
-
-                if (clone.Frames.Count != this.Source.Frames.Count)
-                {
-                    throw new ImageProcessingException($"An error occurred when processing the image using {this.GetType().Name}. The processor changed the number of frames.");
-                }
+                Image<TPixel> clone = this.CreateTarget();
+                this.CheckFrameCount(this.Source, clone);
```
antonfirsov: Not sure if the checks are necessary. The centralized cloning logic should guarantee that the frame counts are matching.

JimBobSquarePants: I actually thought that but was too lazy to check and just merged the code into a single method

antonfirsov: It's an iterative way to delete bloated code 😉

JimBobSquarePants:
Setting a countdown on this one. I want to keep progressing.

PR #1012, opened 2019-09-19
Added: ability to skip unneeded chunks for optimization mode

equinox2k:
### Prerequisites

### Description
This part of work on creating an alternative to TinyPNG / PNGCrush using ImageSharp.


JimBobSquarePants:
Good idea! I think we should make our chunk preservation more granular though, perhaps an enum so we can choose whether to preserve exif? 

Incidentally deleting the relevant metadata will achieve e same result, but this could be a nice shortcut

equinox2k:
@JimBobSquarePants split up optimization into a flagged enum so should let users choose what they would like to do + fixes

antonfirsov:
In decoders, we have an `IgnoreMetadata` property. Can we design it in a way that tries to maintain consistency between decodsrs/encoders accross all codecs?

src/ImageSharp/Formats/Png/PngOptimizeMethod.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+
+namespace SixLabors.ImageSharp.Formats.Png
+{
+    /// <summary>
+    /// Provides enumeration of available PNG optimization methods.
+    /// </summary>
+    [Flags]
+    public enum PngOptimizeMethod
```
antonfirsov: As I mentioned on gitter, it's better if our naming should reflect functional semantics rather than a specific use case (optimization).  In that case "positive" flags may be easier than negative ones expressing the suppression.
Suggestion:
```csharp
public enum PngChunkFilter
{
    IncludeNone = 0,
    IncludePhysicalChunk = 1 << 0,
    IncludeGammaChunk = 1 << 1,
    IncludeExifChunk = 1 << 2,
    IncludeTextChunks = 1 << 3,

    IncludeAll = IncludePhysicalChunk | IncludeGammaChunk | IncludeExifChunk | IncludeTextChunks
}
```

src/ImageSharp/Formats/Png/PngOptimizeMethod.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Suppress the tTXt, iTXt or zTXt chunk.
+        /// </summary>
+        SuppressTextChunks = 8,
+
+        /// <summary>
+        /// Make funlly transparent pixels black.
+        /// </summary>
+        MakeTransparentBlack = 16,
```
antonfirsov: This one is really different from the other ones, I would rather place this one into a separate property inside `PngEncoder`.

src/ImageSharp/Formats/Png/PngOptimizeMethod.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// All possible optimizations.
+        /// </summary>
+        All = 31,
```
antonfirsov: Defining the union of the flags manually is more prone to mistakes when changed in future.

src/ImageSharp/Formats/Png/PngOptimizeMethod.cs
JimBobSquarePants: Agreed. It doesn't seem to fit. Is the naming accurate also? It's not black, it's empty.

src/ImageSharp/Formats/Png/PngOptimizeMethod.cs
JimBobSquarePants: I much prefer the suggested  naming definition.

The overall metadata API is designed to preserve as much metadata as possible so I think we should adopt the negative flags. This to me as a consumer would cognitively reflect that intention as I'm building a rule for what to exclude.

src/ImageSharp/Formats/Png/PngOptimizeMethod.cs
JimBobSquarePants: `All = ~None`.

brianpopow:
I like the idea, i want to continue on this.

src/ImageSharp/Formats/Png/PngChunkFilter.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Excludes the tTXt, iTXt or zTXt chunk from encoding.
+        /// </summary>
+        ExcludeTextChunks = 1 << 3,
+
+        /// <summary>
+        /// All possible optimizations.
+        /// </summary>
+        ExcludeAll = ExcludePhysicalChunk | ExcludeGammaChunk | ExcludeExifChunk | ExcludeTextChunks
```
JimBobSquarePants, line 42:
```suggestion
        ExcludeAll = ~None
```

src/ImageSharp/Formats/Png/PngEncoder.cs
```diff
@@ -42,6 +42,16 @@ public sealed class PngEncoder : IImageEncoder, IPngEncoderOptions
...
 
+        /// <summary>
+        /// Gets or sets the chunk filter. This can be used to exclude some ancillary chunks from being written.
+        /// </summary>
+        public PngChunkFilter? ChunkFilter { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether fully transparent pixels should be converted to black pixels.
+        /// </summary>
+        public bool MakeTransparentBlack { get; set; }
```
JimBobSquarePants, line 53: I don't really understand the use case for this property. Why would we choose `Black` over `Transparent`?

UkooLabs: @JimBobSquarePants the intention is to convert fully transparent pixels that may contain r g b values to a transparent black I.e r g b of 0 and thus should six compression

UkooLabs: So bane maybe confusing but something like MakeTransparentIntoTransparentBlack Is long :)

src/ImageSharp/Formats/Png/PngOptimizeMethod.cs
brianpopow: i think `MakeTransparentBlack` is clear in what it does, but im open for other suggestions

src/ImageSharp/Formats/Png/PngOptimizeMethod.cs
brianpopow: this is changed now to negative flags

brianpopow:
> In decoders, we have an `IgnoreMetadata` property. Can we design it in a way that tries to maintain consistency between decodsrs/encoders accross all codecs?

how can we do that properly? I could add `IgnoreMetadata` to the Encoder options, which would set the ChunkFilter to `ExcludeAll`. Other idea's?

src/ImageSharp/Formats/Png/PngEncoder.cs
JimBobSquarePants: @UkooLabs  That's what I thought it meant to do. However the implementation is converting to `Color.Black` rather than `Color.Transparent`. Probably because the property name needs some love. `ClearTransparency` perhaps or even a two property enum. `PngTransparentColorBehavior` `Clear`, `Preserve`

JimBobSquarePants:
>how can we do that properly? I could add IgnoreMetadata to the Encoder options, which would set the ChunkFilter to ExcludeAll. Other idea's?

@brianpopow I think that's the best route yes. Then we can mirror the API in a neat way and enable the user to save with and without versions without having to delete their metadata.

src/ImageSharp/Formats/Png/PngEncoder.cs
brianpopow: yeah i really misunderstood that. I was thinking it should be `Color.Black`

JimBobSquarePants: Naming is hard...

src/ImageSharp/Formats/Png/PngTransparentColorBehavior.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Six Labors and contributors.
+// Licensed under the GNU Affero General Public License, Version 3.
+
+namespace SixLabors.ImageSharp.Formats.Png
+{
+    /// <summary>
+    /// Enum indicating how the transparency should be handled on encoding.
+    /// </summary>
+    public enum PngTransparentColorBehavior
```
JimBobSquarePants, line 9: My bad, let's call it `PngTransparentColorMode` to keep it in line with othernaming.

Issue #994, opened 2019-09-02
Weird AdaptiveHistogramEqualizationProcessor output for specific dimensions


antonfirsov:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
`AdaptiveHistogramEqualizationProcessor` is producing weird output for dimensions 110x110:
![Issue984_Rgba32_TestPattern110x110](https://user-images.githubusercontent.com/6835152/64082208-c54cf480-cd0b-11e9-93b2-18c0935551d7.png)

### Steps to Reproduce
Run test case `HistogramEqualizationTests.Issue984(110, 110)`

### System Configuration
- ImageSharp version: current main
- Other ImageSharp packages and versions: -
- Environment (Operating system, version and so on): any
- .NET Framework version: any
- Additional information: -

brianpopow:
I know the reason for that now. The clipping is not working as intended for such small tile sizes. I need to rework the way clipping is done. I will provide a PR as soon as i find time.

antonfirsov:
Thanks for the fix!

PR #1017, opened 2019-09-30
Change histogram clipLimit to be an absolute value

brianpopow:
### Prerequisites

### Description

I have looked into how opencv handles clipping and adopted the strategy they are using. The main change is that it is now a absolute value, instead of a percentage value. The default value for clipping has changed accordingly.

I have changed the tests for Issue984 to compare the output to a reference output. I will open a separate PR to add the reference images.

This PR fixes issue #994.


antonfirsov:
@brianpopow just merged SixLabors/Imagesharp.Tests.Images#12

src/ImageSharp/Processing/Processors/Normalization/HistogramEqualizationOptions.cs
```diff
@@ -32,14 +32,13 @@ public class HistogramEqualizationOptions
         public bool ClipHistogram { get; set; } = false;
 
         /// <summary>
-        /// Gets or sets the histogram clip limit in percent of the total pixels in a tile. Histogram bins which exceed this limit, will be capped at this value.
-        /// Defaults to 0.035f.
+        /// Gets or sets the histogram clip limit. Histogram bins which exceed this limit, will be capped at this value. Defaults to 350.
```
antonfirsov: What is the range allowed?

JimBobSquarePants: I would have assumed it was connected to the luminance levels which with the percentile value could be applied to both 8 and 16 bit images. It can't now so we should have some sort of range validation to ensure we're doing the right thing.

brianpopow: This depends on the number of pixels in a tile. So its depended on the number of tiles choosen and the image size itself. For example if the image is 1000x1000 pixels and the tiles choosen is 10, the image would be split up into 10 by 10 tiles with each 100 x 100  pixels. In the worst case all pixels in a tile would have the same color and land in the same histogram bin. So a clipLimit of 100x100 pixels would be the upper limit.

There is a Guard in place in the HistogramEqualization constructor ensuring its at least 1: [ClipLimit Guard](https://github.com/brianpopow/ImageSharp/blob/67ea7b43c2e7027fc020a655be6d39c00b27ace9/src/ImageSharp/Processing/Processors/Normalization/HistogramEqualizationProcessor%7BTPixel%7D.cs#L41)

I am not sure what a reasonable upper limit could be. I think this has to be left to the user to choose. The worst which can happen is that the user chooses a clipLimit which is higher then actually possible. In the above example that would be a value higher than 10000. In this case it would just mean the histogram would not be clipped at all.

antonfirsov: Is it equivalent to the OpenCV implementation now? Does it make sense to reference [their docs](https://docs.opencv.org/3.4/d5/daf/tutorial_py_histogram_equalization.html) from ours as  clarification?

JimBobSquarePants: If not exactly equivalent we should at least expand upon the documentation for the `clipLimit` property to make it more clear. Once we do that I am happy to merge the code. 👍

brianpopow: I have tried to clarify the meaning of the clipLimit in the comment and linked to [wikipedia AHE](https://en.wikipedia.org/wiki/Adaptive_histogram_equalization#Contrast_Limited_AHE)

JimBobSquarePants: Perfect!

PR #1028, opened 2019-10-16
Expose pixelblenders. #967

JimBobSquarePants:
### Prerequisites

### Description
Exposes the PixelBlender API to the public.


src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -189,38 +203,58 @@ public static Vector4 Atop(Vector4 dst, Vector4 src, Vector4 blend)
...
+        /// <param name="source">The source vector.</param>
+        /// <param name="blend">The amount to blend. Range 0..1</param>
+        /// <returns>The <see cref="Vector4"/>.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static Vector4 In(Vector4 dst, Vector4 src, Vector4 blend)
+        public static Vector4 In(Vector4 destination, Vector4 source, Vector4 blend)
         {
-            float alpha = dst.W * src.W;
+            // TODO: blend is not used?
+            float alpha = destination.W * source.W;
```
JimBobSquarePants: This concerns me. We have generated many methods that call this and I would expect it to use the blend vector and yield a different response. I'll need help implementing this.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -189,38 +203,58 @@ public static Vector4 Atop(Vector4 dst, Vector4 src, Vector4 blend)
...
 
+        /// <summary>
+        /// Returns the result of the "Out" compositing equation.
+        /// </summary>
+        /// <param name="destination">The destination vector.</param>
+        /// <param name="source">The source vector.</param>
+        /// <returns>The <see cref="Vector4"/>.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static Vector4 Out(Vector4 dst, Vector4 src)
+        public static Vector4 Out(Vector4 destination, Vector4 source)
```
JimBobSquarePants: `Out` and `Xor` are called by several generated methods but no blend vector is created or used for each method which means identical results . I would expect a blend vector to be generated and passed.

src/ImageSharp/Color/Color.cs
```diff
@@ -178,7 +178,7 @@ public override int GetHashCode()
             where TPixel : struct, IPixel<TPixel>
         {
             ReadOnlySpan<Rgba64> rgba64Span = MemoryMarshal.Cast<Color, Rgba64>(source);
-            PixelOperations<TPixel>.Instance.FromRgba64(Configuration.Default, rgba64Span, destination);
+            PixelOperations<TPixel>.Instance.FromRgba64(configuration, rgba64Span, destination);
```
antonfirsov: 👍

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.Generated.tt
```diff
@@ -17,18 +17,21 @@
...
 #>
-
-using System;
 using System.Numerics;
 using System.Runtime.CompilerServices;
 
 namespace SixLabors.ImageSharp.PixelFormats.PixelBlenders
 {
-    internal static partial class PorterDuffFunctions
+    public static partial class PorterDuffFunctions
```
antonfirsov: Since not directly needed for #967, I wouldn't publish this now. This way we can postpone optimizing the API.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
antonfirsov: Just an assumption, but I think `blend` is only a parameter to simplify method generators by using uniform call sites. (Compare `AddSrcOver` vs `AddSrcIn` in `PorterDuffFunctions.Generated.cs`.

I need more research to be 100% sure though. Maybe @vpenades or @tocsoft can confirm?

src/ImageSharp/PixelFormats/PixelBlender{TPixel}.cs
```diff
@@ -91,7 +90,7 @@ internal abstract class PixelBlender<TPixel>
         /// <param name="source">the source span</param>
         /// <param name="amount">
         /// A span with values between 0 and 1 indicating the weight of the second source vector.
-        /// At amount = 0, "from" is returned, at amount = 1, "to" is returned.
+        /// At amount = 0, "background" is returned, at amount = 1, "source" is returned.
         /// </param>
         public void Blend<TPixelSrc>(
```
antonfirsov: This method has no direct tests. It shouldn't be public yet probably.

JimBobSquarePants: I think it's probably ok. We've got some heavy implicit testing via the calling `Blend` method.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
vpenades: The original Porter Duff formulas are based on a table where all the terms are multipled either by Zero, One or Blend.   There's a few formulas that don't use Blend or it's simply cancelled out, in these cases, the _Source_ is not actually blended with _Dest_ , but is used as a mask. __Xor__ is a good example.

@antonfirsov is right, the Blend parameter is there to simplify code generation, even if some code paths don't use it. The other option would be to hand write all the methods.

vpenades: I'ts been a while since I visited that code.... but I can take a deeper look, let me know if you need some clarification about some specific method.

JimBobSquarePants: @vpenades Thanks for getting back to us, very much appreciated!

This makes sense then, I think I've found the table you were referencing.

http://ssp.impulsetrain.com/porterduff.html

JimBobSquarePants: @vpenades  The only question then I have is with the `In` method. I can see that the template calls the Blend method for each type but the function implementation never uses the result. 

https://github.com/SixLabors/ImageSharp/blob/e484a405dfe8376a55f68cba96129c8c5522c0ff/src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.Generated.tt#L61

It looks like I could simplify the code here and remove the overhead of generating the blend result to pass to the Porter-Duff function.

vpenades: Yes, the code generation can be optimized even further to avoid calling the blend when it's not needed.

At the time I wrote that code I didn't optimize it further because the code was becoming complicated enough, and because I wanted to leave a sense of completeness, in case we wanted to add more composition modes.

Funny thing is, the official [W3 composition specification](https://www.w3.org/TR/compositing-1/#security) states that all composition modes should have the same CPU cost to avoid a timing attack.

In a software implementation it would mean making all composition modes as expensive as the most expensive mode, which is certainly not practical for most use cases.

JimBobSquarePants: Perfect! I've done a bit of cleanup then to remove the unnecessary calculation.

I read the specification you listed and I think we're ok. Your implementation is pixel agnostic anyway.

Thanks again for your help, very much appreciated! 👍

src/ImageSharp/PixelFormats/PixelBlender{TPixel}.cs
antonfirsov: The `TPixel != TPixelSrc` remains untested, but in this case the compiler should guarantee correctness in theory. ReSharper's `MemberCanBePrivate` inspections make me always suspicious though.

JimBobSquarePants:
@antonfirsov Will review your comments tomorrow, it's late now. Still finding my way around this part of the API so I'll probably miss a few things.

antonfirsov:
> Is PixelOperations<T> indeed the best place to hold them? In other words: is it an SRP violation we accept and want to live with?

It means that the scope of the `PixelOperations<TPixel>` class has not yet been explicitly clarified. Stuff just "happens" to be here.
- Is it a class where we only want to expose methods executing **pixel operations** (eg. conversions) only?
- Or is it a facade for different API-s do stuff specific to a known `TPixel`? (Run operations, query pixel blenders or do other stuff.)
In my opinion these are different concerns. Not sure if it is a good idea to mix them.

Publishing API-s is hard ...

JimBobSquarePants:
@antonfirsov I see your concern but am not sure how we could split the two areas of functionality. 

Currently `PixelOperations` is initialized via the `IPixel.CreatePixelOperations()` method which allows us to specify pixel specific optimizations.  If we split pixel blending functionality we lose the ability to create `TPixel` optimal versions of those blenders unless we created a duplicate initialization method in `IPixel`. 

I'm very tempted to leave the structure as-is.

JimBobSquarePants:
>Do we need the virtual keyword? There is no POC for overriding default implementations, which is a must-have when defining extensible API-s in my opinion. Adding the virtual keyword later is not a breaking change

@antonfirsov  The answer to the same question from [Eric Lippart](https://stackoverflow.com/a/49908304/427899) suggests otherwise. I'd rather not introduce the risk should we ever create optimized versions of the blenders.

src/ImageSharp/PixelFormats/PixelBlender{TPixel}.cs
JimBobSquarePants: I think Resharper is wrong there. We use the very same pattern with our `CloneAs<TPixel2>` APIs

antonfirsov:
@JimBobSquarePants you're right on virtual. I've also found a [rule and an explanation](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#members) in the official BCL coding guidelines.

Regarding `PixelOperations<TPixel>`:
Generally, I tend to agree with you. If our decision is explicit, the public API is under control.

JimBobSquarePants:
Yay! One step closer to splitting out drawing!

Issue #1040, opened 2019-10-25
Gradient with opacity not working...


equinox2k:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Gradient's with alpha not working as expected

### Steps to Reproduce

I have created an example that shows gradient not working as expected.. it should do a iOS style gradient overlay, however produces a semi transparent overlay but of a single color where alpha doesnt change...

```
using SixLabors.ImageSharp;
using SixLabors.ImageSharp.PixelFormats;
using SixLabors.ImageSharp.Processing;
using SixLabors.Primitives;
using SixLabors.Shapes;

namespace EQL.ImageProcessors
{
    public static class GlossProcessor
    {
        public static IImageProcessingContext ApplyGloss(this IImageProcessingContext ctx)
        {
            Size size = ctx.GetCurrentSize();
            IPathCollection glossPath = BuildGloss(size.Width, size.Height);
            var graphicsOptions = new GraphicsOptions(true)
            {
                ColorBlendingMode = PixelColorBlendingMode.Normal,
                AlphaCompositionMode = PixelAlphaCompositionMode.SrcAtop
            };
            var linearGradientBrush = new LinearGradientBrush(new Point(0, 0), new Point(0, size.Height / 2), GradientRepetitionMode.Repeat, new ColorStop(0, Color.White.WithAlpha(0.5f)), new ColorStop(1, Color.White.WithAlpha(0.25f)));
            return ctx.Fill(graphicsOptions, linearGradientBrush, glossPath);
        }

        private static IPathCollection BuildGloss(int imageWidth, int imageHeight)
        {
            var pathBuilder = new PathBuilder();
            pathBuilder.AddLine(new PointF(0, 0), new PointF(imageWidth, 0));
            pathBuilder.AddLine(new PointF(imageWidth, 0), new PointF(imageWidth, imageHeight * 0.4f));
            pathBuilder.AddBezier(new PointF(imageWidth, imageHeight * 0.4f), new PointF(imageWidth / 2, imageHeight * 0.6f), new PointF(0, imageHeight * 0.4f));
[...]

PR #1043, opened 2019-11-01
#1040 - Alpha blending incorrect for Gradients with relevant alpha component: Test and first fix

jongleur1983:
### Prerequisites

### Description
Gradients did not properly morph the alpha channel when that was different between start and end.
This was reported by @equinox2k in #1040.

There may be a matching PorterDuff-Function that does this correctly, but I didn't find one. Therefore this PR instead re-calculates the Alpha-Value after PorterDuff has been applied.

I checked the unit test myself, the reference output image must be added as well.

### DIFF
--- a/src/ImageSharp.Drawing/Processing/GradientBrush.cs
+++ b/src/ImageSharp.Drawing/Processing/GradientBrush.cs
@@ -125,11 +125,12 @@ namespace SixLabors.ImageSharp.Processing
                         var toAsVector = to.Color.ToVector4();
                         float onLocalGradient = (positionOnCompleteGradient - from.Ratio) / (to.Ratio - from.Ratio);
 
-                        // TODO: this should be changeble for different gradienting functions
                         Vector4 result = PorterDuffFunctions.NormalSrcOver(
                             fromAsVector,
                             toAsVector,
                             onLocalGradient);
+
+                        result.W = fromAsVector.W + ((toAsVector.W - fromAsVector.W) * onLocalGradient);
 
                         TPixel resultColor = default;
                         resultColor.FromVector4(result);
--- a/tests/ImageSharp.Tests/Drawing/FillLinearGradientBrushTests.cs
+++ b/tests/ImageSharp.Tests/Drawing/FillLinearGradientBrushTests.cs
@@ -8,6 +8,8 @@ using System.Text;
 
 using SixLabors.ImageSharp.PixelFormats;
 using SixLabors.ImageSharp.Processing;
+using SixLabors.Primitives;
+using SixLabors.Shapes;
 
 using Xunit;
 // ReSharper disable InconsistentNaming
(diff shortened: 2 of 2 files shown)


src/ImageSharp.Drawing/Processing/GradientBrush.cs
```diff
@@ -125,12 +125,13 @@ internal abstract class GradientBrushApplicator<TPixel> : BrushApplicator<TPixel
                         var toAsVector = to.Color.ToVector4();
                         float onLocalGradient = (positionOnCompleteGradient - from.Ratio) / (to.Ratio - from.Ratio);
 
-                        // TODO: this should be changeble for different gradienting functions
                         Vector4 result = PorterDuffFunctions.NormalSrcOver(
                             fromAsVector,
                             toAsVector,
                             onLocalGradient);
 
+                        result.W = fromAsVector.W + ((toAsVector.W - fromAsVector.W) * onLocalGradient);
```
JimBobSquarePants: Can you please explain to me what is happening here? I'm struggling to follow. 

The method you are ultimately calling is:

https://github.com/SixLabors/ImageSharp/blob/41c76515f7e8aeda9a2e7c4a1efa52fd9a3f5ac6/src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs#L159

Where `source` and `blend` are the `toAsVector` parameter with the `W` property multiplied by `onLocalGradient`.

jongleur1983: Sure.
I was not able figure out if there's a better matching PorterDuffFunction available...
I guess you're right and it's just Over() with the right parameters.

Looked for something like that but gave up finally  and added the comment to the PR (see above).
So I'll check that later (hope to get to it in the evening).

JimBobSquarePants: I'm a noob with Porter-Duff but I've just found the best explanation of the various equations I've ever seen.

https://www.amanithvg.com/docs/tut/007-blending.html

JimBobSquarePants: Also the old spec docs are waaaaaay better than the new ones.

https://www.w3.org/TR/2009/WD-SVGCompositing-20090430/

jongleur1983: That's not working as expected.
tried what you proposed here locally now:
```csharp
var fromAsVector = from.Color.ToVector4();
var toWithScaledW = to.Color.ToVector4();
float onLocalGradient = (positionOnCompleteGradient - from.Ratio) / (to.Ratio - from.Ratio);
toWithScaledW.W *= onLocalGradient;

Vector4 result = PorterDuffFunctions.Over(fromAsVector, toWithScaledW, toWithScaledW);

TPixel resultColor = default;
resultColor.FromVector4(result);
return resultColor;
```

but got the following result which is not the correct one:
GradientsWithTransparencyOnExistingBackground_Rgba32_Blank200x200

Looks like it's exactly the opposite "direction" of the alpha component than expected.
I checked back with Inkscape, using a red rectangle and a linear gradient on top of that and that agrees with @equinox2k on the expected output.

Have to go for now, will have a look again later.

JimBobSquarePants: Ok.... Not got a solution but looking at `GradientBrush` class I see an issue that should be fixed. We are not using the passed blender.

JimBobSquarePants: Ok Now that I've wired up the correct blender I can see the issue more clearly.

Here's how the alpha component of the returned image is computed in `SrcAtop`. 

https://github.com/SixLabors/ImageSharp/blob/41c76515f7e8aeda9a2e7c4a1efa52fd9a3f5ac6/src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs#L194

I would say that this is correct since overlaying a semi-transparent pixel atop a background should preserve the background opacity.

So what's the fix? Should we adjust the result alpha component following blending or should we adjust the destination alpha component before?

JimBobSquarePants: GradientsWithTransparencyOnExistingBackground_Rgba32_Blank1024x1024

@jongleur1983  So, been putting some more thought to it as I figure my way through the code. 

You shouldn't be using a blender when calculating the gradient here. Lerping between the color stops based on the `onLocalGradient` would, imo, be the correct way to do it.

Here's the current output, based upon that approach (Combined with refactoring I'm doing for #1046. 

Now the funny thing.

@equinox2k By my understanding your reference image is incorrect. If you look at the pixel data on the left hand side of my diff you will see a red component value of 245 - Given that the red component of both the background and the source is 255, this should not be possible.

image

Here's the image with the second color stop updated to `.1F` instead of `.25F`. This looks pretty good to me.

GradientsWithTransparencyOnExistingBackground_Rgba32_Blank1024x1024


.

equinox2k: The expected result image was just something I knocked up in photoshop quickly to try and show what I expected so likely to have some differences than real output. 

Agree the output looks great “SHIP IT” :)

JimBobSquarePants: 😄 See https://github.com/SixLabors/ImageSharp/pull/1046/commits/ff023b833e3dee44c6bd01b6b17099afc397d435 (_Ignore wrong title in the PR commit, I'm not going to rebase and force push to fix_).

@jongleur1983 If you agree with my findings please let me know and we'll close this PR and wait on mine to be completed.

jongleur1983: @JimBobSquarePants agreeing with your changes (well, I just looked at the GradientBrush class of that PR, so don't count it as a complete PR review.

JimBobSquarePants:
@jongleur1983 Closing this since I fixed the issue with #1046 Thanks for your help!

Issue #1002, opened 2019-09-07
Externalize ImageSharp.Drawing


antonfirsov:
Prerequisite: #967

----

**UPDATE**

### Final repository & package structure
![image](https://user-images.githubusercontent.com/6835152/68545258-97e78b80-03cb-11ea-87fc-864be7939dc7.png)

### Refactor steps
I think the best would be to follow the order I suggest. Please let me know if something is missing!

- [ ] Copy all contents of *SixLabors.Core* into the *SixLabors.ImageSharp* repository (test and product code)
- [x] Create a new, 100% self-contained repository for ImageSharp.Drawing, copy the following content into the repository:
    - Current Drawing code:
        - Product code
        - Test code
        - Benchmark code (!!!)
    - Test images used by Drawing tests
    - Reference images used by Drawing tests
    - Product and test code of `SixLabors.Shapes`
- [x] Configure CI for the ImageSharp.Drawing repository
- [ ] Update *ImageSharp.Web* (dogfeeding)

The steps above will also touch #868. If we want to completely resolve that issue, we shall also:
- [ ] Clean-up history of binary artifacts in *SixLabors.ImageSharp* master
- [ ] Copy all reference images to *SixLabors.ImageSharp*

We shall consider introducing GitHub workflows while implementing these steps, but I need more help if we want to do so.

antonfirsov:
I have two competing proposals for reorganizing our GitHub repositories. (Ignoring ImageSharp.Web, since it will be kept as-is.)

### A, 2 Repositories 2 NuGet packages

- SixLabors.ImageSharp
  - Fully internalizing SixLabors.Core
  - Excluding Drawing

- New repository SixLabors.ImageSharp.Drawing
  - Including SixLabors.Fonts, SixLabors.Shapes and Drawing
  - Continue current line of NuGet packages (Next version: beta0008)

### B, 2 Repositories 5 NuGet packages

- Repo: SixLabors.ImageSharp
  - Including SixLabors.Core
    - But define a separate package for it
    - Versions are bound together (Next version: RC1)
  - Excluding Drawing

- Repo: SixLabors.ImageSharp.Drawing
  - Including SixLabors.Fonts, SixLabors.Shapes and Drawing
  - But keep separate NuGet packages
  - Versions are bound together. (Next version: beta0010, also for Drawing)

/cc @tocsoft @dlemstra @JimBobSquarePants

JimBobSquarePants:
[...]

PR #1046, opened 2019-11-05
Expose final internals to allow decoupling Drawing.

JimBobSquarePants:
### Prerequisites

### Description
Fixes accessibility for any remaining blockers that are stopping us decoupling ImageSharp.Drawing

TODO:

- [x] `'ImageFrame.Configuration' is inaccessible due to its protection level`.
- [x] `'float' does not contain a definition for 'Clamp' and no accessible extension method 'Clamp' accepting a first argument of type 'float'`.
- [x] `'Image<TPixel>' does not contain a definition for 'GetMemoryAllocator' and no accessible extension method 'GetMemoryAllocator' accepting a first argument of type 'Image<TPixel>' could be found`
- [x] `'MemoryAllocator' does not contain a definition for 'Allocate2D' and no accessible extension method 'Allocate2D' accepting a first argument of type 'MemoryAllocator' could be found`.
- [x] `'GraphicsOptions.IsOpaqueColorWithoutBlending(Color)' is inaccessible due to its protection level`

I consider this issue the highest priority for now and would really appreciate any assistance that can be given determining the best fix for each outstanding issue.

### DIFF
--- a/src/ImageSharp.Drawing/Common/Extensions/GraphicsOptionsExtensions.cs
+++ b/src/ImageSharp.Drawing/Common/Extensions/GraphicsOptionsExtensions.cs
@@ -1,0 +1,53 @@
+// Copyright (c) Six Labors and contributors.
+// Licensed under the Apache License, Version 2.0.
+
+using System.Numerics;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp
+{
+    /// <summary>
+    /// Extensions methods fpor the <see cref="GraphicsOptions"/> class.
+    /// </summary>
+    internal static class GraphicsOptionsExtensions
+    {
+        /// <summary>
+        /// Evaluates if a given SOURCE color can completely replace a BACKDROP color given the current blending and composition settings.
+        /// </summary>
+        /// <param name="options">The graphics options.</param>
+        /// <param name="color">The source color.</param>
+        /// <returns>true if the color can be considered opaque</returns>
+        /// <remarks>
+        /// Blending and composition is an expensive operation, in some cases, like
+        /// filling with a solid color, the blending can be avoided by a plain color replacement.
+        /// This method can be useful for such processors to select the fast path.
+        /// </remarks>
+        public static bool IsOpaqueColorWithoutBlending(this GraphicsOptions options, Color color)
+        {
+            if (options.ColorBlendingMode != PixelColorBlendingMode.Normal)
+            {
+                return false;
+            }
+
+            if (options.AlphaCompositionMode != PixelAlphaCompositionMode.SrcOver
+                && options.AlphaCompositionMode != PixelAlphaCompositionMode.Src)
+            {
+                return false;
+            }
+
+            const float Opaque = 1F;
+
+            if (options.BlendPercentage != Opaque)
+            {
+                return false;
+            }
+
+            if (((Vector4)color).W != Opaque)
+            {
+                return false;
+            }
+
+            return true;
+        }
+    }
+}
--- a/src/ImageSharp.Drawing/Processing/GradientBrush.cs
+++ b/src/ImageSharp.Drawing/Processing/GradientBrush.cs
@@ -37,9 +37,9 @@ namespace SixLabors.ImageSharp.Processing
         /// <inheritdoc />
         public abstract BrushApplicator<TPixel> CreateApplicator<TPixel>(
             Configuration configuration,
+            GraphicsOptions options,
             ImageFrame<TPixel> source,
-            RectangleF region,
-            GraphicsOptions options)
+            RectangleF region)
             where TPixel : struct, IPixel<TPixel>;
 
         /// <summary>
@@ -58,17 +58,17 @@ namespace SixLabors.ImageSharp.Processing
             /// Initializes a new instance of the <see cref="GradientBrushApplicator{TPixel}"/> class.
             /// </summary>
             /// <param name="configuration">The configuration instance to use when performing operations.</param>
+            /// <param name="options">The graphics options.</param>
             /// <param name="target">The target image.</param>
             /// <param name="colorStops">An array of color stops sorted by their position.</param>
             /// <param name="repetitionMode">Defines if and how the gradient should be repeated.</param>
-            /// <param name="options">The graphics options.</param>
             protected GradientBrushApplicator(
                 Configuration configuration,
+                GraphicsOptions options,
                 ImageFrame<TPixel> target,
                 ColorStop[] colorStops,
-                GradientRepetitionMode repetitionMode,
-                GraphicsOptions options)
-                : base(configuration, target, options)
+                GradientRepetitionMode repetitionMode)
+                : base(configuration, options, target)
             {
                 this.colorStops = colorStops; // TODO: requires colorStops to be sorted by position - should that be checked?
                 this.repetitionMode = repetitionMode;
--- a/src/ImageSharp.Drawing/Processing/GradientBrush.cs
+++ b/src/ImageSharp.Drawing/Processing/GradientBrush.cs
@@ -118,9 +118,6 @@ namespace SixLabors.ImageSharp.Processing
                     else
                     {
                         float onLocalGradient = (positionOnCompleteGradient - from.Ratio) / (to.Ratio - from.Ratio);
-
-                        // TODO: this should be changeble for different gradienting functions.
-                        // TODO: Is the comment above still relevent?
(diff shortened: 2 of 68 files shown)


src/ImageSharp.Drawing/Common/Extensions/GraphicsOptionsExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// </summary>
+        /// <param name="options">The graphics options.</param>
+        /// <param name="color">The source color.</param>
+        /// <returns>true if the color can be considered opaque</returns>
+        /// <remarks>
+        /// Blending and composition is an expensive operation, in some cases, like
+        /// filling with a solid color, the blending can be avoided by a plain color replacement.
+        /// This method can be useful for such processors to select the fast path.
+        /// </remarks>
+        public static bool IsOpaqueColorWithoutBlending(this GraphicsOptions options, Color color)
```
JimBobSquarePants: Why are `TextGraphicsOptions` and `GraphicsOptions` structs? They're both mutable and likely cause unnecessary copying. This extension method is actually copying the struct which cannot be avoided just now.

JimBobSquarePants:
Ok. I'm managed workarounds for everything but the `MemoryAllocator` limitations. @antonfirsov I'll need your help there.

JimBobSquarePants:
I've temporarily enabled internals to test the build cross platform. To test and develop locally comment out the following line in the `ImageSharp.csproj`

``` xml
<InternalsVisibleTo Include="SixLabors.ImageSharp.Drawing" />
```

JimBobSquarePants:
@antonfirsov Is there any reason you can think of why we shouldn't expose the first two extension methods here?

https://github.com/SixLabors/ImageSharp/blob/41c76515f7e8aeda9a2e7c4a1efa52fd9a3f5ac6/src/ImageSharp/Memory/MemoryAllocatorExtensions.cs#L14

src/ImageSharp.Drawing/Common/Extensions/GraphicsOptionsExtensions.cs
JimBobSquarePants: I've converted them to classes.

src/ImageSharp.Drawing/Processing/GradientBrush.cs
```diff
@@ -121,19 +117,8 @@ internal abstract class GradientBrushApplicator<TPixel> : BrushApplicator<TPixel
...
-                        // TODO: this should be changeble for different gradienting functions
-                        Vector4 result = PorterDuffFunctions.NormalSrcOver(
-                            fromAsVector,
-                            toAsVector,
-                            onLocalGradient);
-
-                        TPixel resultColor = default;
-                        resultColor.FromVector4(result);
-                        return resultColor;
+                        return new Color(Vector4.Lerp((Vector4)from.Color, (Vector4)to.Color, onLocalGradient)).ToPixel<TPixel>();
```
JimBobSquarePants: This fixes #1040

src/ImageSharp.Drawing/Utils/NumberUtils.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace SixLabors.ImageSharp
+{
+    /// <summary>
+    /// Utility methods for numeric primitives.
+    /// </summary>
+    internal static class NumberUtils
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static float ClampFloat(float value, float min, float max)
```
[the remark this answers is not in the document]
JimBobSquarePants: We actually covered that ages ago 😄 

https://github.com/SixLabors/ImageSharp/blob/41c76515f7e8aeda9a2e7c4a1efa52fd9a3f5ac6/tests/ImageSharp.Benchmarks/General/BasicMath/ClampFloat.cs#L8

src/ImageSharp.Drawing/Processing/RadialGradientBrush.cs
```diff
@@ -35,16 +35,18 @@ public sealed class RadialGradientBrush : GradientBrush
...
             new RadialGradientBrushApplicator<TPixel>(
+                configuration,
                 source,
-                options,
                 this.center,
                 this.radius,
                 this.ColorStops,
-                this.RepetitionMode);
+                this.RepetitionMode,
+                options);
```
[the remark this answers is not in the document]
JimBobSquarePants: It was all over the place before I refactored. Happy to move it.

JimBobSquarePants:
Thanks for the fast review @antonfirsov I'm keen to move forward with this.

> GraphicsOptions.Default and TextGraphicsOptions.Default are exposing an arbitrary shared mutable state. It was not the case when those were struct-s, and I think we should avoid this, and replace the singletons with default constructor calls.
> 1. The classes need a deep copy logic, just like Configuration
> 2. The classes are showing the symptoms of the Telescoping Constructor anti-pattern. These constructors are unnecessary, since the properties are settable.

Agree with all of the above. That won't take me long.

>Additionally, we should remove <InternalsVisibleTo Include="SixLabors.ImageSharp.Drawing" /> from ImageSharp.csproj to ensure this is indeed the final step to close #1002.

Yeah, good point, will do.

src/ImageSharp.Drawing/Utils/NumberUtils.cs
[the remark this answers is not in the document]
JimBobSquarePants: I'm glad the current `Clamp()` extensions methods are internal actually. I'd like to make them explicit like this in the future, much better readability.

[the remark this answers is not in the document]
antonfirsov: That benchmark is actually using `Math.Clamp()`, which does a conversion internally (for what reason ?? 😮), which explains it entirely. `MathF` has no `Clamp` method actually.

JimBobSquarePants:
@antonfirsov I have updated everything except the csproj `InternalsVisibleTo` declaration as the shared infrastructor classes used in Drawing rely on that call until we can move the library to a separate repository.

src/ImageSharp.Drawing/Processing/TextGraphicsOptions.cs
```diff
@@ -9,120 +10,151 @@ namespace SixLabors.ImageSharp.Processing
...
-
-            this.antialiasSubpixelDepth = 16;
-            this.ColorBlendingMode = PixelColorBlendingMode.Normal;
-            this.AlphaCompositionMode = PixelAlphaCompositionMode.SrcOver;
-            this.blendPercentage = 1;
-            this.antialias = enableAntialiasing;
-            this.dpiX = DefaultTextDpi;
-            this.dpiY = DefaultTextDpi;
-        }
+        public static TextGraphicsOptions Default { get; } = new TextGraphicsOptions();
```
antonfirsov: The singleton is still there. It's completely unnecessary IMO. Shared mutable state that is quite arbitrary. We can replace it with `new TextGraphicsOptions()` on all call sites. Same for `GraphicsOptions`.

tests/ImageSharp.Tests/Drawing/Text/TextGraphicsOptionsTests.cs
```diff
@@ -9,25 +11,139 @@ namespace SixLabors.ImageSharp.Tests.Drawing.Text
 {
     public class TextGraphicsOptionsTests
     {
+        private readonly TextGraphicsOptions newTextGraphicsOptions = new TextGraphicsOptions();
+        private readonly TextGraphicsOptions defaultTextGraphicsOptions = TextGraphicsOptions.Default;
+        private readonly TextGraphicsOptions cloneTextGraphicsOptions = TextGraphicsOptions.Default.Clone();
```
antonfirsov: We need proper tests for `Clone()`, which are actually altering the state of the instance before cloning it.

JimBobSquarePants: Done.

src/ImageSharp.Drawing/Processing/TextGraphicsOptions.cs
JimBobSquarePants: Ah, didn't get your point previously. Removed.

JimBobSquarePants:
> I still think we shall get rid of those build errors in this step, or a separate PR before actually moving the whole thing out. The Drawing project shall include the shared infrastructure independently.

Feel free to try. I got an ambiguity errors when I tried.

PR #1048, opened 2019-11-07
Support palettized transparent bitmaps extracted from an .ico 

GyleIverson:
### Prerequisites

### Description

The format of the 1-, 4-, and 8-bit Bmp images embedded in an .ico is subtly different (and undocumented). I came up with some changes to the Bmp decoder to properly process a Bmp image extracted from an .ico file that has transparency.

### Examples
Here are several images that shows the effect of the changes. The image samples Word document shows the original .ico file, the raw bitmap extraction from the .ico, and the fixed image using the updated code.

[Image samples.docx](https://github.com/SixLabors/ImageSharp/files/3818469/Image.samples.docx)


antonfirsov:
@GyleIverson thanks for raising this! Maybe it's just me, but I fail to follow the use case. Can you post the instances of *input*, *expected*, *actual* images directly here, together with a code snippet reproducing the issue, to avoid misunderstandings for 100%?
We'll also need to add test cases based on the images and the code sample.

JimBobSquarePants:
@brianpopow Do you have any insight here? You know more about bitmap than any of us.

brianpopow:
Unfortunately im also not familiar with the ico format @GyleIverson is refering to.
There are OS/2 icon bitmap variations, but that is not what @GyleIverson is talking about.

My guess is, that its about the icon format introduced with windows 1.x, see this http://fileformats.archiveteam.org/wiki/Windows_1.0_Icon
This is a container format which has its own header format and can contain one or many BMP or even PNG's.
If it contains a BMP the bitmap will have no BitmapFileHeader. And this is why i dont understand the use case either.
The BmpDecoder expects a BitmapFileHeader, so it will fail to decode an ico bitmap byte stream. Not to mention that the user would need to parse the ico header by himself and pass the correct byte stream.

I also dont understand why the bitmaps inside the ico format would be different from normal bitmaps.
The guy who made the buitmap suite has done a great job in collecting valid bitmaps [BmpSuite](http://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html)
Everything what he listed in the folder "g" for good, imagesharp already supports. Even some questionable ones.

In my opinion the proper way to support the icon format is to have a new decoder and encoder for it, 
which is able to parse the ico header and reuse the bitmap decoding part we already have.

It would be great if GyleIverson could provide some official specification where the changes to the color palette are explained. I know he mentioned in his introduction that it is kind of "undocumented", but it feels kind of wrong changing the decoder
without any kind of official mention anywhere why that change is necessary.
Also some example ico images would be great.

GyleIverson:
The submitted changes solves a problem discovered while harvesting usable assets from favicon.ico files from websites. 

All of this is the result of several weeks of reverse engineering work. There is no documentation describing these quirks. There are some additional quirks within the .ico file described at the end. 

In the Word document attached to the pull request you will find three each .ico, .bmp, and .png files. The .ico files were downloaded from the respective website’s favicon.ico.

The .bmp files are the 64x64 versions extracted from the .ico file. Each one demonstrates odd visual artifacts. The Apple does not render at all, where the other two show black where transparency should appear.

(Regarding the issue of the .ico file lacking the bitmap headers… It is a simple exercise to prepend the header using information available from the icon directory entry before serializing the asset.)

The .png files demonstrate the results of loading the .bmp image in ImageSharp and saving as it as a .png file. Without the changes provided by the pull request, the resulting .png file looks identical to the respective .bmp file.

Regarding the .ico quirks… The available .ico file format documentation fails to describe some quirks. The height is always twice the width. No idea why. Overwriting the height by the width solves that. Another quirk is the physical length of the last .bmp image in the file may be shorter than that stated in the icon directory entry. Presumably some tools don’t serialize the trailing filler bytes.

If you want the snippets of code related to dealing with extracting of assets from an .ico file let me know. I’d be happy to share that too for someone else to make a .ico decoder.

brianpopow:
@GyleIverson thanks for the explanation. If i open the attached word document, i cannot see the ico files, they seem to be missing. Could you please attach the ico files and the expected output in a zip file.

> Regarding the .ico quirks… The available .ico file format documentation fails to describe some quirks. The height is always twice the width. No idea why. Overwriting the height by the width solves that.

This sounds like your ico files are similar to O2/2 ico files and it is actually described in the documentation why the height doubled. Citing from the specification:

> For icons and pointers, the cy field in bmp is actually twice the
pel height of the image that appears on the screen. This is because
these types actually contain two full bit-map pel definitions. The
first bit-map definition is the XOR mask, which contains invert
information (0 = no invert, 1 = invert) for the pointer or icon.
The second is the AND mask, which determines whether the pointer
or the screen is shown (0 = black/white, 1 = screen/inverse
screen).

Here is the full documentation: [OS2 icons](https://www.fileformat.info/format/os2bmp/spec/902d5c253f2a43ada39c2b81034f27fd/view.htm)

That said, im not sure if your ico files are OS/2 Icons or if they are just similar. If they are OS2 icons, then there are two version of the icon in the file. The second one is a mask which determines the transparency.

GyleIverson:
@brianpopow After pondering my previous writings on this subject, I now postulate that my analysis of several empirical observations are wrong.

Rather than having some quirks in the embedded .bmp and some in the .ico, I believe all of the quirks are in the .ico file definition. 

The small set of changes I made to support 1-bpp and 4-bpp palettized .bmp remain valid. However, all of the color palette changes made to account for 32-bit color values with alpha is wrong. That code belongs in a .ico decoder.

I will update the pull request to just focus on 1-bpp and 4-bpp palettized .bmp images. All of the bgra32 changes will be removed.

As to other questions… I can confirm that the .ico files available on the web use the Windows NT 3.1+ BITMAPINFOHEADER structure. The size of the BITMAPINFOHEADER is 40 bytes and that is reflected in all of the test .ico files. The 2, 16, or 256 brga32 color palette immediately follows the BITMAPINFOHEADER. The pixels follow the color palette. There are no XOR or AND data areas, ala OS/2, in the image.

When I look at the BITMAPINFOHEADER on https://en.wikipedia.org/wiki/BMP_file_format#Bitmap_file_header, I am left wondering why the .ico tools double the height and don’t express any alpha channel. The lack of alpha data is confusing, given the storage is already allocated (but all the alphas are zero). I am going to use the height doubling as a clue that two modifications are needed to the .bmp image before serializing and passing it along to ImageSharp. First, adjust the height. Second, any color palette entry with a non-zero bgr value will get a saturated alpha value.

After making those changes to my (not ready for prime-time) .ico decoder, I’ll share that code with you.

JimBobSquarePants:
We're going to close this now as it's been sitting around and we don't have the resources to keep it up to date with the main branch. 

Feel free to reopen if you want to continue working on this and have capacity to add tests 👍 .

PR #1050, opened 2019-11-11
Public Vector4 to/from conversions

Sergio0694:
### Prerequisites

### Description
As discussed with @antonfirsov, I've made the following APIs public:
- `PixelOperations<TPixel>.FromVector4Destructive`
- `PixelOperations<TPixel>.ToVector4`

Overloads are included as well. I also made the `PixelConversionModifiers` enum public, otherwise two of those overloads would've been blocked from being exposed.


src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
```diff
@@ -83,7 +83,7 @@ public partial class PixelOperations<TPixel>
         /// <param name="configuration">A <see cref="Configuration"/> to configure internal operations</param>
         /// <param name="sourcePixels">The <see cref="Span{T}"/> to the source colors.</param>
         /// <param name="destVectors">The <see cref="Span{T}"/> to the destination vectors.</param>
-        internal virtual void ToVector4(
+        public virtual void ToVector4(
```
antonfirsov: This method should not be virtual.

Sergio0694:
Hey @antonfirsov - I'm not sure why the build failed there, though if you check the CI you'll see that it's just a seemingly unrelated error for .NET Framework 64 bit, with some luck it might just be one of those random errors that pop up every now and then during automated builds.
Anyway, I'll see if I can remove that `virtual` modifier in the meantime as you suggested 👍

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
Sergio0694: Wait, this method is actually overridden by some pixel-format-specific implementations of `PixelOperations<TPixel>`, like for [`RgbaVector`](https://github.com/Sergio0694/ImageSharp/blob/e98c0ff46bc8fd21a91f74b61bc28484a6f2c0b3/src/ImageSharp/PixelFormats/PixelImplementations/RgbaVector.PixelOperations.cs#L37), [`Rgba32`](https://github.com/Sergio0694/ImageSharp/blob/e98c0ff46bc8fd21a91f74b61bc28484a6f2c0b3/src/ImageSharp/PixelFormats/PixelImplementations/Rgba32.PixelOperations.cs#L23) and others. It's also in the `GenerateRgba32CompatibleVector4ConversionMethods` part in the `_Common.ttinclide` file, [here](https://github.com/Sergio0694/ImageSharp/blob/e98c0ff46bc8fd21a91f74b61bc28484a6f2c0b3/src/ImageSharp/PixelFormats/PixelImplementations/Generated/_Common.ttinclude#L139).

I don't think we can remove that `virtual` right now, given the rest of the codebase as it is now. That is, unless you have some refactoring in mind to make that possible 🤔

JimBobSquarePants:
@Sergio0694 What's with the commit history from July?

Sergio0694:
@JimBobSquarePants I think that's because I created this branch from my existing fork of `ImageSharp`, which hadn't been updated since my bokeh blur processor got merged, so I think those commits are from the initial merge I did to get my fork back in sync with the current official `master`.

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
JimBobSquarePants: Yeah, we're gonna have to have a think about this!

JimBobSquarePants:
@Sergio0694 Still haven't managed to properly sit down and evaluate this. Rest assured though I will do asap.

Sergio0694:
@JimBobSquarePants Hey, I was just about to ask you for a status update on this! Glad to hear it's still on your radar, that's awesome! And no worries, there's no rush, I just wanted to make sure this would make it into the next beta/RC build 😄

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
antonfirsov: @Sergio0694 I did not have the chance to check the code, but if that method is overriden, it's a mistake. All scenarios shall be covered with a single virtual overload which is taking the most arguments, the shorter overloads shall be non-virtual wrappers only.

Otherwise everything else LGTM, if someone could confirm that T4 generation is  working.

Sergio0694: @antonfirsov So, I'm not sure how to proceed here. From what I see, even before my changes in this branch, that `virtual ToVector4` method was already being overridden in virtually all the pixel-format specific `PixelOperations<TPixel>` classes.
For instance, this happens both in the explicit [`RgbaVector`](https://github.com/Sergio0694/ImageSharp/blob/4427e2467a3f46e34b9059c99304266aea17786b/src/ImageSharp/PixelFormats/PixelImplementations/RgbaVector.PixelOperations.cs#L37) and [`Rgba32`](https://github.com/Sergio0694/ImageSharp/blob/4427e2467a3f46e34b9059c99304266aea17786b/src/ImageSharp/PixelFormats/PixelImplementations/Rgba32.PixelOperations.cs#L23) types, as well as in all the generated pixel-format classes (eg. here in [`Argb32.Generated`](https://github.com/Sergio0694/ImageSharp/blob/4427e2467a3f46e34b9059c99304266aea17786b/src/ImageSharp/PixelFormats/PixelImplementations/Generated/Argb32.PixelOperations.Generated.cs#L51)).

I'm not sure how to proceed here if you want to remove the `virtual` modifier: as far as I can see, right now each pixel-format is [returning an instance of its own `PixelOperations<TPixel>` private class](https://github.com/Sergio0694/ImageSharp/blob/4427e2467a3f46e34b9059c99304266aea17786b/src/ImageSharp/PixelFormats/PixelImplementations/Rgba32.cs#L258), exposed as `PixelOperations<TPixel>`, so that when the methods are called on that instance, the right overload for the specific pixel-format is called. And I mean, this makes sense, and it looks great. My question is: if you want to remove the `virtual` modifier from the default implementation of those methods in the base `PixelOperations<TPixel>` class, how can we then have each pixel-format call its own custom version of those APIs?

`PixelOperations<TPixel>` is not `abstract` so we can't even just set that method as `abstract` from there and have each pixel-format specific class implement it. Plus as far as I know there are some pixel-formats that do fallback on that base implementation, so this wouldn't work either.

I'm not sure how to proceed, I might very well be missing something here, I'm confused 🤔

JimBobSquarePants: @Sergio0694 I think you were looking at the wrong method. The one highlighted by @antonfirsov  does not have any overloads. I've removed the `virtual keyword` from it and regenerated the templated code.

Once this builds I'm happy to merge it.

Sergio0694: @JimBobSquarePants @antonfirsov Ooof I was looking at the wrong overload 🤦‍♂️
I did have the feeling I was being stupid, sorry to have wasted your time on this!
Happy to hear the rest of the PR was ok though, thanks for merging it! 😊

JimBobSquarePants: Haha no worries, easy  mistake to make!

PR #1058, opened 2019-11-27
Bokeh blur speedup and memory usage improvement

Sergio0694:
### Prerequisites

### Description

This PR includes a refactoring of the bokeh blur effect, with a complete removal of the two "low memory usage" and "maximum performance" options, and a single implementation that is both faster than the previous "maximum performance", and uses less memory than the previous "low memory usage" mode. Basically, the best of both worlds, and even better than that! 🚀

Here's a benchmark with the 3 different modes. The images are of size 720, 1080 and 2160.

| ImageSize |     Method |       Mean |    Error |   StdDev |    Allocated |
|---------- |-----------:|-----------:|---------:|---------:|-------------:|
|     Small | Low memory |   356.8 ms |  6.98 ms | 10.87 ms |    124.59 KB |
|     Small |  Max speed |   165.1 ms |  1.43 ms |  1.20 ms |    115.46 KB |
|     Small |   **Refactor** |   **165.2 ms** |  2.13 ms |  1.89 ms |    **105.88 KB** |
|           |            |            |          |          |              |
|    Medium | Low memory | 1,355.9 ms | 27.03 ms | 45.16 ms |  66921.5 KB |
|    Medium |  Max speed |   656.0 ms |  3.00 ms |  2.34 ms |  129792.2 KB |
|    Medium |   **Refactor** |   **667.0 ms** |  9.13 ms |  7.62 ms |  **64982.3 KB** |
|           |            |            |          |          |              |
|     Large | Low memory | 5,029.0 ms | 70.54 ms | 65.99 ms | 392994.26 KB |
|     Large |  Max speed | 2,765.6 ms | 84.41 ms | 70.49 ms | 648345.34 KB |
|     Large |   **Refactor** | **2,684.7 ms** | 14.81 ms | 12.36 ms | **389135.74 KB** |

In particular:

- The new implementation completely removes one buffer of `ComplexVector4` values, so that's a reduction of `sizeof(Vector4) 2 * imageHeight * imageWidth` compared to the "maximum performance" mode, and a reduction of `sizeof(Vector4) * 2 * kernelRadius * imageWidth` compared to the previous "low memory usage" mode. This is achieved by modifying the vertical convolution step so that the partial components of the fourier transformation for the bokeh kernels are directly added to the target processing buffer, which makes is possible to only use a temporary buffer for the results of the first horizontal 1D convolution pass.
- The speed improvements are achieved by the fact that the processing partials are being computed right after each vertical convolution pass, so an entire `Parallel.For` invocation on the entire image is not needed anymore. Additionally, this new implementation can always execute the code in parallel, unlike the previous "low memory usage" mode which was forced to execute the second 1D convolution sequentially, to be able to work on a shared buffer.

@JimBobSquarePants @antonfirsov I think you'll be pretty happy with this new PR 😊
Also, please make sure that everything checks out on your end (all the checks are passing for me, so this should be easy to do) and merge this as soon as possible, since it includes a breaking change in a couple of exposed APIs, so it'd be better to do this while the lib is still in beta.


Sergio0694:
Wo-hooo! 😄🍻

PR #1061, opened 2019-12-11
enable using github actions for build

tocsoft:
### Prerequisites

### Description
Adds a github workflow for building PRs using github actions and also build master branch on tags and merges.

On merges/tags packages are built and uploaded to github pacakge repository & for the time being myget (that can be phased out in the future are required.)

### Known differences
Package number changes - due to the fact that github doesn't have build numbers we are generating the number from the date/time of the commit being built which should be unique and sequential enough for our use case.

As part of this PR i've also stopped Appveyor from publishing to the myget feed so we don't get a strange mix of packages up there.

### Update

@JimBobSquarePants (me) added additional changes.
- Build and test scripts are project agnostic and only rely on an environmental variable set in the yml for xunit (required for x86 tests).
- Added targets `netcorapp3.1`, `netstandard2.1` and set langversion to 8
- Removed explicit `net462` target from tests and added `netcoreapp3.1`
- Replaced OpenCover with Coverlet
- Replace custom versioning with compatible SemVer generated via GitVersion
- Replaced obsolete ProjectIconUrl in nuspec properties
- Stylecop is now referenced in all projects with reduced strictness in tests. (They're the Wild West of readability and we need a cleanup)
- Cleanup of R# settings files (R# uses .editorconfig now ) and .vscode startup (It looked incorrect and if someone wants it they can PR)
- Cleanup of other obsolete files.


JimBobSquarePants:
I'm super excited about this! We can finally target NET Core 3+

I'll have a proper dig through the config as soon as I can. Only question I have straight away is that I can only see a subset of the original target frameworks from AppVeyor. Is that intentional?

tocsoft:
It was somewhat intentional, removing the net642 targets was intentional they are so old i feel we can just drop version/target now. Totally forgot about x86 netcore and also mono.

JimBobSquarePants:
It was somewhat intentional, removing the net642 targets was intentional they are so old i feel we can just drop version/target now. Totally forgot about x86 netcore and also mono.

I guess I'm ok with that considering the warning regarding NET 4.6x from the [docs](https://docs.microsoft.com/en-us/dotnet/standard/net-standard)

>While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

I would try and get x86 builds in though. Not sure about Mono.

build.ps1
```diff
@@ -1,20 +1,79 @@
...
+# }else{
+#     if (Get-Command "mono" -ErrorAction SilentlyContinue) 
+#     {
+#         Write-Info "Building full framework targets - mono installed"
+#         $skipFullFramework = 'true'
+#     }
+# }
 
 # we are running on the build server 
 $isVersionTag = $env:APPVEYOR_REPO_TAG_NAME -match $tagRegex
```
JimBobSquarePants: Won't this always be false since we're not using Appveyor anymore?

Directory.Build.props
```diff
@@ -14,7 +14,7 @@
   <PropertyGroup>
     <BaseArtifactsPath>$(MSBuildThisFileDirectory)artifacts/</BaseArtifactsPath>
     <BaseArtifactsPathSuffix>$(ImageSharpProjectCategory)/$(MSBuildProjectName)</BaseArtifactsPathSuffix>
-    <RepositoryUrl>https://github.com/SixLabors/ImageSharp/</RepositoryUrl>
+    <RepositoryUrl Condition="'$(RepositoryUrl)' == ''">https://github.com/SixLabors/ImageSharp/</RepositoryUrl>
```
JimBobSquarePants: What does this change do?

tests/ImageSharp.Tests/ImageSharp.Tests.csproj
```diff
@@ -1,8 +1,9 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;net462;net472</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks Condition="$(skipFullFramework) != 'true'">$(TargetFrameworks);net462;net472</TargetFrameworks>
```
JimBobSquarePants: We still have the ` net462` target here.

antonfirsov:
I really want to do a deeper review on this during the weekend.

For now, only one comment:
If we drop test execution on .NET 4.6.2, we should explicitly state that that platform is no longer supported, and bugs reported against that platform will not be fixed.

build.ps1
```diff
@@ -1,20 +1,79 @@
+param(
+    [string]$targetFramework = 'ALL'
+)
 
 # lets calulat the correct version here
 $fallbackVersion = "1.0.0";
 $version = ''
 
-$tagRegex = '^v?(\d+\.\d+\.\d+)(-([a-zA-Z]+)\.?(\d*))?$'
+$tagRegex = '^v?(\d+\.\d+\.\d+)(?:-([a-zA-Z]+)\.?(\d*))?$'
```
JimBobSquarePants: https://github.com/marketplace/actions/use-actions

tocsoft: we used to use gitversion, it ended up breaking our release numbers so we stopped using it and created our own system for calculating the version numbers.

JimBobSquarePants: Cool. Didn't know that. I've been using it at work for a while now without issues so was just wondering

JimBobSquarePants: Got them working well.

.github/workflows/build-and-test.yml.bak
```diff
@@ -0,0 +1,116 @@
+name: Build
```
JimBobSquarePants: I'll delete this.

.gitignore
```diff
@@ -221,3 +221,5 @@ artifacts/
 # Tests
 **/Images/ActualOutput
 **/Images/ReferenceOutput
+/tests/CodeCoverage/opencover.zip
```
JimBobSquarePants: Don't know how that got added. Will remove

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,105 @@
...
+    pull_request:
+        branches:
+            - master
+
+jobs:
+    Build:
+        strategy:
+            matrix:
+                options:
+                    - os: ubuntu-latest
```
brianpopow: shouldn't that be better a specific version of ubuntu? Same with windows below?

tests/Directory.Build.targets
```diff
@@ -16,4 +16,40 @@
...
+    <DotNetCliToolReference Update="dotnet-xunit" Version="2.3.1" />
+  </ItemGroup>
+
+  <!--Code coverage specific settings-->
+  <!--https://github.com/tonerdo/coverlet-->
+  <PropertyGroup Condition="'$(codecov)' == 'true'">
+    <CollectCoverage>true</CollectCoverage>
+    <UseSourceLink>true</UseSourceLink>
+    <CoverletOutputFormat>opencover</CoverletOutputFormat>
+    <!--Ouput injects target framework into name despite explicit config. See build yml-->
```
brianpopow:
```suggestion
    <!--Output injects target framework into name despite explicit config. See build yml-->
```

tests/ImageSharp.Benchmarks/ImageSharp.Benchmarks.csproj
```diff
@@ -5,9 +5,9 @@
     <AssemblyName>ImageSharp.Benchmarks</AssemblyName>
     <OutputType>Exe</OutputType>
     <RootNamespace>SixLabors.ImageSharp.Benchmarks</RootNamespace>
-    <TargetFrameworks>netcoreapp2.1;net472</TargetFrameworks>
-    <IsPackable>false</IsPackable>
+    <TargetFrameworks>netcoreapp3.1;netcoreapp2.1;net472</TargetFrameworks>
     <GenerateProgramFile>false</GenerateProgramFile>
+    <!--Used to hide test project from dotnet test-->
```
brianpopow: is `<IsTestProject>false</IsTestProject>` missing here?

.github/workflows/build-and-test.yml
JimBobSquarePants: latest is fine 
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners

tests/Directory.Build.targets
JimBobSquarePants: Fixed

tests/ImageSharp.Benchmarks/ImageSharp.Benchmarks.csproj
JimBobSquarePants: Fixed

antonfirsov:
There are massive changes that may effect local dev experience. Gonna check the branch tonight wit Rider.

JimBobSquarePants:
You and your toy IDE 😝

It’s a blanket off and on case with stylecop. You have to specify the rules and turn them off individually.

I’ll disable for now in tests and re-add them with another PR fixing them. It would only take me a couple of hours.

brianpopow:
> You and your toy IDE 😝
> 
> It’s a blanket off and on case with stylecop. You have to specify the rules and turn them off individually.
> 
> I’ll disable for now in tests and re-add them with another PR fixing them. It would only take me a couple of hours.

Im not sure if, we should enable stylecop rules in tests at all. Myself i would not care about it, but i had some colleagues in the past who really disliked those rules. I think writing tests is a chore by itself and we should not make it harder by enforcing style rules on them. I think its ok, if its a bit of a wild west in the tests.

JimBobSquarePants:
>Im not sure if, we should enable stylecop rules in tests at all. Myself i would not care about it, but i had some colleagues in the past who really disliked those rules. I think writing tests is a chore by itself and we should not make it harder by enforcing style rules on them. I think its ok, if its a bit of a wild west in the tests.

The rules are much less strict than the main src and we can tweak them as need be. We need something though. The tests are difficult to follow at points which is dangerous.

brianpopow:
> I’ll disable for now in tests and re-add them with another PR fixing them. It would only take me a couple of hours.

ok then, if you want i can do that. I have some free time next week, i would not mind to do that.

antonfirsov:
I think even relaxed ruleset is worth a review to make sure they don't harm readability by eg. enforcing rules on Member Data formatting which are impractical here. (There might be no such case in practice - just an example to emphasize that we should be careful.)

Fortuanately StyleCop's automatic formatters seem to work much better today (at least with Rider). A few years ago, when I had to fix everything manually, it was a terrible & frustrating productivity killer for me. I used to fight until my last blood to keep SC far from the test code 😄

ci-test.ps1
```diff
@@ -0,0 +1,41 @@
+param(
+  [Parameter(Mandatory, Position = 0)]
+  [string]$os,
+  [Parameter(Mandatory, Position = 1)]
+  [string]$targetFramework,
+  [Parameter(Mandatory, Position = 2)]
+  [string]$platform,
```
antonfirsov: nit: this should be named `architecture`

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,105 @@
...
+                      framework: netcoreapp3.1
+                      runtime: -x64
+                      codecov: false
+                    - os: windows-latest
+                      framework: netcoreapp3.1
+                      runtime: -x64
```
antonfirsov: It's worth to also test x86 + 3.1 because all of the floating point magic we do.

ci-test.ps1
JimBobSquarePants: It's platform in msbuild.

antonfirsov: Ok, makes sense here. However, platform is an overloaded term meaning different things in different contexts 😕

PR #1062, opened 2019-12-12
Add La16 and La32 IPixel formats.

JimBobSquarePants:
### Prerequisites

### Description
This PR acts as a precursor to another planned PR building upon the work done adding a non-generic API to the library #907 

<strike>The planned work will change the behavior of the non-generic `Image.Load()` behavior to select the correct pixel format to decode images to rather than defaulting to `Rgba32`.</strike>

<strike>This PR lays the groundwork to allow this by adding missing pixel formats.</strike>

The above is no longer the plan, rather we make it easier for users to make informed pixel format choices by inspecting metadata.

- Introduces `La32` and `La16`, pixel formats representing luminance and alpha at 8-bit and 16-bit per component.
- Renames `Gray8`, `Gray16`, and `Alpha8` to `L8`, `L16`, and `A8` for consistancy
- Adds `PixelOperations` overloads and tests for the new types and the missing one for `Bgra5551`.
- Fixes rounding issues converting from RGB to L.
- Updates `PngDecoder` to use new pixel formats when decoding.
- <strike>Updates the `JpegDecoder` to decode to `Rgb24` or `L8` instead of `Rgba32`.</strike>
- Adds format specific extensions to `ImageMetadata` and `ImageFrameMetadata` to make it easier to return format specific metadata.

I appreciate there's a lot to read but that's the nature of the `IPixel` API.


tocsoft:
Things to consider with returning anything but `Rgba32` is composability, If `Image.Load` sometimes returns `La16` then you can't now draw over the top a 2nd image that might have loaded as `Rgba32` and save it in color as the draw image call will downscale the `Rgba32` image to `La16` during drawing.

JimBobSquarePants:
>Things to consider with returning anything but `Rgba32` is composability, If `Image.Load` sometimes returns `La16` then you can't now draw over the top a 2nd image that might have loaded as `Rgba32` and save it in color as the draw image call will downscale the `Rgba32` image to `La16` during drawing.

Yeah, I had a think about that and I would expect that behavior. The option is always there to use the generic version if need be.

antonfirsov:
A few remarks (did not have the chance to go through the code changes though):
- We should compare both Load and LoadResizeSave benchmark results before/after. (For both YCbCr Jpegs and whatever is considered to be the most common in PNG. `4 components <=> 3 components` mass packing/unpacking can be tricky, can't recall what's happening there. I want to make sure that we are aware of the perf impacts.
- Regarding @tocsoft 's comment: We should probably document the change in the remarks section of the xml docs for `Image.Load()`, and leave a note to self to also do so in the articles when we get the chance.
- Does the PR effect the default PNG/Jpeg pixel format when we resave an image?

JimBobSquarePants:
@antonfirsov @tocsoft 

I'm beginning to think that returning different pixelformat types is a bad idea. It adds complexity to the internal workings of our decoders that I can't enforce.

 I'm going to revert the JpegDecoder changes in this PR but all other changes still stand.

antonfirsov:
> I'm beginning to think that returning different pixelformat types is a bad idea. 

It's worth to do so in long term, because it is an enabler for memory & CPU optimizations **together with the shuffling intrinsics which are not yet utilized** in the bulk conversions. The speedup and the reduced memory footprint for the most common (3 component, alpha-free) jpeg thumbnail making use case should be significant enough to justify the extra complexity. Basically, all major imaging libraries work like this, for a good reason.

Therefore, tying our own hands by communicating in docs that an untyped `Image` is actually an `Image<Rgba32>` is not wise. If users start to rely on it, it would disable the optimization opportunity.

There are many ways to handle the drawing composability issues - most straightforward is docs.

JimBobSquarePants:
>Therefore, tying our own hands by communicating in docs that an untyped Image is actually an Image<Rgba32> is not wise. If users start to rely on it, it would disable the optimization opportunity.

@antonfirsov   Interestingly enough that was actually the case in some areas of the documentation already, it was quite inconsistent. 

I'll revert that documentation change for now and will wait to change the type until after we have those optimizations. (Hopefully something we can begin very soon once we have merged #1061)

JimBobSquarePants:
Just pushed some changes to fix the casing of any Metadata related files. It was doing my head in.

antonfirsov:
> Interestingly enough that was actually the case in some areas of the documentation already, it was quite inconsistent.

That is most likely an accidental leftover for #907.

> When you have a moment I want to talk Core 3+ and a plan to perform the Jpeg IDCT + Colorspace transforms via intrinsics in a manner that matches other high performance decoders.

Okkay, let's jump into it :) I'm prepering an analysis right now. Sorry in advance, but it will be long as hell. I'm trying best to explain things as deeply and clearly as possible.

src/ImageSharp/Formats/IImageDecoder.cs
```diff
@@ -17,17 +17,16 @@ public interface IImageDecoder
...
         /// <param name="configuration">The configuration for the image.</param>
         /// <param name="stream">The <see cref="Stream"/> containing image data.</param>
-        /// <returns>The decoded image of a given pixel type.</returns>
+        /// <returns>The <see cref="Image{TPixel}"/>.</returns>
         Image<TPixel> Decode<TPixel>(Configuration configuration, Stream stream)
             where TPixel : struct, IPixel<TPixel>;
 
         /// <summary>
         /// Decodes the image from the specified stream to an <see cref="Image"/>.
-        /// The decoder is free to choose the pixel type.
```
antonfirsov: Might be worth to keep the warning, see #1064.

src/ImageSharp/Formats/Png/PngScanlineProcessor.cs
```diff
@@ -128,32 +120,24 @@ internal static class PngScanlineProcessor
...
             {
-                Rgba64 rgba64 = default;
                 for (int x = pixelOffset, o = 0; x < header.Width; x += increment, o += 2)
                 {
                     ushort luminance = BinaryPrimitives.ReadUInt16BigEndian(scanlineSpan.Slice(o, 2));
-                    rgba64.R = luminance;
-                    rgba64.G = luminance;
-                    rgba64.B = luminance;
-                    rgba64.A = luminance.Equals(luminance16Trans.PackedValue) ? ushort.MinValue : ushort.MaxValue;
+                    ushort alpha = luminance.Equals(luminance16Trans.PackedValue) ? ushort.MinValue : ushort.MaxValue;
```
antonfirsov: `alpha` is known outside the loop, doing the comparison each time is redundant.

src/ImageSharp/Formats/IImageDecoder.cs
```diff
@@ -17,17 +17,16 @@ public interface IImageDecoder
         /// <typeparam name="TPixel">The pixel format.</typeparam>
         /// <param name="configuration">The configuration for the image.</param>
         /// <param name="stream">The <see cref="Stream"/> containing image data.</param>
-        /// <returns>The decoded image of a given pixel type.</returns>
```
antonfirsov: Might be worth to keep the warning, see #1064.

src/ImageSharp/Formats/Png/PngScanlineProcessor.cs
```diff
@@ -128,32 +120,24 @@ internal static class PngScanlineProcessor
...
 
-                    pixel.FromRgba64(rgba64);
+                    pixel.FromLa32(new La32(luminance, alpha));
```
antonfirsov: Defining a shared `La32` outside the loop, overwriting, it's `Luminance` would be more efficient than `new La32(luminance, alpha)`.

src/ImageSharp/Formats/Png/PngScanlineProcessor.cs
```diff
@@ -109,7 +101,7 @@ internal static class PngScanlineProcessor
                     for (int x = pixelOffset, o = 0; x < header.Width; x += increment, o += 2)
                     {
                         ushort luminance = BinaryPrimitives.ReadUInt16BigEndian(scanlineSpan.Slice(o, 2));
-                        pixel.FromGray16(new Gray16(luminance));
+                        pixel.FromL16(new L16(luminance));
```
antonfirsov: `Unsafe.As<ushort, L16>()` 😉

src/ImageSharp/Formats/Png/PngScanlineProcessor.cs
```diff
@@ -128,32 +120,24 @@ internal static class PngScanlineProcessor
...
                 {
                     byte luminance = (byte)(Unsafe.Add(ref scanlineSpanRef, o) * scaleFactor);
-                    rgba32.R = luminance;
-                    rgba32.G = luminance;
-                    rgba32.B = luminance;
-                    rgba32.A = luminance.Equals(scaledLuminanceTrans) ? byte.MinValue : byte.MaxValue;
+                    byte alpha = luminance.Equals(scaledLuminanceTrans) ? byte.MinValue : byte.MaxValue;
 
-                    pixel.FromRgba32(rgba32);
+                    pixel.FromLa16(new La16(luminance, alpha));
```
antonfirsov: See previous comments for each consequent for loop.

src/ImageSharp/Formats/Tga/MetadataExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// Extension methods for the <see cref="ImageMetadata"/> type.
+    /// </summary>
+    public static partial class MetadataExtensions
+    {
+        /// <summary>
+        /// Gets the tga format specific metadata for the image.
+        /// </summary>
+        /// <param name="metadata">The metadata this method extends.</param>
+        /// <returns>The <see cref="TgaMetadata"/>.</returns>
+        public static TgaMetadata GetTgaMetadata(this ImageMetadata metadata) => metadata.GetFormatMetadata(TgaFormat.Instance);
```
antonfirsov: Coming out of the cave, now I'm seeing a TGA decoder!

src/ImageSharp/Formats/Png/PngScanlineProcessor.cs
```diff
@@ -55,32 +55,24 @@ internal static class PngScanlineProcessor
...
             {
-                Rgba64 rgba64 = default;
                 for (int x = 0, o = 0; x < header.Width; x++, o += 2)
                 {
                     ushort luminance = BinaryPrimitives.ReadUInt16BigEndian(scanlineSpan.Slice(o, 2));
-                    rgba64.R = luminance;
-                    rgba64.G = luminance;
-                    rgba64.B = luminance;
-                    rgba64.A = luminance.Equals(luminance16Trans.PackedValue) ? ushort.MinValue : ushort.MaxValue;
+                    ushort alpha = luminance.Equals(luminance16Trans.PackedValue) ? ushort.MinValue : ushort.MaxValue;
```
antonfirsov: See my next comment.

src/ImageSharp/PixelFormats/PixelImplementations/Byte4.cs
```diff
@@ -101,11 +101,19 @@ public Vector4 ToVector4()
...
-        public void FromGray16(Gray16 source) => this.FromScaledVector4(source.ToScaledVector4());
+        public void FromL16(L16 source) => this.FromScaledVector4(source.ToScaledVector4());
+
+        /// <inheritdoc/>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public void FromLa16(La16 source) => this.FromScaledVector4(source.ToScaledVector4());
+
+        /// <inheritdoc/>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public void FromLa32(La32 source) => this.FromScaledVector4(source.ToScaledVector4());
```
antonfirsov: We could probably T4 these in the future to reduce the chore work.

tests/ImageSharp.Tests/PixelFormats/La16Tests.cs
```diff
@@ -0,0 +1,287 @@
...
+
+using System.Numerics;
+
+using SixLabors.ImageSharp.PixelFormats;
+using Xunit;
+
+// ReSharper disable InconsistentNaming
+namespace SixLabors.ImageSharp.Tests.PixelFormats
+{
+    public class La16Tests
```
antonfirsov: I wonder if it's worth to add an integration test doing some arbitrary stuff with `Image<La16>` and `Image<La32>` at least for debug support, or do we just trust the unit tests and the implicit coverage given by code in `PngScanlineProcessor`?

src/ImageSharp/PixelFormats/PixelImplementations/Byte4.cs
JimBobSquarePants: Yeah, that would be a great thing to do. It's a proper chore just now.

src/ImageSharp/Formats/Png/PngScanlineProcessor.cs
JimBobSquarePants: It's not known unfortunately, we're comparing the current luminance to the trans value.

src/ImageSharp/Formats/Tga/MetadataExtensions.cs
JimBobSquarePants: Yeah, @brianpopow did the whole thing! Amazing stuff!

tests/ImageSharp.Tests/PixelFormats/La16Tests.cs
JimBobSquarePants: I'm gonna leave it for now. The coverage is good via the `PixelOperations` overloads + tests.

src/ImageSharp/Formats/IImageDecoder.cs
JimBobSquarePants: I think the documented return type is adequate for now.

PR #1089, opened 2020-01-22
Harden memory-intensive tests, repurpose Sandbox46

antonfirsov:
### Prerequisites
- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

## The problem
To enable experiments on memory stuff like #898, it is necessery to put the testing of memory-intensive code on right track. These tests are important but at the same time threatening the execution with sudden OOM-s, which made us to disable many of them.

## Solution

It's possible to sandbox individual test code execution to separate processes with the help of [RemoteExecutor](https://github.com/dotnet/runtime/blob/master/docs/project/writing-tests.md#remoteexecutor) which is a [tool in the arcade project](https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.RemoteExecutor/src).

Limitation: The tool can only execute static methods, which can only consume string arguments.

## Details
- Upgrade to xunit 2.4.1, workaround xunit/xunit#2061 with the help of `TestPixel<TPixel>`
- Add package source containing RemoteExecutor
- Isolate (and re-enable):
  - `JpegDecoderTests`
  - `ArrayPoolMemoryAllocatorTests`
  - `BokehBlurTests`

## Bonus: simplify and repurpose Sandbox46
It's no longer needed to profile tests from Visual Studio, however it can be handy as a separate "profilable" .exe to run individual test blocks in a hard-coded way. This is useful eg. for memory profiling unit tests, since even JetBrains can't do that from IDE UI.
- Renamed to **ImageSharp.Tests.ProfilingSandbox**
- Referemce `ImageSharp.Tests.csproj` instead of including classes

PR #1090, opened 2020-01-22
Use stylecop in Test project

brianpopow:
### Prerequisites

### Description

This PR will introduce stylecop rules to the TestProject. The rules are less strict than in the main project. Those are  the rules:

```
<!--https://github.com/dotnet/roslyn/issues/18117-->
<Rule Id="AD0001" Action="None" />

<!--Missing XML comment for publicly visible type or member-->
<Rule Id="CS1591" Action="None" />

<!--Call 'ConfigureAwait(false)'-->
<Rule Id="RCS1090" Action="None" />
    
<!--XML comment analysis is disabled due to project configuration-->
<Rule Id="SA0001" Action="None" />

<!--Using directive should appear within a namespace declaration-->
<Rule Id="SA1200" Action="None" />

<!--Elements should be documented.-->
<Rule Id="SA1600" Action="None" />

<!--A field should not follow a method-->
<Rule Id="SA1201" Action="None" />

<!--Elements must be ordered by access-->
<Rule Id="SA1202" Action="None" />

<!--Constants must appear before fields-->
<Rule Id="SA1203" Action="None" />

<!--Static members should appear before non-static members-->
<Rule Id="SA1204" Action="None" />

<!--Field names must not contain underscore-->
<Rule Id="SA1310" Action="None" />

<!--Field should be private-->
<Rule Id="SA1401" Action="None" />

<!--A C# code file contains more than one unique type.-->
<Rule Id="SA1402" Action="None" />

<!--The last statement in a multi-line C# initializer or list is missing a trailing comma.-->
<Rule Id="SA1413" Action="None" />

<!--Partial elements should be documented.-->
<Rule Id="SA1601" Action="None" />

<!--Comments should end with a period. I like this but there's 3000+ errors to fix in ImageSharp-->
<Rule Id="SA1629" Action="None" /> 
    
<!--The XML documentation header for a C# constructor does not contain the appropriate summary text.-->
<Rule Id="SA1642" Action="None" />
    
<!--Do not use regions.-->
<Rule Id="SA1124" Action="None" />
    
<!--Constructor is missing documentation for one or more of its parameters.-->
<Rule Id="SA1611" Action="None" />

<!--A C# code file is missing a standard file header.-->
<Rule Id="SA1633" Action="Warning" />
```

Note to @JimBobSquarePants: I have added *SA1642, SA1124, SA1611*


antonfirsov:
I expect painful conflicts with #1089.

JimBobSquarePants:
> Note to @JimBobSquarePants: I have added SA1642, SA1124, SA1611

How were the rules added? Any additional rules should be added to 

https://github.com/SixLabors/SharedInfrastructure/blob/master/SixLabors.Tests.ruleset

and the submodule updated.

Regarding the addition of rules. 
- I want to keep copyright header in every file - We can do that globally via <kbd>ctrl</kbd> + <kbd>.</kbd>
- If a constructor has documentation it should be complete. If you think it's not required simply delete all the constructor documentation and any warnings will go away. Bad docs are worse than no docs.
- Regions can absolutely do one. I hate them 🤢

tests/ImageSharp.Tests/Drawing/DrawImageTests.cs
```diff
@@ -71,49 +70,46 @@ public void ImageBlendingMatchesSvgSpecExamples<TPixel>(TestImageProvider<TPixel
...
+            image.DebugSave(provider, testInfo, encoder: encoder);
+            image.CompareToReferenceOutput(
+                ImageComparer.TolerantPercentage(0.01f),
+                provider,
+                testInfo);
         }
 
         [Theory]
-        [WithTestPatternImages(200, 200, PixelTypes.Rgba32 | PixelTypes.Bgra32)]
+        [WithTestPatternImage(200, 200, PixelTypes.Rgba32 | PixelTypes.Bgra32)]
```
antonfirsov: The plural form might make sense here, since the attribute is feeding an image for each requested pixel type.

tests/ImageSharp.Tests/Formats/Jpg/Utils/ReferenceImplementations.LLM_FloatingPoint_DCT.cs
```diff
@@ -168,27 +166,27 @@ internal static void iDCT2D_llm(Span<float> s, Span<float> d, Span<float> temp)
...
-                c1 = _mm_load_ps(s, 40);
-                c0 = _mm_load_ps(s, 16);
-                Vector4 t2 = (c0 + c1);
-                Vector4 t5 = (c0 - c1);
-
-                c0 = _mm_load_ps(s, 24);
-                c1 = _mm_load_ps(s, 32);
-                Vector4 t3 = (c0 + c1);
-                Vector4 t4 = (c0 - c1);
+                Vector4 c0 = Mm_load_ps(s, 0);
```
antonfirsov: Nope, stylecop, you failed terribly again!

This code has been kept deliberately similar [to the original](https://github.com/norishigefukushima/dct_simd/blob/master/dct/dct8x8_simd.cpp). and `_mm_load_ps` refers to the corresponding C++ intrinsic to emphasize compatibility with that implementation.

tests/ImageSharp.Benchmarks/Codecs/Jpeg/DecodeJpeg_ImageSpecific.cs
```diff
@@ -49,18 +47,15 @@ public ShortClr()
         [Params(
             TestImages.Jpeg.BenchmarkSuite.Lake_Small444YCbCr,
             TestImages.Jpeg.BenchmarkSuite.BadRstProgressive518_Large444YCbCr,
+            /* The scaled result for the large image "ExifGetString750Transform_Huge420YCbCr"
```
antonfirsov: The `/*   */` are harder to work with in most C# IDE-s. Was the change necessary because of the rule? If yes, then I'm afraid that enforcing it may affect our test-code productivity. Or even worse: not placing comments where they should be present.

tests/ImageSharp.Tests/TestUtilities/ImageProviders/ITestImageProvider.cs
```diff
@@ -16,7 +16,9 @@ namespace SixLabors.ImageSharp.Tests
     public interface ITestImageProvider
```
antonfirsov: The file name is misleading now, we should either always move classes to a separate files, or change the rules.

tests/ImageSharp.Tests/TestUtilities/ImageProviders/ITestImageProvider.cs
```diff
@@ -25,29 +27,32 @@ public interface ITestImageProvider
     /// <summary>
     /// Provides <see cref="Image{TPixel}" /> instances for parametric unit tests.
     /// </summary>
-    /// <typeparam name="TPixel">The pixel format of the image</typeparam>
+    /// <typeparam name="TPixel">The pixel format of the image.</typeparam>
     public abstract partial class TestImageProvider<TPixel> : ITestImageProvider
```
antonfirsov: The file name is misleading now, we should either always move classes to a separate files, or change the rules.

antonfirsov:
> If a constructor has documentation it should be complete. If you think it's not required simply delete all the constructor documentation and any warnings will go away.

@JimBobSquarePants what do you mean by completeness?
For me, obvious `<param >` docs do not add any value. We can omit them while still documenting life-saving information in `<summary>`.

JimBobSquarePants:
@antonfirsov Constructor docs are traditionally. 
```c#
Initializes a new instance of the <see cref="MyType" /> class/struct.
```
IMO the params are the important part.

tests/ImageSharp.Tests/TestUtilities/ImageProviders/ITestImageProvider.cs
JimBobSquarePants: In this instance, yes I'd probably move. The relaxing of the rule is designed to allow things like combining enums and classes in the same file

tests/ImageSharp.Benchmarks/Codecs/Jpeg/DecodeJpeg_ImageSpecific.cs
JimBobSquarePants: I'd have added the spaces.

tests/ImageSharp.Benchmarks/General/PixelConversion/PixelConversion_Rgba32_To_Bgra32.cs
```diff
@@ -238,15 +243,29 @@ public void Bitops_Simd()
...
+            public uint Tmp17;
+            public uint Tmp20;
+            public uint Tmp23;
         }
 
         [StructLayout(LayoutKind.Sequential)]
-        struct C
+        private struct C
         {
-            public uint tmp3, tmp6, tmp9, tmp12, tmp15, tmp18, tmp21, tmp24;
```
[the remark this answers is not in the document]
JimBobSquarePants: Personally I don't like grouped declarations. It's too easy to set the variable to the wrong type.

antonfirsov:
> IMO the params are the important part.

In some cases yes. But mostly it becomes `<param name="value">The value.</param>`. Sebastian Aaltonen has [a good point](https://twitter.com/SebAaltonen/status/1082172243050983425) on why is this wrong. (One of the rare cases when I agree with a game developer actually.)

I wish that canonical text was not a thing with constructors :) It takes a whole sentence before the important stuff begins. This is wasted time for both doc writers and readers.

tests/ImageSharp.Tests/Formats/Jpg/Utils/ReferenceImplementations.LLM_FloatingPoint_DCT.cs
JimBobSquarePants: Stylecop hasn't failed here. The rule is good, this is an edge case.

The fix was based on a lack of critical information.

I would add the exception inline with a justification describing why this is important.

antonfirsov: Agree, just trying to be dramatic 😄

The point is that we need to be careful with the automatic application of the rules. If something needs to be suppressed too often, then re-evaluate the rule. Fortunately this doesn't seem to be the case with the current set.

JimBobSquarePants: Agree, the rules are not fixed and as soon as one causes significant issue we should definitely exclude it. 

All I'm looking for by adding rules is a little bit more structure to our tests. A lot of different people have submitted significant parts of code now and it's become, for me, at least difficult to follow at times.

The less draconian the ruleset the better. Let's just be careful to not [throw the baby out with the bathwater](https://en.wikipedia.org/wiki/Don%27t_throw_the_baby_out_with_the_bathwater).

tests/ImageSharp.Benchmarks/Codecs/Jpeg/DecodeJpeg_ImageSpecific.cs
brianpopow: There are two conflicting rules here:

- SA1515	Single-line comment should be preceded by blank line
- SA1115	The parameter should begin on the line after the previous parameter.

I have changed it back now as it was before with an exception for SA1115, see commit: d59ca988a0e730db1bfbac9825593e58f2b8b1e9

tests/ImageSharp.Tests/Formats/Jpg/Utils/ReferenceImplementations.LLM_FloatingPoint_DCT.cs
brianpopow: i have reverted it back, to be more like the original code again: 94dc54b4144394f125673066487c0e5a18036a04

tests/ImageSharp.Tests/Drawing/DrawImageTests.cs
brianpopow: changed with: f897ab6f8da1c3efe770039dd0d9d41ad879afb3

tests/ImageSharp.Benchmarks/General/PixelConversion/PixelConversion_Rgba32_To_Bgra32.cs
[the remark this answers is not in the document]
brianpopow: its back to a single line: d59ca988a0e730db1bfbac9825593e58f2b8b1e9

tests/ImageSharp.Tests/TestUtilities/ImageProviders/ITestImageProvider.cs
brianpopow: I have moved the interface to a separate file

tests/ImageSharp.Benchmarks/Codecs/Jpeg/DecodeJpeg_ImageSpecific.cs
```diff
@@ -47,12 +47,15 @@ public ShortClr()
         [Params(
             TestImages.Jpeg.BenchmarkSuite.Lake_Small444YCbCr,
             TestImages.Jpeg.BenchmarkSuite.BadRstProgressive518_Large444YCbCr,
-            /* The scaled result for the large image "ExifGetString750Transform_Huge420YCbCr"
-             is almost the same as the result for Jpeg420Exif,
-             which proves that the execution time for the most common YCbCr 420 path scales linearly.
-             TestImages.Jpeg.BenchmarkSuite.ExifGetString750Transform_Huge420YCbCr,
-             */
+
+#pragma warning disable SA1115 // Parameter should follow comma
```
dlemstra: Should we not just disable these kind of rules instead?

brianpopow: SA1115  is now disabled

brianpopow:
I have some trouble with the ProfilingSanbox project. There are some errors and alot of wanings im not sure how to resolve.

The errors look like this: 
`Error	CS0121	The call is ambiguous between the following methods or properties: 'SixLabors.ImageSharp.Tests.TestUtils.IsEquivalentTo<TPixel>(SixLabors.ImageSharp.Image<TPixel>, SixLabors.ImageSharp.Image<TPixel>, bool)' and 'SixLabors.ImageSharp.Tests.TestUtils.IsEquivalentTo<TPixel>(SixLabors.ImageSharp.Image<TPixel>, SixLabors.ImageSharp.Image<TPixel>, bool)'	ImageSharp.Tests.ProfilingSandbox (net472), ImageSharp.Tests.ProfilingSandbox (netcoreapp2.1), ImageSharp.Tests.ProfilingSandbox (netcoreapp3.1)	E:\vsprojekte\ImageSharp\tests\ImageSharp.Tests\TestUtilities\Tests\TestUtilityExtensionsTests.cs	59`

And the warnings:
`
Warning	CS0436	The type 'ApproximateColorSpaceComparer' in 'E:\vsprojekte\ImageSharp\tests\ImageSharp.Tests\Colorspaces\Conversion\ApproximateColorspaceComparer.cs' conflicts with the imported type 'ApproximateColorSpaceComparer' in 'SixLabors.ImageSharp.Tests, Version=0.0.1.0, Culture=neutral, PublicKeyToken=null'. Using the type defined in 'E:\vsprojekte\ImageSharp\tests\ImageSharp.Tests\Colorspaces\Conversion\ApproximateColorspaceComparer.cs'.	ImageSharp.Tests.ProfilingSandbox (net472), ImageSharp.Tests.ProfilingSandbox (netcoreapp2.1), ImageSharp.Tests.ProfilingSandbox (netcoreapp3.1)	E:\vsprojekte\ImageSharp\tests\ImageSharp.Tests\Colorspaces\Conversion\RgbAndYCbCrConversionTest.cs	20
`

@JimBobSquarePants have you seen those errors here before? Any ideas how to resolve them would be very welcome.

src/Directory.Build.props
```diff
@@ -30,8 +30,8 @@
     <InternalsVisibleTo Include="DynamicProxyGenAssembly2" PublicKey="0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7" />
     <InternalsVisibleTo Include="DynamicProxyGenAssembly2, PublicKeyToken=null" />
     <InternalsVisibleTo Include="ImageSharp.Benchmarks" />
-    <InternalsVisibleTo Include="SixLabors.ImageSharp.Sandbox46" />
     <InternalsVisibleTo Include="SixLabors.ImageSharp.Tests" />
+    <InternalsVisibleTo Include="ImageSharp.Tests.ProfilingSandbox" />
```
JimBobSquarePants: Isn't this your breaking change?

brianpopow: i have added this because i was getting more than 2k errors of `Error	CS0122	'...' is inaccessible due to its protection level`

JimBobSquarePants: That's because you're including all the files from the test project in the sandbox. 

https://github.com/SixLabors/ImageSharp/pull/1090/files#diff-791cc76ca5e2665ba784771043fd9123R17

We don't do that anymore. The sandbox project simply references the test one.

brianpopow: > That's because you're including all the files from the test project in the sandbox.
> 
> https://github.com/SixLabors/ImageSharp/pull/1090/files#diff-791cc76ca5e2665ba784771043fd9123R17
> 
> We don't do that anymore. The sandbox project simply references the test one.

great, thank you! I was not aware of that. I have the removed the tests from the profiling sandbox, now it work works. See: commit: 0bce7af847619aeae7bce0357a1d2179c128cb14

src/ImageSharp/Primitives/ColorMatrix.cs
```diff
@@ -204,7 +204,7 @@ public bool IsIdentity
         /// <returns>The resulting matrix.</returns>
         public static ColorMatrix operator +(ColorMatrix value1, ColorMatrix value2)
         {
-            ColorMatrix m;
+            var m = default(ColorMatrix);
```
JimBobSquarePants: I don't like this. The declaration was more clear and less verbose before. Why the change?

brianpopow: because of https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0818

JimBobSquarePants: That's not the same thing. Our struct was explicitly declared. I'll leave it for now anyway.

PR #1093, opened 2020-01-25
Update Exif LensInfo

### LINKED ISSUE #1091: EXIF tag LensInfo = 0xA432 is treated as Rational but it is actually Rational[4]
When reading EXIF data ImageSharp tries to read the tag LensInfo = 0xA432 as a Rational but fails. The LensInfo is actually an array of Rational with size 4 according to 

https://exiftool.org/TagNames/EXIF.html

and 

https://www.exiv2.org/tags.html

and the image files I have.

To reproduce the issue:

1. In the method ExifReader.TryReadValue put a break point in the places where the method returns false. It the last case:

            if (!exifValue.TrySetValue(value))
            {
                return false;
            }

that fails since exifValue is expecting Rational but value is Rational[4].

2. Using 

var image= Image.Load("filename.jpg");

load the attached image.

exif

I am using the source code from today (24 January 2020) on a Windows 10 PC and targeting .Net Core 3.1.

The issue can be fixed by changing the value type return by ExifValues.CreateValue(ExifTagValue tag)

I am inexperienced using GitHub but can attempt to make a pull request if you agree that this is a bug.

JimBobSquarePants:
### Prerequisites

### Description
Rename `LenseInfo` to `LensSpecification` and change return type to `Rational[]` to match the specification.

Fix #1091


JimBobSquarePants:
@brianpopow the **0xc630** one isn't actually part of the EXIF specification. There's a copy of the spec in the solution folder.

PR #1095, opened 2020-01-28
Expose Pixel Operations and Configuration

### LINKED ISSUE #1092: expose ToRgba64 + add GetConfiguration to image frame
was trying to do a copy of ImpageCompare code in ImageSharp.Textures however noticed ToRgba64 is internal and only ToVector4 is available, could you please make public...

Example use of ToRgba64

PixelOperations<TPixelA>.Instance.ToRgba64(configuration, aSpan, aBuffer);

Additionally Image has a GetConfiguration() however there is no equivalent for ImageFrane

JimBobSquarePants:
### Prerequisites

### Description
- Exposes the various `IPixel` operations as bulk operations. Individual overriding classes remain internal. 
- Normalize `Configuration` usage and expose `GetConfiguration()` extension for `ImageFrame`
- Normalize `MemoryAllocator` usage and expose only via `Configuration`.

Fix #1092

### DIFF
--- a/src/ImageSharp/ImageFrame.cs
+++ b/src/ImageSharp/ImageFrame.cs
@@ -31,16 +31,10 @@ namespace SixLabors.ImageSharp
             Guard.NotNull(metadata, nameof(metadata));
 
             this.configuration = configuration ?? Configuration.Default;
-            this.MemoryAllocator = configuration.MemoryAllocator;
             this.Width = width;
             this.Height = height;
             this.Metadata = metadata;
         }
-
-        /// <summary>
-        /// Gets the <see cref="MemoryAllocator" /> to use for buffer allocations.
-        /// </summary>
-        public MemoryAllocator MemoryAllocator { get; }
 
         /// <summary>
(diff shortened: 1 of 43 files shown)


src/ImageSharp/ImageFrame.cs
```diff
@@ -13,37 +13,28 @@ namespace SixLabors.ImageSharp
     /// In case of animated formats like gif, it contains the single frame in a animation.
     /// In all other cases it is the only frame of the image.
     /// </summary>
-    public abstract partial class ImageFrame : IDisposable
+    public abstract partial class ImageFrame : IConfigurable, IDisposable
```
antonfirsov: This definition somewhat implies that a frame is configurable separatly from it's owner image.
Shouldn't we rename `IConfigurable` to something like `IConfigurationProvider`?

JimBobSquarePants: Happy to do so. It's an internal interface anyway so not a breaking change.

PR #1096, opened 2020-01-28
Rename Pixel Shader Extensions and Delegates

JimBobSquarePants:
### Prerequisites

### Description
The `PixelShader` extensions, processors and delegates were too suggestive of GPU based functionality so the decision was made to rename them.

All functionality is the same only classes and delegates were renamed.

cc. @Sergio0694


Sergio0694:
Looks great!
And the new `ProcessPixelRowsAsVector4` name is really growing on me, I like the simplicity and intuitiveness of it a lot! Hopefully this will also help drive its adoption even more, which will be a huge win for everyone here, and especially for end users struggling to figure out how to properly process abstract image instances (@JimBobSquarePants you know who I'm talking about 😆).

src/ImageSharp/Processing/PositionAwarePixelRowOperation.cs
```diff
@@ -7,10 +7,10 @@
...
 {
     /// <summary>
-    /// A <see langword="delegate"/> representing a user defined pixel shader.
+    /// A <see langword="delegate"/> representing a user defined, position aware, processing delegate to use to modify image rows.
     /// </summary>
     /// <param name="span">The target row of <see cref="Vector4"/> pixels to process.</param>
     /// <param name="offset">The initial horizontal and vertical offset for the input pixels to process.</param>
     /// <remarks>The <see cref="Vector4.X"/>, <see cref="Vector4.Y"/>, <see cref="Vector4.Z"/>, and <see cref="Vector4.W"/> fields map the RGBA channels respectively.</remarks>
-    public delegate void PositionAwarePixelShader(Span<Vector4> span, Point offset);
+    public delegate void PositionAwarePixelRowOperation(Span<Vector4> span, Point offset);
```
antonfirsov: I don't think we need the verbosity here, the following might be enough:
```C#
public delegate void PixelRowOperation(Span<Vector4> span, Point offset);
```

We need to make sure that we will be able to extend the API with the following variant, since it is easier to use for non-bounded processing:
```C#
public delegate void PositionAwarePixelRowOperation(Span<Vector4> span, int row);
```

src/ImageSharp/Processing/Processors/Effects/PixelRowDelegateProcessor.cs
```diff
@@ -0,0 +1,39 @@
...
+// Licensed under the Apache License, Version 2.0.
+
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Effects
+{
+    /// <summary>
+    /// Applies a user defined row processing delegate to the image.
+    /// </summary>
+    public sealed class PixelRowDelegateProcessor : IImageProcessor
```
antonfirsov: The name may conflict with the future generic variant I'm proposing.
Instead of going deeper in the naming rabbit hole, I suggest internalizing it for now.

JimBobSquarePants: I'm happy to internalize the processors for now.

Sergio0694:
@antonfirsov I'm a bit confused, why do we want to add overloads that work with a specific pixel type, given that those will prevent the JITter from just compiling the `Vector4` operations into SIMD instructions? Is that just to make it easier to work with for users, or am I missing something? 🤔

I mean, my fear is that a general developer presented with both the `Vector4` and `TPixel` variants would just go for a `TPixel` one out of habit and write the delegate working with, say, `Rgba32`, and essentially doing the same exact stuff he could've done with `Vector4`, but losing SIMD altogether and ending up with slower code for no real reason, which would not be ideal.

antonfirsov:
@Sergio0694 it's 4x more compact in memory, conversion is much faster, if done right (NOP for `Rgba32`) . Advanced users can do `System.Runtime.Intrinsics` for SIMD, but most folks just want to play around with the good-old RGBA `byte` values in the easiest possible way. Believe or not, `Vector4` normalized to `0..1` is an abstraction many devs just don't get :unamused:

src/ImageSharp/Processing/PositionAwarePixelRowOperation.cs
JimBobSquarePants: How would the extension methods know which delegate is passed?

antonfirsov: I think C# compiler should be able to distinguish them based on signature, but it's worth a check.

JimBobSquarePants: Figured out a nice way using generics. It also frees us up for using `TPixel` later.

Sergio0694:
@antonfirsov Ah, yeah I hadn't considered that. I personally don't get why the `0..1` range would be that confusing (it's literally just `0.255` scaled down, but that's about it), but yeah you make a very good point there. I hope some devs will take the time to look into the `Vector4` APIs though instead of just settling for the pixel typed versions, because the next step which you mentioned, intrinsics, is way tougher to get to. As in, getting from `TPixel` to `Vector4` is doable, getting to using intrinsics takes years. And getting to _comfortably_ using intrinsics takes a lifetime 🤣

Also I might be wrong here, but I was wondering:
1) I agree that being able to skip the pixel conversions entirely can give a nice performance bump, but isn't the (vectorized?) `ToVector4` pass pretty optimized to be a reasonably small overhead anyway?
2) In non-server scenarios where one's not worried about potential memory issues caused by too many parallel executions, isn't the allocation cost for the `Vector4` buffers negligible, since they're all just just `unmanaged` types being pooled from the same `byte` buffers anyway?
3) Doesn't working with `Vector4` also have the advantage of greater "accuracy" when applying arbitrary effects? As in, since the `0..1` range with floating points is much wider than just `0..255`, doesn't that essentially makes processors work in a larger color space? Which is eventually reduced back, sure, but you can retain more info during the processing steps, similar to how eg. games process intermediate buffers in HDR even if they then only display them in standard SDR 8bit 🤔

Thanks in advance, I'm just trying to learn as much as I can from you guys 😄

JimBobSquarePants:
@SixLabors/core Any blockers here or can I merge?

antonfirsov:
@JimBobSquarePants give me 10 minutes.

src/ImageSharp/Processing/PixelRowOperation.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// A <see langword="delegate"/> representing a user defined processing delegate to use to modify image rows.
+    /// </summary>
+    /// <typeparam name="T">
+    /// The type of the parameter of the method that this delegate encapsulates.
+    /// This type parameter is contravariant.That is, you can use either the type you specified or any type that is less derived.
+    /// </typeparam>
+    /// <param name="span">The target row of <see cref="Vector4"/> pixels to process.</param>
+    /// <param name="value">The parameter of the method that this delegate encapsulates.</param>
+    /// <remarks>The <see cref="Vector4.X"/>, <see cref="Vector4.Y"/>, <see cref="Vector4.Z"/>, and <see cref="Vector4.W"/> fields map the RGBA channels respectively.</remarks>
+    public delegate void PixelRowOperation<in T>(Span<Vector4> span, T value);
```
antonfirsov: Just wanted to double-check if these may conflict with the variants needed for the future `ProcessPixelRowsAs` API, but looks like it's not the case.

```C#
public delegate void PixelRowOperation<in TPixel>(Span<TPixel> span) where TPixel : struct, IPixel<TPixel>;

public delegate void PixelRowOperation<in TPixel, TParam>(Span<TPixel> span, TParam value) where TPixel : struct, IPixel<TPixel>;
```

JimBobSquarePants: Yeah, didn't think it would be an issue. Thanks for reviewing!

antonfirsov:
@Sergio0694 this is the `Rgba32 <=> Vector4` conversion code:
https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Common/Helpers/SimdUtils.ExtendedIntrinsics.cs#L104

It's the fastest thing we can do, but still expensive.

In comparison, a `Bgra32 -> Rgba32` conversion could be done using a few shuffle intrinsics, and you can always fit 8 (!!!) pixels into an AVX2 register instead of only 2 (2x`Vector4`).

Of course, in case of more complex processing operations, it would be compensated by using `Vector4` in the operation body, but if a user only wants to do something basic (like setting the alpha channel to some value, or collecting pixel coordinates over a specific luminosity), the `RGB(A)` variants will be faster.

Sergio0694:
@antonfirsov Ah, yeah I see now, makes perfect sense. Thanks! 😄

PR #1100, opened 2020-01-30
Default Gif RepeatCount to 1

JimBobSquarePants:
### Prerequisites

### Description
We were incorrectly setting the default repeat count to `0` when reading gifs without a Netscape application extension block. This causes multiframe gifs without the block to repeat indefinitely which is incorrect.


JimBobSquarePants:
The behavior matches the specification. If you want different behavior you should update the `GifMetadata` property of the image `Metadata` to reflect your requirements.

https://docs.sixlabors.com/api/ImageSharp/SixLabors.ImageSharp.MetadataExtensions.html#SixLabors_ImageSharp_MetadataExtensions_GetGifMetadata_SixLabors_ImageSharp_Metadata_ImageMetadata_

Issue #355, opened 2017-09-29
Converting Png to Jpg give a color background and not white (Asp.Net Core)


FelixLeChat:
### Description
My goal is to generate renditions (thumbnail, and other size of the image) from an uploaded image (IFormFile, png or jpg).

When uploading a .png with transparency, the resulting image has a background of a color in the image and not a default white one. 

### Steps to Reproduce
```
        public static Stream GenerateCompressedOriginal(IFormFile file)
        {
            Stream outputStream = new MemoryStream();

            using (Image<Rgba32> image = Image.Load(file.OpenReadStream()))
            {
                image.SaveAsJpeg(outputStream, new JpegEncoder() { Quality = 80 });
            }

            return outputStream;
        }
```
From :
![2553905512_listing_logo](https://user-images.githubusercontent.com/8709465/31017762-a7ddefdc-a4f7-11e7-9a18-20aed2765bca.png)

To :
![2553905512_preview](https://user-images.githubusercontent.com/8709465/31017773-b1b7670e-a4f7-11e7-9979-ae30b5e3b569.jpg)

if i save the outputstream as a jpg. the error occur. The image variable has also this background in color and not white. 

### System Configuration

- ImageSharp version: 1.0.0-alpha9-00194
- Environment (Operating system, version and so on): windows 10
- .NET Framework version: .Net Standard 1.6

JimBobSquarePants:
Hi @FelixLeChat ,

The transparent sections of the image are not actually fully transparent, rather only the alpha component of the pixels is set to zero.

You'll need to manually set the background color.

``` csharp
    public static Stream GenerateCompressedOriginal(IFormFile file)
        {
            Stream outputStream = new MemoryStream();

            using (Image<Rgba32> image = Image.Load(file.OpenReadStream()))
            {
                image.Mutate(x => x.BackgroundColor(Rgba32.White));
                image.SaveAsJpeg(outputStream, new JpegEncoder() { Quality = 80 });
            }

            return outputStream;
        }
```

Cheers

James

tocsoft:
I feel we should be doing this by default when encoding into formats that don't support alpha channels.

I think we should add a `NormalizeBackground` property onto `JpegEncoder`  that, if enabled (defaulted to `true`), will do the background color fixes as it encodes each pixel.

JimBobSquarePants:
I disagree. The encoding overheads in my opinion are too high and it's technically incorrect. 
[...]

PR #1104, opened 2020-02-02
Add new Color parsing methods.

### LINKED ISSUE #1103: Add Color.FromHtml
It would be extremely useful for migration from `System.Drawing` to this library to have the `FromHtml` equivalent.
The `FromHtml` takes a string which can be a color name or hex string and converts it to the relevant color.
As a side note, there are 3.5 million downloads to this library, I think it can be released instead of being a pre-release...
Can also be called `Color.FromString`.

JimBobSquarePants:
### Prerequisites

### Description
- Adds `Color.TryParse`
- Adds `Color.Parse`
- Adds `Color.TryParseHex`
- Renames `Color.FromHex` to `Color.ParseHex`
- Adds `Rgba32.TryParseHex`
- Renames `Rgba32.FromHex` to `Rgba32.ParseHex`

`Color.TryParse` and `Color.Parse` recognize both color names and hexadecimal representation as per CSS4 specification in ordinal casing.  Fix #1103

<strike>I haven't added `Rgba32.TryParse` and `Rgba32.Parse` as I'm very tempted to drop the `Rgba32` named colors. I'd much rather channel color creation through the `Color` struct only. 

Please let me know if you agree that this is the best course of action.</strike>

EDIT. I went ahead and deleted them. Much happier now.


JimBobSquarePants:
@SixLabors/core Any objections to this. I'll add it tonight otherwise.

brianpopow:
> @SixLabors/core Any objections to this. I'll add it tonight otherwise.

i have looked trough the changes. Looks good to me, i think it can be merged.

jespersh:
Is this the commit that broke examples that use like?:
```C#
image.Mutate(x => x.BackgroundColor(Rgba32.White));
```
I suppose the replacement is just using `Color.White` instead?

JimBobSquarePants:
There’s nothing broken, this is by design. `Color` is the correct pixel agnostic way to declare colors.

jespersh:
I didn't say it was broken, I said it broke the examples :smile:

JimBobSquarePants:
Which examples are broken? I’ll update if we own them.

jespersh:
I followed this path to here:
https://stackoverflow.com/a/60642706
https://github.com/SixLabors/ImageSharp/issues/355#issuecomment-333133991

Thanks for this library.

JimBobSquarePants:
@jespersh no worries! yeah, that's all out of date. We've got API docs https://docs.sixlabors.com/ that will be much more useful.

PR #1107, opened 2020-02-06
Remove Closure Allocations when Iterating Rows in Parallel

JimBobSquarePants:
### Prerequisites

### Description
The way ImageSharp currently handles work-to-be-executed in parallel, at the simplest level, as a `Action<RowInterval>`. Any state necessary for the operation must therefore be provided as closures, incurring heap allocations. 

By utilizing a trick to exploit the reification in C# generics it's possible completely get rid of `callvirt` instructions when invoking a delegate passed as argument which, in turn, allows the compiler to inline the input delegate. This removes the closure allocations and importantly removes the risk of developers inadvertently passing `this` as an allocation. 

We can perform this optimization by updating the current `ParallelHelper.IterateRows` methods and variants to accept a readonly struct implementing an interface instead  the current action. This optimization technique is already utilized within the `IImageProcessor<TPixel>` API.

The interface pattern is as follows.

```c#
/// <summary>
/// Defines the contract for an action that operates on a row interval.
/// </summary>
public interface IRowIntervalAction
{
    /// <summary>
    /// Invokes the method passing the row interval.
    /// </summary>
    /// <param name="rows">The row interval.</param>
    void Invoke(in RowInterval rows);
}

/// <summary>
/// Defines the contract for an action that operates on a row interval with a temporary buffer.
/// </summary>
/// <typeparam name="TBuffer">The type of buffer elements.</typeparam>
public interface IRowIntervalAction<TBuffer>
    where TBuffer : unmanaged
{
    /// <summary>
    /// Invokes the method passing the row interval and a buffer.
    /// </summary>
    /// <param name="rows">The row interval.</param>
    /// <param name="memory">The contiguous region of memory.</param>
    void Invoke(in RowInterval rows, Memory<TBuffer> memory);
}
```

And the methods as follows.

```c#
/// <summary>
/// Iterate through the rows of a rectangle in optimized batches defined by <see cref="RowInterval"/>-s.
/// </summary>
/// <typeparam name="T">The type of row action to perform.</typeparam>
/// <param name="rectangle">The <see cref="Rectangle"/>.</param>
/// <param name="parallelSettings">The <see cref="ParallelExecutionSettings"/>.</param>
/// <param name="body">The method body defining the iteration logic on a single <see cref="RowInterval"/>.</param>
public static void IterateRows<T>(
    Rectangle rectangle,
    in ParallelExecutionSettings parallelSettings,
    in T body)
    where T : struct, IRowIntervalAction

/// <summary>
/// Iterate through the rows of a rectangle in optimized batches defined by <see cref="RowInterval"/>-s
/// instantiating a temporary buffer for each <paramref name="body"/> invocation.
/// </summary>
internal static void IterateRows<T, TBuffer>(
    Rectangle rectangle,
    in ParallelExecutionSettings parallelSettings,
    in T body)
    where T : struct, IRowIntervalAction<TBuffer>
    where TBuffer : unmanaged
```

### TODO
- [x] Define new interfaces and methods 
- [x] Refactor existing processor to use new methods
- [x] Remove old methods and update tests


JimBobSquarePants:
3... 2... 1... 😉

src/ImageSharp/Advanced/ParallelRowIterator.Wrappers.cs
```diff
@@ -0,0 +1,110 @@
...
+            public WrappingRowIntervalInfo(int minY, int maxY, int stepY, int maxX)
+            {
+                this.MinY = minY;
+                this.MaxY = maxY;
+                this.StepY = stepY;
+                this.MaxX = maxX;
+            }
+        }
+
+        private readonly struct WrappingRowIntervalOperation<T>
```
antonfirsov: Shouldn't this be `RowIntervalOperationWrapper<T>` instead? Took me a for a while to figure out it is exactly the opposite as it's name :)

src/ImageSharp/Advanced/ParallelRowIterator.Wrappers.cs
```diff
@@ -0,0 +1,110 @@
...
+{
+    /// <content>
+    /// Utility methods for batched processing of pixel row intervals.
+    /// Parallel execution is optimized for image processing based on values defined
+    /// <see cref="ParallelExecutionSettings"/> or <see cref="Configuration"/>.
+    /// Using this class is preferred over direct usage of <see cref="Parallel"/> utility methods.
+    /// </content>
+    public static partial class ParallelRowIterator
+    {
+        private readonly struct WrappingRowIntervalInfo
```
antonfirsov: This deserves a better name. `IterationParameters` maybe?

src/ImageSharp/Advanced/ParallelRowIterator.Wrappers.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            public readonly int MinY;
+            public readonly int MaxY;
+            public readonly int StepY;
+            public readonly int MaxX;
```
antonfirsov: Shouldn't this be width? First I thought there is a bug, and we are actually allocating a buffer with the size defined by the maximum `x` coordinate.

src/ImageSharp/Advanced/ParallelRowIterator.Wrappers.cs
JimBobSquarePants: I'm fine with that, will update.

src/ImageSharp/Advanced/ParallelRowIterator.Wrappers.cs
JimBobSquarePants: Sounds good to me.

src/ImageSharp/Advanced/ParallelRowIterator.Wrappers.cs
JimBobSquarePants: Haha, yeah no problem.

PR #1109, opened 2020-02-08
Implement discontiguous buffer handling

### LINKED ISSUE #805: System.ArgumentOutOfRangeException (Parameter name: minimumLength) when loading moderate size jpg image
### Prerequisites

### Description
Trying to load a jpeg image (1024x36000) fails with following stack trace.

```
Unhandled Exception: System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: minimumLength
   at System.Buffers.ConfigurableArrayPool`1.Rent(Int32 minimumLength)
   at SixLabors.Memory.ArrayPoolMemoryAllocator.Allocate[T](Int32 length, AllocationOptions options)
   at SixLabors.ImageSharp.Memory.MemoryAllocatorExtensions.Allocate2D[T](MemoryAllocator memoryAllocator, Int32 width, Int32 height, AllocationOptions options)
   at SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder.JpegComponent.Init()
   at SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder.JpegFrame.InitComponents()
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.ProcessStartOfFrameMarker(Int32 remaining, JpegFileMarker& frameMarker, Boolean metadataOnly)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.ParseStream(Stream stream, Boolean metadataOnly)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.Decode[TPixel](Stream stream)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoder.Decode[TPixel](Configuration configuration, Stream stream)
   at SixLabors.ImageSharp.Image.Decode[TPixel](Stream stream, Configuration config)
   at SixLabors.ImageSharp.Image.Load[TPixel](Configuration config, Stream stream, IImageFormat& format)
   at SixLabors.ImageSharp.Image.Load[TPixel](Configuration config, String path)
   at dotnet_sample.Program.Main(String[] args) in /home/sinan/tmp/dotnet-sample/Program.cs:line 15

```
I believe the size of the image does not hit to 2GB or 4GB buffer limits which are mentioned in some other issues.

### Steps to Reproduce

- Open GIMP and create an image in 1024x36000 width x height.
- Export as jpeg.
- Try to load the jpeg via `Image.Load`.

### System Configuration

- ImageSharp version: 1.0.0-dev002315
- Other ImageSharp packages and versions: SixLabors.Core: 1.0.0-dev000096
- Environment (Operating system, version and so on): Ubuntu 18.04
- .NET Framework version: 2.1 (also tried with 2.2)
- Additional information:
Test file: [sample.zip](https://github.com/SixLabors/ImageSharp/files/2739921/sample.zip)

### LINKED ISSUE #898: Enable images of arbitrary size
Currently our images and internal buffers are limited to `Int32.MaxValue` size, `ArgumentOutOfRange` exception [is thrown](https://github.com/SixLabors/Core/blob/master/src/SixLabors.Core/Memory/ArrayPoolMemoryAllocator.cs#L100) when a code attempts to allocate a larger buffer.

The most robust way to overcome this limitation would be enabling discontinuous buffers in our memory management, ~by adapting `ReadOnlySequence<T>`~ defining a writable memory sequence primitive. *EDIT: `ReadOnlySequence<T>` is not the way to go, it's purpose is completely different.*

This is a breaking change, so we really need to know how many users are interested in this use case, before commiting ourself. If you are one of them, please let us know about your use case in comments! (Platform, purpose of application etc.)

antonfirsov:
### Prerequisites

## Description
~Marked as WIP~ because processors and codecs need adaption, but the core logic is **ready to review**!

- `MemorySource<T>` has been replaced with `MemoryGroup<T>`, which is a list of uniformly sized buffers (`IMemoryOwner<T>` and `Memory<T>`)
- Several helpers intend to provide span-like functionalities in `MemoryGroupExtensions` eg. `.CopyTo()`
- `MemoryGroupView<T>` is an indirection to expose these memory buffers in a safe way, see docs and comments for more info
- The methods exposed over `AdvancedMemoryExtensions` have been changed to reflect the new memory model, `.GetPixelSpan<T>()` has been made obsolete
- `buffer2d.GetSpan<T>()` ---> `buffer2d.GetSingleSpan<T>()`, temporarily marked `[Obsolete]` to guide refactor

## TODO
We should no longer process `Buffer2D` as a single span (except of a few special cases like resize KernelMap). Guidance: go through all `[Obsolote]` warnings.

- [x] Add tests for each codec and processor in a similar manner like it has been added to Jpeg tests. (Artifically limit buffer capacity of allocator to emulate the discontiguous case.)
- [x] Fix remaining codecs
- [x] Fix remaining  processors

Resolves #898, Fixes #805


src/ImageSharp/Advanced/AdvancedImageExtensions.cs
```diff
@@ -40,23 +41,66 @@ public static Configuration GetConfiguration(this ImageFrame source)
...
         public static Span<TPixel> GetPixelSpan<TPixel>(this ImageFrame<TPixel> source)
             where TPixel : struct, IPixel<TPixel>
-            => source.GetPixelMemory().Span;
+        {
+            Guard.NotNull(source, nameof(source));
+
+            IMemoryGroup<TPixel> mg = source.GetPixelMemoryGroup();
+            if (mg.Count > 1)
+            {
+                throw new InvalidOperationException($"GetPixelSpan is invalid, since the backing buffer of this {source.Width}x{source.Height} sized image is discontiguos!");
```
Therzok:
```suggestion
                throw new InvalidOperationException($"GetPixelSpan is invalid, since the backing buffer of this {source.Width}x{source.Height} sized image is discontiguous!");
```

antonfirsov: Thanks! I hate this word, and always fail to type it correctly :smile:

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
```diff
@@ -40,23 +41,66 @@ public static Configuration GetConfiguration(this ImageFrame source)
...
-        /// stored in row major order.
+        /// Gets the representation of the pixels as a <see cref="IMemoryGroup{T}"/> containing the backing pixel data of the image
+        /// stored in row major order, as a list of contiguous <see cref="Memory{T}"/> blocks in the source image's pixel format.
         /// </summary>
+        /// <param name="source">The source image.</param>
         /// <typeparam name="TPixel">The type of the pixel.</typeparam>
-        /// <param name="source">The source.</param>
+        /// <returns>The <see cref="IMemoryGroup{T}"/>.</returns>
+        /// <remarks>
+        /// Certain Image Processors may invalidate the returned <see cref="IMemoryGroup{T}"/> and all it's buffers,
```
JimBobSquarePants: Is this actually possible? All processors that alter the dimensions of an image operate on a new image with the expected output dimensions.

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroupExtensions.cs
```diff
@@ -0,0 +1,231 @@
...
+            private int elementIndex;
+
+            public MemoryGroupCursor(IMemoryGroup<T> memoryGroup)
+            {
+                this.memoryGroup = memoryGroup;
+                this.bufferIndex = 0;
+                this.elementIndex = 0;
+            }
+
+            private bool IsAtLastBuffer => this.bufferIndex == this.memoryGroup.Count - 1;
```
JimBobSquarePants: This appears to be unused.

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroup{T}.Consumed.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Buffers;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SixLabors.ImageSharp.Memory
+{
+    internal abstract partial class MemoryGroup<T>
+    {
+        // Analogous to the "consumed" variant of MemorySource
+        private class Consumed : MemoryGroup<T>
```
JimBobSquarePants: Should seal classes like this.

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroup{T}.Owned.cs
```diff
@@ -0,0 +1,102 @@
...
+                }
+            }
+
+            public override IEnumerator<Memory<T>> GetEnumerator()
+            {
+                this.EnsureNotDisposed();
+                return this.memoryOwners.Select(mo => mo.Memory).GetEnumerator();
+            }
+
+            public override void Dispose()
```
JimBobSquarePants: It's generally wise to add an internal check when disposing to ensure calling `Dispose()` multiple times cannot cause issues.

src/ImageSharp/Memory/Buffer2D{T}.cs
```diff
@@ -62,17 +71,76 @@ internal Buffer2D(MemorySource<T> memorySource, int width, int height)
...
-            this.MemorySource.Dispose();
+            this.MemoryGroup.Dispose();
+            this.cachedMemory = default;
+        }
+
+        /// <summary>
+        /// Same as <see cref="GetRowSpan"/>, but does not validate index in Release mode.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal Span<T> GetRowSpanUnchecked(int y)
```
JimBobSquarePants: I feel like this is something that could be handled better with compiler conditionals rather than two separate methods. It's a little confusing knowing which to call although it does seem the other version is only called via tests.

src/ImageSharp/Memory/Buffer2D{T}.cs
```diff
@@ -62,17 +71,76 @@ internal Buffer2D(MemorySource<T> memorySource, int width, int height)
...
+        }
+
+        /// <summary>
+        /// Gets a <see cref="Memory{T}"/> to the row 'y' beginning from the pixel at the first pixel on that row.
+        /// This method is intended for internal use only, since it does not use the indirection provided by
+        /// <see cref="MemoryGroupView{T}"/>.
+        /// </summary>
+        /// <param name="y">The y (row) coordinate.</param>
+        /// <returns>The <see cref="Span{T}"/>.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
JimBobSquarePants: More of a question than anything. When should we use `InliningOptions`?

antonfirsov: Always! Forgot it by mistake :smile:

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
antonfirsov: Here is a test case that demonstrates the issue:
https://github.com/SixLabors/ImageSharp/blob/0675c80d2731396bdb964c4b5c587fee70e3b0a7/tests/ImageSharp.Tests/Advanced/AdvancedImageExtensionsTests.cs#L44-L53

Without `MemoryGroupView`, this is what would happen:

1. You take a reference to a memory area within the image with `Memory<TPixel>`
2. After Resize, the referenced backing buffer is returned to the pool during the Swap operations. 
3. The `Memory<T>` is referencing a buffer that has nothing to do with the image instance. Worst case: it is got rented by some other code leading to strange bugs, and safety issues.

This problem was originally pointed out by @tocsoft, and this is why we haven't exposed `GetPixelMemory()` and `GetPixelRowMemory()`.
`MemoryGroupView` is protecting against these bugs by adding a layer of indirection, so we can throw an exception when the user aims to access memory area that no longer belongs to the image.

Note that this can still happen by taking a row span, resizing an image and accessing the span later. But normally no one should hit this issue, since spans are intended to be short living objects for quick processing of data. `Memory<T>` is heapable, one can store a long-living reference in an class, or long-running async closure, without being aware that it gets outdated.

We should elaborate on this in our conceptual docs.

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroup{T}.Owned.cs
antonfirsov: This is done by checking `this.memoryOwners == null`. Could be more explicit by adding a property though.

src/ImageSharp/Memory/Buffer2D{T}.cs
antonfirsov: I'm also unsure about this solution. Normally, the checked variant should be good and fast enough, and that is the only public variant, so ImageSharp.Drawing will not be able to rely on the unchecked one.

On the other hand, there are some special cases like Jpeg where we need the extra perf from omitting the bouds checking:
https://github.com/SixLabors/ImageSharp/blob/9c6f1180393145dcc8009b9d563d36ccb42fdaf0/src/ImageSharp/Formats/Jpeg/Components/RowOctet.cs#L30-L37

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
James-Jackson-South: Ah yeah, I totally understand the risk now. Yeah, lot's of warnings.

src/ImageSharp/Memory/Buffer2D{T}.cs
James-Jackson-South: I see the dilemma, ok, since it's not public lets carry on as is.

antonfirsov:
Note that the new `LimitAllocatorBufferCapacity().InBytes(100)` test API I just pushed, will set the capacity to equivalent to a 100x100 byte area (10000 bytes). For me this is easier to follow, but I'm open for better API ideas.

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -294,25 +296,28 @@ private void ReadRle<TPixel>(BmpCompression compression, Buffer2D<TPixel> pixels
             where TPixel : struct, IPixel<TPixel>
         {
             TPixel color = default;
-            using (Buffer2D<byte> buffer = this.memoryAllocator.Allocate2D<byte>(width, height, AllocationOptions.Clean))
-            using (Buffer2D<bool> undefinedPixels = this.memoryAllocator.Allocate2D<bool>(width, height, AllocationOptions.Clean))
+            using (IMemoryOwner<byte> buffer = this.memoryAllocator.Allocate<byte>(width * height, AllocationOptions.Clean))
```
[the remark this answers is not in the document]
brianpopow: If i would use `Buffer2d` i would need to process row wise, which is not really suitable in the current implementation. One run length can span over multiple rows.

I think it is possible to change RLE decompression to work with rows, but it would be a complete overhaul of the implementation. I dont think at the moment it is worth the effort. Mainly because i dont see the usecase of very large RLE bitmaps.

src/ImageSharp/Formats/Bmp/BmpDecoder.cs
```diff
@@ -32,7 +33,20 @@ public Image<TPixel> Decode<TPixel>(Configuration configuration, Stream stream)
...
+            {
+                return decoder.Decode<TPixel>(stream);
+            }
+            catch (InvalidMemoryOperationException ex)
+            {
+                Size dims = decoder.Dimensions;
+
+                // TODO: use InvalidImageContentException here, if we decide to define it
+                // https://github.com/SixLabors/ImageSharp/issues/1110
+                throw new ImageFormatException($"Can not decode image. Failed to allocate buffers for possibly degenerate dimensions: {dims.Width}x{dims.Height}.", ex);
```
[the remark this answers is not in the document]
antonfirsov: @brianpopow we may want to change the text that the other possible reason is a very large RLE bitmap which is is not supported. Then we can remove or change the failing tests I added yesterday.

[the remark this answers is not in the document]
brianpopow: ok i will change the message and increase the limit for the buffer capacity for the tests a bit.

src/ImageSharp/Memory/Buffer2DExtensions.cs
```diff
@@ -14,62 +15,72 @@ namespace SixLabors.ImageSharp.Memory
...
+        /// Gets the backing <see cref="IMemoryGroup{T}"/>.
         /// </summary>
-        /// <param name="buffer">The <see cref="Buffer2D{T}"/>.</param>
-        /// <typeparam name="T">The value type.</typeparam>
-        /// <returns>The <see cref="Span{T}"/> referencing the memory area.</returns>
-        public static Span<T> GetSpan<T>(this Buffer2D<T> buffer)
+        /// <param name="buffer">The buffer.</param>
+        /// <typeparam name="T">The element type.</typeparam>
+        /// <returns>The MemoryGroup.</returns>
+        public static IMemoryGroup<T> GetMemoryGroup<T>(this Buffer2D<T> buffer)
```
JimBobSquarePants: I've been looking at the accessibility of the `IMemoryGroup<T>` of `Buffer2D<T>` property and exposing method and I'm wondering if we should be exposing it at all. At the moment it feels like there's too much choice which is a shade confusing. 

I'd really like to lock the properties of `Buffer2D<T>` right down both internally and externally so there is "one true way" to do everything and ensure we are actually dogfooding our API design.

antonfirsov: Yes, this makes sense, it's aperformance VS clean & safe API dilemma again. I didn't have enough information to solve this properly, but unfortunately it's likely that ImageSharp.Drawing needs to operate on `MemoryGroup<T>`, so we have to figure this out.

Possible solution:
```C#
public class Buffer2D<T>
{
	// ** Returns MemoryGroupView<T> **
	// ---> Safe, PREFERRED way internally
	public IMemoryGroup<T> MemoryGroup { get; } 


	// ** Returns MemoryGroup<T> **
	// Unsafe to the public, useful when we need to be super-performant
	// on frequent accesses to Memory<T> (eg. JpegEncoder)
	internal MemoryGroup<T> GetInternalMemoryGroup() { }
}
```

antonfirsov: This actually leads to a more general design guideline question. (Not like if we had any design guideline haha):

I think we are **overusing unsafe things internally**. It's likely that the changes in this PR caused a buffer overflow because of this. Here I also mean things like `Unsafe.Add()` which became bread-and-butter for us without being really conscious about the consequences. It would be probably wise to push back on this and limit unsafe things to the following cases:
- Core internals like `PixelOperations<T>`
- Hottest execution path for web thumbnail making. (JpegDecoder, Resize, JpegEncoder)
- In other processors / codecs only if:
  - The code is mature enough
  - It's not possible to optimize it further using safe techniques

src/ImageSharp/Memory/Buffer2D{T}.cs
JimBobSquarePants: @antonfirsov I've been rethinking this and I think we should use the unchecked version as the public version. My reasoning is that we do not do any explicit bounds checks for `Image<TPixel>[x,y]` in release mode and rely on the framewrok check/massages instead.

src/ImageSharp/Memory/Buffer2DExtensions.cs
JimBobSquarePants: I couldn't agree more.

JimBobSquarePants: I don't think there's any issue with making `MemoryGroup<T>` public I think we should just be careful when scoping instances in parent classes.

src/ImageSharp/Memory/Buffer2D{T}.cs
antonfirsov: The problem with this is that it goes against [design guidelines for public API exception safety](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types#nullreferenceexception-indexoutofrangeexception-and-accessviolationexception).

It's not only about bounds-checking: we also need to decide what to do with `null` safety, doesn't make sense to deal with only one of thise concerns. (Guideline suggestion is not to leak `NullReferenceException`!)

Of course we may decide to compromise on this (pro: perf, thinner API surface), but we probably don't want to. It would be better IMO to eventually expose the unsafe/unchecked variants to advanced users / downstream libs, since their dangerous nature is obvious from the naming.

src/ImageSharp/Memory/Buffer2DExtensions.cs
antonfirsov: My main argument against publishing `MemoryGroup<T>` is the same as always:

It's a new type, not mature enough, I want to avoid a post-RC1 breaking changes if we decide to improve something as we get more experience by dogfeeding it in the next few months.

src/ImageSharp/Memory/Buffer2D{T}.cs
JimBobSquarePants: Is this something we should fix within the encoder itself. The idea of two methods that do the same thing with and without checking irks me.

antonfirsov: Not with the current design. There is a private member, `cachedMemory`, to speed up serving requests for the most common single-buffer case:
https://github.com/SixLabors/ImageSharp/blob/b35e3cbfbd133eb0d3290a2d537d40f649444133/src/ImageSharp/Memory/Buffer2D%7BT%7D.cs#L112-L114

Even if could manage to solve this (which I doubt we can do in an elegant way), should we duplicate the unchecked code everywhere across the lib? A few places where we definitely need it:
- `image[x,y]` indexer (to avoid doing bounds-checking twice)
- `DenseMatrixUtils.Convolve2DImpl()`
- Resize
- Jpeg encoder & decoder

I can't see a win-all situation here we either:
1. Compromise performance
2. Compromise public API exception guidelines
3. Keep the unchecked variants for internal/advanced scenarios

For me the 3rd is the lesser evil.

Note that I created the two variants by renaming the old `.GetRowSpan(y)` to `.GetRowSpanUnchecked(y)`, and then creating a new `GetRowSpan(y)` that is actually checked, to make sure I don't regress performance. The original `.GetRowSpan(y)` was unchecked and public 😕

antonfirsov: Rethinking it again, actually ignoring the guideline on *advanced types*, might be also an option.

Most important is to make a conscious decision, so far `Buffer2D` was non-conformant by accident.

JimBobSquarePants: I guess for now we have to stick with option 3 then and ensure the method docs indicate its use case so we don't abuse it.

antonfirsov: I'm so good at changing my mind in just a few minutes, I already started coding option 2 right now 😄

Should I stop?

JimBobSquarePants: Hahaha....  Well.... They are guidelines not explicit rules. 

My thinking is this. Those exceptions are "reserved" for the framework but by relying on them instead of relying on our own checks we do not expose the user to any greater risk unless we use unsafe code without bounds checks. You and I both agree that we should severely limit usage of the unsafe code so I would like to say to be damned with the guidelines. The fact we are having this conversation suggests we know what we are doing.

antonfirsov: Those guidelines are here to return useful error information for typical users, saving time for everyone.

But this also indicates, that we can make exceptions for (and only for!) advanced scenarios, if we properly document them. Most people won't even notice that `Buffer2D<T>` exists.

antonfirsov: Pushed changes, including the update of the public indexers, it turned out to be much better than I expected.

James-Jackson-South: Excellent. Will have a look tonight.

src/ImageSharp/ImageFrame{TPixel}.cs
```diff
@@ -148,13 +148,22 @@ internal ImageFrame(Configuration configuration, ImageFrame<TPixel> source)
...
         /// <returns>The <see typeparam="TPixel"/> at the specified position.</returns>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the provided (x,y) coordinates are outside the image boundary.</exception>
         public TPixel this[int x, int y]
         {
-            [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => this.PixelBuffer[x, y];
+            [MethodImpl(InliningOptions.ShortMethod)]
+            get
+            {
+                this.VerifyCoords(x, y);
```
antonfirsov: @James-Jackson-South I added proper bounds checking here. To compensate the performance penalty, I tried to make the underlying `GetElementUnsafe()` implementation as fast as possible, but I expect the aggregate performance change to be still negative. Still don't think we should omit proper input validation on a method with this level of visibility.

I would advise everyone to use `.GetRowSpan(y)` anyways, `[x,y]` is slow by design, whatever we do.

JimBobSquarePants: `[x,y]` is there for convenience in the most part, though it's used a lot in drawing and in algorithms where we are not operating on the current row span. Error diffusion for example. If you think perf is not affected too badly then I'm up for it.

antonfirsov: `[x,y]` should be used on `Buffer2D<T>` in processor code, which has been optimized now as much as we can.

antonfirsov:
Looks like I managed to fix everything, removed the WIP tag, ready for final review.

src/ImageSharp/Formats/Jpeg/Components/RowOctet.cs
```diff
@@ -0,0 +1,60 @@
...
+            this.row1 = y < height ? buffer.GetRowSpan(y++) : default;
+            this.row2 = y < height ? buffer.GetRowSpan(y++) : default;
+            this.row3 = y < height ? buffer.GetRowSpan(y++) : default;
+            this.row4 = y < height ? buffer.GetRowSpan(y++) : default;
+            this.row5 = y < height ? buffer.GetRowSpan(y++) : default;
+            this.row6 = y < height ? buffer.GetRowSpan(y++) : default;
+            this.row7 = y < height ? buffer.GetRowSpan(y) : default;
+        }
+
+        public Span<T> this[int y]
```
JimBobSquarePants: Could we use a ThrowHelper and inline here?

PR #1112, opened 2020-02-13
Use Moment struct in WuQuantizer.

JimBobSquarePants:
### Prerequisites

### Description
Very simple PR. Refactors the `WuFrameQuantizer<TPixel>` to us a single pooled buffer instead of many. Makes code much easier to follow, improves performance, and reduces GC pressure.


src/ImageSharp/Processing/Processors/Quantization/WuFrameQuantizer{TPixel}.cs
```diff
@@ -199,27 +155,18 @@ protected override ReadOnlyMemory<TPixel> GetPalette()
...
 
                 for (int k = 0; k < this.colors; k++)
                 {
                     this.Mark(ref this.colorCube[k], (byte)k);
 
-                    float weight = Volume(ref this.colorCube[k], vwtSpan);
+                    Moment moment = Volume(ref this.colorCube[k], momentsSpan);
 
-                    if (MathF.Abs(weight) > Constants.Epsilon)
+                    if (moment.Weight > 0)
```
antonfirsov: I'm not familiar with the algorithm, but removing `Epsilon` changes the algorithm a bit. Are there any corner cases where this may have an impact?

JimBobSquarePants: None that I could find or test. I know the algorithm pretty well now and the epsilons were actually added as part of some bad design from me in the past. Weight should be a `long`.

antonfirsov: Cool! If the struct is uniform (all members Int64), it opens up nice SIMD optimization opportunities.

PR #1114, opened 2020-02-16
Fix quantization and dithering.

JimBobSquarePants:
### Prerequisites

### Description
A deep investigation into our quantization and dithering algorithms yielded issues we needed to fix.

- Error Diffusion algorithm offset was incorrect.
- Ordered Dithering algorithm was completely incorrect. 
- Quantization could not utilize both algorithms.
- Quantization could not be isolated to a given bounds.
- `Color.Transparent` did not match W3C specification, rather copied incorrect System.Drawing value.
- Tests were leaking memory.

This PR:

- Updates Error Diffusion algorithm with correct implementation.
- Updates Ordered Dithering algorithm with correct implementation.
- Merges `IErrorDiffusion` and `IOrderedDither` APIs into new `IDither` api.
- Updates and simplifies dithering processor to use new API
- Updates and simplifies format quantization and quantization processor to use new API.
- Introduces new `QuantizerOptions` for easier bundling of options.
- Introduces new dither scaling options to allow variable dithering.
- Fixed `Color.Transparent` value.
- Improves performance across the board.
- Fixes memory leak in Bitmap Encoder
- Fixes memory leak in Quantizer tests.

### Examples
Here's an example of the new dithering output vs the old using `BayerDither8x8` ordered dithering. Note the much improved handling of gradients.

**Old**
DitherFilter_WorksWithAllDitherers_Bike_BayerDither8x8-Old

**New**
DitherFilter_WorksWithAllDitherers_Bike_BayerDither8x8-New

With the new `DitherScale` options built into `QuantizerOptions` it's possible to incrementally dither output.

Here's incremental quantization and dithering of a sample image. I've used the websafe palette to accentuate differences.

**Original**
david

**0%**
FloydSteinbergDither 0%

**25%**
FloydSteinbergDither 25%

**50%**
FloydSteinbergDither 50%

**75%**
FloydSteinbergDither 75%

**100%**
FloydSteinbergDither 100%

### Benchmarks

_Note: Previous benchmarks did not have Core 3.1 versions. Memory increase is due to changing of color lookup caching strategy._

``` bash
BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
Intel Core i7-8650U CPU 1.90GHz(Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK = 3.1.101

[Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT
Job-OJKYBT : .NET Framework 4.8 (4.8.4121.0), X64 RyuJIT
Job-RZWLFP : .NET Core 2.1.15 (CoreCLR 4.6.28325.01, CoreFX 4.6.28327.02), X64 RyuJIT
Job-NUYUQV : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT

```
**Before**
|    Method |  Job | Runtime |      Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------- |----- |-------- |----------:|----------:|----------:|------:|------:|------:|----------:|
| DoDiffuse |  Clr |     Clr | 124.93 ms | 33.297 ms | 1.8251 ms |     - |     - |     - |      2 KB |
| DoDiffuse | Core |    Core |  89.63 ms |  9.895 ms | 0.5424 ms |     - |     - |     - |   1.91 KB |

**After**
|    Method |       Runtime |      Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------- |-------------- |----------:|----------:|----------:|------:|------:|------:|----------:|
| DoDiffuse |    .NET 4.7.2 | 30.535 ms | 19.217 ms | 1.0534 ms |     - |     - |     - |  26.25 KB |
|  DoDither |    .NET 4.7.2 | 14.174 ms |  1.625 ms | 0.0891 ms |     - |     - |     - |  31.38 KB |
| DoDiffuse | .NET Core 2.1 | 15.984 ms |  3.686 ms | 0.2020 ms |     - |     - |     - |  25.98 KB |
|  DoDither | .NET Core 2.1 |  8.646 ms |  1.635 ms | 0.0896 ms |     - |     - |     - |  28.99 KB |
| DoDiffuse | .NET Core 3.1 | 16.235 ms |  9.612 ms | 0.5269 ms |     - |     - |     - |  25.96 KB |
|  DoDither | .NET Core 3.1 |  8.429 ms |  1.270 ms | 0.0696 ms |     - |     - |     - |  31.61 KB |


src/ImageSharp/Processing/Processors/Dithering/IDither.cs
```diff
@@ -0,0 +1,45 @@
...
+            ImageFrame<TPixel> image,
+            Rectangle bounds,
+            TPixel source,
+            TPixel transformed,
+            int x,
+            int y,
+            int bitDepth,
+            float scale)
+            where TPixel : struct, IPixel<TPixel>;
+    }
```
antonfirsov: Nice job with the unification, but I'm still wondering if we can find a better abstraction/representation here. The worst thing is that is hard to introduce any optimization with the current design, because of the per-pixel virtual `Dither<T>` method. A it is a clear warning sign for me, that we have 2-element enum defining the two possible algorithms in parallel with the two matching subclasses.

A little bit unorthodox in the world of OO, but I'm wondering about your thoughts on the following idea: Instead of pretending that we are doing something abstract and pluggable in this method, we could simply describe the ditherers by the data (and only the data!) that defines them:

```c#
public struct ErrorDiffusionParameters
{
    public int Offset { get; }

    public DenseMatrix<float> Matrix { get; }
}

public struct OrderedDitherParameters
{
    public DenseMatrix<float> ThresholdMatrix { get; }
}

public sealed class Dither
{
    // Determines which algorithm to use
    public DitherType DitherType { get; }

    // Parameters in case of ErrorDiffusion
    public ErrorDitherParameters? ErrorDiffusionParameters { get; }

    public OrderedDitherParameters? OrderedDitherParameters { get; }

    public static Dither CreateErrorDiffusion(int offset, DenseMatrix<float> matrix) { ... }

    public static Dither CreateOrderedDither(DenseMatrix<float> thresholdMatrix) { .... }
}

public static class KnownDitherings
{
    public static Dither BayerDither2x2 { get; } = CreateBayerDither2x2();

    private static Dither CreateBayerDither2x2()
    {
       DenseMatrix<float> thresholdMatrix;
       
       // calculate thresholdMatrix

       return Dither.CreateOrderedDither(thresholdMatrix);
    }
}

// Implementation could be moved to an internal class.
// Use the per-pixel variants for now, but it would be easy to switch to bulk variants later,
// without getting lost in API design
internal static class DitherExtensions
{
    internal static TPixel ApplyOrderedDither<TPixel>(
            this Dither dither,
            ImageFrame<TPixel> image,
            Rectangle bounds,
            TPixel source,
            TPixel transformed,
            int x,
            int y,
            int bitDepth,
            float scale)
    {
        // ....
    }

    internal static TPixel ApplyErrorDither<TPixel>(
            this Dither dither,
            ImageFrame<TPixel> image,
            Rectangle bounds,
            TPixel source,
            TPixel transformed,
            int x,
            int y,
            int bitDepth,
            float scale)
    {
        // ....
    }
}
```

src/ImageSharp/Processing/Processors/Quantization/QuantizerConstants.cs
```diff
@@ -17,5 +19,20 @@ internal static class QuantizerConstants
...
+
+        /// <summary>
+        /// The max dithering scale used to adjust the amount of dither.
+        /// </summary>
+        public const float MaxDitherScale = 1F;
+
+        /// <summary>
+        /// Gets the default dithering algorithm to use.
+        /// </summary>
+        public static IDither DefaultDither { get; } = KnownDitherings.FloydSteinberg;
```
antonfirsov: Isn't really a constant. Woldn't it better to define it in `KnownDitherings` instead? (`KnownDitherings.Default`)

src/ImageSharp/Processing/Processors/Dithering/IDither.cs
JimBobSquarePants: This is very interesting! I'll have a play around later tonight.

JimBobSquarePants: One issue is the sheer amount of dithering approaches.

- Random,
- BlueNoise
- Halftone

Each one means an extra nullable options property which seems a bit icky.

antonfirsov: I didn't know that, but it changes the picture a lot, because it means we need real extensibility here. The `DitherType` enum makes it hard to add a new dithering, since it needs touching processor/quantizer code as well.

I think we should keep thinking and find a reasonable abstraction. I have an idea for a double-dispatch trick, I will gist it up later today.

src/ImageSharp/Processing/Processors/Quantization/IFrameQuantizer{TPixel}.cs
```diff
@@ -15,22 +15,18 @@ public interface IFrameQuantizer<TPixel> : IDisposable
...
-        /// Gets a value indicating whether to apply dithering to the output image.
+        /// Gets the quantizer options defining quantization rules.
         /// </summary>
-        bool Dither { get; }
-
-        /// <summary>
-        /// Gets the error diffusion algorithm to apply to the output image.
-        /// </summary>
-        IErrorDiffuser Diffuser { get; }
+        QuantizerOptions Options { get; }
```
antonfirsov: All usages of `Options` are in the implementing classes, which means that there is probably no point exposing this on the interface.

src/ImageSharp/Processing/Processors/Dithering/IDither.cs
JimBobSquarePants: Yeah, good point. I'd really like to investigate the possibility of `IDither` implementations being a struct passed via a generic argument also. (_IResampler in the future also in a different PR_).

Been looking at the JIT ASM to see what the difference is between the two approaches and it's chalk and cheese. 

Current class implementation
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYmpBkIYBvGgzNNdxFAwCyACgCUx0+dcvXZhdigMfAXgYuGAB3Bk4uIKg7IND5bGkAEwAxCAhHBw1qD3Modm47cgz3BgBfGjLqLV1cGHiYBL0w7kiTSqzzHSaImTsASRSIBgAzVIcjN3bsjAALXlwWAYYAkYhM1wqJ107+1IZFowBzGAw1CuLOqy67Xh4GBDHi1xm5hdSWAEEEhLt7tfMKs5tTo3DAyIbYFQMHYQVquEEMT7feG/cqaWi6bDAXAYKAQjCNAYgaHOSYMYoAbRsx2mEASvXEkjsVJmtPpYkkAHkxHwIFx5u8DgdYLhcLwFDBelxJDcbgcHABdc66eGI663e6w7JMADsdwYAGoGOQ/mYAWjOjU6g0yAw4olFiA9rtWuUgA=

Generic struct version
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYmpBkIYBvGgzNNdxFAwCyACgCUx0+dcvXZhdigMfAXgYuGAB3Bk4uIKgAHnlsaQATADEICAA+OyDQ2ITkiEcHDWoPcyh2bjtyAvcGAF8aOuoaAG1ZAAtvMQAZbGAynn5BACleDABxGEjlOwwATzEYCAAzO2yYJJSHBwBdLV1cGDi1vTDuSKiAFVTq4NaZGAZzhhAGAElcmhMi8x0TiJk7R6LDZGNxfYoYVq8XAsXIMAJAiCFVwNUGuH6PWFGADmMAwaga1R+Vl+dl4PAYCAcn2KZghUJhKRYAEF4vE7JSkeYGgTGrRdGSMDJFtgVK9ctTzAKGCy2VKOfVNHyGD1cBgoCKMMdciA3ilnGCGNUmjZca0IPEXuJJHYTRDzZaxJIAPJiPgQLjQplYrGwXC4XgKGAvLiSMlkrHbQn88ky0nkykS4rEADsFIYAGoGOROWZuYqfvtDvFjqt1hAngxMYqGkA===

Regarding utility methods or extensions. We'll have to build ones that handle the different behaviors in the `FrameQuantizer<TPixel>` and `PaletteDitherProcessor<TPixel>`.

PaletteDitherProcessor
https://github.com/SixLabors/ImageSharp/blob/b9069de3c87b8889918033e92619518c10747e0a/src/ImageSharp/Processing/Processors/Dithering/PaletteDitherProcessor%7BTPixel%7D.cs#L62


FrameQuantizer
https://github.com/SixLabors/ImageSharp/blob/b9069de3c87b8889918033e92619518c10747e0a/src/ImageSharp/Processing/Processors/Quantization/FrameQuantizer%7BTPixel%7D.cs#L176

src/ImageSharp/Processing/Processors/Quantization/QuantizerConstants.cs
JimBobSquarePants: We'd have to do that across the board for all our `Known***` types.

src/ImageSharp/Processing/Processors/Dithering/IDither.cs
antonfirsov: Good idea, I'll keep that in mind!

antonfirsov: Ok, I mostly figured out `PaletteQuantizer`, but I have no idea how to handle `PaletteDitherProcessor`:
https://gist.github.com/antonfirsov/cbad26bffc39fdd9803b7978a2e87abb

This is getting extremely complex and time-consuming if we want to do it right. Wouldn't it be wiser to close down the API for now, and open a post-RC1 issue?

JimBobSquarePants: Looks like we're both heading in the same direction. Bear with me, I think I can crack it today.

src/ImageSharp/Processing/Processors/Dithering/ErrorDither.cs
```diff
@@ -0,0 +1,162 @@
...
+using SixLabors.ImageSharp.PixelFormats;
+using SixLabors.ImageSharp.Processing.Processors.Quantization;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Dithering
+{
+    /// <summary>
+    /// An error diffusion dithering implementation.
+    /// <see href="http://www.efg2.com/Lab/Library/ImageProcessing/DHALF.TXT"/>
+    /// </summary>
+    public readonly partial struct ErrorDither : IDither, IEquatable<ErrorDither>
```
antonfirsov: I wonder what are the use-cases for equality comparison. Should we override also `==`? Do we also need to implement `IEquatable<IDither>`? Don't we need a few tests to make sure it works as intended?

While analyzing this I found #1116.

JimBobSquarePants: Yeah we should implement `IEquatable<IDither>` plus operators. I'll do that now.

JimBobSquarePants: Done. Also fixed the `DenseMatrix<T>` hashcode generation and added tests.

PR #1118, opened 2020-02-21
Faster Transforms

JimBobSquarePants:
### Prerequisites

### Description
- Inverts the call direction of all transforms to allow inlining and improve performance. 
- Fix invisible pixel bleed in transforms.
- Converts `IResampler` implementations into readonly struct 
 
### Bug Fixes
The attached image shows the difference in output following the updates. RGB color information no longer leaks beyond the bounds of the transformed result.

rotate-compare

### Benchmarks
Massive performance improvements in Affine and Projective transforms.

#### Rotate

Before
|   Method |       Runtime |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |-------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| DoRotate |    .NET 4.7.2 | 64.96 ms | 47.74 ms | 2.617 ms |     - |     - |     - |  11.56 KB |
| DoRotate | .NET Core 2.1 | 38.57 ms | 14.38 ms | 0.788 ms |     - |     - |     - |   4.36 KB |
| DoRotate | .NET Core 3.1 | 38.67 ms | 50.52 ms | 2.769 ms |     - |     - |     - |   5.68 KB |

After
|   Method |       Runtime |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |-------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| DoRotate |    .NET 4.7.2 | 28.77 ms | 3.304 ms | 0.181 ms |     - |     - |     - |    6.5 KB |
| DoRotate | .NET Core 2.1 | 16.27 ms | 1.044 ms | 0.057 ms |     - |     - |     - |   5.25 KB |
| DoRotate | .NET Core 3.1 | 17.12 ms | 4.352 ms | 0.239 ms |     - |     - |     - |   6.57 KB |

#### Skew

Before
| Method |       Runtime |     Mean |     Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |-------------- |---------:|----------:|---------:|------:|------:|------:|----------:|
| DoSkew |    .NET 4.7.2 | 42.78 ms | 11.455 ms | 0.628 ms |     - |     - |     - |   9.33 KB |
| DoSkew | .NET Core 2.1 | 27.71 ms |  3.880 ms | 0.213 ms |     - |     - |     - |   4.34 KB |
| DoSkew | .NET Core 3.1 | 25.06 ms |  6.461 ms | 0.354 ms |     - |     - |     - |   5.63 KB |

After
| Method |       Runtime |     Mean |     Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |-------------- |---------:|----------:|---------:|------:|------:|------:|----------:|
| DoSkew |    .NET 4.7.2 | 24.60 ms | 33.971 ms | 1.862 ms |     - |     - |     - |    6.5 KB |
| DoSkew | .NET Core 2.1 | 12.13 ms |  2.256 ms | 0.124 ms |     - |     - |     - |   5.21 KB |
| DoSkew | .NET Core 3.1 | 12.83 ms |  1.442 ms | 0.079 ms |     - |     - |     - |   6.57 KB |

#### Resize
Resize improvements are less dramatic which is expected. NET Core 3.1 shows a nice improvement though bringing it inline with NETFX. A slight improvement overall. 

Before
|                                   Method |       Runtime | SourceToDest |     Mean |     Error |   StdDev | Ratio | RatioSD | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------- |-------------- |------------- |---------:|----------:|---------:|------:|--------:|------:|------:|------:|----------:|
|                            SystemDrawing |    .NET 4.7.2 |     3032-400 | 82.97 ms |  5.478 ms | 0.300 ms |  1.00 |    0.00 |     - |     - |     - |    1170 B |
| 'ImageSharp, MaxDegreeOfParallelism = 1' |    .NET 4.7.2 |     3032-400 | 51.53 ms |  3.713 ms | 0.203 ms |  0.62 |    0.00 |     - |     - |     - |   16384 B |
|                                          |               |              |          |           |          |       |         |       |       |       |           |
|                            SystemDrawing | .NET Core 2.1 |     3032-400 | 82.83 ms |  2.049 ms | 0.112 ms |  1.00 |    0.00 |     - |     - |     - |      96 B |
| 'ImageSharp, MaxDegreeOfParallelism = 1' | .NET Core 2.1 |     3032-400 | 62.95 ms |  5.588 ms | 0.306 ms |  0.76 |    0.00 |     - |     - |     - |   15560 B |
|                                          |               |              |          |           |          |       |         |       |       |       |           |
|                            SystemDrawing | .NET Core 3.1 |     3032-400 | 86.34 ms | 80.550 ms | 4.415 ms |  1.00 |    0.00 |     - |     - |     - |      96 B |
| 'ImageSharp, MaxDegreeOfParallelism = 1' | .NET Core 3.1 |     3032-400 | 57.54 ms | 86.545 ms | 4.744 ms |  0.67 |    0.08 |     - |     - |     - |   15536 B |

After

|                                   Method |       Runtime | SourceToDest |     Mean |     Error |   StdDev | Ratio | RatioSD | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------- |-------------- |------------- |---------:|----------:|---------:|------:|--------:|------:|------:|------:|----------:|
|                            SystemDrawing |    .NET 4.7.2 |     3032-400 | 80.61 ms | 11.068 ms | 0.607 ms |  1.00 |    0.00 |     - |     - |     - |    1365 B |
| 'ImageSharp, MaxDegreeOfParallelism = 1' |    .NET 4.7.2 |     3032-400 | 49.10 ms |  3.663 ms | 0.201 ms |  0.61 |    0.00 |     - |     - |     - |   21372 B |
|                                          |               |              |          |           |          |       |         |       |       |       |           |
|                            SystemDrawing | .NET Core 2.1 |     3032-400 | 80.30 ms | 13.570 ms | 0.744 ms |  1.00 |    0.00 |     - |     - |     - |      96 B |
| 'ImageSharp, MaxDegreeOfParallelism = 1' | .NET Core 2.1 |     3032-400 | 61.99 ms | 20.643 ms | 1.132 ms |  0.77 |    0.01 |     - |     - |     - |   19816 B |
|                                          |               |              |          |           |          |       |         |       |       |       |           |
|                            SystemDrawing | .NET Core 3.1 |     3032-400 | 81.54 ms | 49.267 ms | 2.701 ms |  1.00 |    0.00 |     - |     - |     - |      96 B |
| 'ImageSharp, MaxDegreeOfParallelism = 1' | .NET Core 3.1 |     3032-400 | 49.75 ms |  0.764 ms | 0.042 ms |  0.61 |    0.02 |     - |     - |     - |   19868 B |


src/ImageSharp/Processing/Processors/Transforms/Resize/ResamplerExtensions.cs
```diff
@@ -0,0 +1,227 @@
...
+            bool compand)
+            where TResampler : unmanaged, IResampler
+            where TPixel : struct, IPixel<TPixel>
+        {
+            PixelConversionModifiers conversionModifiers =
+                PixelConversionModifiers.Premultiply.ApplyCompanding(compand);
+
+            BufferArea<TPixel> sourceArea = source.PixelBuffer.GetArea(sourceRectangle);
+
+            // To reintroduce parallel processing, we would launch multiple workers
```
JimBobSquarePants: @antonfirsov I'm wondering why we haven't done this?

antonfirsov: Shouldn't be rocket science, but still non trivial, had to stop somewhere.

I'm not the biggest fan of parallelization anyways, since it's mostly for desktop and low-load web workflows, not sure if we picked the best defaults for `LevelOfParallelism`.

JimBobSquarePants: I think desktop is a reasonable target and will be a very popular one once we expand our format support. Web applications tend to have less image processing throughput than you might imagine (sensible ones anyway) since they would employ pre generation and caching for the most part. 

Regarding `LevelOfParalleism`, Yeah I don't know, I guess research and experimentation...

Not a major concern anyway. Our Resize is fast enough for now and focus is better spent elsewhere.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
```diff
@@ -28,15 +28,23 @@ internal ResizeKernel(int startIndex, float* bufferPtr, int length)
         /// <summary>
         /// Gets the start index for the destination row.
         /// </summary>
-        public int StartIndex { get; }
+        public int StartIndex
+        {
+            [MethodImpl(InliningOptions.ShortMethod)]
```
antonfirsov: I think the JIT is smart enough to do this without our intervention. I see the .NET BCL code pushing back a lot on applications for explicit `AggressiveInlining` these days, I suppose the JIT does it more reliably with really short methods.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs
```diff
@@ -167,12 +184,14 @@ public void Dispose()
             return result;
         }
 
-        protected virtual void Initialize()
+        /// <summary>
+        /// Initializes the kernel map.
+        /// </summary>
+        protected internal virtual void Initialize()
```
antonfirsov: If we could pass `TResampler` as an argument to this method instead of storing it as a member, we could flip back `ResizeKernelMap` to a non-generic class, with only the methods `Calculate<TResampler>(..)` and `Initialize<TResampler>(...) being generic.

src/ImageSharp/Processing/Processors/Transforms/IResampler.cs
```diff
@@ -21,5 +24,54 @@ public interface IResampler
...
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="configuration">The configuration.</param>
+        /// <param name="source">The source image frame.</param>
+        /// <param name="destination">The destination image frame.</param>
+        /// <param name="matrix">The transform matrix.</param>
+        void ApplyProjectiveTransform<TPixel>(
+            Configuration configuration,
+            ImageFrame<TPixel> source,
+            ImageFrame<TPixel> destination,
+            Matrix4x4 matrix)
```
antonfirsov: The only issue with these methods (and also with `IDither` ones) is that they are extending the public API surface with arbitrary stuff that I would normally consider an implementation detail. It's hard to imagine any use case for consuming them outside the ImageSharp core library.

We should investigate if we can do a trick similar to the one with `TPixel` -> `PixelOperations<TPixel>` so we can hide them from the public. We can do it in a follow-up PR though, since it's not easy to find a solution (maybe there is none).

src/ImageSharp/Processing/Processors/Transforms/Resize/ResamplerExtensions.cs
```diff
@@ -0,0 +1,227 @@
...
+using SixLabors.ImageSharp.Advanced;
+using SixLabors.ImageSharp.Memory;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Transforms
+{
+    /// <content>
+    /// Extensions for <see cref="IResampler"/>.
+    /// </content>
+    public static partial class ResamplerExtensions
```
antonfirsov: We should also try hiding these classes from the public, while keeping extensibility.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
JimBobSquarePants: It's more about expressing intent than anything. The JIT can ignore our suggestion but I believe we should still declare that we want the method inlined.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs
JimBobSquarePants: Ah that would be good yes.

src/ImageSharp/Processing/Processors/Transforms/IResampler.cs
JimBobSquarePants: Yeah, this is something I tried my very best to avoid in both scenarios. 

The problem I found was with virility. Adding the generic constraints mean they had to be propagated all the way up the call stack. `.Resize()` becomes `.Resize<TResampler>()` and `ResizeOptions` becomes `ResizeOptions<TResampler>`. 

Exposing the extension classes and interface members and calling the method from inside ech implementation was my only workaround. 

Quantize was even worse. I found myself stuck requiring the `TDither` type parameter in the `IImageEncoder`. 

If there's a way around it, it's beyond my capabilities to discover.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResamplerExtensions.cs
JimBobSquarePants: See above.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
antonfirsov: Hmm that makes sense, but in trivial cases like property getters it gets overly verbose and hurts readability IMO. (Technically *all* autoproperies should be marked then.)

antonfirsov:
I think I have an idea, will prototype it in a few days (and also do a more comprehensive review here).

JimBobSquarePants:
@antonfirsov I just had another crack at this.

I've introduced a new interface `IResamplingImageProcessor<TPixel>`

```c#
    /// <summary>
    /// Implements an algorithm to alter the pixels of an image via a resampling transforms.
    /// </summary>
    public interface IResamplingImageProcessor<TPixel> : IImageProcessor<TPixel>
    {
        /// <summary>
        /// Applies a resampling transform with the given sampler.
        /// </summary>
        /// <typeparam name="TResampler">The type of sampler.</typeparam>
        /// <param name="sampler">The sampler to use.</param>
        void ApplyTransform<TResampler>(in TResampler sampler)
            where TResampler : struct, IResampler;
    }
```

`IResampler` now looks like this.

```c#
    /// <summary>
    /// Encapsulates an interpolation algorithm for resampling images.
    /// </summary>
    public interface IResampler
    {
        /// <summary>
        /// Gets the radius in which to sample pixels.
        /// </summary>
        float Radius { get; }

        /// <summary>
        /// Gets the result of the interpolation algorithm.
        /// </summary>
        /// <param name="x">The value to process.</param>
        /// <returns>
        /// The <see cref="float"/>
        /// </returns>
        float GetValue(float x);

        /// <summary>
        /// Applies a transformation upon an image.
        /// </summary>
        /// <param name="processor">The transforming image processor.</param>
        void ApplyTransform<TPixel>(IResamplingImageProcessor<TPixel> processor)
            where TPixel : struct, IPixel<TPixel>;
    }
```
With implementations like this.

```c#
    /// <summary>
    /// The function implements the nearest neighbor algorithm. This uses an unscaled filter
    /// which will select the closest pixel to the new pixels position.
    /// </summary>
    public readonly struct NearestNeighborResampler : IResampler
    {
        /// <inheritdoc/>
        public float Radius => 1;

        /// <inheritdoc/>
        public float GetValue(float x) => x;

        /// <inheritdoc/>
        public void ApplyTransform<TPixel>(IResamplingImageProcessor<TPixel> processor)
            => processor.ApplyTransform(in this);
    }
```
Everything else is now internal. I moved things about a little to reorganize also.

Performance remains the same as the benchmarks.

hypeartist:
> ```cs
>     /// <summary>
>     /// Implements an algorithm to alter the pixels of an image via a resampling transforms.
>     /// </summary>
>     /// <typeparam name="TPixel">The pixel format.</typeparam>
>     public interface IResamplingImageProcessor<TPixel> : IImageProcessor<TPixel>
>         where TPixel : struct, IPixel<TPixel>
>     {
>         /// <summary>
>         /// Applies a resampling transform with the given sampler.
>         /// </summary>
>         /// <typeparam name="TResampler">The type of sampler.</typeparam>
>         /// <param name="sampler">The sampler to use.</param>
>         void ApplyTransform<TResampler>(in TResampler sampler)
>             where TResampler : struct, IResampler;
>     }
> ```

Btw, if I am not mistaken, the method ``` void ApplyTransform<TResampler>``` won't be inlined even in case the interface ``` IResamplingImageProcessor<TPixel>``` is implemented by struct due some JIT limitations. I guess @jkotas could clarify this.

JimBobSquarePants:
I’m not so concerned by that method being inlined (it would be nice). Much more important is that the `TResampler` struct is concrete as that’s what all the actual convolution hot paths use.

antonfirsov:
@JimBobSquarePants the `IResamplingImageProcessor<TPixel>` trick is great, actually better than the stuff I wanted, we should apply it to the ditherers as well in a follow-up PR.

src/ImageSharp/Processing/Processors/Transforms/Automorphic/AffineTransformProcessor{TPixel}.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            this.source = source;
+            this.destination = destination;
+            this.resampler.ApplyTransform(this);
+        }
+
+        /// <inheritdoc/>
+        public void ApplyTransform<TResampler>(in TResampler sampler)
+            where TResampler : struct, IResampler
+            => ApplyAffineTransform(
```
antonfirsov: nit: wouldn't it be simpler to just inline the contents of the method here? I don't see many benefits from the separate `.Transforms.cs` files either.

James-Jackson-South: Yeah, was just being lazy and didn't want to bog the classes down with too much code. Will update.

src/ImageSharp/Processing/Processors/Transforms/Automorphic/AutomorphicTransformUtilities.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Runtime.CompilerServices;
+using SixLabors.ImageSharp.Memory;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Transforms
+{
+    /// <summary>
+    /// Utility methods for affine and projective transforms.
+    /// </summary>
+    internal static class AutomorphicTransformUtilities
```
antonfirsov: I also vote for LinearTransform.

src/ImageSharp/Processing/Processors/Transforms/IResamplingImageProcessor{TPixel}.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Transforms
+{
+    /// <summary>
+    /// Implements an algorithm to alter the pixels of an image via a resampling transforms.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    public interface IResamplingImageProcessor<TPixel> : IImageProcessor<TPixel>
```
antonfirsov: `IResamplingTransformProcessor` to make it more explicit and more consistent with the `ApplyTransform` methods?

src/ImageSharp/Processing/Processors/Transforms/Automorphic/AffineTransformProcessor{TPixel}.cs
JimBobSquarePants: Yeah, was just being lazy and didn't want to bog the classes down with too much code. Will update.

src/ImageSharp/Processing/Processors/Transforms/IResamplingImageProcessor{TPixel}.cs
JimBobSquarePants: Yeah will do.

JimBobSquarePants:
Imma gonna merge this.

PR #1133, opened 2020-02-27
Text segment ReadOnlySpan<byte> initialization

Sergio0694:
### Prerequisites

### Description
This PR includes a couple of minor optimizations:

- Replaced some `static readonly byte[]` arrays with static `ReadOnlySpan<byte>` values that directly map over constant data in the `.text` segment (see [here](https://vcsjones.dev/2019/02/01/csharp-readonly-span-bytes-static/)).
- Removed some bound checks when accessing those static sections (added some DEBUG checks)
- Removed some type initialization calls to `Encoding.ASCII.GetBytes`, no longer necessary
- Switched the `Stream.Write` calls to the `ReadOnlySpan<byte>` overloads.

No rush at all for this, just saw those `byte[]` arrays and thought I'd have some fun optimizing them away. Pinging @antonfirsov for the reviews as he's the official performance guru. 😄

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/ZigZag.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/ZigZag.cs
@@ -77,7 +77,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
             ref byte sourceRef = ref MemoryMarshal.GetReference(Unzig);
             ref byte destinationRef = ref Unsafe.AsRef<byte>(result.Data);
 
-            Unsafe.CopyBlock(ref sourceRef, ref destinationRef, Size);
+            Unzig.CopyTo(new Span<byte>(result.Data, Size));
 
             return result;
         }
--- a/src/ImageSharp/Formats/Jpeg/Components/ZigZag.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/ZigZag.cs
@@ -34,12 +34,11 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
         public fixed byte Data[Size];
 
         /// <summary>
-        /// Unzig maps from the zigzag ordering to the natural ordering. For example,
-        /// unzig[3] is the column and row of the fourth element in zigzag order. The
-        /// value is 16, which means first column (16%8 == 0) and third row (16/8 == 2).
+        /// Gets the unzigs map, which maps from the zigzag ordering to the natural ordering.
+        /// For example, unzig[3] is the column and row of the fourth element in zigzag order.
+        /// The value is 16, which means first column (16%8 == 0) and third row (16/8 == 2).
         /// </summary>
-        private static readonly byte[] Unzig =
-        new byte[Size]
+        private static ReadOnlySpan<byte> Unzig => new byte[]
         {
             0,  1,  8, 16,  9,  2,  3, 10,
             17, 24, 32, 25, 18, 11,  4,  5,
@@ -75,8 +74,11 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
         public static ZigZag CreateUnzigTable()
         {
             ZigZag result = default;
-            byte* unzigPtr = result.Data;
-            Marshal.Copy(Unzig, 0, (IntPtr)unzigPtr, Size);
+            ref byte sourceRef = ref MemoryMarshal.GetReference(Unzig);
+            ref byte destinationRef = ref Unsafe.AsRef<byte>(result.Data);
+
+            Unsafe.CopyBlock(ref sourceRef, ref destinationRef, Size);
+
             return result;
         }
 
--- a/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
+++ b/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
@@ -363,10 +363,20 @@ namespace SixLabors.ImageSharp.Formats.Png.Zlib
         [MethodImpl(InliningOptions.ShortMethod)]
         public static short BitReverse(int toReverse)
         {
-            DebugGuard.MustBeLessThan(toReverse & 0xF, Bit4Reverse.Length, nameof(toReverse));
-            DebugGuard.MustBeLessThan((toReverse >> 4) & 0xF, Bit4Reverse.Length, nameof(toReverse));
-            DebugGuard.MustBeLessThan((toReverse >> 8) & 0xF, Bit4Reverse.Length, nameof(toReverse));
-            DebugGuard.MustBeLessThan(toReverse >> 12, Bit4Reverse.Length, nameof(toReverse));
+            /* Use unsafe offsetting and manually validate the input index to reduce the
+             * total number of conditional branches. There are two main cases to test here:
+             *   1. In the first 3, the input value (or some combination of it) is combined
+             *      with & 0xF, which results in a maximum value of 0xF no matter what the
+             *      input value was. That is 15, which is always in range for the target span.
+             *      As a result, no input validation is needed at all in this case.
+             *   2. There are two cases where the input value might cause an invalid access:
+             *      when it is either negative, or greater than 15 << 12. We can test both
+             *      conditions in a single pass by casting the input value to uint, and checking
+             *      whether that value is lower than 15 << 12. If it was a negative value (2-complement),
+             *      the test will fail as the uint cast will result in a much larger value.
+             *      If the value was simply too high, the test will fail as expected.
+             * Doing this reduces the total number of index checks from 4 down to just 1. */
+            Guard.MustBeLessThanOrEqualTo<uint>((uint)toReverse, 15 << 12, nameof(toReverse));
 
             ref byte bit4ReverseRef = ref MemoryMarshal.GetReference(Bit4Reverse);
 
--- a/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
+++ b/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
@@ -381,14 +381,15 @@ namespace SixLabors.ImageSharp.Formats.Png.Zlib
              *      fail as expected. We can't simply check whether the value is lower than
              *      15 << 12, because higher values are acceptable in the first 3 accesses.
              * Doing this reduces the total number of index checks from 4 down to just 1. */
-            Guard.MustBeLessThanOrEqualTo<uint>((uint)(toReverse >> 12), 15, nameof(toReverse));
+            int toReverseRightShiftBy12 = toReverse >> 12;
+            Guard.MustBeLessThanOrEqualTo<uint>((uint)toReverseRightShiftBy12, 15, nameof(toReverse));
 
             ref byte bit4ReverseRef = ref MemoryMarshal.GetReference(Bit4Reverse);
 
             return (short)(Unsafe.Add(ref bit4ReverseRef, toReverse & 0xF) << 12
                            | Unsafe.Add(ref bit4ReverseRef, (toReverse >> 4) & 0xF) << 8
                            | Unsafe.Add(ref bit4ReverseRef, (toReverse >> 8) & 0xF) << 4
-                           | Unsafe.Add(ref bit4ReverseRef, toReverse >> 12));
+                           | Unsafe.Add(ref bit4ReverseRef, toReverseRightShiftBy12));
         }
 
         /// <inheritdoc/>
--- a/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
+++ b/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
@@ -371,12 +371,14 @@ namespace SixLabors.ImageSharp.Formats.Png.Zlib
              *      As a result, no input validation is needed at all in this case.
              *   2. There are two cases where the input value might cause an invalid access:
              *      when it is either negative, or greater than 15 << 12. We can test both
-             *      conditions in a single pass by casting the input value to uint, and checking
-             *      whether that value is lower than 15 << 12. If it was a negative value (2-complement),
-             *      the test will fail as the uint cast will result in a much larger value.
-             *      If the value was simply too high, the test will fail as expected.
+             *      conditions in a single pass by casting the input value to uint and right
+             *      shifting it by 12, which also preserves the sign. If it is a negative
+             *      value (2-complement), the test will fail as the uint cast will result
+             *      in a much larger value. If the value was simply too high, the test will
+             *      fail as expected. We can't simply check whether the value is lower than
+             *      15 << 12, because higher values are acceptable in the first 3 accesses.
              * Doing this reduces the total number of index checks from 4 down to just 1. */
-            Guard.MustBeLessThanOrEqualTo<uint>((uint)toReverse, 15 << 12, nameof(toReverse));
+            Guard.MustBeLessThanOrEqualTo<uint>((uint)(toReverse >> 12), 15, nameof(toReverse));
 
             ref byte bit4ReverseRef = ref MemoryMarshal.GetReference(Bit4Reverse);
 
--- a/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
+++ b/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
@@ -160,13 +160,9 @@ namespace SixLabors.ImageSharp.Formats.Png.Zlib
             this.Pending.WriteBits(this.distTree.NumCodes - 1, 5);
             this.Pending.WriteBits(blTreeCodes - 4, 4);
 
-            ref byte bitLengthOrderRef = ref MemoryMarshal.GetReference(BitLengthOrder);
-
             for (int rank = 0; rank < blTreeCodes; rank++)
             {
-                ref byte bitsRef = ref Unsafe.Add(ref bitLengthOrderRef, rank);
-
-                this.Pending.WriteBits(this.blTree.Length[bitsRef], 3);
+                this.Pending.WriteBits(this.blTree.Length[BitLengthOrder[rank]], 3);
             }
 
             this.literalTree.WriteTree(this.Pending, this.blTree);
@@ -255,14 +251,11 @@ namespace SixLabors.ImageSharp.Formats.Png.Zlib
             // Build bitlen tree
             this.blTree.BuildTree();
 
-            ref byte bitLengthOrderRef = ref MemoryMarshal.GetReference(BitLengthOrder);
             int blTreeCodes = 4;
 
             for (int i = 18; i > blTreeCodes; i--)
             {
-                ref byte bits = ref Unsafe.Add(ref bitLengthOrderRef, i);
-
-                if (this.blTree.Length[bits] > 0)
+                if (this.blTree.Length[BitLengthOrder[i]] > 0)
                 {
                     blTreeCodes = i + 1;
                 }
--- a/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
+++ b/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
@@ -36,10 +36,6 @@ namespace SixLabors.ImageSharp.Formats.Png.Zlib
 
         private const int EofSymbol = 256;
 
-        // The lengths of the bit length codes are sent in order of decreasing
-        // probability, to avoid transmitting the lengths for unused bit length codes.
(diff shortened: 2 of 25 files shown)


src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
```diff
@@ -363,10 +363,17 @@ public bool TallyDist(int distance, int length)
...
+            DebugGuard.MustBeLessThan((toReverse >> 4) & 0xF, Bit4Reverse.Length, nameof(toReverse));
+            DebugGuard.MustBeLessThan((toReverse >> 8) & 0xF, Bit4Reverse.Length, nameof(toReverse));
+            DebugGuard.MustBeLessThan(toReverse >> 12, Bit4Reverse.Length, nameof(toReverse));
+
+            ref byte bit4ReverseRef = ref MemoryMarshal.GetReference(Bit4Reverse);
+
+            return (short)(Unsafe.Add(ref bit4ReverseRef, toReverse & 0xF) << 12
+                           | Unsafe.Add(ref bit4ReverseRef, (toReverse >> 4) & 0xF) << 8
+                           | Unsafe.Add(ref bit4ReverseRef, (toReverse >> 8) & 0xF) << 4
+                           | Unsafe.Add(ref bit4ReverseRef, toReverse >> 12));
```
antonfirsov: DebugGuard won't save us in production. Where does `toReverse` come from? Can a malicious input result in buffer overflow?

src/ImageSharp/Processing/Processors/Quantization/OctreeFrameQuantizer{TPixel}.cs
```diff
@@ -513,8 +514,12 @@ public int GetPaletteIndex(ref Rgba32 pixel, int level)
                 [MethodImpl(InliningOptions.ShortMethod)]
                 private static int GetColorIndex(ref Rgba32 color, int level)
                 {
+                    DebugGuard.MustBeLessThan(level, Mask.Length, nameof(level));
+
                     int shift = 7 - level;
-                    byte mask = Mask[level];
+                    ref byte maskRef = ref MemoryMarshal.GetReference(Mask);
+                    byte mask = Unsafe.Add(ref maskRef, level);
```
antonfirsov: Question is same for `level` as in my previous comment.

src/ImageSharp/Formats/Jpeg/Components/ZigZag.cs
```diff
@@ -75,8 +74,11 @@ internal unsafe struct ZigZag
         public static ZigZag CreateUnzigTable()
         {
             ZigZag result = default;
-            byte* unzigPtr = result.Data;
-            Marshal.Copy(Unzig, 0, (IntPtr)unzigPtr, Size);
+            ref byte sourceRef = ref MemoryMarshal.GetReference(Unzig);
+            ref byte destinationRef = ref Unsafe.AsRef<byte>(result.Data);
+
+            Unsafe.CopyBlock(ref sourceRef, ref destinationRef, Size);
+
```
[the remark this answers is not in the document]
Sergio0694: Ha, fair enough! Trading coolness for clarity 😄

Done in https://github.com/SixLabors/ImageSharp/pull/1133/commits/b251c005004351c259a88aa827b119f0068b5992.

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
Sergio0694: Sure, I mostly added those to validate the code path in our CI tests.. I had not considered possible attempts to exploit this from a security standpoint, you're right 😅

What if we figured out the upper/lower bounds of those 4 indices, and validated just those?
It would still result in just 2 conditional jumps, instead of 4 (one for each original access).

Actually, now that I think about it:

1. The first three indices are always computed with a final `& 0xF`, which means the maximum possible value is `0xF`, so `15`, which is always in range. So those first 3 accesses should always be necessarily valid, without the need to test them
2. That last `>> 12` could result in an out of bound index, either if the value is negative and below `-1 << 12`, or if it's a positive value greater than `1 << 12`. I guess we could just check this last one just to be extra sure, as I don't know where exactly the input is coming from here, as you mentioned.

That's still just a single conditional branch compared to 4, so... Yay? 😁
What do you think?

src/ImageSharp/Processing/Processors/Quantization/OctreeFrameQuantizer{TPixel}.cs
Sergio0694: So, this method is either called with a fixed `level` value that's in range (`0` or `1`), or with an incremental value for each node down the octree, stopping when a leaf is reached. I think the processor/octree is supposed to never have more than 8 levels in total, plus it seems to be something that's built internally by the lib and not just some sequence of arbitrary bytes read by a stream that an attacked could exploit, so that should not be an issue. Still though, if you feel like the extra safety is worth the conditional branch here, I can revert this if you want 👍

antonfirsov: Seems to work in theory, but gut says we need tests which are stressing the depth of the tree to be confident here, which we don't have at the moment. @JimBobSquarePants thoughts?

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
antonfirsov: Sounds good, let's do it like that!

Sergio0694: Awesome, done in https://github.com/SixLabors/ImageSharp/pull/1133/commits/057769037dd01747f93500d14ab0d67c99904566. Also added a detailed comment for future reference 😊

src/ImageSharp/Processing/Processors/Quantization/OctreeFrameQuantizer{TPixel}.cs
JimBobSquarePants: There's an absolute maximum 8 levels determined by the algorithm itself.

 I actually planned to do this very same change with my upcoming quantization and dithering cleanup. 👍

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
```diff
@@ -40,8 +40,6 @@ internal sealed unsafe class DeflaterHuffman : IDisposable
         // probability, to avoid transmitting the lengths for unused bit length codes.
         private static readonly int[] BitLengthOrder = { 16, 17, 18, 0, 8, 7, 9, 6, 10, 5, 11, 4, 12, 3, 13, 2, 14, 1, 15 };
```
JimBobSquarePants: We might well be better off using the `ReadOnlySpan<byte>` here also and casting. 

I know we do that in the jpeg encoder for `uint` and `BitCountLut` with, as I recall, a net gain in performance.

Sergio0694:
@JimBobSquarePants Whoops, fixed in https://github.com/SixLabors/ImageSharp/pull/1133/commits/68387a7737054cf0b715bd0f9b0b407745faece0, my bad 😅

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/1133/commits/5b7ac75dfa637f40763b75bcd378205a79394e5d.
Should I also add some bounds check here, or are all those accesses guaranteed to be safe?

JimBobSquarePants: I couldn't tell you for certain I'm afraid. That port was an act of desperation over understanding.

Sergio0694: I see. Should I just switch those accesses to a direct `Span<T>[int]` access then? 🤔

antonfirsov: If we lack understanding, we shall go for safety.

Btw if we use the data in `Bit4Reverse` for indexing, elements shouldn't be `byte`-s. It's more work for the CPU:
https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcAMACbckB0ASgK4B2wAllAKYEDCA9lAA6UA2NATgMrcBulAMY0IAbnTo4AZjwJs9bAG90uXKrUsulfmGA08SAGzYARgE992AJJkAJjQAet4NIQAKLjQBmZywZYAGmxKChD7JwBKDVwVNDUEvAB2bABVMggwbzoAQTs7Tx9sIPCHR0iJeLUAXxjsOq0dPQN8EwsrWzKAIX9C33aA4IHSqLq4xLU4FPTM7II8gq9fEtCyirratGqgA==

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
antonfirsov: My comment on `BitLengthOrder` also applies here. Additionally, if we already on the microoptimization road, we should cache the value of  `toReverse >> 12` to a local, since we are using it twice.

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
Sergio0694: @antonfirsov not sure what you mean there 🤔

We can't use the `static ReadOnlySpan<T>` trick with `int` values, as those are not supported. By that I mean, you'd get a new array being allocated every time, that's because `byte` is not endianness-aware, while `int` is not. As for the actual value being used for offsetting, I checked all the various `Unsafe.Add` calls, and the offsetting index is always passed as an `int` value, not as a `byte`. Could you clarify what you meant there? I'm not following 😅

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/1133/commits/4b0dfd11c0267905ea8f004833db2618bf4b2ab7.

Also I should really stress out how if we care about micro-optimizations, we should **really** consider refactoring our `Guard` APIs with that trick I suggested. Look at all that asm clutter due to that `string.Format` call we have here: 

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8B6YgAgGEIAHATygEsBzACwzIAowBKMgZQYIyAGWzBouMtgB2AEzKRpGRsACuGCQDoAsAChSIhmBjTcMearkwoZDCxhkAgtWxh7h46ZhoyANWu4DBDSZABMmgAMOrp6AAKhAIxxEWSxCZoAIgzYTNIQuBhGuADcyanpWTl5BUWaVLIwjtLYADa0gSVlaZoASpaF+DB1EPjUDC3WfNYAbkYwnbqxKd19SgyDmgCSStY0U1CzxgtxAMypoWQZMABmLdgY1gASqtfX+DJ6AN56ZL9k1IxpvcHGkkGQejBsLIAPLSNp8FzSAA8wFoDwAfGQAEIMDAoCHTAIOAC8mOkMAA7mRUQ8ANoAXTInzIER8AA4fOgyAlQj5edzWWQkD4ElyEj4AJw+ACsIpOPnl3PFZAA7CLpWQAL6lXR/Mg/P6kchI3CqfDvKC0dEG35G8EwQlQMy2dzAXGSCDXKTcsFu9hAlqqIY2sh2pHEU3m7CW626w0GJEuKDYfBkZqDYkAIg0BKJmfRvlaQdsEDIsEdzr9uHDSZTsb1YdiKtwBaLDgpuJY1PdZYdRNk4abLZDtIAsjA7BBZJtRi0OOPJ9PZ9DqIVgrhNI4mExYLhAoTti0GNJj0xuPSQ7EzqCyLgWNB2DiMLmnTAOMf2Dm+6/uCHvnG9V+D8SxfMwemYNg+BYBhrgwLFaB5MhiRA79nXRTEeR1QC/gAcVUaNZE0UdVAKLEYGEeZcAAFRYGRoSgABRABHfCWioiAkVUD90Q4DguKUbgvwrGBwNYDAoJguCEP5BJZTTFMYE9DghKJbhuB1EM9VgL0aQcP18VQkSbiQ3svXHfBoFoUdozvVpNBwicIWuawTGMDgnwM4T1L0TS/ibTg7wfbgOAAVVMbBnM3WRZA4bTuzxUCjOuHwVNfMgADIWQQAAxXgkSRblQl87CSr1QgyDC3AIqGRxoti4z9MSpyfGUiBErIdCyBQXhMoiHK8oKtlitK0rysq6qopiuLGsM5rOFStDMTZHqstysh8q64aRuwsbwsi2qpoa3FPKJOaFpEiDxOg2D4J5NSsL+TU9CemJ9ATSMLStPQ7QABSgCBZgaSRBkXSQNH+f6HjAdgcmwY8CjIY8AwYeRaxBgJojDCMzU+2NaSuNRt0mB5qBo/7VFYC9dFpBiEDAQMGmy/78HqGAqEdHIYCpj9rGaFpyjBJNClac4yDwgivhDMMPujL6ANDAx/EYa4GHmF17hdBxcGoGAwBg1X5ADYsGEkCY93VkJoDIGAWJF8HsDId4EHWM0yCNmApYMGR5Dsf6KUkGRrbpmBVyCEIYMR9gTbTCAMGiBsE2xqMY09400fkjNM3d/MaIcDBoyYCc3bbHwKWgtxb3vVQWnkYAHGR2RgUImtozrVP1vT9MYCzJ2c/cJ2XdTd3NBb5N8HreM09b1Mu6zNGJ2sAA5BS+4cLuyE9TX/mnhebDsDXo/BuuFHsMAAGtzBH4g0Yn20EwwWgdc7hSsyowtAxgVfbEfhxN7seu2xXwfk/aet8FbGhgMHUOwQFDaSzI4KATAzQmAwLTYw0DpD5nbomaecVZ5ZzbJmYgmJo47khA8PetEQj/0dtgZ2+BXbD2wcQSB6C1zSDAWOCc94lzUDnAuHhM4+ErnYRuLcZC9wMAPHCY8p5zyXmvAkMEsQUBkGIqRcilEaJ0UYrbNiHE35th4oYj+xcP4+BMcWJ2Pg0gpHnhQ5egxfzyz1GXFyZBLEOBAGQTYVBRit2ABMJEniwH/lKhHDgw8/FJhgOxDgTteCYgiM40qYTtoeJYH7BBSDBhKGhOoaE1wegyELmgkO7COD2KXgpHwAASTM79iyfHdpqR2JF2DHzNmDKhG8bA21YiWJkTtNSaEzN5Fxj0QwvRHAIqcQj+HcLmcuDBG5F4QEPLI6QZ4qZ6gBFI4EAtUiqLJhACk2TkF5IKUUkpMAykYI4LY7eY9d6OO8OUFIgw9yc14KSYavtTlpkpE4RBFyMD5LBdcrZtyoEVKqVAV5PhPlVULuMl6QA.

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
antonfirsov: Not an ASM expert, so the terminology might be incorrect, but looks something like this:
Byte can't be directly used to address memory or added to a memory address, therefore it has to be loaded into another register first, thus the extra `movzx eax, dl` in my gist.

A possible workaround is to create `ReadOnlySpan<byte>` of `*4` size with little padding, and unsafe cast it, but then you have to deal with endianness, which will be super-complicated. An other solution is to not use the `.text` block, for these blobs since, you only save 1 `memcpy` at initialization time with `.text`, while indexing will be slower each and every time invoking the methods.

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
antonfirsov: Well this is just terrible, wish we have noticed that earlier. `Guard.MustBeLessThanOrEqualTo` will make this method actually slower than it was before your changes. Can you just use a manual check + ThrowHelper here?

If you have some time, feel free to file a PR in [SixLabors/SharedInfrastructure](https://github.com/SixLabors/SharedInfrastructure). For the comparison methods I'd rather see a set of duplicate non-generic methods, than `TValue : IComparable<TValue>`, which also brings some extra overhead I believe.

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
Sergio0694: I get what you're saying now, yeah 👍
I'm not sure I understand where you're seeing that with `Bit4Reverse` though.
Here's where that is used:
https://github.com/SixLabors/ImageSharp/blob/4b0dfd11c0267905ea8f004833db2618bf4b2ab7/src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs#L384-L392

All the indexing values there are `int` already 🤔
I mean, those are actually the same indices that were being used before as well. I'm not sure I'm following here, where is the `byte` offsetting peformed?

antonfirsov: Ahh sorry, I was wrong the values for `Bit4Reverse` are not used for indexing into data. This is the data that gets indexed.

We should deal with Guard however to make sure we don't regress. Do you plan to fix it globally in SixLabors/SharedInfrastructure, or will you replace it in this PR with a manual check + ThrowHelper calls?

Sergio0694: Ahahah no problem, glad we could figure this out, I was wondering what was I missing there 😄

As for `Guard`, I'd say that's an unrelated improvement, since it's not something that will impact the functionality in any way, it's just a matter of further optimizations. I guess the best thing would be to just make another PR and fix the entire `Guard` class, with all the included APIs. Sounds good?

antonfirsov: If you can do it, that would be the best!

Sergio0694: Sure, always happy to bring more of my discoveries and optimizations to ImageSharp! 😄

JimBobSquarePants:
@Sergio0694 @antonfirsov Are we waiting for this to be completed and merged before finishing here?

https://github.com/SixLabors/SharedInfrastructure/pull/7

Sergio0694:
@JimBobSquarePants Not sure whether we necessarily have to wait for that. I mean, having that would reduce the overhead in those checks, but it wouldn't actually change anything semantically. We might want to wait for it though in case the changes in this PR alone have too much of a performance regression, but that shouldn't be the case (plus it'd be temporary anyways).

Up to you guys 👍

antonfirsov:
@JimBobSquarePants I expect the aggregate result of the changes in `DeflaterHuffman.BitReverse()` to be a regression, unless we finish SixLabors/SharedInfrastructure#7 or replace `Guard` with verbose local code.

On the other hand, I'm fine fixing the regression in a follow-up PR.

JimBobSquarePants:
@Sergio0694 I found the AccessViolation.

https://github.com/SixLabors/ImageSharp/pull/1133/checks?check_run_id=484873388#step:11:698

It's not us it's the ImageMagick reference codec. @dlemstra You'll want to see this.

Sergio0694:
@JimBobSquarePants That's awesome, great work! Props to your investigative skills 😄
Also glad to hear it's not actually coming from `ImageSharp`, that's perfect.

JimBobSquarePants:
Ok.... The submodule is now updated and I've added updated relevant tests.

I've also configured the solution to build the T4 templates on build since we don't import the transformed file.

Will merge once this finishes building.

tests/ImageSharp.Tests/Image/ImageFrameCollectionTests.NonGeneric.cs
```diff
@@ -90,7 +90,7 @@ public void AddNewFrame_Frame_FramesNotBeNull()
                         this.Collection.AddFrame(null);
                     });
 
-                Assert.StartsWith("Value cannot be null.", ex.Message);
+                Assert.StartsWith("Parameter \"source\" must be not null.", ex.Message);
```
dlemstra: Could we also do ` Assert.Throws<ArgumentNullException>("source", () =>` here instead?

JimBobSquarePants: I thought that was dealt with above?

PR #1138, opened 2020-03-03
Dither and Quantize API and Performance Improvements.

JimBobSquarePants:
### Prerequisites

### Description

The [`IDither`](https://github.com/SixLabors/ImageSharp/blob/ac6b6a90715d22225058271c0a686080a1d319b4/src/ImageSharp/Processing/Processors/Dithering/IDither.cs) interface contained implementation specific parameters in the method signatures. This has been simplified to the absolute bare minimum possible parameters. I also managed to remove boxing from the palette dither type.

In addition I simplified the [`IFrameQuantizer`](https://github.com/SixLabors/ImageSharp/blob/ac6b6a90715d22225058271c0a686080a1d319b4/src/ImageSharp/Processing/Processors/Quantization/IFrameQuantizer%7BTPixel%7D.cs) interface to remove palette reference ambiguity in the API.

I noticed some performance problems with the `GifEncoder` during development and as part of my refactoring I tackled some of those issues. Mean performance is around 3-4X faster with 25% less allocations on 2 out of 3 targets. (I have no idea what is so different in NET Core 2.1 memory usage)

_Note: Allocations are due to usage of `ConcurrentDictionary<TPixel,int>` in the pixel map which cannot be avoided. I [normalized the concurrency count across target frameworks](https://github.com/dotnet/corefx/commit/dd6e8994dbce7ffebb02b4b053e8ad505b7cd573) to at least ensure allocations do not go wild on NETFX._

I've introduces some `Unsafe.Add` in the dithering and quantization loops but safe in the knowledge that bounds are carefully constrained long before those methods are called. 

**Before**

|               Method |       Runtime |     Mean |      Error |   StdDev | Ratio | RatioSD |     Gen 0 |    Gen 1 |    Gen 2 |  Allocated |
|--------------------- |-------------- |---------:|-----------:|---------:|------:|--------:|----------:|---------:|---------:|-----------:|
| 'System.Drawing Gif' |    .NET 4.7.2 | 12.93 ms |  23.910 ms | 1.311 ms |  1.00 |    0.00 |  234.3750 |        - |        - | 1093.11 KB |
|     'ImageSharp Gif' |    .NET 4.7.2 | 86.65 ms |  14.582 ms | 0.799 ms |  6.74 |    0.60 | 1142.8571 | 714.2857 | 285.7143 |  8155.4 KB |
|                      |               |          |            |          |       |         |           |          |          |            |
| 'System.Drawing Gif' | .NET Core 2.1 | 12.09 ms |   2.494 ms | 0.137 ms |  1.00 |    0.00 |  234.3750 |        - |        - | 1089.88 KB |
|     'ImageSharp Gif' | .NET Core 2.1 | 97.67 ms | 128.094 ms | 7.021 ms |  8.08 |    0.55 |  800.0000 | 200.0000 |        - |  737.43 KB |
|                      |               |          |            |          |       |         |           |          |          |            |
| 'System.Drawing Gif' | .NET Core 3.1 | 12.01 ms |   1.054 ms | 0.058 ms |  1.00 |    0.00 |   15.6250 |        - |        - |  198.06 KB |
|     'ImageSharp Gif' | .NET Core 3.1 | 71.03 ms |  26.512 ms | 1.453 ms |  5.91 |    0.13 | 1142.8571 | 714.2857 | 285.7143 | 8123.59 KB |

**After**

|               Method |       Runtime |     Mean |     Error |   StdDev | Ratio | RatioSD |     Gen 0 |    Gen 1 |    Gen 2 |  Allocated |
|--------------------- |-------------- |---------:|----------:|---------:|------:|--------:|----------:|---------:|---------:|-----------:|
| 'System.Drawing Gif' |    .NET 4.7.2 | 10.66 ms |  1.325 ms | 0.073 ms |  1.00 |    0.00 |  234.3750 |        - |        - | 1093.01 KB |
|     'ImageSharp Gif' |    .NET 4.7.2 | 27.68 ms |  7.814 ms | 0.428 ms |  2.60 |    0.05 | 1031.2500 | 625.0000 | 281.2500 | 6509.24 KB |
|                      |               |          |           |          |       |         |           |          |          |            |
| 'System.Drawing Gif' | .NET Core 2.1 | 11.01 ms | 10.751 ms | 0.589 ms |  1.00 |    0.00 |  234.3750 |        - |        - | 1089.88 KB |
|     'ImageSharp Gif' | .NET Core 2.1 | 24.18 ms |  3.775 ms | 0.207 ms |  2.20 |    0.10 | 1031.2500 | 656.2500 | 312.5000 |  740.78 KB |
|                      |               |          |           |          |       |         |           |          |          |            |
| 'System.Drawing Gif' | .NET Core 3.1 | 10.49 ms |  1.687 ms | 0.092 ms |  1.00 |    0.00 |   15.6250 |        - |        - |  197.95 KB |
|     'ImageSharp Gif' | .NET Core 3.1 | 27.27 ms | 41.280 ms | 2.263 ms |  2.60 |    0.23 | 1031.2500 | 625.0000 | 281.2500 | 6453.76 KB |

A slight fix to the color matching code was implemented also with led to more accurate matching and distribution of error.


antonfirsov:
`PaletteFrameQuantizer<T>` and `EuclideanPixelMap<T>` were being around here for a while, but unfortunately I did not have a chance to look at the implementation before, so sorry for bringing this question to the table this late, but we need to carefully think this over.

Is the "all memory in for speed" approach really what we want with Gif? Just pushed a change to also run the benchmark against `rgb-48bpp.png`, which is an image of wider color variety:

```
|               Method |         TestImage | Ratio |     Gen 0 |     Gen 1 |    Gen 2 |   Allocated |
|--------------------- |------------------ |------:|----------:|----------:|---------:|------------:|
| 'System.Drawing Gif' |       Bmp/Car.bmp |  1.00 |   15.6250 |         - |        - |   197.95 KB |
|     'ImageSharp Gif' |       Bmp/Car.bmp |  2.33 | 1093.7500 |  656.2500 | 312.5000 |  6453.27 KB |
|                      |                   |       |           |           |          |             |
| 'System.Drawing Gif' | Png/rgb-48bpp.png |  1.00 |         - |         - |        - |   393.23 KB |
|     'ImageSharp Gif' | Png/rgb-48bpp.png |  2.91 | 3250.0000 | 1500.0000 | 500.0000 | 23105.69 KB |
```

Note that the theoretical maximum `distanceCache` for `Rgba32` is `256^4 ~= 4 billion` (!) elements.

There are alternative approaches that will sacrifice some of the speed but allocate much less. The most naive thing I can think of: binary search, probably ~10x slower, could be done without going down the rabbit hole researching literature. Do you remember how much slower was the linear search without caching? Can't find the PR introducing the cache, must be older than history itself :)

antonfirsov:
Ok, I realized binary search won't be enough, so I guess can't find a good solution in a reasonable time :(

Sergio0694:
Looks like this could be the perfect use case scenario for a `ConcurrentDictionarySlim<TKey, TValue>` collection, if it existed. If we care about memory usage, I'm wondering whether maybe it'd be worth it to run some tests with a `DictionarySlim<TKey, TValue>` (see [here](https://github.com/dotnet/corefxlab/blob/master/src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/DictionarySlim.cs)) that is accessed either by a `lock`, or maybe even with just a shared `SpinLock`? Memory usage should go down, especially because `DictionarySlim<TKey, TValue>` doesn't box value types, plus it's meant for types with a fast `GetHashCode` implementation, so it doesn't store a copy of the code in the internal table, which saves an additional 4 bytes per entry. Could be worth looking into maybe? 🤔

antonfirsov:
These micro-optimizations do not worth the efforts now IMO, we should rather focus on finding a better algorithm. An efficient k-d tree implementagion may do the job. [This one](https://github.com/eric-regina/Supercluster.KDTree) looks promising. We can actually combine it with other techniques. I have an idea now, and I may be able to code it in a couple of hours. But it'sin this PR I'll focus on API review from now on.

antonfirsov:
Regarding locking vs ConcurrentDictionary: we can get rid of locking entirely, we just need a first pass to fill the dictionary.

JimBobSquarePants:
@antonfirsov  @Sergio0694  I would personally love to see a better implementation. I deliberately kept the `EuclideanPixelMap<T>` internal because I was not happy with the approach. 

Looking forward to seeing what you can come up with!

tests/ImageSharp.Tests/Formats/Gif/GifEncoderTests.cs
```diff
@@ -25,6 +26,23 @@ public class GifEncoderTests
             { TestImages.Gif.Ratio4x1, 4, 1, PixelResolutionUnit.AspectRatio }
         };
 
+        [Theory]
+        [WithFile(TestImages.Bmp.Car, PixelTypes.Rgba32)]
+        public void EncodeAllocationCheck<TPixel>(TestImageProvider<TPixel> provider)
```
antonfirsov: I usually prefer to put these under `ImageSharp.Tests/ProfilingBenchmarks`, and keep them skipped.

src/ImageSharp/Processing/Processors/Quantization/IFrameQuantizer{TPixel}.cs
```diff
@@ -31,26 +31,23 @@ public interface IFrameQuantizer<TPixel> : IDisposable
...
 
         /// <summary>
         /// Builds the quantized palette from the given image frame and bounds.
         /// </summary>
         /// <param name="source">The source image frame.</param>
         /// <param name="bounds">The region of interest bounds.</param>
-        /// <returns>The <see cref="ReadOnlyMemory{TPixel}"/> palette.</returns>
-        ReadOnlyMemory<TPixel> BuildPalette(ImageFrame<TPixel> source, Rectangle bounds);
+        /// <returns>The <see cref="ReadOnlySpan{TPixel}"/> palette.</returns>
+        ReadOnlySpan<TPixel> BuildPalette(ImageFrame<TPixel> source, Rectangle bounds);
```
antonfirsov: Can't find any related [guideline point](https://docs.microsoft.com/en-us/dotnet/standard/memory-and-spans/memory-t-usage-guidelines), but I have a bad feeling returning a Span as a "result" of an operation, because the ownership of the backing memory is not clear from the API.

I think we should return `void` here, and have a separate `Palette` property instead.

tests/ImageSharp.Tests/Formats/Gif/GifEncoderTests.cs
JimBobSquarePants: I didn't actually mean to check that in!

src/ImageSharp/Processing/Processors/Quantization/IFrameQuantizer{TPixel}.cs
JimBobSquarePants: You've hit upon a point that led to me coding about 5 different version of this. 

I struggled with a `Palette` property for a couple of reasons:

1. The result is temporally coupled since it is not assigned on construction. The palette could be null (If using `ReadOnlyMemory<TPixel>`) or empty `ReadOnlySpan<TPixel>`
2. Using `ReadOnlyMemory<TPixel>` can lead to a larger palette length than actually required.

The `OctreeFrameQuantizer` doesn't return the final number of entries until you build the palette. You have to build it then slice the result reducing the length by the determined index count, returning `ReadOnlySpan` allowed this. 

https://github.com/SixLabors/ImageSharp/pull/1138/files#diff-2e327e0f312f1aae943b78b1145bd8adR90

If you don't do this you can end up with a larger Gif than you actually need. The `WuFrameQuantizer` is similar, but at least you know the length early on.

I'll give everything on last look over and see if I can improve things.

antonfirsov: > The palette could be null (If using `ReadOnlyMemory<TPixel>`) or empty `ReadOnlySpan<TPixel>`

We can also throw an `InvalidOperationException` if the palette has not yet been built.

> Using `ReadOnlyMemory<TPixel>` can lead to a larger palette length than actually required. [...] You have to build it then slice the result reducing the length by the determined index count, returning `ReadOnlySpan` allowed this. 

You can also slice a `ReadOnlyMemory<T>`.

This is no strong opinion however, we can also clarify the ownership model in the docs of `BuildPalette()`.

JimBobSquarePants: > You can also slice a ReadOnlyMemory<T>.

Why oh why has that never occurred to me before?

JimBobSquarePants: Quick question.... 

Would you consider is ok API wise for the QuantizedFrame to reference the FrameQuantizer palette rather than copying it? Currently I pass the span, copy it and use it for the lifetime of the frame.

JimBobSquarePants: Scratch that. It leads to messy code in the Gif Encoder.

JimBobSquarePants:
@antonfirsov I gave it another pass and cleaned it up quite a lot. 

Still not happy with perf and allocations but it's incrementally improving.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -546,59 +545,54 @@ private void WriteHeaderChunk(Stream stream)
...
-                        byte alpha = rgba.A;
+            ref Rgb24 colorTableRef = ref MemoryMarshal.GetReference(MemoryMarshal.Cast<byte, Rgb24>(colorTable.GetSpan()));
+            ref byte alphaTableRef = ref MemoryMarshal.GetReference(alphaTable.GetSpan());
 
-                        Unsafe.Add(ref colorTableRef, offset) = rgba.R;
-                        Unsafe.Add(ref colorTableRef, offset + 1) = rgba.G;
-                        Unsafe.Add(ref colorTableRef, offset + 2) = rgba.B;
+            // Bulk convert our palette to RGBA to allow assignment to tables.
+            // Palette length maxes out at 256 so safe to stackalloc.
+            Span<Rgba32> rgbaPaletteSpan = stackalloc Rgba32[paletteLength];
```
antonfirsov: In theory, yes, but I can't spot any safety mechanism to enforce this. We should at least `Guard` against it, stackallocking a non-const buffer is [super unsafe](https://vcsjones.dev/2020/02/24/stackalloc/).

Another option is to stackalloc when `paletteLength < 64` and create an array otherwise. We should not stackalloc `256 * sizeof(Rgba32) == 1KB` data anyways.

src/ImageSharp/Processing/Processors/Quantization/WuFrameQuantizer{TPixel}.cs
```diff
@@ -634,7 +632,7 @@ private void Mark(ref Box cube, byte label)
         /// </summary>
         private void BuildCube()
         {
-            Span<double> vv = stackalloc double[this.colors];
+            Span<double> vv = stackalloc double[this.maxColors];
```
antonfirsov: See previous comment.

JimBobSquarePants: Ah yes, that's too big! 2048!!

src/ImageSharp/Formats/Png/PngEncoderCore.cs
Sergio0694: Also small sidenote, keep in mind that `stackalloc` zeroes out memory right now, plus the current implementation is quite inefficient since it's basically just a busy loop with a series of `push 0x0`, plus additional code to handle stack overflows etc.

Eg. doing `stackalloc int[256]` produces (among other things) this loop:

```asm
L001c: mov eax, 0x40
L0021: push 0x0
L0023: push 0x0
L0025: dec rax
L0028: jnz L0021
```

Check out the complete sample code here, it actually doesn't look that great right now: https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunHjykDAJYA7DAwCyACgCUXHp2o9lDAMoAHbCIA8ojAD4GuTSIYBeIxmxgA1tgA29iGGFiA2qQCsSALoBueRUGQJViAHYjEzcABn8QgF8aeKA==

You might want to consider just renting an array (which also skips the memory clear), especially if using `stackalloc` could introduce other security/safety concerns.

There's a proposal for a `[SkipLocalsInit]` attribute for C# 9.0 I think, but it's still under prototyping. And the zeroing optimization Ben Adams is doing is still an open PR possibly coming with .NET 5.

JimBobSquarePants: Palette size is limited in a couple of places...
1. In the setter for `QuantizerOptions`.
2. In the constructor for `IndexedImageFrame`.

I can use the constant again and slice though if you want...

I thought 1Kb was the general rule for a max amount, should we be more conservative?

From the linked blog post.
> I won’t prescribe anything specific, but anything larger than a kilobyte is a point of concern.

antonfirsov: Ok, I see now that `IndexedImageFrame` constructor should  prevent anything going wrong.

antonfirsov: 1Kb is the maximum, but probably not an optimum, or at least not always. But it doesn't really matter in `WritePaletteChunk`, it's not a hot path, just converting & copying a short slice of data.

JimBobSquarePants: I think I'll just revert stackalloc for now looking at the codegen.

Sergio0694: I knew that would've scared you 🤣

antonfirsov: You can always make things [even more scary](https://sharplab.io/#v2:EYLgdgpgLgZgHgGiiAhgZwLYB8ACAmARgFgAoHAZgAJ9KBhSgb1MpeqpwIDZKBLMKSgFkAFAEpmrJiVYzKAZQAOKMAB4+UAHyU0SsJQC82qCgDGAaxQAbSwHsTvfgG08AVk4BdANwTZlH7JwAdm1dRwAGL38AX1IooA=) with the Legacy JIT 😆

Sergio0694: I've run that legacy asm through a decompiler, I got this:
```csharp
public static int M()
{
    Span<int> span = stackalloc int[256];
    CalculateMassOfTheSun();
    return span[0];
}
```

antonfirsov:
We should probably merge this, and continue further experiments from here. It's likely that we can introduce all necessary improvements without breaking the current API.

PR #1145, opened 2020-03-14
Fix edge detection and reenable entropy crop tests

JimBobSquarePants:
### Prerequisites

### Description
Edge detection now produces fully opaque output. This stabilizes the output across platforms and also fixes entropy cropping.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Filters/OpaqueProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Filters/OpaqueProcessor{TPixel}.cs
@@ -1,0 +1,67 @@
+// Copyright (c) Six Labors and contributors.
+// Licensed under the Apache License, Version 2.0.
+
+using System;
+using System.Numerics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using SixLabors.ImageSharp.Advanced;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Filters
+{
+    internal sealed class OpaqueProcessor<TPixel> : ImageProcessor<TPixel>
+          where TPixel : unmanaged, IPixel<TPixel>
+    {
+        public OpaqueProcessor(
+            Configuration configuration,
+            Image<TPixel> source,
+            Rectangle sourceRectangle)
+            : base(configuration, source, sourceRectangle)
+        {
+        }
+
+        protected override void OnFrameApply(ImageFrame<TPixel> source)
+        {
+            var interest = Rectangle.Intersect(this.SourceRectangle, source.Bounds());
+
+            var operation = new OpaqueRowOperation(this.Configuration, source, interest);
+            ParallelRowIterator.IterateRows<OpaqueRowOperation, Vector4>(this.Configuration, interest, in operation);
+        }
+
+        private readonly struct OpaqueRowOperation : IRowOperation<Vector4>
+        {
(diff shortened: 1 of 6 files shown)


antonfirsov:
I tried to follow the discussions under #1143, but maybe missed a few points. What is the main reason you've chosen this vs premultiplication?

I wonder how does [OpenCV](https://docs.opencv.org/trunk/da/d22/tutorial_py_canny.html) or [ImageMagic](https://imagemagick.org/discourse-server/viewtopic.php?t=25405) work for images with alpha. @dlemstra do you know or can you quickly check it?

src/ImageSharp/Processing/Processors/Filters/OpaqueProcessor{TPixel}.cs
```diff
@@ -0,0 +1,67 @@
...
+                PixelOperations<TPixel>.Instance.ToVector4(this.configuration, targetRowSpan.Slice(0, span.Length), span, PixelConversionModifiers.Scale);
+                ref Vector4 baseRef = ref MemoryMarshal.GetReference(span);
+
+                for (int x = 0; x < this.bounds.Width; x++)
+                {
+                    ref Vector4 v = ref Unsafe.Add(ref baseRef, x);
+                    v.W = 1F;
+                }
+
+                PixelOperations<TPixel>.Instance.FromVector4Destructive(this.configuration, span, targetRowSpan, PixelConversionModifiers.Scale);
```
antonfirsov: Yet another place where an extended `PixelTypeInfo` would be super useful to eliminate unnecessary scanning. [Related discussion](https://github.com/saucecontrol/PhotoSauce/pull/39#discussion_r392411909).

JimBobSquarePants: Yup.

JimBobSquarePants:
@antonfirsov We do use premultiplication but then reassign the alpha value from the source before unpremultiplying. If you don't do that you end up with a blank image if you have a consistent alpha channel. 

Basically the only thing I changed was to normalize the alpha component before the algorithm runs. 

OpenCV does the [same](https://stackoverflow.com/questions/49071592/opencv-edge-detection-on-png-image-with-alpha-layer). In the link you can see it's picking up changes that would not exist without removing the alpha component first.

PIL doesn't do this and will produce an [entirely transparent image](https://github.com/python-pillow/Pillow/issues/4321). (However in that closed issue they've misidentified the result as having a zero alpha channel, when, in fact, the image has no information in any channels.

src/ImageSharp/Processing/Processors/Filters/OpaqueProcessor{TPixel}.cs
```diff
@@ -0,0 +1,67 @@
...
+using System;
+using System.Numerics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using SixLabors.ImageSharp.Advanced;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Filters
+{
+    internal sealed class OpaqueProcessor<TPixel> : ImageProcessor<TPixel>
```
antonfirsov: Wouldn't be `MakeOpaqueProcessor` better? (Related extension to be added later: `image.MakeOpaque()`.)

JimBobSquarePants: We can do that if/when we choose to expose such an extension.

Issue #1110, opened 2020-02-09
Document exceptions correctly


antonfirsov:
- [x] `Image.Load` overloads are marked as throwing `UnknownImageFormatException` however when the image content is incorrect, we throw `ImageFormatException` instead. We should consider defining, utilizing and documenting a new `InvalidImageContentException` for these cases. Alternatively we can just make the xmldocs to refer to `ImageFormatException`, as an easier solution, but having a separate `ImageFormatException` subclass seems more correct to me.
- [x] In `Image.FromFile.cs` and `Image.FromBytes.cs` we do not document the above exceptions at all.
- [ ] `ImageProcessingException` is not documented on the API surface. I would suggest to add it to: `Mutate`, `Clone` and `IImageProcessor<TPixel>.Execute`, however I haven't analyzed the call hierarchy whether it is the correct thing to do.
- [x] Review our API if there are any other missing exception docs.

If we decide to indroduce `InvalidImageContentException`, that part should be probably done for RC1, since it is a breaking change.

antonfirsov:
Also noticed many `Image.Load` overloads and other public methods, where we are not validating the input at all.

PR #1146, opened 2020-03-14
Add overloads. Fix #1144

### LINKED ISSUE #1144: Add Image.Identify(string path, ...) overloads
See:
https://github.com/saucecontrol/PhotoSauce/pull/39#discussion_r392410253

JimBobSquarePants:
### Prerequisites

### Description
Adds additional overloads to `Image.Identify` to handle both file paths and byte arrays.

### DIFF
--- a/src/ImageSharp/Image.FromFile.cs
+++ b/src/ImageSharp/Image.FromFile.cs
@@ -31,10 +31,50 @@ namespace SixLabors.ImageSharp
         /// <returns>The mime type or null if none found.</returns>
         public static IImageFormat DetectFormat(Configuration config, string filePath)
         {
-            config = config ?? Configuration.Default;
+            config ??= Configuration.Default;
             using (Stream file = config.FileSystem.OpenRead(filePath))
             {
                 return DetectFormat(config, file);
+            }
+        }
+
+        /// <summary>
+        /// Reads the raw image information from the specified stream without fully decoding it.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


src/ImageSharp/Image.FromFile.cs
```diff
@@ -31,13 +31,53 @@ public static IImageFormat DetectFormat(string filePath)
...
+        /// <param name="config">The configuration.</param>
+        /// <param name="filePath">The image file to open and to read the header from.</param>
+        /// <param name="format">The format type of the decoded image.</param>
+        /// <exception cref="NotSupportedException">Thrown if the stream is not readable.</exception>
+        /// <returns>
+        /// The <see cref="IImageInfo"/> or null if suitable info detector is not found.
+        /// </returns>
+        public static IImageInfo Identify(Configuration config, string filePath, out IImageFormat format)
+        {
+            config ??= Configuration.Default;
```
antonfirsov: Wouldn't it be better to guard against `null` on places like this? A call in a form of `Image.Identify(null, "blah", out blah)` is not something I would encourage.

JimBobSquarePants: I just followed the existing pattern.

antonfirsov: Shouldn't we consider changing it? Maybe as part of #1110?

JimBobSquarePants: See above edit.

PR #1147, opened 2020-03-15
WIP: WebP decoding

brianpopow:
### Prerequisites

### Description

This is a port of libweb for decoding webp images with ImageSharp. This is still work in progress. 
It is based on [libwebp](https://chromium.googlesource.com/webm/libwebp)

The specification for the lossless webp format can be found here [lossless](https://developers.google.com/speed/webp/docs/webp_lossless_bitstream_specification)

The specification for the lossy format can be found here [lossy](https://tools.ietf.org/html/rfc6386)

cc @jongleur1983


src/ImageSharp/Formats/WebP/LosslessUtils.cs
```diff
@@ -0,0 +1,656 @@
...
+                        }
+
+                        decodedPixelData[decodedPixels++] = colorMap[(int)(packedPixels & bitMask)];
+                        packedPixels >>= bitsPerPixel;
+                    }
+                }
+
+                decodedPixelData.AsSpan().CopyTo(pixelData);
+
+                return;
```
jongleur1983: I really don't like early returns. Makes code harder to read, but if I recall right, I'm alone or part of a minority with this opinion for ImageSharp. :(
(Arguments against this:
- not faster than else, as the compiler generates this return itself.
- this is semantically an if-then-else construct, but it looks like an if-then - and always do something else.
  The return changes the semantics of the other language construct, and a return in a block is easily hidden when the block is collapsed in an IDE or similar.

src/ImageSharp/Formats/WebP/LosslessUtils.cs
```diff
@@ -0,0 +1,656 @@
...
+        /// <param name="start">The start index of reverse transform.</param>
+        /// <param name="numPixels">The number of pixels to apply the transform.</param>
+        public static void TransformColorInverse(Vp8LMultipliers m, Span<uint> pixelData, int start, int numPixels)
+        {
+            int end = start + numPixels;
+            for (int i = start; i < end; i++)
+            {
+                uint argb = pixelData[i];
+                sbyte green = (sbyte)(argb >> 8);
+                uint red = argb >> 16;
```
jongleur1983: this red isn't red, but "alphaAndRed".
As it's not needed afterwards, I'd collapse lines 153 and 154 to one and call it just "red", initializing "newRed to red + ColorTransformDelta in the (old) line 156.

src/ImageSharp/Formats/WebP/LosslessUtils.cs
```diff
@@ -0,0 +1,656 @@
...
+        /// <param name="pixelData">The pixel data to apply the inverse transform on.</param>
+        /// <param name="start">The start index of reverse transform.</param>
+        /// <param name="numPixels">The number of pixels to apply the transform.</param>
+        public static void TransformColorInverse(Vp8LMultipliers m, Span<uint> pixelData, int start, int numPixels)
+        {
+            int end = start + numPixels;
+            for (int i = start; i < end; i++)
+            {
+                uint argb = pixelData[i];
+                sbyte green = (sbyte)(argb >> 8);
```
jongleur1983: this green contains "arg" (alpha, red and green), right?
if that's the case, why is it called green here?

src/ImageSharp/Formats/WebP/LosslessUtils.cs
```diff
@@ -0,0 +1,656 @@
...
+            int width = transform.XSize;
+            Span<uint> transformData = transform.Data.GetSpan();
+
+            // First Row follows the L (mode=1) mode.
+            PredictorAdd0(pixelData, processedPixels, 1, output);
+            PredictorAdd1(pixelData, 1, width - 1, output);
+            processedPixels += width;
+            yStart++;
+
+            int y = yStart;
```
jongleur1983: yStart is only used here and has a constant value of 1, so we can either use int y = 1 here (and remove lines 178 and 186) or use const int yStart = 1 and remove only line 186.

src/ImageSharp/Formats/WebP/LosslessUtils.cs
```diff
@@ -0,0 +1,656 @@
...
+                    {
+                        xEnd = width;
+                    }
+
+                    // There are 14 different prediction modes.
+                    // In each prediction mode, the current pixel value is predicted from one or more neighboring pixels whose values are already known.
+                    int startIdx = processedPixels + x;
+                    int numberOfPixels = xEnd - x;
+                    switch (predictorMode)
+                    {
```
jongleur1983: This is NOT TESTED and I'm quite unsure if it works, but:
all Predictor calls here have 4 identical parameters, so MAYBE the switch could be optimized as follows:

Those predictors that use the width as an additional parameter, could change their parameter order to use width as last parameter.
My very optimistic guess would be the following general code to be generated:

- push return address on stack
- push pixelData, startIndex, numberOfPixels and output on Stack,
- switch on predictorMode and depending on the value 
  - maybe push "width" to stack
  - jump to Predictor method entrypoint.

Maybe even adding width to ALL predictors and ignoring where not used could be beneficial.
This is an unchecked guess, so don't rely on it to work.

src/ImageSharp/Formats/WebP/LosslessUtils.cs
```diff
@@ -0,0 +1,656 @@
...
+            int endIdx = startIdx + numberOfPixels;
+            int offset = 0;
+            for (int x = startIdx; x < endIdx; ++x)
+            {
+                uint pred = Predictor13(output[x - 1], output, startIdx - width + offset++);
+                output[x] = AddPixels(input[x], pred);
+            }
+        }
+
+        private static uint Predictor0()
```
jongleur1983: why not using a const here as well (private const uint Predictor0 = WebPConstants.ArgbBlack?

src/ImageSharp/Formats/WebP/LosslessUtils.cs
```diff
@@ -0,0 +1,656 @@
...
+
+            output.CopyTo(pixelData);
+        }
+
+        private static void PredictorAdd0(Span<uint> input, int startIdx, int numberOfPixels, Span<uint> output)
+        {
+            int endIdx = startIdx + numberOfPixels;
+            for (int x = startIdx; x < endIdx; ++x)
+            {
+                uint pred = Predictor0();
```
jongleur1983: as Predictor0() returns a constant value, it shouldn't be called once per loop iteration

src/ImageSharp/Formats/WebP/LossyUtils.cs
```diff
@@ -0,0 +1,912 @@
...
+{
+    internal static class LossyUtils
+    {
+        private static void Put16(int v, Span<byte> dst)
+        {
+            for (int j = 0; j < 16; ++j)
+            {
+                Memset(dst.Slice(j * WebPConstants.Bps), (byte)v, 0, 16);
+            }
+        }
```
jongleur1983: maybe faster:

```C#
for (int j = 0; j < 16 * WebPConstants.Bps; j+=WebPConstants.Bps) 
{
    Memset(dst.Slice(j), (byte)v, 0, 16);
}
```
this avoids the repeated multiplication.
(I assume WebPConstants.Bps is const here. If not, WebPConstants.Bps and 16*WebPConstants.Bps should be extracted to variables first).

src/ImageSharp/Formats/WebP/LossyUtils.cs
```diff
@@ -0,0 +1,912 @@
...
+
+        public static void DC16(Span<byte> dst, Span<byte> yuv, int offset)
+        {
```
jongleur1983: Define offsetMinus1 = offset - 1 before the loop, use that in the first yuv[] index.
Define int offsetMinusBps = offset - WebPConstants.Bps before the loop, use that as part of the second yuv[] index.

By that we'd get:
```C#
int offsetMinus1 = offset - 1;
int offsetMinusBps = offset - WebPConstants.Bps;

int dc = 16;
for (int j=0; j<16; ++j) 
{
  dc += yuv[offsetMinus1 + (j * WebPConstants.Bps) + yuv[offsetMinusBps + j];
}

Put16(dc >> 5, dst);
```

src/ImageSharp/Formats/WebP/LossyUtils.cs
```diff
@@ -0,0 +1,912 @@
...
+            }
+        }
+
+        public static void HE16(Span<byte> dst, Span<byte> yuv, int offset)
+        {
+            // horizontal
+            for (int j = 16; j > 0; --j)
+            {
+                // memset(dst, dst[-1], 16);
+                byte v = yuv[offset - 1];
```
jongleur1983: apply offset-- once before the loop and omit the "-1" here.
Instead of subtracting 1 16x this reduces that to once (while it's transparent to the addition in line 55)

src/ImageSharp/Formats/WebP/LossyUtils.cs
```diff
@@ -0,0 +1,912 @@
...
+            Put16(0x80, dst);
+        }
+
+        public static void DC8uv(Span<byte> dst, Span<byte> yuv, int offset)
+        {
+            int dc0 = 8;
+            for (int i = 0; i < 8; ++i)
+            {
+                // dc0 += dst[i - BPS] + dst[-1 + i * BPS];
+                dc0 += yuv[offset + i - WebPConstants.Bps] + yuv[offset - 1 + (i * WebPConstants.Bps)];
```
jongleur1983: see above: pull parts of the calculation outside the loop

brianpopow:
@jongleur1983 i have added better benchmark images. So previous ones decoded in just a few ms. The new ones are bigger and have a alpha channel. I want to use that as a baseline to see what the changes will do.

This is the current benchmark result:

|                     Method |       Runtime |       TestImageLossy |    TestImageLossless |      Mean |     Error |   StdDev |      Gen 0 |     Gen 1 |     Gen 2 |    Allocated |
|--------------------------- |-------------- |--------------------- |--------------------- |----------:|----------:|---------:|-----------:|----------:|----------:|-------------:|
|        'Magick Lossy WebP' |    .NET 4.7.2 | WebP/(...).webp [21] | WebP/(...).webp [24] |  70.37 ms |  9.234 ms | 0.506 ms |          - |         - |         - |     32.05 KB |
|    'ImageSharp Lossy Webp' |    .NET 4.7.2 | WebP/(...).webp [21] | WebP/(...).webp [24] | 211.77 ms |  8.055 ms | 0.442 ms | 19000.0000 |         - |         - |  82297.31 KB |
|     'Magick Lossless WebP' |    .NET 4.7.2 | WebP/(...).webp [21] | WebP/(...).webp [24] |  49.35 ms |  1.099 ms | 0.060 ms |          - |         - |         - |     15.32 KB |
| 'ImageSharp Lossless Webp' |    .NET 4.7.2 | WebP/(...).webp [21] | WebP/(...).webp [24] | 494.34 ms |  5.505 ms | 0.302 ms |  2000.0000 | 1000.0000 | 1000.0000 | 151801.78 KB |
|        'Magick Lossy WebP' | .NET Core 2.1 | WebP/(...).webp [21] | WebP/(...).webp [24] |  70.21 ms |  1.440 ms | 0.079 ms |          - |         - |         - |      14.8 KB |
|    'ImageSharp Lossy Webp' | .NET Core 2.1 | WebP/(...).webp [21] | WebP/(...).webp [24] | 142.32 ms |  6.046 ms | 0.331 ms |  9000.0000 |         - |         - |  40610.23 KB |
|     'Magick Lossless WebP' | .NET Core 2.1 | WebP/(...).webp [21] | WebP/(...).webp [24] |  49.44 ms |  0.258 ms | 0.014 ms |          - |         - |         - |      14.3 KB |
| 'ImageSharp Lossless Webp' | .NET Core 2.1 | WebP/(...).webp [21] | WebP/(...).webp [24] | 206.45 ms | 11.093 ms | 0.608 ms |  2666.6667 | 1666.6667 | 1000.0000 | 151758.87 KB |
|        'Magick Lossy WebP' | .NET Core 3.1 | WebP/(...).webp [21] | WebP/(...).webp [24] |  69.69 ms |  1.147 ms | 0.063 ms |          - |         - |         - |     14.42 KB |
|    'ImageSharp Lossy Webp' | .NET Core 3.1 | WebP/(...).webp [21] | WebP/(...).webp [24] | 121.72 ms |  2.373 ms | 0.130 ms |  9000.0000 |         - |         - |  40050.06 KB |
|     'Magick Lossless WebP' | .NET Core 3.1 | WebP/(...).webp [21] | WebP/(...).webp [24] |  49.37 ms |  1.865 ms | 0.102 ms |          - |         - |         - |     14.27 KB |
| 'ImageSharp Lossless Webp' | .NET Core 3.1 | WebP/(...).webp [21] | WebP/(...).webp [24] | 194.03 ms | 37.759 ms | 2.070 ms |  2000.0000 | 1000.0000 | 1000.0000 | 151756.38 KB |

JimBobSquarePants:
Ooft them allocations! 😉 

I'll try and get some review time in and also get it profiled to reduce that.

brianpopow:
> Ooft them allocations! 😉
> 
> I'll try and get some review time in and also get it profiled to reduce that.

@JimBobSquarePants there is definitely room for improvement. One example i still have on my todo list is the alpha chunk data: its not yet using the memory allocator.

src/ImageSharp/Formats/WebP/AlphaDecoder.cs
```diff
@@ -0,0 +1,309 @@
...
+        /// </summary>
+        public IMemoryOwner<byte> Alpha { get; }
+
+        public int CropTop { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether pre-processing was used during compression.
+        /// 0: no pre-processing, 1: level reduction.
+        /// </summary>
+        private int PreProcessing { get; }
```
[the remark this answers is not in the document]
brianpopow: i dont see how the PreProcessing would affect the decoding process. I will remove it.

src/ImageSharp/Formats/WebP/AlphaDecoder.cs
```diff
@@ -0,0 +1,309 @@
...
+                if (this.AlphaFilterType == WebPAlphaFilterType.None)
+                {
+                    this.Data.AsSpan(0, this.Width * this.Height).CopyTo(alphaSpan);
+                    return;
+                }
+
+                Span<byte> deltas = this.Data.AsSpan();
+                Span<byte> dst = alphaSpan;
+                Span<byte> prev = null;
+                for (int y = 0; y < this.Height; ++y)
```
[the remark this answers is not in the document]
brianpopow: that would lead to duplicate code as the for loop would be in each switch statement. What i could do is use the same approach as in the reference code and use a delegate array for the unfilter methods and the filter type would be the index into that array.

The original code looks like this:

```
for (y = 0; y < num_rows; ++y) {
        WebPUnfilters[alph_dec->filter_](prev_line, deltas, dst, width);
        prev_line = dst;
        dst += width;
        deltas += width;
      }
```

Im not sure though, if i should do that as i think it hurts readability a bit.

src/ImageSharp/Formats/WebP/AlphaDecoder.cs
```diff
@@ -0,0 +1,309 @@
...
+
+        private static bool Is8bOptimizable(Vp8LMetadata hdr)
+        {
+            if (hdr.ColorCacheSize > 0)
+            {
+                return false;
+            }
+
+            // When the Huffman tree contains only one symbol, we can skip the
+            // call to ReadSymbol() for red/blue/alpha channels.
```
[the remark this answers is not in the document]
brianpopow: @jongleur1983 im keen in removing Is8bOptimizable altogether, because its not used yet and i dont see when we would need more than 1 byte per pixel for the alpha channel. I have no test image that would need that. Do you have any insight were this might be the case? 

I dont see anything mentioned in the [spec](https://developers.google.com/speed/webp/docs/riff_container#alpha), but im afraid that it must have some use case im not aware of.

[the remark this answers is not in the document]
brianpopow: i have found one image where Is8bOptimizable is false: `alpha_color_cache.webp`
I will try to figure out what the difference is to the other images.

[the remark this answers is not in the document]
brianpopow: i have figured out how to decode `alpha_color_cache.webp`. The alpha channel data will be in the green channel of the decoded lossless data. See commit: 45bf1064e34a575d77ddbfafc6d7ffb935d172d7

src/ImageSharp/Formats/WebP/WebPLossyDecoder.cs
```diff
@@ -0,0 +1,1414 @@
...
+
+            dec.Vp8BitReaders[lastPart] = new Vp8BitReader(this.bitReader.Data, (uint)sizeLeft, partStart);
+        }
+
+        private void ParseDequantizationIndices(Vp8Decoder decoder)
+        {
+            Vp8SegmentHeader vp8SegmentHeader = decoder.SegmentHeader;
+
+            int baseQ0 = (int)this.bitReader.ReadValue(7);
+            bool hasValue = this.bitReader.ReadBool();
```
jongleur1983: reusing hasValue has no benefit here IMHO,
I'd remove the variable and inline the ReadBool() calls into the ternary operator lines, like:
`int dqy1DC = this.bitReader.ReadBool() ? this.bitReader.ReadSignedValue(4) : 0;`

to make this even more Readable, a local function ReadOptionalValue() is another imrovement (replacing the first suggestion)

src/ImageSharp/Formats/WebP/WebPLossyDecoder.cs
```diff
@@ -0,0 +1,1414 @@
...
+            for (int t = 0; t < WebPConstants.NumTypes; ++t)
+            {
+                for (int b = 0; b < WebPConstants.NumBands; ++b)
+                {
+                    for (int c = 0; c < WebPConstants.NumCtx; ++c)
+                    {
+                        for (int p = 0; p < WebPConstants.NumProbas; ++p)
+                        {
+                            byte prob = WebPLookupTables.CoeffsUpdateProba[t, b, c, p];
+                            int v = this.bitReader.GetBit(prob) != 0
```
jongleur1983: let v be a byte variable, cast the ReadValue() case to byte and remove the - the redundant - cast in line 1298.

src/ImageSharp/Formats/WebP/WebPLossyDecoder.cs
```diff
@@ -0,0 +1,1414 @@
...
+        {
+            var io = default(Vp8Io);
+            io.Width = (int)pictureHeader.Width;
+            io.Height = (int)pictureHeader.Height;
+            io.UseScaling = false;
+            io.ScaledWidth = io.Width;
+            io.ScaledHeight = io.ScaledHeight;
+            io.MbW = io.Width;
+            io.MbH = io.Height;
+            io.YStride = (int)(16 * ((pictureHeader.Width + 15) >> 4));
```
jongleur1983: precalculate (width + 15) >>4

src/ImageSharp/Formats/WebP/WebPLossyDecoder.cs
```diff
@@ -0,0 +1,1414 @@
...
+            {
+                // For complex filter, we need to preserve the dependency chain.
+                dec.TopLeftMbX = 0;
+                dec.TopLeftMbY = 0;
+            }
+            else
+            {
+                // For simple filter, we include 'extraPixels' on the other side of the boundary,
+                // since vertical or horizontal filtering of the previous macroblock can modify some abutting pixels.
+                dec.TopLeftMbX = (-extraPixels) >> 4;
```
jongleur1983: calculate once, use twice

src/ImageSharp/Formats/WebP/WebPLossyDecoder.cs
```diff
@@ -0,0 +1,1414 @@
...
+        {
+            nzCoeffs <<= 2;
+            nzCoeffs |= (uint)((nz > 3) ? 3 : (nz > 1) ? 2 : dcNz);
+            return nzCoeffs;
+        }
+
+        [MethodImpl(InliningOptions.ShortMethod)]
+        private static Vp8BandProbas[] GetBandsRow(Vp8BandProbas[,] bands, int rowIdx)
+        {
+            Vp8BandProbas[] bandsRow = Enumerable.Range(0, bands.GetLength(1)).Select(x => bands[rowIdx, x]).ToArray();
```
jongleur1983: is this linq code optimized somehow by compiler magic?
if not I'd rewrite it to use a loop instead.

src/ImageSharp/Formats/WebP/WebPLossyDecoder.cs
```diff
@@ -0,0 +1,1414 @@
...
+            return bandsRow;
+        }
+
+        [MethodImpl(InliningOptions.ShortMethod)]
+        private static int CheckMode(int mbx, int mby, int mode)
```
jongleur1983: this method is used twice and the only use of the result is to switch over the values.
This get's more readable and should be equally fast (I hope) when we introduce a private enum for CheckMode.
It would allow to use self-describing values instead.

src/ImageSharp/Formats/WebP/LosslessUtils.cs
```diff
@@ -0,0 +1,656 @@
...
+                output[x] = AddPixels(input[x], pred);
+            }
+        }
+
+        private static uint Predictor0()
+        {
+            return WebPConstants.ArgbBlack;
+        }
+
+        private static uint Predictor1(uint left, Span<uint> top)
```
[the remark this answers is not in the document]
jongleur1983: => method is not used currently!

src/ImageSharp/Formats/WebP/WebPLossyDecoder.cs
brianpopow: The linq usage is now removed

src/ImageSharp/Formats/WebP/WebPLookupTables.cs
```diff
@@ -0,0 +1,579 @@
...
+                { 244, 1, 255, 128, 128, 128, 128, 128, 128, 128, 128 },
+                { 238, 1, 255, 128, 128, 128, 128, 128, 128, 128, 128 }
+            }
+          }
+        };
+
+        static WebPLookupTables()
+        {
+            Abs0 = new Dictionary<int, byte>();
+            for (int i = -255; i <= 255; ++i)
```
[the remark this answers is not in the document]
brianpopow: this is not a critical path, also i dont think the change will improve the performance much. I will leave it as is it is.

src/ImageSharp/Formats/WebP/WebPLookupTables.cs
```diff
@@ -0,0 +1,579 @@
...
+            {
+                Abs0[i] = (byte)((i < 0) ? -i : i);
+            }
+
+            Clip1 = new Dictionary<int, byte>();
+            for (int i = -255; i <= 255 + 255; ++i)
```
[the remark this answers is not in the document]
brianpopow: same as above

src/ImageSharp/Formats/WebP/WebPLossyDecoder.cs
```diff
@@ -0,0 +1,1414 @@
...
+            {
+                TPixel color = default;
+                Span<byte> alphaSpan = alpha.Memory.Span;
+                for (int y = 0; y < height; y++)
+                {
+                    Span<TPixel> pixelRow = pixels.GetRowSpan(y);
+                    for (int x = 0; x < width; x++)
+                    {
+                        int offset = (y * width) + x;
+                        int idxBgr = offset * 3;
```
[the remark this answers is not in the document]
jongleur1983: sorry, you're right - ignore that comment - no idea what I had in mind or if I just looked wrong here...

brianpopow:
I will close this for now and start working on the encoder.

Thanks @jongleur1983 for your review feedback!

PR #1152, opened 2020-03-24
Replace Vector4.Clamp

JimBobSquarePants:
### Prerequisites

### Description
`Vector4.Clamp(Vector4, Vector4, Vector4)` is not hardware accelerated and can be sped up by around 5X by using a combination of `Vector4.Min(Vector4, Vector4)` and `Vector4.Max(Vector4, Vector4)`. 

Notable areas we can improve as a result are `Block8x8F.NormalizeColorsInplace(float)` which helps with jpeg decoding..

### DIFF
--- a/src/ImageSharp/Common/Helpers/Vector4Utilities.cs
+++ b/src/ImageSharp/Common/Helpers/Vector4Utilities.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Six Labors and contributors.
+// Copyright (c) Six Labors and contributors.
 // Licensed under the Apache License, Version 2.0.
 
 using System;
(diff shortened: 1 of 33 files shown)


src/ImageSharp/Common/Helpers/Vector4Utilities.cs
```diff
@@ -11,8 +11,20 @@ namespace SixLabors.ImageSharp
...
+        /// <summary>
+        /// Restricts a vector between a minimum and a maximum value.
+        /// 5x Faster then <see cref="Vector4.Clamp(Vector4, Vector4, Vector4)"/>.
+        /// </summary>
+        /// <param name="x">The vector to restrict.</param>
+        /// <param name="min">The minimum value.</param>
+        /// <param name="max">The maximum value.</param>
+        /// <returns>The <see cref="Vector4"/>.</returns>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public static Vector4 FastClamp(Vector4 x, Vector4 min, Vector4 max)
```
tannergooding: Just a note that I've also called out on https://github.com/dotnet/runtime/issues/34017.

Due to how HLSL/GLSL behave and how the underlying `minps`/`maxps` instructions behave, you may need to be careful on the ordering of `x, y` here.

For regular inputs, the order won't matter and you'll get the correct/expected result. However, if either input is `NaN` or both inputs are `Zero`, the result returned might differ based on how you order `x, y`.

tannergooding: I finished the analysis and it looks like you want `Vector4.Min(Vector4.Max(x, min), max)` (so just swap `min` and `x`): https://github.com/dotnet/runtime/issues/34017#issuecomment-603524623

antonfirsov: @tannergooding thanks a lot for the engagement!

JimBobSquarePants: Ah @tannergooding this is fantastic! It explains the issue I was seeing when attempting to replace a part within the fallback intrinsics. The change fixes that.

PR #1153, opened 2020-03-25
More Efficient ToBase64String

JimBobSquarePants:
### Prerequisites

### Description
Prevents allocation of an array when saving to a Base64 encoded string.


src/ImageSharp/ImageExtensions.cs
```diff
@@ -115,8 +118,25 @@ public static string ToBase64String<TPixel>(this Image<TPixel> source, IImageFor
...
+                // Always available.
+                stream.TryGetBuffer(out ArraySegment<byte> buffer);
+
+#if !SUPPORTS_BASE64SPAN
+
+                byte[] sharedBuffer = ArrayPool<byte>.Shared.Rent(buffer.Count);
+                try
+                {
+                    buffer.AsSpan().CopyTo(sharedBuffer);
+                    return $"data:{format.DefaultMimeType};base64,{Convert.ToBase64String(sharedBuffer)}";
```
antonfirsov: Is there a reason to operate on a copy instead of `ArraySegment<byte>.Array`?

JimBobSquarePants: `ArraySegment<byte>.Array` won't be the correct length. I've just realized though that I'm making the same mistake in the unsupported version so will have to accept the allocation in that scenario.

JimBobSquarePants: Haha yes, much better. Will update.

PR #1160, opened 2020-03-29
TGA image origin and support decoding 16bit monochome images

brianpopow:
### Prerequisites

### Description

This PR fixes two issues.
- Images with a bottom right or top right origin now decode correctly.
- Adds support for decoding 16bit gray images.

There are 7 failing tests, but it looks to me that the reference output from Magick.Net is not correct.
Here is an example for image `grayscale_a_UL.tga`, which can be found in `tests\Images\Input\Tga`

This is the output from imagesharp:
grayscale_a_UL_imagesharp

This is the output from imagemagick:
grayscale_a_UL_magick

All the failing tests are with 16 bit monochrome images.


JimBobSquarePants:
@brianpopow Agree regarding output difference. Our output is what I would expect from 16bit grayscale.

src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
```diff
@@ -384,29 +413,65 @@ private void ReadMonoChrome<TPixel>(int width, int height, Buffer2D<TPixel> pixe
...
-                    Span<byte> rowSpan = row.GetSpan();
+                    int newY = InvertY(y, height, origin);
+                    Span<TPixel> pixelSpan = pixels.GetRowSpan(newY);
 
-                    if (!this.hasAlpha)
+                    if (isXInverted)
                     {
-                        // We need to set the alpha component value to fully opaque.
-                        for (int x = 1; x < rowSpan.Length; x += 2)
+                        for (int x = 0; x < width; x++)
```
JimBobSquarePants: We should move the conditionals outside the loop.

src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
```diff
@@ -504,7 +591,7 @@ private void ReadRle<TPixel>(int width, int height, Buffer2D<TPixel> pixels, int
                 this.UncompressRle(width, height, bufferSpan, bytesPerPixel);
                 for (int y = 0; y < height; y++)
                 {
-                    int newY = Invert(y, height, inverted);
+                    int newY = InvertY(y, height, origin);
                     Span<TPixel> pixelRow = pixels.GetRowSpan(newY);
                     int rowStartIdx = y * width * bytesPerPixel;
                     for (int x = 0; x < width; x++)
```
JimBobSquarePants: These per-pixel conditionals will be hurting performance also.

brianpopow:
> @brianpopow Agree regarding output difference. Our output is what I would expect from 16bit grayscale.

@JimBobSquarePants ok then i will add reference images to the ReferenceOutput folder and compare in the tests against those instead of the magick decoder output.

src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
brianpopow: i think i can move it outside the loop. Let me first get the tests green and then i will give it a try.

brianpopow: i have moved the calculation of the x and y positions out of the loop where possible, see: 70a62ead9e6bf66e2b4b85d2212be832380d8c7b.

The only place where i could not manage to do so (without making it overly complicated), was the RLE parts.

brianpopow:
i have added more test cases to cover all the different variants.

src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
JimBobSquarePants: Yeah that's great 👍

PR #1157, opened 2020-03-27
TGA dont ignore alpha channel bits

brianpopow:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

During encoding and decoding the alpha channel bits of the image descriptor was ignored. See page 8 of the specification `Field 5.6 (1 byte) - Image Descriptor`. The spec is inside the tga folder. This PR fixes that. 

edit:
It seems the imagemagick reference decoder is also ignoring the alpha channel bits for 32 bits per pixel images. The images are `32bit_rle_no_alphabits.tga` and `32bit_no_alphabits.tga`
I have added reference output images and compare against those in the tests.

PR #1165, opened 2020-04-05
Additional tga tests for paletted RLE images

brianpopow:
### Prerequisites

### Description

While i was adding more test cases for tga images, i noticed a difference to imagemagick reference decoder for paletted TGA images. I seems i have introduced a mistake with #1157 for paletted images: If there is a palette with 4 bytes for each entry, the alpha channel should not be ignored. It would not make sense to write this extra 4th byte and then ignore it.

I wanted to treat 16 bit paletted images the same, but there are some tga images which would then decode incorrectly, like `ccm8.tga` (alpha bits from the image descriptor is 0 here and all alpha bits from the palette are too). I could not  find a clear statement in the spec about it.  So i decided to do it the same as others do it (imagemagick, gimp and irfanview).


JimBobSquarePants:
Yeah, consistency is key here. Thanks for looking into this.

PR #1169, opened 2020-04-06
Value delegate row processors

Sergio0694:
### Prerequisites

### Description
This PR expands on the APIs to process pixel-type-agnostic rows in a given image, by adding the ability to pass a "value delegate" instead of an actual C# delegate. This offers additional performance increase (skipped `callvirt`, and ability to inline the delegate entirely), at the cost of some extra verbosity for the users. The new APIs are overloads, so the previous delegate-based APIs are still available as well.

The idea here is that the `PixelRowDelegateProcessor` type is defining its own struct internally anyway to wrap the input user-defined delegate, so we might as well just offer more experienced users the ability to provide such a struct directly, to get a "free" performance bump.

Here's how a user would go about using the two different versions to process an image:

```csharp
var image = Image.Load("foo.jpg");

// C# delegate
image.Mutate(c => c.ProcessPixelRowsAsVector4(span =>
{
    for (int i = 0; i < span.Length; i++)
    {
        Vector4 v4 = span[i];
        float avg = (v4.X + v4.Y + v4.Z) / 3f;
        span[i] = new Vector4(avg);
    }
}));

// User defined value delegate
private readonly struct PixelAverageProcessor2 : IPixelRowDelegate
{
    public void Invoke(Span<Vector4> span)
    {
        for (int i = 0; i < span.Length; i++)
        {
            Vector4 v4 = span[i];
            float avg = (v4.X + v4.Y + v4.Z) / 3f;
            span[i] = new Vector4(avg);
        }
    }
}

// Value delegate
image.Mutate(c => c.ProcessPixelRowsAsVector4<PixelAverageProcessor>());
```


src/ImageSharp/Processing/Extensions/Effects/PixelRowDelegateExtensions.cs
```diff
@@ -55,14 +155,114 @@ public static IImageProcessingContext ProcessPixelRowsAsVector4(this IImageProce
         public static IImageProcessingContext ProcessPixelRowsAsVector4(this IImageProcessingContext source, PixelRowOperation rowOperation, Rectangle rectangle, PixelConversionModifiers modifiers)
             => source.ApplyProcessor(new PixelRowDelegateProcessor(rowOperation, modifiers), rectangle);
 
+        /// <summary>
+        /// Applies a user defined processing delegate to the image.
+        /// </summary>
+        /// <typeparam name="TDelegate">The type of user defined, row processing delegate to use.</typeparam>
+        /// <param name="source">The image this method extends.</param>
+        /// <returns>The <see cref="IImageProcessingContext"/> to allow chaining of operations.</returns>
+        public static IImageProcessingContext ProcessPositionAwarePixelRowsAsVector4<TDelegate>(this IImageProcessingContext source)
```
JimBobSquarePants, line 164: I don't like the additional naming here. It was fine as it was with `ProcessPixelRowsAsVector4`. The generic variant was visible enough

Sergio0694:
@JimBobSquarePants It's not so much about saving the memory in the first call, but the fact that with a delegate you get a `callvirt` on that delegate for each row invocation, whereas with the value delegate you can skip those entirely and just basically inline the entire image processing part.

As for the additional code, I think I might be able to reduce it by moving some code around, I'll give it a shot 😄

src/ImageSharp/Processing/Extensions/Effects/PixelRowDelegateExtensions.cs
Sergio0694: The reason why I did that is that you can't have overloads that are only different for the type constraints. Here we'd have two methods with the same name and parameter, with the only difference being that one takes a `TDelegate where TDelegate : struct, IPixelRowDelegate`, and the other takes a `TDelegate where TDelegate : struct, IPixelRowDelegate<Point>`.

Not sure how to avoid having the separate name, unless we dropped the value delegate API entirely for the position aware variant, but we'd then lose half the new APIs basically, and also make the API surface "asymmetrical" in this case, if that matters 🤔

JimBobSquarePants:
>It's not so much about saving the memory in the first call, but the fact that with a delegate you get a callvirt on that delegate for each row invocation

You're talking about this yeah? 

https://github.com/SixLabors/ImageSharp/blob/44b4651a4da5705db6b9b462a5cc93a50d466737/src/ImageSharp/Processing/Processors/Effects/PixelRowDelegateProcessor.cs#L64

Are we sure this comes at a cost? Checking the spec despite the virtual nature of the Invoke methods a delegate is a sealed class. There's a comment here (not definitive) that is intriguing.

>When a delegate's invoke is called, CLR automatically emits a highly optimized body for this method as opposed to compiling IL code to generate body which it does for 'normal' methods. This has nothing to do with being marked virtual in the IL.

https://stackoverflow.com/a/11451035/427899

Calling the delegate via `foo(bar)` should compile down directly to `foo.Invoke(bar)`.

Sergio0694:
I think you might be mixing up the `Invoke` method from our own `IRowProcessor` interface, and the implicit `Invoke` method from the `Delegate` type in C#. Yeah using either `Invoke` or just the `()` syntax in C# on a `Delegate` instance is absolutely equivalent, I agree. Let me try with a code example to show the optimization I'm talking about for this PR, that should hopefully help clear things up 😄

This is the current architecture we have in place (before this PR):

```csharp
// Custom delegate to have a Span<T> parameter
public delegate void RowProcessor(Span<Vector4> span);

// Interface for the wrapper struct
public interface IRowProcessor
{
    void Invoke(Span<Vector4> span);
}

// Internal value delegate wrapping the delegate
public readonly struct RowProcessorWrapper : IRowProcessor
{
    private readonly RowProcessor rowProcessor;
    
    public RowProcessorWrapper(RowProcessor rowProcessor)
    {    
        this.rowProcessor = rowProcessor;
    }

    public void Invoke(Span<Vector4> span) => this.rowProcessor(span);
}
```

When we use `ProcessRowsAsVector4`, the user passes a delegate, which is internally wrapped by an instance of that `RowProcessorWrapper` struct (what we refer to as "value delegate"), which is then passed to the generic row processor. What happens is that for each row, that processor will call the `Invoke` method on _our_ value delegate, which basically acts as a proxy to the wrapped delegate. We can inline the `Invoke` in our wrapping class, since it's just called from within a loop in our generic processor, but the relayed call to the wrapped delegate cannot be inlined there. Se that's why I'm saying we end up with the overhead of a `callvirt` _per row_, since we need to transfer control to the input delegate given by the user. The generic version of the code is the same for each input processor given by the user, so there's no way the JIT might be able to inline that - the actual input function being invoked by that wrapping proxy might change over time and is not fixed.

You can visualize this in a simplified manner like this:

```csharp
public static class Program
{
    public static void ProcessWithDelegate(Vector4[] data, int height, int width)
    {
        ImageSharp.Process(data, height, width, span =>
        {
            foreach (ref Vector4 v in span)
            {
                v /= 2;
            }
        });
    }
}

public static class ImageSharp
{
    // Assume this is our pixel-type generic processor
    public static void Process(Vector4[] data, int height, int width, RowProcessor rowProcessor)
    {
        // Wrap the user-provided delegate
        var wrapper = new RowProcessorWrapper(rowProcessor);
        
        for (int i = 0; i < height; i++)
        {
            Span<Vector4> row = data.AsSpan(i * width, width);
            
            // Proxy call to the input delegate,
            // we can inline the proxy but not the actual target delegate
            wrapper.Invoke(row);
        }
    }
}
```

Compare that with the proposed alternative of this PR: since we're always using that wrapping value delegate internally, we might as well just expose that capability directly to the user, so that they can give us the actual value delegate to inline directy. In this case we'd have one type generic concrete instantiation _per value delegate_, so the JIT will be able to inline that entirely if possible.

We simply reuse the interface defined before, and then the user might define the custom row processing logic manually through a value delegate, like this:

```csharp
public readonly struct RowProcessorValueDelegate : IRowProcessor
{
    public void Invoke(Span<Vector4> span)
    {
        foreach (ref Vector4 v in span)
        {
            v /= 2;
        }
    }
}
```

And then pass that directly to `ImageSharp`, which can then build a processor-specific generic instance, and inline everything as possible. We're basically skipping one entire indirection level, as well as the `callvirt` to that delegate entirely (the one that was being proxied through our wrapper).

You can visualize this new version like so:

```csharp
public static class Program
{    
    public static void ProcessWithValueDelegate(Vector4[] data, int height, int width)
    {
        // Processor specific generic method being JITted
        ImageSharp.Process<RowProcessorValueDelegate>(data, height, width);
    }
}

public static class ImageSharp
{    
     public static void Process<TDelegate>(Vector4[] data, int height, int width)
         where TDelegate : struct, IRowProcessor
    {
        TDelegate processor = default;
        
        for (int i = 0; i < height; i++)
        {
            Span<Vector4> row = data.AsSpan(i * width, width);
            
            // Single call to our generic processor
            // This can be *completely* inlined now!
            processor.Invoke(row);
        }
    }
}
```

I've probably written way too much to detail something that's actually simple, I apologize if my explanation skills are lacking 😅 Anyway, I hope this gets my point across, all in all I think this is a decent (optional) optimization, at no real cost - all the necessary code was already there to begin with. I do like the idea of offering more advanced APIs to users that know how to use them, just like we're now also exposing the various `PixelOperation<TPixel>` APIs and whatnot, which the average casual user wouldn't even look at. Let me know what you think! 😄

JimBobSquarePants:
I'm not confusing anything. 

The line of code I pointed to was the wrapping struct method that wraps our supplied delegate invocation. What I am saying is that the cost of the callvirt is not enough, (without benchmarks) to make up for the additional complexity in the surface API, especially when we have to add an unwieldy method name. 

There comes a point when designing an API where you have balance performance with usability. I'm not sold that the net benefit is worth the complexity. (Without the name change I would not be concerned but that is a degenerate API change).

Sergio0694:
Oh, sorry, I see what you're saying now, I thought you meant that the JITter could already somehow devirtualize that proxy invocation to the wrapped delegate. Read that backwards, and I hope my previous message didn't come across as offensive or anything, that wasn't the tone I was trying to convey at all! I just failed to understand what you meant there, my bad 😅

I agree that having a different name for the public APIs is not ideal, yeah.
Run some benchmarks and it looks like this change brings a ~5% speed improvement 🤔

|        Method |     Mean |   Error |   StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|------------- |---------:|--------:|---------:|------:|--------:|-------:|----------:|
|      Delegate | 464.7 us | 8.11 us | 12.38 us |  1.00 |    0.00 | 1.4648 |   5.97 KB |
| ValueDelegate | 451.3 us | 3.34 us |  2.61 us |  0.95 |    0.03 | 1.4648 |   5.94 KB |

I agree with you that it's not much, I'm just not a fan of hiding this feature away when the necessary code is there already. But yeah those different API names do look ugly. I'll close this for now then, but I hope we'll be able to possibly revisit this in the future if we come up with a better solution for the publicly exposed APIs. It was nice to brainstorm ideas with you though, thank you for your time!

Issue #1027, opened 2019-10-15
PNG Encoder produces too huge images and different sizes in Linux and Windows


odalet:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When saving an image to PNG I end up with a file that is more than 5x larger than what I would obtain by using **System.Drawing** (or **Paint.NET**)
Attached are the images I generated using **ImageSharp**, and the same one opened/saved with **Paint.NET**.
I've tried to tweak the `CompressionLevel` and `FilterMethod` (not so sure what this one is about) 
properties of the encoder, but this changes nearly nothing in the output image size:
* Compression levels of 1 to 5 give me the same size of 310 KB both in my Windows and WSL tests
* Compression levels of 6 to 9 give me a size of 289 KB when run on Windows but 102 KB in WSL
*  With `FilterMethod` set to `Adaptive`, I obtain slightly better results; however, far from my expected 55 KB

I can see there's a related issue (#702); however, regardless of the different output sizes in Linux vs Windows, I still think my issue stands as a bug as I'd very much like to be able to obtain a ~50 KB image instead of 300 KB.

### Steps to Reproduce

Here is the code I used to exhibit this behavior:

```csharp
using System.IO;
using System.Runtime.InteropServices;
using SixLabors.ImageSharp;
using SixLabors.ImageSharp.Formats.Png;
using SixLabors.ImageSharp.PixelFormats;

namespace ImageSharpCompressionTest
{
    internal sealed class Program
    {
        private static void Main(string[] args)
        {
            var encoder = new PngEncoder
            {
                BitDepth = PngBitDepth.Bit8,
                ColorType = PngColorType.RgbWithAlpha
            };

            // encoder.FilterMethod = PngFilterMethod.Adaptive;

            var w = 3500;
            var h = 3500;
            var rgbaBytes = GetImageBytes(w, h);

            var rootdir = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? @"c:\temp\win" : "/mnt/c/temp/lnx/";
[...]

PR #1170, opened 2020-04-06
Faster Deflate

JimBobSquarePants:
### Prerequisites

### Description
Attempting to chip away at the slowness of the managed `ZlibStream`.

I'm in two minds here. Do we continue and run our own version or do we reimplement the wrapper around .NET `DeflateStream` and hope that #1027 is eventually fixed upstream?

We're currently running at 2-2.5x slower than System.Drawing in our benchmarks. I'm managed to save around 10% with these changes.

**I need help optimizing this**. I don't fully understand the deflate process so cannot make any drastic algorithmic improvements. All my changes since copying this from SharpZipLib have been minor and mostly focused around removing bonds checks.

### UPDATE.

Going As-is with this just now. I got the diff down from ~2.5x to the below. Wasn't able to isolate the allocations but can investigate that in the future. 

|               Method |       Runtime | LargeImage |     Mean |    Error |   StdDev | Ratio | RatioSD |    Gen 0 |    Gen 1 |    Gen 2 |  Allocated |
|--------------------- |-------------- |----------- |---------:|---------:|---------:|------:|--------:|---------:|---------:|---------:|-----------:|
| 'System.Drawing Png' |    .NET 4.7.2 |      False | 12.87 ms | 4.162 ms | 0.228 ms |  1.00 |    0.00 | 515.6250 | 234.3750 | 234.3750 | 2079.95 KB |
|     'ImageSharp Png' |    .NET 4.7.2 |      False | 31.00 ms | 4.895 ms | 0.268 ms |  2.41 |    0.06 | 500.0000 | 500.0000 | 500.0000 | 2323.56 KB |
|                      |               |            |          |          |          |       |         |          |          |          |            |
| 'System.Drawing Png' | .NET Core 2.1 |      False | 14.33 ms | 9.995 ms | 0.548 ms |  1.00 |    0.00 | 531.2500 | 250.0000 | 250.0000 | 2076.73 KB |
|     'ImageSharp Png' | .NET Core 2.1 |      False | 30.77 ms | 1.910 ms | 0.105 ms |  2.15 |    0.09 | 562.5000 | 531.2500 | 531.2500 | 2321.25 KB |
|                      |               |            |          |          |          |       |         |          |          |          |            |
| 'System.Drawing Png' | .NET Core 3.1 |      False | 14.08 ms | 7.838 ms | 0.430 ms |  1.00 |    0.00 | 234.3750 | 218.7500 | 218.7500 |  964.29 KB |
|     'ImageSharp Png' | .NET Core 3.1 |      False | 30.79 ms | 0.698 ms | 0.038 ms |  2.19 |    0.07 | 562.5000 | 531.2500 | 531.2500 | 2321.24 KB |


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -380,7 +380,12 @@ private void ReadGammaChunk(PngMetadata pngMetadata, ReadOnlySpan<byte> data)
         private void InitializeImage<TPixel>(ImageMetadata metadata, out Image<TPixel> image)
             where TPixel : unmanaged, IPixel<TPixel>
         {
-            image = new Image<TPixel>(this.configuration, this.header.Width, this.header.Height, metadata);
+            image = Image.CreateUninitialized<TPixel>(
```
JimBobSquarePants, line 383: Easy win on decode here. This was taking a larger percentage of decoding time than I expected.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -633,10 +630,21 @@ private void WriteExifChunk(Stream stream, ImageMetadata meta)
         private void WriteTextChunks(Stream stream, PngMetadata meta)
         {
             const int MaxLatinCode = 255;
-            foreach (PngTextData textData in meta.TextData)
+            for (int i = 0; i < meta.TextData.Count; i++)
```
JimBobSquarePants, line 633: No Linq

src/ImageSharp/Formats/Png/Zlib/Deflater.cs
```diff
@@ -288,8 +288,6 @@ public void Dispose()
                 this.engine = null;
                 this.isDisposed = true;
             }
-
-            GC.SuppressFinalize(this);
```
JimBobSquarePants, line 292: No need with sealed typed and fully managed members.

src/ImageSharp/Formats/Png/Zlib/DeflaterEngine.cs
```diff
@@ -464,6 +462,7 @@ private void SlideWindow()
         /// </summary>
         /// <param name="curMatch">The current match.</param>
         /// <returns>True if a match greater than the minimum length is found</returns>
+        [MethodImpl(InliningOptions.HotPath)]
         private bool FindLongestMatch(int curMatch)
```
JimBobSquarePants, line 466: This method currently take the longest when encoding. I need help if we are to make it any faster.

image

JimBobSquarePants:
>the one in ImageSharp is faster than the BCL one, consistent and not bugged, but just still a bit slower than the one from System.Drawing then

@Sergio0694  No our one is slower than the BCL but not buggy.

Sergio0694:
Oh, gotcha. Well I have to say I find it really weird to hear about such an important class in the BCL being bugged like that, and left in that state for years. Are there open issues on `dotnet/runtime` about it? Also I wonder then given the situation if it's really worth it to spend all this time trying to optimize this self-contained version, as opposed to trying to bring attention to that bug in the BCL and get someone to fix it, or maybe contribute directly there to get it fixed if possible?

I guess it'd just be bad if they eventually fixed it, having to scrap all the hard work for the implementation in this library that would've just acted as a temporary band-aid then 😄

src/ImageSharp/Formats/Png/Zlib/DeflaterEngine.cs
JimBobSquarePants: I'm up for any and all optimization approaches you can think of, feel free to experiment with the branch.

JimBobSquarePants:
@Sergio0694 Take your pick.

https://github.com/dotnet/runtime/issues/30539
https://github.com/dotnet/runtime/issues/1549

Someone proposed an API for ZlibStream but again it's a wrapper round the varying implementation and doesn't expose the Compression level.
https://github.com/dotnet/runtime/issues/2236

Sergio0694:
Wow, I went through the first two issues you linked and man... It might be by design, but it sure looks incredibly inconvenient for devs working on multiple platforms and trying to have fine control 😕

I get that technically the guarantee is that once decompressed you'll always get with the same original data, but I can see how having completely separate implementations under the hood with no real control is probably not what many devs would want to see in their apps.

It just too bad though to have so much time spent into basically rebuilding the same exact API from scratch just to work around this shortcoming in the BCL. If that API proposal for the `DeflateStream` got approved, with also your suggestion for the explicit compression level, it'd be awesome!

saucecontrol:
I don't think the BCL implementation is buggy so much as it's optimized in ways that may not be ideal for PNG use.  The 'bug' in question is an optimization Intel added to zlib that make it quite a bit faster, with negligible impact to compression for the majority of cases.  There was, however, a regression in compression of sparse data, which is more common in images than in text.  The BCL team has simply been hesitant to give up the speed improvement for the majority to restore the compression for the minority.

That said, even the Intel-modified zlib doesn't represent the best performance available.  Cloudflare and Google have also created zlib variants with various optimizations to performance and compression.  There's an alternate project called [zlib-ng](https://github.com/zlib-ng/zlib-ng) that has gathered all of these improvements together and modernized the zlib codebase, which might serve as a good model for a high-performance managed zlib replacement.  It might also be possible to get the BCL team to adopt zlib-ng over the Intel zlib once it reaches release state.

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
```diff
@@ -122,12 +74,77 @@ public DeflaterHuffman(MemoryAllocator memoryAllocator)
             this.pinnedLiteralBuffer = (short*)this.literalBufferHandle.Pointer;
         }
 
+        // See RFC 1951 3.2.6
+        // Literal codes
+        private static short[] StaticLCodes => new short[]
```
Sergio0694, line 79: I'm confused, why is this a property returning an array, instead of a statically initialized array?
This will create a new array every time it's used, and copy the data there.
Am I missing something obvious or is this just an oversight? 🤔

JimBobSquarePants: That's a mistake caused by copy/paste.

Btw... I had a bit of a read and FindLongestMatch should definitely be the focus. Looking at a well known implementation we're doing far too much. I had a quick hack and got around 80% of the tests passing but soon hit a wall.

https://github.com/zlib-ng/zlib-ng/blob/ee1c9bc0be715d788513ac3cbfe6559f91c6b1d3/match_p.h#L49

JimBobSquarePants: > That's a mistake caused by copy/paste.

Though it's a static property so it shouldn't be requesting a new array each time.

Sergio0694: Well no, because it's just a getter there. You don't have a static _field_ that is initialized at some point and that you then use to get a reference to that array, you just have a getter that returns a new array every time. That's exactly the same as this:
```csharp
public static short[] StaticLCodes_Get() => new short[] { ... };
```

JimBobSquarePants:
@saucecontrol It might also be possible to get the BCL team to adopt zlib-ng over the Intel zlib once it reaches release state.

This is a hope of mine. I had a look at the project earlier looking for comparisons and it definitely seemed the best choice. Far more active than any other implementation.

src/ImageSharp/Formats/Png/Zlib/DeflaterHuffman.cs
JimBobSquarePants: Sorry, was getting mixed up with the `ReadonlySpan` optimization https://github.com/dotnet/roslyn/pull/24621/files

src/ImageSharp/Formats/Png/Zlib/DeflaterEngine.cs
JimBobSquarePants: @Sergio0694 If you have the time, do you mind experimenting with that idea?

PR #1174, opened 2020-04-16
Add support for encoding and decoding IPTC metadata

brianpopow:
### Prerequisites

### Description

This PR adds support for reading and writing IPTC metadata in JPEG files. The actual IPTC data parsing is kindly borrowed from Magick.Net.
Description of parsing APP13 Marker can be found here [App13 Marker](https://metacpan.org/pod/Image::MetaData::JPEG::Structures#Structure-of-a-Photoshop-style-APP13-segment)


src/ImageSharp/Metadata/Profiles/IPTC/IptcProfile.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// Sets the value of the specified tag.
+        /// </summary>
+        /// <param name="tag">The tag of the iptc value.</param>
+        /// <param name="encoding">The encoding to use when storing the bytes.</param>
+        /// <param name="value">The value.</param>
+        public void SetValue(IptcTag tag, Encoding encoding, string value)
+        {
+            Guard.NotNull(encoding, nameof(encoding));
+
+            if (!this.IsRepeatable(tag))
```
brianpopow, line 144: @dlemstra i have changed the original code, so it is now possible to add some entrys more than once. If you think that would be useful for Magick.Net, too, i can change that over there, once this PR is done.

brianpopow:
A thing i was unsure about was, if i should follow the specification strictly and cap some IPTC values accordingly. For example the Category should only be max 3 characters long according to the spec. Here is an overview of the tags with the max length [ExifTool](https://exiftool.org/TagNames/IPTC.html)

It seems no other tool which supports IPTC does honor those max length values, for example irfanview.

JimBobSquarePants:
I'm curious about all the values being strings. It looks to me that there are different value types in the specification. Perhaps we should refactor to allow safely using things like `DateTime`?

JimBobSquarePants:
>A thing i was unsure about was, if i should follow the specification strictly and cap some IPTC values accordingly. 

@brianpopow Do you mean writing as well as reading? I'd like to follow the spec for writing if possible.

brianpopow:
> I'm curious about all the values being strings. It looks to me that there are different value types in the specification. Perhaps we should refactor to allow safely using things like `DateTime`?

If you look at page 32, `ReleaseDate` Property, they state the format should be `CCYYMMDD`

JimBobSquarePants:
>If you look at page 32, ReleaseDate Property, they state the format should be CCYYMMDD

@brianpopow How do you intend on enforcing that though?

JimBobSquarePants:
> Maybe an additional bool parameter strict for the SetValue method (which would default to true)?

I'd consider that yes. Would it be hard to implement?

brianpopow:
> > Maybe an additional bool parameter strict for the SetValue method (which would default to true)?
> 
> I'd consider that yes. Would it be hard to implement?

No, i dont think its hard. The only think im unsure yet is how to deal with the datetime fields.

src/ImageSharp/Metadata/Profiles/IPTC/IptcProfile.cs
dlemstra: Would be awesome if you could also get this merged into Magick.NET. Make sure you reference this issue then: https://github.com/dlemstra/Magick.NET/issues/262. And maybe the API then also needs this: `public IEnumerable<IptcValue> GetValues(IptcTag tag)`?

src/ImageSharp/Metadata/Profiles/IPTC/IptcProfile.cs
```diff
@@ -157,37 +162,68 @@ public void SetEncoding(Encoding encoding)
...
+        }
+
+        /// <summary>
+        /// Makes sure the datetime is formatted according to the iptc specification.
+        /// A date will be formatted as CCYYMMDD.
+        /// A time value will be formatted as HHMMSS±HHMM.
+        /// </summary>
+        /// <param name="tag">The tag of the iptc value.</param>
+        /// <param name="dateTime">The datetime.</param>
+        public void SetDateTimeValue(IptcTag tag, DateTime dateTime)
```
JimBobSquarePants: ❤️

src/ImageSharp/Metadata/Profiles/IPTC/IptcTag.cs
```diff
@@ -14,222 +14,241 @@ public enum IptcTag
...
+        /// Release date. Format should be CCYYMMDD,
+        /// e.g. "19890317" for the 17 March 1989.
+        /// Not repeatable, max length is 8.
         /// </summary>
         ReleaseDate = 30,
 
         /// <summary>
-        /// Release time, not repeatable.
+        /// Release time. Format should be HHMMSS±HHMM,
+        /// e.g. "090000-0500" for 9 o'clock New York time (five hours behind UTC).
```
JimBobSquarePants, line 96: Use `<example>` for block examples. Nest in summary to ensure they show up.

https://docs.microsoft.com/en-us/dotnet/csharp/codedoc#example

src/ImageSharp/Metadata/Profiles/IPTC/IptcValue.cs
```diff
@@ -0,0 +1,186 @@
...
+            {
+                return true;
+            }
+
+            if (this.Tag != other.Tag)
+            {
+                return false;
+            }
+
+            byte[] data = other.ToByteArray();
```
JimBobSquarePants, line 168: Do we do this comparison a lot? The allocation here seems troublesome.

Issue #650, opened 2018-07-08
Allow passing custom Configuration instances to processors


antonfirsov:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Problem
At the moment it's not possible to pass `Configuration` instances to image processors. We can't do this without changing the public API which means we need to act before RC1 is out. 

### Suggestion:
- `IImageProcessor<T>.Apply()` needs a `Configuration` argument
- `IImageProcessingContext<TPixel>` should hold a `Configuration` instead of `MemoryAllocator`
- We need overloads for `Mutate()` and `Clone()` taking a `Configuration`

funcylambda:
@antonfirsov - Do you mind if I take this one to get my feet wet with ImageSharp? :)

antonfirsov:
@funcylambda thanks volunteering, and sorry for the late reaction!

Actually, I received some critical voices on this proposal from @JimBobSquarePants  and @tocsoft, because it's possible to control the `Configuration` instances with the current API by binding them to `Image<T>` instances at initalization.

Their opinion is that it should be good enough for 1.0, mine is that we can add more flexibility with this feature. (`IImageProcessingContext<T>.Configuration` > `Image<T>.Configuration` > `Configuration.Default`)

I think a community PR won't do any harm here, because it won't distract the core team from our main priorities. @JimBobSquarePants @tocsoft thoughts?

funcylambda:
@antonfirsov 

Cool, and no worries! I'm probably going to be refactoring some imaging code from an old C# library to ImageSharp at work (you have some good publicity going around thanks to a few bloggers!), so I thought it'd be good to give back as it were, and also learn a bit more about the library. If this isn't a good task, I do have some spare time this weekend so feel free to point me in the direction of another one :)

funcylambda:
@JimBobSquarePants - Cool, I'll send a PR when it's done/message if I have any problems. Probably towards the end of Sunday.

PR #1176, opened 2020-04-18
Add Property bag to Configuration and Image Processing Context

tocsoft:
### Prerequisites

### Description

This adds a property bag to both the `Configuration` class and the `IImageProcessingContext<T>`  interface to allow for passing defaults and any other contextual information between calls against the context. Includes an initial implementation for `GraphicOptions` class to allow setting global blending options etc across the entire application or just the mutation.

This will likely play more into drawing in the future where we will want to be able to define global fallback fonts (enabling default emoji fonts etc), anti-aliasing fill modes etc.


tocsoft:
I was hoping this would trigger a discussion about this.

> We should decide on a several design questions before continuing:
> 
> 1. The new concept somewhat overlaps with the way we are using `Configuration`. When to prefer values in the property bag? When to prefer strongly-typed properties on `Configuration`?

I can see there being some use cases around encoder and decoder options that might make use of this.

> 2. According to decisions on 1: are there any existing API-s we should change for the sake of consistency in the core library?

Only thing to immediately consider is do we remove the overloads that take a `GraphicsOptions` and instead rely on the `.SetGraphicOptions(options)` calls to provide any settings. This lends itself to the composition and make `TextOptions` & `ShapeOptions` simpler.

> 3. Not urgent, since we can wait with API stability, but: Same question for `ImageSharp.Drawing`. I think this enables composition over inheritance techniques: Have a separate `GraphicsOptions` and `TextOptions` classes Instead of `TextGraphicsOptions : GraphicsOptions` inheritance.

> 4. Shouldn't `IImageProcessingContext.Properties` inherit all values of `Configuration.Properties` by default?

The way I setup the `GetDefaultGraphicsOptions()` was to fallback to the `Configuration`, but it could be simpler to just clone the `Configuration`s options functionally its currently the same but there are edge cases that make it subtly different.

antonfirsov:
> The way I setup the `GetDefaultGraphicsOptions()` was to fallback to the `Configuration`

But in that case you need to do the same manually for other properties, I'd rather prefer to solve this by design.

> Only thing to immediately consider is do we remove the overloads that take a `GraphicsOptions` and instead rely on the `.SetGraphicOptions(options)` calls to provide any settings. 

Can we have a quick API comparison? How does a sample drawing mutation call using `GraphicsOptions` look now VS how would it look with `.SetGraphicOptions(options)`?

antonfirsov:
> I can see there being some use cases around encoder and decoder options that might make use of this.

Currently, we configure encoders/decoders with `Configuration`-wide `IImageEncoder` instances. I doubt it makes sense to store some properties in the bag, while storing others in codec instance members, we should pick one approach.

tocsoft:
```c#
img.Mutate(c => c
    .Vignette(new GraphicsOptions
    {
        AlphaCompositionMode = PixelAlphaCompositionMode.Clear,
        ColorBlendingMode = PixelColorBlendingMode.Darken
    }, Color.Red)
    .Vignette(new GraphicsOptions
    {
        AlphaCompositionMode = PixelAlphaCompositionMode.Clear,
        ColorBlendingMode = PixelColorBlendingMode.Lighten
    }, Color.Green)
    .Vignette(new GraphicsOptions
    {
        AlphaCompositionMode = PixelAlphaCompositionMode.Clear,
        ColorBlendingMode = PixelColorBlendingMode.Multiply
    }, Color.Blue)
);
```
vs using SetGraphicsOptions
```c#
img.Mutate(c => c
    .SetGraphicsOptions(new GraphicsOptions
    {
        AlphaCompositionMode = PixelAlphaCompositionMode.Clear,
        ColorBlendingMode = PixelColorBlendingMode.Darken
    })
    .Vignette(Color.Red)
    .SetGraphicsOptions(o => o.ColorBlendingMode = PixelColorBlendingMode.Lighten)
    .Vignette(Color.Green)
    .SetGraphicsOptions(o => o.ColorBlendingMode = PixelColorBlendingMode.Multiply) // alpha still set to Clear
    .Vignette(Color.Blue)
);
```

JimBobSquarePants:
I need to have a proper think about this. I'm not sure why `GraphicsOptions` should be treated any differently to other options types. (`ResizeOptions` for example). Yes, they can be reused for different methods but they a still per-operation options.

tocsoft:
don't worry about it too much... it was just a thought really, we'll leave the overloads on that take an options too. 

We definitely need/want the extensibility to load defaults and pass global parameters otherwise Drawing can't add global config options.

src/ImageSharp/Configuration.cs
```diff
@@ -73,6 +76,12 @@ public int MaxDegreeOfParallelism
             }
         }
 
+        /// <summary>
+        /// Gets a set of properties for the Congiguration.
+        /// </summary>
+        /// <remarks>This can be used for storing global settings and defaults to be accessable to processors.</remarks>
+        public IDictionary<object, object> Properties => this.properties;
```
JimBobSquarePants, line 83: We can use an auto property with initializer here.

tocsoft:
I really want to keep the version on context too  as it opens up a world of options.. I will want to leverage it more on the drawing side and wanted to get it in now.
I don't think it harms anything and only give us flexibility without having to orchestrate multi repo releases especially after RC/1.0.

JimBobSquarePants:
>I really want to keep the version on context too as it opens up a world of options.. I will want to leverage it more on the drawing side and wanted to get it in now.

I don't understand the use case.

The `Configuration` class and it's property bag would already be available to the context and processors either by direct injection or  from the source image. I don't get what the additional property collection offers?

tocsoft:
`Configuration` is a system wide global... or should be.

`ImageProcessingContext` is explicitly a single processing run on a single image.

Its the same difference as there being a Headers collection no `HttpClient` and a separate one on `HttpRequestMessage` one if for globally shared values one is for unit of work wide values.

Main use case is on processing SVGs where we want to be able to record and apply transforms and masks on parts of a processing tree and then revert back later without having to pass loads of complex options around..... not everything in the tree will even need to know about the setting from higher up it would just work and compose into the correct setting. We shouldn't really be encouraging people to create new configuration instances all the time everything should just compose and work. Your point of reference for the API at the moment is based ont he straigh processor base manipulations that are mostly independent of each other with Drawing they are all variants and interaction on the same underlying processor, the way you want to use the API is different... You don't want to have to call a transform on every single vector on the way thru every time you want to be able to do something like `SetOrigin(xxx)`, `SetTransform(yyy)` and it to just apply to all the calls after that point, but you don't want those to apply to this 2nd image for this 2nd request.

JimBobSquarePants:
`Configuration` is considered global yes but we explicitly added the ability to pass instances to individual or chained processors as part of #650 

`ImageProcessingContext` is a single context that can be applied to multiple methods, not just one. This allows sharing of the options but not the fine grained options like you are suggesting.

``` c#
using (var image = new Image<Rgba32>(100, 100))
{
    var contextConfig = new Configuration();
    image.Mutate(
        contextConfig,
         x => x.Hue(180)
               .Pixelate(4)
               .Saturate(.75F)
               .Vignette());
}
``` 

So I see now where you are coming from now API-wise. It's definitely nice to set the options via an action than by splitting up or constructing individual options. `SetGraphicsOptions()` is neat and allows much reduced boilerplate. The idividual property bag works well for this.

Do we need the `GetGraphicsOptions()` methods though since we always have access to the options via the action context for setting?

tocsoft:
> Do we need the GetGraphicsOptions() methods though since we always have access to the options via the action context for setting?

Yeah we do.

Without the `GetGraphicsOptions()` every call site that needs the defaults will need to have the same logic repeated for try getting it from context, oh its not there lets try the try getting it from config instead, still not there let allocate a new instance. (we also no no longer need to allocate every time too as we cache the config wide version.

src/ImageSharp/GraphicOptionsDefaultsExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+            context.Properties[typeof(GraphicsOptions)] = cloned;
+            return context;
+        }
+
+        /// <summary>
+        /// Sets the default options against the configuration.
+        /// </summary>
+        /// <param name="context">The image processing context to store default against.</param>
+        /// <param name="optionsBuilder">The default options to use.</param>
+        public static void SetGraphicsOptions(this Configuration context, Action<GraphicsOptions> optionsBuilder)
```
antonfirsov, line 33:
```suggestion
        /// <param name="configuration">The configuration to store default against.</param>
        /// <param name="optionsBuilder">The default options to use.</param>
        public static void SetGraphicsOptions(this Configuration configuration, Action<GraphicsOptions> optionsBuilder)
```

src/ImageSharp/GraphicOptionsDefaultsExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+            context.Properties[typeof(GraphicsOptions)] = options;
+            return context;
+        }
+
+        /// <summary>
+        /// Sets the default options against the configuration.
+        /// </summary>
+        /// <param name="context">The image processing context to store default against.</param>
+        /// <param name="options">The default options to use.</param>
+        public static void SetGraphicsOptions(this Configuration context, GraphicsOptions options)
```
antonfirsov, line 57: See previous suggestion.

src/ImageSharp/Processing/IImageProcessingContext.cs
```diff
@@ -15,6 +16,12 @@ public interface IImageProcessingContext
         /// </summary>
         Configuration Configuration { get; }
 
+        /// <summary>
+        /// Gets a set of properties for the Image Processing Context.
+        /// </summary>
+        /// <remarks>This can be used for storing global settings and defaults to be accessable to processors.</remarks>
+        IDictionary<object, object> Properties { get; }
```
antonfirsov, line 23: I wonder if we want to consider something more strongly-typed here?
- `System.Type` or a custom object as a key
- Restrictions for the value: do we expect simple values here? If not, maybe a marker interface.
- Our own types `GraphicsProperties` / `IGraphicsProperties` instead of `IDictionary<K,V>`

src/ImageSharp/GraphicOptionsDefaultsExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <param name="context">The image processing context to retrieve defaults from.</param>
+        /// <returns>The globaly configued default options.</returns>
+        public static GraphicsOptions GetGraphicsOptions(this IImageProcessingContext context)
+        {
+            if (context.Properties.TryGetValue(typeof(GraphicsOptions), out var options) && options is GraphicsOptions go)
+            {
+                return go;
+            }
+
+            var configOptions = context.Configuration.GetGraphicsOptions();
```
antonfirsov, line 74: Are we really sure we want the defaulting logic in the `Get***` extension methods, instead of populating `Properties` in the `DefaultImageProcessingContext` ctr?

What are the pros for this decision?

James-Jackson-South: Ah yeah... If our keys are strongly typed that makes comparison faster and easier too.

JimBobSquarePants: Ah yeah... If our keys are strongly typed that makes comparison faster and easier too.

antonfirsov: FYI an interesting API proposal that also contains the concept of property bags: dotnet/runtime#1793

In total, a bit too verbose for my style, but I like the idea of having custom types for the bag, and (in some form) for the keys. I would also apply it for values, if it makes sense.

src/ImageSharp/Processing/IImageProcessingContext.cs
tocsoft: no it should be explicitly be as loosely typed as possible and we should not be limiting anything from being but in the property bag, that's why I even when as far as using `<object,object>` rather than `<string,object>` to allow developers to use it for there needs they can do anything with it... use strings, use private object instances (no way for someone to accidently reuse the key) or anything in between... we would be loosing one of the key flexibility benefits of this extension point is we make the type a custom type with restrictions..

src/ImageSharp/GraphicOptionsDefaultsExtensions.cs
tocsoft: I do kinda like the pattern proposed in the runtime but, (and it is a big but) it feels soo overengineered, sometimes (and in my experience usually) the simple solution is the right one.

PR #1178, opened 2020-04-21
Always read CRC.

### LINKED ISSUE #1177: PngDecoder omits reading CRC, resulting in ImageFormatException
### Prerequisites

### Description
Attempting to download a specific image to stream via an HttpClient and then loading it via Image.Load<Rgba32>() throws the following exception:
`SixLabors.ImageSharp.ImageFormatException: 'PNG Image does not contain a data chunk'`

For an other case it result in 
`SixLabors.ImageSharp.ImageFormatException: 'Bad method for ZLIB header: cmf=132'`

From what i can see in the code PngDecoderCore.ValidateChunk only reads the CRC information on critical chunks. It should however always read the CRC value as it is always part of a chunk.
The current code has a hack around not reading CRC by skipping 4 bytes at the time if next chunk length is bigger than remaining stream length. This fails if CRC value is less than remaining stream length. Then CRC value will be chunk length and chunk length will the chunk type.

Should be fixable by removing 
```
if (!chunk.IsCritical)
            {
                return;
            }
```
And move the is critical to 
```
if (this.crc.Value != crc && chunk.IsCritical)
            {
                string chunkTypeName = Encoding.ASCII.GetString(chunkType);
                PngThrowHelper.ThrowInvalidChunkCrc(chunkTypeName);
            }
```

Have looked at Issue 1127 and 1128, similar but not the same.

### Steps to Reproduce
```
var url = "https://assets-cylindo-com.s3.amazonaws.com/1895/2599314/13/small.png";//PNG 
//var url = "https://assets-cylindo-com.s3.amazonaws.com/1535/2700525/1/small.png";// Bad method..
var response = await new HttpClient( ).GetAsync( url );
var stream = await response.Content.ReadAsStreamAsync( );
var image = Image.Load<Rgba32>( stream ); // throws on this line
```

### System Configuration

- ImageSharp version: 1.0.0-unstable1087
- Other ImageSharp packages and versions: SixLabors.ImageSharp.Drawing 1.0.0-unstable0436
- .NET Framework version: .Net Core 3.1
- Additional information:

JimBobSquarePants:
### Prerequisites

### Description
Fix #1177


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1155,7 +1150,7 @@ private void ValidateChunk(in PngChunk chunk)
             this.crc.Update(chunk.Data.GetSpan());
 
             uint crc = this.ReadChunkCrc();
-            if (this.crc.Value != crc)
+            if (this.crc.Value != crc && chunk.IsCritical)
```
brianpopow, line 1153: what i dont like about this change: we calculate the crc for none critical chunks, but then ignore it.

Maybe we could always read the crc bytes on the start of the method instead?

JimBobSquarePants: Yeah much better

JimBobSquarePants: Yeah much better

PR #1179, opened 2020-04-21
Add exception type and update xml docs. 

JimBobSquarePants:
### Prerequisites

### Description
- Adds new `InvalidImageContentException` type.
- Documents exceptions and adds guards on all loading mechanisms.
- Documents exceptions and adds guards on all saving mechanisms.
- Minor cleanup and performance tweaks.

Fixes. #1110


src/ImageSharp/Image.FromBytes.cs
```diff
@@ -65,13 +69,16 @@ public static IImageFormat DetectFormat(Configuration configuration, byte[] data
...
         /// <param name="data">The byte array containing encoded image data to read the header from.</param>
         /// <param name="format">The format type of the decoded image.</param>
-        /// <exception cref="NotSupportedException">Thrown if the stream is not readable.</exception>
+        /// <exception cref="ArgumentNullException">The configuration is null.</exception>
+        /// <exception cref="ArgumentNullException">The data is null.</exception>
+        /// <exception cref="NotSupportedException">The data is not readable.</exception>
         /// <returns>
         /// The <see cref="IImageInfo"/> or null if suitable info detector is not found.
         /// </returns>
         public static IImageInfo Identify(Configuration configuration, byte[] data, out IImageFormat format)
```
antonfirsov, line 78: Can decoders throw `InvalidImageContentException` in `Identify`?

JimBobSquarePants: Yeah, they can hit a bad value in the image header.

JimBobSquarePants:
> Great job! Will have another one in a few days.

Quick as possible if you can. I'm hoping to ship in less than a week.

brianpopow:
@JimBobSquarePants there are still some places left with a `TODO: use InvalidImageContentException here, if we decide to define it`

For example https://github.com/SixLabors/ImageSharp/blob/5639babe23a438958f5f59db86c332728f15707f/src/ImageSharp/Formats/Gif/GifDecoder.cs#L43.

Was this intentional or should we change those, too?

JimBobSquarePants:
@brianpopow I missed that. Yeah we should update.

brianpopow:
> @brianpopow I missed that. Yeah we should update.

@JimBobSquarePants ok i will do that

PR #1181, opened 2020-04-22
Add DegenerateTransformException

### LINKED ISSUE #966: Consider detecting degenerate transformations in AffineTransformBuilder.BuildMatrix()
### Prerequisites

### Description
The code below:
```csharp
var builder = new AffineTransformBuilder().AppendSkewDegrees(45, 45);
var rectangle = new Rectangle(0, 0, 150, 150);
var matrix = builder.BuildMatrix(rectangle);
```
Will produce:
image

`AffineTransformBuilder` is pretty a high-level class, therefore it should probably help users to detect errors early, so I would consider throwing something like `DegenerateTransformException` instead.

### System Configuration
- ImageSharp version: current main
- Other ImageSharp packages and versions: -
- Environment (Operating system, version and so on): any
- .NET Framework version: any
- Additional information: -

JimBobSquarePants:
### Prerequisites

### Description
Capture and throw degenerate matrices created by the transform builders.

Fix #966


src/ImageSharp/Processing/Processors/Transforms/DegenerateTransformException.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the Apache License, Version 2.0.
+
+using System;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Transforms
+{
+    /// <summary>
+    /// Represents an error that occurs during a transform operation.
+    /// </summary>
+    public sealed class DegenerateTransformException : ImageProcessingException
```
antonfirsov, line 11: I don't think this inheritance is a good idea. `ImageProcessingException` is usually thrown by ImageProcessor-s, the TransformBuilder concept looks fully orthogonal to image processing.

src/ImageSharp/Processing/ProjectiveTransformBuilder.cs
```diff
@@ -300,6 +303,11 @@ public Matrix4x4 BuildMatrix(Rectangle sourceRectangle)
                 matrix *= factory(size);
             }
 
+            if (TransformUtilities.IsNaN(matrix))
```
antonfirsov, line 306: I would also:
- Add this check to `PrependMatrix(...)`  / `AppendMatrix(...)` (better exception locality)
- Add a check to verify if the [determinant](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.matrix4x4.getdeterminant) is a near-zero value. (Both here and to individual append methods.)

src/ImageSharp/Processing/ProjectiveTransformBuilder.cs
```diff
@@ -300,6 +303,11 @@ public Matrix4x4 BuildMatrix(Rectangle sourceRectangle)
...
+            {
+                throw new DegenerateTransformException("Matrix is NaN. Check input values.");
```
antonfirsov, line 308:
```suggestion
                throw new DegenerateTransformException("Matrix is degenerate. Check input values.");
```

src/ImageSharp/Processing/Processors/Transforms/DegenerateTransformException.cs
JimBobSquarePants: I guess... I was thinking in processing pipelines.

src/ImageSharp/Processing/ProjectiveTransformBuilder.cs
JimBobSquarePants: `PrependMatrix(...) / AppendMatrix(...)` only accept factories. We don't know the value of the matrix until we loop through and apply the functions.

A yes, good catch on the determinant! We can't invert the matrix then! 

How large an epsilon do you suggest?

antonfirsov: > `PrependMatrix(...) / AppendMatrix(...)` only accept factories.

That's the generic `Prepend` / `Append`. `PrependMatrix()` / `AppendMatrix()` takes a matrix and converts it to a lambda, so we have a chance to validate it's value in those methods.

> How large an epsilon do you suggest?

I think we should use [Constants.EpsilonSquared](https://github.com/SixLabors/ImageSharp/blob/08dcb0707969cdaa957ac8cce696ca4b21ec2def/src/ImageSharp/Common/Constants.cs#L19). Not 100% sure, but it's not critical.

src/ImageSharp/Processing/Processors/Transforms/DegenerateTransformException.cs
antonfirsov: > I was thinking in processing pipelines.

Do you mean Mutate / Clone pipelines? Is it used there? In my mental model these builders are separate utilities.

src/ImageSharp/Processing/ProjectiveTransformBuilder.cs
JimBobSquarePants: Cool. All done

src/ImageSharp/Processing/Processors/Transforms/DegenerateTransformException.cs
JimBobSquarePants: They are. I just struggle to get my head out of the pipeline way of thinking at times. I've changed it now anyway

src/ImageSharp/Processing/Processors/Transforms/TransformUtilities.cs
```diff
@@ -12,6 +12,57 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
...
+        /// Returns a value that indicates whether the specified matrix is degenerate
+        /// containing one or more values equivalent to <see cref="float.NaN"/> or a
+        /// zero determinant and therefore cannot be used for linear transforms.
+        /// </summary>
+        /// <param name="matrix">The transform matrix.</param>
+        public static bool IsDegenerate(Matrix4x4 matrix)
+            => IsNaN(matrix) || IsZero(matrix.GetDeterminant());
+
+        [MethodImpl(InliningOptions.ShortMethod)]
+        private static bool IsZero(float a)
```
brianpopow, line 34: This should maybe called IsNearZero then?

Other then that this PR looks good to be merged to me.

JimBobSquarePants: For future if it's ever reused yeah, I only extracted it for readability.

tests/ImageSharp.Tests/Drawing/DrawImageExtensionsTests.cs
```diff
@@ -28,7 +20,7 @@ public void DrawImage_OpacityOnly_VerifyGraphicOptionsTakenFromContext()
             this.options.ColorBlendingMode = PixelColorBlendingMode.Screen;
 
             this.operations.DrawImage(null, 0.5f);
-            var dip = this.Verify<DrawImageProcessor>();
+            DrawImageProcessor dip = this.Verify<DrawImageProcessor>();
```
JimBobSquarePants, line 23: @brianpopow Thanks for fixing these. They were starting to annoy me 👍

Issue #410, opened 2017-12-22
Image.Load<Rgba32> Results in 'Bad compression method for ZLIB header: cmf=237' Exception


zeus82:
### Prerequisites

- [ X] I have written a descriptive issue title
- [ X] I have verified that I am running the latest version of ImageSharp
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [ X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Try loading the attached image and see that it generates a 'Bad compression method for ZLIB header: cmf=237' error.  System.Drawing can open this image.

### Steps to Reproduce
Image.Load<Rgba32>(File.OpenRead("1.png"))

### System Configuration
.NET Standard 2.0 Dll hosted in a .NET Framework 4.6.1 windows service

- ImageSharp version: 1.0.0-beta0002
- Other ImageSharp packages and versions: Sizlabors.Core 1.0.0-beta0004
- Environment (Operating system, version and so on): Windows 10
- .NET Framework version: 4.6.1
- Additional information:

![1](https://user-images.githubusercontent.com/6005996/34303888-73d65ca6-e705-11e7-88cf-6c16ad871172.png)

JimBobSquarePants:
Hi @zeus82 

Thanks for the detail and the input image. It turns out that image is not a valid png.

![cgbi](https://user-images.githubusercontent.com/385879/34320495-b3783640-e84f-11e7-8f45-470cbbb1c28b.PNG)

http://iphonedevwiki.net/index.php/CgBI_file_format

There's quite a few differences between the format

- extra critical chunk (CgBI)
- byteswapped (RGBA -> BGRA) pixel data, presumably for high-speed direct blitting to the framebuffer
- zlib header, footer, and CRC removed from the IDAT chunk
- premultiplied alpha (color' = color * alpha / 255)

@antonfirsov @tocsoft @dlemstra 

Theoretically it looks like we could potentially handle these changes with some updates to our decoder based on the following:

http://www.axelbrz.com.ar/?mod=iphone-png-images-normalizer

I don't know if we should attempt this for V1 though. What do you think?

tocsoft:
[...]

PR #1184, opened 2020-04-23
Additional tests

brianpopow:
### Prerequisites

### Description

This PR adds additional unit tests for:

- Setting iptc value with strict option disabled
- Tests for jpg encoder preserving exif, iptc, icc metadata when encoding a image
- Additional test cases for png: Some edge cases and some additional png variants. Note: I have splitted up the `Common` test cases in tests with some more meaningful names.

Another change was made for #410: When a chunk is detected which is from a apple PNG which does not follow the spec, a exception with a better error message is thrown.

I noticed that we do not check the CRC for the IDAT chunks of a PNG. I could not figure out a good solution for that without reading the IDAT chunk twice.


src/ImageSharp/Formats/Png/PngChunkType.cs
```diff
@@ -73,6 +73,58 @@ internal enum PngChunkType : uint
...
+        /// The cHRM chunk may be used to specify the 1931 CIE x,y chromaticities of the red,
+        /// green, and blue display primaries used in the image, and the referenced white point.
+        /// </summary>
+        Chroma = 0x6348524d,
+
+        /// <summary>
+        /// Malformed chunk named CgBI produced by apple, which is not conform to the specification.
+        /// Related issue is here https://github.com/SixLabors/ImageSharp/issues/410
+        /// </summary>
+        MalformedApple = 0x43674249
```
JimBobSquarePants, line 128: `ProprietaryApple`. I think we could theoretically support this in the future.

- Skip Zlib header, footer, and CRC when reading the IDAT chunk
- Read as premultiplied Bgra instead of Rgba

http://iphonedevwiki.net/index.php/CgBI_file_format#CgBI_Chunk

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -215,6 +215,9 @@ public Image<TPixel> Decode<TPixel>(Stream stream)
                             case PngChunkType.End:
                                 this.isEndChunkReached = true;
                                 break;
+                            case PngChunkType.MalformedApple:
+                                PngThrowHelper.ThrowInvalidChunkType("Malformed Apple PNG detected! This PNG file is not conform to the specification and cannot be decoded.");
```
JimBobSquarePants, line 219:
```suggestion
                                PngThrowHelper.ThrowInvalidChunkType("Proprietary Apple PNG detected! This PNG file is not conform to the specification and cannot be decoded.");
```

src/ImageSharp/Formats/Png/PngChunkType.cs
JimBobSquarePants, line 128:
```suggestion
        ProprietaryApple = 0x43674249
```

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -215,6 +215,9 @@ public Image<TPixel> Decode<TPixel>(Stream stream)
                             case PngChunkType.End:
                                 this.isEndChunkReached = true;
                                 break;
+                            case PngChunkType.MalformedApple:
```
JimBobSquarePants, line 218:
```suggestion
                            case PngChunkType.ProprietaryApple:
```

src/ImageSharp/Formats/Png/PngThrowHelper.cs
```diff
@@ -20,11 +20,14 @@ internal static class PngThrowHelper
...
         public static void ThrowInvalidChunkType() => throw new InvalidImageContentException("Invalid PNG data.");
 
+        [MethodImpl(InliningOptions.ColdPath)]
+        public static void ThrowInvalidChunkType(string message) => throw new InvalidImageContentException(message);
+
         [MethodImpl(InliningOptions.ColdPath)]
         public static void ThrowInvalidChunkCrc(string chunkTypeName) => throw new InvalidImageContentException($"CRC Error. PNG {chunkTypeName} chunk is corrupt!");
 
         [MethodImpl(InliningOptions.ColdPath)]
-        public static void ThrowNotSupportedColor() => new NotSupportedException("Unsupported PNG color type");
```
JimBobSquarePants, line 27: Good catch!

src/ImageSharp/Formats/Png/PngChunkType.cs
brianpopow: yeah theoretical, but who knows what else from the spec they do not follow.

JimBobSquarePants: Yeah, could be troublesome. Something for the future anyway....

PR #1188, opened 2020-04-27
Gamma chunk should come before palette chunk

brianpopow:
### Prerequisites

### Description

This PR changes the PNG encoder, so it will write the Gamma chunk before the palette chunk as recommended in the [Specification](https://www.w3.org/TR/2003/REC-PNG-20031110/#table53)

I have also added tests to ensure the order of the chunks.


tests/ImageSharp.Tests/Formats/Png/PngEncoderTests.cs
```diff
@@ -2,6 +2,9 @@
 // Licensed under the Apache License, Version 2.0.
 
 // ReSharper disable InconsistentNaming
```
JimBobSquarePants, line 4: Can you please clean up these tests to remove the warnings. It just reduces the clutter when reading.

brianpopow: sure, somehow i did not noticed that warning.

JimBobSquarePants: Perfect. Thanks!

PR #1190, opened 2020-04-28
Change ImageFormatException to InvalidImageContentException

brianpopow:
### Prerequisites

### Description

Change a few exceptions to `InvalidImageContentException` which were missed in #1179

### DIFF
--- a/src/ImageSharp/Formats/Bmp/BmpDecoder.cs
+++ b/src/ImageSharp/Formats/Bmp/BmpDecoder.cs
@@ -43,9 +43,7 @@ namespace SixLabors.ImageSharp.Formats.Bmp
             {
                 Size dims = decoder.Dimensions;
 
-                // TODO: use InvalidImageContentException here, if we decide to define it
-                // https://github.com/SixLabors/ImageSharp/issues/1110
-                throw new ImageFormatException($"Can not decode image. Failed to allocate buffers for possibly degenerate dimensions: {dims.Width}x{dims.Height}. This error can happen for very large RLE bitmaps, which are not supported.", ex);
+                throw new InvalidImageContentException($"Can not decode image. Failed to allocate buffers for possibly degenerate dimensions: {dims.Width}x{dims.Height}. This error can happen for very large RLE bitmaps, which are not supported.", ex);
             }
         }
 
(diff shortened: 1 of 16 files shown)


src/ImageSharp/Formats/Bmp/BmpDecoder.cs
```diff
@@ -43,9 +43,7 @@ public Image<TPixel> Decode<TPixel>(Configuration configuration, Stream stream)
             {
                 Size dims = decoder.Dimensions;
 
-                // TODO: use InvalidImageContentException here, if we decide to define it
-                // https://github.com/SixLabors/ImageSharp/issues/1110
-                throw new ImageFormatException($"Can not decode image. Failed to allocate buffers for possibly degenerate dimensions: {dims.Width}x{dims.Height}. This error can happen for very large RLE bitmaps, which are not supported.", ex);
+                throw new InvalidImageContentException($"Can not decode image. Failed to allocate buffers for possibly degenerate dimensions: {dims.Width}x{dims.Height}. This error can happen for very large RLE bitmaps, which are not supported.", ex);
```
JimBobSquarePants, line 46: Can you move these and others to the `ThrowHelper` classes for each format please. That's why I missed them.

src/ImageSharp/Formats/Gif/GifDecoder.cs
```diff
@@ -38,9 +38,10 @@ public Image<TPixel> Decode<TPixel>(Configuration configuration, Stream stream)
             {
                 Size dims = decoder.Dimensions;
 
-                // TODO: use InvalidImageContentException here, if we decide to define it
-                // https://github.com/SixLabors/ImageSharp/issues/1110
-                throw new ImageFormatException($"Can not decode image. Failed to allocate buffers for possibly degenerate dimensions: {dims.Width}x{dims.Height}.", ex);
+                GifThrowHelper.ThrowInvalidImageContentException($"Can not decode image. Failed to allocate buffers for possibly degenerate dimensions: {dims.Width}x{dims.Height}.", ex);
+
+                // Not reachable, as the previous statement will throw a exception.
+                return null;
```
brianpopow, line 44: @JimBobSquarePants is there a way to annotate a method, that it will always throw to avoid this return statement?

JimBobSquarePants: Unfortunately not. There's something coming in NET5 I believe but it won't be backwards compatible.

PR #1196, opened 2020-05-06
Async APIs / Fake Async Codecs

tocsoft:
### Prerequisites

### Description
This adds the API surface to support async saving and loading from streams.

This is really a bit of a cheat and doesn't actually process the stream asynchronously just helps prevent people, when loading from aspnetcore setting where syncIO is disabled, having to manually do the memory copy we keep tellign them to do.. we just do it for you if we believe it will be needed and allows us (in the future) to start actually implement the codecs asynchronously as and when we want.

This API addition does raise the whether we need/want to introduce sync `XXXWithFormat()` apis to match the async ones and depricate/hide the overloads with `out` the params to keep the sync and async APIs in sync.

Also should the `FormattedImage` structs be made `IDisposable` and pass the dispose along to the `Image` so you can do 
```csharp
using(var i = async Image.ImageWithFormat(stream)){
     i.Image.DoStuff();
}
```
vs
```csharp
var i = async Image.ImageWithFormat(stream) 
using(var img = i.Image){
     img.DoStuff();
}
```
or do we use some other way of representing those sorts of APIs?


src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -130,8 +131,40 @@ public BmpDecoderCore(Configuration configuration, IBmpDecoderOptions options)
...
             where TPixel : unmanaged, IPixel<TPixel>
+        {
+            // if we can seek then we arn't in a context that errors on async operations
+            if (stream.CanSeek)
+            {
+                return this.Decode<TPixel>(stream);
+            }
+            else
+            {
+                // cheat for now do async copy of the stream into memory stream and use the sync version
```
JimBobSquarePants, line 144: I use a pool backed memory stream that allocates small chunks. Not sure how performant reading would be though.

https://github.com/SixLabors/ImageSharp.Web/blob/master/src/ImageSharp.Web/ChunkedMemoryStream.cs

brianpopow: Is maybe [RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream) an option? Its meant to be a drop in replacement for `MemoryStream`

src/ImageSharp/Formats/Jpeg/JpegDecoder.cs
```diff
@@ -17,6 +18,28 @@ public sealed class JpegDecoder : IImageDecoder, IJpegDecoderOptions, IImageInfo
...
+        /// <inheritdoc/>
+        public async Task<Image<TPixel>> DecodeAsync<TPixel>(Configuration configuration, Stream stream)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            Guard.NotNull(stream, nameof(stream));
+
+            using var decoder = new JpegDecoderCore(configuration, this);
+            try
+            {
+                return await decoder.DecodeAsync<TPixel>(stream);
```
brianpopow, line 30:
```suggestion
                return await decoder.DecodeAsync<TPixel>(stream).ConfigureAwait(false);
```

src/ImageSharp/Formats/Png/PngEncoder.cs
```diff
@@ -56,5 +57,21 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
...
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="image">The <see cref="Image{TPixel}"/> to encode from.</param>
+        /// <param name="stream">The <see cref="Stream"/> to encode the image data to.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        public async Task EncodeAsync<TPixel>(Image<TPixel> image, Stream stream)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            using (var encoder = new PngEncoderCore(image.GetMemoryAllocator(), image.GetConfiguration(), new PngEncoderOptions(this)))
+            {
+                await encoder.EncodeAsync(image, stream);
```
brianpopow, line 73:
```suggestion
                await encoder.EncodeAsync(image, stream).ConfigureAwait(false);
```

src/ImageSharp/Image.Decode.cs
```diff
@@ -107,10 +120,32 @@ private static (Image<TPixel> img, IImageFormat format) Decode<TPixel>(Stream st
...
+        private static async Task<FormattedImage<TPixel>> DecodeAsync<TPixel>(Stream stream, Configuration config)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            IImageDecoder decoder = DiscoverDecoder(stream, config, out IImageFormat format);
+            if (decoder is null)
+            {
+                return (null, null);
+            }
+
+            Image<TPixel> img = await decoder.DecodeAsync<TPixel>(config, stream);
```
brianpopow, line 144:
```suggestion
            Image<TPixel> img = await decoder.DecodeAsync<TPixel>(config, stream).ConfigureAwait(false);
```

src/ImageSharp/Image.Decode.cs
```diff
@@ -119,7 +154,38 @@ private static (Image img, IImageFormat format) Decode(Stream stream, Configurat
...
+
+        private static async Task<FormattedImage> DecodeAsync(Stream stream, Configuration config)
+        {
+            IImageDecoder decoder = DiscoverDecoder(stream, config, out IImageFormat format);
+            if (decoder is null)
+            {
+                return (null, null);
+            }
+
+            Image img = await decoder.DecodeAsync(config, stream);
```
brianpopow, line 168:
```suggestion
            Image img = await decoder.DecodeAsync(config, stream).ConfigureAwait(false);
```

src/ImageSharp/Image.cs
```diff
@@ -157,6 +181,12 @@ public void Visit<TPixel>(Image<TPixel> image)
             {
                 this.encoder.Encode(image, this.stream);
             }
+
+            public Task VisitAsync<TPixel>(Image<TPixel> image)
+                where TPixel : unmanaged, IPixel<TPixel>
+            {
+                return this.encoder.EncodeAsync(image, this.stream);
```
brianpopow, line 188:
```suggestion
                return this.encoder.EncodeAsync(image, this.stream).ConfigureAwait(false);
```

src/ImageSharp/Image.FromStream.cs
```diff
@@ -64,6 +100,39 @@ public static IImageInfo Identify(Stream stream)
...
+        /// <exception cref="ArgumentNullException">The configuration is null.</exception>
+        /// <exception cref="ArgumentNullException">The stream is null.</exception>
+        /// <exception cref="NotSupportedException">The stream is not readable.</exception>
+        /// <exception cref="InvalidImageContentException">Image contains invalid content.</exception>
+        /// <returns>
+        /// The <see cref="IImageInfo"/> or null if suitable info detector is not found.
+        /// </returns>
+        public static async Task<IImageInfo> IdentifyAsync(Configuration configuration, Stream stream)
+        {
+            FormattedImageInfo res = await IdentifyWithFormatAsync(configuration, stream);
```
brianpopow, line 132:
```suggestion
            FormattedImageInfo res = await IdentifyWithFormatAsync(configuration, stream).ConfigureAwait(false);
```

src/ImageSharp/Image.FromStream.cs
```diff
@@ -159,6 +317,23 @@ public static Image Load(Configuration configuration, Stream stream, IImageDecod
...
+        /// <param name="stream">The stream containing image information.</param>
+        /// <exception cref="ArgumentNullException">The configuration is null.</exception>
+        /// <exception cref="ArgumentNullException">The stream is null.</exception>
+        /// <exception cref="NotSupportedException">The stream is not readable.</exception>
+        /// <exception cref="UnknownImageFormatException">Image format not recognised.</exception>
+        /// <exception cref="InvalidImageContentException">Image contains invalid content.</exception>
+        /// <returns>A new <see cref="Image"/>.</returns>>
+        public static async Task<Image> LoadAsync(Configuration configuration, Stream stream)
+        {
+            var fmt = await LoadWithFormatAsync(configuration, stream);
```
brianpopow, line 333:
```suggestion
            var fmt = await LoadWithFormatAsync(configuration, stream).ConfigureAwait(false);
```

src/ImageSharp/Image.FromStream.cs
```diff
@@ -272,6 +507,89 @@ public static Image<TPixel> Load<TPixel>(Configuration configuration, Stream str
...
+        /// <exception cref="ArgumentNullException">The stream is null.</exception>
+        /// <exception cref="NotSupportedException">The stream is not readable.</exception>
+        /// <exception cref="UnknownImageFormatException">Image format not recognised.</exception>
+        /// <exception cref="InvalidImageContentException">Image contains invalid content.</exception>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <returns>A new <see cref="Image{TPixel}"/>.</returns>
+        public static async Task<Image<TPixel>> LoadAsync<TPixel>(Configuration configuration, Stream stream)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            (Image<TPixel> img, _) = await LoadWithFormatAsync<TPixel>(configuration, stream);
```
brianpopow, line 589:
```suggestion
            (Image<TPixel> img, _) = await LoadWithFormatAsync<TPixel>(configuration, stream).ConfigureAwait(false);
```

src/ImageSharp/Image.FromStream.cs
```diff
@@ -336,5 +654,37 @@ private static T WithSeekableStream<T>(Configuration configuration, Stream strea
...
+            // if the stream is seekable then we arn't using one of the aspnetcore wrapped streams that error on sync api calls and we can use it with out
+            // having to further wrap
+            if (stream.CanSeek)
+            {
+                if (configuration.ReadOrigin == ReadOrigin.Begin)
+                {
+                    stream.Position = 0;
+                }
+
+                return await action(stream);
```
brianpopow, line 678:
```suggestion
                return await action(stream).ConfigureAwait(false);
```

src/ImageSharp/Image.FromStream.cs
```diff
@@ -336,5 +654,37 @@ private static T WithSeekableStream<T>(Configuration configuration, Stream strea
...
+            }
+
+            using (var memoryStream = new MemoryStream()) // should really find a nice way to use a pool for these!!
+            {
+                await stream.CopyToAsync(memoryStream);
```
brianpopow, line 683:
```suggestion
                await stream.CopyToAsync(memoryStream).ConfigureAwait(false);
```

src/ImageSharp/Image.FromStream.cs
```diff
@@ -336,5 +654,37 @@ private static T WithSeekableStream<T>(Configuration configuration, Stream strea
...
+                memoryStream.Position = 0;
+
+                return await action(memoryStream);
```
brianpopow, line 686:
```suggestion
                return await action(memoryStream).ConfigureAwait(false);
```

JimBobSquarePants:
@tocsoft I'll review this over the weekend.

antonfirsov:
I will also try my best to have a look, but I need a couple of days, hope to make it before Saturday.

JimBobSquarePants:
Making `FormattedImage` etc implement `IDisposable` would certainly make the code less awkward to consume, I'm not sure about the naming of those types though as they suggest to me that they affect the layout of the image in memory.  I'm, thinking something like `FormatImagePair`

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
```diff
@@ -15,6 +17,47 @@ namespace SixLabors.ImageSharp.Advanced
     /// </summary>
     public static class AdvancedImageExtensions
     {
+        /// <summary>
+        /// For a given path find the best encoder to use
+        /// </summary>
+        /// <param name="source">The source.</param>
+        /// <param name="path">The Path</param>
+        /// <returns>The matching encoder.</returns>
+        public static IImageEncoder FindEncoded(this Image source, string path)
```
JimBobSquarePants, line 26:
```suggestion
        public static IImageEncoder DetectEncoder(this Image source, string path)
```

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
```diff
@@ -76,8 +74,8 @@ public static void AcceptVisitor(this Image source, IImageVisitor visitor)
         /// <param name="source">The source.</param>
         /// <param name="visitor">The visitor.</param>
         /// <returns>A  <see cref="Task"/> representing the asynchronous operation.</returns>
-        public static Task AcceptVisitorAsync(this Image source, IImageVisitorAsync visitor)
-            => source.AcceptAsync(visitor);
+        public static async Task AcceptVisitorAsync(this Image source, IImageVisitorAsync visitor)
+            => await source.AcceptAsync(visitor).ConfigureAwait(false);
```
tocsoft, line 78: This isn't required... the ConfigureAwait only applies inside this `async Task` scope so if you don't await then it doesn't help..

Basically doing this will create the async state machine with no benefit to the caller.

JimBobSquarePants: What about `SaveAsync`?

JimBobSquarePants: You mean delete `ConfigureAwait(false)`?

tocsoft: unless we 'need' to await (because of being in a wrapped using statements or require processing the result etc) then you should not await a task and pass the Task directly down stream, even out of your public api because it'll not help and only hinder due to allocations etc

So in conclusion revert this change and leave it alone, we also shouldn't even call it in the SaveAsync and should just pass everything along to the user.

tocsoft: This also applies to most of the other places you changed in the commit

JimBobSquarePants: I don't know. According to the general guidance the changes I've added are correct.

>This leads to the general guidance of: if you’re writing general-purpose library code, use ConfigureAwait(false). 

https://devblogs.microsoft.com/dotnet/configureawait-faq/

JimBobSquarePants: The only place I would generally not bother with async/await is the result of `Task.FromResult` to remove the overhead.

tocsoft: that only applies if you calling `await` and in all these places we don't need to call await so we don't need to call `ConfigureAwait(false)`

tocsoft: > This leads to the general guidance of: if you’re writing general-purpose library code, use ConfigureAwait(false). This is why, for example, you’ll see every (or almost every) await in the .NET Core runtime libraries using ConfigureAwait(false) on every await; with a few exceptions, in cases where it doesn’t it’s very likely a bug to be fixed. For example, this PR fixed a missing ConfigureAwait(false) call in HttpClient.

Next sentence

JimBobSquarePants: Ah... right, I'm being slow here. Will revert.

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
JimBobSquarePants: I want to have a look at a read-only version of [this](https://github.com/mgravell/Pipelines.Sockets.Unofficial/compare/mgravell:24482d4...mgravell:6740ea4) less overhead and should support all out target frameworks.

src/ImageSharp/Image.FromStream.cs
```diff
@@ -577,8 +576,8 @@ public static async Task<FormattedImage<TPixel>> LoadWithFormatAsync<TPixel>(Con
                 await WithSeekableStreamAsync(
                     configuration,
                     stream,
-                    async s => await DecodeAsync<TPixel>(s, configuration).ConfigureAwait(false))
-                .ConfigureAwait(false);
+                    async s => await DecodeAsync<TPixel>(s, configuration)
```
tocsoft, line 579: this one still doesn't needs awaiting

JimBobSquarePants: Missed that

src/ImageSharp/Image.FromStream.cs
```diff
@@ -99,6 +203,19 @@ public static IImageInfo Identify(Configuration configuration, Stream stream, ou
...
+        /// Decode a new instance of the <see cref="Image"/> class from the given stream.
+        /// The pixel format is selected by the decoder.
+        /// </summary>
+        /// <param name="stream">The stream containing image information.</param>
+        /// <exception cref="ArgumentNullException">The stream is null.</exception>
+        /// <exception cref="NotSupportedException">The stream is not readable.</exception>
+        /// <exception cref="UnknownImageFormatException">Image format not recognised.</exception>
+        /// <exception cref="InvalidImageContentException">Image contains invalid content.</exception>
+        /// <returns>A <see cref="Task{FormattedImage}"/> representing the asynchronous operation.</returns>
+        public static Task<FormattedImage> LoadWithFormatAsync(Stream stream)
```
antonfirsov, line 216: Wondering whether it would be better to move this family of methods under `FormattedImage` and `FormattedImageInfo`. Cleaner naming, but bad for discoverability.

src/ImageSharp/FormattedImage.cs
```diff
@@ -0,0 +1,95 @@
...
+using System;
+using System.Collections.Generic;
+using SixLabors.ImageSharp.Formats;
+
+namespace SixLabors.ImageSharp
+{
+    /// <summary>
+    /// Struct to curry <see cref="Image"/> and <see cref="IImageFormat"/> for return from async overloads.
+    /// </summary>
+    public readonly struct FormattedImage : IEquatable<FormattedImage>
```
antonfirsov, line 13: What is the benefit for defining this type as a `struct` instead of a `class`? (Performance doesn't matter here.)

Is it future proof in it's current form? We may want to extend it to contain more more stuff.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -218,8 +219,32 @@ public static JpegFileMarker FindNextFileMarker(byte[] marker, DoubleBufferedStr
...
+        public async Task<Image<TPixel>> DecodeAsync<TPixel>(Stream stream)
             where TPixel : unmanaged, IPixel<TPixel>
+        {
+            if (stream.CanSeek)
+            {
+                return this.Decode<TPixel>(stream);
+            }
+            else
+            {
+                using (var ms = new MemoryStream())
```
antonfirsov, line 231: In the future we may consider using our own variant of [RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream) to reduce allocation pressure.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -218,8 +219,32 @@ public static JpegFileMarker FindNextFileMarker(byte[] marker, DoubleBufferedStr
         /// <typeparam name="TPixel">The pixel format.</typeparam>
         /// <param name="stream">The stream, where the image should be.</param>
         /// <returns>The decoded image.</returns>
-        public Image<TPixel> Decode<TPixel>(Stream stream)
+        public async Task<Image<TPixel>> DecodeAsync<TPixel>(Stream stream)
             where TPixel : unmanaged, IPixel<TPixel>
+        {
+            if (stream.CanSeek)
+            {
+                return this.Decode<TPixel>(stream);
```
antonfirsov, line 227: I'm not sure whether this execution path is [Stephen Toub proof](https://devblogs.microsoft.com/pfxteam/should-i-expose-asynchronous-wrappers-for-synchronous-methods/), however I can live with it :smile:

I think we got better in total.

src/ImageSharp/Formats/Bmp/BmpDecoder.cs
```diff
@@ -27,6 +28,26 @@ public sealed class BmpDecoder : IImageDecoder, IBmpDecoderOptions, IImageInfoDe
...
+            try
+            {
+                return await decoder.DecodeAsync<TPixel>(stream).ConfigureAwait(false);
+            }
+            catch (InvalidMemoryOperationException ex)
+            {
+                Size dims = decoder.Dimensions;
+
+                throw new InvalidImageContentException($"Can not decode image. Failed to allocate buffers for possibly degenerate dimensions: {dims.Width}x{dims.Height}. This error can happen for very large RLE bitmaps, which are not supported.", ex);
+            }
```
antonfirsov, line 48: We can DRY away these code duplications with the help of an `IImageDecoderCore` interface if we want. (Same for encoders.)

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
antonfirsov: Haven't noticed the ongoing discussion and made a comment about this as well.
`RecyclableMemoryStream` seems more sophisticated at first glance. 

Whichever we choose, I think it would make sense to integrate our adapted solution with `MemoryAllocator`.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
JimBobSquarePants: We don't even need one tbh. We know the length of the input stream, tightly control the lifecycle of the `MemoryStream`. We could simply rend an array from the shared pool and pass to the `MemoryStream` constructor.

JimBobSquarePants:
I need to pick this up as it's likely beneficial to #1205

src/ImageSharp/FormattedImage.cs
JimBobSquarePants: I'm not happy with the name just now... Needs some thought

src/ImageSharp/Image.FromStream.cs
JimBobSquarePants: Structurally cleaner yeah, but I think we'll end up confusing people if we split out the async API. I'd rather avoid the questions.

src/ImageSharp/FormattedImage.cs
JimBobSquarePants: I've been putting some thought to this; I really don't want it to be an extensible object.

My reasoning is that we are merely replicating the synchronous API here.

I'm still not happy with the name `Formatted` suggests the image data is encoded. Do we even need our own types? We could use a named Tuple return type?

``` c#
public static Task<(Image image, IImageFormat Format)> LoadWithFormatAsync (...
```

src/ImageSharp/Formats/Bmp/BmpDecoder.cs
JimBobSquarePants: Yeah, I like this. Can come later though since everything is internal (unless someone is desperately keen to do it).

src/ImageSharp/FormattedImage.cs
JimBobSquarePants: I think this also means we can keep the existing synchronous `out IImageFormat` APIs. It becomes more obvious with a tuple we are working around the syntatic limitations of the language.

JimBobSquarePants: Ah we can't do that can we! We need the implicit operators! 

Back to the drawing board. We need a better name.

JimBobSquarePants: Named Tuples require a minimum language versions though (I'm assuming @tocsoft that's why you went the explicit struct route). Do we want to enforce that? 

Dammit this is tricky.

tocsoft: To be honest I totally didn't think about named tuples having a min version. 

The reason I made an explicit type was because I was working on thought that you shouldn't return Tuples from public APIs, but on this sort of return site (working around language constructs) in hind sight I'm now less bothered in this instance. Even for older language versions just the types of the 2 properties are unique enough to distinguish what the items are.

JimBobSquarePants: Thanks! I think I’m gonna push onwards with them then. The language constraints are low enough and they just feel right as an alternative to the sync API.

antonfirsov: If it's not an extensibility point, anonymous tuple looks better.

PR #1229, opened 2020-06-15
Update Adler32 to correctly filter intrinsics. 

### LINKED ISSUE #1228: PNG encoder throws an exception on x64
Seems there's a lake on supporting this on 64 bit, for more details, please refer to https://github.com/OrchardCMS/OrchardCore/issues/6414

JimBobSquarePants:
### Prerequisites

### Description
Introduces the correct intrinsics support check in the enhanced Adler32 implementation. Fix #1228

@antonfirsov if you have any idea how to test this please let me know if you have any thoughts.

### DIFF
--- a/src/ImageSharp/Formats/Png/Zlib/Adler32.cs
+++ b/src/ImageSharp/Formats/Png/Zlib/Adler32.cs
@@ -63,7 +63,7 @@ namespace SixLabors.ImageSharp.Formats.Png.Zlib
             }
 
 #if SUPPORTS_RUNTIME_INTRINSICS
-            if (Sse3.IsSupported && buffer.Length >= MinBufferSize)
+            if (Ssse3.IsSupported && buffer.Length >= MinBufferSize)
             {
                 return CalculateSse(adler, buffer);
             }
--- a/tests/ImageSharp.Tests/TestUtilities/TestEnvironment.Features.cs
+++ b/tests/ImageSharp.Tests/TestUtilities/TestEnvironment.Features.cs
@@ -10,6 +10,27 @@ namespace SixLabors.ImageSharp.Tests
             public const string On = "1";
             public const string Off = "0";
 
+            // See https://github.com/SixLabors/ImageSharp/pull/1229#discussion_r440477861
+            // * EnableHWIntrinsic
+            //   * EnableSSE
+            //     * EnableSSE2
+            //       * EnableAES
+            //       * EnablePCLMULQDQ
+            //       * EnableSSE3
+            //         * EnableSSSE3
+            //           * EnableSSE41
+            //             * EnableSSE42
+            //               * EnablePOPCNT
+            //               * EnableAVX
+            //                 * EnableFMA
+            //                 * EnableAVX2
+            //   * EnableBMI1
+            //   * EnableBMI2
+            //   * EnableLZCNT
+            //
+            // `FeatureSIMD` ends up impacting all SIMD support(including `System.Numerics`) but not things
+            // like `LZCNT`, `BMI1`, or `BMI2`
+            // `EnableSSE3_4` is a legacy switch that exists for compat and is basically the same as `EnableSSE3`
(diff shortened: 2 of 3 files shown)


antonfirsov:
We can disable runtime features with environment variables , and run some Adler32 tests in external process. (`processStartInfo.Environment["COMPlus_EnableSSSE3"] = "0"`  + feed `processStartInfo` to `RemoteExecutor`)
A list of variables that might be interesting for our code:
https://github.com/dotnet/runtime/blob/f987b3bc8f8a4d49ca1fc8821a48a7614dfee7cd/src/coreclr/tests/testenvironment.proj#L15-L33

I would also add `Assert.False(Ssse3.IsSupported)` to the test, to make sure the JIT knob works.

tannergooding:
That is a reliable way to test the software fallback path and is what we do in the runtime (it's also why we exposed these knobs).

src/ImageSharp/Formats/Png/Zlib/Adler32.cs
```diff
@@ -63,7 +63,7 @@ public static uint Calculate(uint adler, ReadOnlySpan<byte> buffer)
             }
 
 #if SUPPORTS_RUNTIME_INTRINSICS
-            if (Sse3.IsSupported && buffer.Length >= MinBufferSize)
+            if (Ssse3.IsSupported && buffer.Length >= MinBufferSize)
             {
                 return CalculateSse(adler, buffer);
```
dlemstra, line 68: Should we also rename `CalculateSse` to `CalculateSsse`?

JimBobSquarePants: I would say no. It’s still an iteration of SSE technology.

JimBobSquarePants:
@antonfirsov @tannergooding Test added now. Think I've got it right.

tests/ImageSharp.Tests/TestUtilities/TestEnvironment.Features.cs
```diff
@@ -0,0 +1,34 @@
...
+            public const string Off = "0";
+
+            public const string EnableAES = "COMPlus_EnableAES";
+            public const string EnableAVX = "COMPlus_EnableAVX";
+            public const string EnableAVX2 = "COMPlus_EnableAVX2";
+            public const string EnableBMI1 = "COMPlus_EnableBMI1";
+            public const string EnableBMI2 = "COMPlus_EnableBMI2";
+            public const string EnableFMA = "COMPlus_EnableFMA";
+            public const string EnableHWIntrinsic = "COMPlus_EnableHWIntrinsic";
+            public const string EnableIncompleteISAClass = "COMPlus_EnableIncompleteISAClass";
```
tannergooding, line 20: This one isn't a production switch.

For the others, there is a rough hierarchy they follow:
* EnableHWIntrinsic
  * EnableSSE
    * EnableSSE2
      * EnableAES
      * EnablePCLMULQDQ
      * EnableSSE3
        * EnableSSSE3
          * EnableSSE41
            * EnableSSE42
              * EnablePOPCNT
              * EnableAVX
                * EnableFMA
                * EnableAVX2
  * EnableBMI1
  * EnableBMI2
  * EnableLZCNT

`FeatureSIMD` ends up impacting all SIMD support (including `System.Numerics`) but not things like `LZCNT`, `BMI1`, or `BMI2`
`EnableSSE3_4` is a legacy switch that exists for compat and is basically the same as `EnableSSE3`

tannergooding: The hierarchy basically means disabling SSE (`EnableSSE=0`) will also disable everything below it

JimBobSquarePants: Ah lovely. I'll reference that. I intend on doing more intrinsics work in the near future.

PR #1236, opened 2020-06-17
Added Third Party Notices document

### LINKED ISSUE #856: Consider adding a THIRD-PARTY-NOTICES.TXT file
ImageSharp should consider adding a `THIRD-PARTY-NOTICES.TXT` file to contain notices for 3rd party code that is referenced or ported into ImageSharp.

This would be especially useful when code is copied from CoreFX to provide certain functionality downlevel on full framework.

Turnerj:
### Prerequisites

### Description

Adds a Third Party Notices file, closes #856 . Opening language copied from dotnet/runtime, license information copied from the various sources in the code base.

One thing to note, the shared infrastructure repo also has a file with MIT and BSD licensing. I didn't include that reference here as I see the shared infrastructure to be its own dependency. It probably deserves to have its own Third Party Notices file.

### DIFF
--- a/THIRD-PARTY-NOTICES.TXT
+++ b/THIRD-PARTY-NOTICES.TXT
@@ -1,0 +1,32 @@
+ImageSharp uses third-party libraries or other resources that may be
+distributed under licenses different than ImageSharp itself.
+
+In the event that we accidentally failed to list a required notice, please
+bring it to our attention by posting an issue.
+
+The attached notices are provided for information only.
+
+
+License notice for Zlib
+-----
+
+DeflateStream implementation adapted from SharpZipLib.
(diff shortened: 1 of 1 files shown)


Turnerj:
Also not sure if it is then worth then keeping the same license references in the code (like the CoreFX Stream extensions) or a sub-readme (like Zlib). You could just refer them to this Third Party Notices file instead.

JimBobSquarePants:
>Also not sure if it is then worth then keeping the same license references in the code (like the CoreFX Stream extensions) or a sub-readme (like Zlib). You could just refer them to this Third Party Notices file instead.

Let's keep them as-is for the time being. This is great!

PR #1260, opened 2020-07-06
Async version of DiscoverDecoder

pekspro:
### Prerequisites

### Description
It was no async version of DiscoverDecoder, so I added DiscoverDecoderAsync. Since async methods can’t have out parameters, I return a tuple instead. I would not mind doing the same change on DiscoverDecoder so they look the same.


JimBobSquarePants:
Thanks for this @pekspro I think simply transitioning the code internally as you have done is enough coverage for now.

> I would not mind doing the same change on DiscoverDecoder so they look the same.

We've an established pattern using `out` parameters on the synchronous versions of our APIs that I do not want to change.

PR #1262, opened 2020-07-07
Update all dependency references.

JimBobSquarePants:
### Prerequisites

### Description
Just a simple update to all reference dependencies. Might help with any .NET FX type forwarding issues

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -29,7 +29,7 @@
     <PackageReference Update="System.IO.Compression" Version="4.3.0" />
     <PackageReference Update="System.IO.UnmanagedMemoryStream" Version="4.3.0" />
     <PackageReference Update="System.Numerics.Vectors" Version="4.5.0" />
-    <PackageReference Update="System.Memory" Version="4.5.4" />
+    <PackageReference Update="System.Memory" Version="4.5.3" />
     <PackageReference Update="System.Runtime.CompilerServices.Unsafe" Version="4.7" />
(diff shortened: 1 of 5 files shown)


JimBobSquarePants:
No idea what or why this is happening. Can get the tests to run locally without crashing either because ImageMagick decoder always crashes 

https://github.com/SixLabors/ImageSharp/runs/845593121?check_suite_focus=true#step:7:97

JimBobSquarePants:
@antonfirsov I think we need to reevaluate our usage of `RemoteExecutor` to workaround some of the flakiness.

https://github.com/SixLabors/ImageSharp/runs/859311527?check_suite_focus=true#step:7:195

It does look like there's a way with Xunit to run code once at startup which would mean we could move it out of static contructors.
https://stackoverflow.com/a/53143426/427899

antonfirsov:
@JimBobSquarePants the last test failure I see is because of Magic.NET. What if we revert updating that, and address it in a separate PR?

antonfirsov:
Since it's a test-only change, I don't see a point to keep the 1.0 milestone.

JimBobSquarePants:
I was going to update the .NET references also. I'll have one last crack in a mo but if that fails I'll scrap it.

JimBobSquarePants:
@antonfirsov Turns out `System.Memory` is the trigger. I've updated all other .NET dependencies.

JimBobSquarePants:
I'm just gonna go ahead and merge this now since there are no code changes.

PR #1269, opened 2020-07-10
Buffered Decoding.

### LINKED ISSUE #1268: Decoding failes on slow streams
### Prerequisites

### Description
As it is mentioned on the documentation for [Stream.Read](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=netcore-3.1#System_IO_Stream_Read_System_Byte___System_Int32_System_Int32_)

_An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached._

But if you have a stream that is on very bad mood and for instance just reads one byte at the time reading will throw an `SixLabors.ImageSharp.UnknownImageFormatException`. If  you let some reads work as normal, and then starts to read one byte at the time other exceptions will be thrown or just a part of the image till be loaded (at least for PNG-images, have not tried other formats).

My gut feeling is that is something like this that causes the problem in #926, but I haven't been able to replicate exactly that.

### Steps to Reproduce
This code tries to read images from a folder with one byte read at the time:

    using SixLabors.ImageSharp;
    using System;
    using System.IO;

    namespace EvilStreamConsoleApp
    {
        class Program
        {
            {
                Console.WriteLine("Hello World!");

                Directory.CreateDirectory(".\\outputimages");
                var files = Directory.EnumerateFiles(".\\inputimages", "*.*");

                foreach(var file in files)
                {
                    using EvilStream evilStream = new EvilStream(file, FileMode.Open);

                    var image = Image.Load(evilStream);

                    string outputFileName = ".\\outputimages\\" + Path.GetFileName(file);
                    image.Save(outputFileName);
                }
            }
        }

        class EvilStream : FileStream
        {
            public EvilStream(string path, FileMode mode)
                : base(path, mode)
            {

            }

            int beNiceCount = 0;

            public override int Read(byte[] array, int offset, int count)
            {
                if (beNiceCount > 0)
                {
                    beNiceCount--;

                    return base.Read(array, offset, count);
                }

                return base.Read(array, offset, 1);

                //return base.Read(array, offset, Math.Min(256 * 20 - 1, count));
            }
        }
    }

- ImageSharp version: 1.0.0-rc0003
- .NET Framework version: .NET Core 3.1

JimBobSquarePants:
### Prerequisites

### Description
I've been tinkering about with this for a while...

We've used buffered reading via a stream wrapper in our JpegDecoder to reduce the overhead of single byte reads. You end up with a nice performance boost over native reading.
https://github.com/SixLabors/ImageSharp/blob/c0594f7031e76ac862ea5f077d707df76ed6560b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs#L246

However, the shape of the class doesn't allow for transparent reuse with other decoders which is a must. The mileage this  buffering gives for each encoder varies based upon individual implementation but the net benefit is worth the investment. 

With that in mind I refactored the class to turn it into an actual stream that internally wraps our input stream and is transparent to any decoder implementation. A great benefit to this is that we can fix our buffered stream to correctly handle when a stream returns less than expected on `Read(buffer, offset, count)` Fixes #1268 .

The following  benchmarks represent our new new implementation `BufferedReadStream` vs the old jpeg wrapper `BufferedReadStreamWrap`, reading a `MemoryStream` preloaded with a buffer, and reading an array directly.

Things to note:

We're a lot faster on .NET Core + than the native stream reader and ~match the wrapping implementation on those frameworks. 

The performance loss against the wrapper implementation in .NETFX is due to the jitter being unable to inline the appropriate methods on a sealed, inheriting class, I've experimented with several implementations and there seems to be no way around it. Since we are testing against the fastest possible stream reading scenario and we are measuring in `us` I'm not concerned. Our decoder benchmarks on NET FX are unaffected.

|                         Method |       Runtime |      Mean |      Error |    StdDev | Ratio | RatioSD | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------------- |-------------- |----------:|-----------:|----------:|------:|--------:|------:|------:|------:|----------:|
|             StandardStreamRead |    .NET 4.7.2 | 126.07 us | 126.498 us |  6.934 us |  0.99 |    0.08 |     - |     - |     - |         - |
|         BufferedReadStreamRead |    .NET 4.7.2 | 118.08 us |  42.234 us |  2.315 us |  0.92 |    0.03 |     - |     - |     - |         - |
|     BufferedReadStreamWrapRead |    .NET 4.7.2 |  45.33 us |  22.833 us |  1.252 us |  0.35 |    0.00 |     - |     - |     - |         - |
|         StandardStreamReadByte |    .NET 4.7.2 | 128.17 us |  94.616 us |  5.186 us |  1.00 |    0.00 |     - |     - |     - |         - |
|     BufferedReadStreamReadByte |    .NET 4.7.2 | 143.60 us |  92.871 us |  5.091 us |  1.12 |    0.08 |     - |     - |     - |         - |
| BufferedReadStreamWrapReadByte |    .NET 4.7.2 |  32.72 us |  53.708 us |  2.944 us |  0.26 |    0.02 |     - |     - |     - |         - |
|                  ArrayReadByte |    .NET 4.7.2 |  19.40 us |  12.206 us |  0.669 us |  0.15 |    0.01 |     - |     - |     - |         - |
|                                |               |           |            |           |       |         |       |       |       |           |
|             StandardStreamRead | .NET Core 2.1 |  84.82 us |  55.983 us |  3.069 us |  0.75 |    0.15 |     - |     - |     - |         - |
|         BufferedReadStreamRead | .NET Core 2.1 |  49.62 us |  27.253 us |  1.494 us |  0.44 |    0.08 |     - |     - |     - |         - |
|     BufferedReadStreamWrapRead | .NET Core 2.1 |  67.78 us |  87.546 us |  4.799 us |  0.60 |    0.10 |     - |     - |     - |         - |
|         StandardStreamReadByte | .NET Core 2.1 | 115.81 us | 382.107 us | 20.945 us |  1.00 |    0.00 |     - |     - |     - |         - |
|     BufferedReadStreamReadByte | .NET Core 2.1 |  16.32 us |   6.123 us |  0.336 us |  0.14 |    0.02 |     - |     - |     - |         - |
| BufferedReadStreamWrapReadByte | .NET Core 2.1 |  16.68 us |   4.616 us |  0.253 us |  0.15 |    0.03 |     - |     - |     - |         - |
|                  ArrayReadByte | .NET Core 2.1 |  15.13 us |  60.763 us |  3.331 us |  0.14 |    0.05 |     - |     - |     - |         - |
|                                |               |           |            |           |       |         |       |       |       |           |
|             StandardStreamRead | .NET Core 3.1 |  92.44 us |  88.217 us |  4.835 us |  0.94 |    0.06 |     - |     - |     - |         - |
|         BufferedReadStreamRead | .NET Core 3.1 |  36.41 us |   5.923 us |  0.325 us |  0.37 |    0.00 |     - |     - |     - |         - |
|     BufferedReadStreamWrapRead | .NET Core 3.1 |  37.22 us |   9.628 us |  0.528 us |  0.38 |    0.01 |     - |     - |     - |         - |
|         StandardStreamReadByte | .NET Core 3.1 |  98.67 us |  20.947 us |  1.148 us |  1.00 |    0.00 |     - |     - |     - |         - |
|     BufferedReadStreamReadByte | .NET Core 3.1 |  41.36 us | 123.536 us |  6.771 us |  0.42 |    0.06 |     - |     - |     - |         - |
| BufferedReadStreamWrapReadByte | .NET Core 3.1 |  39.11 us |  93.894 us |  5.147 us |  0.40 |    0.05 |     - |     - |     - |         - |
|                  ArrayReadByte | .NET Core 3.1 |  18.84 us |   4.684 us |  0.257 us |  0.19 |    0.00 |     - |     - |     - |         - |


tests/ImageSharp.Tests/Image/ImageTests.Load_FileSystemPath_PassLocalConfiguration.cs
```diff
@@ -36,7 +36,7 @@ public void Configuration_Path_Agnostic()
                 this.TestFormat.VerifyAgnosticDecodeCall(this.Marker, this.TopLevelConfiguration);
             }
 
-            [Fact]
+            [Fact(Skip = "TODO: Enable when someone tells me how this mocking stuff works.")]
```
JimBobSquarePants, line 39: @tocsoft I really struggled to understand how this worked.

tocsoft: Currently the test verifies that the mock `IImageDecoder`s instances `Decode<Rgba32>()` method is called with exactly 2 properties that have reference equality (I believe it reference equality anyway) to `this.TopLevelConfiguration` and `this.DataStream` (which will have been returned from a mock/fake file system configured in `TopLevelConfiguration` when its called with `MockFilePath`).

 For your new code which now wraps all the streams with a `BufferedReadStream` you will need to change the `this.DataStream` parameter to something like `It.Is<Stream>(x=>(BufferedReadStream)x.BaseStream == this.DataStream)` which allows the 'verify' to swap in other sorts of checks. (would require the base stream to be exposed from the buffered read stream)

JimBobSquarePants: Excellent. That fixed it!

JimBobSquarePants:
@SixLabors/core Any objections to this? I'm getting a twitchy merge finger.

src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanScanBuffer.cs
```diff
@@ -11,7 +11,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
     /// </summary>
     internal struct HuffmanScanBuffer
     {
-        private readonly DoubleBufferedStreamReader stream;
+        private readonly Stream stream;
```
antonfirsov, line 14: We probably loose a some part of the benchmarked performance gain here, since the type stored here is the base class `Stream`, therefore the methods being called on it are not inlined, unlike in benchmarks.

JimBobSquarePants: I didn’t actually see any slowdown at all in jpeg but I’ll run/publish tests tomorrow

JimBobSquarePants: @antonfirsov  Before and after benchmarks. Short run so plenty of error but nothing that stands out. 

|                         Method |            TestImage |       Mean |       Error |    StdDev | Ratio | RatioSD | Gen 0 | Gen 1 | Gen 2 |  Allocated |
|------------------------------- |--------------------- |-----------:|------------:|----------:|------:|--------:|------:|------:|------:|-----------:|
| 'Decode Jpeg - System.Drawing' | Jpg/b(...)e.jpg [21] |   5.123 ms |   7.2206 ms | 0.3958 ms |  1.00 |    0.00 |     - |     - |     - |      176 B |
|     'Decode Jpeg - ImageSharp' | Jpg/b(...)e.jpg [21] |  12.219 ms |  15.6786 ms | 0.8594 ms |  2.39 |    0.02 |     - |     - |     - |    15904 B |
|                                |                      |            |             |           |       |         |       |       |       |            |
| 'Decode Jpeg - System.Drawing' | Jpg/b(...)f.jpg [28] |  13.339 ms |   0.5726 ms | 0.0314 ms |  1.00 |    0.00 |     - |     - |     - |      176 B |
|     'Decode Jpeg - ImageSharp' | Jpg/b(...)f.jpg [28] |  29.084 ms |   6.2870 ms | 0.3446 ms |  2.18 |    0.02 |     - |     - |     - |    16880 B |
|                                |                      |            |             |           |       |         |       |       |       |            |
| 'Decode Jpeg - System.Drawing' | Jpg/i(...)e.jpg [43] | 329.528 ms |  56.3769 ms | 3.0902 ms |  1.00 |    0.00 |     - |     - |     - |      216 B |
|     'Decode Jpeg - ImageSharp' | Jpg/i(...)e.jpg [43] | 268.073 ms | 179.0279 ms | 9.8131 ms |  0.81 |    0.04 |     - |     - |     - | 36022424 B |


|                         Method |            TestImage |       Mean |       Error |     StdDev | Ratio | RatioSD | Gen 0 | Gen 1 | Gen 2 |  Allocated |
|------------------------------- |--------------------- |-----------:|------------:|-----------:|------:|--------:|------:|------:|------:|-----------:|
| 'Decode Jpeg - System.Drawing' | Jpg/b(...)e.jpg [21] |   4.965 ms |   0.6894 ms |  0.0378 ms |  1.00 |    0.00 |     - |     - |     - |      176 B |
|     'Decode Jpeg - ImageSharp' | Jpg/b(...)e.jpg [21] |  11.881 ms |   0.3953 ms |  0.0217 ms |  2.39 |    0.02 |     - |     - |     - |    15808 B |
|                                |                      |            |             |            |       |         |       |       |       |            |
| 'Decode Jpeg - System.Drawing' | Jpg/b(...)f.jpg [28] |  13.548 ms |   1.5905 ms |  0.0872 ms |  1.00 |    0.00 |     - |     - |     - |      176 B |
|     'Decode Jpeg - ImageSharp' | Jpg/b(...)f.jpg [28] |  28.147 ms |   1.6038 ms |  0.0879 ms |  2.08 |    0.01 |     - |     - |     - |    16816 B |
|                                |                      |            |             |            |       |         |       |       |       |            |
| 'Decode Jpeg - System.Drawing' | Jpg/i(...)e.jpg [43] | 332.870 ms |  22.0796 ms |  1.2103 ms |  1.00 |    0.00 |     - |     - |     - |      176 B |
|     'Decode Jpeg - ImageSharp' | Jpg/i(...)e.jpg [43] | 286.208 ms | 732.9871 ms | 40.1775 ms |  0.86 |    0.12 |     - |     - |     - | 36022360 B |

antonfirsov: It's ~5% change for the largest image, which is suspicious. These are no microbenchmarks, so the result of a short run is relevant, unless there is other noise, which can ofc happen with local runs.

We have `DecodeJpegParseStreamOnly` as a relatively quick option reassess this.

JimBobSquarePants: Before

|                      Method |        Job |       Runtime |            TestImage |     Mean |     Error |   StdDev | Ratio | RatioSD |    Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------------- |----------- |-------------- |--------------------- |---------:|----------:|---------:|------:|--------:|---------:|------:|------:|----------:|
|       'System.Drawing FULL' | Job-COEKJJ |    .NET 4.7.2 | Jpg/b(...)f.jpg [28] | 16.09 ms |  0.024 ms | 0.001 ms |  1.00 |    0.00 | 156.2500 |     - |     - |  777784 B |
| JpegDecoderCore.ParseStream | Job-COEKJJ |    .NET 4.7.2 | Jpg/b(...)f.jpg [28] | 17.96 ms | 12.218 ms | 0.670 ms |  1.12 |    0.04 |        - |     - |     - |   12544 B |
|                             |            |               |                      |          |           |          |       |         |          |       |       |           |
|       'System.Drawing FULL' | Job-UZXMLV | .NET Core 2.1 | Jpg/b(...)f.jpg [28] | 15.25 ms |  0.973 ms | 0.053 ms |  1.00 |    0.00 | 171.8750 |     - |     - |  775208 B |
| JpegDecoderCore.ParseStream | Job-UZXMLV | .NET Core 2.1 | Jpg/b(...)f.jpg [28] | 12.52 ms |  1.801 ms | 0.099 ms |  0.82 |    0.01 |        - |     - |     - |   12416 B |
|                             |            |               |                      |          |           |          |       |         |          |       |       |           |
|       'System.Drawing FULL' | Job-PIDJAF | .NET Core 3.1 | Jpg/b(...)f.jpg [28] | 14.43 ms |  3.392 ms | 0.186 ms |  1.00 |    0.00 |        - |     - |     - |     191 B |
| JpegDecoderCore.ParseStream | Job-PIDJAF | .NET Core 3.1 | Jpg/b(...)f.jpg [28] | 11.55 ms |  1.149 ms | 0.063 ms |  0.80 |    0.01 |        - |     - |     - |   12408 B |

After

|                      Method |       Runtime |            TestImage |     Mean |    Error |   StdDev | Ratio |    Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------------- |-------------- |--------------------- |---------:|---------:|---------:|------:|---------:|------:|------:|----------:|
|       'System.Drawing FULL' |    .NET 4.7.2 | Jpg/b(...)f.jpg [28] | 13.52 ms | 1.621 ms | 0.089 ms |  1.00 | 171.8750 |     - |     - |  777789 B |
| JpegDecoderCore.ParseStream |    .NET 4.7.2 | Jpg/b(...)f.jpg [28] | 13.01 ms | 1.466 ms | 0.080 ms |  0.96 |        - |     - |     - |   12416 B |
|                             |               |                      |          |          |          |       |
 |       |       |           |
|       'System.Drawing FULL' | .NET Core 2.1 | Jpg/b(...)f.jpg [28] | 13.64 ms | 3.331 ms | 0.183 ms |  1.00 | 171.8750 |     - |     - |  775208 B |
| JpegDecoderCore.ParseStream | .NET Core 2.1 | Jpg/b(...)f.jpg [28] | 11.89 ms | 0.374 ms | 0.021 ms |  0.87 |        - |     - |     - |   12248 B |
|                             |               |                      |          |          |          |       |
 |       |       |           |
|       'System.Drawing FULL' | .NET Core 3.1 | Jpg/b(...)f.jpg [28] | 13.20 ms | 1.116 ms | 0.061 ms |  1.00 |        - |     - |     - |     176 B |
| JpegDecoderCore.ParseStream | .NET Core 3.1 | Jpg/b(...)f.jpg [28] | 12.09 ms | 0.310 ms | 0.017 ms |  0.92 |        - |     - |     - |   12246 B |

JimBobSquarePants: Looks like we've lost something on Core. What do you think, use the concrete implementation? We definitely need something to ensure the reads are complete.

antonfirsov: If single byte reads are really frequent then one benefit of double buffering is that it allows `ReadByte` to be a fast inline  method without virtual calls on hot path.
So using the concrete implementation should definitely help in this case, since it should ideally allow devirtualizing and inlining `ReadByte` the same way it was with `DoubleBufferedStreamReader`.

JimBobSquarePants: That's gonna mean changing `IImageDecoder` to use the concrete type unless I just make the type public and use it individually for each decoder. 

If we do change the signature we should do the equivalent for IImageEncoder also.

antonfirsov: I will think if I can find an another way.

JimBobSquarePants: I think I've got a good way now actually.... I updated the internal interface for the core decoders to use the buffered stream. I'll have something to show very soon.

JimBobSquarePants: @antonfirsov  Less of a difference than I thought. That's with all the core decoder directly referencing the concrete decoder.

|                      Method |        Job |       Runtime |            TestImage |     Mean |    Error |   StdDev | Ratio |    Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------------- |----------- |-------------- |--------------------- |---------:|---------:|---------:|------:|---------:|------:|------:|----------:|
|       'System.Drawing FULL' | Job-RSRDJA |    .NET 4.7.2 | Jpg/b(...)f.jpg [28] | 13.28 ms | 0.976 ms | 0.053 ms |  1.00 | 171.8750 |     - |     - |  777761 B |
| JpegDecoderCore.ParseStream | Job-RSRDJA |    .NET 4.7.2 | Jpg/b(...)f.jpg [28] | 13.54 ms | 0.792 ms | 0.043 ms |  1.02 |        - |     - |     - |   12416 B |
|                             |            |               |                      |          |          |          |       |          |       |       |           |
|       'System.Drawing FULL' | Job-QLXCST | .NET Core 2.1 | Jpg/b(...)f.jpg [28] | 13.24 ms | 2.348 ms | 0.129 ms |  1.00 | 171.8750 |     - |     - |  775208 B |
| JpegDecoderCore.ParseStream | Job-QLXCST | .NET Core 2.1 | Jpg/b(...)f.jpg [28] | 11.72 ms | 1.564 ms | 0.086 ms |  0.89 |        - |     - |     - |   12352 B |
|                             |            |               |                      |          |          |          |       |          |       |       |           |
|       'System.Drawing FULL' | Job-KQKRGS | .NET Core 3.1 | Jpg/b(...)f.jpg [28] | 13.08 ms | 2.922 ms | 0.160 ms |  1.00 |        - |     - |     - |     176 B |
| JpegDecoderCore.ParseStream | Job-KQKRGS | .NET Core 3.1 | Jpg/b(...)f.jpg [28] | 11.49 ms | 0.771 ms | 0.042 ms |  0.88 |        - |     - |     - |   12372 B |

antonfirsov: On .NET Core 3.1:
`11.55` (old) VS  `12.09` (base Stream) VS `11.49` (concrete Stream)

I think it's quite good, or am I missing something?

JimBobSquarePants: I'm looking at the ratios compared to before but error could be creeping in there. It's definitely faster

I'm gonna push with the concrete implementation as it should bring us the best of everything.

JimBobSquarePants: @antonfirsov OK. I think this is ready to review again now.

src/ImageSharp/Formats/Png/PngEncoder.cs
```diff
@@ -72,13 +72,11 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
...
-        public async Task EncodeAsync<TPixel>(Image<TPixel> image, Stream stream)
+        public Task EncodeAsync<TPixel>(Image<TPixel> image, Stream stream)
             where TPixel : unmanaged, IPixel<TPixel>
         {
-            using (var encoder = new PngEncoderCore(image.GetMemoryAllocator(), image.GetConfiguration(), new PngEncoderOptions(this)))
-            {
-                await encoder.EncodeAsync(image, stream).ConfigureAwait(false);
-            }
+            using var encoder = new PngEncoderCore(image.GetMemoryAllocator(), image.GetConfiguration(), new PngEncoderOptions(this));
+            return encoder.EncodeAsync(image, stream);
```
tocsoft, line 79: I'm sure this one needs awaiting otherwise the encoder will be disposed before the `Task` is awaited/activated and things will possibly exploded, but only if we end up having to perform a `CopyToAsync` which is unlikely to be happening the unit test

JimBobSquarePants: I’m cool with switching it back.

src/ImageSharp/Formats/ImageDecoderUtilities.cs
```diff
@@ -14,42 +14,21 @@ internal static class ImageDecoderUtilities
...
-            {
-                return decoder.Decode<TPixel>(stream);
-            }
-
-            using MemoryStream ms = decoder.Configuration.MemoryAllocator.AllocateFixedCapacityMemoryStream(stream.Length);
-            await stream.CopyToAsync(ms).ConfigureAwait(false);
-            ms.Position = 0;
-            return decoder.Decode<TPixel>(ms);
-        }
+            => Task.FromResult(decoder.Decode<TPixel>(stream));
```
antonfirsov, line 32: Was the MemoryStream copying logic an unnecessary duplicate here it's already has been done in `Image.WithSeekableStreamAsync` ?

tests/ImageSharp.Tests/IO/BufferedReadStreamTests.cs
```diff
@@ -0,0 +1,278 @@
...
+            var buffer = new byte[length];
+            var random = new Random();
+            random.NextBytes(buffer);
+
+            return new EvilStream(buffer);
+        }
+
+        // Simulates a stream that can only return 1 byte at a time per read instruction.
+        // See https://github.com/SixLabors/ImageSharp/issues/1268
+        private class EvilStream : MemoryStream
```
antonfirsov, line 265: ❤️

src/ImageSharp/Formats/ImageDecoderUtilities.cs
JimBobSquarePants: Yep, exactly.

src/ImageSharp/Common/Extensions/StreamExtensions.cs
```diff
@@ -51,21 +49,22 @@ public static void Skip(this Stream stream, int count)
...
-                    }
-
-                    count -= bytesRead;
+                    break;
                 }
+
+                count -= bytesRead;
             }
+
+            ArrayPool<byte>.Shared.Return(buffer);
```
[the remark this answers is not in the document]
JimBobSquarePants: Sweet, will do.

src/ImageSharp/Formats/ImageDecoderUtilities.cs
```diff
@@ -14,42 +14,21 @@ internal static class ImageDecoderUtilities
...
-            using MemoryStream ms = decoder.Configuration.MemoryAllocator.AllocateFixedCapacityMemoryStream(stream.Length);
-            await stream.CopyToAsync(ms).ConfigureAwait(false);
-            ms.Position = 0;
-            return decoder.Identify(ms);
-        }
+        /// <param name="stream">The <see cref="BufferedReadStream"/> containing image data.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="stream"/> is null.</exception>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        public static Task<IImageInfo> IdentifyAsync(this IImageDecoderInternals decoder, BufferedReadStream stream)
+            => Task.FromResult(decoder.Identify(stream));
```
antonfirsov, line 21: In case of error we should return `Task.FromException` instead of failing to create the `Task`.

The fix is on the way, as I'm working on a PR for #1280, so don't bother with it.

src/ImageSharp/Common/Extensions/StreamExtensions.cs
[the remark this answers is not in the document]
JimBobSquarePants: Done

src/ImageSharp/Formats/ImageDecoderUtilities.cs
JimBobSquarePants: Sweet. Will wait to see what you come up with then. 👍

PR #1271, opened 2020-07-12
Webp support

brianpopow:
### Prerequisites

### Description

This is a port of libweb for decoding and encoding webp images with ImageSharp.
It is based on [libwebp](https://chromium.googlesource.com/webm/libwebp)

The specification for the lossless webp format can be found here [lossless](https://developers.google.com/speed/webp/docs/webp_lossless_bitstream_specification)

The specification for the lossy format can be found here [lossy](https://tools.ietf.org/html/rfc6386)


stevetemple:
I was hoping you'd be able to share your encoder branch, I'd like to help out if possible

brianpopow:
> I was hoping you'd be able to share your encoder branch, I'd like to help out if possible @brianpopow

@stevetemple im happy to hear that you want to help, i really can need some helping hands with this. 

I have just pushed my changes of the lossless encoding to this branch (still work in progress). I have not yet started to work on lossy encoding. Maybe you can give this a try while i continue with the TODO's in the lossless encoding? Maybe you could start with implementing the VP8BitWriter. I have tried to follow the reference [implementation](https://chromium.googlesource.com/webm/libwebp) as close as possible, to be able to compare the results with it. The corresponding file in the reference Implementation for the bitwriter is `bit_writer_utils.c`. 

Let me know if you have other ideas how to help.

brianpopow:
Good news: I was able to encode the first lossless image!

This input image was:
lena

Unfortunately github does not support webp, so i cannot attach the result.

Still a few things are hardcoded and need to be done, like the quality settings and others, but im happy that it works so far.

stevetemple:
> Good news: I was able to encode the first lossless image!

Awesome was going to try and pick up some more of that tonight, is there anywhere else I can focus on I don't want to tread on toes if you're already looking at the rest of the lossless encoding

brianpopow:
> > Good news: I was able to encode the first lossless image!
> 
> Awesome was going to try and pick up some more of that tonight, is there anywhere else I can focus on?
> 
> I don't want to tread on toes if you're already looking at the rest of the lossless encoding

I have not yet decided how to continue at the moment, so feel free to pick what ever you want. One big construction site is still the `EncoderAnalyze` method. Its where it should be decided which transformations should be used fore the encoding. This is all hard coded at the moment. The method is also called like this in the reference implementation. So i would recommend this, but as i said feel free to pick whatever you want. 

One note if you pick up this: I have decided to not implement the `low_effort` option with intent from the reference decoder. Im not sure if we should do this. Maybe something for the future.

brianpopow:
I was able to fix some issues lately and lossless encoding works better now. There is still a bug with some images, for example: [Car](https://github.com/SixLabors/ImageSharp/blob/master/tests/Images/Input/Bmp/Car.bmp).
Encoding fails with an index out of range exception when writing the Huffman Code. Really hard to find the (hopefully) last issue.

JimBobSquarePants:
If you point me at where you think the issue is an the equivalent source I can take a look for you

brianpopow:
> If you point me at where you think the issue is an the equivalent source I can take a look for you

A second opinion would be really helpful, i think i already looked to much time on the same code
and can't see the wood for the trees.
The problem is, that i have no good idea why this is happening. I spent a good amount of time the 
last two days to find the cause, but without any good progress in that regard.

While the error occurs in `WriteHuffmanCodeWithExtraBits` (its the same method name in the original code),
because the `codeIndex` exceeds the max of 280, i guess its just a follow up error.

There is already a difference in the bitstream compared to the original implementation much earlier.
Only the first 4 bytes are equal and then it differs. It happens in `ApplyPredictFilter` from `Vp8Encoder.cs`
and there in `StoreHuffmanTreeOfHuffmanTreeToBitMask`.

btw. i have kept the original method names usually, except that some had prefixes like `VP8`.
For example `VP8LGetBackwardReferences` which i have changed to just `GetBackwardReferences`

JimBobSquarePants:
@brianpopow I created my own branch to hack against but managed to break the inverse predictor.

https://github.com/SixLabors/ImageSharp/commit/671443f88066ea5ed58a4bdf830c888ba6dd333b

Something I spotted immediately in the encoder is a few areas in your predictor where you are using `i++` instead of `++i`. 

I found the `Span<T>` based API incredibly difficult to follow though so attempted to migrate the code to using fixed pointers to make the methods read much more similarly to the original C source.

However, since the `PredictorX` methods are used in both the decoder and encoder I found myself having to refactor the inverse predictor where I seem to have messed up. I can't spot the issue since the flow within the methods looks identical to me.

I would rather be using fixed pointers over span to avoid all the bounds checks in each `PredictorX` method. We won't get decent performance otherwise.

brianpopow:
@JimBobSquarePants 

> Something I spotted immediately in the encoder is a few areas in your predictor where you are using `i++` instead of `++i`.
> 

If that's outside of just an increment in a for loop, that might be an issue.

> I found the `Span<T>` based API incredibly difficult to follow though so have migrated the code to using fixed pointers to make the methods read much more similarly to the original C source.
> 
> I would rather be using fixed pointers over span to avoid all the bounds checks in each `PredictorX` method. We won't get decent performance otherwise.
> 

Yes i agree that's better. Feel free to merge that into my webp branch.

> My branch doesn't throw in `WriteHuffmanCodeWithExtraBits` but the output still isn't 100% valid (some of the test files seem ok) I'll wait for you to come back before doing much more but I think we'll be on a better path to a working lossless encoder since there are a lot less index tracking variables to manage now.

I still get that index out of bound error in `WriteHuffmanCodeWithExtraBits` with your branch when i test encoding with the Car.bmp file.

JimBobSquarePants:
@brianpopow 

>I still get that index out of bound error in WriteHuffmanCodeWithExtraBits with your branch when i test encoding with the Car.bmp file.

Weird...

I see no errors thrown in the `ImageCanConvertFormat` test. The output folder looks like this.

image

Explorer seems to be incredibly robust though as the file is not valid.

image

I'll throw another couple of hours at it this evening and see if I can make any progress.

brianpopow:
> I'll throw another couple of hours at it this evening and see if I can make any progress.
> 
> UPDATE
> Car.bmp works now in my branch. It's actually the only one that Paint.NET can also open.
> 
> UPDATE2
> Car is broken again but Calliphora now works 😭

@JimBobSquarePants thanks for your help. I will continue today with this and try to figure out whats going on here.

JimBobSquarePants:
@brianpopow No worries. Merge my branch in if you're happy with the changes so far and we can continue from your branch.

brianpopow:
The issue with Car.bmp is fixed now with 86cccfd. We Definitely need some unit tests for the lossless encoding to make sure its stays fixed now. Im not sure if i should use Calliphora and Car.bmp for tests as those images are quite big and i dont want the tests to take forever.

Hopefully this was the last big issue with lossless encoding (took long enough to find the issue). There are some TODO's still, but i think the remaining ones are trivial. I will start looking into lossy encoding now. I guess this will take also quite some time, too.

JimBobSquarePants:
@brianpopow Well done getting that fixed! Yeah definitely plenty of tests required as we'll need to be able to do a performance pass of the code and I found when I did my bit before that I was touching a few places. 

I'm not worried by test duration. We should get it to drop as we work on improving performance.

SebastianStehle:
Is there a ETA for this task or perhaps for a first beta version?

brianpopow:
> Is there a ETA for this task or perhaps for a first beta version?

@SebastianStehle: Unfortunately i cannot give a realistic ETA when this will be done. It already has taken more time then i have expected. I was not able to continue on this the last few weeks. I hope i will be able to continue work on this soon.

SebastianStehle:
I see that you work really actively on this task. Can you give an ETA now?

brianpopow:
> I see that you work really actively on this task. Can you give an ETA now?

I made good progress over the last few weeks with lossy encoding. I think im getting close 
to encoding the first lossy image. But there are still some bigger tasks open. 
For example encoding and decoding `ALPH` chunks is still not done.

I hope i can get a first working version done by the end of the month, so in 4 weeks maybe.

After that i expect the review will take some weeks also, because its a lot of new code.
So my rough estimate for an ETA would be the end of the year.

brianpopow:
good news: I was able to encode the first lossy image today! 

There are still some failing tests and some open tasks, but im happy that the encoding finally starts to work now. The Vp8 code needs also a big refactoring, which i will do in the next few days.

JimBobSquarePants:
AWESOME!!!

brianpopow:
There are still failing tests for the lossy encoder. the reported error is `Total difference: 0.8816%`. Not sure yet where this is coming from, the bitstream is exactly identical to the reference libwebp bitstream.

JimBobSquarePants:
I’ll have a look asap

brianpopow:
With commit e509bbf0111bce3ccaa6efaff7600629ec3bddd7 the bitstreams of libwebp and ImageSharp are now identical for `Lossy.NoFilter06` again with **quality 75** and **method 2**, but there are still failing tests. The reported difference is Total difference: 0,8377%. I have some problems understanding why that is. The images seem to be identical.

I can not attach the results here because github does not support webp, but the images in the `ActualOutput\WebPEncoderTests`: Encode_Lossy_WithDifferentMethods_Works_Rgba32_test_lossy_m2.webp
Is the same as what libweb produces.

JimBobSquarePants:
@brianpopow How are we testing the difference, are we using a third part  decoder?

brianpopow:
> @brianpopow How are we testing the difference, are we using a third part decoder?

@JimBobSquarePants currently im using ImageSharp Webp decoder, but i have tried the ImageMagic Decoder with the same result.

James-Jackson-South:
Cool. That makes it easier to debug then.

JimBobSquarePants:
Cool. That makes it easier to debug then.

brianpopow:
> ... there are still failing tests. The reported difference is Total difference: 0,8377%. I have some problems understanding why that is. The images seem to be identical.
> 

@JimBobSquarePants: Well i think everything is fine, i just made a wrong assumption. The libwebp and ImageSharp results are the same, but in the test i am comparing the decoded image with the encoded image and the encoded image is expected to be slightly different from the input. Its called "lossy" for a reason :smile:

I think i just need to use a tolerant comparer here.

JimBobSquarePants:
Haha... yep that’ll do it! Great to see all these tests passing!

SebastianStehle:
Thank you very much for your amazing work. I hope to see this on nuget soon.

Btw: would be good to have a sponsor package between 2 and 100$

ErcinDedeoglu:
This is amazing and hard feature.
Maybe you should inform Microsoft about your effort.

https://dotnetfoundation.org/member/become-a-member

brianpopow:
> This is amazing and hard feature.
> Maybe you should inform Microsoft about your effort.
> 
> https://dotnetfoundation.org/member/become-a-member

@ErcinDedeoglu thanks, its always nice to hear some encouraging feedback 😄, and yes it was hard to do, it took me a year so far.
Some feature (which i consider not essential) are still not implemented yet also. The main ones are:

- Encoding / Decoding of animated webp images
- Encoding of the alpha channel of lossy images

Those will not be part of the first release of this feature and added in later. We want to get this feature out and get some feedback from users.

btw. ImageSharp is already a member of the dotnet foundation.

tests/ImageSharp.Tests/Formats/WebP/ImageExtensionsTest.cs
```diff
@@ -13,12 +13,15 @@ namespace SixLabors.ImageSharp.Tests.Formats.Webp
...
+        private readonly Configuration configuration;
+
         public ImageExtensionsTest()
         {
-            Configuration.Default.ImageFormatsManager.AddImageFormat(WebpFormat.Instance);
-            Configuration.Default.ImageFormatsManager.AddImageFormatDetector(new WebpImageFormatDetector());
-            Configuration.Default.ImageFormatsManager.SetDecoder(WebpFormat.Instance, new WebpDecoder());
-            Configuration.Default.ImageFormatsManager.SetEncoder(WebpFormat.Instance, new WebpEncoder());
+            this.configuration = new Configuration();
+            this.configuration.ImageFormatsManager.AddImageFormat(WebpFormat.Instance);
```
JimBobSquarePants, line 21: Does anyone else find the existing registration API here clunky?

antonfirsov: We can define helpers like `configuration.AddWebp()`.

src/ImageSharp/Configuration.cs
```diff
@@ -173,6 +174,17 @@ public Configuration Clone()
             };
         }
 
+        /// <summary>
+        /// Registers the webp format detector, encoder and decoder.
+        /// </summary>
+        public void AddWebp()
```
JimBobSquarePants, line 180: I think these should be per format extension methods.

antonfirsov: Yes, definitely. It should live in the `Experimental` namespace.

antonfirsov: Also: I'm not sure how useful our `IConfigurationModule` abstraction is, but since it's the current practice, we need to implement it for WebP:

https://github.com/SixLabors/ImageSharp/blob/78a584e8482b052d7a9885682299e2f37518d83d/src/ImageSharp/Formats/Jpeg/JpegConfigurationModule.cs

brianpopow: yes makes sense, i have moved it to a extension method. 

The WebpConfigurationModule.cs already exists: [here](https://github.com/SixLabors/ImageSharp/blob/webp/src/ImageSharp/Formats/WebP/WebpConfigurationModule.cs)

JimBobSquarePants:
PR is now at #1552 due to Git LFS force push.

PR #1289, opened 2020-07-26
Work out the crunch configs for lossless webp

stevetemple:
### Prerequisites

### Description
I have had a punt at EncoderAnalyze and think the crunch config should now come out of that OK.

The rest is very much WIP, I started carrying these configs through the EncodeStream in the Vp8LEncoder. The reference implementation starts spinning up threads and then comparing sizes and picking the best. There was nothing so far for that, is that something we want to do. It might be simpler, for now, to get it all working on a single thread save some headaches.

### DIFF
--- a/src/ImageSharp/Formats/WebP/EntropyIx.cs
+++ b/src/ImageSharp/Formats/WebP/EntropyIx.cs
@@ -18,6 +18,8 @@ namespace SixLabors.ImageSharp.Formats.WebP
 
         Palette = 4,
 
-        NumEntropyIx = 5
+        PaletteAndSpatial = 5,
+
+        NumEntropyIx = 6
     }
 }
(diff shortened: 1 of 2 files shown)


src/ImageSharp/Formats/WebP/EntropyIx.cs
```diff
@@ -18,6 +18,8 @@ internal enum EntropyIx
 
         Palette = 4,
 
-        NumEntropyIx = 5
+        PaletteAndSpatial = 5,
```
stevetemple, line 21: This was missing when compared to the reference version, something new they've added since this was started?

brianpopow: good spot, yeah this is indeed new

brianpopow:
> The reference implementation starts spinning up threads and then comparing sizes and picking the best. There was nothing so far for that, is that something we want to do. It might be simpler, for now, to get it all working on a single thread save some headaches.

yes i wanted to keep it single threaded to make it easier for the start.

PR #1299, opened 2020-08-01
Better Edge Detection API

JimBobSquarePants:
### Prerequisites

### Description
As discussed in #1297 It's not currently possible to implement your own custom edge detection kernels. 

This goes against the nature of the API and as such I deem it necessary to refactor the API to allow extensibility. This is a breaking change.

The API has been refactored to expose three structs representing three types of edge detection kernels.

- `EdgeDetectorKernel` Contains a single 2D XY kernel.
- `EdgeDetector2DKernel` Contains two separable 1D kernels. 
- `EdgeDetectorCompassKernel` Contains eight kernels representing each point on the compass. 

A new static class `KnownEdgeDetectorKernels` replaces the current enum `EdgeDetectionOperators`. 

Individual edge detection processors `SobelProcessor` etc have been replaced with versions representing each of the three kernel types.

The resulting code is much simpler and more powerful allowing users to implement any series of gradient operators they require.


src/ImageSharp/ImageSharp.csproj
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub"/>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" />
```
dlemstra, line 23: Dp we not need `PrivateAssets="All"` here?

JimBobSquarePants: It's already marked private in the `Directory.Build.targets` file.

JimBobSquarePants:
Perfect @dalingrin I’m glad I was able to improve matters.

JimBobSquarePants:
@SixLabors/core I'm going to merge this unless anyone objects. I'm very happy with it.

PR #1302, opened 2020-08-03
Add "macos-latest" to build matrix

antonfirsov:
### Prerequisites

### Description
Just noticed there is an available [virtual environment](https://github.com/actions/virtual-environments) for macOS. I think we should add it to the list of CI builds.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -19,7 +19,7 @@ jobs:
                       framework: netcoreapp3.1
                       runtime: -x64
                       codecov: false
-                    - os: macos-10.15
+                    - os: macos-latest
                       framework: netcoreapp3.1
                       runtime: -x64
                       codecov: false
@@ -63,8 +63,8 @@ jobs:
 
             - name: Install macos dependencies
               shell: pwsh
-              #if: "${{matrix.options.os}}" == 'macos-10.15'
-              run: if ("${{matrix.options.os}}" -eq "macos-10.15") { brew install mono-libgdiplus } else { write-output "Not macOS" }
+              if: startsWith(matrix.options.os, 'macos')
+              run: brew install mono-libgdiplus
 
             - name: Build
               shell: pwsh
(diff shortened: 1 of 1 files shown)


antonfirsov:
How can we add this to the list of Required checks?

JimBobSquarePants:
In the branch protection settings.

antonfirsov:
Woops 65 failing tests, this will be a bit harder than I naively thought.

Most of them are rooted in `libgdiplus` being missing for some reason. But there are many tests failing because of over-threshold differences in the output. There are two resize tests failing with 18% and 25% difference which may indicate an actual bug.

[macOS-Failure.log](https://github.com/SixLabors/ImageSharp/files/5017373/macOS-Failure.log)

JimBobSquarePants:
They’re transform tests I think, not resize ones.

brianpopow:
> In the branch protection settings. You should have access but if not, I’ll do it
> 
> Looks like we’re missing libgdiplus on the Mac OS. No idea how to add it

could this do the trick? 

```
- name: Install macos dependencies
  if: matrix.os == 'macos-latest'
  run: brew install mono-libgdiplus
```

antonfirsov:
@brianpopow the condition did not evaluate to `true` somehow.

brianpopow:
> @brianpopow the condition did not evaluate to `true` somehow.

@antonfirsov i cannot spot why that is, looks correct.

antonfirsov:
Looks like it has been moved to `matrix.options.os` .

antonfirsov:
29 test failures:
- 3 test code compatibility issues
- 2 of them likely indicating a macOS-only bug in padded resize
- The rest: accuracy issues, mostly in the range of `0.1% - 0,.9%`. (Reminds me of our 32 bit Framework accuracy issues.)

JimBobSquarePants:
The padded resize bugs are interesting. Not sure how that happens...

Transforms I’ll never be 100% happy with, could be the perfectionist in me but the edges never looked quite right to me. Not as crisp as imagemagick.

antonfirsov:
Doesn't make sense to keep around a stale PR, closing this in favor of #1387.

PR #1304, opened 2020-08-04
Simplify jpeg encoder and fix for Windows ARM

### LINKED ISSUE #1275: JpegEncoder producing incorrect output on windows-arm (was working on beta0005)
### Prerequisites

### Description
I already found the issue #871 which seems to be resolved, but as the issue was for linux-arm this one is maybe specific to windows-arm.
On beta0005, the image was saving correctly, on rc3 it is kinda grayish with squares.

Original image:
image

Resized image:
image

### Steps to Reproduce
Just loading and saving the image:

```
// Loading originalImage
using (var targetImage = new MemoryStream())
{
    using (var image = SixLabors.ImageSharp.Image.Load(originalImage))
    {
        image.SaveAsJpeg(targetImage);
    }
}
// Saving targetImage 
```

### System Configuration
- ImageSharp version: 1.0.0-rc0003
- Other ImageSharp packages and versions: N/A
- Environment (Operating system, version and so on): Windows 10 IoT Core 10.0.17763.1217
- .NET Framework version: UWP - Windows 10, version 1903
- Additional information: Application running on Raspberry PI 3 B / ARM architecture

JimBobSquarePants:
### Prerequisites

### Description
Fixes #1275 

- [x] Use safe cast to span where possible
- [x] Simplify pointer code to use known safe paths used during decoding.
- [x] Limit `Unsafe` API usage
- [x] Use simplified code to debug issues
- [x] Prove it works!


src/ImageSharp/PixelFormats/PixelImplementations/Rgb24.cs
```diff
@@ -201,7 +201,17 @@ public void FromLa32(La32 source)
...
+        {
+#if NETSTANDARD2_0
+            // See https://github.com/SixLabors/ImageSharp/issues/1275
+            this.R = source.R;
+            this.G = source.G;
+            this.B = source.B;
+#else
+            this = source.Rgb;
+#endif
+        }
```
antonfirsov, line 214: Very nice! 🎉 

Remember someone talking about opening a dotnet/runtime issue on twitter, if it exists we should provide this info.

JimBobSquarePants: I didn't open one. (I always struggle to provide a cut down solution to help them). Defo wise though.

JimBobSquarePants:
>Can we do a before/after benchmark comparison for EncodeJpeg, with an image different than Car.bmp?

I suggest re-encoding jpeg420exif.jpg, but config can be Config.ShortCore31 to get it done faster.

@antonfirsov I'm not all that fussed if we lose a percentile or two here. We need to create a hardware accelerated version of our transforms anyway.

antonfirsov:
> if we lose a percentile or two here

I'm afraid it might be more, that's why I suggest an extensive benchmark run. Concerned about `Block8x8F` indexer performance, since it's used very frequently now, and it's code doesn't seem optimal to me.

JimBobSquarePants:
>I'm afraid it might be more, that's why I suggest an extensive benchmark run. Concerned about Block8x8F indexer performance, since it's used very frequently now, and it's code doesn't seem optimal to me.

We use it in a few places in the decoder, hence why I chose to reuse it as a known path. 

Anyway.... Here's the before and after results...

**Before**
|                Method |      Mean |     Error |    StdDev | Ratio | RatioSD |   Gen 0 |   Gen 1 |   Gen 2 | Allocated |
|---------------------- |----------:|----------:|----------:|------:|--------:|--------:|--------:|--------:|----------:|
| 'System.Drawing Jpeg' |  4.498 ms |  2.060 ms | 0.1129 ms |  1.00 |    0.00 | 54.6875 | 31.2500 | 31.2500 |   2.22 MB |
|     'ImageSharp Jpeg' | 70.435 ms | 19.521 ms | 1.0700 ms | 15.66 |    0.22 |       - |       - |       - |   2.96 MB |

**After**
|                Method |      Mean |      Error |    StdDev | Ratio | RatioSD |   Gen 0 |   Gen 1 |   Gen 2 | Allocated |
|---------------------- |----------:|-----------:|----------:|------:|--------:|--------:|--------:|--------:|----------:|
| 'System.Drawing Jpeg' |  4.495 ms |  0.6598 ms | 0.0362 ms |  1.00 |    0.00 | 54.6875 | 31.2500 | 31.2500 |   2.22 MB |
|     'ImageSharp Jpeg' | 70.219 ms | 16.7004 ms | 0.9154 ms | 15.62 |    0.19 |       - |       - |       - |   2.96 MB |

I'd take the massive difference in speed between us and System.Drawing in this benchmark with a pinch of salt. They're not subsampling which will be a huge saving.

PR #1314, opened 2020-08-13
Image.WrapMemory<TPixel> APIs wrapping Memory<byte>

### LINKED ISSUE #1097: Creating a new Image<TPixel> by Wrapping Memory<byte>
As discussed in gitter:

With the current API, it is possible to create a new image in two ways:
- LoadPixelData: which accepts a Span<Byte> , but it makes an internal copy of the source data.
- WrapMemory: which requires a Memory<TPixel> to wrap user memory.

The problem is that in many cases, the source data is provided as a plain byte array. this is specially true when dealing with devices at low level, like cameras and sensors.

In these cases, it would be desirable to do have Image,WrapPixelData(Byte[]) or something like that:

```c#
var bits = new Byte[256*256*4];
var image = Image<Argb32>.WrapPixelData(bits, 256, 256);
```

> @antonfirsov  unfortunately, MemoryMarshal can't cast Memory<T>-s only Spans. There is no API for what you want, but you can implement a MemoryManager<T> which does the span-casting on it's API surface. Feel free to raise a feature request so we remember to add this later.

I am aware that with the current ImageSharp architecture this might be difficult, but I believe it's a case scenario that might show quite often, so if you guys can find a solution for this...

Sergio0694:
# Prerequisites

### Description
This PR adds the following APIs to the `Image` class:

```csharp
{
    public abstract partial class Image
    {
        public static Image<TPixel> WrapMemory<TPixel>(
            Configuration configuration,
            Memory<byte> byteMemory,
            int width,
            int height,
            ImageMetadata metadata)

        public static Image<TPixel> WrapMemory<TPixel>(
            Configuration configuration,
            Memory<byte> byteMemory,
            int width,
            int height)

        public static Image<TPixel> WrapMemory<TPixel>(
            Memory<byte> byteMemory,
            int width,
            int height)
    }
}
```

The internal implementation includes a new `ByteMemoryManager<T>` type that acts as a bridge to wrap `Memory<byte>` instances and use that data to produce a `Memory<TPixel>` instance that can be used to load the image as requested.


src/ImageSharp/Memory/ByteMemoryManager{T}.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        /// <inheritdoc/>
+        protected override void Dispose(bool disposing)
+        {
+        }
+
+        /// <inheritdoc/>
+        public override Span<T> GetSpan()
+        {
```
john-h-k, line 38: Just do 
```cs
MemoryMarshal.Cast<byte, T>(memory.Span))
```

(The current logic doesn't work properly for arrays, it needs to return the span of the whole array not the segment iirc)

Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/1314/commits/32380d2bb947737c319ed381d579f34855c0af18, thanks! 👍

src/ImageSharp/Image.WrapMemory.cs
```diff
@@ -150,5 +150,70 @@ public abstract partial class Image
...
+        /// <typeparam name="TPixel">The pixel type</typeparam>
+        /// <param name="configuration">The <see cref="Configuration"/></param>
+        /// <param name="byteMemory">The byte memory representing the pixel data.</param>
+        /// <param name="width">The width of the memory image.</param>
+        /// <param name="height">The height of the memory image.</param>
+        /// <param name="metadata">The <see cref="ImageMetadata"/>.</param>
+        /// <exception cref="ArgumentNullException">The configuration is null.</exception>
+        /// <exception cref="ArgumentNullException">The metadata is null.</exception>
+        /// <returns>An <see cref="Image{TPixel}"/> instance</returns>
+        public static Image<TPixel> WrapMemory<TPixel>(
```
john-h-k, line 167: <https://docs.microsoft.com/en-us/dotnet/standard/memory-and-spans/memory-t-usage-guidelines>

Seems beneficial to consider exposing `IMemoryOwner<T>` overloads of this to allow ownership transfer?

src/ImageSharp/Memory/ByteMemoryManager{T}.cs
```diff
@@ -0,0 +1,67 @@
...
+            // either a byte[] array or a MemoryManager<byte> instance in this case.
+            ThrowHelper.ThrowArgumentException("The input Memory<byte> instance was not valid.", nameof(this.memory));
+
+            return default;
+        }
+
+        /// <inheritdoc/>
+        public override MemoryHandle Pin(int elementIndex = 0)
+        {
+            return this.memory.Pin();
```
john-h-k, line 45: I think this should be, else it just returns pointer to start of mem block always
```cs
this.memory.Slice(elementIndex).Pin()
```

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
```diff
@@ -80,6 +80,52 @@ public override void Unpin()
...
+
+                /// <inheritdoc/>
+                protected override void Dispose(bool disposing)
+                {
+                }
+
+                /// <inheritdoc/>
+                public override Span<TTo> GetSpan()
+                {
+                    if (MemoryMarshal.TryGetArray(this.memory, out ArraySegment<TFrom> arraySegment))
```
john-h-k, line 102: Same as above, just cast the underlying span directly

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
```diff
@@ -80,6 +80,52 @@ public override void Unpin()
...
+
+                    ThrowHelper.ThrowArgumentException("The input Memory<byte> instance was not valid.", nameof(this.memory));
+
+                    return default;
+                }
+
+                /// <inheritdoc/>
+                public override MemoryHandle Pin(int elementIndex = 0)
+                {
+                    return this.memory.Pin();
```
john-h-k, line 120: Same as above

src/ImageSharp/Image.WrapMemory.cs
Sergio0694: I didn't include them here on purpose as they weren't part of the initial API proposal in #1097.
@antonfirsov should this be added here or would that be for another PR?

src/ImageSharp/Memory/ByteMemoryManager{T}.cs
Sergio0694: Right, I was just thinking with respect to pinning the object itself, which wouldn't care about what reference is used.
Fixed in https://github.com/SixLabors/ImageSharp/pull/1314/commits/eb229964e698852693a8a28748020da93f2c3f22, thanks! 👍

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
Sergio0694: Fixed in https://github.com/SixLabors/ImageSharp/pull/1314/commits/eb229964e698852693a8a28748020da93f2c3f22.

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
Sergio0694: Fixed in https://github.com/SixLabors/ImageSharp/pull/1314/commits/eb229964e698852693a8a28748020da93f2c3f22.

Sergio0694:
@JimBobSquarePants my build is failing locally due to a bunch of errors about a missing assembly key in `AssemblyInfo.cs`:

image

Is there some new additional configuration step I'm missing?
My local clone should be up to date and I'm not seeing anything in the readme 🤔

Thanks!

JimBobSquarePants:
@Sergio0694 Likely your submodules are not up to date and your solution requires cleaning.

Sergio0694:
@JimBobSquarePants I had already tried to clean the solution and to run both `git clean -fdx` and `git submodule update --init --recursive`, that didn't work. I guess I can just try nuking the entire clone and reload it again, for good measure 😄

Sergio0694:
@JimBobSquarePants Well surprise, I disabled the .NET 5 Preview SDK and I can build just fine now 😅

JimBobSquarePants:
>Well surprise, I disabled the .NET 5 Preview SDK and I can build just fine now 

How do you disable a preview? I've been terrified to install any of them.

Sergio0694:
@JimBobSquarePants It's pretty easy actually, you need to go to Tools > Options > Environment > Preview Features and then untick the option regarding preview versions of the .NET Core SDK, and restart VS:

image

I could finally debug the tests locally with that out of the picture, fixed crossed for the CI now! 🚀

Sergio0694:
CI passed! 🚀

The PR should be ready for review with the original API surface mentioned in the linked issue.
Added James and Anton as reviewers 👍

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
```diff
@@ -173,6 +215,73 @@ public void WrapSystemDrawingBitmap_WhenOwned()
                 }
             }
 
+            [Fact]
+            public void WrapMemory_FromBytes_CreatedImageIsCorrect()
+            {
+                Configuration cfg = Configuration.Default.Clone();
```
antonfirsov, line 221: Better call `Configuration.CreateDefault()`

src/ImageSharp/Image.WrapMemory.cs
antonfirsov: No preference, other than it's always easier to review multiple smaller (subsequent) PR-s.

src/ImageSharp/Image.WrapMemory.cs
```diff
@@ -150,5 +150,70 @@ public abstract partial class Image
...
+            Configuration configuration,
+            Memory<byte> byteMemory,
+            int width,
+            int height,
```
antonfirsov, line 171: How do we validate, what do we throw if the sizes mismatch? Also needs tests for that case.

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
Sergio0694: I actually had copied that from the other existing test 😄

I proceeded to update the other one too while I was it then, done in https://github.com/SixLabors/ImageSharp/pull/1314/commits/ff5e3015a0567858195841f8941ee8b771189f18.

src/ImageSharp/Image.WrapMemory.cs
Sergio0694: Sounds like a preference to me 😄
Will keep this PR restricted to consumed buffers then, and we can create another one after this one is merged to also add support for owned ones through `IMemoryOwner<byte>` then.

src/ImageSharp/Image.WrapMemory.cs
Sergio0694: I'm really just passing the new `Memory<TPixel>` instance to the `Image<TPixel>` constructor like the previously existing APIs, so I just assumed `Image<TPixel>` would perform the input validation already? My idea here was to avoid repeating the same tests if not needed (eg. the overloads taking a `Memory<TPixel>` argument have no validation either).

Let me know though if you want any changes being done here (possibly to previous overloads too then?).

JimBobSquarePants:
@Sergio0694 I will get to this. Just fixing a critical issue #1316 dirst

Sergio0694:
@JimBobSquarePants Don't worry, no rush! 😄
Just got very excited about that sponsorship from the MS OSS program and wanted to get back and contribute to something 🚀

src/ImageSharp/Image.WrapMemory.cs
JimBobSquarePants: I can't find any validation in the current codebase that ensures the memory size matches. We check in `MemoryGroup<T>.Wrap(...)` that the buffers are uniform but there doesn't seem to be anything else.

Sergio0694: Wait hold on, doesn't that mean the already existing APIs right now lack validation too then? 🤔

I mean, like this:

https://github.com/SixLabors/ImageSharp/blob/85f91f6992da966dc10216b664e3ff833f425e56/src/ImageSharp/Image.WrapMemory.cs#L98-L111

JimBobSquarePants: Yeah, I think so. I could be wrong but I couldn't find anything. A unit test would quickly tell us.

Sergio0694: Yeah, looks like at the end of the day the wrapped memory just gets passed here:

https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Memory/Buffer2D%7BT%7D.cs#L30-L40

And the constructor doesn't validate that the input group actually matches the specified size.
Will add a new unit test and will add the checks to both the new and previous APIs then! 👍

JimBobSquarePants: That matches what I found. Thanks!

Sergio0694:
@JimBobSquarePants Just noticed this bit while adding the unit tests:

https://github.com/SixLabors/ImageSharp/blob/801961dfb96e290ca82a28b24e5ed933d4343a53/src/ImageSharp/Memory/MemoryOwnerExtensions.cs#L16-L22

Wouldn't it be more efficient to just have `GetLength` return the length from the `Memory<T>`?
Accessing the `Span` property of a `Memory<T>` instance is relatively expensive.
Unless there's a specific reason this was done? Not seeing a comment here explaining that though 🤔

JimBobSquarePants:
@Sergio0694 Wow! Never actually looked at the code for `Memory<T>` before. Yeah, can't think of any reason why we would favor getting the length of the span. If you see any other instances where we do that please update them also.

src/ImageSharp/Image.WrapMemory.cs
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/1314/commits/fa59a73c39d6ea25edfb690957be2c494176893f.

src/ImageSharp/Memory/MemoryOwnerExtensions.cs
```diff
@@ -13,20 +14,37 @@ namespace SixLabors.ImageSharp.Memory
     /// </summary>
     internal static class MemoryOwnerExtensions
     {
+        /// <summary>
+        /// Gets a <see cref="Span{T}"/> from an <see cref="IMemoryOwner{T}"/> instance.
+        /// </summary>
+        /// <param name="buffer">The buffer</param>
+        /// <returns>The <see cref="Span{T}"/></returns>
+        [Pure]
```
JimBobSquarePants, line 22: Any particular reason for adding the `Pure` attribute?

JimBobSquarePants (approved):
Not sure we need the new code analysis attributes (we don't use them elsewhere) but otherwise looks great! 👍

src/ImageSharp/Image.WrapMemory.cs
```diff
@@ -150,5 +152,73 @@ public abstract partial class Image
...
+            Configuration configuration,
+            Memory<byte> byteMemory,
+            int width,
+            int height)
+            where TPixel : unmanaged, IPixel<TPixel>
+            => WrapMemory<TPixel>(configuration, byteMemory, width, height, new ImageMetadata());
+
+        /// <summary>
+        /// Wraps an existing contiguous memory area of 'width' x 'height' pixels,
+        /// allowing to view/manipulate it as an ImageSharp <see cref="Image{TPixel}"/> instance.
```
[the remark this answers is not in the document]
Sergio0694: Done in https://github.com/SixLabors/ImageSharp/pull/1314/commits/d7380f0763da99a0aa70b6f4aabb30e56a4d184c.

src/ImageSharp/Memory/MemoryOwnerExtensions.cs
Sergio0694: I just really like to be as verbose and explicit as I can when writing code, and to decorate methods with as many attributes as possible. That includes eg. `[Pure]`, `[DoesNotReturn]`, `[NotNullWhen(bool)]`, etc. Having the code analyzer display all the appropriate hints and warnings when you do so make sthe whole developement experience much better and less error prone. For instance when using `[Pure]`, you make it clear that the method does not modify any state and that calling it without using the return is useless. This way you'll get a warning if you ever call it and you forget to eg. assign its return value to something. Saved me more than once before, highly recommend 😄

Of course, I can remove them here if you prefer to just do a separate PR to just pass over the entire repo and add the missing annotations. Or if you prefer not to use them at all for some other reason. Let me know!

src/ImageSharp/Image.WrapMemory.cs
```diff
@@ -38,6 +38,7 @@ public abstract partial class Image
         {
             Guard.NotNull(configuration, nameof(configuration));
             Guard.NotNull(metadata, nameof(metadata));
+            Guard.IsTrue(pixelMemory.Length == width * height, nameof(pixelMemory), "The length of the input memory doesn't match the specified image size");
```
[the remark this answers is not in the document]
Sergio0694: Wait what allocations? 🤔
Also do you want me to create an issue about this to make sure we don't forget in the future?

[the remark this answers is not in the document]
JimBobSquarePants: Thought that would get you thinking. We always allocate the string whether true or false.

src/ImageSharp/Memory/MemoryOwnerExtensions.cs
JimBobSquarePants: I think I'd like to do without tbh. They're not used in the runtime anywhere I can find and by my understanding it's possible to mark an impure method with the attribute which can lead to masking of potential issues.

src/ImageSharp/Image.WrapMemory.cs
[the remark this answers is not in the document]
Sergio0694: What string though? Both of those are constants, so the JIT will just load the address here.
I feel like I might be missing something obvious 😅

Eg. if you have this (full repro [here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4axAMxNSDITQDemncyRMUDACqqMACgCyMfNACeAHl48AfAwEuoVwBKIxoGcIYASVwrKA4YOwAzbElcGDQGLmwBCES7fzcgjIAiKwALXlwGSoZsBlxssWkGSC5cDGweJnI6YqCNagBfGhoAbScMMogAE0jxSUcYSZm5poB5MT4INpYAOQhIrkkvLwBzIIBdE27zYkto2Pi7YAgISQZEyWxTjOZGbF+PQYwBC1GMQxogyAA)):

```csharp
public static void Test(Memory<int> memory)
{
    IsTrue(false, nameof(memory), "This is a sample constant string");
}

[MethodImpl(MethodImplOptions.NoInlining)]
public static void IsTrue(bool flag, string a, string b)
{
}
```

You get this:

```x86asm
C.Test(System.Memory`1<Int32>)
    L0000: mov r8, 0x16e9fc8ab98
    L000a: mov r8, [r8]
    L000d: mov rdx, 0x16e9fc8ab28
    L0017: mov rdx, [rdx]
    L001a: xor ecx, ecx
    L001c: mov rax, C.IsTrue(Boolean, System.String, System.String)
    L0026: jmp rax
```

So you only pay the two indirect memory accesses to load the strings, but you have no allocations.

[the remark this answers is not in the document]
JimBobSquarePants: Ah yeah, I'm thinking when we interpolate (which I thought we were doing here). It might seem useful but it adds a lot of overhead

src/ImageSharp/Memory/MemoryOwnerExtensions.cs
Sergio0694: That is correct, the attribute is not enforced and you need to take the extra care to use it appropriately.
Remove in https://github.com/SixLabors/ImageSharp/pull/1314/commits/d8618052d17a5a50109e46541bb31a059d4de34a as requested 😊

JimBobSquarePants: Thanks mate.

PR #1316, opened 2020-08-16
Fix non-seekable stream reading.

JimBobSquarePants:
### Prerequisites

### Description
Fixes our non-seekable stream loading which was attempting to use the non-supported `Length` property.
https://github.com/SixLabors/ImageSharp/blob/8e2792e3ab96dd69d8e39255e83afdee493b1b26/src/ImageSharp/Image.FromStream.cs#L734

>// If CanSeek is false, Position, Seek, Length, and SetLength should throw.

https://source.dot.net/#System.Private.CoreLib/Stream.cs,51

The fix utilizes an internal `ChunkedMemoryStream` implementation to allow pooling of byte buffers via the use of non-contiguous chunks. This let's us allocate chunks from the buffer pool on demand upon read. 

I've deliberately kept `FixedCapacityPooledMemoryStream` around for now as I may have use for it in the future experiments and it's internal.


src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -0,0 +1,544 @@
...
+        private MemoryChunk writeChunk;
+
+        // Offset into chunk to write to
+        private int writeOffset;
+
+        // Current chunk to read from
+        private MemoryChunk readChunk;
+
+        // Offset into chunk to read from
+        private int readOffset;
```
antonfirsov, line 45: I'm probably missing something, but why are we writing/reading to/from different chunks?
Shouldn't `Stream.Position` indicate where we are in both cases?

tests/ImageSharp.Tests/Image/ImageTests.Load_FromStream_PassLocalConfiguration.cs
```diff
@@ -39,14 +39,25 @@ public void Configuration_Stream_Agnostic()
             [Fact]
             public void NonSeekableStream()
             {
-                var stream = new NoneSeekableStream(this.DataStream);
+                var stream = new NonSeekableStream(this.DataStream);
```
antonfirsov, line 42: These tests are insufficient.  They are only here to prove the different overloads of extension methods do what they promise by mocking a decoder. `this.DataStream` contains only 16 bytes!

At this point, I would add a couple of integration tests, stressing all decoders with `NonSeekableStream`, and images of different sizes.

antonfirsov (commented):
Wow, this is fairly complex stuff, good job! I'll need to have another round of review after understanding the stuff I'm currently missing.

Even thought it's an urgent and critical bugfix, I believe it's worth to invest a bit more time to get higher confidence this actually works, even if the cost is a delay of few days. For this, I'm suggesting to add integration tests aggressively.

Also: do we have test cases that verify reading/writing correctness, when the operation crosses the boundaries of chunks?

image

Etc ...

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -0,0 +1,544 @@
...
+
+                int length = 0;
+                MemoryChunk chunk = this.memoryChunk;
+                while (chunk != null)
+                {
+                    MemoryChunk next = chunk.Next;
+                    if (next != null)
+                    {
+                        length += chunk.Length;
+                    }
```
antonfirsov, line 93: If we store the number of "fully written" chunks, the loop could be replaced by a multiplication.

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -0,0 +1,544 @@
...
+                this.EnsureNotDisposed();
+
+                if (this.readChunk is null)
+                {
+                    return 0;
+                }
+
+                int pos = 0;
+                MemoryChunk chunk = this.memoryChunk;
+                while (chunk != this.readChunk)
```
antonfirsov, line 120: Consider a trick similar I'm suggesting for `Length`. These properties are heavily used in our decoders as far as I can see.

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -0,0 +1,544 @@
...
+            }
+        }
+
+        /// <inheritdoc/>
+        public override void Flush()
+        {
+        }
+
+        /// <inheritdoc/>
+        public override int Read(byte[] buffer, int offset, int count)
```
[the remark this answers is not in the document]
JimBobSquarePants: We actually never call the `Span<byte>` overload since everything is wrapped by the `BufferedReadStream` implementation.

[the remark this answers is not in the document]
antonfirsov: To me it seems we do use it actually:
https://github.com/SixLabors/ImageSharp/blob/8e2792e3ab96dd69d8e39255e83afdee493b1b26/src/ImageSharp/IO/BufferedReadStream.cs#L349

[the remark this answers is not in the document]
JimBobSquarePants: Yeah, I realized that almost immediately after commenting. Was away from my computer so couldn't correct myself.

JimBobSquarePants:
@antonfirsov Thanks for the thorough review.

I've made a few changes based upon the feedback.

- Added optimized read/write span methods and passed all reading/writing of buffers through those methods.
- Increased the chunk size to 81920 bytes
- Updated unit tests to demonstrate read/write over chunk boundaries
- Added integration tests matching existing seekable tests.
- Added benchmarks to demonstrate no lack of performance during decode.

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants: I had a look but the benchmarks showed no need for the optimization so chose not to complicate things.

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants: See above

tests/ImageSharp.Tests/Image/ImageTests.Load_FromStream_PassLocalConfiguration.cs
JimBobSquarePants: I've copied the existing `Image.Identify` and `Image.IdentifyAsync` to demonstrate accuracy. The unit tests also cover reading and writing across chunk boundaries.

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants: For my sanity mostly. This is hard.

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -0,0 +1,544 @@
...
+    /// Provides an in-memory stream composed of non-contiguous chunks that doesn't need to be resized.
+    /// Chunks are allocated by the <see cref="MemoryAllocator"/> assigned via the constructor
+    /// and is designed to take advantage of buffer pooling when available.
+    /// </summary>
+    internal sealed class ChunkedMemoryStream : Stream
+    {
+        /// <summary>
+        /// The default length in bytes of each buffer chunk.
+        /// </summary>
+        public const int DefaultBufferLength = 4096;
```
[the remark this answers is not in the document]
JimBobSquarePants: Done. Went with 81920

antonfirsov (changes requested):
I pushed the heavyweight integration test I suggested, and it fails for 65 out of 308 images.
We need to `Skip` the new theory, but add some of it's cases to our regular test suite. The lightweight `Identify` test is using the (simple) BMP decoder and doesn't really stress the new complex code we are about to add.

Regarding benchmarks:
I don't think they deliver good indicators in this particular case, since they are only showing a few numbers about primitive operations being executed on small input streams. 

We need to know: How many times do our decoders touch `Length` and `Position` for a typical huge image (I've seen couple of usages in PNG). If the number is high enough (Eg. bc. relevant methods are called in a loop), we need to  proove the perf is not affected when decoding a **large** image.

JimBobSquarePants:
Odd that’s failing, will investigate. Re length and position, the buffered stream only calls that in the underlying stream when the buffer is exhausted so I wouldn’t consider that a hot path.

JimBobSquarePants:
I've just disabled the test on 32bit for now. Will think of something better in a bit.

src/ImageSharp/IO/ChunkedMemoryStream.cs
antonfirsov: Ah ok, so it's a special stream implementation where writes go always to the end of the stream, right? Would add a comment about this.

tests/ImageSharp.Tests/IO/ChunkedMemoryStreamTests.cs
```diff
@@ -0,0 +1,371 @@
...
+        {
+            if (!TestEnvironment.Is64BitProcess)
+            {
+                return;
+            }
+
+            Image<Rgba32> expected;
+            try
+            {
+                expected = Image.Load<Rgba32>(testFileFullPath);
```
antonfirsov, line 325: I know it's my code but I would really like to replace `Image.Load` with `TestFile` stuff now to make the tests at least a bit faster by utilizing it's caching mechanism.  Would do it, but I'm on a train now, may try later, if you also lack time.

antonfirsov (approved):
Just a few wishes, otherwise looks good now!

PR #1329, opened 2020-08-25
Fix missing SDK version for test building it GitHub Workflow

Turnerj:
See the build/test error in #1328 - this *might* patch around that neatly via using the matrix options. This PR is as much of a test of the concept.

### Prerequisites

### Description
Uses the options matrix to define the SDK version to install, thus hopefully guaranteeing the right version is installed for tests.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -18,22 +18,27 @@ jobs:
                       framework: netcoreapp3.1
                       runtime: -x64
                       codecov: false
+                      sdkVersion: "3.1.x"
                     - os: windows-latest
                       framework: netcoreapp3.1
                       runtime: -x64
                       codecov: true
+                      sdkVersion: "3.1.x"
                     - os: windows-latest
                       framework: netcoreapp2.1
                       runtime: -x64
                       codecov: false
+                      sdkVersion: "2.1.x"
                     - os: windows-latest
                       framework: net472
                       runtime: -x64
                       codecov: false
+                      sdkVersion: "3.1.x"
                     - os: windows-latest
                       framework: net472
                       runtime: -x86
                       codecov: false
+                      sdkVersion: "3.1.x"
 
         runs-on: ${{matrix.options.os}}
         if: "!contains(github.event.head_commit.message, '[skip ci]')"
@@ -55,7 +60,7 @@ jobs:
             - name: Setup DotNet SDK
               uses: actions/setup-dotnet@v1
               with:
-                  dotnet-version: "3.1.x"
+                  dotnet-version: ${{matrix.options.sdkVersion}}
 
             - name: Build
               shell: pwsh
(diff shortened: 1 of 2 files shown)


Turnerj:
Slight issue with it - seems that the `Benchmarks` project and the `ProfilingSandbox` project are being built but require specific versions of the SDK themselves. The builds might need to be more selective to avoid building either project as neither is used for CI AFAIK.

JimBobSquarePants:
I think we have to install both for each build now.

https://github.com/actions/setup-dotnet

Turnerj:
Hmmmm, I was thinking going the other way, changing the `ci-build.ps1` script to just build the main library and the test library.
https://github.com/SixLabors/ImageSharp/blob/d95a996ed5ba14a1421dafeb844a56ab08211ead/ci-build.ps1#L11

If that was building the test library, the test library would automatically build the main library.

.github/workflows/build-and-test.yml
```diff
@@ -55,7 +60,7 @@ jobs:
             - name: Setup DotNet SDK
               uses: actions/setup-dotnet@v1
               with:
-                  dotnet-version: "3.1.x"
+                  dotnet-version: ${{matrix.options.sdkVersion}}
```
JimBobSquarePants, line 63: I can't figure out from the docs whether you can simply do. [ "2.1.x", "3.1.x" ]

Turnerj: Digging through the TypeScript code for the action, it looks to only take string values. Internally, the value is getting passed into here: https://github.com/actions/setup-dotnet/blob/e3d81d94533f1e31c11d94a1fdf6f60f096a7bc5/src/installer.ts#L33

JimBobSquarePants: Yeah, ran a quick test  in the Web repo. Big fail

Turnerj:
I need to go get some sleep - I've pushed a change which tweaks what is built for the CI which may work around the issue. Its cool if this isn't the direction you want to go though.

Hopefully it works!

tocsoft:
I don't think we need the "Setup DotNet SDK" step at all anymore, I believe all the sdks we need are all preinstalled on the images... we needed it in the past but I don't think we do anymore.

JimBobSquarePants:
@tocsoft You're right!

JimBobSquarePants:
> I don't think we need the "Setup DotNet SDK" step at all anymore, I believe all the sdks we need are all preinstalled on the images... we needed it in the past but I don't think we do anymore.

@tocsoft You're right!

JimBobSquarePants:
Thanks for your help @Turnerj 

I ended up just pushing the fix in the other open PR. 

Turns out the new VMs have everything!

https://github.com/actions/virtual-environments/releases

Turnerj:
Awesome - glad you got it working!

PR #1362, opened 2020-09-26
Fixed typo in AdaptiveThreshold

rold2007:
### Prerequisites

### Description
I saw a clear typo in the code. So I added a new test which would validate the change. All the images used in the tests were previously grayscale so that the typo would go unnoticed. That being said, using adaptive thresholding on a color image is not a usual use case.


brianpopow (approved):
@rold2007 thank you for the fix!

PR #1373, opened 2020-10-05
WIP: NET5 Targets for Testing

JimBobSquarePants:
### Prerequisites

### Description
Adds relevant targets and compiler conditionals to target NET 5 testing.

- [x] Refactor props to future proof existing compiler conditionals.
- [ ] Add `net5.0` build targets to tests and benchmarks
- [ ] Update benchmark configuration
- [ ] Add CI test profiles (Blocked until RC2)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -77,15 +77,17 @@
     </When>
     <Otherwise>
       <!--NETCORE 3.1, NET5.0, and future versions -->
-      <DefineConstants>$(DefineConstants);SUPPORTS_MATHF</DefineConstants>
-      <DefineConstants>$(DefineConstants);SUPPORTS_HASHCODE</DefineConstants>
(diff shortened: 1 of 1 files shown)


antonfirsov:
Since 3.1 builds will work fine with .NET 5, and we do not need any .NET 5 - specific API-s, I would repurpose this PR to add a CI verification step instead of a new build target in the library.

Ideally this would mean adding `net5.0` targets to ImageSharp.Tests and ImageSharp.Benchmarks and extending `build-and-test.yml`. But I'm wondering now how could the net5.0 test artifacts be built on top of the 3.1 library artifacts?

JimBobSquarePants:
Sounds like a plan. 

I'll keep the new `Choose` constant constructor layout in the props file as that more accurately reflects the inheritance model.

CI wise we should be able to run the tests with the `net5.0` target and everything should run as expected. I'm sure this will be handled gracefully by the runtime.

PR #1374, opened 2020-10-07
Add Avx backed Block8x8F Transpose method

JimBobSquarePants:
### Prerequisites

### Description
Adds a new Avx backed `Block8x8F.TransposeInto` method. Used to help reduce bottleneck in Jpeg Discrete Cosine Transform

https://github.com/SixLabors/ImageSharp/blob/4dadf2494018564c5b7b1199faeb5ab3840a1ee3/src/ImageSharp/Formats/Jpeg/Components/FastFloatingPointDCT.cs#L53-L55

~Benchmarks aren't anywhere near what I hoped they'd be. Maybe I've done something wrong. The only obvious difference I can see so far between this and the implementations in things like `Matrix4x4` is that I'm using `Unsafe.As` over `Avx.Load`~
Benchmarking around 16% faster with some savings in jpeg decoding also. Thanks @tannergooding !

~I've ported both methods from here using a processor directive to hide one from the compiler.~
I've just focused on the implementation that had less instructions.
https://stackoverflow.com/questions/25622745/transpose-an-8x8-float-using-avx-avx2/25627536#25627536

Tagging the brains trust to see if you chaps have any suggestions.

@antonfirsov 
@tannergooding 
@saucecontrol

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19041.508 (2004/?/20H1)
.NET Core SDK=5.0.100-rc.1.20452.10
  [Host]     : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT
  DefaultJob : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT
```
**Block8x8F.Transpose**

|               Method |     Mean |    Error |   StdDev |   Median | Ratio | RatioSD |
|--------------------- |---------:|---------:|---------:|---------:|------:|--------:|
| TransposeIntoVector4 | 46.71 ns | 0.956 ns | 1.277 ns | 45.96 ns |  1.00 |    0.00 |
|     TransposeIntoAvx | 39.97 ns | 0.157 ns | 0.139 ns | 39.97 ns |  0.84 |    0.02 |

**Jpeg Decode**

Before
 
|                         Method |            TestImage |       Mean |      Error |    StdDev | Ratio | RatioSD | Gen 0 | Gen 1 | Gen 2 |  Allocated |
|------------------------------- |--------------------- |-----------:|-----------:|----------:|------:|--------:|------:|------:|------:|-----------:|
| 'Decode Jpeg - System.Drawing' | Jpg/b(...)e.jpg [21] |   5.201 ms |  0.5051 ms | 0.0277 ms |  1.00 |    0.00 |     - |     - |     - |      176 B |
|     'Decode Jpeg - ImageSharp' | Jpg/b(...)e.jpg [21] |  11.378 ms |  0.5299 ms | 0.0290 ms |  2.19 |    0.01 |     - |     - |     - |    15888 B |
|                                |                      |            |            |           |       |         |       |       |       |            |
| 'Decode Jpeg - System.Drawing' | Jpg/b(...)f.jpg [28] |  14.157 ms |  0.6681 ms | 0.0366 ms |  1.00 |    0.00 |     - |     - |     - |      176 B |
|     'Decode Jpeg - ImageSharp' | Jpg/b(...)f.jpg [28] |  27.588 ms |  1.7655 ms | 0.0968 ms |  1.95 |    0.00 |     - |     - |     - |    16896 B |
|                                |                      |            |            |           |       |         |       |       |       |            |
| 'Decode Jpeg - System.Drawing' | Jpg/i(...)e.jpg [43] | 337.594 ms | 39.9392 ms | 2.1892 ms |  1.00 |    0.00 |     - |     - |     - |      216 B |
|     'Decode Jpeg - ImageSharp' | Jpg/i(...)e.jpg [43] | 264.551 ms | 84.5057 ms | 4.6320 ms |  0.78 |    0.02 |     - |     - |     - | 36022512 B |

After

|                         Method |            TestImage |       Mean |       Error |    StdDev | Ratio | RatioSD | Gen 0 | Gen 1 | Gen 2 |  Allocated |
|------------------------------- |--------------------- |-----------:|------------:|----------:|------:|--------:|------:|------:|------:|-----------:|
| 'Decode Jpeg - System.Drawing' | Jpg/b(...)e.jpg [21] |   5.464 ms |   0.8390 ms | 0.0460 ms |  1.00 |    0.00 |     - |     - |     - |      176 B |
|     'Decode Jpeg - ImageSharp' | Jpg/b(...)e.jpg [21] |  10.433 ms |   0.6750 ms | 0.0370 ms |  1.91 |    0.02 |     - |     - |     - |    15918 B |
|                                |                      |            |             |           |       |         |       |       |       |            |
| 'Decode Jpeg - System.Drawing' | Jpg/b(...)f.jpg [28] |  15.075 ms |  32.4878 ms | 1.7808 ms |  1.00 |    0.00 |     - |     - |     - |      176 B |
|     'Decode Jpeg - ImageSharp' | Jpg/b(...)f.jpg [28] |  28.504 ms |   8.1014 ms | 0.4441 ms |  1.91 |    0.19 |     - |     - |     - |    16896 B |
|                                |                      |            |             |           |       |         |       |       |       |            |
| 'Decode Jpeg - System.Drawing' | Jpg/i(...)e.jpg [43] | 339.520 ms |  24.3315 ms | 1.3337 ms |  1.00 |    0.00 |     - |     - |     - |      216 B |
|     'Decode Jpeg - ImageSharp' | Jpg/i(...)e.jpg [43] | 254.203 ms | 108.8514 ms | 5.9665 ms |  0.75 |    0.02 |     - |     - |     - | 36022512 B |


src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
```diff
@@ -596,5 +600,160 @@ private static void GuardBlockIndex(int idx)
...
+            Vector256<float> r0 = Unsafe.As<Vector4, Vector256<float>>(ref this.V0L);
+            Vector256<float> r1 = Unsafe.As<Vector4, Vector256<float>>(ref this.V1L);
+            Vector256<float> r2 = Unsafe.As<Vector4, Vector256<float>>(ref this.V2L);
+            Vector256<float> r3 = Unsafe.As<Vector4, Vector256<float>>(ref this.V3L);
+            Vector256<float> r4 = Unsafe.As<Vector4, Vector256<float>>(ref this.V4L);
+            Vector256<float> r5 = Unsafe.As<Vector4, Vector256<float>>(ref this.V5L);
+            Vector256<float> r6 = Unsafe.As<Vector4, Vector256<float>>(ref this.V6L);
+            Vector256<float> r7 = Unsafe.As<Vector4, Vector256<float>>(ref this.V7L);
+
+            Vector256<float> t0 = Avx.UnpackLow(r0, r1);
```
tannergooding, line 641: I'd expect there are a few spills as part of this due to having 8-16 live registers at a time.

You might see better perf by intermixing the three steps given here. That is, handle
t0, t2, t4, t6 to produce r0, r1, r4, r5 and store them
and then handle t1, t3, t5, t7 to produce r2, r3, r6, r7 and store them

I imagine this will help minimize the number of stack spills that happen and may also allow better pipelining on the latest CPUs (without hindering older CPUs)

tannergooding: Noting that I did not check the disassembly here, I am just speculating based on past experience 😄

tests/ImageSharp.Benchmarks/Codecs/Jpeg/BlockOperations/Block8x8F_Transpose.cs
```diff
@@ -0,0 +1,45 @@
...
+using BenchmarkDotNet.Attributes;
+using SixLabors.ImageSharp.Formats.Jpeg.Components;
+
+namespace SixLabors.ImageSharp.Benchmarks.Codecs.Jpeg.BlockOperations
+{
+    public class Block8x8F_Transpose
+    {
+        private static readonly Block8x8F Source = Create8x8FloatData();
+
+        [Benchmark]
```
antonfirsov, line 13:
```suggestion
        [Benchmark(Baseline=true)]
```
Will lead to better comparison.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
JimBobSquarePants: It helped! 🎉

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
```diff
@@ -596,5 +600,98 @@ private static void GuardBlockIndex(int idx)
...
+            Vector256<float> t1 = Avx.UnpackHigh(r0, r1);
+            Vector256<float> t3 = Avx.UnpackHigh(r2, r3);
+            v = Avx.Shuffle(t1, t3, 0x4E);
+            Unsafe.As<Vector4, Vector256<float>>(ref d.V2L) = Avx.Blend(t1, v, 0xCC);
+            Unsafe.As<Vector4, Vector256<float>>(ref d.V3L) = Avx.Blend(t3, v, 0x33);
+
+            Vector256<float> t5 = Avx.UnpackHigh(r4, r5);
+            Vector256<float> t7 = Avx.UnpackHigh(r6, r7);
+            v = Avx.Shuffle(t5, t7, 0x4E);
+            Unsafe.As<Vector4, Vector256<float>>(ref d.V6L) = Avx.Blend(t5, v, 0xCC);
```
antonfirsov, line 692: @tannergooding can we rely on codegen to produce optimal assembly when working with ` Unsafe.As`? (In comparison to `Store` / `LoadVector256` )

We wanted to avoid pinning, if possible.

antonfirsov (commented):
For a while I was frightened that the gains are negligible, but seems like a visible improvement, good job! 

Mostly LGTM, but let's see what do others think.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
JimBobSquarePants: I would also like to subscribe to this newsletter.

tannergooding: We should still fold the loads/stores. I believe there are a couple edge cases where we may generate an additional `lea` instruction, but the disassembly would nerd to be checked to see if it's an issue.

JimBobSquarePants: Is there a new language feature I am unaware of that allows this without pinning?

https://source.dot.net/#System.Private.CoreLib/Matrix4x4.cs,273

```c#
Vector128<float> M11 = AdvSimd.LoadVector128(&value1.M11);
```

antonfirsov: In this case `Matrix4x4` is passed by value to the method, so the compiler can make sure `value1` is living on the stack, which means it never gets moved around by the GC. => the address can be used safely without pinning anything

antonfirsov: Actually: `JpegBlockPostprocessor` is the owner of all 3 blocks passed to `TransformIDCT`, and that struct always lives on the stack:
https://github.com/SixLabors/ImageSharp/blob/78a584e8482b052d7a9885682299e2f37518d83d/src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegComponentPostProcessor.cs#L79

Therefore, you can get a pointer to the whole `Block8x8` buffer at the beginning of the method "safely" with Unsafe:
```C#
float* dPtr = (float*)Unsafe.AsPointer(ref d);
```

JimBobSquarePants: Ah yes of course. This all makes sense thanks! 

I might give the pointer approach a try tomorrow just to see if there is any difference.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
```diff
@@ -596,5 +600,98 @@ private static void GuardBlockIndex(int idx)
...
+        /// AVX-only variant for executing <see cref="TransposeInto(ref Block8x8F)"/>.
+        /// <see href="https://stackoverflow.com/questions/25622745/transpose-an-8x8-float-using-avx-avx2/25627536#25627536"/>
+        /// </summary>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public void TransposeIntoAvx(ref Block8x8F d)
+        {
+            Vector256<float> r0 = Avx.InsertVector128(
+               Unsafe.As<Vector4, Vector128<float>>(ref this.V0L).ToVector256(),
+               Unsafe.As<Vector4, Vector128<float>>(ref this.V4L),
+               1);
```
Turnerj, line 634: This doesn't change the assembly that is generated but it might show the intention better as:

```suggestion
            Vector256<float> r0 = Vector256.Create(
                Unsafe.As<Vector4, Vector128<float>>(ref this.V0L), 
                Unsafe.As<Vector4, Vector128<float>>(ref this.V4L)
            );
```

JimBobSquarePants: Can anyone explain the difference in the precision of the output instruction here?

`vmovups` to `vmovupd` at `L0006`

[Original](https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcAMACOSB0AcgK5QCmATgJYDGEA3OlrgQErEB2wV5+AwgHsoAByoAbSgGVKAN1pkGTHHnzsuPMvgCSXahwi1FGZW07deO4HoN18ADQAcANkZp02D7gDM2CFeI0wNgAQmICNADWDgAeDgBi7p4A3omennA+NAL6QVRc2JJUAF5k2AC82E4ALK6pad7YAGpkgQIUVU2YADKu9ek+za3tnay1aH0eGU0twG0djUg9dWlTg7PDC6Poy/3TQ/MIS+MTqzNzTQhbbsd9p/tNXkcTDWvnjV5XO5MDZxtVTycfvdGlVPjd6nd1vMAKwA25AqFNaFg56Qt5OOEQhHolGAvaIxoAdkxK2xG0JozSX2w1IA2gBZMjAAAWAgAJloRGIABSMlnsznCMQAeWE3GyEHwAEEAOYyigKAwyMg6MR5PIygCUAF1qVM4B0ACoUMD6YQCCAqrgCKUyaLchUAMxCYUiMXi2DZmupKXBz1eFAQ0KcAB5HWEwMAAHzYCg4Cq26LafSUYABpAIBzc6kTACq+jAjs0UogIYDVQANPi2hmHGGI9Gow6yM6WVRJY1upr8IaBAGg05uZqKzm+vmIIXi6Xy1X05n6wJI1Gm07sG2O/9h6P6khNb0JgBfal1A9AA=)

```
Block8x8F.TransposeIntoAvx(Block8x8F ByRef)
    L0000: sub esp, 0x20
    L0003: vzeroupper
    L0006: vmovups xmm0, [ecx]
    L000a: add ecx, 0x80
    L0010: vmovupd xmm1, [ecx]
    L0014: vinsertf128 ymm0, ymm0, xmm1, 1
    L001a: vmovupd [esp], ymm0
    L001f: vzeroupper
    L0022: add esp, 0x20
    L0025: ret
```

[Suggested Change](https://sharplab.io/#v2:C4LghgzgtgPgAgBgARwIwDoByBXKBTAJwEsBjCAbgFgAoRFDAJWwDtgj90BhAeygAciAG0IBlQgDdSeCjTpp0TVuzzoAkq2LMIpGbWTzFbDuuCbtZdAA0AHADYq1GkVaFmYQUj5gCbd0gim2CTASABCgtwkANbWAB7WAGI0SClIAN7JqVkA9LlIADwQuFDeAJ4AfJlZKbnZSACCSMy4AEaESNwAZv4k7t5IJNx4nZ2kRHisEEjOSGAFEHh4AwTDALwAROGRMfEJ69mV1NWptQXZRVAlBBVV1XAAzAPcWiHOISJEAF5Lq0i2ACwOW5ZB5IABqeGC3AI/3BCAAMg5jilQRCoTC4QwgUdkajIcBobCwahEcDUnj0UTUFiaGSUY80QSMWCAEyknHHClMoksmmODl3Bn4wng+7s5EoIWU0V8umS8HC5n/cW4qXc8H/WUCkFqkVggCsKs5uuZ+q1Eq5etsRsFCulYNs5tVdvVYIA7DadS69W6+VlbqC4LCACoEMBaPjcBYmbj1cSxAAUK26W2icUSSAAJgBKOUZbXIuVZRnQln62z5ToRMDAcpIAjIX4lghl2xcFY1vAJovIgCqWjAnRU9Qg+Wb/wANN6CKgWdZK9Xa+Uk8MkMAABZECDoMEI7NTnvHfsQQfD0fjqfN2fzqvcGvlZfJteb7dg5W5gvHbNI5EAXzpv5AA==)

```
Block8x8F.TransposeIntoAvx(Block8x8F ByRef)
    L0000: sub esp, 0x20
    L0003: vzeroupper
    L0006: vmovupd xmm0, [ecx]
    L000a: add ecx, 0x80
    L0010: vmovupd xmm1, [ecx]
    L0014: vinsertf128 ymm0, ymm0, xmm1, 1
    L001a: vmovupd [esp], ymm0
    L001f: vzeroupper
    L0022: add esp, 0x20
    L0025: ret
```

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
JimBobSquarePants: Decided against it. I don't want to risk changes to external code breaking this.

antonfirsov:
@JimBobSquarePants wouldn't it make sense to isolate testinfra changes into a follow-up PR?

(1) I'm getting unsure about the right approach about providing proper test coverage for the different codepaths, (2) we also need to deal with #1376.

Regarding (1): We may probably want to introduce one or more entire test targets instead/besides the unit test helpers to make 100% sure we cover end to end scenarios on systems where certain instruction set extensions are unavailable.
- Pro: less reliance on Unit Test quality, and extensiveness of code reviews when adding SIMD features
- Con: running more targets

JimBobSquarePants:
I could revert. I just didn’t want to add more clutter (I don’t like having to write duplicate tests for the same feature)

JimBobSquarePants:
I don’t fancy extra targets either. Builds already take far too long

antonfirsov:
My point is that the transpose stuff is good to merge as is. Would be nice to have it done, and take our time to discuss/finish/review the rest.

tests/ImageSharp.Tests/TestUtilities/FeatureTesting/FeatureTestRunner.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Runs the given test <paramref name="action"/> within an environment
+        /// where the given <paramref name="intrinsics"/> features.
+        /// </summary>
+        /// <param name="action">The test action to run.</param>
+        /// <param name="intrinsics">The intrinsics features.</param>
+        public static void RunWithHwIntrinsicsFeature(
```
antonfirsov, line 109: Anyways, great stuff! Would be nice to add tests for this, but this is currently blocked.

JimBobSquarePants: I do have tests waiting to push. Just hit the executor issue so stopped. I’ll revert later, merge the current state and add as a separate PR.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
JimBobSquarePants: Could be error but I spotted a slight slow down in the change. Keeping as-is for now.

PR #1378, opened 2020-10-12
Update Dependencies

JimBobSquarePants:
### Prerequisites

### Description
Updating all possible dependencies to their latest versions.

~There is a known issue https://github.com/SixLabors/ImageSharp/issues/1380 with the reference BMP encoder on Ubuntu that I have applied a workaround for now by skipping on Ubuntu CI. This affects our coverage report but is tested on all other platforms.~

Fixed.

~I have a single outstanding upgrade of `Microsoft.Net.Compilers.Toolset` required but there's an [issue](https://github.com/dotnet/arcade/issues/6393) that either needs fixing upstream for that or we can use the `ActiveIssueAttribute` to skip the affected tests on NET 4.7.2~

~Thoughts please on that.~

I've disabled the codec tests using `ActiveIssueAttribute` and removed unnecessary implementations of the executor for the BokehBlur tests.

Looks like the executor issue now fails on Ubuntu also. 😞 
https://github.com/SixLabors/ImageSharp/pull/1378/checks?check_run_id=1256202313#step:6:351

### DIFF
--- a/tests/Directory.Build.targets
+++ b/tests/Directory.Build.targets
@@ -37,7 +37,7 @@
     <PackageReference Update="SharpZipLib" Version="1.3.0" />
     <PackageReference Update="System.Drawing.Common" Version="4.7.0" />
     <PackageReference Update="xunit" Version="2.4.1" />
-    <PackageReference Update="xunit.runner.visualstudio" Version="2.4.1" />
+    <PackageReference Update="xunit.runner.visualstudio" Version="2.4.3" />
   </ItemGroup>
 
 </Project>
--- a/tests/Directory.Build.targets
+++ b/tests/Directory.Build.targets
@@ -31,7 +31,7 @@
     <PackageReference Update="coverlet.collector" Version="1.3.1-preview.27.gdd2237a3be" PrivateAssets="All"/>
     <PackageReference Update="Magick.NET-Q16-AnyCPU" Version="7.22.0" />
     <PackageReference Update="Microsoft.DotNet.RemoteExecutor" Version="5.0.0-beta.20069.1" />
-    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.7.1" />
     <PackageReference Update="Moq" Version="4.14.6" />
     <PackageReference Update="Pfim" Version="0.9.1" />
     <PackageReference Update="SharpZipLib" Version="1.3.0" />
--- a/tests/Directory.Build.targets
+++ b/tests/Directory.Build.targets
@@ -26,11 +26,12 @@
   <ItemGroup>
     <!--Test Dependencies-->
     <PackageReference Update="BenchmarkDotNet" Version="0.12.1" />
-    <PackageReference Update="BenchmarkDotNet.Diagnostics.Windows" Version="0.12.1" Condition="'$(OS)' == 'Windows_NT'" />    
+    <PackageReference Update="BenchmarkDotNet.Diagnostics.Windows" Version="0.12.1" Condition="'$(OS)' == 'Windows_NT'" />
     <PackageReference Update="Colourful" Version="2.0.5" />
     <PackageReference Update="coverlet.collector" Version="1.3.1-preview.27.gdd2237a3be" PrivateAssets="All"/>
     <PackageReference Update="Magick.NET-Q16-AnyCPU" Version="7.22.0" />
-    <PackageReference Update="Microsoft.DotNet.RemoteExecutor" Version="5.0.0-beta.20069.1" />
+    <PackageReference Update="Microsoft.DotNet.RemoteExecutor" Version="6.0.0-beta.20513.1" />
+    <PackageReference Update="Microsoft.DotNet.XUnitExtensions" Version="6.0.0-beta.20513.1" />
     <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.7.1" />
     <PackageReference Update="Moq" Version="4.14.6" />
     <PackageReference Update="Pfim" Version="0.9.1" />
--- a/tests/Directory.Build.targets
+++ b/tests/Directory.Build.targets
@@ -29,12 +29,12 @@
     <PackageReference Update="BenchmarkDotNet.Diagnostics.Windows" Version="0.12.1" Condition="'$(OS)' == 'Windows_NT'" />    
     <PackageReference Update="Colourful" Version="2.0.5" />
     <PackageReference Update="coverlet.collector" Version="1.3.1-preview.27.gdd2237a3be" PrivateAssets="All"/>
-    <PackageReference Update="Magick.NET-Q16-AnyCPU" Version="7.15.5" />
+    <PackageReference Update="Magick.NET-Q16-AnyCPU" Version="7.22.0" />
     <PackageReference Update="Microsoft.DotNet.RemoteExecutor" Version="5.0.0-beta.20069.1" />
     <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.5.0" />
-    <PackageReference Update="Moq" Version="4.14.5" />
+    <PackageReference Update="Moq" Version="4.14.6" />
     <PackageReference Update="Pfim" Version="0.9.1" />
-    <PackageReference Update="SharpZipLib" Version="1.2.0" />
+    <PackageReference Update="SharpZipLib" Version="1.3.0" />
     <PackageReference Update="System.Drawing.Common" Version="4.7.0" />
     <PackageReference Update="xunit" Version="2.4.1" />
     <PackageReference Update="xunit.runner.visualstudio" Version="2.4.1" />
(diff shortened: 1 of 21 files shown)


tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.cs
```diff
@@ -129,10 +129,10 @@ public async Task DecodeAsnc_DegenerateMemoryRequest_ShouldTranslateTo_ImageForm
         [Theory]
         [InlineData(TestImages.Jpeg.Baseline.Jpeg420Small, 0)]
         [InlineData(TestImages.Jpeg.Issues.ExifGetString750Transform, 1)]
-        [InlineData(TestImages.Jpeg.Issues.ExifGetString750Transform, 10)]
+        [InlineData(TestImages.Jpeg.Issues.ExifGetString750Transform, 15)]
```
JimBobSquarePants, line 132: Test was flaky on NET Core 2.1, this seemed to make it less so.

tests/ImageSharp.Tests/TestUtilities/TestEnvironment.cs
```diff
@@ -170,7 +170,10 @@ private static void PrepareRemoteExecutor()
             }
 
             string testProjectConfigPath = TestAssemblyFile.FullName + ".config";
-            File.Copy(testProjectConfigPath, remoteExecutorConfigPath);
+            if (File.Exists(testProjectConfigPath))
```
JimBobSquarePants, line 173: This file only exists if there are binding redirects.

tests/Directory.Build.targets
```diff
@@ -26,18 +26,19 @@
...
-    <PackageReference Update="BenchmarkDotNet.Diagnostics.Windows" Version="0.12.1" Condition="'$(OS)' == 'Windows_NT'" />    
+    <PackageReference Update="BenchmarkDotNet.Diagnostics.Windows" Version="0.12.1" Condition="'$(OS)' == 'Windows_NT'" />
     <PackageReference Update="Colourful" Version="2.0.5" />
     <PackageReference Update="coverlet.collector" Version="1.3.1-preview.27.gdd2237a3be" PrivateAssets="All"/>
     <PackageReference Update="Magick.NET-Q16-AnyCPU" Version="7.22.0" />
-    <PackageReference Update="Microsoft.DotNet.RemoteExecutor" Version="5.0.0-beta.20069.1" />
-    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.5.0" />
-    <PackageReference Update="Moq" Version="4.14.5" />
+    <PackageReference Update="Microsoft.DotNet.RemoteExecutor" Version="6.0.0-beta.20513.1" />
+    <PackageReference Update="Microsoft.DotNet.XUnitExtensions" Version="6.0.0-beta.20513.1" />
```
antonfirsov, line 34: Are we only adding this because of `ActiveIssueAttribute`?

tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
```diff
@@ -39,22 +39,40 @@ public class BmpDecoderTests
...
+
+        [ActiveIssue("https://github.com/dotnet/arcade/issues/6393", TargetFrameworkMonikers.NetFramework)]
+        [Theory]
+        [WithFileCollection(nameof(MiscBmpFiles), PixelTypes.Rgba32)]
+        public void BmpDecoder_CanDecode_MiscellaneousBitmaps_WithLimitedAllocatorBufferCapacity<TPixel>(
+            TestImageProvider<TPixel> provider)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            // dotnet xunit doesn't respect filter.
+            if (TestEnvironment.IsFramework)
```
antonfirsov, line 63: We are keeping  `dotnet xunit`  around only to do the x86 test run, which eventually happens only on Framework for some reason. Would be important to get the x86 coverage across all platforms with `dotnet test`, where `dotnet(.exe)` has to come from the 32 bit SDK, see https://github.com/xunit/xunit/issues/1123#issuecomment-449642224

Is this something we can easily get with `build-and-test.yml`?

tests/ImageSharp.Tests/Formats/Gif/GifDecoderTests.cs
```diff
@@ -195,20 +195,27 @@ public void GifDecoder_DegenerateMemoryRequest_ShouldTranslateTo_ImageFormatExce
...
+
             static void RunTest(string providerDump, string nonContiguousBuffersStr)
             {
                 TestImageProvider<TPixel> provider = BasicSerializer.Deserialize<TestImageProvider<TPixel>>(providerDump);
 
                 provider.LimitAllocatorBufferCapacity().InPixelsSqrt(100);
 
                 using Image<TPixel> image = provider.GetImage(GifDecoder);
-                image.DebugSave(provider);
+                image.DebugSave(provider, nonContiguousBuffersStr);
```
antonfirsov, line 218: good catch

tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.Progressive.cs
```diff
@@ -44,7 +62,7 @@ static void RunTest(string providerDump, string nonContiguousBuffersStr)
             RemoteExecutor.Invoke(
                 RunTest,
                 providerDump,
-                enforceDiscontiguousBuffers ? "Disco" : string.Empty)
+                "Disco")
```
antonfirsov, line 65: We no longer need the parameter I guess.

tests/ImageSharp.Tests/Processing/Processors/Convolution/BokehBlurTest.cs
```diff
@@ -138,21 +138,10 @@ public void Serialize(IXunitSerializationInfo info)
...
-                provider.RunValidatingProcessorTest(
-                    x => x.BokehBlur(value.Radius, value.Components, value.Gamma),
-                    testOutputDetails: value.ToString(),
-                    appendPixelTypeToFileName: false);
-            }
-
-            RemoteExecutor
-                .Invoke(RunTest, BasicSerializer.Serialize(provider), BasicSerializer.Serialize(value))
-                .Dispose();
+            provider.RunValidatingProcessorTest(
```
antonfirsov, line 141: This was remote executed to reduce memory pressure, especially on x86. `BokehBlur` is very memory intensive.

antonfirsov (commented):
A few concerns/questions.

tests/Directory.Build.targets
antonfirsov: Ahh, I see it's not a trivial feature because of the optional `TargetFrameworkMonikers` attribute. Cool 👍

tests/ImageSharp.Tests/Processing/Processors/Convolution/BokehBlurTest.cs
JimBobSquarePants: They seem to be ok now. As I recall it uses less memory than it did originally. Haven't seen any failures in the last few test runs due to this.

tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
JimBobSquarePants: I don't know. Originally I wanted to do it via setup-dotnet but choosing the target framework there is not supported.

https://github.com/actions/setup-dotnet/issues/72

tests/Directory.Build.targets
JimBobSquarePants: Yeah, it's pretty nifty. Some cool other attributes available also.

tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
JimBobSquarePants: I'm not sure the x86 SDKs are installed by default.

https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#net-core-sdk

tests/ImageSharp.Tests/Processing/Processors/Convolution/BokehBlurTest.cs
JimBobSquarePants: Definitely rock solid. Multiple test runs with zero issues.

tests/Directory.Build.targets
JimBobSquarePants: Not actually needed now but keeping it as it's far too useful!

brianpopow (approved):
LGTM

JimBobSquarePants:
@antonfirsov I'm gonna merge this as I'm happy with the result and have cleaned up anything questionable.

tests/ImageSharp.Tests/Processing/Processors/Convolution/BokehBlurTest.cs
antonfirsov: Ok for now then, but if we'll see OOM-s after introducing new 32bit targets, this is a good place to get back.

Note that we were getting OOM-s not because of a specific test but rather because of reaching a critical mass of memory intentse-tests. I also pushed some Jpeg tests out-process together with the Bokeh ones back in the past. The price is  execution time OFC. Testing image-processing is hard :)

JimBobSquarePants: Preach!

PR #1385, opened 2020-10-15
Advanced Feature Testing

JimBobSquarePants:
### Prerequisites

### Description
As we add more hardware intrinsics into the library we need a good way to test them.

Ideally we want to follow the [same pattern](https://source.dot.net/#System.Private.CoreLib/Matrix4x4.cs,265) the runtime uses for abstracting intrinsics so our tests should be able to call the same method multiple times with difference features disabled.

This PR adds two bits of functionality that allow this pattern
- `FeatureTestRunner.RunWithHwIntrinsicsFeature`. Which utilizes `RemoteExecutor` + environmental constants to disable intrinsics features.
- `HwIntrinsics_SSE_AVX` which is the first implementation of config classes for benchmarking a single method with separate runs utilizing environmental constants to disable intrinsics features.


antonfirsov (approved):
Looks good. Let's see where this concept brings us.

PR #1388, opened 2020-10-16
Pixel remapping/Swizzling

### LINKED ISSUE #1115: Pixel remapping/Swizzling
This feature is implemented in many graphics hardwares at least of most gaming consoles. Adding this feature would enable ImageSharp to natively allow for these specifications of graphics hardware to be supported out of the box.

As per OpenCV naming and implementation this feature is most likely known as "Remapping" or "Point remapping". So as the naming goes, it should incorporate either one.

Point remappings are, as the name suggests, a process to rearrange the pixels from a source to a destination point, where each source point is modified as per following method:
`dst(x,y) = src(fx(x,y), fy(x,y))`

As for the implementation following this function, it should allow executing an action for both the x- and y-coordinate, and return its modified value respectively. Or an even simpler approach as seen in the Kanvas library of the Kuriimu2 project, where it just takes in a `Point` and returns a new one with the modified coordinates.

Sources:
OpenCV documentation on remapping: https://docs.opencv.org/2.4/modules/imgproc/doc/geometric_transformations.html?highlight=remap
Kuriimu2's Kanvas Swizzle interface:https://github.com/FanTranslatorsInternational/Kuriimu2/blob/dev/src/Kontract/Kanvas/IImageSwizzle.cs

Evangelink:
### Prerequisites

### Description

Fix #1115


src/ImageSharp/Processing/Processors/Transforms/SwizzleProcessor{TSwizzler,TPixel}.cs
```diff
@@ -0,0 +1,33 @@
...
+        protected override void OnFrameApply(ImageFrame<TPixel> source)
+        {
+            for (int y = 0; y < source.Height; y++)
+            {
+                var pixelRowSpan = source.GetPixelRowSpan(y);
+                for (int x = 0; x < source.Width; x++)
+                {
+                    var newPoint = this.swizzler.Transform(new Point(x, y));
+                }
+            }
```
antonfirsov, line 30: If you are afraid of perf penalty from `new Point(x, y)` and similar, you can do sg like:

```C#
            var p = new Point(x, y);
            var newPoint = default;
            for (p.y = 0; y < source.Height; p.y++)
            {
                var pixelRowSpan = source.GetPixelRowSpan(p.y);
                for (p.x = 0; x < source.Width; p.x++)
                {
                    this.swizzler.Transform(p, out newPoint);
                }
            }
```

antonfirsov:
@Evangelink are you still interested / do you have time to push this to completion?

Evangelink:
Sorry @antonfirsov I got distracted by some other work. I will give it a try tomorrow, I hope to be able to make it work!

Evangelink:
@antonfirsov I have made some change but I have no confidence at all in what I have written, I have tried to put up 2 tests based on others but I also don't feel confident about them (not sure I really got how the test infra is working).

I am happy to try to finish the work but I feel like I need a full baby-sitting so I am afraid that's pointless for you... Let me know if you'd rather close this.

Evangelink:
Also I am not sure what's the process to update your submodule.

antonfirsov (commented):
@Evangelink you don't need reference images to test this feature. I think it's better to implement some basic "unswizzle" code right in the tests then call `ImageComparer.Exact.VerifySimilarity` to compare the unswizzled image to the original one.

antonfirsov (commented):
Looks good with one suggestion.

@onepiecefreak3 I know it's been ages, but any chance you can take a look if this is something that would meet your expectations?

src/ImageSharp/Processing/Processors/Transforms/ISwizzler.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// Gets the size of the image after transformation.
+        /// </summary>
+        Size DestinationSize { get; }
+
+        /// <summary>
+        /// Applies the swizzle transformation to a given point.
+        /// </summary>
+        /// <param name="point">Point to transform.</param>
+        /// <param name="newPoint">The transformed point.</param>
+        void Transform(Point point, out Point newPoint);
```
JimBobSquarePants, line 21: Are we sure about `void` and `out` here? Why not return the point?

antonfirsov: Ah yeah. Unless there are benchmarks (or a sharplab.io decompilation link) proving that the `out` variant is significantly better, I would prefer the retval version.

Evangelink:
@antonfirsov @JimBobSquarePants Issues are unrelated to this PR, e.g.:
```
     Decode_DegenerateMemoryRequest_ShouldTranslateTo_ImageFormatException<Rgba32>(provider: Jpg/progressive/Festzug.jpg[Rgba32]) [PASS]
      Output:
        Cannot decode image. Failed to allocate buffers for possibly degenerate dimensions: 1443x1071.

Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```

antonfirsov (approved):
LGTM

JimBobSquarePants:
Should we add the ability to swizzle a sub rectangle of the source or are we happy just doing the whole image?

Evangelink:
@JimBobSquarePants I am not sure how you could apply it only to some sub-rectangle. The new coordinates could be on some space outside of the rectangle (either on an existing location in the image or even outside of the image), what would you do in such case?

JimBobSquarePants:
@Evangelink Let's not bother for now then. I'm happy with this as is. 👍

JimBobSquarePants (approved):

Evangelink:
@JimBobSquarePants You are welcome! @antonfirsov Thank you for the strong support :)

PR #1391, opened 2020-10-16
Fix ARM Build for UWP Release

### LINKED ISSUE #1204: ImageSharp-rc1 fails to build on UWP Release ARM
### Prerequisites

- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode

### Description

Building a UWP app in Release ARM fails to build when using `ImageSharp-rc1`, and the UWP compiler version `6.2.10` (latest). Downgrading to `-beta007` works for now.
Here's the issue I'm getting:

> ILT0005: 'C:\Users\Sergio\.nuget\packages\runtime.win10-arm.microsoft.net.native.compiler\2.2.8-rel-28605-00\tools\ARM\ilc\Tools\nutc_driver.exe @"C:\Users\Sergio\source\repos\ImageSharpRepro\ImageSharpRepro\obj\ARM\Release\ilc\intermediate\MDIL\ImageSharpRepro.rsp"' returned exit code -1073741819	ImageSharpRepro			

Building Release x86, x64 and ARM64 works fine 🤷‍♂️

Pinging @MattWhilden for his infinite .NET Native knowledge and wisdom 🧙‍♂️

### Steps to Reproduce

- Create a new UWP project (`19041` target, `17763` minimum)
- Add the `ImageSharp-rc1` package
- (Try to) build in Release ARM

### System Configuration

- ImageSharp version: `ImageSharp-rc1`
- Other ImageSharp packages and versions: `.NETCore.UniversalWindowsPlatform 6.2.10`
- Environment (Operating system, version and so on): Windows 10 x64 `19041.264`

JimBobSquarePants:
### Prerequisites

### Description
Fixes #1204

From conversations with the .NET Native team. Thanks @Sergio0694 for arranging that.

>the crash happens while compiling CdfApplicationRowOperation.Invoke. I’m unclear what the best workaround would be. You can always place a NoOptimization flag on the method and that will probably get rid of the issue because the optimizer won’t be involved. From what I see on the stack, the optimizer is trying to hoist some operation out of the loop when it hits the problem.

The compiler actually fails unless both `CdfApplicationRowOperation.Invoke` and `GrayscaleLevelsRowOperation.Invoke` are marked with `MethodImplOptions.NoOptimization`. 

I attempted to work around the problem by manually inlining the luminance method call and better managing scope per loop iteration. While this didn't work they should result in better performance across all platforms.

I've also added notes for potential further operations.

Proof

### DIFF
--- a/src/ImageSharp/Processing/Processors/Normalization/GlobalHistogramEqualizationProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Normalization/GlobalHistogramEqualizationProcessor{TPixel}.cs
@@ -106,15 +106,23 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
             }
 
             /// <inheritdoc/>
+#if NETSTANDARD2_0
+            // https://github.com/SixLabors/ImageSharp/issues/1204
+            [MethodImpl(MethodImplOptions.NoOptimization)]
+#else
             [MethodImpl(InliningOptions.ShortMethod)]
+#endif
             public void Invoke(int y)
             {
                 ref int histogramBase = ref MemoryMarshal.GetReference(this.histogramBuffer.GetSpan());
                 ref TPixel pixelBase = ref MemoryMarshal.GetReference(this.source.GetPixelRowSpan(y));
+                int levels = this.luminanceLevels;
 
                 for (int x = 0; x < this.bounds.Width; x++)
                 {
-                    int luminance = GetLuminance(Unsafe.Add(ref pixelBase, x), this.luminanceLevels);
+                    // TODO: We should bulk convert here.
+                    var vector = Unsafe.Add(ref pixelBase, x).ToVector4();
+                    int luminance = ImageMaths.GetBT709Luminance(ref vector, levels);
                     Unsafe.Add(ref histogramBase, luminance)++;
                 }
             }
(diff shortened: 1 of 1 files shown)


Sergio0694:
Hey, this is awesome! 🎉🎉🎉

I mean, I'm sad to see that at the end of the day this build error was a consequence of the whole value delegate trick, but thankfully it still seems to work just fine in all other processors on UWP, and the issue was only when building for ARM32 anyway. It's great that the .NET Native team could identified exactly the faulting methods so we can just deoptimize them instead of burning performance for the whole library, yay! 🚀

src/ImageSharp/Processing/Processors/Normalization/GlobalHistogramEqualizationProcessor{TPixel}.cs
```diff
@@ -106,15 +106,23 @@ protected override void OnFrameApply(ImageFrame<TPixel> source)
...
+            // https://github.com/SixLabors/ImageSharp/issues/1204
+            [MethodImpl(MethodImplOptions.NoOptimization)]
+#else
             [MethodImpl(InliningOptions.ShortMethod)]
+#endif
             public void Invoke(int y)
             {
                 ref int histogramBase = ref MemoryMarshal.GetReference(this.histogramBuffer.GetSpan());
                 ref TPixel pixelBase = ref MemoryMarshal.GetReference(this.source.GetPixelRowSpan(y));
+                int levels = this.luminanceLevels;
```
Sergio0694, line 119: I see some bonus loop hoisting, nice 🚀

src/ImageSharp/Processing/Processors/Normalization/GlobalHistogramEqualizationProcessor{TPixel}.cs
```diff
@@ -147,18 +155,27 @@ public void Invoke(int y)
             }
 
             /// <inheritdoc/>
+#if NETSTANDARD2_0
+            // https://github.com/SixLabors/ImageSharp/issues/1204
+            [MethodImpl(MethodImplOptions.NoOptimization)]
```
Sergio0694, line 160: Shame that we can't just target UWP specifically for this, but .NET Standard 2.0 as a whole should mostly be for back-compat anyway, so this seems absolutely fine to me. This whole processor is not one of the most used either, which is good.

Sergio0694 (approved):
Ship it! 🚀

JimBobSquarePants:
@Sergio0694 Yeah, really glad to have made some progress here. Would have been nice to have been able to target a little more specifically but I don't believe it's possible (is it?)

src/ImageSharp/Processing/Processors/Normalization/GlobalHistogramEqualizationProcessor{TPixel}.cs
JimBobSquarePants: Ha! Just typed the same thing below 😝

Sergio0694:
@JimBobSquarePants It actually is, though it requires some more setup.
We'd need to build the library using `MSBuild.Sdk.Extras` and add `uap10.0.17763` (or above) as one of the targets.
This way UWP devs would get the UWP class library directly. In theory this should also work if they're using `ImageSharp` from a .NET Standard 2.0 project and then referencing it from the UWP app, as NuGet should then resolve the most specific target at build time, which would be the UWP one. You can see an example of this [here](https://github.com/windows-toolkit/WindowsCommunityToolkit/blob/c6c66022a967969d6f7c6d4a04ad0a6a3951a830/Microsoft.Toolkit.Uwp/Microsoft.Toolkit.Uwp.csproj#L4), we're using `MSBuild.Sdk.Extras` for the UWP packages in the toolkit as well.

This might not be worth it though considering that both UWP as a whole (I mean, .NET Native) and .NET Standard 2.0 are basically mostly considered legacy at this point, as people gradually just move to .NET 5. Given that the change is so small to fix this, it might be worth it to just keep the library as is. Thoughts?

JimBobSquarePants:
Sounds like a lot of extra work/testing for little reward. Let's stick to what we have for now.

Sergio0694:
The day UWP finally switches to .NET 6 I will pop the biggest bottle of champagne ever made 🤣

JimBobSquarePants:
I'll drink to that! 🍾

Issue #1387, opened 2020-10-15
Extend CI test execution to macOS


antonfirsov:
Initial attempt failed, we need to fix the following issues:
https://github.com/SixLabors/ImageSharp/pull/1302#issuecomment-668183518

Disabling the two (likely) bugs with `[ActiveIssue]` might be also fine.

rohansp:
@antonfirsov I am interested in contributing to this project.Can I help?

JimBobSquarePants:
@rohansp Absolutely! That would be awesome. 

Basically the process would be to build and run the tests on your Mac and then raise a PR adding the new CI option to the workflow like #1302  Any questions you have on any failing tests can be raised either here or in a draft PR and we can work together to ensure everything passes.

antonfirsov:
@rohansp we validate our output by comparing the resulting image to a reference output under (`tests/Images/External/ReferenceOutput` -- comes from a submodule). The comparison is tolerant. The majority of the failures are because of a small difference. These could be fixed by relaxing the tolerance threshold a bit, preferably in a selective way.

There are a few failures which may indicate a bug. By running the tests locally, there is a way to observe the difference visually. Actual output is saved under `tests/Images/ActualOutput` when running in debug.

rohansp:
@JimBobSquarePants Small problem.I have a Windows 10 machine. MacOS is not installed on my machine.Sorry, didn't realize earlier.

AlexNDRmac:
@antonfirsov I can help you with cross-platform code coverage and codecov reports

antonfirsov:
@AlexNDRmac this issue is not about metrics or reports, the problem is that at the moment we are not able to do stable test runs on macOS at all. I believe this is easy to fix, see the opening post and https://github.com/SixLabors/ImageSharp/issues/1387#issuecomment-712183926 for details.

If you think you want to help with this, I'm happy to answer any questions!

AlexNDRmac:
I see the problem with tests on Mac, I'll try to sort out with it. I'll ping you when I'll be ready to ask smth :)

JimBobSquarePants:
Love it. Thanks!

AlexNDRmac:
@antonfirsov I've fixed error with `libgdiplus`... but I see some any problem with tests, like `allocation error`. Maybe you can recommend more stable branch than `master` for isolated testing macOS CI with `libgdiplus` issue?

JimBobSquarePants:
[...]

PR #1403, opened 2020-10-26
Enable support to use libgdiplus library on macOS

AlexNDRmac:
### Prerequisites

### Description

- Fixed error with libgdiplus on macOS (`Unable to load DLL 'libgdiplus'`)
- Added cache action to CI workflow for all OS in build matrix

Related issue: #1387


.github/workflows/build-and-test.yml
```diff
@@ -14,23 +14,33 @@ jobs:
         strategy:
             matrix:
                 options:
-                    - os: ubuntu-latest
```
JimBobSquarePants, line 17: Can you revert the formatting change here please.

.github/workflows/build-and-test.yml
```diff
@@ -52,6 +62,25 @@ jobs:
                   git fetch --prune --unshallow
                   git submodule -q update --init --recursive
 
+            - name: Setup DotNet SDK
```
JimBobSquarePants, line 65: There's currently an [issue](https://github.com/actions/setup-dotnet/issues/136) with the setup-dotnet action which is preventing us from using it. 

The macOs and Ubuntu images already have the required SDK installed. 
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#language-and-runtime

src/ImageSharp/ImageSharp.csproj
```diff
@@ -28,6 +28,14 @@
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" />
   </ItemGroup>
 
+  <!-- Enable support for using libgdiplus on macOS -->
```
JimBobSquarePants, line 31: This isn't a requirement for the main library. ImageSharp does not use System.Drawing

tests/ImageSharp.Benchmarks/ImageSharp.Benchmarks.csproj
```diff
@@ -25,7 +25,9 @@
     <PackageReference Include="BenchmarkDotNet.Diagnostics.Windows" Condition="'$(OS)' == 'Windows_NT'" />
     <PackageReference Include="Colourful" />
     <PackageReference Include="Pfim" />
-    <PackageReference Include="SharpZipLib"  />
+    <!-- Enable support for using libgdiplus on macOS -->
```
JimBobSquarePants, line 28: This reference should be in the [Directory.Build.targets](https://github.com/SixLabors/ImageSharp/blob/120080b99f3f1be0fc4ecec6906ee683ea999a3d/tests/Directory.Build.targets) file. 

The reference can then be included in the relevant test projects and filtered by OS in a manner similar to the following.
https://github.com/SixLabors/ImageSharp/blob/120080b99f3f1be0fc4ecec6906ee683ea999a3d/tests/ImageSharp.Benchmarks/ImageSharp.Benchmarks.csproj#L25

JimBobSquarePants (changes requested):
Some reference changes required here and a little cleanup. The unit test memory issues are known to us and only appear randomly. They relate to one of our test reference codecs.

.github/workflows/build-and-test.yml
AlexNDRmac: I'm not sure that is a bug. In my projects I'm also had such problems with netcore 2.x and netcore 3.x.
I can propose one solution for this:
- Change build matrix
```yaml
matrix:
  framework:
    - 'netcoreapp2.1'
    - 'netcoreapp3.1'
    - 'net472'
```
- Configure matrix options:
```yaml
matrix:
  options:
    - os: ubuntu-latest
      runtime: -x64
      codecov: true
   - os: macos-latest
      runtime: -x64
      codecov: true
  - os: windows-latest
      runtime: -x64
      codecov: true

  exclude:
   - os: ubuntu-latest
      framework: 'net472'
   - os: macOS-latest
      framework: 'net472'
```
- Change runner:
```yaml
jobs:
  Build:
    name: ${{ matrix.options.os }} ${{ matrix.framework }}
    runs-on: ${{ matrix.options.os }}
```

After that - you'll have next builds:
```
macOS-latest netcoreapp2.1
macOS-latest netcoreapp3.1
ubuntu-latest netcoreapp2.1
ubuntu-latest netcoreapp3.1
windows-latest netcoreapp2.1
windows-latest netcoreapp3.1
windows-latest net472
```

And little change for `actions/setup-dotnet@v1`
```yaml
- name: Setup DotNet SDK 2.1.x
  if: (runner.os == 'macOS' || runner.os == 'Linux') && matrix.framework == 'netcoreapp2.1'
  uses: actions/setup-dotnet@v1
  with:
     dotnet-version: "2.1.x"

- name: Setup DotNet SDK 3.1.x
  if: (runner.os == 'macOS' || runner.os == 'Linux') && matrix.framework == 'netcoreapp3.1'
  uses: actions/setup-dotnet@v1
  with:
     dotnet-version: "3.1.x"
```

tests/ImageSharp.Benchmarks/ImageSharp.Benchmarks.csproj
AlexNDRmac: '$(OS)' have only two options: 'Windows_NT' and 'Unix'. So... using `$([MSBuild]::IsOSPlatform('OSX'))` more precise and we do not use non-relevant package in Linux.

.github/workflows/build-and-test.yml
JimBobSquarePants: I'd consider it a bug since it used to work well enough until recent updates. I also chatted to the repo owner on Twitter and he encouraged me to raise the issue. 

I don't understand the purpose of the `exclude:` additions here?

I'm not convinced your `dotnet-setup` filter will work. The application is multi targeted so if you setup `2.1.x` then that will update the path and the build will likely fail. (_Or does it work on unix for some bizarre reason?_)

What I'm trying to get across here is that we want the absolute minimum of changes in order to make this work. 

`macos-latest` on `3.1.x` is the only target we need for now and we don't want to be running the tests with code coverage on multiple platforms.

tests/ImageSharp.Benchmarks/ImageSharp.Benchmarks.csproj
JimBobSquarePants: Ah yes, I forgot this was updated. Yep that's a better filter!

.github/workflows/build-and-test.yml
AlexNDRmac: exclude - means, CI will skip some build matrix options which you'd like to do not run.

`dotnet-setup` filter wil always work - it triggered by `if condition`. I agree that looks ugly, but it works explicitly.

I think, you can toying with all of these when you'll need. So, do you want to drop `setup-dotnet` action?
And... 6 tests on macOs failed, what I need to do with it? Because these fails not related to `libgdiplus`

JimBobSquarePants: Thanks for the updates. The tests are failing because the test environment is expecting a different reference codec. We need to update the tests and reference codec registration to cater for the new environment.

https://github.com/SixLabors/ImageSharp/blob/120080b99f3f1be0fc4ecec6906ee683ea999a3d/tests/ImageSharp.Tests/TestUtilities/Tests/TestEnvironmentTests.cs#L84-L93

https://github.com/SixLabors/ImageSharp/blob/120080b99f3f1be0fc4ecec6906ee683ea999a3d/tests/ImageSharp.Tests/TestUtilities/TestEnvironment.Formats.cs#L53-L77

antonfirsov:
I'm not sure whether the current state is acceptable (minimal workflows knowledge here), but it looks like there were only 6 failing tests ATM. 
@AlexNDRmac can you decorate those with `[PlatformSpecific(TestPlatforms.OSX)]` to see if we can get a clean CI run?

AlexNDRmac:
@antonfirsov pls look at CI. 503 code for `myget` when restoring packages. I can't restart CI jobs manually to check if all macOS test passed.

antonfirsov:
@AlexNDRmac re-runned, failed again. URI seems to have an extra dot at the end:  `https://www.myget.org/F/sixlabors/api/v3/index.json.`

JimBobSquarePants:
Weird. That ref hasn't changed. I wonder if we should switch to a nuget.config file instead?

https://github.com/AlexNDRmac/ImageSharp/blob/af/macos-libgdiplus-fix/Directory.Build.props#L119

JimBobSquarePants:
Looks like they're having issues
http://status.myget.org/

JimBobSquarePants:
IT LIIIIVES! Well done @AlexNDRmac 🚀

.github/workflows/build-and-test.yml
```diff
@@ -52,6 +56,19 @@ jobs:
                   git fetch --prune --unshallow
                   git submodule -q update --init --recursive
 
+            # - name: Setup nuget cache
+            #   uses: actions/cache@v2
+            #   id: nuget-cache
+            #   with:
+            #       path: ~/.nuget
+            #       key: ${{ runner.os }}-nuget-${{ hashFiles('**/src/ImageSharp/*.csproj') }}
+            #       restore-keys: ${{ runner.os }}-nuget-
```
antonfirsov, line 65: Is this something we shall try to re-enable? If not then we should probably delete it.
Or if it might be useful in the future, let's document the purpose of the commented out code (and also the reason why is it commented out).

tests/ImageSharp.Tests/Processing/Processors/Transforms/ResizeTests.cs
```diff
@@ -355,6 +355,7 @@ public void Resize_ThrowsForWrappedMemoryImage<TPixel>(TestImageProvider<TPixel>
         }
 
         [Theory]
+        [PlatformSpecific(~TestPlatforms.OSX)]
```
antonfirsov, line 358: @JimBobSquarePants reminder to ourselves that we shall open an issue for the disabled tests when this gets merged.

@AlexNDRmac I suppose you work on Mac. Can you help us by posting the output images saved by these tests on your machine? (`tests/Images/ActualOutput/ResizeTests/ResizeFromSourceRectangle_Rgba32_CalliphoraPartial.png`). Same for the other 2 disabled tests.

Expected images are under `tests/Images/External/ReferenceOutput/ResizeTests/` if interested.

antonfirsov (commented):
Thank you very much for taking the time to work on this! 🥇 Almost done, just a few things.

When I was doing my experiments in #1302, I had 29 failing tests, now I see only 3 disabled ones. Majority was reporting a minor image difference, in stuff like `AffineTransformTests` or `ProjectiveTransformTests`:
https://pipelines.actions.githubusercontent.com/L8wwCFkW4FUCsGOMOcVKecb9uNpo8kI6vyU6NkDyYB5YnVAih3/_apis/pipelines/1/runs/689/signedlogcontent/3?urlExpires=2020-11-01T15%3A28%3A08.8086782Z&urlSigningMethod=HMACV1&urlSignature=nVb2mk%2FFsOwQSPIhd7lDVNYBnRxjNayA9q%2BcYI9Kqc4%3D

Any idea why are they gone now? Have you seen them locally?

.github/workflows/build-and-test.yml
AlexNDRmac: It's disabled temporary. I'll enable it. If you don't mind, I recommend you to use cache for packages, it's seriously speedup build time.

tests/ImageSharp.Tests/Processing/Processors/Transforms/ResizeTests.cs
AlexNDRmac: I propose to use save-action and fail scenario. For example: tests on mac failed, then starts build step with condition 'on-fail' and save any details (pictures, logs... etc) as build artifacts. Next, you can download it for investigation :)
I can add this scenario to workflow, it's very simple and useful.

antonfirsov: Sounds awesome, let's give it a try!
I wonder though what's simplest: here or in a separate PR?

JimBobSquarePants: Do the tests not fail locally or something?

AlexNDRmac: I think, better - is separate PR, because this improvement not related to `libgdiplus`.

On my Mac some tests failed, but on CI such tests - passed :) I use 10.14.16, but CI using 10.15.x.... and on my Mac some tests kills some dotnet sub-processes... It's very interesting case, maybe I can foud whick component have "magic" influence on tests, but experiments requires some time.

AlexNDRmac:
@antonfirsov pls restart CI... smh strange with builds :) macOS passed, but windows not.

antonfirsov:
@AlexNDRmac done, let's see ..

AlexNDRmac:
I see there is no progress with testing on CI... Why?

antonfirsov:
@AlexNDRmac in the last few days there were problems with GitHub, those seem to be fixed now, actions in other PR-s run nicely now, so there must be an issue with the current workflow setup.

image

I may remember it wrong but I think there was a point with a green CI run in this PR. Can we get back to the minimum set of changes that address the topic in the title and enable macOS runs? We can then try further improvements in follow-up PR-s.

antonfirsov:
@AlexNDRmac if you are still interested contributing the reference output exporter action to our builds, feel free to do it in a separate PR. (Providing an explanation how to use it!)

Otherwise ... I guess we should close this one, since the original goal has been completed in #1424. Or is there anything else?

AlexNDRmac:
@antonfirsov if you dont mind, I'd like to add cache to CI (it may speedup builds) and add step to store actual images for failed jobs (it can be useful for you to debug or investigate the reason of fail).

antonfirsov:
@AlexNDRmac cool, and thanks for the interest! I'm open for both, but prefer to see those as 2 separate (new) PR-s. Much easier to review them and reason about the changes.

Issue #809, opened 2019-01-12
JpegDecoder: AVX2 optimize non-YCbCr color spaces


antonfirsov:
- [x] Grayscale
- [x] Ycck
- [x] Cmyk
- [x] RGB

JimBobSquarePants:
Can close this now. Thanks again to @tkp1n !!

Issue #1410, opened 2020-11-02
Non-generic Image.Load should decode Jpeg into Image<Rgb24>


antonfirsov:
Prerequsite: #1354

- Top gain: reduced memory footprint
- There will be a noticeable LoadResizeSave speed regression on 2.1 and Framework (no SIMD for shuffles), but I hope the ROTR / ROTL path will keep it in an acceptable range.
- On 3.1 and .NET 5 the LoadResizeSave  speed regression should be negligible thanks to the SIMD path in #1409
- #1396 combined with omitting [premultipication](https://github.com/SixLabors/ImageSharp/blob/adf56a4d77dc88252709fd7d080471ce2e2d0a3c/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeProcessor%7BTPixel%7D.cs#L173-L174) for `Rgb24` will cancel out these regressions, bringing noticeable speedup on 3.1 and .NET 5. We may want to consider milestoning that issue to 1.1.0

PR #1402, opened 2020-10-23
Add AVX2 Vector4Octet.Pack implementation

JimBobSquarePants:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description
This should add some performance boost until #1242 and others are complete. 

I struggled writing this; any performance tips are most welcome. 

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19041.572 (2004/?/20H1)
Intel Core i7-8650U CPU 1.90GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.100-rc.2.20479.15
  [Host]          : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  AVX             : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  No HwIntrinsics : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  SSE             : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT

Runtime=.NET Core 3.1
```
_Note: No difference with SIMD disabled as we're not using it._

| Method |             Job |                              EnvironmentVariables |     Mean |    Error |   StdDev | Ratio | RatioSD | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |---------------- |-------------------------------------------------- |---------:|---------:|---------:|------:|--------:|------:|------:|------:|----------:|
|   Pack |             AVX |                                             Empty | 10.69 ns | 0.097 ns | 0.075 ns |  1.00 |    0.00 |     - |     - |     - |         - |
|   Pack | No HwIntrinsics | COMPlus_EnableHWIntrinsic=0,COMPlus_FeatureSIMD=0 | 13.70 ns | 0.297 ns | 0.558 ns |  1.27 |    0.07 |     - |     - |     - |         - |
[...]

PR #1242, opened 2020-06-19
Add PackFromRgbPlanes AVX2 vectorised implementation for Rgba32 and Rgba24 pixels

john-h-k:
WIP. A few initial questions raised

* Is there a specific reason for no SSE intrinsics? They would be useful here
* Why are the remainder paths for vectorised pixel operations marked as `NoInlining`? They are often going to be hot code
* There were a few places where there were unnecessary span slices. Although not directly in the scope of this PR, I removed them
* What other pixel types need this impl?

PR #1411, opened 2020-11-03
Vectorize (AVX2) JPEG Color Converter

tkp1n:
This is an initial implementation of the remaining (not yet vectorized) JPEG color converters using AVX(2) runtime intrinsics as well as `Vector<float>`.

I'm opening this as a draft PR to get initial feedback before investing additional time in polishing, tests and benchmarks.

See also #809


src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromCmykSimdAvx2.cs
```diff
@@ -0,0 +1,145 @@
...
+                    Vector256<float> yoHi = Avx.UnpackHigh(y, one);
+
+                    ref Vector256<float> destination = ref Unsafe.Add(ref resultBase, i * 4);
+
+                    destination = Avx.Shuffle(cmLo, yoLo, 0b01_00_01_00);
+                    Unsafe.Add(ref destination, 1) = Avx.Shuffle(cmLo, yoLo, 0b11_10_11_10);
+                    Unsafe.Add(ref destination, 2) = Avx.Shuffle(cmHi, yoHi, 0b01_00_01_00);
+                    Unsafe.Add(ref destination, 3) = Avx.Shuffle(cmHi, yoHi, 0b11_10_11_10);
+                }
+#else
```
antonfirsov, line 98: Didn't have chance to react, but I'm not entirely happy with the practice established in #1402. I would try putting the different paths into different classes (`FromCmykVector8` VS `FromCmykVectorAvx2`).
`FromCmykVectorAvx2` can derive from `FromCmykVector8` for convenience.

antonfirsov (commented):
EPIC!

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromCmykSimdAvx2.cs
JimBobSquarePants: Yeah, that was a stopgap. I’d like to revisit

tkp1n: If we don't shy away from inheritance here, I would suggest a hierarchy like this:
```csharp
abstract class JpegColorConverter
{
    public abstract bool IsAvailable { get; } // new on this class
}

abstract class BasicJpegColorConverter : JpegColorConverter // new base class for all non-vectorized color converters
{
    public override bool IsAvailable => true;
}

abstract class VectorizedJpegColorConverter : JpegColorConverter // new base class for all vectorized color converters
{
    private readonly int vectorSize;

    protected VectorizedJpegColorConverter(JpegColorSpace colorSpace, int precision, int vectorSize)
        : base(colorSpace, precision)
    {
        this.vectorSize = vectorSize;
    }

    public override void ConvertToRgba(in ComponentValues values, Span<Vector4> result)
    {
        int remainder = result.Length % vectorSize;
        int simdCount = result.Length - remainder;
        if (simdCount > 0)
        {
            ConvertCoreVectorized(values.Slice(0, simdCount), result.Slice(0, simdCount));
        }

        ConvertCore(values.Slice(simdCount, remainder), result.Slice(simdCount, remainder));
    }

    protected abstract void ConvertCoreVectorized(in ComponentValues values, Span<Vector4> result);

    protected abstract void ConvertCore(in ComponentValues values, Span<Vector4> result);
}

abstract class Avx2JpegColorConverter : VectorizedJpegColorConverter // new base class for all AVX2-based converters
{
    protected Avx2JpegColorConverter(JpegColorSpace colorSpace, int precision)
        : base(colorSpace, precision, 8)
    {
    }

    public override bool IsAvailable
    {
        get
        {
            return Avx2.IsSupported;
#else
            return false;
#endif
        }
    }
}

// another base class for Vector<float>-based converters
```

We could then instantiate all converters initially into a static array and determine which one to use based on `ColorSpace`, `Precision` and `IsAvailable`. This would address your point and maximize code re-use at the cost of some virtual method calls.

Let me know what you think. I didn't want to refactor to much for the initial draft-PR and just add the vectorized logic into the existing code structure.

antonfirsov: @tkp1n I like the concept, go for it! Hopefully the runtime costs of the virtual dispatches are be still negligible.

tkp1n: I refactored as discussed above, PTAL...

Next up: Benchmarks 🚀

tkp1n:
The numbers are in 💯 Looks like some nice improvements across the board.

👍 YCCK is the big winner.
👎 Vectorizing the gray scale conversion does not seem profitable. 

``` ini

Intel Core i7-9750H CPU 2.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.100-rc.2.20479.15
  [Host]     : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  Job-ZMVTJV : .NET Framework 4.8 (4.8.4250.0), X64 RyuJIT
  Job-VQWAVA : .NET Core 2.1.23 (CoreCLR 4.6.29321.03, CoreFX 4.6.29321.01), X64 RyuJIT
  Job-ETAOPA : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT

```

## CMYK

|         Method |        Job |       Runtime |     Mean |     Error |   StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |----------- |-------------- |---------:|----------:|---------:|------:|--------:|-------:|------:|------:|----------:|
|         Scalar | Job-ZMVTJV |    .NET 4.7.2 | 790.7 ns | 373.12 ns | 20.45 ns |  1.00 |    0.00 | 0.0048 |     - |     - |      32 B |
|    SimdVector8 | Job-ZMVTJV |    .NET 4.7.2 | 441.6 ns |  69.54 ns |  3.81 ns |  0.56 |    0.01 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ZMVTJV |    .NET 4.7.2 |       NA |        NA |       NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |          |           |          |       |         |        |       |       |           |
|         Scalar | Job-VQWAVA | .NET Core 2.1 | 376.8 ns | 125.78 ns |  6.89 ns |  1.00 |    0.00 | 0.0048 |     - |     - |      32 B |
|    SimdVector8 | Job-VQWAVA | .NET Core 2.1 | 320.1 ns |  18.45 ns |  1.01 ns |  0.85 |    0.01 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-VQWAVA | .NET Core 2.1 |       NA |        NA |       NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |          |           |          |       |         |        |       |       |           |
|         Scalar | Job-ETAOPA | .NET Core 3.1 | 395.4 ns |  89.28 ns |  4.89 ns |  1.00 |    0.00 | 0.0048 |     - |     - |      32 B |
|    SimdVector8 | Job-ETAOPA | .NET Core 3.1 | 303.2 ns |  28.49 ns |  1.56 ns |  0.77 |    0.01 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ETAOPA | .NET Core 3.1 | 195.4 ns |  18.34 ns |  1.01 ns |  0.49 |    0.01 | 0.0062 |     - |     - |      40 B |

## YCCK

|         Method |        Job |       Runtime |       Mean |       Error |    StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |----------- |-------------- |-----------:|------------:|----------:|------:|--------:|-------:|------:|------:|----------:|
|         Scalar | Job-ZMVTJV |    .NET 4.7.2 | 6,590.0 ns | 2,572.12 ns | 140.99 ns |  1.00 |    0.00 |      - |     - |     - |      32 B |
|    SimdVector8 | Job-ZMVTJV |    .NET 4.7.2 |   484.9 ns |   177.17 ns |   9.71 ns |  0.07 |    0.00 | 0.0057 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ZMVTJV |    .NET 4.7.2 |         NA |          NA |        NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |            |             |           |       |         |        |       |       |           |
|         Scalar | Job-VQWAVA | .NET Core 2.1 | 4,880.5 ns |   296.83 ns |  16.27 ns |  1.00 |    0.00 |      - |     - |     - |      32 B |
|    SimdVector8 | Job-VQWAVA | .NET Core 2.1 |   380.2 ns |   162.87 ns |   8.93 ns |  0.08 |    0.00 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-VQWAVA | .NET Core 2.1 |         NA |          NA |        NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |            |             |           |       |         |        |       |       |           |
|         Scalar | Job-ETAOPA | .NET Core 3.1 | 3,655.0 ns |   966.89 ns |  53.00 ns |  1.00 |    0.00 | 0.0038 |     - |     - |      32 B |
|    SimdVector8 | Job-ETAOPA | .NET Core 3.1 |   346.4 ns |    21.49 ns |   1.18 ns |  0.09 |    0.00 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ETAOPA | .NET Core 3.1 |   229.5 ns |    55.78 ns |   3.06 ns |  0.06 |    0.00 | 0.0062 |     - |     - |      40 B |

## RGB

|         Method |        Job |       Runtime |     Mean |     Error |   StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |----------- |-------------- |---------:|----------:|---------:|------:|--------:|-------:|------:|------:|----------:|
|         Scalar | Job-ZMVTJV |    .NET 4.7.2 | 507.3 ns |  47.95 ns |  2.63 ns |  1.00 |    0.00 | 0.0048 |     - |     - |      32 B |
|    SimdVector8 | Job-ZMVTJV |    .NET 4.7.2 | 386.7 ns | 217.45 ns | 11.92 ns |  0.76 |    0.02 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ZMVTJV |    .NET 4.7.2 |       NA |        NA |       NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |          |           |          |       |         |        |       |       |           |
|         Scalar | Job-VQWAVA | .NET Core 2.1 | 298.8 ns |  43.56 ns |  2.39 ns |  1.00 |    0.00 | 0.0048 |     - |     - |      32 B |
|    SimdVector8 | Job-VQWAVA | .NET Core 2.1 | 297.8 ns |  53.72 ns |  2.94 ns |  1.00 |    0.01 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-VQWAVA | .NET Core 2.1 |       NA |        NA |       NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |          |           |          |       |         |        |       |       |           |
|         Scalar | Job-ETAOPA | .NET Core 3.1 | 293.0 ns |  28.07 ns |  1.54 ns |  1.00 |    0.00 | 0.0048 |     - |     - |      32 B |
|    SimdVector8 | Job-ETAOPA | .NET Core 3.1 | 287.9 ns |  22.87 ns |  1.25 ns |  0.98 |    0.01 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ETAOPA | .NET Core 3.1 | 178.3 ns | 155.38 ns |  8.52 ns |  0.61 |    0.03 | 0.0062 |     - |     - |      40 B |

## Grayscale

|         Method |        Job |       Runtime |     Mean |    Error |  StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |----------- |-------------- |---------:|---------:|--------:|------:|--------:|-------:|------:|------:|----------:|
|         Scalar | Job-ZMVTJV |    .NET 4.7.2 | 212.1 ns | 70.83 ns | 3.88 ns |  1.00 |    0.00 | 0.0050 |     - |     - |      32 B |
|    SimdVector8 | Job-ZMVTJV |    .NET 4.7.2 | 321.9 ns | 32.69 ns | 1.79 ns |  1.52 |    0.03 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ZMVTJV |    .NET 4.7.2 |       NA |       NA |      NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |          |          |         |       |         |        |       |       |           |
|         Scalar | Job-VQWAVA | .NET Core 2.1 | 215.7 ns | 81.12 ns | 4.45 ns |  1.00 |    0.00 | 0.0050 |     - |     - |      32 B |
|    SimdVector8 | Job-VQWAVA | .NET Core 2.1 | 258.6 ns | 68.91 ns | 3.78 ns |  1.20 |    0.02 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-VQWAVA | .NET Core 2.1 |       NA |       NA |      NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |          |          |         |       |         |        |       |       |           |
|         Scalar | Job-ETAOPA | .NET Core 3.1 | 145.3 ns | 28.68 ns | 1.57 ns |  1.00 |    0.00 | 0.0050 |     - |     - |      32 B |
|    SimdVector8 | Job-ETAOPA | .NET Core 3.1 | 249.5 ns |  9.49 ns | 0.52 ns |  1.72 |    0.02 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ETAOPA | .NET Core 3.1 | 141.3 ns |  2.46 ns | 0.13 ns |  0.97 |    0.01 | 0.0062 |     - |     - |      40 B |

## YCbCr

|         Method |        Job |       Runtime |       Mean |       Error |    StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |----------- |-------------- |-----------:|------------:|----------:|------:|--------:|-------:|------:|------:|----------:|
|         Scalar | Job-ZMVTJV |    .NET 4.7.2 | 6,087.3 ns | 1,964.34 ns | 107.67 ns |  9.08 |    0.24 |      - |     - |     - |      32 B |
|     SimdVector | Job-ZMVTJV |    .NET 4.7.2 |   670.4 ns |   121.26 ns |   6.65 ns |  1.00 |    0.00 | 0.0057 |     - |     - |      40 B |
|    SimdVector8 | Job-ZMVTJV |    .NET 4.7.2 |   430.5 ns |   216.63 ns |  11.87 ns |  0.64 |    0.02 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ZMVTJV |    .NET 4.7.2 |         NA |          NA |        NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |            |             |           |       |         |        |       |       |           |
|         Scalar | Job-VQWAVA | .NET Core 2.1 | 4,533.9 ns | 1,935.28 ns | 106.08 ns |  7.83 |    0.22 |      - |     - |     - |      32 B |
|     SimdVector | Job-VQWAVA | .NET Core 2.1 |   579.2 ns |    79.18 ns |   4.34 ns |  1.00 |    0.00 | 0.0057 |     - |     - |      40 B |
|    SimdVector8 | Job-VQWAVA | .NET Core 2.1 |   338.5 ns |    45.89 ns |   2.52 ns |  0.58 |    0.01 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-VQWAVA | .NET Core 2.1 |         NA |          NA |        NA |     ? |       ? |      - |     - |     - |         - |
|                |            |               |            |             |           |       |         |        |       |       |           |
|         Scalar | Job-ETAOPA | .NET Core 3.1 | 3,442.3 ns |   465.54 ns |  25.52 ns |  7.68 |    0.06 | 0.0038 |     - |     - |      32 B |
|     SimdVector | Job-ETAOPA | .NET Core 3.1 |   448.4 ns |   116.31 ns |   6.38 ns |  1.00 |    0.00 | 0.0062 |     - |     - |      40 B |
|    SimdVector8 | Job-ETAOPA | .NET Core 3.1 |   323.5 ns |    81.56 ns |   4.47 ns |  0.72 |    0.02 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2 | Job-ETAOPA | .NET Core 3.1 |   204.0 ns |    62.41 ns |   3.42 ns |  0.45 |    0.01 | 0.0062 |     - |     - |      40 B |

JimBobSquarePants:
Looking good @tkp1n !! I'll do a thorough review as soon as possible!

antonfirsov:
So guys, here's my offer:

I want to put together a PR for #1410 that also alters the color converters *to pack into a 3-channel RGB float buffer*, but I want to "go lazy" and spare solving AVX2 shuffling riddles, because every hour I'd spend there would delay SixLabors/ImageSharp.Drawing#96 further.

So the question is: @tkp1n do you have any willingness and chance to spend a bit more time here, and finish such a refactor with @JimBobSquarePants 's help?

If we can make it, this trick would deliver superior performance compared to any alternative, since we could spare a completely unnecessary alpha padding.

Let me know what you think. If you agree we'd figure out the "what PR follows what" details. Cheers!

JimBobSquarePants:
@antonfirsov Isn't this essentially what https://github.com/SixLabors/ImageSharp/pull/1242 should be doing?

antonfirsov:
That corresponds to #1121 ("Optimized pipeline"), which is a more expensive refactor, probably not realistic for 1.1.

The late night idea was that after all the "low hanging" refactors that happened, with a single extra step we could profit a lot now, and It's probably easier to do while this shuffling thing is fresh in your heads. (This would remove the need of implementing a one-step `Vector4 -> Rgb24` converter, and bring more wins.)

However I'm not sure now, since what I'm asking might be tricky and time-consuming because the way 3-component RGB-s overlap `Vector256<float>`-s. If it's too much, let's stop now and continue with #1121 later.

JimBobSquarePants:
I don't think it's _that_ much effort since the existing converters are all doing most of the work already. It's a case of moving most of it and filling out some gaps using `Vector3` instead of `Vector4` for the fallback packing and some offsetting for the HW Intrinsics version.

I'm happy to look at it after this is merged for simplicities sake.

tkp1n:
Starting next Monday, I'll be away from "work" like this for 4 weeks with very limited access to the web. 😢 
While I can likely address PR feedback, I won't be able to contribute to any larger-scale refactoring in this area.

I've removed the "[WIP]" and marked this PR as ready for review.  Feel free to let me know what you need to be fixed before this can be merged.

If the timing of this PR doesn't suit you due to other planned or ongoing refactoring, feel free to set this one aside, and we'll reconsider it when the timing is right.

tkp1n:
Starting next Monday, I'll be away from "work" like this for 4 weeks with very limited access to the web. 😢 
While I can likely address PR feedback, I won't be able to contribute to any larger-scale refactoring in this area.

I've removed the "[WIP]" and marked this PR as ready for review.  Feel free to let me know what you need to be fixed before this can be merged.

If the timing of this PR doesn't suit you due to other planned or ongoing refactoring, feel free to set this one aside, and we'll reconsider it when the timing is right.

antonfirsov:
@tkp1n no worries, enjoy your vacation, and thanks for the contribution! Your PR should be the first to be merged, we'll review it ASAP.

JimBobSquarePants:
I can't see any issues with this. It all looks fantastic and the performance is consistent with master for the existing optimized conversion. 👍 🚀 

If we drop `FromGrayscaleVector8` I'd be very happy for this to go in.

JimBobSquarePants (approved):

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector.cs
```diff
@@ -5,36 +5,22 @@
...
 
-            public override void ConvertToRgba(in ComponentValues values, Span<Vector4> result)
-            {
-                int remainder = result.Length % 8;
-                int simdCount = result.Length - remainder;
-                if (simdCount > 0)
-                {
-                    ConvertCore(values.Slice(0, simdCount), result.Slice(0, simdCount), this.MaximumValue, this.HalfValue);
-                }
+            protected override bool IsAvailable => true;
```
antonfirsov, line 21:
```suggestion
            protected override bool IsAvailable => Vector.IsHardwareAccelerated && Vector<float>.Count == 4;
```

This is essentially an SSE path based on `Vector4`. The converter is not safe to run if the above conditions are not met, because of the `RoundAndDownscalePreVector8` call.

@tannergooding is there an environment variable to enforce a `(Vector.IsHardwareAccelerated && Vector<float>.Count == 4) == true` configuration for testing?

If not, I would suggest to remove this converter.

tkp1n:
> I can't see any issues with this. It all looks fantastic and the performance is consistent with master for the existing optimized conversion. 👍 🚀

Thanks a lot 😃 

> If we drop `FromGrayscaleVector8` I'd be very happy for this to go in.

Done, along with some final polishing of the benchmarks.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector.cs
```diff
@@ -5,36 +5,22 @@
...
 using System.Runtime.InteropServices;
-
 using SixLabors.ImageSharp.Tuples;
 
 namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder.ColorConverters
 {
     internal abstract partial class JpegColorConverter
     {
-        internal sealed class FromYCbCrSimd : JpegColorConverter
+        internal sealed class FromYCbCrVector : VectorizedJpegColorConverter
```
antonfirsov, line 14:
```suggestion
        internal sealed class FromYCbCrVector4 : VectorizedJpegColorConverter
```

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromCmykAvx2.cs
```diff
@@ -0,0 +1,81 @@
...
+        internal sealed class FromCmykAvx2 : Avx2JpegColorConverter
+        {
+            public FromCmykAvx2(int precision)
+                : base(JpegColorSpace.Cmyk, precision)
+            {
+            }
+
+            protected override void ConvertCoreVectorized(in ComponentValues values, Span<Vector4> result)
+            {
+#if SUPPORTS_RUNTIME_INTRINSICS
```
antonfirsov, line 27: Can we hide the whole file behind the condition instead, and add this guard to `GetYCbCrConverters`:

```C#
yield return new FromYCbCrAvx2(precision);
#endif
```

Same for the other color spaces.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromGrayScaleVector8.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Numerics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using SixLabors.ImageSharp.Tuples;
+
+namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder.ColorConverters
+{
+    internal abstract partial class JpegColorConverter
+    {
+        internal sealed class FromGrayscaleVector8 : Vector8JpegColorConverter
```
antonfirsov: Since it performs worse than the `Vector4` variant, let's remove this one.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromGrayScaleAvx2.cs
```diff
@@ -0,0 +1,63 @@
...
+                    Vector256<float> g = Avx2.PermuteVar8x32(Unsafe.Add(ref gBase, i), vcontrol);
+
+                    g = Avx.Multiply(g, scale);
+
+                    ref Vector256<float> destination = ref Unsafe.Add(ref resultBase, i * 4);
+
+                    destination = Avx.Blend(Avx.Permute(g, 0b00_00_00_00), one, 0b1000_1000);
+                    Unsafe.Add(ref destination, 1) = Avx.Blend(Avx.Permute(g, 0b01_01_01_01), one, 0b1000_1000);
+                    Unsafe.Add(ref destination, 2) = Avx.Blend(Avx.Permute(g, 0b10_10_10_10), one, 0b1000_1000);
+                    Unsafe.Add(ref destination, 3) = Avx.Blend(Avx.Permute(g, 0b11_11_11_11), one, 0b1000_1000);
```
antonfirsov, line 54: If you have any idea what should we try to improve this path, let's add a comment. If you think it's unfixable, consider removing this variant, since it doesn't add noticable perf in comparison to the "Basic" (`Vector4`) variant.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrAvx2.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                    // r = y + (1.402F * cr);
+                    // g = y - (0.344136F * cb) - (0.714136F * cr);
+                    // b = y + (1.772F * cb);
+                    // Adding & multiplying 8 elements at one time:
+                    Vector256<float> r = HwIntrinsics.MultiplyAdd(y, cr, rCrMult);
+                    Vector256<float> g = HwIntrinsics.MultiplyAdd(HwIntrinsics.MultiplyAdd(y, cb, gCbMult), cr, gCrMult);
+                    Vector256<float> b = HwIntrinsics.MultiplyAdd(y, cb, bCbMult);
+
+                    // TODO: We should be saving to RGBA not Vector4
```
antonfirsov, line 75: @JimBobSquarePants we should either do this or #1121, but let's not keep around contradicting plans in our notes!

antonfirsov (commented):
One major concern about `FromYCbCrVector`, and a few simplification ideas.

Test coverage seems to prove that everything works 100% fine, however it might worth for @JimBobSquarePants to do a quick sanity check on the shuffling bits.

antonfirsov (changes requested):
https://github.com/SixLabors/ImageSharp/pull/1411#pullrequestreview-525302469

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromGrayScaleVector8.cs
antonfirsov: I see this has been addressed in the meanwhile.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector.cs
tannergooding: If you set `COMPlus_EnableAVX2=0` it will force `Vector<T>` to be 16-bytes. We don't have a switch that explicitly controls the size of `Vector<T>` today.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
```diff
@@ -41,54 +42,40 @@ public JpegColorConverterTests(ITestOutputHelper output)
...
                 3,
                 inputBufferLength,
                 resultBufferLength,
                 seed);
         }
 
-        private static void ValidateYCbCr(in JpegColorConverter.ComponentValues values, Vector4[] result, int i)
+        [Theory]
+        [MemberData(nameof(CommonConversionData))]
+        public void FromYCbCrVector(int inputBufferLength, int resultBufferLength, int seed)
```
antonfirsov, line 57: Doesn't have to happen in this PR, but for 1.1 we need a variant of this test that works with:

```C#
                RunTest,
                args,
                HwIntrinsics.DisableAVX2);
```

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector.cs
antonfirsov: @tannergooding thanks!

@JimBobSquarePants https://github.com/SixLabors/ImageSharp/pull/1411#discussion_r518919719

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrAvx2.cs
JimBobSquarePants: I wrote that a while back and forgot about it. I still believe there's merit in cutting out the `FromVector4` call though. I need to make some diagrams and think things through.

antonfirsov: > I still believe there's merit in cutting out the `FromVector4`

I think #1121 is the most efficient way to do that with the float pipelines.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromCmykAvx2.cs
JimBobSquarePants: If we add the conditionals to the files we have to refactor the tests as I don't want to pepper the tests with conditionals and we don't have cross platform remote executor available to test everything on Core2.1 Win yet.

Adding to the yield though is fine.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
JimBobSquarePants: What I would like to do is a single test which I can then pass through the runner. That will test all the versions then.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrAvx2.cs
JimBobSquarePants: Sorry I wasn't clear I don't mean change the pipeline what I mean is to pack triplets so we can avoid `Vector4` => `Rgba32` => `Rgb24`. That comment should have said RGB.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromGrayScaleAvx2.cs
JimBobSquarePants: @saucecontrol do you have ideas here? Those permutes look expensive.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector.cs
JimBobSquarePants: I've done this manually

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
antonfirsov: Can you parameterize it to also exercise the Vector8 the Vector4 and the scalar path?

JimBobSquarePants: Yep, since the call change the environmental paths like `COMPlus_EnableAVX2=0` we can do the following.

```
                RunTest,
                args,
                HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX | HwIntrinsics.DisableSSE );
```

antonfirsov: How to get `Vector.IsHardwareAccelerated == false`?

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromGrayScaleAvx2.cs
saucecontrol: These are in-lane permutes (`vpermilps`), so they're the same cost as a normal shuffle @ 1 cycle latency.  The encoding is 1 byte shorter for `vshufps`, which makes that instruction kind of pointless, but it's not bad.  See: https://stackoverflow.com/a/54170092/4926931

The cross-lane permutes above (`Avx2.PermuteVar8x32`) are higher cost, though.  This might be a case where an SSE implementation is faster, but I haven't tested that one specifically.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
saucecontrol: `COMPlus_FeatureSIMD=0` should do it

JimBobSquarePants: Yep that's it. We have `HwIntrinsics.DisableSIMD` for that.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromGrayScaleAvx2.cs
JimBobSquarePants: Thanks!

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.cs
```diff
@@ -89,10 +81,79 @@ public static JpegColorConverter GetConverter(JpegColorSpace colorSpace, int pre
...
+            var converters = new List<JpegColorConverter>();
+
+            // 8-bit converters
+            converters.AddRange(GetYCbCrConverters(8));
+            converters.AddRange(GetYccKConverters(8));
+            converters.AddRange(GetCmykConverters(8));
+            converters.AddRange(GetGrayScaleConverters(8));
+            converters.AddRange(GetRgbConverters(8));
+
+            // 8-bit converters
```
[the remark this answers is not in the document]
JimBobSquarePants: Done manually

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromCmykAvx2.cs
antonfirsov: Ok lets skip it for now. Comment could be useful.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromGrayScaleAvx2.cs
saucecontrol: @JimBobSquarePants If you want to get a better idea how the ports + latency and throughput numbers for the various SIMD instructions interact, have a read through chapter 9 of [Agner's optimization guide](https://www.agner.org/optimize/optimizing_assembly.pdf).  Once you have a handle on that, you can model things pretty quickly to get a sense of how they'll perform without necessarily having to write and benchmark them.  The best data for the ports/latency/thoughput currently is on uops.info or in Agner's instruction table PDFs.

I did another crappy diagram™ to show how this one models out with the instruction latency and parallelism:

```
cycle ______input_______
1     vmovups
2     ...
3     ...
4        => vpermd
5           ...
6           ...
7              => vmulps
8                 ...
9                 ...
10                ...       ________out0_______    ________out1_______    ________out2_______    _______out3_______
11                       => vpermilps                     
12                             => vblendps         vpermilps
13                                   => vmovups       => vblendps         vpermilps
14                                      ...                 => vmovups       => vblendps         vpermilps
15                                      ...                    ...                 => vmovups      => vblendps
16                                                             ...                    ...                => vmovups
17                                                                                    ...                    ...
18                                                                                                           ...
```

So ignoring loop overhead, and assuming 32-byte aligned data and perfect cache usage, the SIMD code models out to 18 cycles per iteration for 8 pixels.  The key thing here is that it takes 10 cycles to prepare the data before the output shuffling can start, and then the 4 outputs parallelize nicely, doing the rest of the work in 8 cycles.

If you were to take 2 input vectors at a time, you'd have the same 10 cycles latency before your first shuffle, but the second vector is ready 1 cycle after that, and the rest can be paralleled in the same way.  You end up with 10 cycles for the input prep and 12 for the shuffle/blend/store, making it 22 cycles for 16 pixels.  You could go even further and get 26 cycles for 24 pixels by unrolling the input by 3, or 30 cycles for 32 pixels, etc.  At some point, you run out of registers, your loop body gets too big for the uop cache, or you simply have chunks so big your remainder gets out of hand, so there's a balance to be found somewhere.

If it were me, I'd unroll by at least 2, swap the order of the multiply and permute8x32, and change the `vpermilps` to `vshufps` to minimize the loop body code size.  That should give enough gain to make the AVX2 implementation worthwhile.

tkp1n: So I changed the input size of the benchmarks from 128 to 512 and re-ran it with the following changes:

* `SimdVectorAvx2` unchanged
* `SimdVectorAvx2v2` swapped multiply and permute and changed `vpermilps` to `vshufps`
* `SimdVectorAvx2By2` All suggested changes (`SimdVectorAvx2v2` + unrolled by 2)

|            Method |        Job |       Runtime |     Mean |     Error |   StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------ |----------- |-------------- |---------:|----------:|---------:|------:|--------:|-------:|------:|------:|----------:|
|            Scalar | Job-KDKLQO | .NET Core 3.1 | 400.3 ns | 107.51 ns |  5.89 ns |  1.00 |    0.00 | 0.0048 |     - |     - |      32 B |
|    SimdVectorAvx2 | Job-KDKLQO | .NET Core 3.1 | 240.7 ns |  36.36 ns |  1.99 ns |  0.60 |    0.01 | 0.0062 |     - |     - |      40 B |
|  SimdVectorAvx2v2 | Job-KDKLQO | .NET Core 3.1 | 233.2 ns |   4.93 ns |  0.27 ns |  0.58 |    0.01 | 0.0062 |     - |     - |      40 B |
| SimdVectorAvx2By2 | Job-KDKLQO | .NET Core 3.1 | 223.9 ns |  55.07 ns |  3.02 ns |  0.56 |    0.00 | 0.0062 |     - |     - |      40 B |


--> I'll apply your suggested improvements but ignore unrolling for now, as the gains are relatively small vs. code size increase.

src/ImageSharp/Common/Helpers/SimdUtils.cs
```diff
@@ -25,6 +25,25 @@ internal static partial class SimdUtils
         public static bool HasVector8 { get; } =
             Vector.IsHardwareAccelerated && Vector<float>.Count == 8 && Vector<int>.Count == 8;
 
+        /// <summary>
+        /// Gets a value indicating whether <see cref="Vector{T}"/> code is being JIT-ed to SSE instructions
+        /// where float and integer registers are of size 128 byte.
+        /// </summary>
+        public static bool HasVector4 { get; } =
```
antonfirsov, line 32: I find this name misleading, one may think it has to do something with `Vector4`.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector4.cs
```diff
@@ -87,7 +74,7 @@ internal static void ConvertCore(in ComponentValues values, Span<Vector4> result
                     tmp.MultiplyInplace(1.772F);
                     b.AddInplace(ref tmp);
 
-                    if (Vector<float>.Count == 4)
+                    if (SimdUtils.HasVector4)
```
antonfirsov, line 77: For perf, we should remove this branch now and use `RoundAndDownscalePreVector8` only. `IsAvailable` has already done the capability check by this time.

src/ImageSharp/Common/Helpers/SimdUtils.cs
tkp1n: Any suggestions?

antonfirsov: I would just inline it, there would be only one usage, if we'd address my other suggestion.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector4.cs
tkp1n: Good catch, I removed the branch

antonfirsov (approved):
LGTM

Will open an issue later for the test gap on the `FromYCbCrVector4 ` test gap.

tkp1n:
Thank you @JimBobSquarePants and @antonfirsov for your inputs and hands-on time to get this merged so quickly 🚀

Issue #1396, opened 2020-10-20
Introduce PixelTypeInfo.HasAlpha


antonfirsov:
## Motivation

By introducing a property to observe whether a given pixel type has an alpha, we can then use it in our processors to omit unnecessary alpha premultiplication. Might be also interesting for users working with `Image.Identify`.

We need to make sure it's value is properly filled:
- In decoders with `Identify` implementations  (PNG, BMP)
- For all built-in pixel formats (`IPixel`-s)

## API alternatives

In https://github.com/SixLabors/ImageSharp/discussions/1394#discussioncomment-106875 I forgot that extending `IPixel` is actually a breaking change.  Non-breaking variant:

```C#
internal enum PixelAlphaRepresentation
{
    None,
    Associated,
    Unassociated
}

public class PixelTypeInfo
{
     // 'null' means unknown/not implemented
     public PixelAlphaRepresentation? AlphaRepresentation { get; }    
}

public PixelOperations<TPixel>
{
     // Return AlphaRepresentation == null by default, but for important pixel types, introduce proper overrides with T4:
     public virtual PixelTypeInfo GetPixelTypeInfo();
}
```

JimBobSquarePants:
I can see us adding `ChannelCount` to this type at some point so I guess that would be nullable also.

antonfirsov:
@JimBobSquarePants so you prefer the non-breaking variant? Honestly, now it looks like a safer way to go to me.
Also allows us to extend codecs incrementally to detect `PixelTypeInfo` details.

JimBobSquarePants:
Yeah, it's a safe clever approach. Big fan!

saucecontrol:
I might be misreading the purpose of this, but did you consider support for pixel formats that are already premultiplied?  [TIFF](https://www.awaresystems.be/imaging/tiff/tifftags/extrasamples.html) and [DirectX](https://docs.microsoft.com/en-us/windows/win32/api/dcommon/ne-dcommon-d2d1_alpha_mode) (among others) support alpha in both forms.

I recall we had a chat about this somewhere, although I've since lost track of it... I set up a generic pixel format descriptor type [here](https://github.com/saucecontrol/PhotoSauce/blob/master/src/MagicScaler/Core/PixelFormats.cs) which borrows heavily from WIC's pluggable pixel format model and describes alpha using an enum if you're looking for inspiration.

JimBobSquarePants:
[...]

PR #1420, opened 2020-11-08
Add new PixelAlphaRepresentation property and implement for TPixel types

JimBobSquarePants:
### Prerequisites

### Description
The first in several PRs targetting #1396 

- Adds a new nullable `PixelAlphaRepresentation` property to the `PixelTypeInfo` class which describes the pixel alpha behavior for the image pixel format.
- Implements the property values for all `TPixel` types. 
- Adds tests for each implementation.

I've automated the test generation using T4 as there's a lot of repetition. 

Follow up PRs will wire up image formats and optimize methods based upon the new information.


src/ImageSharp/ImageSharp.csproj
```diff
@@ -62,62 +62,102 @@
...
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>HalfVector2.PixelOperations.Generated.tt</DependentUpon>
+    </Compile>
+    <Compile Update="PixelFormats\PixelImplementations\PixelOperations\Generated\HalfVector4.PixelOperations.Generated.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>HalfVector4.PixelOperations.Generated.tt</DependentUpon>
+    </Compile> 
+    <Compile Update="PixelFormats\PixelImplementations\PixelOperations\Generated\L8.PixelOperations.Generated.cs">
```
antonfirsov, line 120: I don't see these files.

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Generated/_Common.ttinclude
```diff
@@ -8,19 +8,85 @@
...
+        "Bgra32",
+        "Rgb24",
+        "Bgr24"
+    };
 
-    // Types with Rgba32-combatible to/from Vector4 conversion
-    static readonly string[] Rgba32CompatibleTypes = { "Argb32", "Bgra32", "Rgb24", "Bgr24" };
+    void GeneratePixelTypeInfo(string pixelType)
+    {
+        string alpha = "PixelAlphaRepresentation.None";
```
antonfirsov, line 76: When we add a new pixel type with a `.tt`, we should be careful and include it in `UnassociatedAlphaPixelTypes` when it has an alpha, otherwise `PixelTypeInfo.AlphaRepresentation` will be incorrectly report `None`.

antonfirsov (commented):
I think it's not clear from the changes if we intended to control pixel types like `HalfVector4` from T4 or from regular `.cs`.

The product changes look good otherwise, but also want to have a look at tests.

src/ImageSharp/ImageSharp.csproj
JimBobSquarePants: Weird... Those includes shouldn't be there. Will delete.

src/ImageSharp/Formats/PixelTypeInfo.cs
```diff
@@ -16,18 +15,32 @@ public class PixelTypeInfo
         /// Initializes a new instance of the <see cref="PixelTypeInfo"/> class.
         /// </summary>
         /// <param name="bitsPerPixel">Color depth, in number of bits per pixel.</param>
-        internal PixelTypeInfo(int bitsPerPixel)
+        /// <param name="alpha">Tthe pixel alpha transparency behavior.</param>
+        internal PixelTypeInfo(int bitsPerPixel, PixelAlphaRepresentation? alpha = null)
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah... I'm getting more chill in my old age.

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Generated/_Common.ttinclude
JimBobSquarePants: Yeah... My thinking was since this was only used when we are adding via `.tt` we have to check the various lists anyway for correctness.

JimBobSquarePants:
>Unless I'm missing something, it's not clear from the PR if we intended to control `PixelTypeInfo` for pixel types like `HalfVector4` from `T4` or from regular `.cs`.

Normally `cs`. That's what I've done with all the existing types that are not already using `T4` for the pixel operations. I added the codegen to the original ones because I didn't want to create multiple classes but I've just pushed a change that separates out the generation so we can't accidently assign `None`

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Generated/_Common.ttinclude
JimBobSquarePants: I've moved the code to an explicit declaration

JimBobSquarePants:
How are we doing here? I'd like to merge and move on.

antonfirsov:
@JimBobSquarePants I can take a final look tomorrow (sill want to check the tests), but if its too late feel free to merge. You can also base next PR on current state since the product changes look good.

JimBobSquarePants:
Yeah, no worries. This PR adds a lot of (admittedly autogenerated) tests so well worth a review.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
```diff
@@ -0,0 +1,119 @@
...
+            {
+                var bits = <#=pixelType#>.PixelOperations.Instance.GetPixelTypeInfo().BitsPerPixel;
+                Assert.Equal(Unsafe.SizeOf<<#=pixelType#>>() * 8, bits);
+            }
+
+            [Fact]
+            public void PixelTypeInfoHasCorrectAlphaRepresentation()
+            {
+                var alphaRepresentation = <#=pixelType#>.PixelOperations.Instance.GetPixelTypeInfo().AlphaRepresentation;
+                Assert.Equal(<#=alpha#>, alphaRepresentation);
```
antonfirsov, line 96: I'm concerned whether this type of testing is valuable here.

I would rather try something like this in the base type:

```C#
[Fact]
public void PixelAlphaRepresentation_DefinesPresenceOfAlphaChannel()
{
    TPixel pixel = default;
    pixel.FromRgba32(new Rgba32(0, 0, 0, 123));

    Rgba32 dest;
    pixel.ToRgba32(ref dest);

    bool hasAlpha = GetPixelTypeInfo().AlphaRepresentation != PixelAlphaRepresentation.None;
    byte exectedAlpha = hasAlpha ? 123 : 255;    

    Assert.Equal(expectedAlpha, dest.A);
}
```

src/ImageSharp/PixelFormats/PixelAlphaRepresentation.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Indicates that the transparency behavior is premultiplied.
+        /// Each color is first scaled by the alpha value. The alpha value itself is the same
+        /// in both straight and premultiplied alpha. Typically, no color channel value is
+        /// greater than the alpha channel value.
+        /// If a color channel value in a premultiplied format is greater than the alpha
+        /// channel, the standard source-over blending math results in an additive blend.
+        /// </summary>
+        Associated,
```
antonfirsov, line 24: Unless we know image formats where we can report this by `Identify` this would be a completely unimplemented virtual feature for now.  I know it's me who proposed the API, but now I'm concerned exposing as-is, feels a bit misleading and confusing.

How about replacing the summary with a text *"Reserved for future use"*, and moving the docs into a comment?

antonfirsov (commented):
1. Concerns about exposing an unimplemented feature.
2. I think in tests we should use T4 only when it's absolutely necessary, and whenever it's possible to implement a test in the base type, we should push the code up.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
JimBobSquarePants: All I'm testing here is the correct assignment of the value. Any behavior based upon that value is the responsibility of the caller.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/PixelOperationsTests.Specialized.Generated.tt
```diff
@@ -0,0 +1,11 @@
...
+<#@ output extension=".cs" #>
+
+namespace SixLabors.ImageSharp.Tests.PixelFormats.PixelOperations
+{
+    public partial class PixelOperationsTests
+    {
+        <# GenerateAllSpecializedClasses(); #>
+    }
+}
+
```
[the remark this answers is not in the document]
JimBobSquarePants: Honestly.... I tried to write the damn thing without the include and I kept making mistakes. I hate T4.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
```diff
@@ -0,0 +1,119 @@
...
+    void GenerateSpecializedClass(string pixelType, string alpha)
+    {
+#>
+        public partial class <#=pixelType#>_OperationsTests : PixelOperationsTests<<#=pixelType#>>
+        {
+            public <#=pixelType#>_OperationsTests(ITestOutputHelper output)
+                : base(output)
+            {
+                var alphaRepresentation =  <#=pixelType#>.PixelOperations.Instance.GetPixelTypeInfo().AlphaRepresentation;
+                this.HasUnassociatedAlpha = alphaRepresentation == PixelAlphaRepresentation.Unassociated;
```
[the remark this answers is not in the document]
JimBobSquarePants: It could but I was just keeping things together.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
```diff
@@ -0,0 +1,119 @@
...
+
+            [Fact]
+            public void IsSpecialImplementation() => Assert.IsType<<#=pixelType#>.PixelOperations>(PixelOperations<<#=pixelType#>>.Instance);
+
+            [Fact]
+            public void PixelTypeInfoHasCorrectBitsPerPixel()
+            {
+                var bits = <#=pixelType#>.PixelOperations.Instance.GetPixelTypeInfo().BitsPerPixel;
+                Assert.Equal(Unsafe.SizeOf<<#=pixelType#>>() * 8, bits);
+            }
```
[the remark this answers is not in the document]
JimBobSquarePants: See above.

src/ImageSharp/PixelFormats/PixelAlphaRepresentation.cs
JimBobSquarePants: I dunno... We'd then have to have a tracking issue etc. What if someone want to implement their own pixel format type and use it to represent already premultiplied pixel data?

tocsoft: If a pixel had premultiplied alpha would it break any of our processors methods? if it would then either we need to go though and add checks to any that it will break even if we don't do anything else with it or we need to fix them so they function with eather type.

JimBobSquarePants: It's unlikely _I think_ that it would break anything. I couldn't be 100% sure though and since no one has raised an issue suggesting that it does so far I think we're ok.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
JimBobSquarePants: That test could be useful though to check I assigned it correctly.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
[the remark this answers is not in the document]
antonfirsov: But why to repeat the same implementation across multiple classes, when it can be unified?

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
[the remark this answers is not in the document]
antonfirsov: What I mean is:
if we *can* implement something without T4 in test code while still keeping things DRY, we *should*. I view T4 as a tool for DRY in perf-critical scenarios.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/PixelOperationsTests.Specialized.Generated.tt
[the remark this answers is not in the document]
antonfirsov: Then let's leave it, this remark is not critical.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
antonfirsov: > All I'm testing here is the correct assignment of the value.

My point explained further:
it's very easy to make a mistake in such a test here (tempting to copy-paste product code to test code brainlessly, especially if it's generated), but practically impossible, if the test is validating against actual behavior.

But I'm fine with keeping this as an improvement opportunity for later.

src/ImageSharp/PixelFormats/PixelAlphaRepresentation.cs
antonfirsov: It will only make difference when we start utilizing it in `ResizeProcessor`. There's still some non-trivial work out there: Should we watch for the value before calling `PixelOperations.To/FromVector4`? Or should we do it right in the converter method overwriting `PixelConversionModifiers.Premultiply`?

Other than this I don't really see open questions here *according to my understanding of the feature*, but I'm generally reluctant for defining public API-s for stuff that is not even prototyped.

antonfirsov: Might be overthinking in this case, IDK.

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/A8.PixelOperations.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// </content>
+    public partial struct A8
+    {
+        /// <summary>
+        /// Provides optimized overrides for bulk operations.
+        /// </summary>
+        internal class PixelOperations : PixelOperations<A8>
+        {
+            /// <inheritdoc />
+            public override PixelTypeInfo GetPixelTypeInfo() => PixelTypeInfo.Create<A8>(PixelAlphaRepresentation.Unassociated);
```
antonfirsov, line 19: Note for later:
we may want to cache the value into a field/lazy property so we can query it efficiently including it's potential usage in `To/FromVector4`.

src/ImageSharp/PixelFormats/PixelAlphaRepresentation.cs
JimBobSquarePants: I think the best place to do the check is in `PixelConversionModifiers.Premultiply`. However, there are still per-pixel calls to `Premultiply` in the convolution API.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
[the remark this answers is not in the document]
JimBobSquarePants: Moved it. Can't be static though as Operations can't.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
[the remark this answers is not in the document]
JimBobSquarePants: Done

tests/ImageSharp.Tests/PixelFormats/PixelOperations/Generated/_Common.ttinclude
JimBobSquarePants: I've implemented a version of the test in the base class. The original test remains in the generated versions though as I do not use code generation for this property in the source.

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/A8.PixelOperations.cs
JimBobSquarePants: Done.

antonfirsov (approved):
LGTM, great job!

PR #1428, opened 2020-11-15
Use bulk conversion in Write8BitColor

brianpopow:
### Prerequisites

### Description

Just a small improvement in writing 8 bit bitmaps with a palette: Converting the quantized colors to rgb is now done as bulk convert.

### DIFF
--- a/src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
+++ b/src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
@@ -343,7 +343,7 @@ namespace SixLabors.ImageSharp.Formats.Bmp
             Span<Rgba32> rgbColors = rgbColorsBuffer.GetSpan();
 
             ReadOnlySpan<TPixel> quantizedColors = quantized.Palette.Span;
-            PixelOperations<TPixel>.Instance.ToRgba32(Configuration.Default, quantizedColors, rgbColors);
+            PixelOperations<TPixel>.Instance.ToRgba32(this.configuration, quantizedColors, rgbColors);
 
             int idx = 0;
             foreach (Rgba32 color in rgbColors)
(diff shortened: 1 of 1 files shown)


src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
```diff
@@ -340,15 +339,15 @@ private void Write8BitColor<TPixel>(Stream stream, ImageFrame<TPixel> image, Spa
         {
             using IQuantizer<TPixel> frameQuantizer = this.quantizer.CreatePixelSpecificQuantizer<TPixel>(this.configuration);
             using IndexedImageFrame<TPixel> quantized = frameQuantizer.BuildPaletteAndQuantizeFrame(image, image.Bounds());
+            using IMemoryOwner<Rgba32> rgbColorsBuffer = this.memoryAllocator.Allocate<Rgba32>(quantized.Palette.Length);
+            Span<Rgba32> rgbColors = rgbColorsBuffer.GetSpan();
 
             ReadOnlySpan<TPixel> quantizedColors = quantized.Palette.Span;
-            var color = default(Rgba32);
+            PixelOperations<TPixel>.Instance.ToRgba32(Configuration.Default, quantizedColors, rgbColors);
```
JimBobSquarePants, line 346: This should be the configuration that the memory allocator is derived from.

src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
```diff
@@ -340,15 +339,15 @@ private void Write8BitColor<TPixel>(Stream stream, ImageFrame<TPixel> image, Spa
...
 
-            // TODO: Use bulk conversion here for better perf
             int idx = 0;
-            foreach (TPixel quantizedColor in quantizedColors)
+            foreach (Rgba32 color in rgbColors)
```
antonfirsov: Why not something like:
```C#
PixelOperations<TPixel>.Instance.ToRgba32(Configuration.Default, quantizedColors, MemoryMarshal.Cast<byte, Rgba32>(colorPalette));
ref Span<uint> colorPaletteAsUInt= MemoryMarshal.Cast<byte, uint>(colorPalette);
for (int i =0;i<colorPaletteAsUInt;i++)
  colorPaletteAsUInt[i] = colorPaletteAsUInt[i] & 0xFFFFFF00;
```

JimBobSquarePants: Tidy!

brianpopow: yeah that will do, it just needs to be ToBgra instead

JimBobSquarePants: Which should be fast enough with the new shuffle intrinsics

JimBobSquarePants (approved):

Issue #1429, opened 2020-11-16
AccessViolationException when applying adaptive histogram to large images


nullpainter:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

When applying an adaptive histogram equalisation to a 21,969 x 9802px image, I receive the following:

```
Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at SixLabors.ImageSharp.Processing.Processors.Normalization.AdaptiveHistogramEqualizationProcessor`1.CdfTileData.RowIntervalOperation.Invoke(RowInterval& rows)
   at SixLabors.ImageSharp.Advanced.ParallelRowIterator.RowIntervalOperationWrapper`1.Invoke(Int32 i)
   at System.Threading.Tasks.Parallel.<>c__DisplayClass17_0`1.<ForWorker>b__1()
   at System.Threading.Tasks.Task.InnerInvokeWithArg(Task childTask)
   at System.Threading.Tasks.Task.<>c__DisplayClass176_0.<ExecuteSelfReplicating>b__0(Object <p0>)
   at System.Threading.Tasks.Task.Execute()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
   at System.Threading.Tasks.Task.ExecuteEntry(Boolean bPreventDoubleExecution)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

### Steps to Reproduce
This can be easily replicated using the following code, which fails fast:

```c#
using var image = new Image<Rgba32>(20000, 10000);
image.Mutate(c => c.HistogramEqualization(new HistogramEqualizationOptions
{
    Method = HistogramEqualizationMethod.AdaptiveTileInterpolation, 
[...]

PR #1431, opened 2020-11-16
Use GetPixelRowSpan to access pixel data in Histogram equalization

brianpopow:
### Prerequisites

### Description

This PR changes the histogram equalization to use GetPixelRowSpan to access pixel data to fix an Access Violation issue for very large images reported in #1429.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
@@ -145,9 +145,9 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
                 Span<TPixel> rowSpan = source.GetPixelRowSpan(dy);
                 for (int dx = xStart; dx < xEnd; dx++)
                 {
-                    TPixel pixel = rowSpan[dx];
+                    ref TPixel pixel = ref rowSpan[dx];
                     float luminanceEqualized = cdfData.RemapGreyValue(cdfX, cdfY, GetLuminance(pixel, luminanceLevels));
-                    rowSpan[dx].FromVector4(new Vector4(luminanceEqualized, luminanceEqualized, luminanceEqualized, pixel.ToVector4().W));
+                    pixel.FromVector4(new Vector4(luminanceEqualized, luminanceEqualized, luminanceEqualized, pixel.ToVector4().W));
                 }
             }
         }
@@ -191,9 +191,9 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
                     Span<TPixel> rowSpan = source.GetPixelRowSpan(dy);
                     for (int dx = xStart; dx < xEnd; dx++)
                     {
-                        TPixel pixel = rowSpan[dx];
+                        ref TPixel pixel = ref rowSpan[dx];
                         float luminanceEqualized = InterpolateBetweenTwoTiles(pixel, cdfData, cdfX, cdfY, cdfX, cdfY + 1, tileY, tileHeight, luminanceLevels);
-                        rowSpan[dx].FromVector4(new Vector4(luminanceEqualized, luminanceEqualized, luminanceEqualized, pixel.ToVector4().W));
+                        pixel.FromVector4(new Vector4(luminanceEqualized, luminanceEqualized, luminanceEqualized, pixel.ToVector4().W));
                     }
 
                     tileY++;
@@ -243,9 +243,9 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
                     int xLimit = Math.Min(x + tileWidth, sourceWidth - 1);
                     for (int dx = x; dx < xLimit; dx++)
                     {
-                        TPixel pixel = rowSpan[dx];
+                        ref TPixel pixel = ref rowSpan[dx];
                         float luminanceEqualized = InterpolateBetweenTwoTiles(pixel, cdfData, cdfX, cdfY, cdfX + 1, cdfY, tileX, tileWidth, luminanceLevels);
-                        rowSpan[dx].FromVector4(new Vector4(luminanceEqualized, luminanceEqualized, luminanceEqualized, pixel.ToVector4().W));
+                        pixel.FromVector4(new Vector4(luminanceEqualized, luminanceEqualized, luminanceEqualized, pixel.ToVector4().W));
                         tileX++;
                     }
                 }
@@ -423,7 +423,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
                             int tileX = 0;
                             for (int dx = x; dx < xEnd; dx++)
                             {
-                                TPixel pixel = rowSpan[dx];
+                                ref TPixel pixel = ref rowSpan[dx];
                                 float luminanceEqualized = InterpolateBetweenFourTiles(
                                     pixel,
                                     this.cdfData,
@@ -437,7 +437,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
                                     this.tileHeight,
                                     this.luminanceLevels);
 
-                                rowSpan[dx].FromVector4(new Vector4(luminanceEqualized, luminanceEqualized, luminanceEqualized, pixel.ToVector4().W));
+                                pixel.FromVector4(new Vector4(luminanceEqualized, luminanceEqualized, luminanceEqualized, pixel.ToVector4().W));
                                 tileX++;
                             }
 
--- a/src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
@@ -86,42 +86,39 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
                     new Rectangle(0, 0, sourceWidth, tileYStartPositions.Count),
                     in operation);
 
-                ref TPixel pixelsBase = ref source.GetPixelReference(0, 0);
-
                 // Fix left column
-                ProcessBorderColumn(ref pixelsBase, cdfData, 0, sourceWidth, sourceHeight, this.Tiles, tileHeight, xStart: 0, xEnd: halfTileWidth, luminanceLevels);
+                ProcessBorderColumn(source, cdfData, 0, sourceHeight, this.Tiles, tileHeight, xStart: 0, xEnd: halfTileWidth, luminanceLevels);
 
                 // Fix right column
                 int rightBorderStartX = ((this.Tiles - 1) * tileWidth) + halfTileWidth;
-                ProcessBorderColumn(ref pixelsBase, cdfData, this.Tiles - 1, sourceWidth, sourceHeight, this.Tiles, tileHeight, xStart: rightBorderStartX, xEnd: sourceWidth, luminanceLevels);
+                ProcessBorderColumn(source, cdfData, this.Tiles - 1, sourceHeight, this.Tiles, tileHeight, xStart: rightBorderStartX, xEnd: sourceWidth, luminanceLevels);
 
                 // Fix top row
-                ProcessBorderRow(ref pixelsBase, cdfData, 0, sourceWidth, this.Tiles, tileWidth, yStart: 0, yEnd: halfTileHeight, luminanceLevels);
+                ProcessBorderRow(source, cdfData, 0, sourceWidth, this.Tiles, tileWidth, yStart: 0, yEnd: halfTileHeight, luminanceLevels);
 
                 // Fix bottom row
                 int bottomBorderStartY = ((this.Tiles - 1) * tileHeight) + halfTileHeight;
-                ProcessBorderRow(ref pixelsBase, cdfData, this.Tiles - 1, sourceWidth, this.Tiles, tileWidth, yStart: bottomBorderStartY, yEnd: sourceHeight, luminanceLevels);
+                ProcessBorderRow(source, cdfData, this.Tiles - 1, sourceWidth, this.Tiles, tileWidth, yStart: bottomBorderStartY, yEnd: sourceHeight, luminanceLevels);
 
                 // Left top corner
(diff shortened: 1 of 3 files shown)


src/ImageSharp/Processing/Processors/Normalization/GlobalHistogramEqualizationProcessor{TPixel}.cs
```diff
@@ -165,18 +165,18 @@ public void Invoke(int y)
...
-                ref TPixel pixelBase = ref MemoryMarshal.GetReference(this.source.GetPixelRowSpan(y));
+                Span<TPixel> pixelRow = this.source.GetPixelRowSpan(y);
                 int levels = this.luminanceLevels;
                 float noOfPixelsMinusCdfMin = this.numberOfPixelsMinusCdfMin;
 
                 for (int x = 0; x < this.bounds.Width; x++)
                 {
                     // TODO: We should bulk convert here.
-                    ref TPixel pixel = ref Unsafe.Add(ref pixelBase, x);
+                    TPixel pixel = pixelRow[x];
```
JimBobSquarePants, line 175: You can still use `ref` here and avoid the double access.

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
```diff
@@ -146,23 +142,22 @@ protected override void OnFrameApply(ImageFrame<TPixel> source)
         {
             for (int dy = yStart; dy < yEnd; dy++)
             {
-                int dyOffSet = dy * sourceWidth;
```
JimBobSquarePants, line 149: It seems to me like `int dyOffSet = dy * sourceWidth;` (_and the same below_) were the actual source of the issue and we could have mitigated the issue with rows whilekeeping `Unsafe.Add`.

I'd like to see a before/after benchmark if possible.

antonfirsov: > while keeping Unsafe.Add

Do we really want it? What I meant in https://github.com/SixLabors/ImageSharp/issues/1429#issuecomment-727931317 is a policy against `Unsafe.*` regardless of whether it can be justified in a particular use-case.

These processors are the spice in the library, it's not the core functionality where we really need the 1-2 extra percents got by micro-optimization. I'd rather go low-risk here. It's much better to get `IndexOutOfRangException` than an access violation which is a security-critical bug.

JimBobSquarePants: I'd like to see benchmarks before I make a decision.

brianpopow:
after the changes of this PR:

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19041.630 (2004/?/20H1)
.NET Core SDK=5.0.100
  [Host]     : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  Job-ZJWYUQ : .NET Framework 4.8 (4.8.4250.0), X64 RyuJIT
  Job-PDUVYP : .NET Core 2.1.23 (CoreCLR 4.6.29321.03, CoreFX 4.6.29321.01), X64 RyuJIT
  Job-ZXDATS : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT

|                                               Method |        Job |       Runtime |      Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------------------- |----------- |-------------- |----------:|----------:|----------:|------:|------:|------:|----------:|
|                      'Global Histogram Equalization' | Job-YJFVOZ |    .NET 4.7.2 |  2.992 ms | 0.5856 ms | 0.0321 ms |     - |     - |     - |   5.22 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-YJFVOZ |    .NET 4.7.2 | 38.121 ms | 2.7466 ms | 0.1505 ms |     - |     - |     - |   6.77 KB |
|                      'Global Histogram Equalization' | Job-RDLXEE | .NET Core 2.1 |  2.299 ms | 0.6028 ms | 0.0330 ms |     - |     - |     - |   5.15 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-RDLXEE | .NET Core 2.1 | 18.264 ms | 2.2435 ms | 0.1230 ms |     - |     - |     - |   5.95 KB |
|                      'Global Histogram Equalization' | Job-VSRZSV | .NET Core 3.1 |  2.268 ms | 1.1815 ms | 0.0648 ms |     - |     - |     - |    6.9 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-VSRZSV | .NET Core 3.1 | 15.433 ms | 1.9711 ms | 0.1080 ms |     - |     - |     - |   5.95 KB |
```

before that:

```
|                                               Method |        Job |       Runtime |      Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------------------- |----------- |-------------- |----------:|----------:|----------:|------:|------:|------:|----------:|
|                      'Global Histogram Equalization' | Job-XFMRZJ |    .NET 4.7.2 |  3.061 ms | 1.3141 ms | 0.0720 ms |     - |     - |     - |   5.16 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-XFMRZJ |    .NET 4.7.2 | 48.401 ms | 6.5210 ms | 0.3574 ms |     - |     - |     - |    8.8 KB |
|                      'Global Histogram Equalization' | Job-WTJMNW | .NET Core 2.1 |  2.315 ms | 0.9744 ms | 0.0534 ms |     - |     - |     - |   5.13 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-WTJMNW | .NET Core 2.1 | 37.697 ms | 2.5760 ms | 0.1412 ms |     - |     - |     - |   5.95 KB |
|                      'Global Histogram Equalization' | Job-AVGHFY | .NET Core 3.1 |  2.287 ms | 0.6242 ms | 0.0342 ms |     - |     - |     - |   6.82 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-AVGHFY | .NET Core 3.1 | 16.601 ms | 3.7207 ms | 0.2039 ms |     - |     - |     - |   5.93 KB |
```

netcore 3.1 is almost the same as before, but net471 and core 2.1 are slower now.

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
JimBobSquarePants: I do agree we should be a lot more careful with `Unsafe` though. I'm guilty of reaching for it far too often.

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
```diff
@@ -244,12 +238,12 @@ protected override void OnFrameApply(ImageFrame<TPixel> source)
             {
                 for (int dy = yStart; dy < yEnd; dy++)
                 {
-                    int dyOffSet = dy * sourceWidth;
+                    Span<TPixel> rowSpan = source.GetPixelRowSpan(dy);
                     int tileX = 0;
                     int xLimit = Math.Min(x + tileWidth, sourceWidth - 1);
                     for (int dx = x; dx < xLimit; dx++)
```
JimBobSquarePants, line 244: I think if we sliced the `rowSpan` and used `.Length` then we should be able avoid all bounds checks and maintain previous performance in a safe manner on all targets.

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
JimBobSquarePants: Think I've spotted a way to keep perf for all in a safe manner.

antonfirsov: Try a microbenchmark iterating through a span. You shall see no difference  or almost no difference if the loop is defined as `for (int i = 0; i < span.Length; i++)`.

Since real life code does much more than that, benefits shall be negligible or nonexistent. I barely see such code in BCL, I believe the only visible benefit is on .NET Framework, or maybe older .NET Core. 

If there is no proof for visible benefits, I would strongly prefer to see a guideline point against `Unsafe.*`.

antonfirsov:
@brianpopow I guess before/after are the other way around in https://github.com/SixLabors/ImageSharp/pull/1431#issuecomment-728149117?

Any explanation for the significant difference for Framework VS Core? I would be surprised if it's only about indexers.

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
JimBobSquarePants: That's why I suggested the slice outside the loop in this instance so we can use `.Length`. There's little change in Core 3.1 with the current approach but Core 2.1 and NET 472 are suffering according to the benchmarks. 

>If there is no proof for visible benefits, I would strongly prefer to see a guideline point against Unsafe.*.

Yep. Totally with you. I know of several places in the library where it would be smart to revert to safer code.

brianpopow:
> @brianpopow I guess before/after are the other way around in [#1431 (comment)](https://github.com/SixLabors/ImageSharp/pull/1431#issuecomment-728149117)?
> 
> Any explanation for the significant difference for Framework VS Core? I would be surprised if it's only about indexers.

Give me a moment i will do the benchmark again. I really think i made a mistake.

brianpopow:
> @brianpopow I guess before/after are the other way around in [#1431 (comment)](https://github.com/SixLabors/ImageSharp/pull/1431#issuecomment-728149117)?
> 
> Any explanation for the significant difference for Framework VS Core? I would be surprised if it's only about indexers.

@antonfirsov  @JimBobSquarePants: I have run the benchmark now many times, but with the same results as above. The current master branch is actually **slower** than the change in this PR. I could not believe my eyes, maybe some else wants to double check this.

**After the change of this PR:**

```
|                                               Method |        Job |       Runtime |      Mean |    Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------------------- |----------- |-------------- |----------:|---------:|----------:|------:|------:|------:|----------:|
|                      'Global Histogram Equalization' | Job-KEXZPE |    .NET 4.7.2 |  3.028 ms | 1.898 ms | 0.1040 ms |     - |     - |     - |   5.16 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-KEXZPE |    .NET 4.7.2 | 38.288 ms | 3.331 ms | 0.1826 ms |     - |     - |     - |   7.43 KB |
|                      'Global Histogram Equalization' | Job-GHKMRJ | .NET Core 2.1 |  2.313 ms | 1.010 ms | 0.0553 ms |     - |     - |     - |   5.14 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-GHKMRJ | .NET Core 2.1 | 18.522 ms | 4.412 ms | 0.2419 ms |     - |     - |     - |   5.94 KB |
|                      'Global Histogram Equalization' | Job-DUVFTK | .NET Core 3.1 |  2.350 ms | 2.915 ms | 0.1598 ms |     - |     - |     - |    6.9 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-DUVFTK | .NET Core 3.1 | 15.641 ms | 4.962 ms | 0.2720 ms |     - |     - |     - |   5.91 KB |
```

**Current master branch:**

```
|                                               Method |        Job |       Runtime |      Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------------------- |----------- |-------------- |----------:|----------:|----------:|------:|------:|------:|----------:|
|                      'Global Histogram Equalization' | Job-WDJRLD |    .NET 4.7.2 |  3.008 ms | 0.5885 ms | 0.0323 ms |     - |     - |     - |   5.19 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-WDJRLD |    .NET 4.7.2 | 49.309 ms | 4.5882 ms | 0.2515 ms |     - |     - |     - |   7.88 KB |
|                      'Global Histogram Equalization' | Job-TYBBMO | .NET Core 2.1 |  2.249 ms | 0.7262 ms | 0.0398 ms |     - |     - |     - |   5.14 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-TYBBMO | .NET Core 2.1 | 37.419 ms | 2.8973 ms | 0.1588 ms |     - |     - |     - |   5.96 KB |
|                      'Global Histogram Equalization' | Job-OCZNHU | .NET Core 3.1 |  2.281 ms | 0.8250 ms | 0.0452 ms |     - |     - |     - |   6.89 KB |
| 'AdaptiveHistogramEqualization (Tile interpolation)' | Job-OCZNHU | .NET Core 3.1 | 16.639 ms | 4.2431 ms | 0.2326 ms |     - |     - |     - |   5.88 KB |
```

I cannot really tell where the large difference between the framework and core comes from.

JimBobSquarePants:
Haha... I read it the same as Anton. I say we merge this then!

JimBobSquarePants (approved):

antonfirsov:
Can we add a test based on what we know about the root cause? (Maybe doesn't need the `20k x 10k` size.)

brianpopow:
> Can we add a test based on what we know about the root cause? (Maybe doesn't need the `20k x 10k` size.)

It seems it has to be large to trigger this. 15k x 10k was the "smallest" image where i have seen this. This seems too large for a unit test in my opinion. @antonfirsov what do you think?

antonfirsov (approved):
We can add such a test with `RemoteExecutor`, but it will probably run for seconds, and we probably don't want the additional load on our CI, in that case the PR is fine as is.

PR #1439, opened 2020-11-24
Add cache for Nuget packages

AlexNDRmac:
### Prerequisites

### Description

- Added cache action for Nuget packages for speedup Nuget restore before each CI builds

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -56,6 +56,14 @@ jobs:
                   git fetch --prune --unshallow
                   git submodule -q update --init --recursive
 
+            - name: Setup nuget cache
+              uses: actions/cache@v2
+              id: nuget-cache
+              with:
+                  path: ~/.nuget
+                  key: ${{ runner.os }}-nuget-${{ hashFiles('**/ImageSharp*.csproj') }}
+                  restore-keys: ${{ runner.os }}-nuget-
+
             - name: Build
               shell: pwsh
               run: ./ci-build.ps1 "${{matrix.options.framework}}"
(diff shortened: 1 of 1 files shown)


AlexNDRmac:
@antonfirsov current summary restoring time for all projects is about ~1 minutes. With cache (next time, when you will build with enabled cache action) summary restore time should be ~5-10 seconds.
I guess, you can compare build time for each OS by triggering build action again.
Cache key based on all files in repository with .csproj extension, so any changes with packages will be change hash of cache kay.

.github/workflows/build-and-test.yml
```diff
@@ -56,6 +56,14 @@ jobs:
                   git fetch --prune --unshallow
                   git submodule -q update --init --recursive
 
+            - name: Setup nuget cache
+              uses: actions/cache@v2
+              id: nuget-cache
+              with:
+                  path: ~/.nuget
+                  key: ${{ runner.os }}-nuget-${{ hashFiles('**/ImageSharp*.csproj') }}
```
tocsoft, line 64: you will need to add the `*.props`/`*.targets` too as we define most of our package versions in those shared locations i.e. https://github.com/SixLabors/ImageSharp/blob/0e0dc2ae9cafcdf5bde9d185919cf073ddf4f186/Directory.Build.targets#L19-L37

tocsoft (changes requested):
haven't captured all the files that could end up changing the packages

.github/workflows/build-and-test.yml
AlexNDRmac: Thanks. Added with next commit.

JimBobSquarePants (approved):
Ok. Looks like everything is covered here now. Let's give it a whirl!

Issue #1442, opened 2020-11-26
Some (28) tests are failing on master branch


Evangelink:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

On a clean master branch, running all tests from VS 2019, I end up with 28 failing tests (see image below):
![image](https://user-images.githubusercontent.com/11340282/100328635-098d7a00-2fcd-11eb-9a80-49b10e258816.png)

### System Configuration

- ImageSharp version: latest master commit
- Environment (Operating system, version and so on): Win 10 (Version 10.0.18363 Build 18363)

brianpopow:
thanks for reporting this. I see the reason for the bitmap tests failing: Its a DebugGuard which ensures the Spans are the same length when converting to Bgra32. This is sometimes not the case during quantization where it can happen that the source is smaller then the destination. I will fix that.

Evangelink:
Just FYI there seem to be also an `AccessViolationException` and a `OutOfMemoryException`, I don't know if they relates to the spans issues.

antonfirsov:
Is it a 32bit run?  OOM-s are not extraordinary. 

AccessViolationException is a serious concern, what test is that? Is it sporadic or deterministic?

JimBobSquarePants:
What’s the stack trace of the AccessViolationException. I’ve seen that one pop up a few times during CI and it’s always Magick.NET

brianpopow:
> What’s the stack trace of the AccessViolationException. I’ve seen that one pop up a few times during CI and it’s always Magick.NET

yes the AccessViolationException is related to `MagickReferenceDecoder` and it only happens with `net472` as far as i can tell. It only happens sporadic and not always.

Evangelink:
It's sporadic, I will put the stacktrace next time I face it.

Regarding the OOM, they appear with Any CPU, x86 and x64 but it's also something quite sporadic. Here is the stacktrace:

```
  Message: 
    System.OutOfMemoryException : Exception of type 'System.OutOfMemoryException' was thrown.
  Stack Trace: 
[...]

PR #1443, opened 2020-11-26
Ensure Span length of source and destination are equal during pixel conversion

brianpopow:
### Prerequisites

### Description

As reported in #1442, there are some failing tests in the master branch. The reason for this is a `Debug.Guard` which ensures the span length are equal during pixel conversion. Sometimes the source span was smaller the the destination. This got unnoticed during **Release** CI runs.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -56,6 +56,14 @@ jobs:
                   git fetch --prune --unshallow
                   git submodule -q update --init --recursive
 
+            - name: Setup nuget cache
+              uses: actions/cache@v2
+              id: nuget-cache
+              with:
+                  path: ~/.nuget
+                  key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.csproj', '**/*.props', '**/*.targets') }}
+                  restore-keys: ${{ runner.os }}-nuget-
+
             - name: Build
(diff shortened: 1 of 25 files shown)


antonfirsov (approved):
LGTM.

antonfirsov:
> The reason for this is a `Debug.Guard` which ensures the span length are equal during pixel conversion. 

This shouldn't be the case, the `PixelOperations` contract allows `destination` to be longer. @JimBobSquarePants looks like `SimdUtils.Shuffle*` does a strict verification, which is fine (modeled after the `Vector4` conversion stuff I guess), but we should slice down `destination` in `PixelOperations` then.

JimBobSquarePants:
Aha! I actually had to do a hack to allow for this in a debug check. 

https://github.com/SixLabors/ImageSharp/commit/1ad9fcdad63259ee628b04c81b85108e9948bbfb#diff-ba27a760948c3e593e3ac6a18ad7b8ea04bc0255376575088b7fad70814040e3R227-R231

I'll be able to set it right.

brianpopow:
i will leave this as it is now. Maybe someone should give a quick review of 7c77e7e11cd49c39e057523aa0147bf0ae6cccff again

JimBobSquarePants (approved):
LGTM 👍

PR #1444, opened 2020-11-27
Reduce code duplication due to reified generics

Sergio0694:
### Prerequisites

### Description
The `BokehBlurProcessor<TPixel>` type had one of the internal row iterators that worked on non-generic types, but having that nested type defined within the generic processor class caused the code to be duplicated and compiled again for every single pixel type that the process was applied to. I moved that to the non-generic processor type so the same generic instantiation can be reduced. This will reduce the binary size in AOT scenarios, and reduce the load on the JIT otherwise.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
@@ -127,7 +127,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Convolution
                     in verticalOperation);
 
                 // Compute the horizontal 1D convolutions and accumulate the partial results on the target buffer
-                var horizontalOperation = new ApplyHorizontalConvolutionRowOperation(sourceRectangle, processingBuffer, firstPassBuffer, kernel, parameters.Z, parameters.W);
+                var horizontalOperation = new BokehBlurProcessor.ApplyHorizontalConvolutionRowOperation(sourceRectangle, processingBuffer, firstPassBuffer, kernel, parameters.Z, parameters.W);
                 ParallelRowIterator.IterateRows(
                     configuration,
                     sourceRectangle,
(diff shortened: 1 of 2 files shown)


src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
```diff
@@ -127,7 +127,7 @@ protected override void OnFrameApply(ImageFrame<TPixel> source)
                     in verticalOperation);
 
                 // Compute the horizontal 1D convolutions and accumulate the partial results on the target buffer
-                var horizontalOperation = new ApplyHorizontalConvolutionRowOperation(sourceRectangle, processingBuffer, firstPassBuffer, kernel, parameters.Z, parameters.W);
+                var horizontalOperation = new BokehBlurProcessor.ApplyHorizontalConvolutionRowOperation(sourceRectangle, processingBuffer, firstPassBuffer, kernel, parameters.Z, parameters.W);
```
JimBobSquarePants, line 130: Can you not reference the processor passed to the constructor?

JimBobSquarePants:
@Sergio0694 Nice! Always on the lookout for things like this. We need to ensure the amount of code we produce is as small as possible to avoid issues with AOT.

src/ImageSharp/Processing/Processors/Convolution/BokehBlurProcessor{TPixel}.cs
Sergio0694: Not sure what you mean - this is a static reference, we're not accessing an instance member. That's just the qualified type that's declared within `BokehBlurProcess`, so we need the parent type name when using it 🤔

JimBobSquarePants: I misread, sorry, carry on. 👍

JimBobSquarePants (approved):
LGTM 🚀

PR #1451, opened 2020-12-01
Update to use new props/targets plus cleanup.

JimBobSquarePants:
### Prerequisites

### Description

Vastly simplifies the solution by reusing props/targets defined in our infrastructure. I've also done some tests cleanup to match rules and added missing images to the solution plus replaced the EXIF standard PDF broken by old gitattributes.


brianpopow (approved):
Looks good, nice improvement 👍

PR #1456, opened 2020-12-03
Tiff format support

brianpopow:
### Prerequisites

### Description
This PR adds support for TIFF decoding and encoding.

The specification can be found here: [TIFF spec](https://www.adobe.io/content/udp/en/open/standards/TIFF/_jcr_content/contentbody/download_1370394226/file.res/TIFFphotoshop.pdf)
Or as a PDF in the tiff folder (ImageSharp\Formats\Tiff).

Currently the Encoder and Decoder support the following:

## Baseline TIFF

- [x] Bilevel Images aka Black and White images.
- [x] Grayscale Images
- [x] Palette-color Images
- [x] RGB Full Color Images
- [x] PackBits Compression
- [x] Modified Huffman Compression (Ccitt1D)

## TIFF Extensions

- [x] CcittGroup3Fax
- [x] LZW compression
- [x] Deflate compression
- [x] Differencing Predictor

The following extension features will probably not be part of the first release of the TIFF format support.
- CcittGroup4Fax
- JPEG Compression [Technote 2](http://simplesystems.org/libtiff/TIFFTechNote2.html)
- Tiled Images
- TransparencyMask
- Separated (TIFF Extension color spaces)
- YCbCr (TIFF Extension color spaces)
- CieLab (TIFF Extension color spaces)
- IccLab (TechNote 1)

## TODO's

- [x] ~Issue with LZW decoding with some images, for example: `rgb_lzw_no_predictor.tiff`~ Fixed with #1457 
- [x] Deflate compression uses deflate implementation of PNG, we should move that now to a common place 
- [x] ~There is still an issue with LZW. Compression a byte array with LZW and then decompress it again does not yield in the original byte array. See failing tests.~ fixed with da51a4f9


src/ImageSharp/Formats/Png/DeflateCompressionLevel.cs
```diff
@@ -6,7 +6,7 @@ namespace SixLabors.ImageSharp.Formats.Png
     /// <summary>
     /// Provides enumeration of available PNG compression levels.
     /// </summary>
-    public enum PngCompressionLevel
+    public enum DeflateCompressionLevel
```
antonfirsov, line 9: While reasonable, this will be a hard breaking change, introducing an experimental codec, doesn't justify it for the rest of the users IMO. Can we try a softer way? 
- Duplicating `DeflateCompressionLevel` as `PngCompressionLevel` while marking the latter `[EditorBrowsable(false)]`
- Duplicating a `PngCompressionLevel` variant of `PngEncoder.CompressionLevel` and marking it `[EditorBrowsable(false)]`

We [may or may not want](https://github.com/SixLabors/ImageSharp/discussions/1454) to obsolete the old API-s, no strong opinion.

brianpopow: yes you are right, i will change that back

src/ImageSharp/Formats/ImageExtensions.Save.tt
```diff
@@ -17,6 +17,7 @@ using SixLabors.ImageSharp.Advanced;
         "Jpeg",
         "Png",
         "Tga",
+        "Tiff",
```
antonfirsov, line 20: I suggest to add something like this in the `foreach (string fmt in formats)` loop:

```C#
string experimentalString =  fmt != "Tiff" ? "" : @"EXPERIMENTAL! ";
#>

        /// <summary>
        /// <#= experimentalString #>Saves the image to the given stream with the <#= fmt #> format.	        /// Saves the image to the given stream with the <#= fmt #> format.
        /// </summary>

// same in other overloads
```

src/ImageSharp/Formats/Png/DeflateCompressionLevel.cs
antonfirsov: Changing back is fine for now. Later we may still want to introduce the "soft" ranaming, since `PngCompressionLevel` is weird when using with TIFF :)

src/ImageSharp/Formats/Tiff/Constants/TiffByteOrder.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Formats.Experimental.Tiff.Constants
+{
+    /// <summary>
+    /// The tiff data stream byte order enum.
+    /// </summary>
+    public enum TiffByteOrder
```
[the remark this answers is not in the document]
brianpopow: i have moved this to the Common folder and renamed it to just ByteOrder: 13eb5c2a

src/ImageSharp/Formats/Tiff/ImageExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <returns>
+        /// The <see cref="Image{TPixel}"/>.
+        /// </returns>
+        public static Image<TPixel> SaveAsTiff<TPixel>(this Image<TPixel> source, Stream stream, TiffEncoder encoder)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            encoder = encoder ?? new TiffEncoder();
+            encoder.Encode(source, stream);
+
+            return source;
```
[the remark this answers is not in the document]
brianpopow: yes it is, its removed now: b8f8f17b

src/ImageSharp/Formats/ImageExtensions.Save.tt
brianpopow: done

src/ImageSharp/Formats/Png/DeflateCompressionLevel.cs
brianpopow: I have duplicated `PngCompressionLevel` to `DeflateCompressionLevel` and marked `PngCompressionLevel` as `[EditorBrowsable(EditorBrowsableState.Never)]`.

I was unsure if that's what you suggested. Is that 7357ae38 ok?

antonfirsov: Also wanted to somehow duplicate the `PngEncoder.CompressionLevel` but did not think it through. We can't have two properties of the same name, not sure how to proceed now.

antonfirsov: If noone can figure it out, lets revert to the old name for the first release maybe?

brianpopow: `PngEncoder.CompressionLevel` was already changed back with a09cefe080e8f86ae60a12a16adb8f509c17b97d

tests/ImageSharp.Tests/Formats/Tiff/TiffMetadataTests.cs
```diff
@@ -17,6 +17,14 @@ namespace SixLabors.ImageSharp.Tests.Formats.Tiff
     [Trait("Format", "Tiff")]
     public class TiffMetadataTests
     {
+        public TiffMetadataTests()
+        {
+            Configuration.Default.ImageFormatsManager.AddImageFormat(TiffFormat.Instance);
+            Configuration.Default.ImageFormatsManager.AddImageFormatDetector(new TiffImageFormatDetector());
+            Configuration.Default.ImageFormatsManager.SetDecoder(TiffFormat.Instance, new TiffDecoder());
+            Configuration.Default.ImageFormatsManager.SetEncoder(TiffFormat.Instance, new TiffEncoder());
```
antonfirsov, line 25: I suggest to use an instance configuration in tiff tests.

src/ImageSharp/Formats/Tiff/Ifd/EntryReader.cs
```diff
@@ -0,0 +1,312 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EntryReader" /> class.
+        /// </summary>
+        /// <param name="stream">The stream.</param>
+        public EntryReader(TiffStream stream)
+        {
+            this.stream = stream;
+        }
+
+        public IExifValue ReadNext()
```
JimBobSquarePants, line 29: I'm confused by this class. Shouldn't the existing existing `ExifReader` class be handling this?

src/ImageSharp/Formats/Tiff/TiffFrameMetadata.cs
```diff
@@ -0,0 +1,332 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System.Collections.Generic;
+using System.Linq;
+
+using SixLabors.ImageSharp.Formats.Experimental.Tiff.Constants;
+using SixLabors.ImageSharp.Metadata.Profiles.Exif;
+
+namespace SixLabors.ImageSharp.Formats.Experimental.Tiff
```
JimBobSquarePants, line 10: The way Exif data is handled here is troublesome. I would expect a profile per frame if we are to split the data up. We can't use our existing APIs to update the data.

src/ImageSharp/Formats/Tiff/TiffMetadata.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        /// <summary>
+        /// Gets or sets the compression used to create the TIFF file.
+        /// </summary>
+        public TiffCompression Compression { get; set; } = TiffCompression.None;
+
+        /// <summary>
+        /// Gets or sets the XMP profile.
+        /// </summary>
+        public byte[] XmpProfile { get; set; }
```
JimBobSquarePants, line 49: Shouldn't we have a dedicated type for this?

src/ImageSharp/Formats/Tiff/Utils/SubStream.cs
```diff
@@ -0,0 +1,176 @@
...
+namespace SixLabors.ImageSharp.Formats.Experimental.Tiff.Utils
+{
+    /// <summary>
+    /// Utility class to encapsulate a sub-portion of another <see cref="Stream"/>.
+    /// </summary>
+    /// <remarks>
+    /// Note that disposing of the <see cref="SubStream"/> does not dispose the underlying
+    /// <see cref="Stream"/>.
+    /// </remarks>
+    internal class SubStream : Stream
```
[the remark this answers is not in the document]
IldarKhayrutdinov: @JimBobSquarePants @brianpopow yes I am preparing PR to remove this class

src/ImageSharp/Formats/Tiff/TiffMetadata.cs
IldarKhayrutdinov: @JimBobSquarePants ImageSharp don't support XMP profile yet, so here the data is in raw format

src/ImageSharp/Formats/Tiff/Ifd/EntryReader.cs
IldarKhayrutdinov: @JimBobSquarePants Exif is extension of Tiff and it seems that reading for Tif is a little different from reading existing `ExifReader`, I'll figure it out more and come back with an answer

src/ImageSharp/Formats/Tiff/TiffMetadata.cs
JimBobSquarePants: This should be internal then. We cannot introduce a change we will likely break

tests/ImageSharp.Tests/Formats/Tiff/TiffMetadataTests.cs
brianpopow: changed with 1c734b34

src/ImageSharp/Formats/Tiff/Utils/SubStream.cs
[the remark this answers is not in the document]
IldarKhayrutdinov: Please review #1467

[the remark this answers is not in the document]
brianpopow: its merged now

src/ImageSharp/Formats/Tiff/TiffMetadata.cs
brianpopow: i will make that internal tomorrow

IldarKhayrutdinov:
#1471 - here I'm workin on metadata, while there are errors and improvements are required

src/ImageSharp/Formats/Tiff/TiffMetadata.cs
IldarKhayrutdinov: @JimBobSquarePants @brianpopow didn't notice this message, yes, it's done!

src/ImageSharp/Formats/Tiff/Ifd/EntryReader.cs
IldarKhayrutdinov: @JimBobSquarePants @brianpopow  tiff and exif have differences in tags reading, but I think it's possible to combine them (although it will probably be ugly), it will take some work, coming days I can't do, perhaps next week

brianpopow: @IldarKhayrutdinov maybe you could elaborate what you want to change beforehand. If its really "ugly", we probably should not do it in the first place. Maybe we can come up with a better solution.

IldarKhayrutdinov: @brianpopow yes, you still need to think about how to do it beautifully 🤔

IldarKhayrutdinov:
Hi everyone! Please review #1494

IldarKhayrutdinov:
@brianpopow thanks for corrections of big benchmarks, I haven't yet looked you recommended `Calliphora_x` files. I used these scripts (and big benchmarks) for development purposes, now the tiff format is almost ready, you can either leave them or somehow improve.

brianpopow:
> @brianpopow thanks for corrections of big benchmarks, I haven't yet looked you recommended `Calliphora_x` files. I used these scripts (and big benchmarks) for development purposes, now the tiff format is almost ready, you can either leave them or somehow improve.

@IldarKhayrutdinov: I think we can leave the benchmark as it is right now.

IldarKhayrutdinov:
> I think we can leave the benchmark as it is right now.

yes they can be used for stress tests

IldarKhayrutdinov:
Hi, @brianpopow, I'm currently working on multiStrip support for encoder

brianpopow:
> Hi, @brianpopow, I'm currently working on multiStrip support for encoder

@IldarKhayrutdinov That's great! Its actually recommended to encode with multiStrip. Because i have read it like 10 minutes ago, 8k bytes per strip seems to be the recommendation.

IldarKhayrutdinov:
As first part of multiStrip for encoder task - #1527 - only splitting and move classes, no change algorithms.

src/ImageSharp/Metadata/Profiles/Exif/Tags/ExifTag.ByteArray.cs
```diff
@@ -21,6 +21,16 @@ public abstract partial class ExifTag
         /// </summary>
         public static ExifTag<byte[]> XMP => new ExifTag<byte[]>(ExifTagValue.XMP);
 
+        /// <summary>
+        /// Gets the IPTC exif tag.
+        /// </summary>
+        public static ExifTag<byte[]> IPTC => new ExifTag<byte[]>(ExifTagValue.IPTC);
```
brianpopow, line 27: @IldarKhayrutdinov I have trouble with decoding some images from the tiff test-suite.
It seems that some encoders treat the IPTC datatype as **long[]** instead of **byte[]**.
For [example](http://www.luotain.uku.fi/~kraka/uudet_sivut/data/pressi/kraka_pressi02.tif) or [example2](http://korejtko.net/wp-content/uploads/2008/11/148222981.tif)

There are some [sources](https://www.awaresystems.be/imaging/tiff/tifftags/iptc.html) which claim
treating it as **long** is incorrect. I would totally agree, but some encoders do not.
Take for example `iptc.jpg` from our testimages in `Input\Jpg\baseline` and convert it to tif
with either imagemagick or gimp. Both will treat IPTC as **long** for some reason.

IldarKhayrutdinov: I think we can remove this line ` public static ExifTag<byte[]> IPTC => new ExifTag<byte[]>(ExifTagValue.IPTC);` the iptc tag object will be created correctly and needs to add the correct processing of iptc tag when IptcProfile creating.

IldarKhayrutdinov: Although, the data must be in byte [] format, even if long. Probably should add workaround when reading (exif reader)

brianpopow: i think **byte** is the correct datatype. I will create a workaround for the cases when its long.

IldarKhayrutdinov:
Please review MultiStrip encoding  #1537

JimBobSquarePants:
@IldarKhayrutdinov @brianpopow Have either of you seen this library? I wonder if there's some features we can compare there. 

https://github.com/yigolden/TiffLibrary/blob/master/Documentation/supported-tiff-features.md

IldarKhayrutdinov:
> @IldarKhayrutdinov @brianpopow Have either of you seen this library? I wonder if there's some features we can compare there.
> 
> https://github.com/yigolden/TiffLibrary/blob/master/Documentation/supported-tiff-features.md

I haven't seen, but it seems we can benefit from her, for example features: BigTIFF, tiled; TiffGray16 Pixel, CMYK, YCbCr Photometric; T4, T6, JPEG, ThunderScan, NeXT compressions; and we can compare performance for existing features.

brianpopow:
> @IldarKhayrutdinov @brianpopow Have either of you seen this library? I wonder if there's some features we can compare there.
> 
> https://github.com/yigolden/TiffLibrary/blob/master/Documentation/supported-tiff-features.md

I have not seen that either. As @IldarKhayrutdinov said, good to see how others have implemented some features and to compare performance against.

JimBobSquarePants:
Dunno why this got closed. Something to do with the LFS changes I think.

JimBobSquarePants:
PR is now at #1553 due to Git LFS force push.

PR #1462, opened 2020-12-05
Implement PackFromRgbPlanes for Rgba32 and Rgb24

antonfirsov:
### Prerequisites

### Description
This is a replacement for #1242, implementing the algorithm suggested by @saucecontrol in https://github.com/SixLabors/ImageSharp/pull/1242#issuecomment-735346693.
Contributes to #1121 and #1410. 

### Results

The baseline is the `float` AVX2 packing which is the final step of all JpegColorConverters.

```
.NET Core 3.1.1
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores


|                           Method | Count |       Mean |    Error |   StdDev | Ratio | RatioSD |
|--------------------------------- |------ |-----------:|---------:|---------:|------:|--------:|
|     Rgb24_Scalar_PerElement_Span |  1024 | 1,634.6 ns | 26.56 ns | 24.84 ns |  3.12 |    0.05 |
|   Rgb24_Scalar_PerElement_Unsafe |  1024 | 1,284.7 ns |  4.70 ns |  4.16 ns |  2.46 |    0.01 |
| Rgb24_Scalar_PerElement_Batched8 |  1024 | 1,182.3 ns |  5.12 ns |  4.27 ns |  2.26 |    0.01 |
| Rgb24_Scalar_PerElement_Batched4 |  1024 | 1,146.2 ns | 16.38 ns | 14.52 ns |  2.19 |    0.02 |
|                Rgba32_Avx2_Float |  1024 |   522.7 ns |  1.78 ns |  1.39 ns |  1.00 |    0.00 |
|                 Rgb24_Avx2_Bytes |  1024 |   243.3 ns |  1.56 ns |  1.30 ns |  0.47 |    0.00 |
|                Rgba32_Avx2_Bytes |  1024 |   146.0 ns |  2.48 ns |  2.32 ns |  0.28 |    0.01 |
```

#### Conclusions:
- The trick seems to worth it
- Although staying on `Rgba32` would be faster, we can bring `Rgb24` closer in the final implementation because we need to convert 25% less `float`-s to `byte`-s with `Image<Rgb24>`.

### Follow-up
Finishing  #1121 and #1410 should be a trivial refactor from now on, I can describe all the steps if someone wants to take it. (I likely won't have the time before January.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -105,8 +105,11 @@
 *.pvr               binary
 *.snk               binary
 *.tga               binary
+*.tif               binary
+*.tiff              binary
 *.ttc               binary
 *.ttf               binary
+*.wbmp              binary
 *.webp              binary
 *.woff              binary
 *.woff2             binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -105,11 +105,8 @@
 *.pvr               binary
 *.snk               binary
 *.tga               binary
-*.tif               binary
-*.tiff              binary
 *.ttc               binary
 *.ttf               binary
-*.wbmp              binary
 *.webp              binary
 *.woff              binary
 *.woff2             binary
--- a/src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
+++ b/src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
@@ -4,6 +4,8 @@
 using System;
 using System.Buffers;
 using System.Numerics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using SixLabors.ImageSharp.Formats;
 using SixLabors.ImageSharp.Memory;
 
@@ -177,13 +179,26 @@ namespace SixLabors.ImageSharp.PixelFormats
             Span<TPixel> destination)
         {
             Guard.NotNull(configuration, nameof(configuration));
+
+            int count = redChannel.Length;
+            Guard.IsTrue(greenChannel.Length == count, nameof(greenChannel), "Channels must be of same size!");
+            Guard.IsTrue(blueChannel.Length == count, nameof(blueChannel), "Channels must be of same size!");
+            Guard.IsTrue(destination.Length > count + 2, nameof(destination), "'destination' must contain a padding of 3 elements!");
+
             Guard.DestinationShouldNotBeTooShort(redChannel, destination, nameof(destination));
 
-            for (int i = 0; i < destination.Length; i++)
-            {
-                var rgb24 = new Rgb24(redChannel[i], greenChannel[i], blueChannel[i]);
+            Rgb24 rgb24 = default;
+            ref byte r = ref MemoryMarshal.GetReference(redChannel);
+            ref byte g = ref MemoryMarshal.GetReference(greenChannel);
+            ref byte b = ref MemoryMarshal.GetReference(blueChannel);
+            ref TPixel d = ref MemoryMarshal.GetReference(destination);
 
-                destination[i].FromRgb24(rgb24);
+            for (int i = 0; i < count; i++)
+            {
+                rgb24.R = Unsafe.Add(ref r, i);
+                rgb24.G = Unsafe.Add(ref g, i);
+                rgb24.B = Unsafe.Add(ref b, i);
+                Unsafe.Add(ref d, i).FromRgb24(rgb24);
             }
         }
     }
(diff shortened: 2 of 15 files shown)


src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
```diff
@@ -159,5 +161,45 @@ public partial class PixelOperations<TPixel>
 
             PixelOperations<TDestinationPixel>.Instance.From(configuration, sourcePixels, destinationPixels);
         }
+
+        /// <summary>
+        /// Bulk operation that packs 3 seperate RGB channels to <paramref name="destination"/>.
+        /// The destination must have a padding of 3.
```
antonfirsov, line 167: We expect `destination` to be padded, because of quirks of the Rgb24 AVX2 implementation. This will need some extra trickery (a copy) when using it arond the last line(s) of Jpeg, but that shouldn't be too hard.

.gitattributes
```diff
@@ -105,11 +105,8 @@
 *.pvr               binary
 *.snk               binary
 *.tga               binary
-*.tif               binary
```
JimBobSquarePants: This change comes from the automatic copy from the shared-infrastructure submodule when building. 

I think you need to rebuild the solution and push again making sure all submodules are up-to-date and that the root `.gitattributes` and `.editorconfig` files are updated from the submodule on build.

JimBobSquarePants: I can do this if you want?

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
JimBobSquarePants: Yeah when I did the slice from 4 to 3channels I offset the write so there wasn't any padding required but that could only be done with SSE so likely wouldn't be fast enough here.

.gitattributes
antonfirsov: Reverted this change.

src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
antonfirsov: The problem is that I'm doing overlapped stores of 256 byte vectors in the following format:

```
rgb1   = <R00, G00, B00, R01, G01, B01, R02, G02 | B02, R03, G03, B03, R04, G04, B04, R05 || B05, G05, R06, G06, B06, R07, G07, B07 | ___, ___, ___, ___, ___, ___, ___, ___>
rgb2   = <R08, G08, B08, R09, G09, B09, R10, G10 | B10, R11, G11, B11, R12, G12, B12, R13 || B13, G13, R14, G14, B14, R15, G15, B15 | ___, ___, ___, ___, ___, ___, ___, ___>
rgb3   = <R16, G16, B17, R17, G17, B17, R18, G18 | B18, R19, G19, B19, R20, G20, B20, R21 || G21, B21, R22, G22, B22, R23, G23, B23 | ___, ___, ___, ___, ___, ___, ___, ___> 
rgb4   = <R24, G24, B24, R25, G25, B25, R26, G26 | B26, R27, G27, B27, R28, G28, B28, R29 || G29, B29, R30, G30, B30, R31, G31, B31 | ___, ___, ___, ___, ___, ___, ___, ___>  
```
Storing `rgb4` would result in an overflow, since `B31` is the last element.

JimBobSquarePants: Would it perhaps be better to vectorize one less round than max and slice accordingly after? I'm concerned about the padding requirement.

antonfirsov: That's something we can do a good solution on the call-site yeah ... but the input span has to be still padded, there is no other way for us to make sure there is no overflow within the `PackFromRgbPlanes` method.

Idea:
- Define a utility `Buffer2D.TryGetPaddedRowSpan(y, padding)`, which will return `true` for everything except the last row, or last few rows for the `width < padding` corner case.
- If there is no padding do an automatic copy in `PackFromRgbPlanes()` instead of throwing an exception. (Slow(er) path for the last row)

JimBobSquarePants: Sounds good to me!

JimBobSquarePants:
@antonfirsov I can't see any issues here. Love how quickly you did this, would have taken me weeks! 👍

PR #1465, opened 2020-12-07
Optimize Convolution

JimBobSquarePants:
### Prerequisites

### Description

Refactors all the convolution algorithms to take advantage of bulk operations. 

Operationally the way the algorithm works has changed as follows:
```
Before
Per Interest Row
        |____ Per Interest Column
                |____ Per Kernel Row
                        |____ Per Kernel Column

After
Per Interest Row
        |____ Per Kernel Row
                |____ Per Interest Column
                        |____ Per Kernel Column
```

This allows us to use our bulk conversion methods per row.

Benchmarks are very healthy.

**Before**
blur-before

**After**
blur-after


src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.Buffer{T}.cs
```diff
@@ -53,8 +53,13 @@ public override Span<T> GetSpan()
                 {
                     ThrowObjectDisposedException();
                 }
-
+#if SUPPORTS_CREATESPAN
```
JimBobSquarePants, line 56: This came up during profiling. On NET Core 3.1 We can use a shortcut since we know the length.

src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.Buffer{T}.cs
```diff
@@ -53,8 +53,13 @@ public override Span<T> GetSpan()
...
+                ref byte r0 = ref MemoryMarshal.GetReference<byte>(this.Data);
```
Sergio0694, line 57: Small note - this is implicitly creating a `Span<byte>` from the array. When we add .NET 5 support we can optimize this even further by using `MemoryMarshal.GetArrayDataReference` instead and completely bypass all checks 😊

JimBobSquarePants: Yeah... I was looking at that. I still don't know what to do re .NET 5 though. It's be our first target framework which isn't LTS.

Sergio0694: Ah, right. The way I see it, we could maybe do:
- Add a .NET 5 target to start preparing the codebase for .NET 6. When .NET 6 lands, immediately switch to it and drop .NET 5.
- Just hold our breath until .NET 6 lands and stick to .NET Core 3.1 as max until then.

I'd say it might be worth doing an initial investigation for starters, to see how many places in the lib could actually benefit from .NET 5 exclusive APIs that are not on .NET Core 3.1? Then we could decide based on that 🙂

JimBobSquarePants: ARM intrinsics is a big one.

Sergio0694: Ah, right, right. Yeah that alone might be worth it. Properly supporting ARM64 gives a ton of visibility right now, plus being ready for .NET 6 would be very nice. And if we are to work on ARM64 right now anyway, I don't see the harm in publishing .NET 5 packages already, assuming we'll get a working build soon. After all, companies would likely remain on .NET Core 3.1 until .NET 6 is out, so those wouldn't be affected. And devs on .NET 5 now would be fine with having to jump to .NET 6 immediately anyway, after all they jsut did the same with .NET 5 right now. I'd say if you have time you could look into this and then when you have a branch that shows a nice boost on ARM64, consider shipping if it's really good? 😄

src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.Buffer{T}.cs
```diff
@@ -53,8 +53,13 @@ public override Span<T> GetSpan()
...
+                return MemoryMarshal.CreateSpan(ref Unsafe.As<byte, T>(ref r0), this.length);
+#else
                 return MemoryMarshal.Cast<byte, T>(this.Data.AsSpan()).Slice(0, this.length);
```
Sergio0694, line 60: In theory here we know that the `byte` size is an exact multiple of our target length for the `T` span, so I was thinking we could be able to remove that final `Slice` call by pre-slicing the `Data` array. Something like:

```csharp
return MemoryMarshal.Cast<byte, T>(this.Data.AsSpan(0, this.length * sizeof(T)));
```

Tried out on sharplab ([here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8YUCAAcAymIBuvMDFwBuGvSasAchwFQFKtYykALbFAkAZbMHbc+A1dTUBmJqQYBhGgG8aDfwwSerLYGDAMwACeYQDaALoMACah2I4BgcGh4bw8DAA2MFwA5hjGaQF+ATEAUrwYAOKFYgoAFBiREjAQAGYtAGowYBjQKACUo3GV/sSu3LjY3eFSEthcADwAKgB8DI0Yy6vkm1sto1PpDADuxmLhGwwgDNz4q9hFMInnvtQXAcQA7AwALIwfDQSJAsy4Ux5FgePAYNZRMJoBjbNrGXi4FjJDDYFgAQVwBy4p1GLCkeQUMBadFRpSxLAKxVKo3K/gAvudzjU6o0uM0wG0Ol1egMhiNxpMfn9Zlx5osGCTjrsYPsVlxSMdTud0tdbmiHk8uC8uG8Pl9dX9ASCwVAIVCYXCEUjojBUeiGdjcfiiSTafTMdjmSVjAwAFQMXC8ABeopaG3GbPOXOoHKAA==)) but apparently the codegen is worse for some reason. So yeah I'm not really seeing a way to improve this without having access to the .NET 5 APIs 🤔

src/ImageSharp/Primitives/DenseMatrix{T}.cs
```diff
@@ -109,7 +109,7 @@ public DenseMatrix(T[,] data)
         /// <returns>The <see typeparam="T"/> at the specified position.</returns>
         public ref T this[int row, int column]
         {
-            [MethodImpl(InliningOptions.ShortMethod)]
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
Sergio0694, line 112: Why is it `AggressiveInlining` here and not the internal `ShortMethod`? Was this intentional or did you just forget to change this back after your initial experiments?

src/ImageSharp/Processing/Processors/Convolution/Convolution2DProcessor{TPixel}.cs
```diff
@@ -43,12 +40,12 @@ internal class Convolution2DProcessor<TPixel> : ImageProcessor<TPixel>
         }
 
         /// <summary>
-        /// Gets the horizontal gradient operator.
+        /// Gets the horizontal convolution kernel.
```
Sergio0694, line 43: ❤

src/ImageSharp/Processing/Processors/Convolution/Convolution2DProcessor{TPixel}.cs
```diff
@@ -60,102 +57,39 @@ internal class Convolution2DProcessor<TPixel> : ImageProcessor<TPixel>
...
             var interest = Rectangle.Intersect(this.SourceRectangle, source.Bounds());
-            var operation = new RowOperation(interest, targetPixels, source.PixelBuffer, this.KernelY, this.KernelX, this.Configuration, this.PreserveAlpha);
 
-            ParallelRowIterator.IterateRows<RowOperation, Vector4>(
-                this.Configuration,
-                interest,
-                in operation);
+            // We use a rectangle 3x the interest width to allocate a buffer big enough
+            // for source and target bulk pixel conversion.
+            var operationBounds = new Rectangle(interest.X, interest.Y, interest.Width * 3, interest.Height);
```
Sergio0694, line 69: I'm a bit confused about the temporary memory usage of this solution, as I remember you being a bit worried about this back when I was first working on the bokeh blur processor. Wouldn't this mean we're now allocating a `3 x [image size]` buffer every time now? As in, if I'm processing a `1920x1080` image, this would allocate a `5760 x 1080` temporary buffer, correct? Is that ok? At 4K image that'd be a buffer of 24 million pixels 🤔

Looking at the rest of the code I think I get why you're doing this (as it allows you to access pixels on the Y access in row major order), but I'd be curious about a diff in max peak memory usage compared to `master`, if you have made such a benchmark? If nothing else, I think it'd be an interesting bit of into to include 🙂

src/ImageSharp/Processing/Processors/Convolution/Convolution2PassProcessor{TPixel}.cs
```diff
@@ -42,12 +42,12 @@ internal class Convolution2PassProcessor<TPixel> : ImageProcessor<TPixel>
...
-        /// Gets the horizontal gradient operator.
+        /// Gets the horizontal convolution kernel.
         /// </summary>
         public DenseMatrix<float> KernelX { get; }
 
         /// <summary>
-        /// Gets the vertical gradient operator.
+        /// Gets the vertical convolution kernel.
         /// </summary>
         public DenseMatrix<float> KernelY { get; }
```
Sergio0694, line 52: Small side note, I'm still convinced that using `DenseMatrix<T>` for separable 1D kernels makes the code less intuitive (those are 1D vectors, not matrices) and potentially more overhead for the unnecessary 2D coordinate calculation. I think we could possibly look into having a different DenseVector<T> type in the future that would just map to a 1D vector, to use in cases such as this.

src/ImageSharp/Processing/Processors/Convolution/ReadOnlyKernel.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace SixLabors.ImageSharp.Processing.Processors.Convolution
+{
+    /// <summary>
+    /// A stack only, readonly, kernel matrix that can be indexed without
+    /// bounds checks when compiled in release mode.
+    /// </summary>
+    internal readonly ref struct ReadOnlyKernel
```
Sergio0694, line 15: Also to get back to my point above with respect to clarity, imho this should be called `ReadOnlyKernel2D`.

Sergio0694 (approved):
Looks great, and love the speed improvements!
Just left a few notes with respect to code style and other minor things. My main concern is more about peak memory usage int his case, so curious to see more info on that front. Will get to work on the bokeh blur version using this new pattern once this is merged 😄

src/ImageSharp/Primitives/DenseMatrix{T}.cs
JimBobSquarePants: Intentional. `ShortMethod` was initially designed for certain jpeg profiling but it kinda spread. I don't think we need it any more wit the tooling we have available.

src/ImageSharp/Processing/Processors/Convolution/Convolution2DProcessor{TPixel}.cs
JimBobSquarePants: We're actually only allocating a single of `Span<Vector4>` of length `3x interest width` per parallel region and only for 2D convolution. For 1D or 2Pass it's  `1x interest width`.

src/ImageSharp/Processing/Processors/Convolution/Convolution2PassProcessor{TPixel}.cs
JimBobSquarePants: It makes my life easier as I can use the same code throughout when doing things like offset mapping. 

We also use the same code for the non-separable convolution operation so that can be arbitrary dimensions.

src/ImageSharp/Processing/Processors/Convolution/ReadOnlyKernel.cs
JimBobSquarePants: See above, there's only be one type.

src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.Buffer{T}.cs
JimBobSquarePants: Yep .NET 5 makes this easier.

src/ImageSharp/Primitives/DenseMatrix{T}.cs
Sergio0694: Ah, got it. Well we might want to revert that in another PR then 😄

src/ImageSharp/Processing/Processors/Convolution/Convolution2DProcessor{TPixel}.cs
Sergio0694: Oooh I see, yeah that makes sense. Perfect then! 🚀

PR #1475, opened 2020-12-12
Optimize bokeh blur convolution

Sergio0694:
### Prerequisites

### Description

Optimized the bokeh blur effect, with a more specialized variation of the changes introduced in #1465.

### Benchmark

image

**EDIT:** updated benchmark after https://github.com/SixLabors/ImageSharp/pull/1475/commits/2e53a44acf4ab393e9827f41e74f61b8459e5dfe.

image

**EDIT:** updated benchmark after https://github.com/SixLabors/ImageSharp/pull/1475/commits/e4ba017dcf245d57e2e9c4faca3bc3390024bcfa.

image

**EDIT:** updated benchmark after https://github.com/SixLabors/ImageSharp/pull/1475/commits/442e46781196276ce4a2c96b18557508c67f53ff and https://github.com/SixLabors/ImageSharp/pull/1475/commits/c38fc814a856166ad5c48d7eaceeac4c19f38fcf.

image


Sergio0694:
Investigating the test failures with the effect on smaller image regions, and also noticed a previously existing bug that was likely causing some out of bounds writes in the past, very weird we hadn't spotted that one before. Glad it popped up now! 😄

Sergio0694:
Removed the `WorksWithDiscoBuffers` test for the bokeh blur processor, as it was just always throwing due to insufficient memory. @JimBobSquarePants I guess you could reintroduce it later on if you wanted to, but for now I wanted to have the CI run to verify all the other tests and the changes in this PR. Also, noticed that eg. the gaussian blur doesn't have this test either, so... 🤷‍♂️

https://github.com/SixLabors/ImageSharp/blob/ff94d20ceda7d0aedc6af7e0e801c6444012128b/tests/ImageSharp.Tests/Processing/Processors/Convolution/BokehBlurTest.cs#L177-L183

JimBobSquarePants:
>Removed the WorksWithDiscoBuffers test for the bokeh blur processor, as it was just always throwing due to insufficient memory.

This is concerning. We weren't getting any throws for this in previous runs.

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
```diff
@@ -99,8 +99,8 @@ public void Dispose()
         {
             if (!this.isDisposed)
             {
-                this.yOffsets.Dispose();
-                this.xOffsets.Dispose();
+                this.yOffsets?.Dispose();
```
JimBobSquarePants, line 102: Are you failing to call `BuildSamplingOffsetMap` at some point?

Sergio0694: Yes, in that `WorksWithDiscoBuffers` test. One of the `Allocate` calls in the `BuildOffsetsMap` call was throwing due to insufficient memory, so `Dispose` was called, and then it was throwing here because the other buffer was still `null`, as it crashed before that was assigned 🤔

JimBobSquarePants: You running in 32bit? How big are the buffers?

Sergio0694: The one crashing is for `xOffsets`, which is `100` in width and `65` in kernel size, so `6500` in total.
I'm running in 64-bit by default (double checked and `sizeof(IntPtr) == 8`).

JimBobSquarePants: Hmmm... That should be nothing. I'll pull it down tomorrow and see what is happening.

Sergio0694:
Reverted to draft as I just had some new ideas for some other possibly juicy optimizations to add 😋

Sergio0694:
Alright, happy with the progress so far, I'd say this PR is good for review, and we can always revisit later on 🎉

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
JimBobSquarePants:
``` 
Requested allocation: 19500 elements of Int32 is over the capacity of the MemoryAllocator
```

JimBobSquarePants: That number is given by bounds.Height of 300 x kernel.Height of 65. Is that large a kernel expected?

JimBobSquarePants: So... You're actually requesting 78000 bytes during the request which is not an unreasonable amount. I think that test limit is simply set far to low. 

The available allocation is calculated as follows:

`no. of rows * width * sizeof(pixel)`

Now since the test is using a `300 * 100 * Bgr24` with a limit of 41 rows this means the maximum allowable buffer is. 

`41 * 100 * 3 = 12300` bytes which is far too low. We need 260 rows which still allows for discontiguous testing.

Sergio0694: > Is that large a kernel expected?

The default radius is 32, so yeah that matches the expected size for the default parameters 🙂

JimBobSquarePants (approved):
LGTM. Very nice improvement! 🚀

PR #1477, opened 2020-12-15
1D convolution optimization and general codegen tweaks

Sergio0694:
### Prerequisites

### Description
This PR does a few things:

- Speed optimizations to the 2D pass convolution processor (powering gaussian blur, sharpen, etc.)
- Speed optimizations to the bokeh blur
- Some general codegen optimizations that should apply to all common pixel conversions, etc.

### Benchmarks

Here's a preview of the current improvements for the gaussian blur benchmark:

image

And here's some more bokeh blur optimizations compared to master, after #1475 got merged:

image


src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
```diff
@@ -90,4 +94,4 @@ public static void Compress(ref Vector4 vector)
         [MethodImpl(InliningOptions.ShortMethod)]
         public static float Compress(float channel) => channel <= 0.0031308F ? 12.92F * channel : (1.055F * MathF.Pow(channel, 0.416666666666667F)) - 0.055F;
```
JimBobSquarePants, line 95: If we ever figure out how to do an accurate SIMD enable approximation of this we would be laughing.

src/ImageSharp/Processing/Processors/Convolution/Convolution2PassProcessor{TPixel}.cs
```diff
@@ -64,44 +59,364 @@ protected override void OnFrameApply(ImageFrame<TPixel> source)
             // for source and target bulk pixel conversion.
             var operationBounds = new Rectangle(interest.X, interest.Y, interest.Width * 2, interest.Height);
 
-            using (var mapX = new KernelSamplingMap(this.Configuration.MemoryAllocator))
+            // We can create a single sampling map with the size as if we were using the non separated 2D kernel
+            // the two 1D kernels represent, and reuse it across both convolution steps, like in the bokeh blur.
+            using var mapXY = new KernelSamplingMap(this.Configuration.MemoryAllocator);
+
+            mapXY.BuildSamplingOffsetMap(this.Kernel.Length, this.Kernel.Length, interest);
```
JimBobSquarePants, line 66: I was initially concerned about the increased size of the map but doing the math I think it's acceptable.

JimBobSquarePants (approved):
Very, very nice! 🚀

src/ImageSharp/Processing/Processors/Convolution/Convolution2PassProcessor{TPixel}.cs
Sergio0694: From what I see in the actual map code, the allocations are the same as before, just done at the same time, no?
As in, looking at the allocation bit in the `KernelSamplingMap` class:

https://github.com/SixLabors/ImageSharp/blob/cd1a25efeeb4b243cbfcfff8950fc215523077b7/src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs#L44-L45

I did this change mostly just to avoid the extra computations for the horizontal/vertical maps that would not be used at all when doing either the horizontal or vertical passes for the specialized 1D convolution processors 😄

JimBobSquarePants: Haha... yeah you're right! Awesome.

src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
benaadams: pow(channel, 0.416666666666667F) => exp(channel * log(0.416666666666667F))

log(0.416666666666667F) == -0.875468737353899935628f

So...
```csharp
public static void Compress(ref Vector4 vector)
{
    var channels = Unsafe.As<Vector4, Vector128<float>>(ref vector);
    var log = Vector128.Create(-0.875468737353899935628f);

    channels = Sse.Multiply(channels, log);

    channels = Exp(channels); // Isn't simd intrinsic

    if (Fma.IsSupported)
    {
        channels = Fma.MultiplyAdd(Vector128.Create(1.055F), channels, Vector128.Create(-0.055F));
    }
    else
    {
        channels = Sse.Add(Sse.Multiply(Vector128.Create(1.055F), channels), Vector128.Create(-0.055F));
    }

    Unsafe.As<Vector4, Vector128<float>>(ref vector) = channels;
}
```
But `Exp` isn't a Simd intrinsic; however you can approximate it with these sequences [sse_mathfun](https://github.com/RJVB/sse_mathfun/blob/0f462645698320643decdb5f79d211733b2a97d6/sse_mathfun.h#L393-L469) or [avx_mathfun](https://github.com/reyoung/avx_mathfun/blob/be617bbcf66993c4f7e7d267ed9926cd8e7d4f02/avx_mathfun.h#L250-L304)?

Issue #1433, opened 2020-11-20
Optimize pixel blending with integer arithmetics


antonfirsov:
### Problem
The current `float` and `Vector4` -based pixel blender API does not give us too much space for introducing the rasterization perf improvements needed for SixLabors/ImageSharp.Drawing#102. With our current bulk API, the maximum we can do is to process 2 pixels in one AVX batch, since we can fit only 8 `float`-s into one AVX register.  This means that the expected speedup for blending is around or below 2x. With this we would keep lagging behind Skia and GDI significantly.

### Idea
We should explore API-s and implementations working with `UInt16`-based fixed point arithmetics. This is technically very similar to approach taken by the libjpeg decoder SIMD pipelines which we eventually also want to adapt. In theory, `UInt16`-based bulk processing should reduce the time spent in pixel blenders by ~4x (or more) when AVX2 is present.

This will require API additions similar to the following:
```C#
public abstract class PixelBlender<TPixel> 
{
    public void Blend<TPixelSrc>(
            Configuration configuration,
            Span<TPixel> destination,
            ReadOnlySpan<TPixel> background,
            ReadOnlySpan<TPixelSrc> source,
            // 'amount' is scaled to 0-255. Could be byte, but with UInt16 we will avoid some unnecessary conversions
            ReadOnlySpan<UInt16> amount); 

    protected virtual BlendFunction(
            Configuration configuration,
            Span<Rgba32> destination,
            ReadOnlySpan<Rgba32> background,
            ReadOnlySpan<Rgba32> source,
            ReadOnlySpan<UInt16> amount);
}


public static class PorterDuffFunctions
{
    /*public*/ static Vector4 NormalSrcOver(Span<Rgba32> destination,
            ReadOnlySpan<Rgba32> background,
            ReadOnlySpan<Rgba32> source,
            ReadOnlySpan<UInt16> opacity);
}
```

**Update:** 
In the first API variant there was a type `ScaledUInt16Vector4`, but after thinking it through, I realized it is unnecessary. We should work with `Rgba32` to maximize perf.

JimBobSquarePants:
Agreed. This seems to be the smartest approach while keeping work on the CPU. Other libraries [defer to the GPU for perf](https://github.com/google/skia/blob/master/src/gpu/effects/GrPorterDuffXferProcessor.cpp)

JimBobSquarePants:
Pinning this as It's super important for ImageSharp.Drawing

[...]

PR #1481, opened 2020-12-18
Much Faster sRGB Companding

JimBobSquarePants:
### Prerequisites

### Description

Updates the bulk sRGB Companding method to use the similar LUT approach as MagicScaler (_thanks @saucecontrol !_ )

Results are awesome. Massive speedup in real world scenarios.

companding

I noticed some oddities (_I'd previously introduced_) in the existing unit tests during the work which I cleaned up also.

### DIFF
--- a/src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
+++ b/src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
@@ -69,8 +69,9 @@ namespace SixLabors.ImageSharp.ColorSpaces.Companding
             return result;
         });
 
-        private static readonly float[] ExpandTable = LazyExpandTable.Value;
-        private static readonly float[] CompressTable = LazyCompressTable.Value;
+        private static float[] ExpandTable => LazyExpandTable.Value;
+
+        private static float[] CompressTable => LazyCompressTable.Value;
 
         /// <summary>
         /// Expands the companded vectors to their linear equivalents with respect to the energy.
--- a/src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
+++ b/src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
@@ -5,6 +5,10 @@ using System;
 using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+#if SUPPORTS_RUNTIME_INTRINSICS
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.X86;
+#endif
 
 namespace SixLabors.ImageSharp.ColorSpaces.Companding
 {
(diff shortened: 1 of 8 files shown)


src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
```diff
@@ -18,49 +22,114 @@ namespace SixLabors.ImageSharp.ColorSpaces.Companding
...
+                    d = Math.Pow((d + 0.055) / 1.055, 2.4);
+                }
+
+                result[i] = (float)d;
+            }
+
+            return result;
+        });
+
+        private static readonly float[] ExpandTable = LazyExpandTable.Value;
```
antonfirsov, line 72: Why Lazy if this is a field?

antonfirsov:
@saucecontrol can we use this approach for pixel blending? Seems cheaper than #1433.

src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
JimBobSquarePants: Yeah, doesn't need to be,. Will update.

antonfirsov: I would make `ExpandTable` a property.  The buffers seem to be huge, lazy init makes sense here. (16kbytes if I get it right)

JimBobSquarePants: Updated.

src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
```diff
@@ -18,49 +22,115 @@ namespace SixLabors.ImageSharp.ColorSpaces.Companding
...
+                else
+                {
+                    d = (1.055 * Math.Pow(d, 1.0 / 2.4)) - 0.055;
+                }
+
+                result[i] = (float)d;
+            }
+
+            return result;
+        });
```
antonfirsov, line 48:
```suggestion
        }, LazyThreadSafetyMode.PublicationOnly);
```

src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
```diff
@@ -18,49 +22,115 @@ namespace SixLabors.ImageSharp.ColorSpaces.Companding
...
+                else
+                {
+                    d = Math.Pow((d + 0.055) / 1.055, 2.4);
+                }
+
+                result[i] = (float)d;
+            }
+
+            return result;
+        });
```
antonfirsov, line 70:
```suggestion
        }, LazyThreadSafetyMode.PublicationOnly);
```

tests/ImageSharp.Tests/PixelFormats/PixelOperations/PixelOperationsTests.cs
```diff
@@ -1177,14 +1152,22 @@ public void Verify()
...
+                else if (typeof(IPixel).IsAssignableFrom(typeof(TDest)))
+                {
+                    Span<TDest> expected = this.ExpectedDestBuffer.AsSpan();
+                    Span<TDest> actual = this.ActualDestBuffer.GetSpan();
+                    var comparer = new ApproximateFloatComparer(TestEnvironment.Is64BitProcess ? 0.0001F : 0.001F);
 
-                        // ReSharper restore PossibleNullReferenceException
+                    for (int i = 0; i < count; i++)
+                    {
+                        Assert.Equal((IPixel)expected[i], (IPixel)actual[i], comparer);
```
antonfirsov, line 1170: I don't really understand this path.

How can [`ApproximateFloatComparer`](https://github.com/SixLabors/ImageSharp/blob/a1784a67bf492b1b02e99655a39aa5e5acf81988/tests/ImageSharp.Tests/TestUtilities/ApproximateFloatComparer.cs) compare `IPixel`-s?

If I understand your intent right, you wanted to loosen the strictness for the cases where `TDest` is a pixel (almost everything). I don't think it's always a good idea. For example `Rgb32 -> Rgb24` conversion should be always exact.

antonfirsov (commented):
Let's make the lazy code thread-safe + a major concern regarding tests.

Can't comment on the SIMD code (too much time to review properly), but I trust the test coverage as usual.

src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
antonfirsov: Ok, realized it's unnecessary, since writing `float[]` (`object`) should be atomic.

tests/ImageSharp.Tests/PixelFormats/PixelOperations/PixelOperationsTests.cs
JimBobSquarePants: They will be exact at that bit depth since our approach uses 16bits per component. I noticed very minor fluctuations for Short4 and HalfVector4 though in one of the longer array runs. Small enough to disappear but large enough to fail the exact comparer.

JimBobSquarePants: When we attach more metadata to a TPixel type we could be more explicit

antonfirsov: I missed that you added `ApproximateFloatComparer : IEqualityComparer<IPixel>` in my previous comment.

Still somewhat concerned about tolerant comparison in cases we don't need it. I would add a cheap heuristic approach like this:
- To take a default-valued parameter `preferExactComparison = true` on `TestOperation` and pass it forward to `Verify`
- Pass `false` in `FromCompandedScaledVector4` and similar
- Change the condition:

```C#
else if (typeof(IPixel).IsAssignableFrom(typeof(TDest)) && !preferExactComparison && (IsComplexPixel<TSource>() || IsComplexPixel<TDest>()))
```

and 

```C#
static bool IsComplexPixel<TPixel>() => Unsafe.SizeOf<TPixel>() > sizeof(int);
```

JimBobSquarePants:
```
static bool IsComplexPixel<TPixel>() => Unsafe.SizeOf<TPixel>() > sizeof(int);
```

Nearly... There's a few types that would fail this test. I've written a version that explicitly captured them. We really need a `PixelTypeInfo.BitsPerComponent` property.

src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
JimBobSquarePants: I've explicitly added the "true" second parameter.

saucecontrol:
> @saucecontrol can we use this approach for pixel blending? Seems cheaper than #1433.

Yep, I use it for all my linear pixel conversions.  It also works great in combination with fixed-point math, which I still use as a fallback on non-RyuJIT netfx.  In that case, you can have a 15-bit LUT with direct lookups and no interpolation.

src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
saucecontrol: `true` is also redundant.  Thread-safe is the default mode: https://source.dot.net/#System.Private.CoreLib/Lazy.cs,236

JimBobSquarePants: I remember checking that back in reference source when I did the other stuff but must have misread. I'll leave it for now as it's doing no harm. I can cleanup another time.

PR #1490, opened 2021-01-03
Target Image<TPixel> loading/decoding APIs

Sergio0694:
### Prerequisites

### Description
This PR adds new APIs to load/decode images directly into reused `Image<TPixel>` instances.

### Details
I've decided to add the overloads for the `Image<TPixel>` class to improve interoperability with other APIs.
I didn't feel great about either adding a new, separate `IImageMemoryDecoder` type, so this should help with consustency 😄

<details>
    <summary><b>API additions (click to expand):</b></summary>
<br/>

```csharp
namespace SixLabors.ImageSharp.Formats
{
    public interface IImageDecoder
    {
        void Decode<TPixel>(Stream stream, Image<TPixel> image)
        Task DecodeAsync<TPixel>(Stream stream, Image<TPixel> image, CancellationToken cancellationToken)
    }
}

{
    public abstract partial class Image
    {
        // byte[]
        public static void Load<TPixel>(byte[] data, Image<TPixel> image, out IImageFormat format)
        public static void Load<TPixel>(byte[] data, Image<TPixel> image, IImageDecoder decoder)

        // ReadOnlySpan<byte>
        public static unsafe void Load<TPixel>(
            ReadOnlySpan<byte> data,
            Image<TPixel> image,
            IImageDecoder decoder)
        public static unsafe void Load<TPixel>(
            ReadOnlySpan<byte> data,
            Image<TPixel> image,
            out IImageFormat format)

        // Stream
        public static void Load<TPixel>(Stream stream, Image<TPixel> image, IImageDecoder decoder)
        public static Task LoadAsync<TPixel>(
            Stream stream,
            Image<TPixel> image,
            IImageDecoder decoder,
            CancellationToken cancellationToken = default)
        public static void Load<TPixel>(Stream stream, Image<TPixel> image, out IImageFormat format)
        public static async Task<IImageFormat> LoadWithFormatAsync<TPixel>(
            Stream stream,
            Image<TPixel> image,
            CancellationToken cancellationToken = default)

        // string
        public static Task LoadAsync<TPixel>(
            string path,
            Image<TPixel> image,
            IImageDecoder decoder,
            CancellationToken cancellationToken = default)
        public static void Load<TPixel>(string path, Image<TPixel> image, out IImageFormat format)
        public static void Load<TPixel>(string path, Image<TPixel> image, IImageDecoder decoder)
    }
}
```
</details>

> **NOTE:** opening the PR to have the CI run on it, but I still need to add unit tests.

### Use case example

The new APIs allow for eg. GPU processing (assuming DirectX 12 buffers on NUMA architecture) like so (using ComputeSharp):

```csharp
IImageInfo info = Image.Identify("cat.jpg", out IImageFormat format);

using ReadWriteBuffer<uint> gpuBuffer = Gpu.Default.AllocateReadWriteBuffer<uint>(info.Width * info.Height);

using (var buffer = gpuBuffer.GetBufferForWrite())
fixed (void* pointer = buffer.Span)
using (var image = Image.WrapMemory<Rgb24>(pointer, info.Width, info.Height))
{
    Image.Load("cat.jpg", image, out _);
}

// Do GPU processing...

using (var buffer = gpuBuffer.GetBufferForRead())
fixed (void* pointer = buffer.Span)
using (var image = Image.WrapMemory<Rgb24>(pointer, info.Width, info.Height))
{
    image.SaveAsJpeg("cat_processed.jpg");
}
```

The big advantage is that we can completely skip one entire copy of the image, as we can use the new APIs to load/decode the image directly into the temporary upload buffer on the GPU. Without these APIs, we'd instead have to load the image into CPU memory, then copy from there to the temporary upload GPU buffer, then copy again from there to the shader visible GPU buffer.

### Open questions

❓ Right now I just throw a `NotSupportedException` for GIF files. Is this ok?
❓ Right now I just throw an `ArgumentException` if the image has the wrong size. Do we want to try to mutate it instead?


src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -158,6 +157,11 @@ public Image<TPixel> Decode<TPixel>(BufferedReadStream stream, CancellationToken
             return image;
         }
 
+        /// <inheritdoc />
+        public void Decode<TPixel>(BufferedReadStream stream, Image<TPixel> image, CancellationToken cancellationToken)
+            where TPixel : unmanaged, IPixel<TPixel>
+            => throw new NotSupportedException("The current decoder doesn't support decoding to a target image");
```
JimBobSquarePants, line 163: Why not decode a single frame?

DaZombieKiller: What to do in the case of a multi-frame image was something I was unsure about in the initial discussion proposal as well. Decoding the first frame is definitely what I'd expect when using the API. It might also make sense to offer additional overloads for specifying a frame index, but I'm not completely sold on that being the best approach.

src/ImageSharp/Image.cs
```diff
@@ -67,7 +67,7 @@ protected Image(Configuration configuration, PixelTypeInfo pixelType, ImageMetad
         public int Height => this.size.Height;
 
         /// <inheritdoc/>
-        public ImageMetadata Metadata { get; }
+        public ImageMetadata Metadata { get; internal set; }
```
DaZombieKiller, line 70: Should there to be a public API for setting this somewhere? With the new setter being `internal`, a third party image decoder can't properly re-use an `Image` as it can't change the metadata.

JimBobSquarePants: You set individual properties on the metadata not the metadata itself. 

See [ImageMetadata](https://docs.sixlabors.com/api/ImageSharp/SixLabors.ImageSharp.Metadata.ImageMetadata.html#properties)

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
JimBobSquarePants: I would probably leave it at frame 0 for now. Having the value settable might well complicate things to much for a single PR.

src/ImageSharp/Image.cs
DaZombieKiller: I see, that makes sense. Seeing different `*Metadata` classes had me confused for a moment, not noticing that `ImageMetadata` is sealed.

src/ImageSharp/Formats/IImageDecoder.cs
```diff
@@ -24,6 +24,16 @@ public interface IImageDecoder
...
             where TPixel : unmanaged, IPixel<TPixel>;
 
+        /// <summary>
+        /// Decodes the image from the specified stream to an <see cref="Image{TPixel}"/> of a specific pixel type.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="stream">The <see cref="Stream"/> containing image data.</param>
+        /// <param name="image">The target <see cref="Image{TPixel}"/> instance.</param>
+        // TODO: Document ImageFormatExceptions (https://github.com/SixLabors/ImageSharp/issues/1110)
+        void Decode<TPixel>(Stream stream, Image<TPixel> image)
```
antonfirsov, line 34: This is gonna be a breaking change for users who implemented their `IImageDecoder`. The reason @DaZombieKiller proposed `IImageMemoryDecoder` was to avoid this.

I'm undecided on this, with a slight preference to not break.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -1087,27 +1104,37 @@ private ushort ReadUint16(BufferedReadStream stream)
...
+                    this.Configuration,
+                    this.ImageWidth,
+                    this.ImageHeight,
+                    this.Metadata);
             }
+            else
+            {
+                if (this.ImageWidth != image.Width || this.ImageHeight != image.Height)
+                {
+                    ThrowHelper.ThrowArgumentException("The input image has an invalid size", nameof(image));
```
antonfirsov, line 1129:
```suggestion
                    ThrowHelper.ThrowArgumentException("The input image should match the dimensions of the image in the stream.", nameof(image));
```

Same for the other decoders. We can throw on this much earlier in the decoding workflow.

antonfirsov (commented):
We should decide on API questions before proceeding.

I like the idea of having a `Load(Async)` overloads that work on an existing `Image<T>` instance over the static variants in the original proposal, since they separate the concerns of wrapping the memory and loading the image. This reduces the number of overloads we need to create and maintain to `NumOf(WrapVariants) + NumOf(LoadVariants)` from `NumOf(WrapVariants) * NumOf(LoadVariants)`. (The numbers will likely keep growing.) On the other hand, the user needs to know the dimensions in advance, I'm not sure if it's always possible. @DaZombieKiller? (Also see my notes in the very end.)

If we stick with the instance API, we need to decide on the location of the new overloads. Here my preference is to have them as extension methods in either `AdvancedImageExtensions` or a new extension class under the `Advanced` namespace.

We need to decide if we are ok with the breaking changes of extending `IImageDecoder` (break a small number of advanced users to deliver a feature for another group of advanced users ... and kinda also break our promise to not break).

@JimBobSquarePants suggestions?

----

> Right now I just throw an ArgumentException if the image has the wrong size. Do we want to try to mutate it instead?

If the stream image size is bigger, we clearly need to throw. For other cases, I see 3 options here:
1. The user wants to change dimensions (mutate)
2. The user wants to decode to a sub rectangle of the target image
3. The user is expecting an exception if sizes mismatch (current behavior)

DaZombieKiller:
@antonfirsov 
> On the other hand, the user needs to know the dimensions in advance, I'm not sure if it's always possible. @DaZombieKiller?

In my use case they're always known as I just use `Image.Identify` to work out the dimensions and format so I can create a texture to fill.

src/ImageSharp/Formats/IImageDecoder.cs
JimBobSquarePants: Hmmmm.... It's a fairly fundamental break and there are decoder out in the wild. I say new interface.

DaZombieKiller: In that case, now comes the hard part: naming it. `IImageMemoryDecoder` doesn't seem descriptive enough, especially since it's no longer operating on `Memory<T>` like in the original proposal. Maybe `IImageTargetDecoder` or `IImageTargetedDecoder`, based on the name of this PR?

JimBobSquarePants: We'd have to expose the interface and use the type as the method parameter type. Can't call what you don't have then.

DaZombieKiller: What about the overloads that don't take an `IImageDecoder` currently? Would they just be removed?

JimBobSquarePants: When someone calls the new `Image.Load` without a specified decoder we will loop through our config to find a matching decoder that implements the interface. If there is none it will through. 

I generally don't recommend people use specific decoders unless they have to as they skip the format validation checks.

src/ImageSharp/Image.FromStream.cs
```diff
@@ -441,7 +442,28 @@ public static async Task<(Image<TPixel> Image, IImageFormat Format)> LoadWithFor
...
+        /// <param name="image">The target <see cref="Image{TPixel}"/> instance.</param>
+        /// <param name="decoder">The decoder.</param>
+        /// <exception cref="ArgumentNullException">The stream is null.</exception>
+        /// <exception cref="NotSupportedException">The stream is not readable.</exception>
+        /// <exception cref="ArgumentException">The size of the image is invalid.</exception>
+        /// <exception cref="NotSupportedException">The image format is not supported for the operation.</exception>
+        /// <exception cref="UnknownImageFormatException">Image format not recognised.</exception>
+        /// <exception cref="InvalidImageContentException">Image contains invalid content.</exception>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        public static void Load<TPixel>(Stream stream, Image<TPixel> image, IImageDecoder decoder)
```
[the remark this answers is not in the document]
JimBobSquarePants: I'd say `AdvancedImageExtensions`

PR #1508, opened 2021-01-17
Vectorize Jpeg Encoder Color Conversion

### LINKED ISSUE #1476: Speed Up Jpeg Encoder Color Conversion
Some analysis of the performance of the encoder based upon a breakdown of this benchmark indicates that encoding a large jpeg takes 80% of the entire processing time. 

https://github.com/kleisauke/net-vips/tree/master/tests/NetVips.Benchmarks

This is due to the lack of hardware acceleration in our color conversion approach. 

The current Jpeg encoder utilizes predefined tables to convert a span of `Rgb24` pixels into separate Y Cb Cr `Block8x8F` planes.

https://github.com/SixLabors/ImageSharp/blob/f1a0fb61798e718314b26af37d4470e7ec381793/src/ImageSharp/Formats/Jpeg/Components/Encoder/YCbCrForwardConverter%7BTPixel%7D.cs#L58-L83

While this is faster than naïve per-pixel floating point calculation it can be heavily optimized. 

### Short Term Goal
Add AVX2? acceleration directly to the converter to optimize conversion for .NET Core 3.1+. This should be a few hours work for someone with SIMD knowledge.

### Long Term Goal 
Establish an architecture similar to the [Jpeg decoder ColorConverters](https://github.com/SixLabors/ImageSharp/tree/f1a0fb61798e718314b26af37d4470e7ec381793/src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters) allowing incremental addition accelerated converters for all platforms and color spaces.

tkp1n:
### Prerequisites

### Description
Touches #1476 with a vectorized implementation of the RGB -> YCbCr conversions.

It is worth noting, that the following benchmarks don't show the entire picture. The current lookup table converter uses 3kB of lookup tables that pollute the cache and likely negatively impact subsequent operations in the encoding chain.

### Benchmark 🚀 

``` ini

Intel Core i7-9750H CPU 2.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.102
  [Host]     : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT
  DefaultJob : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT


```
|            Method |      Mean |    Error |   StdDev | Ratio |
|------------------ |----------:|---------:|---------:|------:|
|        ConvertLut | 224.68 ns | 1.059 ns | 0.938 ns |  1.00 |
| ConvertVectorized |  81.07 ns | 0.404 ns | 0.337 ns |  0.36 |

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterLut.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterLut.cs
@@ -111,7 +111,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
             // float cb = 128F + ((-0.168736F * r) - (0.331264F * g) + (0.5F * b));
             cbResult[i] = (this.CbRTable[r] + this.CbGTable[g] + this.CbBTable[b]) >> ScaleBits;
 
-            // float cr = 128F + ((0.5F * r) - (0.418688F * g) - (0.081312F * b))
+            // float cr = 128F + ((0.5F * r) - (0.418688F * g) - (0.081312F * b));
             crResult[i] = (this.CbBTable[r] + this.CrGTable[g] + this.CrBTable[b]) >> ScaleBits;
         }
 
--- a/src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterLut.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterLut.cs
@@ -1,16 +1,17 @@
 // Copyright (c) Six Labors.
 // Licensed under the Apache License, Version 2.0.
 
+using System;
 using System.Runtime.CompilerServices;
+using SixLabors.ImageSharp.PixelFormats;
 
 namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
 {
     /// <summary>
     /// Provides 8-bit lookup tables for converting from Rgb to YCbCr colorspace.
     /// Methods to build the tables are based on libjpeg implementation.
-    /// TODO: Replace this logic with SIMD conversion (similar to the one in the decoder)!
     /// </summary>
-    internal unsafe struct RgbToYCbCrTables
+    internal unsafe struct RgbToYCbCrConverterLut
     {
         /// <summary>
         /// The red luminance table
@@ -63,10 +64,10 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
         /// <summary>
         /// Initializes the YCbCr tables
         /// </summary>
-        /// <returns>The initialized <see cref="RgbToYCbCrTables"/></returns>
-        public static RgbToYCbCrTables Create()
+        /// <returns>The initialized <see cref="RgbToYCbCrConverterLut"/></returns>
+        public static RgbToYCbCrConverterLut Create()
         {
-            RgbToYCbCrTables tables = default;
+            RgbToYCbCrConverterLut tables = default;
 
             for (int i = 0; i <= 255; i++)
             {
@@ -92,11 +93,10 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
         }
 
         /// <summary>
-        /// TODO: Replace this logic with SIMD conversion (similar to the one in the decoder)!
         /// Optimized method to allocates the correct y, cb, and cr values to the DCT blocks from the given r, g, b values.
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void ConvertPixelInto(
+        private void ConvertPixelInto(
             int r,
             int g,
             int b,
@@ -111,8 +111,27 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
             // float cb = 128F + ((-0.168736F * r) - (0.331264F * g) + (0.5F * b));
             cbResult[i] = (this.CbRTable[r] + this.CbGTable[g] + this.CbBTable[b]) >> ScaleBits;
 
-            // float cr = MathF.Round(y + (1.772F * cb), MidpointRounding.AwayFromZero);
+            // float cr = 128F + ((0.5F * r) - (0.418688F * g) - (0.081312F * b))
             crResult[i] = (this.CbBTable[r] + this.CrGTable[g] + this.CrBTable[b]) >> ScaleBits;
+        }
+
+        public void Convert(Span<Rgb24> rgbSpan, ref Block8x8F yBlock, ref Block8x8F cbBlock, ref Block8x8F crBlock)
+        {
+            ref Rgb24 rgbStart = ref rgbSpan[0];
+
+            for (int i = 0; i < 64; i++)
+            {
+                ref Rgb24 c = ref Unsafe.Add(ref rgbStart, i);
+
+                this.ConvertPixelInto(
+                    c.R,
+                    c.G,
+                    c.B,
+                    ref yBlock,
+                    ref cbBlock,
+                    ref crBlock,
+                    i);
+            }
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
--- a/src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
@@ -15,97 +15,43 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
 {
     internal static class RgbToYCbCrConverterVectorized
     {
-        private static ReadOnlySpan<byte> ExtractionMasks => new byte[]
-        {
-            0x0, 0xFF, 0xFF, 0xFF, 0x1, 0xFF, 0xFF, 0xFF, 0x2, 0xFF, 0xFF, 0xFF, 0x3, 0xFF, 0xFF, 0xFF,   0x10, 0xFF, 0xFF, 0xFF, 0x11, 0xFF, 0xFF, 0xFF, 0x12, 0xFF, 0xFF, 0xFF, 0x13, 0xFF, 0xFF, 0xFF,
-            0x4, 0xFF, 0xFF, 0xFF, 0x5, 0xFF, 0xFF, 0xFF, 0x6, 0xFF, 0xFF, 0xFF, 0x7, 0xFF, 0xFF, 0xFF,   0x14, 0xFF, 0xFF, 0xFF, 0x15, 0xFF, 0xFF, 0xFF, 0x16, 0xFF, 0xFF, 0xFF, 0x17, 0xFF, 0xFF, 0xFF,
-            0x8, 0xFF, 0xFF, 0xFF, 0x9, 0xFF, 0xFF, 0xFF, 0xA, 0xFF, 0xFF, 0xFF, 0xB, 0xFF, 0xFF, 0xFF,   0x18, 0xFF, 0xFF, 0xFF, 0x19, 0xFF, 0xFF, 0xFF, 0x1A, 0xFF, 0xFF, 0xFF, 0x1B, 0xFF, 0xFF, 0xFF,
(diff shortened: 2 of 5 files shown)


tkp1n:
Looks like some of my final touches broke the tests.. I'll look into it and mark this as a draft in the meantime.

src/ImageSharp/Formats/Jpeg/Components/Encoder/YCbCrForwardConverter{TPixel}.cs
```diff
@@ -48,7 +47,12 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
         public static YCbCrForwardConverter<TPixel> Create()
         {
             var result = default(YCbCrForwardConverter<TPixel>);
-            result.colorTables = RgbToYCbCrTables.Create();
+            if (RgbToYCbCrConverterVectorized.IsSupported)
```
brianpopow: shouldn't this be `if (!RgbToYCbCrConverterVectorized.IsSupported)` instead?

tkp1n: Just fixed this a second ago, well spotted 👍

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
```diff
@@ -0,0 +1,182 @@
...
+
+                    // 2nd part of unrolled loop
+                    mask = Unsafe.Add(ref Unsafe.As<byte, Vector256<byte>>(ref MemoryMarshal.GetReference(ExtractionMasks)), (j * 2) + 1);
+
+                    r = Avx.ConvertToVector256Single(Avx2.Shuffle(red, mask).AsInt32());
+                    g = Avx.ConvertToVector256Single(Avx2.Shuffle(green, mask).AsInt32());
+                    b = Avx.ConvertToVector256Single(Avx2.Shuffle(blue, mask).AsInt32());
+
+                    // (0.299F * r) + (0.587F * g) + (0.114F * b);
+                    Vector256<float> yy1 = Fma.MultiplyAdd(f0299, r, Fma.MultiplyAdd(f0587, g, Avx.Multiply(f0114, b)));
```
JimBobSquarePants, line 157: We have a `SimdUtils.HWIntrinsics.MultiplyAdd` method that can be used to abstract this away. It falls back to `Avx.Add/Multiply`

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
```diff
@@ -0,0 +1,182 @@
...
+#endif
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
+{
+    internal static class RgbToYCbCrConverterVectorized
+    {
+        private static ReadOnlySpan<byte> ExtractionMasks => new byte[]
+        {
+            0x0, 0xFF, 0xFF, 0xFF, 0x1, 0xFF, 0xFF, 0xFF, 0x2, 0xFF, 0xFF, 0xFF, 0x3, 0xFF, 0xFF, 0xFF,   0x10, 0xFF, 0xFF, 0xFF, 0x11, 0xFF, 0xFF, 0xFF, 0x12, 0xFF, 0xFF, 0xFF, 0x13, 0xFF, 0xFF, 0xFF,
```
JimBobSquarePants, line 20: I prefer to keep the masks in decimal rather than hex notation personally as I find it far more readable since we're looking at indices.

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
```diff
@@ -0,0 +1,182 @@
...
+        ///      r32, ... r63, g32, ..., g63, b31, ..., b63 }
+        /// </summary>
+        /// <remarks>
+        /// SSE is used for this operation as it is significantly faster than AVX in this specific case.
+        /// Solving this problem with AVX requires too many instructions that cross the 128-bit lanes of YMM registers.
+        /// </remarks>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        private static void SeparateRgb(ReadOnlySpan<Rgb24> rgbSpan)
+        {
+            var selectRed0 = Vector128.Create(0x00, 0x03, 0x06, 0x09, 0x0C, 0x0F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF);
```
JimBobSquarePants, line 62: Correct me if I'm wrong but these would be better off using ROS pre NET 5 wouldn't they?

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
```diff
@@ -0,0 +1,182 @@
...
+            var selectGreen1 = Vector128.Create(0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x03, 0x06, 0x09, 0x0C, 0x0F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF);
+            var selectGreen2 = Vector128.Create(0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x02, 0x05, 0x08, 0x0B, 0x0E);
+
+            var selectBlue0 = Vector128.Create(0x02, 0x05, 0x08, 0x0B, 0x0E, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF);
+            var selectBlue1 = Vector128.Create(0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x01, 0x04, 0x07, 0x0A, 0x0D, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF);
+            var selectBlue2 = Vector128.Create(0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x03, 0x06, 0x09, 0x0C, 0x0F);
+
+            for (int i = 0; i < 2; i++)
+            {
+                ref Vector128<byte> inRef = ref Unsafe.Add(ref Unsafe.As<Rgb24, Vector128<byte>>(ref MemoryMarshal.GetReference(rgbSpan)), i * 6);
```
JimBobSquarePants, line 76: I feel like this would be better outside the loop.

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterLut.cs
```diff
@@ -63,10 +64,10 @@ internal unsafe struct RgbToYCbCrTables
         /// <summary>
         /// Initializes the YCbCr tables
         /// </summary>
-        /// <returns>The initialized <see cref="RgbToYCbCrTables"/></returns>
-        public static RgbToYCbCrTables Create()
+        /// <returns>The initialized <see cref="RgbToYCbCrConverterLut"/></returns>
+        public static RgbToYCbCrConverterLut Create()
```
JimBobSquarePants, line 68: Is this masked by a condition now? I can't spot anything in the code. It would be good to avoid it if possible.

JimBobSquarePants (commented):
@tkp1n Thanks for having a crack at this! Very much appreciated! The deinterleaving code had me beat the other week when I had a look so I was hoping someone smarter than me would have a go. 

I'm gonna leave this to the brains trust to properly review because I'm still very much a SIMD noob but I've added  a few comments.

@antonfirsov @saucecontrol If you have ideas and time please comment. I would have though we could have sped up the operation a bit more than the current (welcome) improvement.

saucecontrol:
Rather than having two methods with a round-trip through memory, I'd try to get this all done in one step.  The best option would actually be to do the conversion fixed-point if AVX2 is available, but I assume that's a much larger change if the encoder is already taking the YCbCr input in floating point.

This version could be simplified using something like this:

```
// read 32 bytes, of which 24 will be used
rgb = vmovdqu(rgbptr)
RGBRGBRGBRGBRGBR|GBRGBRGBxxxxxxxx

// move 12 good bytes into each lane
rgb = vpermd(rgb, mask 0,1,2,6,3,4,5,7)
RGBRGBRGBRGBxxxx|RGBRGBRGBRGBxxxx

// group channel values together
rgb = vpshufb(rgb, mask 0,3,6,9,1,4,7,10,2,5,8,11,-1,-1,-1,-1)
RRRRGGGGBBBBxxxx|RRRRGGGGBBBBxxxx

// unpack to int32 width
rg = vpunpcklbw(rgb, zero)
R0R0R0R0G0G0G0G0|R0R0R0R0G0G0G0G0
bx = vpunpckhbw(rgb, zero)
B0B0B0B0x0x0x0x0|B0B0B0B0x0x0x0x0

r = vpunpcklwd(rg, zero)
R000R000R000R000|R000R000R000R000
g = vpunpckhwd(rg, zero)
G000G000G000G000|G000G000G000G000
b = vpunpcklwd(bx, zero)
B000B000B000B000|B000B000B000B000

// convert to float, then use the existing ConvertInternal math
// values will already be properly ordered for 32-byte stores
```

That's untested but should get you going in the right direction.  You'll need to allow for 8 bytes of overrun on the last read, or you can stop short and process the last 24 bytes using the LUT fallback or with the SIMD loop using a read/shift from the end.

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
saucecontrol: Yep, the codegen for these is bad on 3.1, but hopefully they're not needed.

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
saucecontrol: All processors that support AVX2 also support FMA, so it's fine to require both explicitly.  As long as you protect against use with the JIT flags disabling FMA by including the IsSupported check, it won't make any difference to real users.  Just comes down to a matter of preference.

antonfirsov:
I'm super curious how much the algorithm in https://github.com/SixLabors/ImageSharp/pull/1508#issuecomment-761924703 could give us, but if there's no time to implement it, we can merge the PR as is after changing the ` Vector128.Create`-s.

In any case let's not close #1476 with the PR, there is definitely space left for improvements by rearranging the pipeline steps within the `Encode420` / `Encode444` loops. (Needs further analysis to determine what exactly has to be done.)

tkp1n:
@antonfirsov I'm already on it.. Seems to reduce times roughly from 170ns to 100ns 🚀 (initial numbers)
@saucecontrol Thanks for the input!

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
tkp1n: Switched to `SimdUtils.HWIntrinsics.MultiplyAdd`

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
tkp1n: Done

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
tkp1n: The remaining masks were moved to ROS

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterLut.cs
tkp1n: `Create` is only called in `YCbCrForwardConverter<TPixel>` if the Vectorized variant isn't available. One leftover would be that the space for the tables is still allocated on the stack even if it is never used.

JimBobSquarePants:
I’ll do a final pass tonight but this is looking great! 👍

JimBobSquarePants (approved):
LGTM!

PR #1513, opened 2021-01-19
Speed improvements to resize kernel (w/ SIMD)

Sergio0694:
### Prerequisites

### Description
Related to #1476, this PR includes some speed improvements to the resize kernels.
In particular, it introduces a new vectorized path using AVX2/FMA operations to perform the convolutions.
For those interested, [here is](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgDkBXfGKASzFwG4BYAKDMq0ASgwB2GVsxoBhCPgAOrADYsAyiwBu7GFz4DqNEeMkwaASXFtRudrv4UDRiVIsYrNjjQAaADiQ9ePlZxFlFsJXJYbAATCFElAE9yMVxsADMYclw3BjAMciEdVgAvGABpUJglPgBvPnIG8nk2DWwMTKjY+KS0pQg2gCpyYAY0jKgABTcAxvJ6xuIAZnJg/IAZGFEAcwwAC3mGut5Z2YBtAFkYPYhoswUlAApL69v7gHl5CTjcGgBBLa2sFwNg0MAsSmCwS2AEoALoHE5bK4zRoAXwRCKW5AAajA8tAUORZKINBAlKDZLAHrA0ji8RgCZEIAB3VQYbBQDCFNLQhFHE6NXr9DBDEZjNTsznkAC85D2rB+YvGUygKIFQsGw1G4wAoqJojKteKoGyOfkANTkB7yn4bbZ7cgAMnIAD9FtC1SdcfioAAmACsSAAPBqMAA+SI6BhKDABpAAfQosu9DL9gZDfTaYZoAC0WBBPbMU9A4xnhRGgdHY4H41RDcW08HQ9m81ACwjZq0oOR8HgANb1+klwMyKLtB6kNDkSfTqczqhThfkJdUD18DuNZm7ZSZB5KiVm8hBo26/W844C8j8y+zSsxuOJw0AMV7NHOVdY8kSv2i0QeG5vBoAFVrHSUxflwIMG3QOkfVLZsw2pGBaTbVlJS5ZDoTQADAN+DQEF9GgJhYfAGHabEOR8BBFl9B4GzjUcYDaGBVDAcIORA1IMgeAYHliBhgBUAZoX3E10OhP5cFUKEVAeLCe37LCcJvO9qwTUg10CC9AMjXAqwfOtZRfbA3w/L8Eh/P9lMvTiwMkqChygGD6PTBCkNpWyMj+X93KZNCzW5KdfWhJTtJ0xo8IIoiSLImAKKgKiaLoxyGOkMcWLYpQONA7jeP4wSYGEvdtQPKVLWCiSIOk7ZZPk3tcD7ULwpOVSDM06yGlE00ytlFBC0vVDuow2lZRpchPPAnyxsG9DAvIFBNJvdEtJUqN7xrJNyEi7y/1ajapz2hNVwCHDWFpB5rW3W1Nh2XYnXId1yDDWVgpw68dMOx8jNfd8Y0/b8fI6k4Jvs6CpxcptM3DRDppZIbuSa5qIvwwjiKgUjyMo6jaIhxjmNY9ioAmni+IgAShJEkqxLNSqpJkmA5KnerGuwsLms+jSTrZy8uvQ8hzRe/qBRmgLkMNMaQcs3yRc5ObgqF8hlpwhsqF9HwyyzXSq1VnxDVUXBJt2ta1MTGgAHErjWFkWEZrX1vU82riA+R5BtkKuZvM6rUuhUaDtW77tXcgAEJZQ0t6gdUnXnx+syAas7mdMlyCwdg1MdY16Hpbh2bMNZpHGhVtW8fHAZeZp/OC7tjAdcWy8lcTygAHZyF46DhMdKO1c9ZblqAA=) a sharplab link with the current codegen for the `ConvolveCore` method, when using SIMD operations.

Still running benchmarks and work in progress...

### DIFF
--- a/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
+++ b/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
@@ -71,7 +71,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
         public Vector4 ConvolveCore(ref Vector4 rowStartRef)
         {
 #if SUPPORTS_RUNTIME_INTRINSICS
-            if (Fma.IsSupported)
+            if (Avx2.IsSupported)
             {
                 float* bufferStart = this.bufferPtr;
                 float* bufferEnd = bufferStart + (this.Length & ~1);
@@ -82,8 +82,9 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
                 {
                     Vector256<float> rowItem256 = Unsafe.As<Vector4, Vector256<float>>(ref rowStartRef);
                     Vector256<float> bufferItem256 = Avx2.PermuteVar8x32(Vector256.Create(*(double*)bufferStart).AsSingle(), mask);
+                    Vector256<float> multiply256 = Avx.Multiply(rowItem256, bufferItem256);
 
-                    result256 = Fma.MultiplyAdd(rowItem256, bufferItem256, result256);
+                    result256 = Avx.Add(multiply256, result256);
 
                     bufferStart += 2;
                     rowStartRef = ref Unsafe.Add(ref rowStartRef, 2);
@@ -95,8 +96,9 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
                 {
                     Vector128<float> rowItem128 = Unsafe.As<Vector4, Vector128<float>>(ref rowStartRef);
                     var bufferItem128 = Vector128.Create(*bufferStart);
+                    Vector128<float> multiply128 = Sse.Multiply(rowItem128, bufferItem128);
 
-                    result128 = Fma.MultiplyAdd(rowItem128, bufferItem128, result128);
+                    result128 = Sse.Add(multiply128, result128);
                 }
 
                 return *(Vector4*)&result128;
@@ -114,8 +116,8 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
                     // Vector4 v = offsetedRowSpan[i];
                     result += rowStartRef * *bufferStart;
 
-                    rowStartRef = ref Unsafe.Add(ref rowStartRef, 1);
                     bufferStart++;
+                    rowStartRef = ref Unsafe.Add(ref rowStartRef, 1);
                 }
 
                 return result;
--- a/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
+++ b/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
@@ -5,6 +5,7 @@ using System;
 using System.Numerics;
 using System.Runtime.CompilerServices;
 #if SUPPORTS_RUNTIME_INTRINSICS
+using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics;
 using System.Runtime.Intrinsics.X86;
 #endif
@@ -77,7 +78,14 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
                 float* bufferEnd = bufferStart + (this.Length & ~3);
                 Vector256<float> result256_0 = Vector256<float>.Zero;
                 Vector256<float> result256_1 = Vector256<float>.Zero;
-                var mask = Vector256.Create(0, 0, 0, 0, 1, 1, 1, 1);
+                ReadOnlySpan<byte> maskBytes = new byte[]
+                {
+                    0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0,
+                    1, 0, 0, 0, 1, 0, 0, 0,
+                    1, 0, 0, 0, 1, 0, 0, 0,
+                };
+                Vector256<int> mask = Unsafe.ReadUnaligned<Vector256<int>>(ref MemoryMarshal.GetReference(maskBytes));
 
                 while (bufferStart < bufferEnd)
                 {
--- a/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
+++ b/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
@@ -4,6 +4,10 @@
 using System;
 using System.Numerics;
 using System.Runtime.CompilerServices;
+#if SUPPORTS_RUNTIME_INTRINSICS
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.X86;
+#endif
 
 namespace SixLabors.ImageSharp.Processing.Processors.Transforms
 {
@@ -66,21 +70,55 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
         [MethodImpl(InliningOptions.ShortMethod)]
         public Vector4 ConvolveCore(ref Vector4 rowStartRef)
         {
-            ref float horizontalValues = ref Unsafe.AsRef<float>(this.bufferPtr);
-
-            // Destination color components
-            Vector4 result = Vector4.Zero;
-
-            for (int i = 0; i < this.Length; i++)
+#if SUPPORTS_RUNTIME_INTRINSICS
+            if (Fma.IsSupported)
             {
-                float weight = Unsafe.Add(ref horizontalValues, i);
+                float* bufferStart = this.bufferPtr;
+                float* bufferEnd = bufferStart + (this.Length & ~1);
+                Vector256<float> result256 = Vector256<float>.Zero;
 
-                // Vector4 v = offsetedRowSpan[i];
-                Vector4 v = Unsafe.Add(ref rowStartRef, i);
-                result += v * weight;
+                while (bufferStart < bufferEnd)
+                {
+                    Vector256<float> rowItem256 = Unsafe.As<Vector4, Vector256<float>>(ref rowStartRef);
+                    var bufferItem256 = Vector256.Create(Vector128.Create(bufferStart[0]), Vector128.Create(bufferStart[1]));
+
+                    result256 = Fma.MultiplyAdd(rowItem256, bufferItem256, result256);
+
+                    bufferStart += 2;
+                    rowStartRef = ref Unsafe.Add(ref rowStartRef, 2);
+                }
+
+                Vector128<float> result128 = Sse.Add(result256.GetLower(), result256.GetUpper());
+
+                if ((this.Length & 1) != 0)
+                {
+                    Vector128<float> rowItem128 = Unsafe.As<Vector4, Vector128<float>>(ref rowStartRef);
+                    var bufferItem128 = Vector128.Create(*bufferStart);
+
+                    result128 = Fma.MultiplyAdd(rowItem128, bufferItem128, result128);
+                }
+
+                return *(Vector4*)&result128;
             }
+            else
+#endif
+            {
+                // Destination color components
+                Vector4 result = Vector4.Zero;
+                float* bufferStart = this.bufferPtr;
+                float* bufferEnd = this.bufferPtr + this.Length;
 
-            return result;
+                while (bufferStart < bufferEnd)
+                {
+                    // Vector4 v = offsetedRowSpan[i];
+                    result += rowStartRef * *bufferStart;
+
+                    rowStartRef = ref Unsafe.Add(ref rowStartRef, 1);
+                    bufferStart++;
+                }
+
+                return result;
+            }
         }
 
         /// <summary>
--- a/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
+++ b/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
@@ -71,7 +71,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
         public Vector4 ConvolveCore(ref Vector4 rowStartRef)
         {
 #if SUPPORTS_RUNTIME_INTRINSICS
-            if (Avx2.IsSupported)
+            if (Fma.IsSupported)
             {
                 float* bufferStart = this.bufferPtr;
                 float* bufferEnd = bufferStart + (this.Length & ~1);
@@ -80,11 +80,20 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
 
                 while (bufferStart < bufferEnd)
                 {
-                    Vector256<float> rowItem256 = Unsafe.As<Vector4, Vector256<float>>(ref rowStartRef);
-                    Vector256<float> bufferItem256 = Avx2.PermuteVar8x32(Vector256.Create(*(double*)bufferStart).AsSingle(), mask);
-                    Vector256<float> multiply256 = Avx.Multiply(rowItem256, bufferItem256);
-
-                    result256 = Avx.Add(multiply256, result256);
+                    // It is important to use a single expression here so that the JIT will correctly use vfmadd231ps
+                    // for the FMA operation, and execute it directly on the target register and reading directly from
+                    // memory for the first parameter. This skips initializing a SIMD register, and an extra copy.
+                    // The code below should compile in the following assembly on .NET 5 x64:
+                    //
+                    // vmovsd xmm2, [rax]               ; load *(double*)bufferStart into xmm2 as [ab, _]
+                    // vpermps ymm2, ymm1, ymm2         ; permute as a float YMM register to [a, a, a, a, b, b, b, b]
+                    // vfmadd231ps ymm0, ymm2, [r8]     ; result256 = FMA(pixels, factors) + result256
+                    //
+                    // For tracking the codegen issue with FMA, see: https://github.com/dotnet/runtime/issues/12212.
+                    result256 = Fma.MultiplyAdd(
+                        Unsafe.As<Vector4, Vector256<float>>(ref rowStartRef),
+                        Avx2.PermuteVar8x32(Vector256.CreateScalarUnsafe(*(double*)bufferStart).AsSingle(), mask),
+                        result256);
 
                     bufferStart += 2;
                     rowStartRef = ref Unsafe.Add(ref rowStartRef, 2);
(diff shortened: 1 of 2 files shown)


saucecontrol:
Looking good on Skylake 👍 

``` ini
.NET Core SDK=5.0.102
  [Host]     : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT
  Job-XNQYOQ : .NET Framework 4.8 (4.8.4300.0), X64 RyuJIT
  Job-ALEINB : .NET Core 2.1.24 (CoreCLR 4.6.29518.01, CoreFX 4.6.29518.01), X64 RyuJIT
  Job-LVTEAU : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT
```

Before:

|                                   Method |        Job |       Runtime |      Mean |     Error |    StdDev | Ratio | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------- |----------- |-------------- |----------:|----------:|----------:|------:|------:|------:|------:|----------:|
|                            SystemDrawing | Job-ILLXJH |    .NET 4.7.2 | 17.320 ms | 0.0918 ms | 0.0814 ms |  1.00 |     - |     - |     - |     256 B |
| &#39;ImageSharp, MaxDegreeOfParallelism = 1&#39; | Job-ILLXJH |    .NET 4.7.2 |  9.572 ms | 0.0701 ms | 0.0656 ms |  0.55 |     - |     - |     - |   40624 B |
|                                          |            |               |           |           |           |       |       |       |       |           |
|                            SystemDrawing | Job-MQEJZE | .NET Core 2.1 | 17.535 ms | 0.2673 ms | 0.2500 ms |  1.00 |     - |     - |     - |      96 B |
| &#39;ImageSharp, MaxDegreeOfParallelism = 1&#39; | Job-MQEJZE | .NET Core 2.1 |  8.645 ms | 0.0660 ms | 0.0618 ms |  0.49 |     - |     - |     - |   40472 B |
|                                          |            |               |           |           |           |       |       |       |       |           |
|                            SystemDrawing | Job-BXMGKS | .NET Core 3.1 | 17.498 ms | 0.1888 ms | 0.1766 ms |  1.00 |     - |     - |     - |     140 B |
| &#39;ImageSharp, MaxDegreeOfParallelism = 1&#39; | Job-BXMGKS | .NET Core 3.1 |  8.008 ms | 0.0335 ms | 0.0314 ms |  0.46 |     - |     - |     - |   40472 B |

After:

|                                   Method |        Job |       Runtime |      Mean |     Error |    StdDev | Ratio |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------- |----------- |-------------- |----------:|----------:|----------:|------:|-------:|------:|------:|----------:|
|                            SystemDrawing | Job-KIBPBM |    .NET 4.7.2 | 17.318 ms | 0.1773 ms | 0.1572 ms |  1.00 |      - |     - |     - |     256 B |
| &#39;ImageSharp, MaxDegreeOfParallelism = 1&#39; | Job-KIBPBM |    .NET 4.7.2 |  9.562 ms | 0.0685 ms | 0.0607 ms |  0.55 |      - |     - |     - |   40624 B |
|                                          |            |               |           |           |           |       |        |       |       |           |
|                            SystemDrawing | Job-AOLXLZ | .NET Core 2.1 | 17.362 ms | 0.2023 ms | 0.1892 ms |  1.00 |      - |     - |     - |      96 B |
| &#39;ImageSharp, MaxDegreeOfParallelism = 1&#39; | Job-AOLXLZ | .NET Core 2.1 |  8.621 ms | 0.0707 ms | 0.0662 ms |  0.50 |      - |     - |     - |   40472 B |
|                                          |            |               |           |           |           |       |        |       |       |           |
|                            SystemDrawing | Job-INXCIO | .NET Core 3.1 | 17.418 ms | 0.1830 ms | 0.1712 ms |  1.00 |      - |     - |     - |     140 B |
| &#39;ImageSharp, MaxDegreeOfParallelism = 1&#39; | Job-INXCIO | .NET Core 3.1 |  6.545 ms | 0.0714 ms | 0.0668 ms |  0.38 | 7.8125 |     - |     - |   40424 B |

Sergio0694:
@saucecontrol That's awesome, thank you for running the benchmarks on your machine! 🚀

Here are mine, it looks like this PR is actually ever so slightly slower than master.
I'm thinking it might just be that my Ryzen 2700X (Zen+ arch) is trash at FMA stuff, wouldn't surprise me 🤔

## Master

|                                   Method |      Runtime |      Mean |     Error |    StdDev | Ratio |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------- |------------- |----------:|----------:|----------:|------:|--------:|------:|------:|----------:|
| ImageSharp | .NET Core 3.1 |  7.506 ms | 0.0110 ms | 0.0098 ms |  0.44 |  7.8125 |     - |     - |   40772 B |

## PR

|                                   Method |      Runtime |      Mean |     Error |    StdDev | Ratio |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------- |------------- |----------:|----------:|----------:|------:|--------:|------:|------:|----------:|
| ImageSharp | .NET Core 3.1 |  7.646 ms | 0.0986 ms | 0.1174 ms |  0.45 |  7.8125 |     - |     - |   40712 B |

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
```diff
@@ -66,21 +70,87 @@ public Vector4 Convolve(Span<Vector4> rowSpan)
...
+                    // vmovsd xmm2, [rax]               ; load *(double*)bufferStart into xmm2 as [ab, _]
+                    // vpermps ymm2, ymm1, ymm2         ; permute as a float YMM register to [a, a, a, a, b, b, b, b]
+                    // vfmadd231ps ymm0, ymm2, [r8]     ; result256_0 = FMA(pixels, factors) + result256_0
+                    //
+                    // For tracking the codegen issue with FMA, see: https://github.com/dotnet/runtime/issues/12212.
+                    // Additionally, we're also unrolling two computations per each loop iterations to leverage the
+                    // fact that most CPUs have two ports to schedule multiply operations for FMA instructions.
+                    result256_0 = Fma.MultiplyAdd(
+                        Unsafe.As<Vector4, Vector256<float>>(ref rowStartRef),
+                        Avx2.PermuteVar8x32(Vector256.CreateScalarUnsafe(*(double*)bufferStart).AsSingle(), mask),
```
antonfirsov, line 98: According to what I [learned](https://github.com/SixLabors/ImageSharp/pull/1242#issuecomment-735346693) from @saucecontrol, moving permutes out from an operation (dependency) chain and running them in a separate sequence might help performance.

Thinking it further:
Would make the code more tricky, but maybe we can try to process two `Vector256<float>`-s in the loop body, so we can run 4 permutes in a row, then do 4+4 FMA-s.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
```diff
@@ -66,21 +70,87 @@ public Vector4 Convolve(Span<Vector4> rowSpan)
...
         {
-            ref float horizontalValues = ref Unsafe.AsRef<float>(this.bufferPtr);
+#if SUPPORTS_RUNTIME_INTRINSICS
+            if (Fma.IsSupported)
+            {
+                float* bufferStart = this.bufferPtr;
+                float* bufferEnd = bufferStart + (this.Length & ~3);
+                Vector256<float> result256_0 = Vector256<float>.Zero;
+                Vector256<float> result256_1 = Vector256<float>.Zero;
+                var mask = Vector256.Create(0, 0, 0, 0, 1, 1, 1, 1);
```
antonfirsov, line 80: Should be `Vector.Load` with a ROS.

antonfirsov:
According to the test failures, there is some noticeable (but visually still insignificant) difference between the `Vector4` and the FMA output. @Sergio0694 I wonder if we get a better understanding on this before changing tolerances?

antonfirsov:
Regarding my https://github.com/SixLabors/ImageSharp/pull/1513#issuecomment-763597089, I think Wikipedia answered my concerns:

>  a fused multiply–add would compute the entire expression a+b×c to its full precision before rounding the final result down to N significant bits.

If I get it right, the FMA code is actually more accurate.

Sergio0694:
Yeah, in theory the FMA path should be more accurate. From what @tannergooding said on Discord:

> "So, for reference, an fma(x, y, z) vs x * y + z should differ by no more than a single bit (ignoring the special case around infinity)
this can compound over many operations to produce even more significant differences
this falls out because the former rounds once while the latter rounds twice"

So in theory the difference should mean the FMA is ever so slightly better, as you mentioned.
It's not 100% that in theory since if the number of pixels per row is not evenly divisible, the last pixel will be computed with a slightly different precision compared to the previous, and also there's the fact that we're using 4 different accumulators in parallel and then just summing them all together at the end, so I'd expect that would alter the overall precision a bit too. But all things considered this new version should be at least equally as precise as the original, I think 🤔
Also if we can't spot any differences when looking at the images, I'd say we're probably good 😄

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
Sergio0694: The codegen in this very specific case seems to actually be just a `vxorps` (which is actually... weird), so I'm looking into this. It seems to be ok even though we're not using a `ROS` though, or actually better than that too.
Will update in a bit 🙂

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
Sergio0694: There was an issue with using locals here (I documented that in the comments here), where the JIT was picking the wrong instruction for the FMA operation and adding extra unnecessary memory copies. Doing this inline instead picked the right one that directly loaded the first argument from memory, which resulted in much better asseembly. I'm worried that moving things around will make the codegen worse again there. Also from what we discussed on Discord, there's usually 2 ports to perform FMA multiplications, so it might not be beneficial to do more than 2 in the same loop? I mean, other than the general marginal improvements just due to more unrolling, possibly.
I think @saucecontrol is doing only two ops per iteration as well in his own lib because of this? 🤔

saucecontrol:
> using 4 different accumulators in parallel and then just summing them all together at the end, so I'd expect that would alter the overall precision a bit too

Yes, this also improves precision.  Take the following example:

```c#
var rand = new Random(42);
var vals = new float[128];
for (int i = 0; i < vals.Length; i++)
	vals[i] = (float)rand.NextDouble();

float fa = 0;
for (int i = 0; i < vals.Length; i++)
	fa += vals[i];

double da = 0;
for (int i = 0; i < vals.Length; i++)
	da += vals[i];

float a0 = 0, a1 = 0, a2 = 0, a3 = 0;
for (int i = 0; i < vals.Length / 4; i++)
{
	int j = i * 4;
	a0 += vals[j];
	a1 += vals[j + 1];
	a2 += vals[j + 2];
	a3 += vals[j + 3];
}
float pa = a0 + a1 + a2 + a3;

Console.WriteLine("float accumulator:    " + fa);
Console.WriteLine("double accumulator:   " + da);
Console.WriteLine("4x float accumulator: " + pa);
```

Outputs:

```
float accumulator:    54.136913
double accumulator:   54.136910965666175
4x float accumulator: 54.13691
```

saucecontrol:
> I'm thinking it might just be that my Ryzen 2700X (Zen+ arch) is trash at FMA stuff

I looked at the Zen+ perf numbers at [uops.info](https://uops.info/table.html?search=vfmadd231ps&cb_lat=on&cb_tp=on&cb_uops=on&cb_ports=on&cb_SKL=on&cb_ICL=on&cb_ZEN%2B=on&cb_ZEN2=on&cb_measurements=on&cb_doc=on&cb_avx=on&cb_avx2=on&cb_fma=on), and they're not great.  On Intel processors, FMA gives you the add for free after the multiply, saving 4 cycles latency.  On AMD processors, FMA is only 1 cycle faster than MUL+ADD, but it runs on the multiply ports the whole time, whereas ADD can use a different set of ports if they are split.  This was the point @tannergooding was making on discord yesterday.  If there were contention on the MUL ports, the FMA version ties them up longer so may not be a win.  That's particularly true of Zen+, where a 256-bit FMA ties up both ports because it's split to 2 128-bit uops.  In this specific case, there is no other work to be done, so it's not a negative -- but it's also not much of a win at all.

The permute, however, is particularly bad on Zen+.  From the [perf numbers](https://uops.info/table.html?search=vpermps&cb_lat=on&cb_tp=on&cb_uops=on&cb_ports=on&cb_SKL=on&cb_ICL=on&cb_ZEN%2B=on&cb_ZEN2=on&cb_measurements=on&cb_doc=on&cb_avx=on&cb_avx2=on&cb_fma=on), it appears AMD processors emulate this instruction with microcode, and the Zen+ version is extra slow.  Since the throughput is so low there, you can't really take advantage of the parallel FMAs.  Looks like you're lucky not to be too much slower than the Vector4 code on that machine.

So, on Intel, this is strictly a win.  On Zen2/3, it should be a win but less so (I don't have access to one at the moment).  And Zen+ would appear to be a loss but one worth taking for the gains everywhere else.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
saucecontrol: Yeah, 2 is the max number of FMAs that can be scheduled at once, but it's a pipelined instruction, so you can get more benefit from scheduling more sequentially.  I had an unroll by 4 in MagicScaler previously, but it wasn't a ton faster so I dropped it to reduce complexity.

The way I get around having to shuffle/permute the weights in the inner loop is by pre-duplicating them in my kernel map.  So my inner loop is just 2 reads of pixel values and 2 FMAs (with the weight reads contained in the FMA instruction).  That approach also has the benefit of being backward compatible with `Vector<T>`, allowing AVX processing on netfx.

Sergio0694:
Thanks for all the extra info @saucecontrol and also the perf analysis, that's super interesting! 😄
Will bump the image comparison threshold to `0.004%` for those resize tests then to make the CI happy.

JimBobSquarePants (approved):

JimBobSquarePants:
LGTM. Great work (and code comments) @Sergio0694 and thanks @saucecontrol for the additional input.

JimBobSquarePants:
For anyone curious here's what the benchmarks against other libraries looks like currently on my SB2. 

```
BenchmarkDotNet=v0.12.1.1467-nightly, OS=Windows 10.0.19042
.NET SDK=5.0.102
  [Host]       : .NET 5.0.2 (5.0.220.61120), X64 RyuJIT
  .Net 5.0 CLI : .NET 5.0.2 (5.0.220.61120), X64 RyuJIT

Job=.Net 5.0 CLI  Arguments=/p:DebugType=portable  Toolchain=.NET 5.0
IterationCount=5  LaunchCount=1  WarmupCount=5
```

|                    Method |        Mean |       Error |      StdDev | Ratio | RatioSD |    Gen 0 |    Gen 1 |    Gen 2 | Allocated |
|-------------------------- |------------:|------------:|------------:|------:|--------:|---------:|---------:|---------:|----------:|
|   'System.Drawing Resize' | 11,789.6 us |   530.93 us |   137.88 us |  1.00 |    0.00 |        - |        - |        - |     136 B |
|       'ImageSharp Resize' |  2,793.8 us |    87.18 us |    22.64 us |  0.24 |    0.00 |        - |        - |        - |   9,152 B |
|      'ImageMagick Resize' | 57,661.5 us | 6,566.96 us | 1,705.42 us |  4.89 |    0.14 |        - |        - |        - |   5,449 B |
|        'FreeImage Resize' |  8,188.4 us |   133.72 us |    34.73 us |  0.69 |    0.01 | 500.0000 | 500.0000 | 500.0000 |     136 B |
|      'MagicScaler Resize' |    797.7 us |    21.21 us |     3.28 us |  0.07 |    0.00 |        - |        - |        - |   1,872 B |
| 'SkiaSharp Canvas Resize' |  2,508.5 us |   331.56 us |    86.11 us |  0.21 |    0.01 |        - |        - |        - |   1,584 B |
| 'SkiaSharp Bitmap Resize' |  2,452.5 us |   377.57 us |    58.43 us |  0.21 |    0.01 |        - |        - |        - |     488 B |
|          'NetVips Resize' |  6,148.8 us |   195.83 us |    30.31 us |  0.52 |    0.01 |        - |        - |        - |   4,104 B |




```
BBenchmarkDotNet=v0.12.1.1467-nightly, OS=Windows 10.0.19042
.NET SDK=5.0.102
  [Host]       : .NET 5.0.2 (5.0.220.61120), X64 RyuJIT
  .Net 5.0 CLI : .NET 5.0.2 (5.0.220.61120), X64 RyuJIT

Job=.Net 5.0 CLI  Arguments=/p:DebugType=portable  Toolchain=.NET 5.0
IterationCount=5  LaunchCount=1  WarmupCount=5
```

|                                Method |     Mean |     Error |   StdDev | Ratio | RatioSD |    Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------------- |---------:|----------:|---------:|------:|--------:|---------:|------:|------:|----------:|
|   'System.Drawing Load, Resize, Save' | 420.2 ms |  37.87 ms |  9.83 ms |  1.00 |    0.00 |        - |     - |     - |     12 KB |
|       'ImageSharp Load, Resize, Save' | 211.6 ms |   7.87 ms |  2.04 ms |  0.50 |    0.01 | 333.3333 |     - |     - |  2,179 KB |
|      'ImageMagick Load, Resize, Save' | 469.7 ms |  10.39 ms |  2.70 ms |  1.12 |    0.02 |        - |     - |     - |     58 KB |
|      'MagicScaler Load, Resize, Save' | 105.0 ms |   5.45 ms |  1.41 ms |  0.25 |    0.00 |        - |     - |     - |     57 KB |
| 'SkiaSharp Canvas Load, Resize, Save' | 263.6 ms |   4.55 ms |  0.70 ms |  0.63 |    0.02 |        - |     - |     - |    107 KB |
| 'SkiaSharp Bitmap Load, Resize, Save' | 278.6 ms | 101.39 ms | 15.69 ms |  0.66 |    0.05 |        - |     - |     - |     91 KB |
|          'NetVips Load, Resize, Save' | 184.5 ms |  17.62 ms |  4.58 ms |  0.44 |    0.01 |        - |     - |     - |     50 KB |

saucecontrol:
Looking good!  I reckon with some more codec work, Vips is within reach.  I've just started working on codecs myself and should be able to help out with that this year.

JimBobSquarePants:
> I've just started working on codecs myself and should be able to help out with that this year.

BD7FC273-69BC-4F7D-A687-98387A8546B8

Sergio0694:
My takeaways from the benchmarks that James shared:

- The overall performance for ImageSharp looks pretty good!
- Definitely agreed that with some improvements to encoding/decoding we can beat NetVips 🚀
- MagicScaler is just stupid fast, it's ridiculous 🤣

Issue #1432, opened 2020-11-16
Add .NET 5 as an explicit target framework.


JimBobSquarePants:
We need to add ARM intrinsics in order to [maintain performance on platforms such as Windows ARM 64](https://github.com/bleroy/core-imaging-playground/pull/21). 

Adding these intrinsics requires .NET 5. 

@tannergooding has already indicated he would be interested in helping out with implementations, I'm keen to learn also.

Adding the target should be trivial enough. Any individual optimization/feature tasks can be raised as and when required.

antonfirsov:
There are several questions that need to be answered before a commiting to this.

1. How many ARM users do we have? What are their use cases? Are the costs worth the benefits? Aren't there other (more important) topics where we can benefit them? (Eg. guaranteeing AOT-friendliness)

2. How can we provide reliable CI coverage for ARM 64?

JimBobSquarePants:
There's no way to answer 1. We have no metrics. For 2. We'd need a Docker image for now as there are no native images available.

brianpopow:
> There's no way to answer 1. We have no metrics. For 2. We'd need a Docker image for now as there are no native images available.

I tried to run the ImageSharp tests with docker / ARM Image in the past without success so far. It seems docker for windows uses QEMU to emulate an ARM cpu when the host CPU is not ARM. This fails with some serious fatal errors when executing the tests.
Something like this:

`qemu: uncaught target signal 6 (Aborted) - core dumped`

Last time i checked is already a year ago, but if nothing essentially changed with docker for windows and QEMU, i dont have high hopes that it will work.
There are new docker ARM images from .net now with the new .Net 5 release, maybe i should give it another try.

kleisauke:
> 2\. How can we provide reliable CI coverage for ARM 64?

Not many CI providers supports ARM64, unfortunately. At the time of writing:
- [Travis supports Linux ARM64](https://docs.travis-ci.com/user/multi-cpu-architectures/) in an early beta stage, only for OSS-projects. Windows ARM64 is unsupported.
- GitHub actions only supports self-hosted Linux ARM64 jobs. The Windows ARM64 runner is [work-in-progress](https://github.com/actions/runner/pull/785).
- Azure Pipelines only supports self-hosted Linux ARM64 jobs. The Windows ARM64 agent is [unsupported in Azure DevOps](https://developercommunity.visualstudio.com/idea/1015752/please-add-a-windows-arm64-hosted-agent.html).
[...]

PR #1519, opened 2021-01-23
Add net5.0 targetframework

stefannikolei:
#1432

### DIFF
--- a/src/ImageSharp/ImageSharp.csproj
+++ b/src/ImageSharp/ImageSharp.csproj
@@ -12,7 +12,7 @@
     <PackageProjectUrl>$(RepositoryUrl)</PackageProjectUrl>
     <PackageTags>Image Resize Crop Gif Jpg Jpeg Bitmap Png Tga NetCore</PackageTags>
     <Description>A new, fully featured, fully managed, cross-platform, 2D graphics API for .NET</Description>
-    <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/ImageSharp/ImageSharp.csproj
+++ b/src/ImageSharp/ImageSharp.csproj
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <Choose>
-    <When Condition="$(CI) == true">
+    <When Condition="$(SIXLABORS_TESTING) == true">
       <PropertyGroup>
         <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
       </PropertyGroup>
--- a/src/ImageSharp/ImageSharp.csproj
+++ b/src/ImageSharp/ImageSharp.csproj
@@ -14,13 +14,18 @@
     <Description>A new, fully featured, fully managed, cross-platform, 2D graphics API for .NET</Description>
   </PropertyGroup>
 
-  <PropertyGroup Condition="$(CI) == false" >
-    <TargetFrameworks>netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="$(CI) == true">
-    <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
-  </PropertyGroup>
+  <Choose>
+    <When Condition="$(CI) == true">
+      <PropertyGroup>
+        <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
+      </PropertyGroup>
+    </When>
+    <Otherwise>
+      <PropertyGroup>
+        <TargetFrameworks>netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


antonfirsov:
We had several discussions about this with @JimBobSquarePants, and if I'm not mistaken the conclusion was that `net5.0` should be a **test-only** target.
The reason is that we don't expect to add 5.0-specific code to the library anytime soon, therefore it's not worth to pay the costs of an additional target for now (larger package, longer local builds, complexity etc.)

Nevertheless, it's very important to make our CI run tests against `net5.0` so we can ensure that there are no regressions in the library on that platform. @stefannikolei any chance you can turn this into a test/CI-only change?

stefannikolei:
@antonfirsov Sure i can make that change. The pr in shared infrastructure needs to be merged too. Because I got only errors when I  added net50 as target

stefannikolei:
Build is not working without the .net50 target framework

```
/usr/share/dotnet/sdk/5.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(241,5): error NETSDK1005: Assets file '/home/runner/work/ImageSharp/ImageSharp/artifacts/obj/src/ImageSharp/project.assets.json' doesn't have a target for 'net5.0'. Ensure that restore has run and that you have included 'net5.0' in the TargetFrameworks for your project. [/home/runner/work/ImageSharp/ImageSharp/src/ImageSharp/ImageSharp.csproj]
```

brianpopow:
> Build is not working without the .net50 target framework
> 

I am able to build it as it is right now (without the .net50 target framework in the `ImageSharp.csproj`). Im using .Net SDK version `5.0.101`. It also works on ubuntu (with docker image `mcr.microsoft.com/dotnet/sdk:5.0.101`).

Not sure why it does not work for @stefannikolei and the CI. @stefannikolei: which .Net SDK version are you using?

stefannikolei:
On my Mac and rider I do not see that problem. Gonna look into the build script what they call there.

stefannikolei:
Ok i now know where the problem is. 
In The build script we explicitly call the build with .net5 this is the call which fails.

brianpopow:
> Ok i now know where the problem is.
> In The build script we explicitly call the build with .net5 this is the call which fails.

The build works now, but the test execution fails on `mac-os` in the CI. I do not have a mac to test this. @stefannikolei can you run `./ci-test.ps1 "macos-latest" "net5.0" "-x64" "false"` and see if it works for you?

stefannikolei:
@brianpopow it is a vicious circle. It can not run because the .net5.0 dll was not build ...

JimBobSquarePants:
Wrap separate `TargetFrameworks` in the csproj files in a condition based on a custom environmental variable, one with NET5.0 one without. Then we can target the framework only when we need to during CI testing.

stefannikolei:
@JimBobSquarePants I added the env variable. But now i am hitting the problem why i opened https://github.com/SixLabors/SharedInfrastructure/pull/15 on SharedInfrastructure...

src/ImageSharp/ImageSharp.csproj
```diff
@@ -12,9 +12,16 @@
     <PackageProjectUrl>$(RepositoryUrl)</PackageProjectUrl>
     <PackageTags>Image Resize Crop Gif Jpg Jpeg Bitmap Png Tga NetCore</PackageTags>
     <Description>A new, fully featured, fully managed, cross-platform, 2D graphics API for .NET</Description>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="$(CI) == false" >
```
JimBobSquarePants, line 17: I don't think comparing to false will work with local builds since there's no matching environmental variable to coerce.

The following will work and is explicit.

```xml
<Choose>
  <When Condition="$(CI) == true">
    <PropertyGroup>
      <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
    </PropertyGroup>
  </When>
  <Otherwise>
    <PropertyGroup>
      <TargetFrameworks>netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
    </PropertyGroup>
  </Otherwise>
</Choose>
```

JimBobSquarePants (changes requested):
See the comment regarding the targetframework options.

src/ImageSharp/ImageSharp.csproj
JimBobSquarePants: Once this change is made run the following command from the repository root to update to the submodule version containing your fix from https://github.com/SixLabors/SharedInfrastructure/pull/15 (Thanks for that fix!)
```
```

stefannikolei: I changed it. Thank you for your patience in this small change. I learned a lot about MSBuild variables and things you can do in CSProj. Kind of Black magic

JimBobSquarePants: Thanks for doing the PR! You and I have both been learning. 😄

JimBobSquarePants (approved):

JimBobSquarePants:
Oooh @stefannikolei I approved this but just realized there's an issue. 

We only want NET5 on the build/test part, not on publish so we should be using a custom environmental property as the filter instead. Use `SIXLABORS_TESTING` both the `Build` and `Test` steps and as the MsBuild filter.

We're explicitly setting the `CI` variable in the `Test` step but we shouldn't be since it's a default for all steps.
https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables

stefannikolei:
> SIXLABORS_TESTING

So just change CI with SIXLABORS_TESTING?

JimBobSquarePants:
>So just change CI with SIXLABORS_TESTING?

Yep but you need to set it in both steps not just `Test`

https://github.com/SixLabors/ImageSharp/blob/954d23399edb7128123671099fb2d7a248dae504/.github/workflows/build-and-test.yml#L67-L76

stefannikolei:
@JimBobSquarePants  Want me to change TestEnvironment.cs to check for SIXLABORS_TESTING rather than CI, or should i add the CI back to the build script

JimBobSquarePants:
The failure is unrelated. It happens every so often when there’s heavy traffic in GitHub Actions. The CI environmental variable is always true here. GitHub set it. Check the link I shared for a list of defaults.

Issue #12, opened 2016-10-30
No Tiff support.


JimBobSquarePants:
_From @JimBobSquarePants on August 31, 2016 1:49_

We should be able to support TIFF

Useful links.

http://partners.adobe.com/public/developer/tiff/index.html

http://www.remotesensing.org/libtiff/

http://www.awaresystems.be/imaging/tiff.html

Current implementation can be found in the [tiff-codec](https://github.com/SixLabors/ImageSharp/tree/tiff-codec) branch.

### Feature Checklist

- [x] Reading TIFF file structure
- [x] Parsing of relevant metadata
- [x] Extraction of image data blocks
- [x] Construction of image (from strip data)
- [ ] Construction of image (from tile data)

### Compression Types

- [x] None
- [ ] Ccitt1D
- [x] PackBits
- [ ] CcittGroup3Fax
- [ ] CcittGroup4Fax
- [x] Lzw
- [ ] Old Jpeg
- [ ] Jpeg (Technote 2)
- [x] Deflate (Technote 2)
- [x] Old Deflate (Technote 2)

### Photometric Interpretation Formats

- [x] WhiteIsZero
- [x] BlackIsZero
- [x] Rgb (Chunky)
- [x] Rgb (Planar)
- [x] PaletteColor
- [ ] TransparencyMask
- [ ] Separated (TIFF Extension color spaces)
- [ ] YCbCr (TIFF Extension color spaces)
- [ ] CieLab (TIFF Extension color spaces)
- [ ] IccLab (TechNote 1)

JimBobSquarePants:
_From @kierenj on September 7, 2016 20:27_

I realise how unhelpful this comment right here is, but TIFF decoding across the board can be really tricky to get right - there are loads of compression formats, and even fairly respected tools (from MS / Atalasoft) get it wrong even with 1-bit images.  I think some archiving formats even require special licensing to implement correctly.  It might be worth identifying a small subset of formats to support!

JimBobSquarePants:
@kierenj It's certainly not an unhelpful comment. Tiff would be hard and I'm sorely tempted to let someone else figure it out and concentrate on what we have already. I'm just dreading the day someone tells me that the library is useless without it. There's always someone!

chrisvandijk:
I needed to import Tiff files so I created a temporary [Tiff encoder and decoder](https://github.com/chrisvandijk/ImageSharp-Tiff) based on LibTiff.Net.

[...]

PR #1527, opened 2021-01-31
Separate the tiff encoder classes

IldarKhayrutdinov:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

* Split `TiffWriter` class.
* Move decompressor classes.
 
Only splitting and moving classes, no change algorithms.

PR #1537, opened 2021-02-07
#12 Support multi strip encoding. Improve performance and memory usage.

IldarKhayrutdinov:
### Prerequisites

### Description

* Refactory color encoders and data compressors - it will make it easier to add new color modes (4-bit gray, CMYK, RGB 565/555, YCbCr), compressions (CcittGroup4Fax, Jpeg) and multi page encoding.
* Support multi strip encoding.
* Improve performance and memory usage of encoders and small improve of decoders.
* [**fixed** by @brianpopow] Report palette lzw bug

Encoing performance improved: **6-69%**, memory usage: **8-70%**.
Decoder performance stayed ±10%, maybe big tests will show better results.
I don't provided profiling, I think with profiling there is potential for more improvements.


IldarKhayrutdinov:
Please attention:

https://github.com/SixLabors/ImageSharp/commit/367d8bb0fa13114e609466564c812a2c9c3ebbbe#diff-916bcb855ed09d4491b4ccc51fee218af1aa197ec22d499cc4dddc31f6942a20R98-R125
This test I added before my changes, here reproduction incorrect setting color mode from compression. The `CcittGroup3Fax` and `Ccitt1D` compressions can use only with `mono color` mode.

Benchmark code also was corrected (instead of `biColor` it tested `uncompressed rgb`):
https://github.com/SixLabors/ImageSharp/pull/1537/commits/367d8bb0fa13114e609466564c812a2c9c3ebbbe#diff-fdeb1a0d0f29bf3fdecc5db3c8f7110377795fe75843f8858d8e89f5d1e7fdfaL71-L74
The `biColor` results became a very slowest: https://github.com/SixLabors/ImageSharp/pull/1537/commits/da856dadf5ebb0393a9486e5f34a5c1a888aafdf

After my changes the test work correctly:
https://github.com/SixLabors/ImageSharp/pull/1537/commits/8cca66761798260768fed650c08c54511071359e#diff-916bcb855ed09d4491b4ccc51fee218af1aa197ec22d499cc4dddc31f6942a20R97-R116

IldarKhayrutdinov:
**Attention 2**

In current [ImageSharp:tiff-format](https://github.com/SixLabors/ImageSharp/tree/tiff-format) branch following tests generate `black output images`. I don't understand, why they are successful:
https://github.com/SixLabors/ImageSharp/blob/c57664dee01d7b527ee17347161484a53a4ee978/tests/ImageSharp.Tests/Formats/Tiff/TiffEncoderTests.cs#L183-L201

Also I tested this bug in revision befor merging https://github.com/SixLabors/ImageSharp/pull/1527, the bug was there too.

This bug I found during multistrip tests. Therefore I add workaround for this tests and add bug reprodusion test:
https://github.com/SixLabors/ImageSharp/commit/5fc3d37794bc1ece5bf024d85157123bccb6ce71#diff-916bcb855ed09d4491b4ccc51fee218af1aa197ec22d499cc4dddc31f6942a20R210-R237

brianpopow:
> **Attention 2**
> 
> In current [ImageSharp:tiff-format](https://github.com/SixLabors/ImageSharp/tree/tiff-format) branch following tests generate **black output images**. I don't understand, why they are successful:

The test do not fail because with the palette image tests, we do not compare the result against the input image. There is a quantization step involved to create the palette, which means the encoded image will always be slightly different from the original.
Instead we load the encoded image with a reference decoder and compare against that. Not sure how we can improve that.

I tried to figure out the issue, but so far i cannot see any obvious bug. `tiffinfo` shows no error with the encoded image:

```
TIFF Directory at offset 0x1c1e (7198)
  Image Width: 804 Image Length: 1198
  Resolution: 96, 96 pixels/inch
  Bits/Sample: 8
  Compression Scheme: LZW
  Photometric Interpretation: palette color (RGB from colormap)
  Samples/Pixel: 1
  Rows/Strip: 1198
  Planar Configuration: single image plane
  Color Map: (present)
  DocumentName: .\Calliphora_palette_uncompressed.tiff
  ImageDescription: File source: http://commons.wikimedia.org/wiki/File:Calliphora_sp_Portrait.jpg
  Software: ImageSharp
```

IldarKhayrutdinov:
@brianpopow there are my outputs [EncodeColorPalette_WithLzw.zip](https://github.com/SixLabors/ImageSharp/files/5939173/EncodeColorPalette_WithLzw.zip)
from `\ImageSharp\tests\Images\ActualOutput\TiffEncoderTests\` path.

There may be a problem on my computer, please check your output?

brianpopow:
> @brianpopow there are my outputs [EncodeColorPalette_WithLzw.zip](https://github.com/SixLabors/ImageSharp/files/5939173/EncodeColorPalette_WithLzw.zip)
> from `\ImageSharp\tests\Images\ActualOutput\TiffEncoderTests\` path.
> 
> There may be a problem on my computer, please check your output?

@IldarKhayrutdinov i can reproduce the isssue, i do see those black images, too. I just dont know yet why this is happening.

IldarKhayrutdinov:
ok, I want to note that this problem was before this PR

> > @brianpopow there are my outputs [EncodeColorPalette_WithLzw.zip](https://github.com/SixLabors/ImageSharp/files/5939173/EncodeColorPalette_WithLzw.zip)
> > from `\ImageSharp\tests\Images\ActualOutput\TiffEncoderTests\` path.
> > There may be a problem on my computer, please check your output?
> 
> @IldarKhayrutdinov i can reproduce the isssue, i do see those black images, too. I just dont know yet why this is happening.

ok, I want to note that this problem was before this PR

IldarKhayrutdinov:
**Encoder performance report**
|    | Method                | Runtime       | Compression     | Mean       | Ratio | Allocated  | Performance | Memory |
|----|-----------------------|---------------|-----------------|------------|-------|------------|-------------|--------|
|    | 'System.Drawing Tiff' | .NET 4.7.2    | None            | 6.520 ms   | 1.00  | 11570062 B |             |        |
|    | 'ImageSharp Tiff'     | .NET 4.7.2    | None            | 5.698 ms   | 0.87  | 9919288 B  |             |        |
| PR | 'ImageSharp Tiff'     | .NET 4.7.2    | None            | 4.844 ms   | 0.73  | 7445922 B  | 6%          | 25%    |
|    | 'System.Drawing Tiff' | .NET Core 2.1 | None            | 6.851 ms   | 1.00  | 11562768 B |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 2.1 | None            | 4.294 ms   | 0.63  | 9918144 B  |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 2.1 | None            | 3.189 ms   | 0.46  | 7444718 B  | 27%         | 25%    |
|    | 'System.Drawing Tiff' | .NET Core 3.1 | None            | 5.835 ms   | 1.00  | 8672224 B  |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 3.1 | None            | 5.167 ms   | 0.89  | 9918112 B  |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 3.1 | None            | 3.342 ms   | 0.57  | 7444631 B  | 36%         | 25%    |
|    | 'System.Drawing Tiff' | .NET 4.7.2    | Deflate         | NA         | ?     | -          |             |        |
|    | 'ImageSharp Tiff'     | .NET 4.7.2    | Deflate         | 125.909 ms | ?     | 11167960 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET 4.7.2    | Deflate         | 87.815 ms  | ?     | 6617521 B  | 30%         | 40%    |
|    | 'System.Drawing Tiff' | .NET Core 2.1 | Deflate         | NA         | ?     | -          |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 2.1 | Deflate         | 125.041 ms | ?     | 11164792 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 2.1 | Deflate         | 84.005 ms  | ?     | 6605507 B  | 33%         | 40%    |
|    | 'System.Drawing Tiff' | .NET Core 3.1 | Deflate         | NA         | ?     | -          |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 3.1 | Deflate         | 125.139 ms | ?     | 11168428 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 3.1 | Deflate         | 81.102 ms  | ?     | 6604792 B  | 35%         | 40%    |
|    | 'System.Drawing Tiff' | .NET 4.7.2    | Lzw             | 49.024 ms  | 1.00  | 10673371 B |             |        |
|    | 'ImageSharp Tiff'     | .NET 4.7.2    | Lzw             | 411.728 ms | 8.41  | 23265464 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET 4.7.2    | Lzw             | 125.569 ms | 2.66  | 8423760 B  | 69%         | 64%    |
|    | 'System.Drawing Tiff' | .NET Core 2.1 | Lzw             | 47.288 ms  | 1.00  | 10668688 B |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 2.1 | Lzw             | 201.643 ms | 4.26  | 27451168 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 2.1 | Lzw             | 96.217 ms  | 2.03  | 8422488 B  | 52%         | 70%    |
|    | 'System.Drawing Tiff' | .NET Core 3.1 | Lzw             | 46.526 ms  | 1.00  | 8001741 B  |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 3.1 | Lzw             | 170.276 ms | 3.66  | 27451445 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 3.1 | Lzw             | 93.635 ms  | 2.02  | 8422504 B  | 44%         | 70%    |
|    | 'System.Drawing Tiff' | .NET 4.7.2    | PackBits        | NA         | ?     | -          |             |        |
|    | 'ImageSharp Tiff'     | .NET 4.7.2    | PackBits        | 28.948 ms  | ?     | 9943858 B  |             |        |
| PR | 'ImageSharp Tiff'     | .NET 4.7.2    | PackBits        | 27.449 ms  | ?     | 7453052 B  | 5%          | 25%    |
|    | 'System.Drawing Tiff' | .NET Core 2.1 | PackBits        | NA         | ?     | -          |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 2.1 | PackBits        | 22.611 ms  | ?     | 9942792 B  |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 2.1 | PackBits        | 19.935 ms  | ?     | 7451912 B  | 12%         | 25%    |
|    | 'System.Drawing Tiff' | .NET Core 3.1 | PackBits        | NA         | ?     | -          |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 3.1 | PackBits        | 23.465 ms  | ?     | 9942772 B  |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 3.1 | PackBits        | 19.664 ms  | ?     | 7451974 B  | 16%         | 25%    |
|    | 'System.Drawing Tiff' | .NET 4.7.2    | CcittGroup3Fax  | 43.618 ms  | 1.00  | 1169683 B  |             |        |
|    | 'ImageSharp Tiff'     | .NET 4.7.2    | CcittGroup3Fax  | 191.602 ms | 4.39  | 24829048 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET 4.7.2    | CcittGroup3Fax  | 191.413 ms | 4.42  | 22714336 B | -1%         | 8%     |
|    | 'System.Drawing Tiff' | .NET Core 2.1 | CcittGroup3Fax  | 43.258 ms  | 1.00  | 1169200 B  |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 2.1 | CcittGroup3Fax  | 177.930 ms | 4.11  | 24772997 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 2.1 | CcittGroup3Fax  | 180.059 ms | 4.13  | 22658509 B | -           | 8%     |
|    | 'System.Drawing Tiff' | .NET Core 3.1 | CcittGroup3Fax  | 43.330 ms  | 1.00  | 850189 B   |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 3.1 | CcittGroup3Fax  | 168.846 ms | 3.90  | 24774571 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 3.1 | CcittGroup3Fax  | 171.370 ms | 3.94  | 22658261 B | -1%         | 8%     |
|    | 'System.Drawing Tiff' | .NET 4.7.2    | ModifiedHuffman | 17.106 ms  | 1.00  | 11561706 B |             |        |
|    | 'ImageSharp Tiff'     | .NET 4.7.2    | ModifiedHuffman | 192.530 ms | 11.27 | 24826163 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET 4.7.2    | ModifiedHuffman | 191.066 ms | 11.18 | 22710384 B | <1%         | 8%     |
|    | 'System.Drawing Tiff' | .NET Core 2.1 | ModifiedHuffman | 16.988 ms  | 1.00  | 11555088 B |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 2.1 | ModifiedHuffman | 180.265 ms | 10.61 | 24769453 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 2.1 | ModifiedHuffman | 177.379 ms | 10.41 | 22656395 B | 1%          | 8%     |
|    | 'System.Drawing Tiff' | .NET Core 3.1 | ModifiedHuffman | 15.989 ms  | 1.00  | 8666467 B  |             |        |
|    | 'ImageSharp Tiff'     | .NET Core 3.1 | ModifiedHuffman | 181.295 ms | 11.34 | 24770275 B |             |        |
| PR | 'ImageSharp Tiff'     | .NET Core 3.1 | ModifiedHuffman | 167.231 ms | 10.49 | 22659275 B | 7%          | 8%     |

IldarKhayrutdinov:
@all there were a lot of changes, don't hesitate to comments 😉

src/ImageSharp/Formats/Tiff/TiffEncoderPixelStorageMethod.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Formats.Experimental.Tiff
+{
+    /// <summary>
+    /// The tiff encoder pixel storage method.
+    /// </summary>
+    public enum TiffEncoderPixelStorageMethod
```
IldarKhayrutdinov, line 9: I could not think of a better name, it can be renamed 😄

src/ImageSharp/Formats/Tiff/TiffEncoderCore.cs
```diff
@@ -30,6 +31,8 @@ internal sealed class TiffEncoderCore : IImageEncoderInternals
                 ? TiffConstants.ByteOrderLittleEndianShort
                 : TiffConstants.ByteOrderBigEndianShort;
 
+        private const int DefaultStripSize = 8 * 1024;
```
IldarKhayrutdinov, line 34: Default strip size 8Kb

src/ImageSharp/Formats/Tiff/TiffEncoderCore.cs
```diff
@@ -159,27 +150,51 @@ public long WriteHeader(TiffStreamWriter writer)
...
             writer.WriteMarker(ifdOffset, (uint)writer.Position);
             long nextIfdMarker = this.WriteIfd(writer, entriesCollector.Entries);
+        }
+
+        private int CalcRowsPerStrip(ImageFrame image, int bytesPerRow)
+        {
+            switch (this.storageMode)
+            {
+                default:
+                case TiffEncoderPixelStorageMethod.Auto:
```
IldarKhayrutdinov, line 188: default use multi strip method

src/ImageSharp/Formats/Tiff/TiffEncoderCore.cs
```diff
@@ -239,37 +254,51 @@ public long WriteIfd(TiffStreamWriter writer, List<IExifValue> entries)
...
-        /// <param name="image">The <see cref="Image{TPixel}" /> to encode from.</param>
-        /// <param name="entriesCollector">The entries collector.</param>
-        /// <param name="imageDataStartOffset">The start of the image data in the stream.</param>
-        /// <param name="imageDataBytes">The image data in bytes to write.</param>
-        public void AddStripTags<TPixel>(Image<TPixel> image, TiffEncoderEntriesCollector entriesCollector, uint imageDataStartOffset, int imageDataBytes)
-            where TPixel : unmanaged, IPixel<TPixel>
+        private void SetMode(Image image)
         {
-            var stripOffsets = new ExifLongArray(ExifTagValue.StripOffsets)
+            if (this.CompressionType == TiffEncoderCompression.CcittGroup3Fax || this.CompressionType == TiffEncoderCompression.ModifiedHuffman)
```
IldarKhayrutdinov, line 259: for mon `CcittGroup3Fax` and `ModifiedHuffman` force use `BiColor` mode.

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
```diff
@@ -3,208 +3,97 @@
...
             // TODO: Should we allow to skip this by the user, if its known to be black and white already?
-            using Image<TPixel> imageBlackWhite = image.Clone();
-            imageBlackWhite.Mutate(img => img.BinaryDither(default(ErrorDither)));
-
-            if (compression == TiffEncoderCompression.Deflate)
-            {
-                return this.WriteBiColorDeflate(imageBlackWhite, pixelRowAsGraySpan, outputRow, compressionLevel);
-            }
+            this.imageBlackWhite = new Image<TPixel>(configuration, new ImageMetadata(), new[] { image.Clone() });
+            this.imageBlackWhite.Mutate(img => img.BinaryDither(default(ErrorDither)));
```
IldarKhayrutdinov, line 30: `BiColor` encoder use hard `Mutate` operation, I think this is the reason for the slow results of benchmarks

src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter.cs
```diff
@@ -3,58 +3,47 @@
...
+        private const int ColorPaletteSize = ColorsPerChannel * 3;
+        private const int ColorPaletteBytes = ColorPaletteSize * 2;
+
+        private readonly IndexedImageFrame<TPixel> quantized;
+
+        public TiffPaletteWriter(ImageFrame<TPixel> image, IQuantizer quantizer, MemoryAllocator memoryAllocator, Configuration configuration, TiffEncoderEntriesCollector entriesCollector)
+             : base(image, memoryAllocator, configuration, entriesCollector)
         {
+            using IQuantizer<TPixel> frameQuantizer = quantizer.CreatePixelSpecificQuantizer<TPixel>(this.Configuration);
+            this.quantized = frameQuantizer.BuildPaletteAndQuantizeFrame(image, image.Bounds());
```
IldarKhayrutdinov, line 29: `Palette` encoder also use hard operations

src/ImageSharp/Formats/Tiff/TiffEncoderCore.cs
```diff
@@ -159,27 +150,51 @@ public long WriteHeader(TiffStreamWriter writer)
...
             long nextIfdMarker = this.WriteIfd(writer, entriesCollector.Entries);
+        }
+
+        private int CalcRowsPerStrip(ImageFrame image, int bytesPerRow)
+        {
+            switch (this.storageMode)
+            {
+                default:
+                case TiffEncoderPixelStorageMethod.Auto:
+                case TiffEncoderPixelStorageMethod.MultiStrip:
```
IldarKhayrutdinov, line 189: default use multi strip method

src/ImageSharp/Formats/Tiff/TiffEncoderPixelStorageMethod.cs
IldarKhayrutdinov: to this enum possible to add tile, planar

src/ImageSharp/Formats/Tiff/ITiffEncoderOptions.cs
```diff
@@ -37,5 +37,15 @@ internal interface ITiffEncoderOptions
...
+
+        /// <summary>
+        /// Gets the pixel storage method.
+        /// </summary>
+        TiffEncoderPixelStorageMethod PixelStorageMethod { get; }
+
+        /// <summary>
+        /// Gets the maximum size of strip (bytes).
+        /// </summary>
+        int MaxStripBytes { get; }
```
IldarKhayrutdinov, line 49: this is a limit of one strip length (the field also can be used for tiles)

IldarKhayrutdinov: this field can also be renamed 😄

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
IldarKhayrutdinov: we can check the source image, if it is already `bicolor`, we can use source directly

src/ImageSharp/Formats/Tiff/TiffEncoderPixelStorageMethod.cs
JimBobSquarePants: I would have thought whatever the enum is then it would be used for both decoder and encoder no? Would this not be part of the retrieved metadata?

IldarKhayrutdinov: this is settings for decoder only, like [TiffEncodingMode](https://github.com/SixLabors/ImageSharp/blob/c57664dee01d7b527ee17347161484a53a4ee978/src/ImageSharp/Formats/Tiff/TiffEncodingMode.cs)

src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter.cs
JimBobSquarePants: This should utilize whatever bit depth property we store.

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
JimBobSquarePants: `default(ErrorDither)` is a big red flag here, it won't produce any output since there's no matrix. We should be using one of the static properties. `ErrorDither.FloydSteinberg` or similar. Looks like LibTiff uses that one. http://www.libtiff.org/man/tiffmedian.1.html

>we can check the source image, if it is already bicolor, we can use source directly

We cannot do that since it will almost always be different from the output.

src/ImageSharp/Formats/Tiff/TiffEncoderPixelStorageMethod.cs
IldarKhayrutdinov: it seems, this enum can be removed and always use multiStrip mode

src/ImageSharp/Formats/Tiff/ITiffEncoderOptions.cs
brianpopow: i think `MaxStripBytes` is an ok name, its clear what it does.

JimBobSquarePants: Is that interface public? I’m curious as to what the property means as a configuration value to the average user

src/ImageSharp/Formats/Tiff/TiffEncoderPixelStorageMethod.cs
brianpopow: i would say remove this enum, and always use multiStrip, as its recommended by the spec anyway.

src/ImageSharp/Formats/Tiff/Compression/Compressors/DeflateCompressor.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            if (this.Predictor == TiffPredictor.Horizontal)
+            {
+                HorizontalPredictor.ApplyHorizontalPrediction(rows, this.BytesPerRow, this.BitsPerPixel);
+            }
+
+            stream.Write(rows);
+
+            stream.Flush();
+            //// stream.Dispose(); // todo: dispose write crc
```
brianpopow, line 41: the stream should really be disposed before getting the size, otherwise the crc will not be written as the comment states.
I wonder how this actually works without it.

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
IldarKhayrutdinov: > `default(ErrorDither)` is a big red flag here, it won't produce any output since there's no matrix. We should be using one of the static properties. `ErrorDither.FloydSteinberg` or similar. Looks like LibTiff uses that one. http://www.libtiff.org/man/tiffmedian.1.html

@JimBobSquarePants yeah, it's required to use `ErrorDither.FloydSteinberg` (or another that we can add to encoder options ), in this PR I didn't change this instruction, I just moved this instruction to a new method. 

> > we can check the source image, if it is already bicolor, we can use source directly
> 
> We cannot do that since it will almost always be different from the output.
Yeah, it's a narrow cases, for example, when re-saving or joining several monochrome scans into single tiff (in the future).
I mean, if we had `L1` (`L8` analogy) we can skipping hard dithering operation.

src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter.cs
IldarKhayrutdinov: @JimBobSquarePants Sorry, I don't understand, what did you mean?

src/ImageSharp/Formats/Tiff/ITiffEncoderOptions.cs
IldarKhayrutdinov: The interface is `internal`. 
I looked `StripByteCounts` on different `tiff` files and watched that they have different lengths, not only 8Kb, in some cases it reaches 1Mb. So I figured that the user might want to set the strip size.

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
brianpopow: > @JimBobSquarePants yeah, it's required to use ErrorDither.FloydSteinberg (or another that we can add to encoder options > ?), in this PR I didn't change this instruction, I just moved this instruction to a new method.

@IldarKhayrutdinov this was my mistake, i can change that in the tiff branch, you dont have to do that (but that might result in a merge conflict here)

src/ImageSharp/Formats/Tiff/Compression/Compressors/DeflateCompressor.cs
IldarKhayrutdinov: I think that readers are ignored the `CRC`, I'll add crc writing.

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
IldarKhayrutdinov: @brianpopow Yeah, there are still such mistakes, and mine too 😄 but we're close to RC!
So that there are no conflicts, it is better on make in this PR, one of these days I will deal with remarks

JimBobSquarePants: Do you think it could be the cause of your blank images?

src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter.cs
JimBobSquarePants: I'm seeing a passed `IQuantizer` instance but also hardcoded `ColorsPerChannel` values which indicate 256 colors each time. I would assume that there is some sort of `BitDepth` property that is captured during decoding and can be optionally set to dictate the encoding bit depth. There should be consistency in the palette size and desired bit depth.

src/ImageSharp/Formats/Tiff/ITiffEncoderOptions.cs
JimBobSquarePants: Pardon my ignorance but what do the different strip count length mean in terms of encoding?

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
IldarKhayrutdinov: > Do you think it could be the cause of your blank images?

shouldn't be out of this, on weekend I'll debugging empty images

src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter.cs
IldarKhayrutdinov: Yeah, it's required to bit depth settings, but at the moment support only 8 bit palette and grayscale.
I would like add bit depth later, in future PR.

src/ImageSharp/Formats/Tiff/ITiffEncoderOptions.cs
IldarKhayrutdinov: This is the size of encoder buffer, the image is divided into `strips` and encoding one by one, so there is no need to allocate additional large buffers. It seems that there is no single agreement, there is only a recommendation about `8Kb` (also I read about `64Kb`), I think most users will be satisfied with default size. But maybe someone wants to adjust the strip size himself. Maybe @brianpopow will give more complete information?

IldarKhayrutdinov: frankly, for modern computers, I think it won't be a problem using `1Mb` buffers

JimBobSquarePants: Thanks... I guess we can determine a sensible default by benchmarking.

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
brianpopow: > Do you think it could be the cause of your blank images?

Im still trying to figure out the reason for the black images. At the moment i think its not actually in the encoder, but rather an issue with the decoder.
If you just save the image like this:

```
image.Save("output.tiff", new TiffEncoder()
{
    Compression = TiffEncoderCompression.Lzw,
    Mode = TiffEncodingMode.ColorPalette
});
```

The image looks fine and `tiffinfo` from libtiff does not complain about it.

brianpopow:
@IldarKhayrutdinov there seems to be an issue now with the `packedbits` compression. While the thumbnail of the tiffs look fine, if you open the image it looks broken. See for example in the Actual output folder of the tests: `TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.tiff`

Maybe we should add a CompareToReference output step in the tests to cover those cases.

src/ImageSharp/Formats/Tiff/TiffEncoderPixelStorageMethod.cs
IldarKhayrutdinov: Done 4895521f63395962d06555062292b79229f1e109

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
IldarKhayrutdinov: @brianpopow could you attach your output?

src/ImageSharp/Formats/Tiff/TiffEncoderPixelStorageMethod.cs
IldarKhayrutdinov: > @IldarKhayrutdinov there seems to be an issue now with the `packedbits` compression. While the thumbnail of the tiffs look fine, if you open the image it looks broken. See for example in the Actual output folder of the tests: `TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.tiff`
> 
> Maybe we should add a CompareToReference output step in the tests to cover those cases.

@brianpopow here is my output, it seems it looks normal, maybe can be different results on different computers?
[TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.zip](https://github.com/SixLabors/ImageSharp/files/5975500/TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.zip)

I think adding CompareToReference is a good idea.

IldarKhayrutdinov:
> @IldarKhayrutdinov there seems to be an issue now with the `packedbits` compression. While the thumbnail of the tiffs look fine, if you open the image it looks broken. See for example in the Actual output folder of the tests: `TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.tiff`
> 
> Maybe we should add a CompareToReference output step in the tests to cover those cases.

@brianpopow here is my output, it seems it looks normal, maybe can be different results on different computers?
[TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.zip](https://github.com/SixLabors/ImageSharp/files/5975500/TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.zip)

I think adding CompareToReference is a good idea.

brianpopow:
> > @IldarKhayrutdinov there seems to be an issue now with the `packedbits` compression. While the thumbnail of the tiffs look fine, if you open the image it looks broken. See for example in the Actual output folder of the tests: `TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.tiff`
> > Maybe we should add a CompareToReference output step in the tests to cover those cases.
> 
> @brianpopow here is my output, it seems it looks normal, maybe different results on different computers?
> [TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.zip](https://github.com/SixLabors/ImageSharp/files/5975500/TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed.zip)
> 
> I think adding CompareToReference is a good idea.

I doubt that the output can be different on different computers. We do not use any hardware intrinsics when decoding tiffs.

After digging  a bit deeper and that you mentioned it does not look corrupted on your pc, it seems an issue with IrfanView, gimp and XnView. On the other hand viewing it wit paint works. I definitely think there is a regression introduced with this PR and PackedBits compression, as i dont see this issue with the current tiff branch.

I will try to improve on the tests so we will spot those issues.

Here is an example output:
TiffEncoder_EncodeColorPalette_WithPackBitsCompression_Works_Rgba32_Calliphora_palette_uncompressed tiff

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
brianpopow: here is an output example with the above mentioned settings:

[lzw_colorpalette.zip](https://github.com/SixLabors/ImageSharp/files/5975930/lzw_colorpalette.zip)

Im working currently on an replacement implementation for the lzw decoder. Looks good so far but i dont want to promise too much until its done. Im pretty sure the black image issue is the decoder.

IldarKhayrutdinov:
> I doubt that the output can be different on different computers. We do not use any hardware intrinsics when decoding tiffs.

yes, it confused me too...

now I also saw through `IrfanView` (earlier I looked through the standard `windows` viewer and `MS Paint`)

IldarKhayrutdinov:
@brianpopow I investigated the `lzw` issue and finally understood, cause of `decoder` affecting for `encoder` test. 
The encoder test method: https://github.com/SixLabors/ImageSharp/blob/ed6078e8a70ebc78972a82ca001e98472a72eca3/tests/ImageSharp.Tests/Formats/Tiff/TiffEncoderTests.cs#L222-L226

line `using var encodedImage = (Image<TPixel>)Image.Load(this.configuration, memStream);`

here output image reload and resave, so in output directory `tests\Images\ActualOutput\TiffEncoderTests` placed resaved image.

brianpopow:
> @brianpopow I investigated the `lzw` issue and finally understood, cause of `decoder` affecting for `encoder` test.
> The encoder test method:
> 
> https://github.com/SixLabors/ImageSharp/blob/ed6078e8a70ebc78972a82ca001e98472a72eca3/tests/ImageSharp.Tests/Formats/Tiff/TiffEncoderTests.cs#L222-L226
> 
> line `using var encodedImage = (Image<TPixel>)Image.Load(this.configuration, memStream);`
> 
> here output image reload and resave, so in output directory `tests\Images\ActualOutput\TiffEncoderTests` placed resaved image.

yeah that was an unlucky workaround for the fact that the quantization was changing the output image slightly. I have changed that part now and use a TolerantComparer for the color palette tests: a8f75b06. This should fail now if the output has too much different from the input.

IldarKhayrutdinov:
> yeah that was an unlucky workaround for the fact that the quantization was changing the output image slightly. I have changed that part now and use a TolerantComparer for the color palette tests: [a8f75b0](https://github.com/SixLabors/ImageSharp/commit/a8f75b06f7e687e9edff452a390b287d137781e0). This should fail now if the output has too much different from the input.

👍 

I researching loading issue with `IrfanView`

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter.cs
brianpopow: The black image issue is now fixed in the `tiff-format` branch.

JimBobSquarePants:
```c#
using var encodedImage = (Image<TPixel>)Image.Load(this.configuration, memStream); 
```

Why the cast here and not using the generic version directly? That cast will only work if `TPixel` is `Rgba32`

tests/ImageSharp.Tests/Formats/Tiff/TiffEncoderTests.cs
```diff
@@ -121,11 +121,10 @@ public void TiffEncoder_CorrectBiMode<TPixel>(TestImageProvider<TPixel> provider
         [InlineData(TiffEncodingMode.ColorPalette, TiffEncoderCompression.ModifiedHuffman)]
         [InlineData(TiffEncodingMode.Gray, TiffEncoderCompression.ModifiedHuffman)]
         [InlineData(TiffEncodingMode.Rgb, TiffEncoderCompression.ModifiedHuffman)]
-        public void TiffEncoder_IncompatibilityOptions<TPixel>(TiffEncodingMode mode, TiffEncoderCompression compression)
-            where TPixel : unmanaged, IPixel<TPixel>
+        public void TiffEncoder_IncompatibilityOptions(TiffEncodingMode mode, TiffEncoderCompression compression)
         {
             // arrange
-            using var input = new Image<TPixel>(10, 10);
+            using var input = new Image<Rgb24>(10, 10);
```
JimBobSquarePants, line 127: This should remain `TPixel`

IldarKhayrutdinov:
> ```cs
> using var encodedImage = (Image<TPixel>)Image.Load(this.configuration, memStream); 
> ```
> 
> Why the cast here and not using the generic version directly? That cast will only work if `TPixel` is `Rgba32`

@JimBobSquarePants now this method has removed, now use `TestTiffEncoderCore` method:  https://github.com/SixLabors/ImageSharp/blob/0b8eac5faf395b5b95e5a7535dcc42a5543da2c6/tests/ImageSharp.Tests/Formats/Tiff/TiffEncoderTests.cs#L344

IldarKhayrutdinov:
@brianpopow I fixed `PackBits` bug, please check, is everything fixed?

brianpopow:
> @brianpopow I fixed `PackBits` bug, please check, is everything fixed?

@IldarKhayrutdinov great job in finding this issue! I will try to get this reviewed tomorrow.

brianpopow (approved):
Looks good, @IldarKhayrutdinov thank you once again for your help!

Issue #1498, opened 2021-01-10
Resize: enable opting-out for alpha premultiplication


ptasev:
**Edit:** this turned into a feature request for a `ResizeOptions.PremultiplyAlpha` property defaulting to `true`.

Adding up-for-grabs, since it's easy, here are steps for a potential community PR:
https://github.com/SixLabors/ImageSharp/issues/1498#issuecomment-758564510

----

### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When using Resize(int width, int height) I am getting very odd black bands in the output image. The results seems quite wrong. I am getting a different results when using GIMP's Bicubic Scale Image function.

### Steps to Reproduce
I have attached sample code that results in this issue. I have included the input image in the sample, as well as the output I get on my PC for reference.

### System Configuration

- ImageSharp version: 1.0.2
- Other ImageSharp packages and versions: N/A
- Environment (Operating system, version and so on): I am using Win 10 Pro 2004, with Ryzen 2600, and AMD Radeon R9 380.
- .NET Framework version: .NET 5.0
- Additional information: N/A

Input:
![image](https://user-images.githubusercontent.com/23424044/104130596-7020ea00-5326-11eb-93da-3cfb7ebf8ded.png)

My ImageSharp output as seen in GIMP:
![image](https://user-images.githubusercontent.com/23424044/104130404-71054c00-5325-11eb-9a6a-e5aebfbd108a.png)

GIMP's bicubic scale:
![image](https://user-images.githubusercontent.com/23424044/104130694-fd643e80-5326-11eb-8160-e3847ddb27ca.png)

[ResizeBugProj.zip](https://github.com/SixLabors/ImageSharp/files/5792985/ResizeBugProj.zip)

JimBobSquarePants:
Can you post a side by side comparison of the two outputs?

ptasev:
I have updated the post.

JimBobSquarePants:
I’m sorry but I still do not understand what the issue you are trying to highlight is? Is that second image showing the ImageSharp output image as black/white?

ptasev:
[...]

PR #1540, opened 2021-02-08
Binary threshold on saturation instead of luminescence value.

Franz1960:
### Prerequisites


antonfirsov (changes requested):
This is gonna be a breaking change in the proposed form. (New overload with a default arg replacing the existing one.) You should refactor default args. to separate overloads.

You also need to add tests in form of a new `[Theory]` to the following test class:
https://github.com/SixLabors/ImageSharp/blob/a1784a67bf492b1b02e99655a39aa5e5acf81988/tests/ImageSharp.Tests/Processing/Processors/Binarization/BinaryThresholdTest.cs

See second part of https://github.com/SixLabors/ImageSharp/issues/1498#issuecomment-758564510 for guidance on extending tests. (Difference is that the recommendations there for `ResizeTests`, replace them with `BinaryThresholdTest` when reading.)

Summary: 3 things to do in the PR:
1. Enum instead of `bool` arg
2. Separate overloads
3. Add tests

src/ImageSharp/Processing/Extensions/Binarization/BinaryThresholdExtensions.cs
```diff
@@ -16,9 +16,12 @@ public static class BinaryThresholdExtensions
         /// </summary>
         /// <param name="source">The image this method extends.</param>
         /// <param name="threshold">The threshold to apply binarization of the image. Must be between 0 and 1.</param>
+        /// <param name="useSaturationNotLuminance">
+        /// Use saturation value instead of luminance.
+        /// </param>
         /// <returns>The <see cref="IImageProcessingContext"/> to allow chaining of operations.</returns>
-        public static IImageProcessingContext BinaryThreshold(this IImageProcessingContext source, float threshold) =>
-            source.ApplyProcessor(new BinaryThresholdProcessor(threshold));
+        public static IImageProcessingContext BinaryThreshold(this IImageProcessingContext source, float threshold, bool useSaturationNotLuminance = false) =>
```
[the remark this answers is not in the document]
Franz1960: Thank you for the recommendations. I should find time til tomorrow.

Franz1960:
@antonfirsov: Is the overload in the current form still a breaking change?

antonfirsov:
@Franz1960 yeah, the problem is that at IL level `Foo(a,b=0)` is not the same as `Foo(a)` + a separate `Foo(a,b)`. Here is a short explanation with IL:
https://stackoverflow.com/a/12671182

And here is a (somewhat) longer article:
https://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/

TLDR: If we were designing this API from scratch, we could just use default args, but since it's not the case, we need to add separate overloads for each version in order to avoid unexpected `MissingMethodException`-s for users who already took dependency on any of the overloads, and are updating ImageSharp *without rebuilding their app*.

src/ImageSharp/Processing/Extensions/Binarization/BinaryThresholdExtensions.cs
```diff
@@ -28,12 +29,14 @@ public static class BinaryThresholdExtensions
...
         /// The <see cref="Rectangle"/> structure that specifies the portion of the image object to alter.
         /// </param>
+        /// <param name="valueType">The color component to be compared to threshold.</param>
         /// <returns>The <see cref="IImageProcessingContext"/> to allow chaining of operations.</returns>
         public static IImageProcessingContext BinaryThreshold(
             this IImageProcessingContext source,
             float threshold,
-            Rectangle rectangle) =>
-            source.ApplyProcessor(new BinaryThresholdProcessor(threshold), rectangle);
+            Rectangle rectangle,
```
antonfirsov, line 37: Additionally, `Rectangle` should be the last argument when there `treshold` and/or `valueType`  is present (for consistency).

src/ImageSharp/Processing/Processors/Binarization/BinaryThresholdProcessor.cs
```diff
@@ -5,6 +5,22 @@
 
 namespace SixLabors.ImageSharp.Processing.Processors.Binarization
 {
+    /// <summary>
+    /// The color component to be compared to threshold.
+    /// </summary>
+    public enum BinaryThresholdValueType : int
```
JimBobSquarePants, line 11: `BinaryThresholdComponentKind` ? Naming is hard...

Franz1960: Oh yes, that's true.
My naming candidates have been
- BinaryThresholdColorComponent
- BinaryThresholdValue
- BinaryThresholdValueType
Now Jim added
- BinaryThresholdComponentKind
My intention would be to make the enumeration expandable for each color component of each color model.
I'm open regarding names. How can we find the best one?

src/ImageSharp/Processing/Extensions/Binarization/BinaryThresholdExtensions.cs
Franz1960: Ok, thank you for clarification. I will implement it soon.

One more question: Code coverage test continues to fail even though I have extended the test methods to pass the branch with saturation. What am I doing wrong?

src/ImageSharp/Processing/Processors/Binarization/BinaryThresholdProcessor.cs
brianpopow: What about `BinaryThresholdMethod`?

antonfirsov: If I get it right, it's not a method, it's just just a value selector.

src/ImageSharp/Processing/Extensions/Binarization/BinaryThresholdExtensions.cs
antonfirsov: I think the problem is on the codecov (tool) side not on yours. I'm not an expert on that, @JimBobSquarePants any idea what's going on?

JimBobSquarePants: What’s happening with that new test? We’re doing a new comparison with saturation but I do not see how since the submodule has not been updated.

antonfirsov: There was no validation in the old tests, so there is none in the new one (just debug save).

JimBobSquarePants: https://github.com/SixLabors/ImageSharp/pull/1540/files#diff-88fae6a5fdda36fa3f6f2ba33c87415448b8db16c5320bdf437ee75235381be1R83

JimBobSquarePants: I’m on my phone so can’t check what that verify method is doing

antonfirsov: If memory serves well, that method just checks if the area outside the rect was left unchanged.

src/ImageSharp/Processing/Processors/Binarization/BinaryThresholdProcessor.cs
antonfirsov: `BinaryThresholdComponentKind` feels better than `BinaryThresholdValueType`, but based on the semantics you described, my favorite from the list above is `BinaryThresholdColorComponent`. I don't think we need a postfix like `*Type` or `*Kind` when the name already ends to `*Component`. @JimBobSquarePants thoughts?

tests/ImageSharp.Tests/Processing/Processors/Binarization/BinaryThresholdTest.cs
```diff
@@ -53,5 +54,34 @@ public void ImageShouldApplyBinaryThresholdInBox<TPixel>(TestImageProvider<TPixe
...
+
+        [Theory]
+        [WithFileCollection(nameof(CommonTestImages), nameof(BinaryThresholdValues), PixelTypes.Rgba32)]
+        public void ImageShouldApplyBinarySaturationThresholdFilter<TPixel>(TestImageProvider<TPixel> provider, float value)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            using (Image<TPixel> image = provider.GetImage())
+            {
+                image.Mutate(x => x.BinaryThreshold(value, BinaryThresholdValueType.Saturation));
+                image.DebugSave(provider, value);
```
antonfirsov, line 66: @Franz1960 can you please double check if the test results saved to your local `tests/Images/ActualOutput/BinaryThresholdTest/` meet your expectations (Saturation vs Luminance)?

Normally we'd have a validation call here in these functional tests (`image.CompareToReferenceOutput(...)`), comparing the result to a pre-existing reference output, but unfortunately `BinaryThresholdTest` haven't been updated to have this validation.

I don't want to block this PR on this, but my testing OCD tells that we should at least visually validate the output.

src/ImageSharp/Processing/Processors/Binarization/BinaryThresholdProcessor.cs
JimBobSquarePants: Yeah... `BinaryThresholdComponent` works well. We've managed to avoid the `**Kind` suffix elsewhere.

tests/ImageSharp.Tests/Processing/Processors/Binarization/BinaryThresholdTest.cs
Franz1960: I included comparisons to ref outputs. Used a rgb image which is better fitted to saturation operations. Added a PR in 
Imagesharp.Tests.Images repo with new ReferenceOutput images. Is this the right way to bring the ref images to work?

JimBobSquarePants: Yep, now use 

``` bash
```
In your branch root to pull those reference images in. 

I'm going to pull down your fork later and do a performance pass as I've spotted scope for some improvement.

src/ImageSharp/Processing/Processors/Binarization/BinaryThresholdProcessor.cs
```diff
@@ -19,6 +19,12 @@ public enum BinaryThresholdColorComponent : int
         /// Saturation color component as value to be compared to threshold.
         /// </summary>
         Saturation = 1,
+
+        /// <summary>
+        /// Colorfulness (saturated with min luminance 10%) color component as value to be
+        /// compared to threshold.
+        /// </summary>
+        Colorfulness_L10 = 2,
```
JimBobSquarePants, line 27: This isn't part of the original specification.

Franz1960: Ok, I see your point. I would see a certain degree of general purpose because saturation is really a problem at the dark end of the color space. At the light end saturation is always low, but rgb(0,1,0) has high saturation though it's practically black. The proposed Colorfulness_L10 works well in our toad application, and different L thresholds might be realised. A generalization with L threshold as argument would be possible but perhaps less convenient for users.

I fully understand if you don't want to have this in your lib, but we do need it. Would you recommend a fork? or is there a way to make extensions like that without integrating it?

Thanks and cheers,
Franz

antonfirsov: I don't like the idea of introducing enum-driven arbitrary strategies (they may be useless for most users), but we can generalize the feature by passing something like a `Func<Rgba32,byte,bool>` or `Func<Vector4,float,bool>` or equivalent to the processor. It's not gonna be trivial work, but I'm happy to guide you with such a contribution if you're interested.

Alternatively, you can just copy the code of `BinaryThresholdProcessor` and `BinaryThresholdProcessor<T>` into your application, and apply the changes there. Forking the entire project is unnecessary.

JimBobSquarePants: Apologies for the silence. Day job is eating up all my waking hours. 

I think even a func overload would have to be very carefully designed so it wasn’t a performance killer and feels very niche. I’m happy with the general enum but specific cases can be delivered like Anton said. The library is designed to be extensible.

Franz1960: First I want to thank you for your time spent on this PR, both of you.

My arbitrarily defining a 'Colorfulness_L10' color component was too pragmatic. I will have a closer look on color standards like BT.709 or JPEG YCbCr. We need a scalar value to compare to a threshold. This value has to represent colorfulness or chromaticity to the human eye for my purposes. The HSL saturation value comes near but fails for dark colors. The distance of Cb or Cr to the achromatic level AL would be a candidate.

    Chroma = Max(Abs(Cb - AL), Abs(Cr - AL))

Would that be general purpose in your eyes?

antonfirsov: I like the approach but my understanding of colorspaces is not good enough not to evaluate the proposal.

Quick sanity test: How would you name this "mode"? How would you explain it in docs? Have you tested it already?

Franz1960: I think I understand the software requirements better now and will investigate the issue more thoroughly first. When I have a suitable solution from my point of view, I will make a corresponding proposal. Until then I will work on a branch without PRs. Maybe I will do a merge to saturation_threshold from time to time to see the results of the tests. Please just ignore and let the PR stay open! I will get back to you when I have a state that is ok from my point of view.

antonfirsov:
Marking this as a draft while discussing the new proposal.

@Franz1960 thanks for your interest and patience, I really appreciate it!

Franz1960:
If my PR #17 on Imagesharp.Tests.Images will be merged, these tests should run through.

JimBobSquarePants:
Test images merged. I'll do that performance pass later.

antonfirsov:
@Franz1960 do the new tests pass locally for you? I see test failures both because of missing reference output and big difference in output.

SixLabors/Imagesharp.Tests.Images#17 should be most likely reverted and redone.

When doing so, I recommend to run the [optipng tool](https://github.com/SixLabors/Imagesharp.Tests.Images/tree/master/tools) on the new reference images, they seem to be unreasonably large resulting in >1MB addition to the repo.

Franz1960:
@antonfirsov Locally all tests passed. The optipng tool only shrinked the small files a little (black white only) so I did not use it.
I think the merge of Imagesharp.Tests.Images PR #17 had not been complete (it had been originally rejected by @JimBobSquarePants), so I make a new branch and PR #18, which has to be merged before tests will pass.

JimBobSquarePants:
@Franz1960 I didn't reject anything...

https://github.com/SixLabors/Imagesharp.Tests.Images/pull/17

Btw you keep linking to the wrong locations. Hashes only work within the same repository.

Franz1960:
Sorry I obvously don't understand the workflow with the Imagesharp.Tests.Images submodule. I forked it (Franz1960/Imagesharp.Tests.Images), made a branch "saturation_threshold", pushed the new and changed test images and made a PR (#17). @JimBobSquarePants commented it was beyond the accepted spec and I made changes accordingly and made commits of test images. Now in SixLabors/Imagesharp.Tests.Images / master all images are present, but the PR tests on ImageSharp failed.

How would I do this correctly?

antonfirsov:
@Franz1960 based on what you describe you did things conceptually right, but some mistake happened on the way. What I see is that there are only 4 images pushed to the PR, while you modified 22 images in SixLabors/Imagesharp.Tests.Images#17. Maybe you did not point to the right commit with your submodule update? I recommend to do a fresh clone of your fork to an empty directory, and running the tests there. That should give you some clues on what went wrong.

FYI: We realized that the submodule workflow is a huge burden, so @JimBobSquarePants is currently working on getting rid of it by merging the reference images into the repository + introducing [git LFS](https://git-lfs.github.com/) for images. It will be a radical step with history rewrite on master. So an alternative to the approach I suggested in the previous paragraph is to wait until that work is done.

The preferred solution also depends on @JimBobSquarePants 's ETA on nuking down the submodules.

JimBobSquarePants:
Ok.... So introducing Git LFS requires rewriting Git history which means this PR's history is now out of sync with our repository and GitHub has now closed it. 

My advice @Franz1960 , for simplicities sake, is to create a fresh clone of our repository and add the required changes to a new PR. You will be able to add the images now directly to the references folder but first ensure you have [Git Large File Storage](https://docs.github.com/en/github/managing-large-files/installing-git-large-file-storage) installed.

PR #1555, opened 2021-02-17
New overloads for binary threshold operations. The new argument Color…

Franz1960:
…Component defines the scalar color component to be used for threshold comparison: Luminance, Saturation or MaxChroma. Luminance is default and identical to previous versions. Saturation is the HSL saturation component. MaxChroma is calculated as the maximum of YCbCr chroma value, i.e. Cb and Cr distance from achromatic value. Background: This component shall discriminate colorful parts from achromatic parts in human perception. Very dark pixels, which are perceived as near black, can have high HSL saturation values if e.g. (rgb)==(4,0,0); this would definitely not be perceived as colorful by a human. The MaxChroma component will calculate them low.

### Prerequisites


JimBobSquarePants:
Thanks for redoing this @Franz1960 and sorry about the runaround. The Git docs tend to describe situations that should be prefaced with "In a vacuum...".

src/ImageSharp/Processing/Processors/Binarization/BinaryThresholdProcessor{TPixel}.cs
```diff
@@ -90,14 +97,55 @@ public void Invoke(int y)
...
+                        color.ToRgba32(ref rgba);
+
+                        // Extract saturation and compare to threshold.
+                        float sat = this.colorSpaceConverter.ToHsl(rgba).S;
+                        color = (sat >= fThreshold) ? this.upper : this.lower;
+                    }
+                }
+                else if (this.colorComponent == BinaryThresholdColorComponent.MaxChroma)
+                {
+                    float fThreshold = this.threshold / 2F;
```
JimBobSquarePants, line 128: Hey @Franz1960 Can you explain this bit to me please? I'm struggling to figure out why we're halving the threshold.

Franz1960: The achromatic level is in the center of the Cb and Cr range, so the distance is half the range.

JimBobSquarePants: Thanks... Makes sense.

One last thing. Can you please give me push permissions to your fork? I just did some optimization work on the process but do not have permissions to push for some reason (_Normally I do with others_).

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

Franz1960: The box had been checked:
image
but as it was not giving you the necessary rights (for whatever reason) I gave you write permission to the forked repo.

JimBobSquarePants (approved):

JimBobSquarePants:
@Franz1960 Thanks for your patience here. Still trying to figure out LFS as I had to manually add your new files to it. 

I made some changes to optimize the methods which should keep them snappy also if you're curious but I'm going to merge this now. It's awesome to have a new community feature!

Franz1960:
@JimBobSquarePants Thanks too for your patience, and @antonfirsov equally. I had to learn many things. Next time will hopefully go better. Replacing submodules by LFS was for sure the right thing to do.

antonfirsov:
@Franz1960 thanks for the great work and also the patience towards our processes (with the submodules they were way more complicated than they should be)!

One last thing: can you add your opinion on #1555? I'd like to rename it today.

PR #1568, opened 2021-03-01
Fix ordered dithering for small palette lengths.

JimBobSquarePants:
### Prerequisites

### Description

Triggered by this runtime feature request https://github.com/dotnet/runtime/issues/48878 I decided to test our output to compare results and I discovered our implementation to be buggy. This PR fixes the issues and updates the expected output. 

I also added a new `KnownDitherings.Bayer16x16` option since System.Drawing has that option and did some cleanup to remove some of our abuse of S.R.C.Unsafe APIs..

**Before**
immo - Copy

**After**
DitherFilter_WorksWithAllDitherers_immo_Bayer8x8


src/ImageSharp/Processing/Processors/Dithering/OrderedDither.cs
```diff
@@ -145,24 +144,25 @@ public OrderedDither(uint length)
                 in ditherOperation);
         }
 
-        [MethodImpl(InliningOptions.ShortMethod)]
+        // Spread assumes an even colorspace distribution and precision.
+        // Cubed root used because we always compare to Rgb.
+        // https://bisqwit.iki.fi/story/howto/dither/jy/
+        // https://en.wikipedia.org/wiki/Ordered_dithering#Algorithm
+        internal static int CalculatePaletteSpread(int colors) => (int)(255 / (Math.Pow(colors, 1.0 / 3) - 1));
```
saucecontrol, line 151: This will still give you bad results with some palettes.  It's not really a question of how many color channels are in the image but rather how many color channels are represented in the palette.  Consider the following 4-step greyscale palette

```
new[] { 0x000000u, 0x555555u, 0xAAAAAAu, 0xFFFFFFu };
```

Here the spread should be 85 (0x55), which is `255/(4 - 1)`.  The above calc will give you `255 / (cbrt(4) - 1)`, which comes out to 434 instead and will lead to overblown contrast again.  A black and white palette would give you a spread of 981 instead of the correct 255.

What you want to do here is build a check like `palette.IsGreyscale` and only take the cube root if that's false.  And just to be absolutely safe, I'd clamp the value of the denominator to a min of 1 (or clamp the result to a max of 255).  You could still be given an RGB palette that's less than 8 colors.

JimBobSquarePants: I have been wondering about this tbh. We currently have no mechanism in place to determine whether or not the palette is greyscale. 

We also use the same operations to handle automatically resolved palettes from quantization so whatever we did we would have to be able to provide that option there also.

src/ImageSharp/Processing/Processors/Dithering/OrderedDither.cs
```diff
@@ -23,7 +22,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Dithering
         /// Initializes a new instance of the <see cref="OrderedDither"/> struct.
         /// </summary>
         /// <param name="length">The length of the matrix sides</param>
-        [MethodImpl(InliningOptions.ShortMethod)]
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
antonfirsov, line 25: This method does not seem trivial. I would keep it as `ShortMethod` so it shows up in profiler results when asked for.

src/ImageSharp/Processing/Processors/Dithering/OrderedDither.cs
```diff
@@ -218,23 +218,24 @@ public override int GetHashCode()
                 this.source = source;
                 this.destination = destination;
                 this.bounds = bounds;
-                this.bitDepth = ColorNumerics.GetBitsNeededForColorDepth(destination.Palette.Length);
+                this.spread = CalculatePaletteSpread(destination.Palette.Length);
             }
 
-            [MethodImpl(InliningOptions.ShortMethod)]
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
antonfirsov, line 224: Same as above.

src/ImageSharp/Processing/Processors/Dithering/OrderedDither.cs
```diff
@@ -262,19 +263,23 @@ public void Invoke(int y)
                 this.source = source;
                 this.bounds = bounds;
                 this.scale = processor.DitherScale;
-                this.bitDepth = ColorNumerics.GetBitsNeededForColorDepth(processor.Palette.Length);
+                this.spread = CalculatePaletteSpread(processor.Palette.Length);
             }
 
-            [MethodImpl(InliningOptions.ShortMethod)]
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
antonfirsov, line 269: Same as above.

src/ImageSharp/Processing/Processors/Dithering/OrderedDither.cs
```diff
@@ -145,24 +144,25 @@ public OrderedDither(uint length)
...
             TPixel source,
             int x,
             int y,
-            int bitDepth,
+            int spread,
             float scale)
             where TPixel : unmanaged, IPixel<TPixel>
         {
-            Rgba32 rgba = default;
+            Unsafe.SkipInit(out Rgba32 rgba);
```
antonfirsov, line 162: 👍

antonfirsov (commented):
Some concerns about the removal of `InliningOptions.ShortMethod` in non-trivial methods, otherwise looks good.

src/ImageSharp/Processing/Processors/Dithering/OrderedDither.cs
JimBobSquarePants: Happy to revert

antonfirsov (approved):
LGTM

src/ImageSharp/Processing/Processors/Dithering/OrderedDither.cs
JimBobSquarePants: Left a TODO for now. But thinking about some sort of enhancement for `PixelTypeInfo`

Issue #808, opened 2019-01-12
Feature request: allow Color Space choices in JpegEncoder


antonfirsov:
This is a useful image size optimization for users dealing with grayscale images. (See #807) 

Additional colorspace encoders to implement:
- [x] Grayscale (actually important)
- [ ] Ycck (nice to have)
- [ ] Cmyk (nice to have)
- [x] RGB (nice to have)

brianpopow:
@antonfirsov with #1734 merged, encoding with RGB colorspace is now possible.

antonfirsov:
@brianpopow nice!  Only two more left.

JimBobSquarePants:
Fixed via #2120

PR #1586, opened 2021-03-28
Grayscale Jpeg encoding

ynse01:
### Prerequisites

### Description
Implemented Grayscale JPEG encoding, as requested in issue #808 (first bullet).

This means, that for L8 or L16 pixel types, by default the JpegEncoder creates Grayscale, 1-component jpegs. This can also be forced upon other pixel type, by setting the ColorType property in the JpegEncoderOptions.

Grayscale Jpeg encoding performs better and results in a smaller file size.

Extended the JpegEncoderTests to cover Grayscale encoding and L8 pixel types also.


src/ImageSharp/Formats/Jpeg/JpegSubsample.cs
```diff
@@ -8,6 +8,11 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
     /// </summary>
     public enum JpegSubsample
     {
+        /// <summary>
+        /// Only luminance - No chrome channels.
+        /// </summary>
+        Grayscale,
```
antonfirsov, line 14: What happens if someone tries to use a `JpegEncoder` with `Subsample` being incompatible with `ColorType`?

src/ImageSharp/Formats/Jpeg/Components/Encoder/L8ToYConverter.cs
```diff
@@ -0,0 +1,49 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
+{
+    /// <summary>
+    /// Provides 8-bit lookup tables for converting from L8 to Y colorspace.
+    /// </summary>
+    internal unsafe struct L8ToYConverter
```
antonfirsov, line 13: There are no lookup tables or any other information to work with during grayscale conversion.

All methods in this type can be just moved into `static` methods within `LuminanceForwardConverter<TPixel>` and the `L8ToYConverter` struct can be removed.

src/ImageSharp/Formats/Jpeg/JpegEncoder.cs
```diff
@@ -50,7 +56,21 @@ public Task EncodeAsync<TPixel>(Image<TPixel> image, Stream stream, Cancellation
...
         {
             var encoder = new JpegEncoderCore(this);
+            this.EnrichColorType<TPixel>();
             return encoder.EncodeAsync(image, stream, cancellationToken);
         }
+
+        /// <summary>
+        /// If ColorType was not set, set it based on the given <typeparamref name="TPixel"/>.
+        /// </summary>
+        private void EnrichColorType<TPixel>()
```
antonfirsov, line 66: `InitializeColorType` would be a better name for this method.

tests/ImageSharp.Tests/Formats/Jpg/JpegEncoderTests.cs
```diff
@@ -83,6 +83,12 @@ public void Encode_PreserveQuality(string imagePath, int quality)
         public void EncodeBaseline_WorksWithDifferentSizes<TPixel>(TestImageProvider<TPixel> provider, JpegSubsample subsample, int quality)
             where TPixel : unmanaged, IPixel<TPixel> => TestJpegEncoderCore(provider, subsample, quality);
 
+        [Theory]
+        [WithFile(TestImages.Png.BikeGrayscale, nameof(BitsPerPixel_Quality), PixelTypes.L8)]
+        [WithSolidFilledImages(nameof(BitsPerPixel_Quality), 1, 1, 100, 100, 100, 255, PixelTypes.L8)]
+        public void EncodeBaseline_GrayscaleWorksWithDifferentSizes<TPixel>(TestImageProvider<TPixel> provider, JpegSubsample subsample, int quality)
```
antonfirsov, line 89: We also need test cases for:
- `PixelTypes.Rgba32` (general case with explicit setting of  `colorType:JpegColorType.Luminance)`, the actual image can be grayscale, so the validation can be done easily
- A case with `PixelTypes.L8`, but  `colorType:null` to test the automatic detection of `ColorType` when not specified

brianpopow:
> Can anyone in @SixLabors/core check if the `[WithFile(TestImages.Png.BikeGrayscale, nameof(BitsPerPixel_Quality), PixelTypes.L8)]` case works for them and produces output locally? I have some troubles with ImageMagic on my PC.

I have tested it (on windows), it works fine, it produces a gray image of the bike testimage. @antonfirsov what problem do you see on your pc?

antonfirsov:
@brianpopow don't mind it, must be local to my environment (+ want the PR review to stay on topic). Will report back on Slack if it will keep bothering me 😄 Thanks for checking!

src/ImageSharp/Formats/Jpeg/JpegEncoder.cs
```diff
@@ -50,7 +56,21 @@ public Task EncodeAsync<TPixel>(Image<TPixel> image, Stream stream, Cancellation
...
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            if (this.ColorType == null)
+            {
+                bool isGrayscale = typeof(TPixel) == typeof(L8) || typeof(TPixel) == typeof(L16);
```
JimBobSquarePants, line 71: We should be storing the `JpegColorType` in the `JpegMetadata` class upon decode and querying both places with the encoder options taking precedence. 

`La16` and `La32` should trigger grayscale also.

src/ImageSharp/Formats/Jpeg/JpegSubsample.cs
ynse01: Subsample is only configured on the Chroma components in the current settings.

If in the future, there would be a Subsample for Luminance: Then ColorType takes precedence. See line 193 of JpegEncoderCore.cs

src/ImageSharp/Formats/Jpeg/JpegEncoder.cs
ynse01: Renamed it as you suggested

src/ImageSharp/Formats/Jpeg/Components/Encoder/L8ToYConverter.cs
ynse01: I modeled this to be similar to the existing YCbCr classes. But Grayscale conversion is much simpler, so moved it into the LuminanceForwardConverter class.

tests/ImageSharp.Tests/Formats/Jpg/JpegEncoderTests.cs
ynse01: All these test cases added. Renamed the new testcase, as now it is fully focussed on Grayscale.

src/ImageSharp/Formats/Jpeg/JpegEncoder.cs
ynse01: Added La16 and La32 to JpegEncoder and added test cases for both.

ynse01:
Thank you for the great feedback.

I have refactored my solution such, that the JpegSubsample.Grayscale option is no longer needed. I realized that it would have been hard to explain to users what it did. 
Adopted the JpegEncoderTests accordingly.

Please have another look.

src/ImageSharp/Formats/Jpeg/JpegEncoder.cs
antonfirsov: Unresolving the conversation, since I think @JimBobSquarePants also asked for a decoder extension here, will try to explain  what he likely meant in more detailed manner:
- Define an optional `JpegColorType?` property in `JpegMetadata`
- Detect `JpegColorType` in `JpegDecoderCore` and store it into the image's `JpegMetadata`.
- In the encoder, also trigger the grayscale path in case if  `JpegMetadata.JpegColorType` is set to `JpegColorType.GrayScale`

@ynse01 let us know if you feel comfortable and confident adding this. If not, I recommend to not block the PR on this,  and open a tracking issue for the addition.

ynse01: I gave it a try. Please indicate if this is what you guys expect.

JimBobSquarePants (approved):
This all looks fantastic. Thanks so much for taking your time to contribute, we really appreciate it!

PR #1591, opened 2021-04-06
Faster Linear Transforms

JimBobSquarePants:
### Prerequisites

### Description
I haven't been happy with the existing linear transform code since I originally wrote it so thought I'd have another crack. 

Benchmarks are very good. Large speedup on all targets with a 2x speedup on .NET Core 3.1+. However they don't tell the whole story. Originally two 2D buffers were rented from the pool of _DestinationHeight x MaxDiameter_ to store kernel weight buffers but, since the kernel weights must be calculated based on the exact transformed subpixel point, it was pointless to do so. Those buffers are now gone.

In addition I discovered a rounding issue where the incorrect scaled filter radius was being calculated which lead to rogue pixels containing RGB values but 0 alpha to be generated outside transform edges. I've updated two references to reflect the improved accuracy. 

I consider readability much improved now also. It's much easier to understand how the transform operation actually works. 

Ignore individual commits during review, I generated a lot of noise during my experimentation (_basically properly learning how the process actually worked_ 😇)  so I'll be squashing to a single commit on merge.

### Benchmarks

```
.NET Core SDK=5.0.201
  [Host]     : .NET Core 5.0.4 (CoreCLR 5.0.421.11614, CoreFX 5.0.421.11614), X64 RyuJIT
  Job-WMQSXI : .NET Framework 4.8 (4.8.4341.0), X64 RyuJIT
  Job-ZPQJBS : .NET Core 2.1.26 (CoreCLR 4.6.29812.02, CoreFX 4.6.29812.01), X64 RyuJIT
  Job-LHSZUP : .NET Core 3.1.13 (CoreCLR 4.700.21.11102, CoreFX 4.700.21.11602), X64 RyuJIT
```

### ROTATE

**BEFORE**
|   Method |        Job |       Runtime |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |----------- |-------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| DoRotate | Job-OBWWBY |    .NET 4.7.2 | 44.56 ms | 0.464 ms | 0.411 ms |     - |     - |     - |  11.33 KB |
| DoRotate | Job-MMNLYB | .NET Core 2.1 | 19.18 ms | 0.320 ms | 0.355 ms |     - |     - |     - |   5.91 KB |
| DoRotate | Job-UKOETF | .NET Core 3.1 | 24.90 ms | 0.328 ms | 0.274 ms |     - |     - |     - |   7.23 KB |

**AFTER**
|   Method |        Job |       Runtime |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |----------- |-------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| DoRotate | Job-BAUEPW |    .NET 4.7.2 | 30.73 ms | 0.397 ms | 0.331 ms |     - |     - |     - |   6.75 KB |
| DoRotate | Job-SNWMCN | .NET Core 2.1 | 16.31 ms | 0.317 ms | 0.352 ms |     - |     - |     - |   5.25 KB |
| DoRotate | Job-MRMBJZ | .NET Core 3.1 | 12.21 ms | 0.239 ms | 0.245 ms |     - |     - |     - |   6.61 KB |

### SKEW

**BEFORE**
| Method |        Job |       Runtime |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |----------- |-------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| DoSkew | Job-GBTORI |    .NET 4.7.2 | 37.08 ms | 0.463 ms | 0.433 ms |     - |     - |     - |  10.29 KB |
| DoSkew | Job-RAEJFO | .NET Core 2.1 | 17.27 ms | 0.187 ms | 0.175 ms |     - |     - |     - |   5.85 KB |
| DoSkew | Job-LFXFDP | .NET Core 3.1 | 19.34 ms | 0.149 ms | 0.140 ms |     - |     - |     - |   7.17 KB |

**AFTER**
| Method |        Job |       Runtime |      Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |----------- |-------------- |----------:|----------:|----------:|------:|------:|------:|----------:|
| DoSkew | Job-YEGFRQ |    .NET 4.7.2 | 23.563 ms | 0.0731 ms | 0.0570 ms |     - |     - |     - |   6.75 KB |
| DoSkew | Job-HZHOGR | .NET Core 2.1 | 13.700 ms | 0.2727 ms | 0.5122 ms |     - |     - |     - |   5.25 KB |
| DoSkew | Job-LTEUKY | .NET Core 3.1 |  9.971 ms | 0.0254 ms | 0.0225 ms |     - |     - |     - |   6.61 KB |


src/ImageSharp/Processing/Processors/Transforms/Linear/LinearTransformUtils.cs
```diff
@@ -1,104 +0,0 @@
...
-    /// <summary>
-    /// Utility methods for affine and projective transforms.
-    /// </summary>
-    internal static class LinearTransformUtils
-    {
-        [MethodImpl(InliningOptions.ShortMethod)]
-        internal static int GetSamplingRadius<TResampler>(in TResampler sampler, int sourceSize, int destinationSize)
-            where TResampler : struct, IResampler
-        {
-            double scale = sourceSize / destinationSize;
```
JimBobSquarePants, line 21: Spot the rounding error 😄

JimBobSquarePants:
@antonfirsov I'm gonna need some help here. MacOS .NET 3.1 was showing a 4% error on projective transforms but I couldn't pull down the test output because it was incomplete. (DebugSave disabled on CI). I enabled saving and now I'm getting a bunch of MemoyAllocator errors. Any idea?

antonfirsov:
`DebugSave` uses `image.GetConfiguration().MemoryAllocator` which is artificially limited to a very low number to test discontinuous decoding:

https://github.com/SixLabors/ImageSharp/blob/5ab768c7fff09350487dc09e35e699ce5a14b6c8/tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs#L414-L417

I can't think of any better solution than changing the following line to use `Configuration.Default` instead of the `image`'s configuration (with a comment on why are we doing so). I hope it won't lead to unwanted sideffects:

https://github.com/SixLabors/ImageSharp/blob/a1784a67bf492b1b02e99655a39aa5e5acf81988/tests/ImageSharp.Tests/TestUtilities/ImagingTestCaseUtility.cs#L181

JimBobSquarePants:
@antonfirsov Thanks for you help! I've just created a clone of the encoder and use that when registering for tests.

JimBobSquarePants:
OK... tried everything I can think of here. Rewrote all the sampling to use (since reverted) to use double precision and it still fails for NET Core 3.1 on Mac OS.

src/ImageSharp/Processing/Processors/Transforms/Linear/AffineTransformProcessor{TPixel}.cs
```diff
@@ -129,100 +110,119 @@ public void ApplyTransform<TResampler>(in TResampler sampler)
...
+                        for (int yK = top; yK <= bottom; yK++)
+                        {
+                            float yWeight = sampler.GetValue(yK - pY);
+
+                            for (int xK = left; xK <= right; xK++)
+                            {
+                                float xWeight = sampler.GetValue(xK - pX);
+
+                                Vector4 current = sourceBuffer.GetElementUnsafe(xK, yK).ToScaledVector4();
+                                Numerics.Premultiply(ref current);
```
antonfirsov, line 211: Idea 3:
Also log the premultiplied `current` value, compare the log files, look for highest difference.

src/ImageSharp/Processing/Processors/Transforms/Linear/AffineTransformProcessor{TPixel}.cs
```diff
@@ -129,100 +110,119 @@ public void ApplyTransform<TResampler>(in TResampler sampler)
...
-                for (int x = 0; x < this.maxX; x++)
-                {
-                    // Use the single precision position to calculate correct bounding pixels
-                    // otherwise we get rogue pixels outside of the bounds.
-                    var point = Vector2.Transform(new Vector2(x, y), this.matrix);
-                    LinearTransformUtils.Convolve(
-                        in this.sampler,
-                        point,
-                        sourceBuffer,
+                    Numerics.UnPremultiply(span);
```
antonfirsov, line 219: Have we used the bulk `UnPremultiply` in the old code?

antonfirsov (approved):
The new transform code looks good and easy to follow, great job!

Can you post some Beyond Compare screenshots for the most problematic, and the less problematic failures? 

For troubleshooting I would log and compare as many internals as possible. Since it may take a lot of time, we can just disable the Mac tests with `[ActiveIssue]` instead of trying to figure out the problem as part of the PR.

src/ImageSharp/Processing/Processors/Transforms/Linear/AffineTransformProcessor{TPixel}.cs
JimBobSquarePants: We didn't. Have just ran a test commit using the non-bulk variant and the major issues disappeared leaving only 2 minor (_max 0.03% diff_) remaining. 

Below is the output of the worst issue. The last 3 pixels of each row appear to be affected which suggests to me that something is going wrong here. My guess is that it has something to do with unpremultiplying Vector4.Zero 

https://github.com/SixLabors/ImageSharp/blob/bb727699686cba9cf55b67c5b114bb45311e6f3f/src/ImageSharp/Common/Helpers/Numerics.cs#L556-L576

image

The question is, what do we do? The issue is clearly fixed for .NET 5 but .NET Core 3.1 is LTS.

antonfirsov: I think we should still keep the bulk code for the rest of the platforms, because it likely takes an important part in the perf improvements we see in your benchmarks. I would add a check for the OS platform at this line with an explanatory comment:

https://github.com/SixLabors/ImageSharp/blob/bb727699686cba9cf55b67c5b114bb45311e6f3f/src/ImageSharp/Common/Helpers/Numerics.cs#L543

```C#
if (Avx2.IsSupported && RuntimeInformation.IsOSPlatform(OSPlatform.OSX) && vectors.Length >= 2)
```

In no case I would complicate this with additional logic to detect if we are running on .NET 5 (Mac is not *that* important).

JimBobSquarePants:
I’m done with this. I give up. I can’t replicate it on my Mac and I’m absolutely fed up with guessing what the cause could be.

JimBobSquarePants:
You better re-review this @antonfirsov there's been some awful hackery.

src/ImageSharp/Common/Helpers/RuntimeEnvironment.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether the .NET installation is .NET Core
+        /// </summary>
+        public static bool IsNetCore => IsNetCoreLazy.Value;
+
+        /// <summary>
+        /// Gets the name of the .NET installation on which an app is running.
+        /// </summary>
+        public static string FrameworkDescription => RuntimeInformation.FrameworkDescription;
```
antonfirsov, line 25: Does this have to be exposed? I would make it private.

src/ImageSharp/Common/Helpers/RuntimeEnvironment.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// Many methods defer to <see cref="RuntimeInformation"/> when available.
+    /// </summary>
+    internal static class RuntimeEnvironment
+    {
+        private static readonly Lazy<bool> IsNetCoreLazy = new Lazy<bool>(() => FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase));
+
+        /// <summary>
+        /// Gets a value indicating whether the .NET installation is .NET Core
+        /// </summary>
+        public static bool IsNetCore => IsNetCoreLazy.Value;
```
antonfirsov, line 20: I guess this definition of ".NET Core" excludes .NET 5. We need to be more careful here, the terminology already became ambiguous. (Same product rebranded TBH ...)

```suggestion
        /// <summary>
        /// Gets a value indicating whether the .NET installation is .NET Core 3.1 or lower.
        /// </summary>
        public static bool IsNetCoreClassic => IsNetCoreLazy.Value;
```

antonfirsov (commented):
I'm wondering why did you decide to duplicate? Wasn't there a mistake in 35065c7f479fed1411fed2eff66fee692e5f1fc1 adding the new conditions checks to `Premultiply` instead of `Unpremultiply`?

If there is a perf concern regarding the long expression, I recommend to cache `!RuntimeEnvironment.IsOSPlatform(OSPlatform.OSX) && RuntimeEnvironment.IsNetCore` into a `static` variable.

src/ImageSharp/Common/Helpers/RuntimeEnvironment.cs
JimBobSquarePants: I did make a mistake yeah but corrected it in a later commit. 

What I've tried so far:

- Filtered out the runtime instrinsics part of the `UnPremultiply` method to aviod Mac on Core 3.1
- Rewrote the scalar overhang portion of the method to use a simple for loop over the span.

Neither worked. The only thing I could get working was duplicating the code which is simply awful. I couldn't replicate the issue locally either to effectively debug.

Regarding the method, no it doesn't need to be exposed.

src/ImageSharp/Common/Helpers/RuntimeEnvironment.cs
JimBobSquarePants: I would probably call other querying methods. 

- `IsNet`
- `IsNetFramework`

and flesh out the intellisense docs. .NET 5 returns `.NET` as the prefix, while < .NET 4.8 return `.NET Framework`. 

It's a crappy solution anyway. .NET Core was and always shall be a terrible name.

antonfirsov: At least let's change the comment for now to avoid shooting ourselves in the feet!

```suggestion
        /// <summary>
        /// Gets a value indicating whether the .NET installation is .NET Core 3.1 or lower.
        /// </summary>
        public static bool IsNetCore => IsNetCoreLazy.Value;
```

antonfirsov (approved):
If it doesn't work without the duplication, let's keep it. LGTM with one suggestion.

PR #1600, opened 2021-04-15
Remove mandatory naming prefix and suffix from .editorconfig

saucecontrol:
### Prerequisites

### Description

I'm getting some wacky errors trying to build in VS 16.10p2, including this nonsensical fix suggestion

image

Which apparently comes from these .editorconfig lines.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -261,8 +261,8 @@ dotnet_naming_style.prefix_type_parameters_with_t_style.capitalization = pascal_
 dotnet_naming_style.prefix_type_parameters_with_t_style.required_prefix = T
 # disallowed_style - Anything that has this style applied is marked as disallowed
 dotnet_naming_style.disallowed_style.capitalization  = pascal_case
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Ah, this change actually has to happen upstream. We [copy](https://github.com/SixLabors/SharedInfrastructure/blob/9b1179f0ebe6a4dfed998252b860fa07fee54363/msbuild/targets/SixLabors.Src.targets#L8-L18) the editorconfig on build from the submodule [here](https://github.com/SixLabors/SharedInfrastructure/blob/master/.editorconfig).

JimBobSquarePants:
I need to do a comparison with the latest version here. They're removed that stuff now and added some other updates.

https://github.com/RehanSaeed/EditorConfig/blob/main/.editorconfig

saucecontrol:
Ah, cool.  Will close then if you're looking at the upstream fix :)

PR #1606, opened 2021-04-16
Add InnerLoop Configuration Profiles for Faster Prototyping

JimBobSquarePants:
### Prerequisites

### Description
Adds two new solution configuration profiles that target on a single target framework `netcoreapp3.1` in order to allow faster development prototyping due to reduce build times. See #1603
- Debug-InnerLoop
- Release-InnerLoop

In addition MinVer and SourceLink have been configured upstream via our Shared Infrastructure submodule to only be trigger when running in CI.

I've also updated our editorconfig profile to work better with VS 16.8+ #1600. There's a lot more analyzers running in VS now so we get more warnings but nothing critical.

NOTE: Visual Studio is a little flaky when you switch between the profiles (fixed by a restart) so I suggest picking a profile and sticking with it for the most part.


antonfirsov:
Want to try this out with both VS and Rider this week, and come back.

src/ImageSharp/ImageSharp.csproj
```diff
@@ -20,6 +21,11 @@
         <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp2.1;netstandard2.1;netstandard2.0;netstandard1.3;net472</TargetFrameworks>
       </PropertyGroup>
     </When>
+    <When Condition="$(Configuration.EndsWith('InnerLoop')) == true">
+      <PropertyGroup>
+        <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
```
antonfirsov, line 26: How about this?
https://github.com/adamralph/minver#can-i-disable-minver

```suggestion
        <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
        <MinVerSkip>true</MinVerSkip>
```

antonfirsov (approved):
Seems to work nicely for me, great change, thanks!

There is an easy way to disable MinVer without touching shared infra.

src/ImageSharp/ImageSharp.csproj
JimBobSquarePants: Didn't know that was possible! I'll leave the changes in the submodule for now though as I don't really see a use for MinVer outside of CI.

PR #1610, opened 2021-04-18
Added overloads to `Image.WrapMemory` for `IMemoryOwner<byte>`.

ptasev:
### Prerequisites

### Description
Adds an overload as discussed here:
https://github.com/SixLabors/ImageSharp/discussions/1580


src/ImageSharp/Image.WrapMemory.cs
```diff
@@ -303,6 +303,82 @@ public abstract partial class Image
...
+            int height,
+            ImageMetadata metadata)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            Guard.NotNull(configuration, nameof(configuration));
+            Guard.NotNull(metadata, nameof(metadata));
+
+            var pixelMemoryOwner = new ByteMemoryOwner<TPixel>(byteMemoryOwner);
+
+            Guard.IsTrue(pixelMemoryOwner.Memory.Length >= width * height, nameof(pixelMemoryOwner), "The length of the input memory is less than the specified image size");
```
Sergio0694, line 335: This could technically overflow (`width * height`), which would cause the test to pass and then something else to break later down the line. I'm seeing that we're doing the same in other existing APIs though, so... Shrugs? 🤷‍♂️

src/ImageSharp/Memory/ByteMemoryOwner{T}.cs
```diff
@@ -0,0 +1,55 @@
...
+                this.disposedValue = true;
+            }
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+            this.Dispose(disposing: true);
+            GC.SuppressFinalize(this);
```
Sergio0694, line 52: Why is this needed? This type doesn't have a finalizer 🤔

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
```diff
@@ -413,6 +413,32 @@ public void WrapMemory_IMemoryOwnerOfT_ValidSize(int size, int height, int width
...
+            [InlineData(26, 5, 5)]
+            [InlineData(2, 1, 1)]
+            [InlineData(1024, 32, 32)]
+            [InlineData(2048, 32, 32)]
+            public void WrapMemory_IMemoryOwnerOfByte_ValidSize(int size, int height, int width)
+            {
+                var array = new byte[size * Unsafe.SizeOf<Rgba32>()];
+                var memory = new TestMemoryOwner<byte> { Memory = array };
+
+                Image.WrapMemory<Rgba32>(memory, height, width);
```
Sergio0694, line 439: We should probably add some additional test to actually validate the image size after creation, and also grabbing a few image row spans and checking that references to their items actually match the correct locations in the initial wrapped arrays, just to be extra safe here 🙂

src/ImageSharp/Memory/ByteMemoryOwner{T}.cs
```diff
@@ -0,0 +1,55 @@
...
+            this.memoryOwner = memoryOwner;
+            this.memoryManager = new ByteMemoryManager<T>(memoryOwner.Memory);
+        }
+
+        /// <inheritdoc/>
+        public Memory<T> Memory => this.memoryManager.Memory;
+
+        private void Dispose(bool disposing)
+        {
+            if (!this.disposedValue)
```
Sergio0694, line 36: This is not thread safe, is this an issue?
We might want to replace this with `Interlocked.CompareExchange` and make the flag an `int` to enable this, but I'm not sure if the rest of the library actually gives strong thread-safety guarantees on other APIs anyway, so not entirely sure whether this is needed. I mean, wouldn't hurt either though, I guess 🤔
@JimBobSquarePants thoughts?

Sergio0694 (changes requested):
Looks nice!
Left a couple review notes 🙂

src/ImageSharp/Image.WrapMemory.cs
ptasev: I added a cast to `long` in this case. Not sure if I should touch the others in this file that I didn't add.

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
```diff
@@ -413,6 +413,51 @@ public void WrapMemory_IMemoryOwnerOfT_ValidSize(int size, int height, int width
...
+                        var expected = (byte)random.Next(0, 256);
+
+                        Span<Rgba32> rowSpan = img.GetPixelRowSpan(i);
+                        array[arrayIndex] = expected;
+
+                        Assert.Equal(expected, rowSpan[0].R);
+                    }
+                }
+
+                Assert.True(memory.Disposed);
```
antonfirsov, line 458: Mega 👍 👍  for adding this check. [Whoever takes an `IMemoryOwner<T>` as a parameter, takes ownership -> should dispose it](https://gist.github.com/GrabYourPitchforks/4c3e1935fd4d9fa2831dbfcab35dffc6#rule-8-if-you-have-an-imemoryownert-parameter-in-your-api-surface-you-are-accepting-ownership-of-that-instance).

I wish we added the same check for the `IMemoryOwner<TPixel>` test.

antonfirsov (approved):
Looks good with the latest fixes, thanks!

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
```diff
@@ -413,6 +413,51 @@ public void WrapMemory_IMemoryOwnerOfT_ValidSize(int size, int height, int width
...
+
+                    for (int i = 0; i < height; ++i)
+                    {
+                        var arrayIndex = pixelSize * width * i;
+                        var expected = (byte)random.Next(0, 256);
+
+                        Span<Rgba32> rowSpan = img.GetPixelRowSpan(i);
+                        array[arrayIndex] = expected;
+
+                        Assert.Equal(expected, rowSpan[0].R);
```
Sergio0694, line 454: Nit: it'd be more correct to just compare the reference directly (it'd also avoid possible false positives).
Something like:
```suggestion
                        Span<Rgba32> rowSpan = img.GetPixelRowSpan(i);
                        ref Rgba32 r0 = ref rowSpan[0];
                        ref Rgba32 r1 = ref Unsafe.As<byte, Rgba32>(ref array[arrayIndex]);
                        
                        Assert.IsTrue(Unsafe.AreSame(ref r0, ref r1));
```

You'd also need to remove that `var random = new Random()` at line 436 🙂

Sergio0694 (changes requested):
Just one last minor nitpick, looks good to me otherwise! 🚀

tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
ptasev: Thanks! I started down this path, but couldn't figure out how to compare the memory locations of the ref vars (`Unsafe.AreSame`).

Sergio0694: Ahah glad I could help and thanks for making the change!
And yeah, the `Unsafe` class has a lot of great and forbidden stuff, it's lots of fun 😈

Sergio0694 (approved):
:shipit:

JimBobSquarePants:
Great work everyone 😄

Issue #87, opened 2017-01-20
Feature Request: autolevel (like ImageMagick)


aviflax:
ImageMagick has a very useful “autolevel” feature that my company has used to good effect. It would be very cool if ImageSharp had such a feature.

Thank you!

dlemstra:
I am actually one of the developers for the ImageMagick project and just looked through our code to see how it is done. It should not be very difficult to add this to ImageSharp.

aviflax:
That's great! Thank you!

JimBobSquarePants:
I've just realized we've had this for nearly 4 years now. Closing!

PR #1619, opened 2021-04-30
Add AutoLevel Processor

CoenraadS:
Closes https://github.com/SixLabors/ImageSharp/issues/87

### Prerequisites

### Description
Add AutoLevel Normalization (only works for L8, L16 types)

Test Image generated from following command (ImageMagick)

`cd ImageSharp\tests\Images\Input\Jpg\baseline`
`magick 640px-Unequalized_Hawkes_Bay_NZ.jpg -auto-level 640px-Unequalized_Hawkes_Bay_NZ.png`


src/ImageSharp/Processing/Processors/Normalization/AutoLevelHistogramProcessor{TPixel}.cs
```diff
@@ -0,0 +1,188 @@
...
+        /// <param name="configuration">The configuration which allows altering default behaviour or extending the library.</param>
+        /// <param name="source">The source <see cref="Image{TPixel}"/> for the current processor instance.</param>
+        /// <param name="sourceRectangle">The source area to process for the current processor instance.</param>
+        public AutoLevelHistogramProcessor(
+            Configuration configuration,
+            Image<TPixel> source,
+            Rectangle sourceRectangle)
+            : base(configuration, source, sourceRectangle)
+        {
+            // TODO I don't know how to get channel bit depth for non-grayscale types
```
CoenraadS, line 39: I didn't know what to do here

src/ImageSharp/Processing/Processors/Normalization/AutoLevelProcessor{TPixel}.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// <param name="source">The source <see cref="Image{TPixel}"/> for the current processor instance.</param>
+        /// <param name="sourceRectangle">The source area to process for the current processor instance.</param>
+        public AutoLevelProcessor(
+            Configuration configuration,
+            Image<TPixel> source,
+            Rectangle sourceRectangle)
+            : base(configuration, source, sourceRectangle)
+        {
+            // TODO I don't know how to get channel bit depth for non-grayscale types
+            if (!(typeof(TPixel) == typeof(L16) || typeof(TPixel) == typeof(L8)))
```
JimBobSquarePants, line 36: I think it would be ok to take the `PixelType.BitsPerPixel` and divide it by `Unsafe.SizeOf<T>()`. That will give you a per-pixel-component average (_since all our pixel formats are blittable_) which should be close enough.

CoenraadS: If I would try that with `L16`

`Unsafe.SizeOf<L16>()` would return `2`
`BitsPerPixel` would return `16`

It's half of what it needs to be?

Or did you mean it's an operation only for non-luminescence types?

JimBobSquarePants: Ah right yep, my bad We need a `ChannelCount` property on `PixelTypeInfo`. 

@antonfirsov This would be a breaking change (_which we can probably allow for 1.1 release_) but do you have any ideas how we could introduce it in a non breaking way?

JimBobSquarePants: A hack you could do just now is something similar to what we do in the png encoder, I.E use a switch for known types and a default for unknown.
 
 https://github.com/SixLabors/ImageSharp/blob/a8cb71158493096121439d937b7e79aa5ea2727a/src/ImageSharp/Formats/Png/PngEncoderOptionsHelpers.cs#L159-L174

CoenraadS:
Looking through the ImageMagick implementation, they also support autolevel on the color channels (I thought it would just create a grayscale image but its not), but I must admit I'm not to sure how to implement that, nor do I have a use for it. Would full color support be required for this to be merged?

JimBobSquarePants:
What's the science behind working for all color channels? Might it be a case of using the resultant normalized luminance as some sort of multiplier against each pixel? `(luminance/bitdepth/bitdepth) * pixel`.

I'm not convinced greyscale only is an adequate implementation for the base library (_I dearly wish our existing Histogram equalization operations supported full color_) simply because every use case for automated equalization I've ever seen is in the application against full color photos in an album.

brianpopow:
> What's the science behind working for all color channels? Might it be a case of using the resultant normalized luminance as some sort of multiplier against each pixel? `(luminance/bitdepth/bitdepth) * pixel`.
> 
> I'm not convinced greyscale only is an adequate implementation for the base library (_I dearly wish our existing Histogram equalization operations supported full color_) simply because every use case for automated equalization I've ever seen is in the application against full color photos in an album.

Im not convinced that applying the histogram equalization to each individual color channel of an RGB image makes much sense.
OpenCV for example does not support it for the adaptive histogram equalization either [CLAHE](https://docs.opencv.org/4.5.2/d6/db6/classcv_1_1CLAHE.html).

One approach which sounds reasonable to me is converting the image from the RGB color space to the LAB color space and then apply the histogram equalization to the L channel. Then convert it back to RGB.
I had once a Proof of concept done for this when i was working on CLAHE for ImageSharp, but i was not fully convinced with the results and its like 2 years ago (cant find the source code anymore).

In my opinion applying the equalization to all color channels should not be the scope of this PR. Maybe its worth looking into how ImageMagic does it, but that could be something for a future PR.

Issue #1350, opened 2020-09-14
Quantization: reduce memory footprint without compromising speed and quality


antonfirsov:
All our current quantizers rely on a very heavy cache in order to produce acceptable quality:
https://github.com/SixLabors/ImageSharp/blob/b06cb32b7114961fd5473f7645d38f8fee04ec64/src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap%7BTPixel%7D.cs#L21

With large images this could consume up to several megabytes of memory every time a larger GIF (or palettized PNG) gets saved. We need to reduce memory usage without compromising speed and quality.

We made a couple of experiments and exchanged many ideas a couple of months ago, but unfortunately all those discussions got lost in the noise of the gitter chatroom. I suggest to recollect and rediscuss those ideas.

Improvements we may consider (according to what I remember from the gitter chat):
1. Improving the RGB Octree implementation to make it very fast by default so we hopefully don't need any cache. Most important: flattening the tree nodes into array(s) of structs (instead of heap objects).
2. Extend the octree to RGBA
3. Replace the dictionary with [this cache](https://github.com/SixLabors/ImageSharp/blob/7e451ea0acbb19e1ab742a869968480daa0fbf78/tests/ImageSharp.Benchmarks/Quantization/ColorLookupConcepts.cs#L463-L613). It produced promising results in my experiments.
4. New idea from @JimBobSquarePants: replace the dictionary with an LRU cache implementation like [the one in BitFaster.Caching](https://github.com/bitfaster/BitFaster.Caching/blob/main/BitFaster.Caching/Lru/TemplateConcurrentLru.cs).

@saucecontrol since I remember you had really valuable ideas here, I would really appreciate your feedback on all 4 options, and apologies if I'm making you to repeat yourself because of my terrible memory :smile: If I remember correctly, you also had some promising results in your repos.

Bravely assigning milestone 1.1 for now.

saucecontrol:
> apologies if I'm making you to repeat yourself because of my terrible memory

I'm afraid I don't recall much detail of those conversations myself 😄 

[...]

PR #1623, opened 2021-05-05
Add support for encoding 1 bit per pixel bitmaps

brianpopow:
### Prerequisites

### Description

Add support for encoding 1 bit per pixel bitmaps.

@JimBobSquarePants: The bitmap encoder uses Octree Quantizer as a default. This quantizer does seem to have problems with just 2 colors. Im not so familiar with the Octree Quantizer. Is not possible to quantize with just to colors or is it maybe an issue we should look into? All indices seem to be 0.
If i change it to Wu quantizer it works with two colors.

### DIFF
--- a/src/ImageSharp/Formats/Bmp/BmpBitsPerPixel.cs
+++ b/src/ImageSharp/Formats/Bmp/BmpBitsPerPixel.cs
@@ -9,6 +9,11 @@ namespace SixLabors.ImageSharp.Formats.Bmp
     public enum BmpBitsPerPixel : short
     {
         /// <summary>
+        /// 1 bit per pixel.
+        /// </summary>
+        Pixel1 = 1,
+
+        /// <summary>
         /// 4 bits per pixel.
         /// </summary>
         Pixel4 = 4,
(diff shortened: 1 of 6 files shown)


JimBobSquarePants:
>This quantizer does seem to have problems with just 2 colors. Im not so familiar with the Octree Quantizer. Is not possible to quantize with just to colors or is it maybe an issue we should look into? All indices seem to be 0.

@brianpopow Yep there's issues with how the Octree Quantizer currently distributes and reduces colors. (In addition to rampant memory usage issues)

It does look like we should be able to do a fairly quick port based on the code here to a version that should allow accurate reduction.
https://github.com/SixLabors/ImageSharp/issues/1350#issuecomment-817367334

It's something I keep meaning to take the time to attempt.

brianpopow:
> @brianpopow Yep there's issues with how the Octree Quantizer currently distributes and reduces colors. (In addition to rampant memory usage issues)
> 

@JimBobSquarePants: Should I switch the default quantizer then to the Wu-Quantizer?

JimBobSquarePants (approved):
Awesome work. I can't believe we have all the bit depth support now! 😄

rickbrew:
> It does look like we should be able to do a fairly quick port based on the code here to a version that should allow accurate reduction.
> [#1350 (comment)](https://github.com/SixLabors/ImageSharp/issues/1350#issuecomment-817367334)

I can confirm that my new Octree code correctly quantizes down to 1-bit depth. Paint.NET uses this in production today with the latest 4.2.16 release.

JimBobSquarePants:
@rickbrew Awesome, that's great to hear, thanks!

PR #1629, opened 2021-05-14
Fixed Issue#1628

### LINKED ISSUE #1628: Image & underlying ImageFrameCollection aren't behaving like they were disposed after calling Dispose()
### Prerequisites

### Description
Image<TPixel>& Image underlying frames aren't disposed correctly, can be accessed **after** parent image disposal and **can be mutated** like it's a valid instance.

Frames access properties do not ensure that they are not disposed:
https://github.com/SixLabors/ImageSharp/blob/993f96eb7d90e05dbf2039c116d44be8fb081cd8/src/ImageSharp/Image%7BTPixel%7D.cs#L144
https://github.com/SixLabors/ImageSharp/blob/993f96eb7d90e05dbf2039c116d44be8fb081cd8/src/ImageSharp/Image%7BTPixel%7D.cs#L149

Dispose method does not nullify `Frames` property:
https://github.com/SixLabors/ImageSharp/blob/993f96eb7d90e05dbf2039c116d44be8fb081cd8/src/ImageSharp/Image%7BTPixel%7D.cs#L258-L271

Nearly all ImageFrameCollection methods affected leading to odd situations:

```csharp
var image = Image.Load<ImageSharp.PixelFormats.Rgba32>(...);
var someOtherFrame = image.Frames.RootFrame.Clone();
image.Dispose();

// accepting root frame
var rootFromImage = image.GetRootFramePixelBuffer(); // throws ObjectDisposed as expected
var rootFromFrames = image.Frames.RootFrame; // returns null

// mutation still works
image.Frames.AddFrame(someOtherFrame);

// and this mutation can lead to image returning spans like it was never disposed
var span = image.GetPixelRowSpan(0);
```

Last line works kinda like expected because everything else is broken

*All above also works for base Image class

### Steps to Reproduce
Use this code snippet for any image:
```csharp
var image = Image.Load<ImageSharp.PixelFormats.Rgba32>("path");
var someOtherFrame = image.Frames.RootFrame.Clone();
image.Dispose();

var rootFromFrames = image.Frames.RootFrame;

image.Frames.AddFrame(someOtherFrame);
var span = image.GetPixelRowSpan(0);
```

br3aker:
### Prerequisites

- [x] This PR does not break existing public API in any way

### Description
Fixes #1628
Some public properties can't break anything (e.g image size or metadata), they didn't throw before this PR, this behaviour wasn't altered. `ToString` weren't affected either.

Due to slight internal (within `Image` class) API change pixel indexer might be a bit more performant: `EnsureNotDisposed()` call is no longer virtual. Although JIT in some runtimes even could inline such virtual calls in previous versions, it's still a nice change of having one less virtual method I guess.
This change also affects any public method but their calls are too rare to affect performance.


JimBobSquarePants:
I'll get this reviewed ASAP. Thanks for the contribution!

src/ImageSharp/Image.cs
```diff
@@ -140,15 +148,20 @@ public Image<TPixel2> CloneAs<TPixel2>()
...
 
         /// <summary>
         /// Throws <see cref="ObjectDisposedException"/> if the image is disposed.
         /// </summary>
-        internal abstract void EnsureNotDisposed();
+        internal void EnsureNotDisposed()
+        {
+            if (this.isDisposed)
+            {
+                throw new ObjectDisposedException("Trying to execute an operation on a disposed image.");
```
JimBobSquarePants, line 162: We should be using a ThrowHelper here to allow inlining on sealed instances.

src/ImageSharp/ImageFrameCollection.cs
```diff
@@ -11,8 +11,10 @@ namespace SixLabors.ImageSharp
     /// Encapsulates a pixel-agnostic collection of <see cref="ImageFrame"/> instances
     /// that make up an <see cref="Image"/>.
     /// </summary>
-    public abstract class ImageFrameCollection : IEnumerable<ImageFrame>
+    public abstract class ImageFrameCollection : IDisposable, IEnumerable<ImageFrame>
```
JimBobSquarePants, line 14: Wow. I hadn't realized this wasn't already implementing the type. Good catch!

src/ImageSharp/Image.cs
```diff
@@ -80,8 +82,14 @@ protected Image(Configuration configuration, PixelTypeInfo pixelType, ImageMetad
         /// <inheritdoc />
         public void Dispose()
         {
-            this.Dispose(true);
-            GC.SuppressFinalize(this);
```
JimBobSquarePants, line 84: We shouldn't be removing the finalizer suppression in the base class. 

https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#the-dispose-method

src/ImageSharp/Image{TPixel}.cs
```diff
@@ -245,39 +258,17 @@ public override Image<TPixel2> CloneAs<TPixel2>(Configuration configuration)
...
-
-        /// <inheritdoc/>
-        internal override void EnsureNotDisposed()
-        {
-            if (this.isDisposed)
-            {
-                throw new ObjectDisposedException("Trying to execute an operation on a disposed image.");
-            }
-        }
+        protected override void DisposeManaged() => this.frames.Dispose();
```
JimBobSquarePants, line 271: The original pattern was correct here.

JimBobSquarePants (changes requested):
Definitely the right direction but we need to fix up the Disposal pattern changes to match convention and prevent breaking the API.

src/ImageSharp/Image.cs
br3aker: You mean ThrowHelper class from ThrowHelper.cs? Should I add ThrowArgumentException method or am I missing something? (there is none as far as I can see :P)

src/ImageSharp/Image.cs
br3aker: Didn't know `SuppressFinalize` do nothing if type has no finalizer, thanks!

br3aker:
@JimBobSquarePants seems like I've messed up with ThrowHelper from shared infrastructure. Although I'd added it localy, I've totally missed that it wasn't in the commit and it isn't tracked by git. How to fix it?
First time working with such a huge project, sorry for being such a burden.

JimBobSquarePants:
>First time working with such a huge project, sorry for being such a burden.

Not at all!! You're contributions are most welcome!

That's my mistake for not being clear. `ThrowHelper` in the `ShareInfrastructure` project cannot be edited (As it's included via a submodule) but it is a partial so I would make another part for that class in the ImageSharp project and add your method there.

br3aker:
Got it, thanks for answering yet another question!

JimBobSquarePants:
@br3aker Can you please give me write access to your fork. For some reason I'm unable to push directly to your PR so we can complete this.

br3aker:
@JimBobSquarePants that's strange. Edit was enabled from the start, I re-checked it now, can you once more please?

JimBobSquarePants:
@br3aker This isn't the first PR I've had issues with and I'm not sure why it happens. I tried again and it didn't work. The last time the contributor had to explicitly invite me as a contributor to the repo as a workaround.

br3aker:
@JimBobSquarePants invite sent, thanks for fixing this!

JimBobSquarePants (approved):
LGTM 👍 Thanks!

Issue #1597, opened 2021-04-14
JpegDecoder: post-process baseline spectral data per MCU-row


antonfirsov:
Currently, Huffmann decoding (done by `HuffmanScanDecoder`) is strictly separated from postprocessing/color conversion (done by `JpegImagePostProcessor`) for simplicity. This means that [`JpegComponent.SpectralBlocks`](https://github.com/SixLabors/ImageSharp/blob/97dde7fb2f0ac97b5b1f9e6b0555a53d2e270c46/src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegComponent.cs#L63) are allocated upfront for the whole image.

I did a second round of memory profiling using `SimpleGcMemoryAllocator` to get rid of pooling for more auditable results. This shows that `SpectralBlocks` are responsible for the majority of our memory allocation:

![image](https://user-images.githubusercontent.com/6835152/114704223-a5bee000-9d26-11eb-95b4-a68f683dfa8d.png)

This can be eliminated with some non-trivial, but still limited refactoring:
- `JpegDecoderCore` and `HuffmanScanDecoder` needs a mode where `JpegComponent.SpectralBlocks` is interpreted as a sliding window of blocks instead of full set of decoded spectral blocks
- `HuffmanScanDecoder` can then push the rows of the sliding window, directly calling an instance of `JpegComponentPostprocessor` in the end of it's [MCU-row decoding loop](https://github.com/SixLabors/ImageSharp/blob/97dde7fb2f0ac97b5b1f9e6b0555a53d2e270c46/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanScanDecoder.cs#L154-L205)

JimBobSquarePants:
I can definitely get behind this. As I recall looking at other libraries they were working on a per MCU process but I think per MCU row is fine. 

It's a shame this optimization is only limited to sequential jpegs but is definitely worth the effort.

antonfirsov:
I hope to be able to have a look at this within ~3 weeks.

JimBobSquarePants:
@br3aker is this something you'd be interested in? You've been doing amazing work in the encoder!

br3aker:
@JimBobSquarePants I can definately get behind this after PR with that deBruijn table I was talking about at memory allocator PR.

br3aker:
Did a little study to understand decoder architecture. Long story short: code explosion due to generic `TPixel` type propagation. Current workflow for jpeg encoder:
```
[...]

Issue #1596, opened 2021-04-13
More efficient MemoryAllocator


antonfirsov:
One way to address the concerns in #1590 is to come up with a new `MemoryAllocator` that meets the following two requirements even under very high load:
(A) Steady allocation patterns, instead of GC fluctuations
(B) Lower amount of memory being retained, at least "after some time"

Some ideas to explore:
1. Allocate native memory over a certain threshold (~1MB) deferring the memory management to the OS
    1.1 Consider pooling unmanaged memory, especially if we implement the next point.
2. Go discontigous, and build all large buffers from fix-sized blocks of memory similarly to [RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream). It's worth to check how does this work with both pooled arrays and unmanaged memory.
3. Have a threshold on the reatained memory, and release some of the pooled arrays (or pooled unmanaged buffers) when we grow over it.
4. Mentioning an idea I really hate for the sake of completeness: Have a synchronization mechanism around large buffer acquisition, similarly to System.Drawing (WIC?). To do this properly, memory allocation should become an asynchronous operation, otherwise SpinLocks will just add more fuel to the fire.

Point 1. seems to be very simple to prototype. We need an allocator that uses `Marshal.AllocHGlobal` over some threshold, and an `ArrayPool` below it, and see how the memory timeline goes with the [bee heads MemoryStress benchmark](https://github.com/saucecontrol/core-imaging-playground/tree/beeees/MemoryStress) in comparison to `ArrayPoolMemoryAllocator`.

@saucecontrol any thoughts or further ideas? (Especially on point 2.)

saucecontrol:
Idea 2 is interesting. I wonder, though, if you can manage discontinous buffers, is it much of a stretch from there to go fully streamed?

For the record, I also hate idea 4, but that may be necessary if you really do have to materialize large images and care at all about 32-bit hosts.  The machine I ran my tests on was only a quad core as well.  I wonder what would happen on a 16+ core machine... I reckon even in 64-bit you'd start paging.

antonfirsov:
>  I wonder, though, if you can manage discontinous buffers, is it much of a stretch from there to go fully streamed?

[...]

Issue #1376, opened 2020-10-11
Test coverage not reliable because of RemoteExecutor bug


antonfirsov:
Some of our tests are using the RemoteExecutor tool from dotnet/arcade.

It looks like these tests are not reporting potential failures because of dotnet/arcade#6371. We should resolve this ASAP.

JimBobSquarePants:
Switching to the Ubuntu build might be enough for now. That seems to work according to my tests

JimBobSquarePants:
I've removed the blocking label as we have a workaround but I'm keeping the issue open so we can track it as it makes testing a little more difficult on Windows in some situations..

antonfirsov:
> I've removed the blocking label

I see your point, but I believe that after all the SIMD refactors are done, it would be important to get a reliable RemoteExecutor run on Windows on too before pushing out a new release. Doesn't have to be on CI, could be an ugly one-time workaround, but we probably want to avoid surprises, even if it would cost us waiting 1-2 more weeks.

I can have a look at this, after finishing the active edges stuff.

JimBobSquarePants:
Yeah... I've been doing all this manually so far by writing/testing SSE first then adding AVX. It's not been fun!

I can't seem to get any movement upstream to get a fix in either. I'm hoping after the .NET5 launch I'll see something.

antonfirsov:
Considering all the SIMD PR-s happening, I wouldn't block new releases until this is fixed.

Note adapting https://github.com/dotnet/arcade/pull/7426 became a non-trivial question, because arcade dropped 2.1 support in https://github.com/dotnet/arcade/pull/6891, and our tests still target 2.1.

The solution depends on our .NET Core 2.1 end-of-support strategy. Official support [ends on August 21, 2021](https://dotnet.microsoft.com/platform/support/policy/dotnet-core).

We can do 3 things:
1. Hard cut 2.1 from both product and test code. I wouldn't do it before September, and we probably don't want to block the fix until that time.
2. Cut 2.1 from tests, but keep product code building against 2.1 for a few more months
3. Do some hack in msbuild, eg. build 2.1 target can refer to earlier version of RemoteExecutor

JimBobSquarePants:
Wow. They’re really doing their best to make an open source project as useless for the general public as possible aren’t they? Kinda defeats the purpose. 

Does RemoteExecutor actually have any dependencies on other projects? Could we strip it out and maintain it separately? 

What was that alternative library?

[...]

Issue #449, opened 2018-01-24
Feature Request: Save Progressive jpeg


nla-brandonjames:
### Description
I can't find an API in ImageSharp to save my images in the progressive jpeg format. It is a requirement for my application. I am capable of saving as a baseline image.

- ImageSharp version: 1.0.0-beta2
- Other ImageSharp packages and versions:
- Environment (Operating system, version and so on): Windows 10 (dev) Ubuntu 16.04 (prod)
- .NET Framework version: .NET Core 2.0.5

JimBobSquarePants:
@nla-brandonjames 

Unfortunately we don't have support for saving progressive jpegs. See #10 It's something I REALLY want though.

We'd need support from someone in the community if we are in any way likely to get it for v1.0 as jpeg is.... complicated. At the moment you would have to post process the images using something like mozjpeg. 

Cheers

James

P.S In the future could you please ask questions in the Gitter channel? We're trying to keep the issue tracker clear so we can stay on top of things.

nla-brandonjames:
@JimBobSquarePants Okay, will do. I have a couple more things to add to this.

JimBobSquarePants:
Thanks! 👍

maartenmensink:
Is this still on the roadmap?

JimBobSquarePants:
If someone wants to implement it yes.

maartenmensink:
We are willing to help with the implementation for sure. My first gut feeling is that we would expand the current JpegEncoder.

With a property Progressive which should default to false.

![image](https://user-images.githubusercontent.com/339653/120468145-3027d580-c3a1-11eb-97e7-ee5b0c432525.png)

Jpeg Progressive Encoder

- Second we need to port a good working progressive encoder like libjpeg-turbo if that is offcourse allowed with respect to their license 
- Create one from scratch.

What are your considerations in regards to this?

antonfirsov:
> port a good working progressive encoder like libjpeg-turbo

Note that libjpeg-turbo has it's own memory management primitives, a bunch of C indirections, and the optimized SIMD methods are implemented in assembly, which makes it extremely hard to port it in an efficient manner. There is a naive port of classic LibJpeg under https://github.com/BitMiracle/libjpeg.net/tree/master/LibJpeg/Classic/Internal. Such a port will leave you in a place where both the performance and the code quality is quite bad, and a huge amount of additional work has to be spent to bring it to an acceptable level.

[...]

PR #1632, opened 2021-05-22
Jpeg encoding code optimization

br3aker:
### Prerequisites

### Description
**This is a draft due to 4 and some small internal jpeg changes I will adress in a separate comment.** Want to hear your thoughts on this.

Jpeg format encoding optimization & code cleanup.
1. Moved scan encoding logic to a separate class `HuffmanScanEncoder` following logic of `HuffmanScanDecoder` in jpeg decoding code (did todo comment from Encode444/Encode420, basically).
2. Previously, encoder was writing encoded scan byte-by-byte, buffer logic existed but it wasn't used properly. This PR fixes it and introduces new buffer size - `1024`, Looks like it's  optimal for general use case, lesser size would lead to worse performance, bigger size doesn't bring any performance boost at all (on my machine atleast). Right now it is defined by constant at `HuffmanScanEncoder` class, it can be injected from the callee/global config but I'm not sure it is really needed.
3. Introduced 256bit 8x8 matrix forward/inverse DC transforms via avx instruction set.
4. **(This is questionable)** Introduced explicit layout for `Block8x8F`. Most, if not all, avx-intrinsic code use a lot of `Unsafe.As` casts like here:
```csharp
var valueVec = Vector256.Create(value);
Unsafe.As<Vector4, Vector256<float>>(ref this.V0L) = Avx.Multiply(Unsafe.As<Vector4, Vector256<float>>(ref this.V0L), valueVec);
```
With new layout it can be simplified to:
```csharp
var valueVec = Vector256.Create(value);
this.V0 = Avx.Multiply(this.V0, valueVec);
```
There's also a performance gain but it's too small to really affect any heavy algorithms like jpeg decoding/encoding.
**Note**: This PR does not change any existing code with `Unsafe.As` calls. If it gets merged - this can be a separate `codequality` issue (good-first-issue maybe?)

### Stats
Benchmarks are based on this noise image with different sizes:
200x200

Also tested some random images from beeheads demo but they are too large for gaining stable results in a reasonable amount of time :P
Even so, got ~100-200ms performance gain on those.

```
.NET Core SDK=6.0.100-preview.3.21202.5
```

Jpeg encoding to `MemoryStream` (disk writes are less stable):
|                   Method |     Mean |    Size |   Iterations |
|------------------------- |---------:|---------:|---------:|
|            Current | 3 ms | 200x200 | 250 |
|            Current | 11 ms | 400x400 | 250 |
|            Current | 44 ms | 800x800 | 250 |
|            Current | 177 ms | 1600x1600 | 250 |
|            New | 2 ms | 200x200 | 250 |
|            New | 8 ms | 400x400 | 250 |
|            New | 32 ms | 800x800 | 250 |
|            New | 133 ms | 1600x1600 | 250 |

Avx accelerated cosine transforms:

|                   Method |     Mean |    Error |   StdDev | Ratio |
|------------------------- |---------:|---------:|---------:|------:|
|            TransformFDCT | 64.43 ns | 0.541 ns | 0.506 ns |  1.00 |
| TransformFDCT_intrinsics | 50.24 ns | 0.361 ns | 0.320 ns |  0.78 |

|                   Method |     Mean |    Error |   StdDev | Ratio |
|------------------------- |---------:|---------:|---------:|------:|
|            TransformIDCT | 75.21 ns | 0.382 ns | 0.339 ns |  1.00 |
| TransformIDCT_intrinsics | 58.96 ns | 0.306 ns | 0.271 ns |  0.78 |


src/ImageSharp/Formats/Jpeg/Components/FastFloatingPointDCT.cs
```diff
@@ -309,11 +198,106 @@ public static void FDCT8x4_RightPart(ref Block8x8F s, ref Block8x8F d)
...
+            // 2 6
+            d.V2 = SimdUtils.HwIntrinsics.MultiplyAdd(Avx.Multiply(c2, C_V_0_5411), c3, C_V_1_3065);
+            if (Fma.IsSupported)
+            {
+                d.V6 = Fma.MultiplySubtract(c3, C_V_0_5411, Avx.Multiply(c2, C_V_1_3065));
+            }
+            else
+            {
+                d.V6 = Avx.Subtract(Avx.Multiply(c3, C_V_0_5411), Avx.Multiply(c2, C_V_1_3065));
+            }
```
antonfirsov, line 239: I would define a method `SimdUtils.HwIntrinsics.MultiplySubtract` to get rid of the if blocks.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
```diff
@@ -27,29 +27,64 @@ internal partial struct Block8x8F : IEquatable<Block8x8F>
...
+        public Vector256<float> V3;
+        [FieldOffset(128)]
+        public Vector256<float> V4;
+        [FieldOffset(160)]
+        public Vector256<float> V5;
+        [FieldOffset(192)]
+        public Vector256<float> V6;
+        [FieldOffset(224)]
+        public Vector256<float> V7;
+#endif
```
antonfirsov, line 87: @saucecontrol are you aware of any codegen bugs or perf issues with `[FieldOffset]` + intrinsics?

antonfirsov (commented):
Thanks! We need to change `DCTTests.FastFloatingPoint.TransformFDCT` to stress the 3 possible paths (Vector4 vs AVX vs AVX+FMA) using `FeatureTestRunner` and this is good to go.

/cc @saucecontrol if you can spot something trivial in the SIMD code.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
br3aker: There shouldn't be any, either way a lot of avx code already assumes that left/right vectors are aligned this way:
```csharp
Avx.Multiply(Unsafe.As<Vector4, Vector256<float>>(ref this.V0L), valueVec)
```
If there are any issues with that, all `Unsafe.As` casts in avx should be affected.

antonfirsov:
@br3aker re benchmarking:

we have existing benchmarks here:
https://github.com/SixLabors/ImageSharp/blob/master/tests/ImageSharp.Benchmarks/Codecs/Jpeg/EncodeJpeg.cs

Can you run this and post the results before/after?

If you have some extra time you feel free to add your FDCT benchmark into that project.

src/ImageSharp/Formats/Jpeg/Components/FastFloatingPointDCT.cs
br3aker: Forgot about this, will do!

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -635,37 +296,47 @@ private void WriteDefineHuffmanTables(int componentCount)
                 markerlen += 1 + 16 + s.Values.Length;
             }
 
+            // TODO: this magic constant (array size) should be defined by HuffmanSpec class
+            // This is a one-time call which can be stackalloc'ed or allocated directly in memory as method local array
+            // Allocation here would be better for GC so it won't live for entire encoding process
+            // TODO: if this is allocated on the heap - pin it right here or following copy code will corrupt memory
```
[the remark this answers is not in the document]
br3aker: Main point of this is that previously this array was allocated for entire encoding run, including scan and everything else, not a big hit on gc but better to free sooner then later. Haven't though about pooling here, nice catch!
The only question is: use pool from provided image or from global configuration?

br3aker:
@antonfirsov sure thing!
Should I add FDCT/IDCT benchmarks to `ImageSharp.Benchmarks/Codecs/Jpeg/BlockOperations/` or directly to `ImageSharp.Benchmarks/Codecs/Jpeg/`?

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
antonfirsov: Just for clarity: this is not about layout, your explicit layout is equivalent to the implicit sequential layout before the change.

I expect the runtime having different implementations for codegening `Unsafe.As` and `[FieldOffset]`. Normally they should be equivalent, but we've seen a couple of weird bugs in the past around similar areas.

While I trust `Unsafe.As`, seeing `[FieldOffset]` together with intrinsics kinda frightens me :) Let's see if @saucecontrol can react.

antonfirsov:
`BlockOperations` is a better place.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
br3aker: I see, never thought this could cause any troubles, interested to hear his opinion on this :)

br3aker:
Benchmarks from `EncodeJpeg` builtin benchmark:

Old
|                Method |      Mean |     Error |    StdDev | Ratio | RatioSD |
|---------------------- |----------:|----------:|----------:|------:|--------:|
| 'System.Drawing Jpeg' |  3.215 ms | 0.0623 ms | 0.0583 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 45.137 ms | 0.5645 ms | 0.4713 ms | 14.01 |    0.26 |

New
|                Method |      Mean |     Error |    StdDev | Ratio | RatioSD |
|---------------------- |----------:|----------:|----------:|------:|--------:|
| 'System.Drawing Jpeg' |  3.204 ms | 0.0329 ms | 0.0292 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 41.426 ms | 0.8184 ms | 0.7655 ms | 12.95 |    0.29 |

@antonfirsov although results are better for this PR, it's differs from commented benchmark by a LOT:

|                Method |     Mean |     Error |    StdDev | Ratio | RatioSD |
|---------------------- |---------:|----------:|----------:|------:|--------:|
| 'System.Drawing Jpeg' | 4.297 ms | 0.0244 ms | 0.0228 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 5.286 ms | 0.1034 ms | 0.0967 ms |  1.23 |    0.02 |

Am I doing something wrong?

antonfirsov:
@br3aker I changed the benchmark to use a larger test image in 1968d3724fa81ac381145d4acc1dd1d36e89154f, but forgot to delete the old test results from the file. The new image provides more realistic results, the encoder perf gap is still really huge.

I benchmarked your stuff on my surface book (Core i7-8650U CPU 1.90GHz), results are similar to yours:

#### 

#### Current master
```
|                Method |      Mean |     Error |    StdDev | Ratio | RatioSD |
|---------------------- |----------:|----------:|----------:|------:|--------:|
| 'System.Drawing Jpeg' |  3.668 ms | 0.0684 ms | 0.1085 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 55.544 ms | 0.4942 ms | 0.3858 ms | 14.96 |    0.47 |
```


#### PR

```
|                Method |      Mean |     Error |    StdDev | Ratio | RatioSD |
|---------------------- |----------:|----------:|----------:|------:|--------:|
| 'System.Drawing Jpeg' |  3.597 ms | 0.0716 ms | 0.0824 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 43.468 ms | 0.3445 ms | 0.3054 ms | 12.02 |    0.26 |
```

Feel free to replace the old results with yours!

br3aker:
@antonfirsov I think I found an issue with this benchmark, it's possible that System.Drawing is using cached results so that `3.5ms` thingy is not valid to compare. Gimme a moment, will be back soon and check it.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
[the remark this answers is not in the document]
antonfirsov: But now it's stackalloced only for the `WriteDefineHuffmanTables` run isn't it? For a 179 byte array stackalloc is optimal, I still don't see why you consider to change it.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
saucecontrol: I'm not aware of any issues, but I'd be happy to pull this down and give the codegen a look.  Will try to get to that this afternoon.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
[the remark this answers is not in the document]
br3aker: Oh, I misunderstood you, it **was allocated** as class field for entire run before I moved everything to a separate class and stackallocted it :P
So that was a question: is stackalloc here good enough or should I change it to pooling/actual allocation.
I will remove TODOs and introduce constant for the size then.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
br3aker: @saucecontrol thanks!

br3aker:
@antonfirsov yep, it wasn't a fair benchmark for ImageSharp, System.Drawing is indeed using cached encoded data for [short pass saving if saving encoder format is equal to decoder format](https://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/Image.cs,707).

Providing coder & parameters explicitely yields 'true' results:
```csharp
bmpDrawing.Save(this.destinationStream, this.jpegCodec, this.jpegEncoderParameters);
```
90 quality was yielding somewhat similar pictures for both System.Drawing and ImageSharp

**[90 quality] Current master**
|                Method |     Mean |    Error |   StdDev | Ratio | RatioSD |
|---------------------- |---------:|---------:|---------:|------:|--------:|
| 'System.Drawing Jpeg' | 33.23 ms | 0.353 ms | 0.313 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 49.52 ms | 0.817 ms | 0.724 ms |  1.49 |    0.02 |

**[90 quality] PR**
|                Method |     Mean |    Error |   StdDev | Ratio | RatioSD |
|---------------------- |---------:|---------:|---------:|------:|--------:|
| 'System.Drawing Jpeg' | 33.16 ms | 0.346 ms | 0.324 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 40.26 ms | 0.442 ms | 0.369 ms |  1.22 |    0.02 |

Not so far behind actually :P
Furthermore, ImageSharp creates image with bigger file size for the same quality than System.Drawing does.
Now this is really interesting: for 100 quality both ImageSharp and System.Drawing produce almost identical results for this PR:

**[100 quality] PR**
|                Method |     Mean |    Error |   StdDev | Ratio |
|---------------------- |---------:|---------:|---------:|------:|
| 'System.Drawing Jpeg' | 39.77 ms | 0.345 ms | 0.323 ms |  1.00 |
|     'ImageSharp Jpeg' | 40.48 ms | 0.446 ms | 0.396 ms |  1.02 |

What's still not clear to me - why is less than 100 quality gap is so large. `Quality=75` yields same performance as `Quality=100` for ImageSharp:

**[75 quality] PR**
|                Method |     Mean |    Error |   StdDev | Ratio | RatioSD |
|---------------------- |---------:|---------:|---------:|------:|--------:|
| 'System.Drawing Jpeg' | 29.79 ms | 0.259 ms | 0.217 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 49.10 ms | 0.949 ms | 0.842 ms |  1.65 |    0.04 |

@JimBobSquarePants you should definately see this, has System.Drawing been finally defeated in this performance showdown? :D

saucecontrol:
> System.Drawing is indeed using cached encoded data for [short pass saving if encoder format is equal to decoder format](https://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/Image.cs,707).

Good catch!

> What's still not clear to me - why is less than 100 quality gap is so large.

I can answer that one.  The original IJG libjpeg implementation (which most implementations are based on) includes both a 'fast' and 'slow' DCT calculation, where the fast one does 16-bit fixed-point math and the slow one uses floating-point or 32-bit integer math (can't remember which offhand).  They give the same results up to about q=95, so the fast one is used. Above that, they revert to the slow one.

JimBobSquarePants:
Oh wow! Great catch! We should check our other encoding benchmarks also. 

This is turning into an excellent day 😀

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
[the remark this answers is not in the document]
antonfirsov: Yeah, the whole comment was about removing the TODO-s since they are misleading/unnecessary, stackalloc is more than good here.

antonfirsov:
> Providing codec & parameters explicitely yields 'true' results

Feel free to push the benchmark changes!

JimBobSquarePants:
Thinking about the DCT stuff, do you k ow by what criteria either is used during decode? We are able to determine the encoded quality so (unless it was too much of a refactor) we could potentially choose that kind of optimisation.

saucecontrol:
AFAIK, the decoder always uses the slow IDCT, although libjpeg-turbo has an option to override that in its alternate turbojpeg API.  But analyzing the coefficient tables and picking the fast version when appropriate would seem to me to be a viable option.

JimBobSquarePants:
>Furthermore, ImageSharp creates image with bigger file size for the same quality than System.Drawing does.

@br3aker  System.Drawing always uses 2x2 (4:2:0)  subsampling as I recall.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
saucecontrol: Codegen around the ExplicitLayout struct looks good.  The codegen around the vector constants could be a bit better, for example, from the FDCT:

```C#
d.V2 = SimdUtils.HwIntrinsics.MultiplyAdd(Avx.Multiply(c2, C_V_0_5411), c3, C_V_1_3065);
d.V6 = SimdUtils.HwIntrinsics.MultiplySubtract(c3, C_V_0_5411, Avx.Multiply(c2, C_V_1_3065));
```
(assuming you get that `MultiplySubtract` helper in 😸)
compiles to:

```asm
mov         rax,16E6ABBA590h                 ' load C_V_0_5411 offset
mov         rax,qword ptr [rax]              ' load C_V_0_5411 address
vmulps      ymm9,ymm4,ymmword ptr [rax+8]  
mov         rax,16E6ABBA598h                 ' load C_V_1_3065 offset
mov         rax,qword ptr [rax]              ' load C_V_1_3065 address
vfmadd231ps ymm9,ymm2,ymmword ptr [rax+8]  
vmovupd     ymmword ptr [rdx+40h],ymm9  
mov         rax,16E6ABBA590h                 ' C_V_0_5411 again
mov         rax,qword ptr [rax]              ' "
mov         rcx,16E6ABBA598h                 ' C_V_1_3065 again  
mov         rcx,qword ptr [rcx]              ' "
vmulps      ymm4,ymm4,ymmword ptr [rcx+8]  
vfmsub132ps ymm2,ymm4,ymmword ptr [rax+8]
vmovupd     ymmword ptr [rdx+0C0h],ymm2  
```

You could try loading them into temp locals for re-use, but I'm not sure if you'll run out of SIMD registers doing that.  I only see up to ymm10 being used, so you should have a few to play with.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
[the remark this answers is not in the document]
br3aker: Actually, isn't it better to just write buffers like this?
```csharp
for (int i = 0; i < specs.Length; i++)
{
    this.outputStream.WriteByte(headers[i]);
    this.outputStream.Write(specs[i].Count);
    this.outputStream.Write(specs[i].Values);
}
```

No extra buffers, no extra constants, no memory copies and code is simplier by a lot.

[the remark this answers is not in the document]
JimBobSquarePants: Yeah, those methods shouldn’t require anything more complicated. They’re called so few times.

Note to self: implement BufferedWriteStream.

br3aker:
@antonfirsov @saucecontrol encountered a little confusion with `MultiplyAdd` method in `SimdUtils`.
Multiply/Add looks like this:
```csharp
public static Vector256<float> MultiplyAdd(va, vm0, vm1)
{
    if (Fma.IsSupported)
    {
        return Fma.MultiplyAdd(vm1, vm0, va);
    }
    else
    {
        return Avx.Add(Avx.Multiply(vm0, vm1), va);
    }
}
```

So `ma(a, b, c) => b * c + a` looks good. 
But `MultiplySubstract` with similar parameter 'placement' looks a little misleading to me: `ms(a, b, c) => b * c - a`.
Isn't `ms(a, b, c) => a * b - c` parameter placement better? What's your opinion on this?

br3aker:
> System.Drawing always uses 2x2 (4:2:0) subsampling as I recall.
@JimBobSquarePants it's really hard to find info about this without manually checking header size of resulting System.Drawing image, looks like it's using 4:1:1 subsampling so 4:2:0 should be close enough in benchmarking I think.

saucecontrol:
> So `ma(a, b, c) => b * c + a` looks good.
> But hipothetical `MultiplySubstract` with similar parameter 'placement' looks a little misleading to me: `ms(a, b, c) => b * c - a`.
> Isn't `ms(a, b, c) => a * b - c` parameter placement better?

I'm not sure, but it looks like @JimBobSquarePants lifted that `MultiplyAdd` helper from my version here https://github.com/saucecontrol/PhotoSauce/blob/cf81241695b4c8e361a519886974f524f61f6aa0/src/MagicScaler/Utilities/HWIntrinsics.cs#L110-L126

Mine was written with that parameter order specifically to work around https://github.com/dotnet/runtime/issues/12212 in the overload that takes a pointer for one of the multiply operands.  And because I'm almost always accumulating into the add operand, it makes the method parameter order match the actual operand order of the instruction.  I made the one that takes all vectors match the parameter order for consistency.

I haven't spent much time looking at the codegen around `MultiplySubtract` so I don't know if it has the exact same quirks around operand containment, but if you're moving the constant to a local for re-use, that won't even be a factor anymore.  I'm not sure you'd want to have the two of them be wildly different from a consistency perspective, but there's no technical reason you can't do it.

> it's really hard to find info about this without manually checking header of resulting System.Drawing file, looks like it's using 4:1:1 subsampling

It's definitely using 4:2:0 subsampling, which confusingly is a 4:1:1 actual pixel ratio by plane.  The easiest way to get that kind of info is with [ExifTool](https://exiftool.org/install.html), which will give you output like this:

```
Encoding Process                : Baseline DCT, Huffman coding
Bits Per Sample                 : 8
Color Components                : 3
Y Cb Cr Sub Sampling            : YCbCr4:2:0 (2 2)
```

br3aker:
Yeah, inconsistency between those two is the main reason I don't like it so I'll stick with MultiplyAdd parameter order. Thanks for indepth explanation behind codegen!

br3aker:
One more (probably the last one for this PR) question.
There's a bit count lookup table:
```csharp
uint bt;
if (a < 0x100)
{
    bt = BitCountLut[a];
}
else
{
    bt = 8 + (uint)BitCountLut[a >> 8];
}
```
Maybe it's better to use `Lzcnt` op thought `BitOperations.LeadingZeroesCount` if it's supported?
```csharp
uint bt;
if (Lzcnt.IsSupported) // Same as sse/avx runtime constant branch elimination
{
    bt = LzcntImplementation(value); // (uint)(32 - BitOperations.LeadingZeroCount((uint)value));
}
else
{
    bt = BitCountLookup(value); // current implementation
}
```
It is a little more performant than lookup but doesn't really make any difference on a global scale though.

antonfirsov:
> Maybe it's better to use `Lzcnt` op thought `BitOperations.LeadingZeroesCount` if it's supported?

Don't you mean `Lzcnt.LeadingZeroCount`? I don't follow the relationship between the content of that LUT and lzctn, can you elaborate the math?

br3aker:
Oh, I didn't know `Lzcnt.LeadingZeroCount` intrinsic existed, that's even better. `BitOperations.LeadingZeroCount` is basically the same as `Lzcnt.LeadingZeroCount` with `Lzcnt.IsSupported` check and fallback code.

That lookup table is used to find how many minimum bits is need to store some 16bit arbitary value (thus if-else in `EmitHuffRLE` method).
Same can be evaluated via Lzcnt instruction:
Let's take 170 for example:
00000000_00000000_00000000_10101010 - 24 zeroes before first 1, lzcnt would return 24
(sizeof(uint) * 8) - lzcnt(170) => 32 - 24 => 8 minimum bits needed to store this value, same as LuT would return

This should be a bit faster - no branching, less instructions and no extra memory usage.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -638,34 +299,16 @@ private void WriteDefineHuffmanTables(int componentCount)
...
-                    for (int v = 0; v < spec.Values.Length; v++)
-                    {
-                        huffman[len++] = values[v];
-                    }
-                }
-
-                this.outputStream.Write(this.huffmanBuffer, 0, len);
+                this.outputStream.WriteByte(headers[i]);
+                this.outputStream.Write(specs[i].Count);
+                this.outputStream.Write(specs[i].Values);
```
antonfirsov, line 304: Much better 👍

antonfirsov:
@br3aker sounds good, thanks for clarifying!

saucecontrol:
> didn't know `Lzcnt.LeadingZeroCount` intrinsic existed, that's even better

I'd recommend using `BitOperations.LeadingZeroCount` over `Lzcnt.LeadingZeroCount` in every case.  The former is optimized on all x86/x64 and arm64, rather than being tied to support for a specific ISA.  In fact, I removed all uses of `Lzcnt.LeadingZeroCount` (except those used to implement `BitOperations` methods) from the BCL in a PR last year.

Note that it's not optimized for x86/x64 without LZCNT support in netcoreapp3.1 (I added that for net5.0), but the branch-free fallback in 3.1 is probably no worse than the current LUT implementation in ImageSharp and should be hit pretty rarely.

br3aker:
Looks like `Lzcnt` actually did matter in performance:

Lookup table
|                Method |     Mean |    Error |   StdDev | Ratio | RatioSD |
|---------------------- |---------:|---------:|---------:|------:|--------:|
| 'System.Drawing Jpeg' | 39.54 ms | 0.269 ms | 0.225 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 47.25 ms | 0.937 ms | 1.219 ms |  1.20 |    0.02 |

Lzcnt
|                Method |     Mean |    Error |   StdDev | Ratio | RatioSD |
|---------------------- |---------:|---------:|---------:|------:|--------:|
| 'System.Drawing Jpeg' | 39.67 ms | 0.774 ms | 0.828 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg' | 45.39 ms | 0.415 ms | 0.346 ms |  1.14 |    0.03 |

-------------------------------------------------------------------------------

antonfirsov:
> error CS0103: The name 'BitOperations' does not exist in the current context

I'm afraid we need the LUT for `netstandard <= 2.0` and .NET Framework.

antonfirsov:
Should be easy to ifdef after https://github.com/SixLabors/SharedInfrastructure/pull/18 is merged. (Will require a submodule update though.)

JimBobSquarePants:
> Should be easy to ifdef after SixLabors/SharedInfrastructure#18 is merged. (Will require a submodule update though.)

Merged. 

Use the following in the repository root to update.

``` bash
```

antonfirsov:
Does this need anything other than tests to be finished?

br3aker:
@antonfirsov nope, still working. Optimized 4:2:0 subsampling with following results:
(4:2:0) Current master
|            Method | Quality |     Mean |    Error |   StdDev |
|------------------ |-------- |---------:|---------:|---------:|
| 'ImageSharp Jpeg' |      50 | 33.35 ms | 0.657 ms | 1.004 ms |
| 'ImageSharp Jpeg' |      75 | 37.31 ms | 0.692 ms | 0.797 ms |
| 'ImageSharp Jpeg' |      95 | 51.04 ms | 1.020 ms | 1.941 ms |
| 'ImageSharp Jpeg' |     100 | 55.79 ms | 1.043 ms | 0.924 ms |


(4:2:0) Stream write fix
|            Method | Quality |     Mean |    Error |   StdDev |
|------------------ |-------- |---------:|---------:|---------:|
| 'ImageSharp Jpeg' |      50 | 31.37 ms | 0.260 ms | 0.217 ms |
| 'ImageSharp Jpeg' |      75 | 34.90 ms | 0.614 ms | 0.656 ms |
| 'ImageSharp Jpeg' |      95 | 44.59 ms | 0.852 ms | 1.078 ms |
| 'ImageSharp Jpeg' |     100 | 48.80 ms | 0.962 ms | 0.900 ms |


(4:2:0) Pre YCbCr rgb 4-to-1 conversion
|            Method | Quality |     Mean |    Error |   StdDev |
|------------------ |-------- |---------:|---------:|---------:|
| 'ImageSharp Jpeg' |      50 | 27.67 ms | 0.519 ms | 0.533 ms |
| 'ImageSharp Jpeg' |      75 | 31.16 ms | 0.610 ms | 0.653 ms |
| 'ImageSharp Jpeg' |      95 | 40.31 ms | 0.797 ms | 0.745 ms |
| 'ImageSharp Jpeg' |     100 | 45.64 ms | 0.898 ms | 0.998 ms |

Didn't have enough time to work on it this week - these results are for simd supported code only, fallback code is yet to be implemented.

JimBobSquarePants:
Looking forward to seeing the changes. I've just pushed a fix for the BitCount fallback.

JimBobSquarePants:
Thought I'd just share how this branch fairs in the playground benchmarks. Finally beating NETVips! 🎉 

@saucecontrol is still way ahead of the pack of course (_and untouchable_) but I think we can set ourselves away from the pack once this PR is complete and we tackle #1596 and #1597

```csv
BenchmarkDotNet=v0.12.1.1514-nightly, OS=Windows 10.0.19043
.NET SDK=5.0.300
  [Host]       : .NET 5.0.6 (5.0.621.22011), X64 RyuJIT
  .Net 5.0 CLI : .NET 5.0.6 (5.0.621.22011), X64 RyuJIT

Job=.Net 5.0 CLI  Arguments=/p:DebugType=portable  Toolchain=.NET 5.0
IterationCount=5  LaunchCount=1  WarmupCount=5

|                                Method |      Mean |     Error |   StdDev | Ratio |     Gen 0 |     Gen 1 |     Gen 2 | Allocated | Allocated native memory | Native memory leak |
|-------------------------------------- |----------:|----------:|---------:|------:|----------:|----------:|----------:|----------:|------------------------:|-------------------:|
|   'System.Drawing Load, Resize, Save' | 406.55 ms |  6.168 ms | 0.954 ms |  1.00 |         - |         - |         - |     12 KB |               11,376 KB |             738 KB |
|       'ImageSharp Load, Resize, Save' | 161.16 ms |  2.470 ms | 0.642 ms |  0.40 |  250.0000 |         - |         - |  1,614 KB |                    9 KB |                  - |
|      'ImageMagick Load, Resize, Save' | 435.60 ms | 17.824 ms | 2.758 ms |  1.07 |         - |         - |         - |     54 KB |               61,313 KB |                  - |
|        'ImageFree Load, Resize, Save' | 273.11 ms |  8.509 ms | 1.317 ms |  0.67 | 6000.0000 | 6000.0000 | 6000.0000 |     91 KB |               49,137 KB |              58 KB |
|      'MagicScaler Load, Resize, Save' |  74.51 ms |  2.178 ms | 0.566 ms |  0.18 |         - |         - |         - |    104 KB |                3,274 KB |              18 KB |
| 'SkiaSharp Canvas Load, Resize, Save' | 253.56 ms |  4.067 ms | 1.056 ms |  0.62 |         - |         - |         - |     99 KB |               69,710 KB |           1,662 KB |
| 'SkiaSharp Bitmap Load, Resize, Save' | 253.00 ms |  3.964 ms | 0.613 ms |  0.62 |         - |         - |         - |     85 KB |               70,424 KB |             949 KB |
|          'NetVips Load, Resize, Save' | 165.82 ms | 20.405 ms | 5.299 ms |  0.41 |         - |         - |         - |     48 KB |               44,198 KB |           3,514 KB |

// * Legends *
  Mean                    : Arithmetic mean of all measurements
  Error                   : Half of 99.9% confidence interval
  StdDev                  : Standard deviation of all measurements
  Ratio                   : Mean of the ratio distribution ([Current]/[Baseline])
  Gen 0                   : GC Generation 0 collects per 1000 operations
  Gen 1                   : GC Generation 1 collects per 1000 operations
  Gen 2                   : GC Generation 2 collects per 1000 operations
  Allocated               : Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)
  Allocated native memory : Allocated native memory per single operation
  Native memory leak      : Native memory leak size in byte.
  1 ms                    : 1 Millisecond (0.001 sec)
```

br3aker:
@JimBobSquarePants what's the name of that benchmark? Wanna try it with current optimization which isn't pushed yet :P

JimBobSquarePants:
@br3aker Here it is. 
https://github.com/bleroy/core-imaging-playground

I just modified my local copy to include ImageSharp as a project as a reference rather than a NuGet one.

br3aker:
1. Added tests
2. Optimized 420 subsampling for both simd and non-simd paths (results below).
3. There's actually some other places encoder can be potentially optimized but it won't affect code globally so it can be merged after review so folks can work on https://github.com/SixLabors/ImageSharp/issues/449
4. I didn't run *pack* performance benchmark but it should be faster by a couple of ms than results above
5. There's an unused method and even a struct - Block8x8, @JimBobSquarePants should I mark/delete them and their tests here or separate PR would be better?

These benchmarks are **not** master vs PR, it's previous PR optimization (stream write fix) vs new specific 420 subsampling optimization.

**[simd]** 4:2:0 Subsampling
|            Method | Quality |     Mean |    Error |   StdDev |
|------------------ |-------- |---------:|---------:|---------:|
| 'ImageSharp Jpeg' |      50 | 31.37 ms | 0.260 ms | 0.217 ms |
| 'ImageSharp Jpeg optimized' |      50 | 27.67 ms | 0.519 ms | 0.533 ms |
| 'ImageSharp Jpeg' |      75 | 34.90 ms | 0.614 ms | 0.656 ms |
| 'ImageSharp Jpeg optimized' |      75 | 31.16 ms | 0.610 ms | 0.653 ms |
| 'ImageSharp Jpeg' |      95 | 44.59 ms | 0.852 ms | 1.078 ms |
| 'ImageSharp Jpeg optimized' |      95 | 40.31 ms | 0.797 ms | 0.745 ms |
| 'ImageSharp Jpeg' |     100 | 48.80 ms | 0.962 ms | 0.900 ms |
| 'ImageSharp Jpeg optimized' |     100 | 45.64 ms | 0.898 ms | 0.998 ms |

**[non-simd]** 4:2:0 Subsampling
|            Method | Quality |     Mean |    Error |   StdDev |
|------------------ |-------- |---------:|---------:|---------:|
| 'ImageSharp Jpeg' |      50 | 36.40 ms | 0.604 ms | 0.565 ms |
| 'ImageSharp Jpeg optimized' |      50 | 34.92 ms | 0.480 ms | 0.425 ms |
| 'ImageSharp Jpeg' |      75 | 38.57 ms | 0.509 ms | 0.425 ms |
| 'ImageSharp Jpeg optimized' |      75 | 37.97 ms | 0.515 ms | 0.430 ms |
| 'ImageSharp Jpeg' |      95 | 48.44 ms | 0.839 ms | 0.785 ms |
| 'ImageSharp Jpeg optimized' |      95 | 46.44 ms | 0.257 ms | 0.201 ms |
| 'ImageSharp Jpeg' |     100 | 53.50 ms | 1.070 ms | 1.464 ms |
| 'ImageSharp Jpeg optimized' |     100 | 50.22 ms | 0.324 ms | 0.270 ms |

antonfirsov:
> There's an unused method and even a struct - Block8x8

`Block8x8` is being used by the decoder:
https://github.com/SixLabors/ImageSharp/blob/97dde7fb2f0ac97b5b1f9e6b0555a53d2e270c46/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanScanDecoder.cs#L179

If you find any methods that are unused by product code, feel free to delete or move them to test code. (Depending whether they are needed for reference implementations or similar test infra stuff.)

tests/ImageSharp.Tests/Formats/Jpg/DCTTests.cs
```diff
@@ -169,7 +288,10 @@ public void TransformFDCT(int seed)
                 var temp1 = new float[64];
                 var temp2 = default(Block8x8F);
 
+                // reference
                 ReferenceImplementations.LLM_FloatingPoint_DCT.FDCT2D_llm(src, expectedDest, temp1, downscaleBy8: true);
+
+                // testee
```
antonfirsov, line 294: There are 3 possible paths in `TransformFDCT` depending on hardware capabilities (Vector4 vs AVX vs AVX+FMA).

To stress the method in a robust manner (including the  selection logic), we need to do out of process test execution with capabilities being selectively disabled using environment variables. We have a tool `FeatureTestRunner` to do this:

https://github.com/SixLabors/ImageSharp/blob/97dde7fb2f0ac97b5b1f9e6b0555a53d2e270c46/tests/ImageSharp.Tests/Common/SimdUtilsTests.Shuffle.cs

Note that it swallows errors and produces false greens in Visual Studio (#1376), but should work on CI.

src/ImageSharp/Formats/Jpeg/Components/FastFloatingPointDCT.cs
```diff
@@ -327,14 +290,226 @@ public static void FDCT8x4_RightPart(ref Block8x8F s, ref Block8x8F d)
...
+            mz2 = SimdUtils.HwIntrinsics.MultiplyAdd(mz4, mz2, C_V_n1_9615);
+            mz3 = SimdUtils.HwIntrinsics.MultiplyAdd(mz4, mz3, C_V_n0_3901);
+            mz0 = Avx.Multiply(mz0, C_V_n0_8999);
+            mz1 = Avx.Multiply(mz1, C_V_n2_5629);
+
+            Vector256<float> mb3 = Avx.Add(SimdUtils.HwIntrinsics.MultiplyAdd(mz0, my7, C_V_0_2986), mz2);
+            Vector256<float> mb2 = Avx.Add(SimdUtils.HwIntrinsics.MultiplyAdd(mz1, my5, C_V_2_0531), mz3);
+            Vector256<float> mb1 = Avx.Add(SimdUtils.HwIntrinsics.MultiplyAdd(mz1, my3, C_V_3_0727), mz2);
+            Vector256<float> mb0 = Avx.Add(SimdUtils.HwIntrinsics.MultiplyAdd(mz0, my1, C_V_1_5013), mz3);
+
```
antonfirsov, line 469: Looks like there are extra newlines here and there. @JimBobSquarePants shouldn't StyleCop complain about this?

br3aker:
> 
> `Block8x8` is being used by the decoder:
> https://github.com/SixLabors/ImageSharp/blob/97dde7fb2f0ac97b5b1f9e6b0555a53d2e270c46/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanScanDecoder.cs#L179

Ups, missed that, thanks. Encoder no longer needs it for code reusage reasons (performance is not affected).

Will delete unused method & its tests in this PR then.

tests/ImageSharp.Tests/Formats/Jpg/DCTTests.cs
br3aker: Same to IDCT then, will try to wrap my head around `FeatureTestRunner` today.

br3aker: And while we are here, there's two tests which test IDCT transform with `LLM_FloatingPoint_DCT.TransformIDCT` and another one with `AccurateDCT.TransformIDCT`. Pre-pr tests used for LLM implementation for testing IDCT parts, I used same reference for new methods. So there are some questions:

1. Should I use LLM of Accurate for tests? Both?
2. If only one reference is used, should the other one exists in test code at all?
3. If both used, should both asserts be checks at the same test method or 2 different methods per 1 test?

src/ImageSharp/Formats/Jpeg/Components/FastFloatingPointDCT.cs
br3aker: It does complain, my oversight, will fix asap!

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
```diff
@@ -94,27 +130,130 @@ public static void Convert(ReadOnlySpan<Rgb24> rgbSpan, ref Block8x8F yBlock, re
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Vector256<float> Scale_8x4_4x2(Span<Vector256<float>> v)
+        {
+            Vector256<int> switchInnerDoubleWords = Unsafe.As<byte, Vector256<int>>(ref MemoryMarshal.GetReference(SimdUtils.HwIntrinsics.PermuteMaskSwitchInnerDWords8x32));
+            var f025 = Vector256.Create(0.25f);
+
+            Vector256<float> topPairSum = SumHorizontalPairs(v[0], v[2]);
+            Vector256<float> botPairSum = SumHorizontalPairs(v[1], v[3]);
+
+            return Avx2.PermuteVar8x32(Avx.Multiply(SumVerticalPairs(topPairSum, botPairSum), f025), switchInnerDoubleWords);
```
saucecontrol, line 247: Looks like this is refactored from a previous PR I missed, but the SIMD here is a bit wacky.  If I'm reading this right, it does something like this:

```
v0 = 0 | 1 | 2 | 3 || 4 | 5 | 6 | 7
v1 = 8 | 9 | a | b || c | d | e | f
v2 = g | h | i | j || k | l | m | n
v3 = o | p | q | r || s | t | u | v

vs0 = vshufps v0,v2, mask 0,2,0,2
vs1 = vshufps v1,v3, mask 0,2,0,2
vs2 = vshufps v0,v2, mask 1,3,1,3
vs3 = vshufps v1,v3, mask 1,3,1,3

vs0 = 0 | 2 | g | i || 4 | 6 | k | m
vs1 = 8 | a | o | q || c | e | s | u
vs2 = 1 | 3 | h | j || 5 | 7 | l | n
vs3 = 9 | b | p | r || d | f | t | v

v0 = vaddps vs0,vs2
v1 = vaddps vs1,vs3

v0 = 0+1|2+3|g+h|i+j||4+5|6+7|k+l|m+n
v1 = 8+9|a+b|o+p|q+r||c+d|e+f|s+t|u+v

vs0 = vshufps v0,v1 mask 0,1,0,1
vs1 = vshufps v0,v1 mask 2,3,2,3

vs0 = 0+1|2+3|8+9|a+b||4+5|6+7|c+d|e+f
vs1 = g+h|i+j|o+p|q+r||k+l|m+n|s+t|u+v

v0 = vaddps vs0,vs1
v0 = vmulps v0, 0.25f

v0 = 01gh|23ij|89op|abqr||45kl|67mn|cdst|efuv

v0 = vpermd v0, mask 0,1,4,5,2,3,6,7

v0 = 01gh|23ij|45kl|67mn||89op|abqr|cdst|efuv
```

Which could be dramatically simplified by using horizontal adds, like such

```
v0 = 0 | 1 | 2 | 3 || 4 | 5 | 6 | 7 
v1 = 8 | 9 | a | b || c | d | e | f
v2 = g | h | i | j || k | l | m | n
v3 = o | p | q | r || s | t | u | v

v0 = vhaddps v0,v1
v1 = vhaddps v2,v3

v0 = 0+1|2+3|8+9|a+b||4+5|6+7|c+d|e+f
v1 = g+h|i+j|o+p|q+r||k+l|m+n|s+t|u+v

v0 = vaddps v0,v1
v0 = vmulps v0, 0.25f

v0 = 01gh|23ij|89op|abqr||45kl|67mn|cdst|efuv

v0 = vpermq v0, mask 0,2,1,3

v0 = 01gh|23ij|45kl|67mn||89op|abqr|cdop|efqr
```

Note that `vpermq` (Avx2.Permute4x64) works here because float pairs are kept together, and that instruction takes an immediate byte mask rather than a vector mask.

Usual disclaimer I haven't tested this, but it should be close if not completely correct.

saucecontrol (commented):
Really nice improvements!  Left a comment on the SIMD 2:1 box filter if you want to push it a bit further

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbToYCbCrConverterVectorized.cs
br3aker: Completely forgot about horizontal adds, thanks a lot! Will try it today with benchmarks.

br3aker:
@saucecontrol did a little research about 8x4 -> 4x2 downscaling based on your ideas.

1) A lot of total method time was spent on the final permutation. I guess line crossing was a thing, your approach with `vpermq` works 25% faster:

|              Method |     Mean |     Error |    StdDev | Ratio |
|-------------------- |---------:|----------:|----------:|------:|
|         Shuffle_vec | 1.962 ns | 0.0146 ns | 0.0136 ns |  1.00 |
|       Shuffle_qword | 1.485 ns | 0.0124 ns | 0.0110 ns |  0.76 |

Note that this benchmark uses current `add(shuffle, shuffle)` implementation, not horizontal add.

---------------------------------------------

2) `vhaddps` doesn't really give any performance boost:

|              Method |     Mean |     Error |    StdDev | Ratio |
|-------------------- |---------:|----------:|----------:|------:|
|         Shuffle_vec | 1.962 ns | 0.0146 ns | 0.0136 ns |  1.00 |
|   HorizontalAdd_vec | 1.863 ns | 0.0148 ns | 0.0131 ns |  0.95 |
|       Shuffle_qword | 1.485 ns | 0.0124 ns | 0.0110 ns |  0.76 |
| HorizontalAdd_qword | 1.497 ns | 0.0114 ns | 0.0101 ns |  0.76 |

`vpermq` gives a little to none performance boost **but** it reduces code by a lot, it completely eliminates 2 custom methods and looks cleaner!

---------------------------------------------

3) Horizontal adds are slower than usual adds as far as I know. Your implementation uses 2 hadd & 1 add, we can change it to 1 hadd and 2 add:

```
//v0 = 0 | 1 | 2 | 3 || 4 | 5 | 6 | 7
//v1 = 8 | 9 | a | b || c | d | e | f
//v2 = g | h | i | j || k | l | m | n
//v3 = o | p | q | r || s | t | u | v

v0 = vaddps v0 ,v2
v1 = vaddps v1, v3
//v0 = 0g | 1h | 2i | 3j || 4k | 5l | 6m | 7n
//v1 = 8o | 9p | aq | br || cs | dt | eu | fv

v0 = vhaddps v0, v1
v0 = vmulps v0, 0.25f
//v0 = 01gh|23ij|89op|abqr||45kl|67mn|cdst|efuv

vpermq v0, mask 0,2,1,3
//per = 01gh|23ij|45kl|67mn||89op|abqr|cdst|efuv
```

And it runs almost **50% faster**:

|              Method |     Mean |     Error |    StdDev | Ratio |
|-------------------- |---------:|----------:|----------:|------:|
|         Current | 1.962 ns | 0.0146 ns | 0.0136 ns |  1.00 |
| HorizontalAdd_qword | 1.497 ns | 0.0114 ns | 0.0101 ns |  0.76 |
|       brand_new | 1.018 ns | 0.0151 ns | 0.0134 ns |  0.52 |

Will push this change & test coverage today.

saucecontrol:
That's great!  I had the same thought about the horizontal adds late last night but was too lazy to go back upstairs to my computer 😆 

Horizontal add is implemented in microcode in the processor, and it's effectively broken down into 2 shifts (one of which doesn't map to an instruction exposed publicly) and an add.  So it's actually only one cycle faster than the current code that does 3 shifts and an add.  But yeah, doing the vertical portion first really cleans it up.

And hopefully the smaller code makes it possible to get rid of the code in the outer method that spills those to the stack.  (I didn't spend enough time looking at that to figure out what it's doing entirely).

Oh, and you don't need `Unsafe.As` to cast between vector types.  `avg2x2.AsDouble()` is the same as `Unsafe.As<Vector256<float>, Vector256<double>>(ref avg2x2)`.  All the `AsXXX` methods on the vector types are intrinsics that compile to a no-op.

br3aker:
Yep, code is **much** cleaner and smaller with your additions. Didn't know about no-op for that casts.

Thank you for your contribution to this pr, getting smarter everyday! :smile:

br3aker:
@JimBobSquarePants waiting for new commits to be built & checked and we are ready to merge I guess!
If fresh pre/post full encoding benchmarks are needed - will gladly execute and provide them tomorrow.

JimBobSquarePants:
@br3aker Looks like you've got a `PlatformNotSupportedException` around `System.Runtime.Intrinsics.X86.Avx2.Permute4x64` on macOS .NET 5.

https://github.com/SixLabors/ImageSharp/pull/1632/checks?check_run_id=2789197791#step:10:256

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -577,6 +577,24 @@ public static class HwIntrinsics
                 }
             }
 
+            /// <summary>
+            /// Scales 8x8 matrix to 4x2 using 2x2 average
+            /// </summary>
+            /// <param name="v">Input matrix consisting of 4 256bit vectors, first row: (v[0], v[2]), second row: (v[1], v[3])</param>
+            /// <returns>256bit vector containing upper and lower scaled parts of the input matrix</returns>
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static Vector256<float> Scale16x2_8x1(ReadOnlySpan<Vector256<float>> v)
```
JimBobSquarePants, line 586: Does this method have a use case outside of the jpeg encoder? If not, it shouldn't be in this class. `SimdUtils` is supposed to be for reusable utility methods. 

Anything public should also have explicit `IsSupported` checks.

br3aker: It's not, moved to jpeg pixel converter. Added `Debug` check for Avx2.

br3aker:
@JimBobSquarePants my bad, did check for `Avx`, not `Avx2`. Pushed fixes for this and your comment about `SimdUtils` class.

tests/ImageSharp.Benchmarks/Codecs/Jpeg/EncodeJpeg.cs
```diff
@@ -12,23 +13,44 @@ namespace SixLabors.ImageSharp.Benchmarks.Codecs.Jpeg
...
                 this.bmpStream = File.OpenRead(Path.Combine(TestEnvironment.InputImagesDirectoryFullPath, TestImage));
+
                 this.bmpCore = Image.Load<Rgba32>(this.bmpStream);
                 this.bmpCore.Metadata.ExifProfile = null;
+                this.encoder = new JpegEncoder { Quality = Quality, Subsample = EncodingSubsampling };
+
                 this.bmpStream.Position = 0;
                 this.bmpDrawing = SDImage.FromStream(this.bmpStream);
+                this.jpegCodec = GetEncoder(ImageFormat.Jpeg);
+                this.encoderParameters = new EncoderParameters(1);
```
JimBobSquarePants, line 50: Note. `EncoderParameters` implement `IDisposable` but the GC should clean up eventually here.

JimBobSquarePants (approved):
I can't see any issues here. @br3aker really, really great work!! 

I'll leave it one more day so others can have a final look though before merging though s this is tricky stuff and I might have missed something.

tests/ImageSharp.Benchmarks/Codecs/Jpeg/EncodeJpeg.cs
br3aker: Shouldn't be hard to fix. And I completely forgot to update benchmarks in the comments of that file. Plus It's better to implement full set of benchmarks for: both subsampling methods (skip 444 for System.Drawing) and a set of quality parameters, something like 75, 90, 100 for a full picture.

JimBobSquarePants: If you have time that would be great!

br3aker: Done!

br3aker:
@JimBobSquarePants fixed disposable thingy in benchmarks & updated them:

```
.NET Core SDK=6.0.100-preview.3.21202.5
  [Host]     : .NET Core 3.1.13 (CoreCLR 4.700.21.11102, CoreFX 4.700.21.11602), X64 RyuJIT  [AttachedDebugger]
  DefaultJob : .NET Core 3.1.13 (CoreCLR 4.700.21.11102, CoreFX 4.700.21.11602), X64 RyuJIT
```

|                      Method | Quality |     Mean |    Error |   StdDev | Ratio | RatioSD |
|---------------------------- |-------- |---------:|---------:|---------:|------:|--------:|
| 'System.Drawing Jpeg 4:2:0' |      75 | 30.60 ms | 0.496 ms | 0.464 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg 4:2:0' |      75 | 29.86 ms | 0.350 ms | 0.311 ms |  0.98 |    0.02 |
|     'ImageSharp Jpeg 4:4:4' |      75 | 45.36 ms | 0.899 ms | 1.036 ms |  1.48 |    0.05 |
|                             |         |          |          |          |       |         |
| 'System.Drawing Jpeg 4:2:0' |      90 | 34.05 ms | 0.669 ms | 0.687 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg 4:2:0' |      90 | 37.26 ms | 0.706 ms | 0.660 ms |  1.10 |    0.03 |
|     'ImageSharp Jpeg 4:4:4' |      90 | 52.54 ms | 0.579 ms | 0.514 ms |  1.55 |    0.04 |
|                             |         |          |          |          |       |         |
| 'System.Drawing Jpeg 4:2:0' |     100 | 39.36 ms | 0.267 ms | 0.237 ms |  1.00 |    0.00 |
|     'ImageSharp Jpeg 4:2:0' |     100 | 42.44 ms | 0.410 ms | 0.383 ms |  1.08 |    0.01 |
|     'ImageSharp Jpeg 4:4:4' |     100 | 70.88 ms | 0.508 ms | 0.450 ms |  1.80 |    0.02 |

br3aker:
Waaaat, first time ever seeing `OutOfMemory` exception in tests in this PR for... decoding?

antonfirsov:
Looks like the ImageMagick reference decoder is also very greedy, don't mind it, nothing to do with your code.

antonfirsov (approved):
LGTM, nice work!

JimBobSquarePants:
Lets get this merged!! 🎉

PR #1641, opened 2021-05-28
AdaptiveHistogramEqualizationProcessor. Fix for large images

JimBobSquarePants:
### Prerequisites

### Description

Fixes the issue identified in #1640 

Ensures that the `cdfLutSpan` buffer is cleared before each use. I also did a couple of code style cleanup bits to remove squigglies and an indexer check removal.

I spotted areas where we could improve performance but did not touch them at this time.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
@@ -619,9 +619,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
                                 for (int dx = x; dx < xlimit; dx++)
                                 {
                                     int luminance = GetLuminance(rowSpan[dx], this.luminanceLevels);
-
-                                    // This is safe. The index maxes out to the span length.
-                                    Unsafe.Add(ref histogramBase, luminance)++;
+                                    histogram[luminance]++;
                                 }
                             }
 
--- a/src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
@@ -459,10 +459,14 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
             private readonly Configuration configuration;
             private readonly MemoryAllocator memoryAllocator;
 
-            // Used for storing the minimum value for each CDF entry.
+            /// <summary>
+            /// Used for storing the minimum value for each CDF entry.
+            /// </summary>
             private readonly Buffer2D<int> cdfMinBuffer2D;
 
-            // Used for storing the LUT for each CDF entry.
+            /// <summary>
+            /// Used for storing the LUT for each CDF entry.
+            /// </summary>
             private readonly Buffer2D<int> cdfLutBuffer2D;
             private readonly int pixelsInTile;
             private readonly int sourceWidth;
@@ -596,6 +600,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Normalization
                         int y = this.tileYStartPositions[index].y;
                         int endY = Math.Min(y + this.tileHeight, this.sourceHeight);
                         Span<int> cdfMinSpan = this.cdfMinBuffer2D.GetRowSpan(cdfY);
+                        cdfMinSpan.Clear();
 
                         using IMemoryOwner<int> histogramBuffer = this.allocator.Allocate<int>(this.luminanceLevels);
                         Span<int> histogram = histogramBuffer.GetSpan();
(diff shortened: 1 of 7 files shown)


tests/ImageSharp.Tests/Processing/Normalization/HistogramEqualizationTests.cs
```diff
@@ -162,5 +163,40 @@ public void Issue984<TPixel>(TestImageProvider<TPixel> provider)
...
+                var options = new HistogramEqualizationOptions
+                {
+                    Method = HistogramEqualizationMethod.AdaptiveTileInterpolation,
+                    LuminanceLevels = 4096,
+                    ClipHistogram = false,
+                    ClipLimit = 350,
+                    NumberOfTiles = 8
+                };
+
+                Image<TPixel> processed = image.Clone(ctx =>
```
brianpopow, line 191: missing using here

brianpopow (approved):
looks good 👍

tests/ImageSharp.Tests/Processing/Normalization/HistogramEqualizationTests.cs
```diff
@@ -162,5 +163,40 @@ public void Issue984<TPixel>(TestImageProvider<TPixel> provider)
                 image.CompareToReferenceOutput(ValidatorComparer, provider);
             }
         }
+
+        [Theory]
+        [WithTestPatternImages(5120, 9234, PixelTypes.L16)]
+        public unsafe void Issue1640<TPixel>(TestImageProvider<TPixel> provider)
```
antonfirsov, line 169: How long does this run?
Instead of adding a 0.5MB reference image, I would prefer to create the test input dynamically, and would also make sure the test is isolated by using an custom `MemoryAllocator`.  Will post follow up.

antonfirsov:
In case the test runs for more than 20 seconds I would delete it. This is not the processor that is worth the extra cost IMO.

In case if it's fast enough,  an you try if the following change keeps reproducing the issue (after removing the fix):

```C#
provider.Configuration = Configuration.CreateDefault(); // isolate `MemoryAllocator` usage

using Image<TPixel> referenceResult = provider.GetImage().Clone( ctx => {                    
    ctx.HistogramEqualization(options);
    ctx.Resize(image.Width / 4, image.Height / 4, KnownResamplers.Bicubic)
});

using Image<TPixel> image = provider.GetImage();

// ...

// validation:
ValidatorComparer.VerifySimilarity(referenceResult, processed);
```

tests/ImageSharp.Tests/Processing/Normalization/HistogramEqualizationTests.cs
JimBobSquarePants: That’ll teach me to copy/paste! 🤪

tests/ImageSharp.Tests/Processing/Normalization/HistogramEqualizationTests.cs
JimBobSquarePants: About 7 seconds on my machine. Wasn't concerned by the size since it's all LFS now but if you fancy working out a better approach go for it.

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationProcessor{TPixel}.cs
```diff
@@ -614,7 +619,9 @@ public void Invoke(in RowInterval rows)
                                 for (int dx = x; dx < xlimit; dx++)
                                 {
                                     int luminance = GetLuminance(rowSpan[dx], this.luminanceLevels);
-                                    histogram[luminance]++;
+
+                                    // This is safe. The index maxes out to the span length.
+                                    Unsafe.Add(ref histogramBase, luminance)++;
```
antonfirsov, line 624: Feel free to ignore, but I would prefer to stick with span indexers systematically in less critical processors. Span indexer is quite fast these days and `GetLuminance` is way more expensive than the indexer.

JimBobSquarePants: Will revert. Should be replaced anyway with a span based version.

JimBobSquarePants:
>provider.Configuration = Configuration.CreateDefault(); // isolate `MemoryAllocator` usage

@antonfirsov Wouldn't this create an entirely new set up pooled buffers. That's quite expensive no?

tests/ImageSharp.Tests/Processing/Normalization/HistogramEqualizationTests.cs
antonfirsov: Undecided if it's worth the extra time spent on minimizing the sizes or not. I'm a bit afraid that with growing test suite it will be a trouble in the future even with LFS, but maybe it's a false fear.

antonfirsov:
> Wouldn't this create an entirely new set up pooled buffers. That's quite expensive no?

Not significantly. It's equivalent of using non-pooled large arrays, which should be fine to do extensively in the tests. GC should clean up on 64bit IMO.

tests/ImageSharp.Tests/Processing/Normalization/HistogramEqualizationTests.cs
JimBobSquarePants: Removed, using your test now.

PR #1644, opened 2021-05-28
Stop using timeouts to test cancellation token compliance.

tocsoft:
### Prerequisites

### Description

This removes the use of large images and hoping that decoding/encoding is slow to test cancellation token compliance. Will make these tests reliable. 

We now have a `PausedStream` class that will hang all its API calls until it is `Release()`ed. `PausedStream` also has a task we can await for when there is at least one caller waiting the stream to unlock.


tests/ImageSharp.Tests/TestUtilities/PausedStream.cs
```diff
@@ -0,0 +1,145 @@
...
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace SixLabors.ImageSharp.Tests.TestUtilities
+{
+    public class PausedStream : Stream
+    {
+        private readonly SemaphoreSlim slim = new SemaphoreSlim(0);
```
antonfirsov, line 13: Nit: I would  name it `semaphore`.

tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.cs
```diff
@@ -127,60 +128,52 @@ public async Task DecodeAsync_DegenerateMemoryRequest_ShouldTranslateTo_ImageFor
...
-                {
-                    using Image image = await Image.LoadAsync(hugeFile, cts.Token);
-                }
-                catch (TaskCanceledException)
-                {
-                    // Successfully observed a cancellation
-                    return;
-                }
-            }
+            AsyncLocalSwitchableFilesystem.ConfigureFileSystemStream(pausedStream);
```
antonfirsov, line 153: Since xunit tests are running in parallel, we should not modify any global state, but rather pass a new `Configuration` instance to `Image.LoadAsync`. It can be created by `Configuration.Default.DeepClone()`.

tests/ImageSharp.Tests/TestUtilities/AsyncLocalSwitchableFilesystem.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System.IO;
+using System.Threading;
+using SixLabors.ImageSharp.IO;
+
+namespace SixLabors.ImageSharp.Tests.TestUtilities
+{
+    public class AsyncLocalSwitchableFilesystem : IFileSystem
```
antonfirsov, line 10: By altering local `Configuration` only, this becomes unnecessary, we'll need `SingleStreamFileSystem` only.

tocsoft: That means we don't test the code that make sue of the global configuration however

tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.cs
tocsoft: That means we don't actually test our usage of the global context here.... This is exactly why I created this specific version of `IFileStream` that proxies to the same time as default global version in all cases unless in an async context where its specifically been overridden... this should be perfectly safe along as this is the only mechanism used to switch out the global file system.

tocsoft: hmm...actually changed my mind, will switch over to the overload that take a configuration.

tests/ImageSharp.Tests/TestUtilities/AsyncLocalSwitchableFilesystem.cs
tocsoft: makes sense (changed my mind based on what we are actually trying to test)

antonfirsov: I'd manage this as following:
1. We need a test that somehow checks if the Default configuration is forwarded in all the parameterless `Load` can be  be independent of the individual methods. This is likely a gap, but can be easily fixed by adding `Assert.Same(Configuration.Default, img.GetConfiguration());` to the following method:
https://github.com/SixLabors/ImageSharp/blob/841f8965bedc349d2cf0b8217451d27d054b7ca7/tests/ImageSharp.Tests/Image/ImageTests.Load_FileSystemPath_UseDefaultConfiguration.cs#L20-L23
2. We can test individual features with local `Configuration` when necessary

Otherwise we would have to move the test case into a separate class and do magic with `[CollectionDefinition(DisableParallelization = true)]` (then reset the state change), or use `RemoteExecutor` to separate at process level. I think sticking with isolated local state fits Xunit best practices better.

I'm actually surprised that other (independent) tests are passing while the shared configuration is using `SingleStreamFileSystem`. Maybe it's because `TestImageProvider` also uses [an isolated configuration](https://github.com/SixLabors/ImageSharp/blob/97dde7fb2f0ac97b5b1f9e6b0555a53d2e270c46/tests/ImageSharp.Tests/TestUtilities/ImageProviders/TestImageProvider.cs#L28) and there is a very low amount of tests that will call the shared Configuration's filesystem.

tocsoft: not that I matters anymore as I've switched to local configuration, but the reason the other tests where working fine is becasue they where not using the `SingleStreamFileSystem` the singleton `AsyncLocalSwitchableFilesystem` used an `AsyncLocal<T>`  to store the the `SingleStreamFileSytem` and as that was only being overridden in the test(s) the called `ConfigureFileSystemStream()` then all other tests where falling back to proxying to an `LocalFileSytem` which is our default version.

antonfirsov (approved):
LGTM, thanks for fixing this reliably!

Sidenote: now we don't test the case when cancellation happens during Jpeg color conversion (stream is entirely eaten by that point), but #1597 will automatically fix that.

JimBobSquarePants:
@tocsoft This is excellent! 💯

PR #1646, opened 2021-05-31
Make sure encoding 4bit paletted tiff rows are byte aligned

brianpopow:
### Prerequisites

### Description

This **PR** makes sure to write padding when encoding 4-bit paletted tiff's with uneven width.

### DIFF
--- a/src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter{TPixel}.cs
+++ b/src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter{TPixel}.cs
@@ -60,7 +60,7 @@ namespace SixLabors.ImageSharp.Formats.Tiff.Writers
             {
                 int width = this.Image.Width;
                 int excess = (width % 2) * height;
-                int rows4BitBufferLength = indexedPixels.Length + excess;
+                int rows4BitBufferLength = (width / 2 * height) + excess;
                 using IMemoryOwner<byte> rows4bitBuffer = this.MemoryAllocator.Allocate<byte>(rows4BitBufferLength);
                 Span<byte> rows4bit = rows4bitBuffer.GetSpan();
                 int idxPixels = 0;
--- a/src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter{TPixel}.cs
+++ b/src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter{TPixel}.cs
@@ -55,23 +55,38 @@ namespace SixLabors.ImageSharp.Formats.Tiff.Writers
         /// <inheritdoc />
         protected override void EncodeStrip(int y, int height, TiffBaseCompressor compressor)
         {
-            Span<byte> pixels = GetStripPixels(((IPixelSource)this.quantizedImage).PixelBuffer, y, height);
+            Span<byte> indexedPixels = GetStripPixels(((IPixelSource)this.quantizedImage).PixelBuffer, y, height);
             if (this.BitsPerPixel == 4)
             {
-                using IMemoryOwner<byte> rows4bitBuffer = this.MemoryAllocator.Allocate<byte>(pixels.Length / 2);
+                int width = this.Image.Width;
+                int excess = (width % 2) * height;
+                int rows4BitBufferLength = indexedPixels.Length + excess;
+                using IMemoryOwner<byte> rows4bitBuffer = this.MemoryAllocator.Allocate<byte>(rows4BitBufferLength);
                 Span<byte> rows4bit = rows4bitBuffer.GetSpan();
-                int idx = 0;
-                for (int i = 0; i < rows4bit.Length; i++)
+                int idxPixels = 0;
+                int idx4bitRows = 0;
+                int halfWidth = width / 2;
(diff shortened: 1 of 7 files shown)


src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter{TPixel}.cs
```diff
@@ -55,23 +55,38 @@ internal sealed class TiffPaletteWriter<TPixel> : TiffBaseColorWriter<TPixel>
...
         protected override void EncodeStrip(int y, int height, TiffBaseCompressor compressor)
         {
-            Span<byte> pixels = GetStripPixels(((IPixelSource)this.quantizedImage).PixelBuffer, y, height);
+            Span<byte> indexedPixels = GetStripPixels(((IPixelSource)this.quantizedImage).PixelBuffer, y, height);
             if (this.BitsPerPixel == 4)
             {
-                using IMemoryOwner<byte> rows4bitBuffer = this.MemoryAllocator.Allocate<byte>(pixels.Length / 2);
+                int width = this.Image.Width;
+                int excess = (width % 2) * height;
+                int rows4BitBufferLength = indexedPixels.Length + excess;
```
IldarKhayrutdinov, line 63: it seems that here you can:
`int rows4BitBufferLength = (indexedPixels.Length / 2) + excess;`

brianpopow: yeah your right, this was a bit too much

JimBobSquarePants (approved):

src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter{TPixel}.cs
```diff
@@ -55,23 +55,38 @@ internal sealed class TiffPaletteWriter<TPixel> : TiffBaseColorWriter<TPixel>
...
+                int width = this.Image.Width;
+                int excess = (width % 2) * height;
+                int rows4BitBufferLength = (width / 2 * height) + excess;
+                using IMemoryOwner<byte> rows4bitBuffer = this.MemoryAllocator.Allocate<byte>(rows4BitBufferLength);
                 Span<byte> rows4bit = rows4bitBuffer.GetSpan();
-                int idx = 0;
-                for (int i = 0; i < rows4bit.Length; i++)
+                int idxPixels = 0;
+                int idx4bitRows = 0;
+                int halfWidth = width / 2;
```
gfoidl, line 68:
```suggestion
                int halfWidth = width >> 1;
```
And move it before L63, so this variable can be re-used in `rows4BitBufferLength`.

Note: instead of the bit-operations (here and comment above) you could use unsigned-arthmetics, as there the JIT does these "tricks" on your behalf.

antonfirsov: Wow, I'm surprised JIT still can't optimize this .

@brianpopow in case you change these, I would recommend a form like
```C#
int excess = (width & 1) * height; // (width % 2) * height
```
to keep the intention clear.

gfoidl: > I'm surprised JIT still can't optimize this .

Please don't blame the JIT here. It's signed arithmetics that need to work for negative values too, and the JIT can't conclude that width is positive. So one could use `uint` to give the JIT the necessary hint(s) and produce bit-operations, or do it manually.

GCC or Clang can't handle this too (because of signed arithmetics).

src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter{TPixel}.cs
```diff
@@ -55,23 +55,38 @@ internal sealed class TiffPaletteWriter<TPixel> : TiffBaseColorWriter<TPixel>
...
         protected override void EncodeStrip(int y, int height, TiffBaseCompressor compressor)
         {
-            Span<byte> pixels = GetStripPixels(((IPixelSource)this.quantizedImage).PixelBuffer, y, height);
+            Span<byte> indexedPixels = GetStripPixels(((IPixelSource)this.quantizedImage).PixelBuffer, y, height);
             if (this.BitsPerPixel == 4)
             {
-                using IMemoryOwner<byte> rows4bitBuffer = this.MemoryAllocator.Allocate<byte>(pixels.Length / 2);
+                int width = this.Image.Width;
+                int excess = (width % 2) * height;
+                int rows4BitBufferLength = (width / 2 * height) + excess;
```
IldarKhayrutdinov:
```suggestion
                int rows4BitBufferLength = ((width >> 1 ) + (width & 1)) * height; 
```

src/ImageSharp/Formats/Tiff/Writers/TiffPaletteWriter{TPixel}.cs
antonfirsov: > It's signed arithmetics that need to work for negative values too, and the JIT can't conclude that width is positive. 

Missed that aspect, makes sense, thanks!

PR #1647, opened 2021-06-01
Add support for decoding 6, 12, 30, 42 bits per pixel tiff's

brianpopow:
### Prerequisites

### Description

This **PR** adds support for decoding:
- 12 bit (4 bits per channel, RGB chunky) TIFF images.
- 6 bit (2 bits per channel, RGB chunky) TIFF images.
- 30 bit (10 bits per channel, RGB chunky) TIFF images.
- 42 bit (14 bits per channel, RGB chunky) TIFF images.

/cc @IldarKhayrutdinov

### DIFF
--- a/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/TiffColorDecoderFactory{TPixel}.cs
+++ b/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/TiffColorDecoderFactory{TPixel}.cs
@@ -60,9 +60,9 @@ namespace SixLabors.ImageSharp.Formats.Tiff.PhotometricInterpretation
                 case TiffColorType.Rgb222:
                     DebugGuard.IsTrue(
                         bitsPerSample.Length == 3
-                        && bitsPerSample[0] == 2
+                        && bitsPerSample[2] == 2
                         && bitsPerSample[1] == 2
-                        && bitsPerSample[2] == 2,
+                        && bitsPerSample[0] == 2,
                         "bitsPerSample");
                     DebugGuard.IsTrue(colorMap == null, "colorMap");
                     return new RgbTiffColor<TPixel>(bitsPerSample);
@@ -70,9 +70,9 @@ namespace SixLabors.ImageSharp.Formats.Tiff.PhotometricInterpretation
                 case TiffColorType.Rgb444:
                     DebugGuard.IsTrue(
                         bitsPerSample.Length == 3
-                        && bitsPerSample[0] == 4
+                        && bitsPerSample[2] == 4
                         && bitsPerSample[1] == 4
-                        && bitsPerSample[2] == 4,
+                        && bitsPerSample[0] == 4,
                         "bitsPerSample");
                     DebugGuard.IsTrue(colorMap == null, "colorMap");
                     return new Rgb444TiffColor<TPixel>();
@@ -80,9 +80,9 @@ namespace SixLabors.ImageSharp.Formats.Tiff.PhotometricInterpretation
                 case TiffColorType.Rgb888:
                     DebugGuard.IsTrue(
                         bitsPerSample.Length == 3
-                        && bitsPerSample[0] == 8
+                        && bitsPerSample[2] == 8
                         && bitsPerSample[1] == 8
-                        && bitsPerSample[2] == 8,
+                        && bitsPerSample[0] == 8,
                         "bitsPerSample");
                     DebugGuard.IsTrue(colorMap == null, "colorMap");
                     return new Rgb888TiffColor<TPixel>();
@@ -90,9 +90,9 @@ namespace SixLabors.ImageSharp.Formats.Tiff.PhotometricInterpretation
                 case TiffColorType.Rgb101010:
                     DebugGuard.IsTrue(
                         bitsPerSample.Length == 3
-                        && bitsPerSample[0] == 10
+                        && bitsPerSample[2] == 10
                         && bitsPerSample[1] == 10
-                        && bitsPerSample[2] == 10,
+                        && bitsPerSample[0] == 10,
                         "bitsPerSample");
                     DebugGuard.IsTrue(colorMap == null, "colorMap");
                     return new RgbTiffColor<TPixel>(bitsPerSample);
@@ -100,9 +100,9 @@ namespace SixLabors.ImageSharp.Formats.Tiff.PhotometricInterpretation
                 case TiffColorType.Rgb141414:
                     DebugGuard.IsTrue(
                         bitsPerSample.Length == 3
-                        && bitsPerSample[0] == 14
+                        && bitsPerSample[2] == 14
                         && bitsPerSample[1] == 14
-                        && bitsPerSample[2] == 14,
+                        && bitsPerSample[0] == 14,
                         "bitsPerSample");
                     DebugGuard.IsTrue(colorMap == null, "colorMap");
                     return new RgbTiffColor<TPixel>(bitsPerSample);
--- a/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/TiffColorDecoderFactory{TPixel}.cs
+++ b/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/TiffColorDecoderFactory{TPixel}.cs
@@ -97,6 +97,16 @@ namespace SixLabors.ImageSharp.Formats.Tiff.PhotometricInterpretation
                     DebugGuard.IsTrue(colorMap == null, "colorMap");
                     return new RgbTiffColor<TPixel>(bitsPerSample);
 
+                case TiffColorType.Rgb141414:
+                    DebugGuard.IsTrue(
+                        bitsPerSample.Length == 3
+                        && bitsPerSample[0] == 14
+                        && bitsPerSample[1] == 14
+                        && bitsPerSample[2] == 14,
+                        "bitsPerSample");
+                    DebugGuard.IsTrue(colorMap == null, "colorMap");
+                    return new RgbTiffColor<TPixel>(bitsPerSample);
+
                 case TiffColorType.PaletteColor:
                     DebugGuard.NotNull(bitsPerSample, "bitsPerSample");
                     DebugGuard.NotNull(colorMap, "colorMap");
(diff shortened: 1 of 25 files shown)


src/ImageSharp/Formats/Tiff/TiffEncoderEntriesCollector.cs
```diff
@@ -274,6 +278,7 @@ public void Process(TiffEncoderCore encoder)
                     Value = (ushort)encoder.PhotometricInterpretation
                 };
 
+                this.collector.AddOrReplace(planarConfig);
```
brianpopow, line 281: note that I have explicitly added the chunky planar configuration, otherwise it could have accidentally be possible that it would be set to planar, if the decoded images was planar.

IldarKhayrutdinov:
The patch is workaround support for `RGB {2,2,2}` using [RgbTiffColor<TPixel>](https://github.com/SixLabors/ImageSharp/blob/bp/tiff12bit/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/RgbTiffColor%7BTPixel%7D.cs): [custom_rgb.zip](https://github.com/SixLabors/ImageSharp/files/6583058/custom_rgb.zip)

Test images: 
[tiger-rgb-float-strip-contig-16.tif](https://github.com/BitMiracle/libtiff.net/blob/master/TestCase/tiger-rgb-float-strip-contig-16.tif)
[tiger-rgb-strip-contig-03.tif](https://github.com/BitMiracle/libtiff.net/blob/master/TestCase/tiger-rgb-strip-contig-03.tif)
[tiger-rgb-strip-contig-02.tif](https://github.com/BitMiracle/libtiff.net/blob/master/TestCase/tiger-rgb-strip-contig-02.tif)

brianpopow:
> The patch is workaround support for `RGB {2,2,2}` using [RgbTiffColor](https://github.com/SixLabors/ImageSharp/blob/bp/tiff12bit/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/RgbTiffColor%7BTPixel%7D.cs):
> [custom_rgb.zip](https://github.com/SixLabors/ImageSharp/files/6583058/custom_rgb.zip)
> 
> Test images:
> [tiger-rgb-float-strip-contig-16.tif](https://github.com/BitMiracle/libtiff.net/blob/master/TestCase/tiger-rgb-float-strip-contig-16.tif)
> [tiger-rgb-strip-contig-03.tif](https://github.com/BitMiracle/libtiff.net/blob/master/TestCase/tiger-rgb-strip-contig-03.tif)
> [tiger-rgb-strip-contig-02.tif](https://github.com/BitMiracle/libtiff.net/blob/master/TestCase/tiger-rgb-strip-contig-02.tif)

yeah you a right, we can do it this way, but I think this should be another PR. Feel free to open another PR with it.

brianpopow:
@dlemstra could you please double check, if the usage of ImageMagick for 4 bit and 2 bit is correct here: bc723d30

IldarKhayrutdinov:
> we can do it this way, but I think this should be another PR. Feel free to open another PR with it.

yeah, this PR nothing breaks, it should be another PR

src/ImageSharp/Formats/Tiff/TiffBitsPerSampleExtensions.cs
```diff
@@ -48,6 +50,13 @@ public static TiffBitsPerSample GetBitsPerSample(this ushort[] bitsPerSample)
                         return TiffBitsPerSample.Bit24;
                     }
 
+                    if (bitsPerSample[0] == TiffConstants.BitsPerSampleRgb4Bit[0] &&
+                        bitsPerSample[1] == TiffConstants.BitsPerSampleRgb4Bit[1] &&
+                        bitsPerSample[2] == TiffConstants.BitsPerSampleRgb4Bit[2])
```
gfoidl, line 55:
```suggestion
                    if (bitsPerSample[2] == TiffConstants.BitsPerSampleRgb4Bit[2] &&
                        bitsPerSample[1] == TiffConstants.BitsPerSampleRgb4Bit[1] &&
                        bitsPerSample[0] == TiffConstants.BitsPerSampleRgb4Bit[0])
```
By flipping the order the JIT only needs to emit bound checks for index 2, thus saving some branches.
You can do the same for `Bit24` above.

I believe for .NET 6 JIT got some improvements in that area, but older runtimes will profit from that change.

src/ImageSharp/Formats/Tiff/TiffDecoderOptionsParser.cs
```diff
@@ -179,7 +180,18 @@ private static void ParseColorType(this TiffDecoderCore options, ExifProfile exi
...
+                        switch (options.BitsPerSample)
+                        {
+                            case TiffBitsPerSample.Bit24:
+                                options.ColorType = TiffColorType.Rgb888;
+                                break;
+                            case TiffBitsPerSample.Bit12:
+                                options.ColorType = TiffColorType.Rgb444;
+                                break;
+                            default:
+                                TiffThrowHelper.ThrowNotSupported("Bits per sample is nut supported.");
```
gfoidl, line 192:
```suggestion
                                TiffThrowHelper.ThrowNotSupported("Bits per sample is not supported.");
```

src/ImageSharp/Formats/Tiff/TiffEncoderEntriesCollector.cs
JimBobSquarePants: I don't really understand this change. Could you please explain to me what the issue is with pulling the value from the decoder if present?

JimBobSquarePants:
>`SamplesPerPixel` tag can have arbitrary value and can have different bits per component for each component corresponding to a pixel.
>
>We don't need to support different different bits per component other than 565 (and 555), but we can support the same >arbitrary bit values: {2,2,2}, {3,3,3}...

@IldarKhayrutdinov So what does this mean in regards to this PR?

src/ImageSharp/Formats/Tiff/TiffEncoderEntriesCollector.cs
brianpopow: There are two possible values of the `PlanarConfiguration`:
- *Chunky*: The data is stored as RGBRGBRGB. This is the default mode.
- *Planar*: This is more uncommon. The data is stored as: 
RRR
GGG
BBB

The Tiff Encoder does always encodes the data as `Chunky`, but does not make this explicitly by storing a value for the `PlanarConfiguration`. The value is omitted and because decoders will assume `Chunky` this will work. 

Now in the case of when we decode an actual 'Planar' Tiff and the reencode this TIFF again, the `PlanarConfiguration` would carry over to the encoded image, because its present in the EXIF profile. This would change the interpretation of the data and make the encoded image invalid. Therefore i thought it would be a good idea to make it explicit and always set it to `Chunky`.

brianpopow:
> > `SamplesPerPixel` tag can have arbitrary value and can have different bits per component for each component corresponding to a pixel.
> > We don't need to support different different bits per component other than 565 (and 555), but we can support the same >arbitrary bit values: {2,2,2}, {3,3,3}...
> 
> @IldarKhayrutdinov So what does this mean in regards to this PR?

Now that I thought about it again, I think Ildar meant we can use `RgbTiffColor{TPixel}` for all bit values which have the same values for all channels. So we dont actually need Rgb444TiffColor{TPixel}.cs and can use `RgbTiffColor{TPixel}`. I think i misunderstood that when I first read it. Is that correct @IldarKhayrutdinov? I will convert this to WIP and check it again.

IldarKhayrutdinov:
> > `SamplesPerPixel` tag can have arbitrary value and can have different bits per component for each component corresponding to a pixel.
> > We don't need to support different different bits per component other than 565 (and 555), but we can support the same >arbitrary bit values: {2,2,2}, {3,3,3}...
> 
> @IldarKhayrutdinov So what does this mean in regards to this PR?

@JimBobSquarePants 
This PR doesn't change anything , but only adds support for 4-bit RGB, so can separate different `SamplesPerPixel` problem into new PR. It seems like once upon a time support was conceived and it was supposed to work (but were no tests), but at some point it stopped being supported, I missed this moment. 
In principle, this is not a bug, but we must support it, since it is worth a little effort for us.

IldarKhayrutdinov:
> Now that I thought about it again, I think Ildar meant we can use `RgbTiffColor{TPixel}` for all bit values which have the same values for all channels. So we dont actually need Rgb444TiffColor{TPixel}.cs and can use `RgbTiffColor{TPixel}`. I think i misunderstood that when I first read it. Is that correct @IldarKhayrutdinov? I will convert this to WIP and check it again.


@brianpopow `Rgb444TiffColor{TPixel}` is needed because `RgbTiffColor{TPixel}` is slow, also worth adding most relevant `555/565`. 
By the way, `RgbTiffColor{TPixel}`  is can to decode any arbitrary `SamplesPerPixel` for 1-16 (can expand to 32) bits per sample: {2,2,2}, {3,3,3}, {5,6,5}, {7,6,5}..

brianpopow:
> In principle, this is not a bug, but we must support it, since it is worth a little effort for us.
> Do it in this PR or in the new PR, you decide ☺️

@IldarKhayrutdinov thanks for the clarification. 
I have added support for 6, 10 and 14 bit per channel and use `RgbTiffColor{TPixel}` for that. Note that I decided to not add 3 bits per channel support, since I think its not really a valid option. 9 bit per pixel would be completely unaligned at the byte border and I did not see those in the tiff-testsuite either. Also 565 should be a separate PR since i think its worth to implement its own `TiffColorDecoder`.

src/ImageSharp/Formats/Tiff/PhotometricInterpretation/TiffColorDecoderFactory{TPixel}.cs
```diff
@@ -57,6 +57,16 @@ public static TiffBaseColorDecoder<TPixel> Create(TiffColorType colorType, ushor
                     DebugGuard.IsTrue(colorMap == null, "colorMap");
                     return new RgbTiffColor<TPixel>(bitsPerSample);
 
+                case TiffColorType.Rgb222:
+                    DebugGuard.IsTrue(
+                        bitsPerSample.Length == 3
+                        && bitsPerSample[0] == 2
```
JimBobSquarePants, line 63: You can use the indexer check skipping trick discussed the other day here I reckon.

JimBobSquarePants (approved):
Very cool! Love how well structured this all is. 👍

PR #1652, opened 2021-06-05
Add support for decoding tiff's with additional pixel variants

brianpopow:
### Prerequisites

### Description

This **PR** will add support for decoding tiff's with:
- Single channel gray tiff's with 2, 6, 10, 12, 14, 16 bits per pixel
- 12 bits per channel RGB
- 16 bits per channel RGB

/cc @IldarKhayrutdinov


src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
```diff
@@ -50,9 +50,9 @@ public TiffDecoderCore(Configuration configuration, ITiffDecoderOptions options)
         }
 
         /// <summary>
-        /// Gets or sets the number of bits per component of the pixel format used to decode the image.
+        /// Gets or sets the bits per sample.
         /// </summary>
-        public TiffBitsPerSample BitsPerSample { get; set; }
+        public ushort[] BitsPerSample { get; set; }
```
IldarKhayrutdinov, line 55: @brianpopow 👍  it looks easier and better

src/ImageSharp/Formats/Tiff/TiffDecoderOptionsParser.cs
```diff
@@ -173,30 +185,39 @@ private static void ParseColorType(this TiffDecoderCore options, ExifProfile exi
...
 
                     if (options.PlanarConfiguration == TiffPlanarConfiguration.Chunky)
                     {
-                        switch (options.BitsPerSample)
+                        ushort bitsPerChannel = options.BitsPerSample[0];
+                        switch (bitsPerChannel)
                         {
-                            case TiffBitsPerSample.Bit42:
+                            case 16:
+                                options.ColorType = TiffColorType.Rgb161616;
```
IldarKhayrutdinov, line 199: What do you think, is it possible to simplify parsing [TiffColorType](https://github.com/SixLabors/ImageSharp/blob/aba1050bae1a1dc0aee0a72d275933a8f6b41254/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/TiffColorType.cs)? For example, maybe, can move parsing logic directly to [TiffColorDecoderFactory](https://github.com/SixLabors/ImageSharp/blob/aba1050bae1a1dc0aee0a72d275933a8f6b41254/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/TiffColorDecoderFactory%7BTPixel%7D.cs)?

src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
JimBobSquarePants: I'm curious, why is this a public property (_I haven't reviewed the usage_)? I'm always cautious around exposing arrays.

src/ImageSharp/Formats/Tiff/TiffFrameMetadata.cs
```diff
@@ -35,6 +35,11 @@ private TiffFrameMetadata(TiffFrameMetadata other)
         /// </summary>
         public TiffBitsPerPixel? BitsPerPixel { get; set; }
 
+        /// <summary>
+        /// Gets or sets number of bits per component.
+        /// </summary>
+        public ushort[] BitsPerSample { get; set; }
```
JimBobSquarePants, line 41: Yeah, I really do not like this. It's an unbounded array.

src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
IldarKhayrutdinov: @JimBobSquarePants oh yes, arrays aren't safe, can make as `ReadOnlyCollection`

JimBobSquarePants: We reuse the property for TiffFrameMetadata. Surely there’s limited scope for allowable bits per sample no?

JimBobSquarePants: Thanks! Ok, then we're gonna have to sanitize that property or use some sort of struct that takes up to 3 values. 

I see [Aspose](https://apireference.aspose.com/imaging/net/aspose.imaging.imageoptions/tiffoptions/properties/bitspersample) use an array

brianpopow: I have added back in now the `TiffBitsPerSample` enum instead of using the array directly: 67f7b782

src/ImageSharp/Formats/Tiff/TiffDecoderOptionsParser.cs
brianpopow: while i think this sounds like a sensible idea, I could not come up with a good solution for it. I will leave it now as it is, but maybe we can improve that in the future.

src/ImageSharp/Formats/Tiff/TiffBitsPerSampleExtensions.cs
```diff
@@ -13,27 +13,42 @@ internal static class TiffBitsPerSampleExtensions
         /// </summary>
         /// <param name="tiffBitsPerSample">The tiff bits per sample.</param>
         /// <returns>Bits per sample array.</returns>
-        public static ushort[] Bits(this TiffBitsPerSample tiffBitsPerSample)
+        public static ushort[] BitsPerChannel(this TiffBitsPerSample tiffBitsPerSample)
```
JimBobSquarePants, line 16: You're looking to avoid having to do this and also support any arbitrary arrays of up to 3 elements with each value maxed at 16 right?

Couldn't we define a struct made up of 3 `ushort` fields? and have `TryParse(ushort[])` and `ToArray()` properties. The array allocations would be so small an uncommon they wouldn't be an issue.

Something like this?

```c#
public readonly struct TiffBitsPerSample : IEquatable<TiffBitsPerSample>
{
    public readonly ushort Channel0;
    public readonly ushort Channel1;
    public readonly ushort Channel2;

    public TiffBitsPerSample(ushort channel0, ushort channel1, ushort channel2)
    {
        this.Channel0 = (ushort)Numerics.Clamp(channel0, 1, 16);
        this.Channel1 = (ushort)Numerics.Clamp(channel1, 0, 16);
        this.Channel2 = (ushort)Numerics.Clamp(channel2, 0, 16);
    }

    public static bool TryParse(ushort[] value, out TiffBitsPerSample sample)
    {
        if (value is null || value.Length == 0)
        {
            sample = default;
            return false;
        }

        ushort c2;
        ushort c1;
        ushort c0;
        switch (value.Length)
        {
            case 3:
                c2 = value[2];
                c1 = value[1];
                c0 = value[0];
                break;
            case 2:
                c2 = 0;
                c1 = value[1];
                c0 = value[0];
                break;
            default:
                c2 = 0;
                c1 = 0;
                c0 = value[0];
                break;
        }

        sample = new TiffBitsPerSample(c0, c1, c2);
        return true;
    }

    public override bool Equals(object obj)
        => obj is TiffBitsPerSample sample && this.Equals(sample);

    public bool Equals(TiffBitsPerSample other)
        => this.Channel0 == other.Channel0
        && this.Channel1 == other.Channel1
        && this.Channel2 == other.Channel2;

    public override int GetHashCode()
        => HashCode.Combine(this.Channel0, this.Channel1, this.Channel2);

    public ushort[] ToArray()
    {
        if (this.Channel1 == 0)
        {
            return new ushort[] { this.Channel0 };
        }

        if (this.Channel2 == 0)
        {
            return new ushort[] { this.Channel0, this.Channel1 };
        }

        return new ushort[] { this.Channel0, this.Channel1, this.Channel2 };
    }

    public override readonly string ToString()
        => $"TiffBitsPerSample({this.Channel0}, {this.Channel1}, {this.Channel2})";
}
```

You could then create static instances of known types for easy comparison during encoding.

brianpopow: Good idea, I have changed BitsPerSample now to a struct aa848d74.

> You're looking to avoid having to do this and also support any arbitrary arrays of up to 3 elements with each value maxed at 16 right?

Arbitrary is maybe strong word here. As Ildar said, the spec allows arbitrary values, but i have not yet seen a tiff in the wild which has not the same values for all channels. The maximum is 32 bits per channel, so we currently only missing 24 and 32 bits per channel as far as i can tell.

src/ImageSharp/Formats/Tiff/TiffBitsPerSample.cs
```diff
@@ -1,96 +1,242 @@
...
+
+            return new[] { this.Channel0, this.Channel1, this.Channel2 };
+        }
+
+        /// <summary>
+        /// Maps an array of bits per sample to a concrete struct value.
+        /// </summary>
+        /// <param name="bitsPerSample">The bits per sample array.</param>
+        /// <returns>TiffBitsPerSample enum value.</returns>
+        public static TiffBitsPerSample? GetBitsPerSample(ushort[] bitsPerSample)
```
JimBobSquarePants, line 120: I don't think we should expose this. Ideally the caller would use the result of `TryParse` within which we can incorporate some of this logic.

brianpopow: yeah your right, that method is totally unnecessary now. I overlooked that. Its removed now.

src/ImageSharp/Formats/Tiff/TiffFrameMetadata.cs
```diff
@@ -77,7 +77,12 @@ internal static void Parse(TiffFrameMetadata meta, ExifProfile profile)
         {
             if (profile != null)
             {
-                meta.BitsPerSample = profile.GetValue(ExifTag.BitsPerSample) != null ? TiffBitsPerSample.GetBitsPerSample(profile.GetValue(ExifTag.BitsPerSample)?.Value) : null;
+                ushort[] bitsPerSampleValue = profile.GetValue(ExifTag.BitsPerSample)?.Value;
+                if (bitsPerSampleValue != null && TiffBitsPerSample.TryParse(bitsPerSampleValue, out TiffBitsPerSample bitsPerSample))
```
JimBobSquarePants, line 81: This is 😍 though you don't need the null check since the struct already handles that.

src/ImageSharp/Formats/Tiff/TiffDecoderOptionsParser.cs
```diff
@@ -69,7 +69,7 @@ public static void VerifyAndParse(this TiffDecoderCore options, ExifProfile exif
             options.Predictor = frameMetadata.Predictor ?? TiffPredictor.None;
             options.PhotometricInterpretation = frameMetadata.PhotometricInterpretation ?? TiffPhotometricInterpretation.Rgb;
             options.BitsPerPixel = frameMetadata.BitsPerPixel != null ? (int)frameMetadata.BitsPerPixel.Value : (int)TiffBitsPerPixel.Bit24;
-            options.BitsPerSample = GetBitsPerSample(frameMetadata.BitsPerPixel);
+            options.BitsPerSample = frameMetadata.BitsPerSample != null ? frameMetadata.BitsPerSample?.ToArray() : Array.Empty<ushort>();
```
JimBobSquarePants, line 72: I'm curious, why are we using `ToArray()` here? the only use case for that I would have thought would have been writing to the EXIF tag when encoding.

src/ImageSharp/Formats/Tiff/TiffFrameMetadata.cs
brianpopow: removed the null check

src/ImageSharp/Formats/Tiff/TiffDecoderOptionsParser.cs
brianpopow: BitsPerSample was still a `ushort[]`, I have changed that now to the TiffBitsPerSample struct

JimBobSquarePants (approved):
LGTM. 👍  I think we've definitely improved the API also!

src/ImageSharp/Formats/Tiff/TiffBitsPerSampleExtensions.cs
JimBobSquarePants: Glad I could help here!

PR #1654, opened 2021-06-10
Quantization: reduce memory footprint without compromising speed and quality

### LINKED ISSUE #1350: Quantization: reduce memory footprint without compromising speed and quality

JimBobSquarePants:
### Prerequisites

### Description

Fixes #1350.

This PR does a couple of things. 
- Replaces the unbounded dictionary based cache used for the `EuclideanPixelMap<TPixel>` with a cache based upon the Wu quantizer pixel map. This reduces the memory usage to a set 2.2MB for all images.
- Replaces any parallel code using the map since the new cache is not thread safe. This actually fixes an issue where the output was not idempotent.
- Fixes a few issues with the `OctreeQuantizer`'s color distribution and transparency handling.
- Replaces the `BmpEncoder` default quantizer with `OctreeQuantizer`

Performance is much better ~4x faster, with Gif encoding comparing well with System.Drawing   

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19043
.NET Core SDK=5.0.300
  [Host]     : .NET Core 3.1.15 (CoreCLR 4.700.21.21202, CoreFX 4.700.21.21402), X64 RyuJIT
  Job-UDUUQH : .NET Framework 4.8 (4.8.4341.0), X64 RyuJIT
  Job-SIPGYB : .NET Core 2.1.28 (CoreCLR 4.6.30015.01, CoreFX 4.6.30015.01), X64 RyuJIT
  Job-CSWJHY : .NET Core 3.1.15 (CoreCLR 4.700.21.21202, CoreFX 4.700.21.21402), X64 RyuJIT

```
Master
|               Method |        Job |       Runtime |         TestImage |      Mean |     Error |   StdDev | Ratio | RatioSD |     Gen 0 |     Gen 1 |    Gen 2 |   Allocated |
|--------------------- |----------- |-------------- |------------------ |----------:|----------:|---------:|------:|--------:|----------:|----------:|---------:|------------:|
| 'System.Drawing Gif' | Job-BHOJYB |    .NET 4.7.2 |       Bmp/Car.bmp |  12.76 ms |  1.518 ms | 0.083 ms |  1.00 |    0.00 |  234.3750 |         - |        - |  1093.11 KB |
|     'ImageSharp Gif' | Job-BHOJYB |    .NET 4.7.2 |       Bmp/Car.bmp |  39.02 ms | 19.385 ms | 1.063 ms |  3.06 |    0.08 |  846.1538 |  307.6923 |  76.9231 |  5992.04 KB |
|                      |            |               |                   |           |           |          |       |         |           |           |          |             |
| 'System.Drawing Gif' | Job-CZDZPN | .NET Core 2.1 |       Bmp/Car.bmp |  12.66 ms |  1.692 ms | 0.093 ms |  1.00 |    0.00 |  234.3750 |         - |        - |  1089.88 KB |
|     'ImageSharp Gif' | Job-CZDZPN | .NET Core 2.1 |       Bmp/Car.bmp |  37.16 ms | 25.210 ms | 1.382 ms |  2.93 |    0.09 |  928.5714 |  428.5714 | 142.8571 |   203.71 KB |
|                      |            |               |                   |           |           |          |       |         |           |           |          |             |
| 'System.Drawing Gif' | Job-YJADTU | .NET Core 3.1 |       Bmp/Car.bmp |  12.68 ms |  2.301 ms | 0.126 ms |  1.00 |    0.00 |         - |         - |        - |   197.95 KB |
|     'ImageSharp Gif' | Job-YJADTU | .NET Core 3.1 |       Bmp/Car.bmp |  37.24 ms | 13.228 ms | 0.725 ms |  2.94 |    0.09 |  928.5714 |  428.5714 | 142.8571 |  5941.78 KB |
|                      |            |               |                   |           |           |          |       |         |           |           |          |             |
| 'System.Drawing Gif' | Job-BHOJYB |    .NET 4.7.2 | Png/rgb-48bpp.png |  43.99 ms |  8.502 ms | 0.466 ms |  1.00 |    0.00 |         - |         - |        - |   539.82 KB |
|     'ImageSharp Gif' | Job-BHOJYB |    .NET 4.7.2 | Png/rgb-48bpp.png | 134.89 ms | 40.334 ms | 2.211 ms |  3.07 |    0.06 | 3250.0000 | 1500.0000 | 500.0000 | 22261.71 KB |
|                      |            |               |                   |           |           |          |       |         |           |           |          |             |
| 'System.Drawing Gif' | Job-CZDZPN | .NET Core 2.1 | Png/rgb-48bpp.png |  43.28 ms |  8.130 ms | 0.446 ms |  1.00 |    0.00 |         - |         - |        - |   538.94 KB |
|     'ImageSharp Gif' | Job-CZDZPN | .NET Core 2.1 | Png/rgb-48bpp.png | 127.82 ms | 60.645 ms | 3.324 ms |  2.95 |    0.05 | 3200.0000 | 1400.0000 | 400.0000 |   228.45 KB |
|                      |            |               |                   |           |           |          |       |         |           |           |          |             |
| 'System.Drawing Gif' | Job-YJADTU | .NET Core 3.1 | Png/rgb-48bpp.png |  43.97 ms | 16.380 ms | 0.898 ms |  1.00 |    0.00 |         - |         - |        - |   393.88 KB |
|     'ImageSharp Gif' | Job-YJADTU | .NET Core 3.1 | Png/rgb-48bpp.png | 127.14 ms | 18.737 ms | 1.027 ms |  2.89 |    0.07 | 3200.0000 | 1400.0000 | 400.0000 | 22133.64 KB |

This PR
|               Method |        Job |       Runtime |         TestImage |     Mean |     Error |   StdDev | Ratio | RatioSD |    Gen 0 | Gen 1 | Gen 2 |  Allocated |
|--------------------- |----------- |-------------- |------------------ |---------:|----------:|---------:|------:|--------:|---------:|------:|------:|-----------:|
| 'System.Drawing Gif' | Job-SSUAIX |    .NET 4.7.2 |       Bmp/Car.bmp | 11.94 ms |  0.847 ms | 0.046 ms |  1.00 |    0.00 | 234.3750 |     - |     - | 1093.11 KB |
|     'ImageSharp Gif' | Job-SSUAIX |    .NET 4.7.2 |       Bmp/Car.bmp | 11.93 ms |  3.801 ms | 0.208 ms |  1.00 |    0.01 |  15.6250 |     - |     - |  198.59 KB |
|                      |            |               |                   |          |           |          |       |         |          |       |       |            |
| 'System.Drawing Gif' | Job-MNGTMS | .NET Core 2.1 |       Bmp/Car.bmp | 11.90 ms |  1.515 ms | 0.083 ms |  1.00 |    0.00 | 234.3750 |     - |     - | 1089.88 KB |
|     'ImageSharp Gif' | Job-MNGTMS | .NET Core 2.1 |       Bmp/Car.bmp | 10.39 ms |  0.988 ms | 0.054 ms |  0.87 |    0.00 |  15.6250 |     - |     - |   198.3 KB |
|                      |            |               |                   |          |           |          |       |         |          |       |       |            |
| 'System.Drawing Gif' | Job-YGEDPI | .NET Core 3.1 |       Bmp/Car.bmp | 11.33 ms |  0.116 ms | 0.006 ms |  1.00 |    0.00 |  15.6250 |     - |     - |  197.95 KB |
|     'ImageSharp Gif' | Job-YGEDPI | .NET Core 3.1 |       Bmp/Car.bmp | 10.21 ms |  2.127 ms | 0.117 ms |  0.90 |    0.01 |  15.6250 |     - |     - |  198.32 KB |
|                      |            |               |                   |          |           |          |       |         |          |       |       |            |
| 'System.Drawing Gif' | Job-SSUAIX |    .NET 4.7.2 | Png/rgb-48bpp.png | 42.68 ms | 76.627 ms | 4.200 ms |  1.00 |    0.00 |        - |     - |     - |  539.82 KB |
|     'ImageSharp Gif' | Job-SSUAIX |    .NET 4.7.2 | Png/rgb-48bpp.png | 38.06 ms |  6.984 ms | 0.383 ms |  0.90 |    0.08 |        - |     - |     - |  198.64 KB |
|                      |            |               |                   |          |           |          |       |         |          |       |       |            |
| 'System.Drawing Gif' | Job-MNGTMS | .NET Core 2.1 | Png/rgb-48bpp.png | 39.87 ms |  1.479 ms | 0.081 ms |  1.00 |    0.00 |        - |     - |     - |  538.94 KB |
|     'ImageSharp Gif' | Job-MNGTMS | .NET Core 2.1 | Png/rgb-48bpp.png | 35.99 ms |  4.689 ms | 0.257 ms |  0.90 |    0.01 |        - |     - |     - |   198.3 KB |
|                      |            |               |                   |          |           |          |       |         |          |       |       |            |
| 'System.Drawing Gif' | Job-YGEDPI | .NET Core 3.1 | Png/rgb-48bpp.png | 42.76 ms |  4.836 ms | 0.265 ms |  1.00 |    0.00 |        - |     - |     - |  393.38 KB |
|     'ImageSharp Gif' | Job-YGEDPI | .NET Core 3.1 | Png/rgb-48bpp.png | 39.47 ms | 13.996 ms | 0.767 ms |  0.92 |    0.02 |        - |     - |     - |  198.45 KB |

Quality is still great also. Here's an example

Original 73.8kb
bike-small

[PngQuant](https://pngquant.org/) (_Best in class dedicated software which clamps dithering output, I'm happy to not be able to beat this._) 22.8kb
Ducati_side_shadow-fs8

Wu 22.7kb
QuantizeImageShouldPreserveMaximumColorPrecision_Rgba32_bike-small_Wu

Original 239kb
splash

Wu 52.8kb
QuantizeImageShouldPreserveMaximumColorPrecision_Rgba32_splash_Wu

Original 179kb
blur

Wu 68.5kb
QuantizeImageShouldPreserveMaximumColorPrecision_Rgba32_blur_Wu


tests/ImageSharp.Tests/Formats/Tiff/TiffEncoderTests.cs
```diff
@@ -475,12 +476,20 @@ private static void TestStripLength<TPixel>(TestImageProvider<TPixel> provider,
                 }
             }
 
+            // TODO: Ask Brian about this. It seems like some of the images used
```
brianpopow, line 479: The reason the image is different from the input image is because a palette is used. Due to the quantization step the output can differ slightly. I have checked the output and it looks visually identical. 

@JimBobSquarePants I will go ahead and change the particular test to use a tolerant comparer, if you dont mind.

JimBobSquarePants: Yeah go ahead, thanks!

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -97,9 +97,92 @@ private int GetClosestColorSlow(TPixel color, ref TPixel paletteRef, out TPixel
...
         }
+
+        /// <summary>
+        /// Returns the Euclidean distance squared between two specified points.
+        /// </summary>
+        /// <param name="a">The first point.</param>
+        /// <param name="b">The second point.</param>
+        /// <returns>The distance squared.</returns>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        private static float DistanceSquared(Rgba32 a, Rgba32 b)
```
antonfirsov, line 112: Looks like an integral value with an upper bound of `4*256^2`, safe to return `int` instead of `float`.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -97,9 +97,92 @@ private int GetClosestColorSlow(TPixel color, ref TPixel paletteRef, out TPixel
...
+            private const int IndexAlphaCount = (1 << IndexAlphaBits) + 1;
+            private const int RgbShift = 8 - IndexBits;
+            private const int AlphaShift = 8 - IndexAlphaBits;
+            private const int TableLength = IndexCount * IndexCount * IndexCount * IndexAlphaCount;
+            private short[] table;
+
+            public static ColorDistanceCache Create()
+            {
+                ColorDistanceCache result = default;
+                short[] entries = new short[TableLength];
```
antonfirsov, line 146: We need to use `MemoryAllocator` here. Fortunately `IQuantizer<TPixel> : IDisposable`, so it shouldn't be an issue.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -97,9 +97,92 @@ private int GetClosestColorSlow(TPixel color, ref TPixel paletteRef, out TPixel
...
+
+        /// <summary>
+        /// A cache for storing color distance matching results.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The granularity of the cache has been determined based upon the current
+        /// suite of test images and provides the lowest possible memory usage while
+        /// providing enough match accuracy.
+        /// Entry count is currently limited to 1221858 entries at 1.17MB.
```
antonfirsov, line 129:
```suggestion
        /// Entry count is currently limited to 610929 entries (1221858 bytes ~1.17MB).
```

antonfirsov (commented):
A few remarks, can do deeper review next week, if you don't get one by that time.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
antonfirsov: Although realized we need fast access through `Span<short>` instead of `Memory<short>`, which seems to be impossible with the current design.

We can keep the LOH allocation or pin the pooled buffer like in `ResizeKernelMap`:
https://github.com/SixLabors/ImageSharp/blob/381dff8640b721a34b1227c970fcf6ad6c5e3e72/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs#L55
https://github.com/SixLabors/ImageSharp/blob/381dff8640b721a34b1227c970fcf6ad6c5e3e72/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs#L258

JimBobSquarePants: Should be doable. I’ll have a look

antonfirsov: Pinning a 1MB buffer is not ideal, but if it happens for a <100ms time, it's probably better than a LOH allocation of the same amount.

JimBobSquarePants: Is there a risk using `MemoryAllocator` that we could end up with a discontiguous memory buffer.

antonfirsov: Only if you use Allocate2D

tests/ImageSharp.Tests/Processing/Processors/Dithering/DitherTests.cs
```diff
@@ -155,7 +156,7 @@ public void DitherFilter_ShouldNotDependOnSinglePixelType<TPixel>(TestImageProvi
                 appendPixelTypeToFileName: false);
         }
 
-        [Theory]
+        [Theory(Skip = "Unable to assign capacity smaller than the image.")]
```
JimBobSquarePants, line 159: @antonfirsov This is problematic. The maximum height of the image in rows provides less of a buffer than is required by the cache.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
br3aker: @antonfirsov isn't 1mb array a LOH allocation? Most likely it won't be moved by GC anyway, only if user manually force `GCSettings.LargeObjectHeapCompactionMode`. So pinning shouldn't change anything, doesn't it? If memory pool gives native memory, it isn't moved either, no?

antonfirsov: Yeah thats true, forgot that LOH objects are not moved. Pinning wont hurt at all here.

JimBobSquarePants: The pinning is more a performance thing to workaround the API to prevent us having to call `IMemoryOwner.GetSpan()` on a per pixel basis.

tests/ImageSharp.Tests/Processing/Processors/Dithering/DitherTests.cs
JimBobSquarePants: I think I should just drop pooling here. I wrote code to workaround our pooler limitations using `ArrayPool<byte>.Shared` but that's gonna just allocate the buffer anyway since we just exceed the maximum buffer amount.

JimBobSquarePants: Scratch that. After reading the [source](https://source.dot.net/#System.Private.CoreLib/TlsOverPerCoreLockedStacksArrayPool.cs) I realized that it's not pre-filling multiple buckets but filling a bucket on demand. Using `ArrayPool<short>.Shared` seems appropriate for this use case.

saucecontrol: You had that right the first time.  The comment [here](https://source.dot.net/#System.Private.CoreLib/TlsOverPerCoreLockedStacksArrayPool.cs,32) is weird because `NumBuckets` is the length of the buckets array, so max index is one less than that, which is the bucket for 1MiB arrays.  Anything larger ends up with an invalid bucket index and will new up an array.

JimBobSquarePants: Yeah but I was using `byte` and casting to avoid using a pool that I couldn't find usage of in the runtime. Now I'm using `short` my buffer length is less so it get pooled. The index is determined purely by the minimum length required.

saucecontrol: Ohhhh, gotcha :)

saucecontrol:
Looking at the way the Wu palette matching buckets by just chopping of the least significant bit, I suspect you're going to lose more quality here than would be acceptable on some images.  I've just run the PR version against a simple 0-255 ramp, which demonstrates the problem.

Given this input:

ramp

1.0.3's Octree Quantizer outputs the following, which is perfect except for the one column that is duplicated because the palette reserves 1 slot for transparent and is therefore missing one shade of red.:

rampis103

The PR's output is this, which is only using 32 colors from the palette, with some dithering trying to fix the errors.

rampispr

JimBobSquarePants:
@saucecontrol oh gosh, yeah that’s unpleasant. I guess we could bump the bit depth to match the default Wu values but that comes with increased memory usage 2MB I think which will require pooling tweaks.

I’m trying to do all this with the minimum of architectural changes

saucecontrol:
Yeah, the tradeoffs are super tricky.  Just wanted to help you visualize what you're giving up from the current quality.

Wu really does an amazing job on that paint splash image above, BTW.  My quantizer really does poorly on that one.  I'll be doing some tweaking based on that :)

JimBobSquarePants:
@saucecontrol Ok so here's Octree with my adjusted precision using 5 bits for each component. By using the `int` pool an casting as I pin I can keep the buffer pooled.

It's much better but not as perfect as 1.0.3 but I don't know if I can match that. 

QuantizeImageShouldPreserveMaximumColorPrecision_Rgba32_gradient_Octree

Interestingly Wu is shocking at gradients like this.

QuantizeImageShouldPreserveMaximumColorPrecision_Rgba32_gradient_Wu

These are all definitely improved for Wu though

Blur looks ace 
QuantizeImageShouldPreserveMaximumColorPrecision_Rgba32_blur_Wu

As does the bike
QuantizeImageShouldPreserveMaximumColorPrecision_Rgba32_bike-small_Wu

And splash
QuantizeImageShouldPreserveMaximumColorPrecision_Rgba32_splash_Wu

saucecontrol:
Oh yeah, blur is much better!  Weird that Wu makes a 32 color palette for the ramp and then only uses 16 of them when mapping.

JimBobSquarePants:
Yep, super weird. I’ll have to investigate that some time. Thanks for sharing that image btw, I’m really glad I updated the precision now!

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -97,9 +98,111 @@ private int GetClosestColorSlow(TPixel color, ref TPixel paletteRef, out TPixel
...
+                int b = rgba.B >> RgbShift;
+                int a = rgba.A >> AlphaShift;
+                int idx = GetPaletteIndex(r, g, b, a);
+                match = this.tablePointer[idx];
+                return match > -1;
+            }
+
+            [MethodImpl(InliningOptions.ShortMethod)]
+            private static int GetPaletteIndex(int r, int g, int b, int a)
+                => (r << ((IndexBits * 2) + IndexAlphaBits))
```
antonfirsov, line 189: See discussion: https://github.com/SixLabors/ImageSharp/pull/1646#discussion_r643132801

```suggestion
                => (r << ((IndexBits << 1) + IndexAlphaBits))
```

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -97,9 +98,111 @@ private int GetClosestColorSlow(TPixel color, ref TPixel paletteRef, out TPixel
...
+                + (r << (IndexBits + IndexAlphaBits + 1))
+                + (g << (IndexBits + IndexAlphaBits))
+                + (r << (IndexBits * 2))
```
antonfirsov, line 192:
```suggestion
                + (r << (IndexBits << 1))
```

antonfirsov (commented):
We should do something about the pooling issue, otherwise looks good.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -97,9 +98,111 @@ private int GetClosestColorSlow(TPixel color, ref TPixel paletteRef, out TPixel
...
+
+        /// <summary>
+        /// A cache for storing color distance matching results.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The granularity of the cache has been determined based upon the current
+        /// suite of test images and provides the lowest possible memory usage while
+        /// providing enough match accuracy.
+        /// Entry count is currently limited to 1185921 entries (2371842 bytes ~2.26MB).
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah without alpha I think you can do it in 600kb or something like that.  We can track alpha requirements easily enough as we build the palette.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -97,9 +98,111 @@ private int GetClosestColorSlow(TPixel color, ref TPixel paletteRef, out TPixel
...
+            private readonly int[] table;
+            private readonly short* tablePointer;
+
+            private ColorDistanceCache(int bufferLength, int entries)
+            {
+                // We use ArrayPool<int>.Shared for several reasons.
+                // 1. To avoid out of range issues caused by configuring small discontiguous buffers rented via MemoryAllocator
+                // 2. To ensure that the rented buffer is actually pooled.
+                // 3. The .NET runtime already uses this pool so we might already have a pooled array present.
+                this.table = ArrayPool<int>.Shared.Rent(bufferLength);
```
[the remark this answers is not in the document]
JimBobSquarePants: Yep that’s exactly why plus the allocator was throwing. Absolutely a hack but wasn’t sure when we’d be making progress on the allocator.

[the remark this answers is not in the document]
antonfirsov: With that topic I definitely want to push for going slowly and carefully, and do all the possible benchmarks/validations/reviews before merging :)

Not sure if it's worth to block on those improvements with this PR, we can just remove the condition check in  ArrayPoolMemoryAllocator and move on.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
saucecontrol: This is actually a deoptimization.  JIT can always optimize multiply by powers of 2.  In the case of `x * 2`, it optimizes to either `add` or, even better `lea`, which doesn't set flags.

The comment you linked was referring specifically to signed division, which can't safely be optimized to only a shift because right shift of a negative int rounds toward negative infinity, while a shift of a positive int rounds toward zero.  JIT must therefore emit extra code to force negative values back toward zero after the shift.

[sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunGBLAHYYGAZT4BzATAAmAET4A3UgApBwhAEoGAXgB8DBAwD0DUgG4uPegzWiJUuYoAcqoQa16Dxhk4vVL3NYArrYAqgL2MvJKyiFumjr6hibmAbwMccLhkY4KLpnuiV4mvmlp1rZikjIAskEANiq2CZ6GAFSmfjzplTl19S7NHkkMHaXU3eWMBdnV0v0qBS0jHakTVtNhEXP9+UNF7T5+AL5AA)

saucecontrol: Heh, I just noticed JIT will also rewrite `<< 1` as `add` or `lea` as well.  So no harm in writing it either way.  Whatever's more clear will do.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -150,8 +150,8 @@ private void EncodeGlobal<TPixel>(Image<TPixel> image, IndexedImageFrame<TPixel>
             // The palette quantizer can reuse the same pixel map across multiple frames
             // since the palette is unchanging. This allows a reduction of memory usage across
             // multi frame gifs using a global palette.
-            EuclideanPixelMap<TPixel> pixelMap = default;
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah will clean up

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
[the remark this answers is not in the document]
JimBobSquarePants: I think I'm gonna leave this for a future optimization. It's waaaaay better than a dictionary.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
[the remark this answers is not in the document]
JimBobSquarePants: Will do!

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
antonfirsov: > right shift of a negative int rounds toward negative infinity, while a shift of a positive int rounds toward zero

I didn't realize this is the real problem there, two's complement is tricky, thanks!

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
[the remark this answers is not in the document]
JimBobSquarePants: Done

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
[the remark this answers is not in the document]
antonfirsov: Yeah it's definitely a topic for another PR, we should remember to open a follow-up issue.

Directory.Build.props
```diff
@@ -26,5 +26,5 @@
   <PropertyGroup Condition="$(Configuration.StartsWith('Release')) == true">
     <Optimize>true</Optimize>
   </PropertyGroup>
-
+
```
antonfirsov, line 29: Extra space?

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter{TPixel}.cs
```diff
@@ -36,50 +36,38 @@ public TiffBiColorWriter(ImageFrame<TPixel> image, MemoryAllocator memoryAllocat
         /// <inheritdoc/>
         protected override void EncodeStrip(int y, int height, TiffBaseCompressor compressor)
         {
-            int width = this.Image.Width;
+            this.pixelsAsGray ??= this.MemoryAllocator.Allocate<byte>(height * this.Image.Width);
+
+            Span<byte> pixelAsGraySpan = this.pixelsAsGray.Slice(0, height * this.Image.Width);
+
+            Span<TPixel> pixelsBlackWhite = GetStripPixels(this.imageBlackWhite.GetRootFramePixelBuffer(), y, height);
```
antonfirsov, line 43: This is using `GetSingleSpan()`, which might be an issue when the buffer is discontiguous. We should do the conversion line-by-line instead.

In long term, it would be cool to have `PixelOperations<TPixel>` overloads which work with `Buffer2DRegion<T>`s, so we can work with "image slices" more easily.

antonfirsov (commented):
Quantizer changes look good, one major issue in `TiffBaseColorWriter{TPixel}.cs`.

I recommend to add tiff [tests stressing discontiguous buffers](https://github.com/SixLabors/ImageSharp/blob/7a550f8830e318d2683bea367ade2514c044c16c/tests/ImageSharp.Tests/Formats/Png/PngEncoderTests.cs#L512-L530).

For the tiff refactor I defer to @brianpopow and @IldarKhayrutdinov.

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter{TPixel}.cs
brianpopow: That's weird, I have changed that yesterday with #1662. Did something went wrong here with merging master? Usage of  `GetSingleSpan()` was removed. See: [TiffBiColorWriter, master](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter%7BTPixel%7D.cs)

JimBobSquarePants: WTH? Something bust have been funky about one of the recent PRs to do this. I've just been pulling master straight in. I'll have to fix it all manually.

brianpopow: I cannot really tell how this happened. 

I have added tests for discontiguous buffers: 23093821

JimBobSquarePants: I'm just about to force push a restored branch which will overwrite your commit.

brianpopow: The test commit? Ok that's easy to replace, go ahead.

JimBobSquarePants: Done

src/ImageSharp/Processing/Processors/Quantization/PaletteQuantizer{TPixel}.cs
```diff
@@ -17,25 +17,31 @@ internal struct PaletteQuantizer<TPixel> : IQuantizer<TPixel>
...
+        /// <param name="leaveMap">
+        /// <see langword="true"/> to leave the pixel map undisposed after disposing the <see cref="PaletteQuantizer{TPixel}"/> object; otherwise, <see langword="false"/>.
+        /// </param>
         [MethodImpl(InliningOptions.ShortMethod)]
         public PaletteQuantizer(
             Configuration configuration,
             QuantizerOptions options,
-            EuclideanPixelMap<TPixel> pixelMap)
+            EuclideanPixelMap<TPixel> pixelMap,
+            bool leaveMap)
```
antonfirsov, line 36: Nit: typical pattern I've seen for this is to have a bool parameter indicating ownership (not the lack of it), (-> `ownsMap`)

Feel free to ignore in this PR.

antonfirsov (approved):
LGTM.

brianpopow (approved):
Looks good, nice work!

src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter{TPixel}.cs
br3aker: > In long term, it would be cool to have `PixelOperations<TPixel>` overloads which work with `Buffer2DRegion<T>`s, so we can work with "image slices" more easily.

+1 for this.
It can be a game changer for jpeg encoding.

src/ImageSharp/Processing/Processors/Quantization/PaletteQuantizer{TPixel}.cs
JimBobSquarePants: I copied `DeflateStream`. That uses a `leaveOpen` parameter for the same thing.

PR #1655, opened 2021-06-11
Use GreatestCommonDivisor to prevent overflow

### LINKED ISSUE #1616: Overflow in `LeastCommonMultiple` leads to exception during resize operation
### Prerequisites

### Description
When doing a resize operation of an image with a particularly large height, I got an exception that ImageSharp was trying to allocate memory of negative total length.  I tracked this down to this line where period is calculated. The period calculation overflows making period become negative which ends up failing when it creates the kernel map several line below (158) while allocating memory in the constructor.
https://github.com/SixLabors/ImageSharp/blob/a6c17854684edaa998080233273d19d987174bb5/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs#L135

One possible solution for this particular case could be to change that line to use greatest common divisor instead and calculate period using:
```csharp
int period = destinationSize / Numerics.GreatestCommonDivisor(sourceSize, destinationSize);
```

### Steps to Reproduce
With a height of 207773, and a destination height of 51943, the GCD is 1 causing an overflow when multiplying both of those numbers in `Numerics.LeastCommonMultiple(a, b)`.
```csharp
var width = 5000;
var height = 207773;
using (var img = new Image<L16>(width, height))
{
    using var clone = img.Clone(ctx =>
    {
        var factor = 4;
        ctx.Resize(img.Width / factor, img.Height / factor, KnownResamplers.Bicubic);
    });
}
```

### System Configuration
I don't think it's relevant.

- ImageSharp version: 1.0.3
- Other ImageSharp packages and versions: N/A
- Environment (Operating system, version and so on): Win 10
- .NET Framework version: NET 3.1, NET 5.0
- Additional information: N/A

JimBobSquarePants:
### Prerequisites

### Description
Fixes #1616
Updates the period calculation to use `GreatestCommonDivisor` as per issue.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs
+++ b/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs
@@ -130,9 +130,9 @@ namespace SixLabors.ImageSharp.Processing.Processors.Transforms
             int radius = (int)TolerantMath.Ceiling(scale * sampler.Radius);
 
             // 'ratio' is a rational number.
-            // Multiplying it by LCM(sourceSize, destSize)/sourceSize will result in a whole number "again".
+            // Multiplying it by destSize/GCD(sourceSize, destSize) will result in a whole number "again".
             // This value is determining the length of the periods in repeating kernel map rows.
-            int period = Numerics.LeastCommonMultiple(sourceSize, destinationSize) / sourceSize;
+            int period = destinationSize / Numerics.GreatestCommonDivisor(sourceSize, destinationSize);
 
             // the center position at i == 0:
             double center0 = (ratio - 1) * 0.5;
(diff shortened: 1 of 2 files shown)


src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs
```diff
@@ -130,9 +130,9 @@ protected virtual void Dispose(bool disposing)
             int radius = (int)TolerantMath.Ceiling(scale * sampler.Radius);
 
             // 'ratio' is a rational number.
-            // Multiplying it by LCM(sourceSize, destSize)/sourceSize will result in a whole number "again".
+            // Multiplying it by destSize/GCD(sourceSize, destSize) will result in a whole number "again".
             // This value is determining the length of the periods in repeating kernel map rows.
-            int period = Numerics.LeastCommonMultiple(sourceSize, destinationSize) / sourceSize;
+            int period = destinationSize / Numerics.GreatestCommonDivisor(sourceSize, destinationSize);
```
dlemstra, line 135: Cannot place a comment on line `122` but do we need to protect this method from division by zero?

JimBobSquarePants: GCD should always yield a positive integer unless both inputs are 0. That can never happen in our code since the `ResizeHelper` will always return positive integers.

antonfirsov (approved):
LGTM

PR #1657, opened 2021-06-12
More reliable runtime feature dependent tests skip

br3aker:
### Prerequisites

### Description
Quite some tests for intrinsic-only code use naive skip code **without skip reason**:
https://github.com/SixLabors/ImageSharp/blob/381dff8640b721a34b1227c970fcf6ad6c5e3e72/tests/ImageSharp.Tests/Common/SimdUtilsTests.cs#L399-L402

`SimdUtilsTests` tests set has exclusive skip check with skip reason but with rather *strange check* for Avx2: `SimdUtils.HasVector8` instead of `Avx2.IsSupported` but it at least writes something to the output:
https://github.com/SixLabors/ImageSharp/blob/381dff8640b721a34b1227c970fcf6ad6c5e3e72/tests/ImageSharp.Tests/Common/SimdUtilsTests.cs#L108-L117

The idea of this proposal is to provide custom fact/theory attributes which check & skip tests on the testing framework side rather than testing code:

```csharp
// Outcome: SKIPPED
// Output message
// Required: AVX, FMA
// Not supported: FMA
[IntrinsicFact(HwIntrinsics.AVX | HwIntrinsics.FMA)]
public void SomeTest()
{
    // test code
}
```

Instead of this:
```csharp
// Outcome: PASSED
public void SomeTest()
{
    if(!Avx.IsSupported || !Fma.IsSupported)
    {
        // possible this.output.WriteLine("Reason")
        return;
    }

    // test code
}
```

The only thing that interferes - Test project already has `HwIntrinsics` enumeration for `FeatureTestRunner` (that's why pr has `_HwIntrinsics` naming). This can be fixed by either unifiing them or having both enumerations but under different naming I guess?


tests/ImageSharp.Tests/Common/SimdUtilsTests.cs
```diff
@@ -105,29 +105,13 @@ public void FastRound_RandomValues(int seed, float scale)
...
-            {
-                this.Output.WriteLine("Skipping AVX2 specific test case: " + testCaseName);
-                return true;
-            }
-
-            return false;
-        }
-
-        [Theory]
+        [IntrinsicTheory(_HwIntrinsics.AVX2)]
```
antonfirsov, line 108: This change will skip some tests which should not be skipped.

The tests and the code under test are here from the times HW Intrinsics did not exist. `SkipOnNonAvx2` isn't really checking for AVX2, but rather for `Vector.IsHardwareAccelerated && Vector<float>.Count == 8`, which is poor man's AVX2 on older platforms.

When we started to add `System.Runtime.Intrinsics.X86` stuff to the repo, I updated the product code to use the term *"Vector8"* instead of *"Avx2"* for the old `Vector<float>` stuff, but looks like the test code still keeps using Avx2, which is confusing.

tests/ImageSharp.Tests/TestUtilities/Attributes/Intrinsics/IntrinsicTestsUtils.cs
```diff
@@ -0,0 +1,96 @@
...
+            if (isSupported)
+            {
+                set &= ~value;
+            }
+        }
+#endif
+    }
+
+    [Flags]
+    public enum _HwIntrinsics
```
antonfirsov, line 64: This goes against our naming conventions, which are for some reason not strictly checked now (see #1658).

I think this could be more general than just "HW intrinsics", maybe let's name it `RuntimeFeatureFilter` ?

tests/ImageSharp.Tests/TestUtilities/Attributes/Intrinsics/IntrinsicTheoryAttribute.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System;
+using Xunit;
+
+namespace SixLabors.ImageSharp.Tests
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class IntrinsicTheoryAttribute : TheoryAttribute
```
antonfirsov, line 10: `RuntimeFeatureConditionalTheory` ?

antonfirsov (commented):
Good addition! We need to fix the naming and the "Vector8" stuff.

tests/ImageSharp.Tests/TestUtilities/Attributes/Intrinsics/IntrinsicTestsUtils.cs
br3aker: Yeah, it's checked by stylecop, I named it this way because `FeatureTestRunner` already has enum with this name but it's used for disabling features. `RuntimeFeatureFilter` looks cleaner considering Vector8 stuff, good suggestion! Will fix asap.

tests/ImageSharp.Tests/TestUtilities/Attributes/Intrinsics/IntrinsicTheoryAttribute.cs
br3aker: Your naming is more clear to undestand considering `RuntimeFeatureFilter` enum naming but isn't it too long?

tests/ImageSharp.Tests/Common/SimdUtilsTests.cs
br3aker: Hm, makes sense now, I will add `Vector8` value separate from avx2 and check it with `Vector.IsHardwareAccelerated && Vector<float>.Count == 8`. Will also try to figure out which tests need that specific check. As this can be crucial to some tests will not unmark draft thingy. Thanks for clarifying, this one was the most confusing bit about intrinsics tests.

tests/ImageSharp.Tests/TestUtilities/Attributes/Intrinsics/IntrinsicTheoryAttribute.cs
antonfirsov: Yeah it's super verbose, but it's in sync with arcade's `[ConditionalTheory]` which we already use. Maybe @JimBobSquarePants  has some suggestions.

tests/ImageSharp.Tests/Common/SimdUtilsTests.cs
br3aker: A message so I would understand this properly, `Vector8` basically translates to avx intructions on runtimes without `Intrinsic` API platforms?

What is also a bit confusing, isn't first avx version also have vector8 support (256bit vectors to be hardware-friendly)? Wouldn't `Vector.IsHardwareAccelerated && Vector<float>.Count == 8` be true for avx but **not** avx2 CPUs?

And the most important question: should we care? 😂
Maybe checks should be like something like this:
// not sure if any other simd extension provides 256bit vectors
Avx -> `Vector.IsHardwareAccelerated && Vector<float>.Count == 8`
Avx2 -> `Avx2.IsSupported`

br3aker: And yet another message, *maybe* we should rename `SimdUtils.HasVector8` to `SimdUtils.HasAvx`?
it would comply with `HasAvx2` property:
```csharp
SimdUtils.HasVector8
SimdUtils.HasAvx2
```
vs
```csharp
SimdUtils.HasAvx  // either Vector8 check OR Avx.IsSupported
SimdUtils.HasAvx2 // Avx2.IsSupported
```

antonfirsov: The `Vector<float>.Count == 8` optimization actually uses AVX2, so `HasAvx` would be misleading, I prefer `Vector8` since it makes it clear we refer to the `Vector<float>` stuff.

br3aker: Oh now I got it, condition `Vector<float>.Count == 8` could mean Avx OR Avx2 and would compile operations to intrinsics if it's supported. So I guess it's better to provide all 3 variants: `Avx, Avx2, Vector8` so Avx2 specific code tests can use this.

saucecontrol: Not sure if you got this, but the way Vector<T> is implemented in the JIT, it sets the vector size to 128 bit unless AVX2 is supported.  So even though all float operations could be done at 256-bit width on processors supporting AVX and not AVX2, those processors still get `Vector<float>.Count == 4`.  `Vector<float>.Count == 8` really does imply AVX2 support.

br3aker: @saucecontrol that's super weird, thanks for clarifying. I've decided to implement Vector8 and avx/avx2 flags separately.

saucecontrol: The short explanation is that they map `Vector<T>` as a whole to either `TYP_SIMD16` or `TYP_SIMD32` in the JIT.  It can only be mapped to `TYP_SIMD32` if all operations on all `T` are supported, so that means AVX2.

The JIT code is really fascinating if you ever get a chance to dig into it.  The code that sets `Vector<T>` size is here, for example: https://github.com/dotnet/runtime/blob/main/src/coreclr/jit/ee_il_dll.cpp#L307-L345

br3aker: Had yet another question about another `Vector<T>` detail but now I can have a truly full picture myself, super interesting. Thanks a lot!

br3aker:
@antonfirsov added `Vector8` & `Vector4` flags, fixed tests with that flag usage, added docs.

For some unknown reason `RemoteExecutor` fails to execute some tests:
https://github.com/SixLabors/ImageSharp/pull/1657/checks#step:10:225

Testable method contains 2 implmentations: Avx2 & Sse2
https://github.com/SixLabors/ImageSharp/blob/5966a2eddfddef39187322b994fdfa032fcdb131/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs#L756-L760

I implemented separate tests for both paths, and Avx2 path test chooses Sse2 path for some reason even though `RemoteExecutor` is launched without `DisableAvx2`:
https://github.com/SixLabors/ImageSharp/pull/1657/checks#step:10:234

Line 822 is a Sse2 path which should not be taken in that test:
https://github.com/SixLabors/ImageSharp/blob/5966a2eddfddef39187322b994fdfa032fcdb131/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs#L822

 **And** `Avx2.IsSupported` is true for testing environment so attributes are working just fine:
https://github.com/SixLabors/ImageSharp/pull/1657/checks#step:10:246

tests/ImageSharp.Tests/TestUtilities/Attributes/Intrinsics/IntrinsicTheoryAttribute.cs
br3aker: @JimBobSquarePants sorry for namedropping but can you give your opinion on this?

JimBobSquarePants:
@br3aker ill pull down your branch and have a look ASAP.

br3aker:
@JimBobSquarePants forgot to mention - it works just fine on local my machine, all tests are passed.

tests/ImageSharp.Tests/Common/SimdUtilsTests.cs
```diff
@@ -212,7 +212,7 @@ static void RunTest(string serialized)
             FeatureTestRunner.RunWithHwIntrinsicsFeature(
                 RunTest,
                 count,
-                HwIntrinsics.DisableSSE41);
+                HwIntrinsics.AllowAll);
```
JimBobSquarePants, line 215: Rules are inherited. 

HwIntrinsics.DisableSSE41 will disable anything SSE41 and above . That includes AVX, AVX2 etc.

br3aker: That is super confusing considering this from jit source:
```
// Enable AVX instruction set for wide operations as default. When both AVX and SSE3_4 are set, we will use the most
// capable instruction set available which will prefer AVX over SSE3/4.
```

Further looking into jit initialization step for intrinsics it become clear that avx/avx2 check comes only after sse42. I've actually seen some kind of a tree for intrinsics checks somewhere in disscussions or even issues maybe and can't find it. I'll try restore it or recreate from https://blog.weghos.com/coreclr/CoreCLR/jit/compiler.cpp.html and add it to the docs of the HwIntrinsics enum so this problem won't happen again.

Thanks for taking a look at this!

tests/ImageSharp.Tests/TestUtilities/Attributes/RuntimeFeatures/IntrinsicTestsUtils.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// Equivalent of Ssse3.IsSupported
+        /// </summary>
+        /// <remarks>Not supported in certain runtimes even if hardware supports it</remarks>
+        SSSE3 = 1 << 4,
+
+        /// <summary>
+        /// Equivalent of Sse41.IsSupported
+        /// </summary>
+        /// <remarks>Not supported in certain runtimes even if hardware supports it</remarks>
+        SSE41 = 1 << 5,
```
antonfirsov, line 305: [Naming] The BCL naming conventions for enums are the same as for method or class names. These should be `Sse`, `Sse2`, `Sse41`, `Popcnt` etc. The test project's analyzers may be more relaxed on this rule, but in this case I don't see a reason to not conform them.

br3aker: Done!

tests/ImageSharp.Tests/Common/SimdUtilsTests.cs
```diff
@@ -221,9 +191,30 @@ static void RunTest(string serialized)
             FeatureTestRunner.RunWithHwIntrinsicsFeature(
                 RunTest,
                 count,
-                HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX2 | HwIntrinsics.DisableSSE41);
```
JimBobSquarePants, line 224: I can't see a test for when SSE41 is disabled now. The method falls back to SSE2 when it's not available.

br3aker:
This has the lowest priority atm so I'll just close it for now.

PR #1660, opened 2021-06-13
WIP. More Efficient MemoryAllocator

JimBobSquarePants:
### Prerequisites

### Description

A breaking (_due to the removal of unused configuration options_) rewrite of ArrayPoolMemoryAllocator to modelled on the design described at https://github.com/SixLabors/ImageSharp/issues/1596#issuecomment-821146358 by @antonfirsov 

- The allocator will use `ArrayPool<byte>.Shared` for buffers <= 1MB
- For > 1MB the allocator will use a custom pool that has a maximum allowed buffer length of 2MB. This pool uses the same trick  [`Gen2GcCallback` in `TlsOverPerCoreLockedStacksArrayPool`](https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/libraries/System.Private.CoreLib/src/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs#L201) to trim the pool under pressure.
- `Buffer2D<T>` will use 2MB for each discontiguous ([_did you know that was a Scottish word_](https://www.collinsdictionary.com/dictionary/english/discontiguous)?) chunk.

#### Things to consider
- <strike>For `Allocate<T>` only we should consider using unmanaged memory on supported platforms for > 2MB allocations using the approach described by @saucecontrol https://github.com/SixLabors/ImageSharp/issues/1596#issuecomment-822733344 </strike> Done!
- Remove all factory methods and potentially all configuration from the pool.
- We need tests for the GC trimming. Will check the runtime tests to see what they have.

#### Issues
<strike>Several Tiff encoding tests fail with the lower buffer chunk threshold set. This is due to `TiffBaseColorWriter<TPixel>` calling `Buffer2D<T>.GetSingleSpan()`. @brianpopow I'll need you or @IldarKhayrutdinov to help me there as I don't know enough about the format to do a fix. </strike> Fixed, thanks @brianpopow !!
https://github.com/SixLabors/ImageSharp/blob/381dff8640b721a34b1227c970fcf6ad6c5e3e72/src/ImageSharp/Formats/Tiff/Writers/TiffBaseColorWriter%7BTPixel%7D.cs#L82-L84

<strike>We have a failing test for `ResizeKernelMap` due to the use of `Buffer2D<T>.GetSingleMemory()`. @antonfirsov I'll need help there. </strike>. Fixed!
https://github.com/SixLabors/ImageSharp/blob/381dff8640b721a34b1227c970fcf6ad6c5e3e72/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs#L54-L55


src/ImageSharp/Memory/Allocators/Internals/UnmanagedBuffer{T}.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        private sealed class SafeHGlobalHandle : SafeHandle
+        {
+            private readonly int byteCount;
+
+            public SafeHGlobalHandle(int size)
+                : base(IntPtr.Zero, true)
+            {
+                this.SetHandle(Marshal.AllocHGlobal(size));
+                GC.AddMemoryPressure(this.byteCount = size);
```
br3aker, line 74: Should this be done? This would lead to a faster gen2 collection and object itself would behave like it's a managed allocation which I assume is not what we want to be happening. 

Microsoft docs states that ["The AddMemoryPressure and RemoveMemoryPressure methods improve performance only for types that exclusively depend on finalizers to release the unmanaged resources"](https://docs.microsoft.com/en-us/dotnet/api/system.gc.addmemorypressure?view=net-5.0). Which is not the case here as safe handle is used to prevent memory leaking only when user forgot to call dispose which is very unlikely to be caused by library code (this should be tested somehow to be honest).

Otherwise this would lead to a memory throttle as it is with current implementation.

src/ImageSharp/Memory/Allocators/Internals/Utilities.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+#if !SUPPORTS_BITOPERATIONS
+        /// <summary>
+        /// Returns the integer (floor) log of the specified value, base 2.
+        /// Note that by convention, input value 0 returns 0 since Log(0) is undefined.
+        /// Does not directly use any hardware intrinsics, nor does it incur branching.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        private static int Log2SoftwareFallback(uint value)
```
br3aker, line 61: Jpeg encoder uses similar logic for internal operations, this could potentially be placed somewhere in `Numerics.cs` class for reusage.

P.S.
Jpeg needs to calculate minimum bitsize of a given number which can be calculated via Log2DeBruijn logic. Not really a game changer but would be nice to have these under single class & single test fixture.

src/ImageSharp/Memory/Allocators/Internals/UnmanagedBuffer{T}.cs
saucecontrol: That advice is awkwardly worded.  In the case where a user fails to Dispose an object backed by an unmanaged buffer, it does rely on the SafeHandle's finalizer to release the memory, so it is advantageous to advise the GC that it might be able to reclaim the memory by doing its GC thing.  This should only result in more frequent gen2 GCs if the memory limits are being reached, which should only happen if the unmanaged buffer is long-lived or if the system is actually low on memory.  For properly Disposed ephemeral buffers, the GC will be aware of the added memory pressure but will also see it freed quickly, so no harm.

br3aker: Not sure that these buffers can be named 'ephemeral', 2mb is a lot (at least it looks like a lot for something temporal) which *should* lead to a lot of execution time spent on this memory. Primary idea I had while writing this is that even if user forgets to dispose memory backed by a handle it'll be freed by either gen0/gen1 collection or full gen2 collection when `OutOfMemoryException` kicks in.

While I'm not so sure about this now, I'm still concerned that this might be an overkill due to freachable queue. At least I'd want to benchmark this on something big like your parallel beeheads demo.

saucecontrol: unmanaged buffers will only be used for allocations that are over the pooled size limit.  They should be used once and then released, hence 'ephemeral'.  The question of whether `GC.AddMemoryPressure` is appropriate ultimately comes down to "is there any way a GC could free this memory?".  The answer here is yes, but only if the memory has been leaked because someone didn't dispose.  Unfortunately there's no way to know whether someone is *going to* leak; you can only tell after they've done it, when your finalizer runs.

What the MS docs should have said is something along the lines of "don't tell the GC about memory it couldn't possibly reclaim".  If the allocation and free were always 100% deterministic, there would be no point in telling the GC about them.  But since we can't know whether the GC *could* reclaim the memory, it's better to tell it about the allocation than to not.

src/ImageSharp/Memory/Allocators/Internals/Utilities.cs
JimBobSquarePants: That uses `BitOperations.LeadingZeroCount` does it not?

src/ImageSharp/Memory/Allocators/Internals/UnmanagedBuffer{T}.cs
br3aker: Once again, thanks for a deep dive! I've never looked at GC from "memory it can or cannot not claim" point of view.

src/ImageSharp/Memory/Allocators/Internals/Utilities.cs
br3aker: If `Lzcnt` intrinsic is supported - yes. But for fallback code it now uses LuT of 255 values with possible branching if value exceeds 255 only once - thus 16 *bits* is the maximum value vitsize this method would safely calculate. DeBruijn sequence is should be a bit faster and would eliminate the 16 bit maximum value constraint. Code would be a bit different, that table is what can be shared between jpeg & pool code. I'll open a PR with naming fix & algorithms after finishing my current open pr for testing.

br3aker: Oh wait I'm dumb, `BitOperations.LeadingZeroCount` already contains fallback code with deBruijn fallback implementation, totally forgot that we ended up using that instead of raw lzcnt intrinsic. Afaik that table is internal so yeah, no point in actually putting this code to `Numerics.cs` as only pool would use it.

JimBobSquarePants: Your maths is better than mine! 😀

Cool I’ll leave that to you then. Thanks!

br3aker: No problem! I'll try to do it today so you could use that in this pr before merging.

JimBobSquarePants: Lovely, thanks!

JimBobSquarePants:
@saucecontrol I don't know why the second graph isn't loading but I had a go at profiling. I'm fairly confident I messed things up.

image

Here's your original sample.

image

src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.cs
```diff
@@ -12,140 +13,131 @@ namespace SixLabors.ImageSharp.Memory
...
         /// </summary>
-        private ArrayPool<byte> normalArrayPool;
+        internal const int DefaultMaxArrayLengthInBytes = 2 * SharedPoolThresholdInBytes;
 
         /// <summary>
-        /// The <see cref="ArrayPool{T}"/> for huge buffers, which is not kept clean.
+        /// The default bucket count for <see cref="largeArrayPool"/>.
         /// </summary>
-        private ArrayPool<byte> largeArrayPool;
+        private const int DefaultMaxArraysPerBucket = 16;
```
antonfirsov, line 30: Since we utilize only one bucket now, an intensive parallel load will defer to the unmanaged stuff almost all the time. I would try benchmarking with different bucket counts.

JimBobSquarePants: I'm gonna need help writing those tests and improving others to ensure they work in CI.

src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.cs
```diff
@@ -12,140 +13,131 @@ namespace SixLabors.ImageSharp.Memory
...
             int bufferSizeInBytes = length * itemSizeBytes;
-            if (bufferSizeInBytes < 0 || bufferSizeInBytes > this.BufferCapacityInBytes)
+
+            IMemoryOwner<T> memory;
+            if (bufferSizeInBytes > this.MaxPooledArrayLengthInBytes)
             {
-                ThrowInvalidAllocationException<T>(length);
+                // For anything greater than our pool limit defer to unmanaged memory
+                // to prevent LOH fragmentation.
+                memory = new UnmanagedBuffer<T>(length);
```
antonfirsov, line 126: Would be great to see some metrics, how many times do we go here VS to the pool.
I can help with the EventSource stuff.

src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.cs
```diff
@@ -12,140 +13,131 @@ namespace SixLabors.ImageSharp.Memory
...
+        /// This matches the upper pooling length of <see cref="ArrayPool{T}.Shared"/>.
+        /// </summary>
+        private const int SharedPoolThresholdInBytes = 1024 * 1024;
 
         /// <summary>
-        /// The <see cref="ArrayPool{T}"/> for small-to-medium buffers which is not kept clean.
+        /// The default value for the maximum size of pooled arrays in bytes. 2MB.
         /// </summary>
-        private ArrayPool<byte> normalArrayPool;
+        internal const int DefaultMaxArrayLengthInBytes = 2 * SharedPoolThresholdInBytes;
```
antonfirsov, line 25: This is also a parameter we can tweak!

antonfirsov (commented):
In order to have a fair comparison, you need to benchmark the old and the new allocator on the same PC. You can make a copy of the old allocator class in a different namespace for simplicity.

I would try to tweak different parameters and see how the results change.

src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.cs
JimBobSquarePants: I know the BMP codec hits this a few times. I saw that when I made a mistake in the `UnmanagedBuffer<T>` type. Hopefully my EventSource stuff works ok.

src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.cs
JimBobSquarePants: Yep. Time is against me now though, it's nearly 3am.

Directory.Build.props
```diff
@@ -26,5 +26,41 @@
...
+    </When>
+    <When Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
+      <PropertyGroup>
+        <DefineConstants>$(DefineConstants);SUPPORTS_CRITICALFINALIZER</DefineConstants>
+      </PropertyGroup>
+    </When>
+    <When Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','netcoreapp3.1'))">
+      <!--NETCORE 3.1. NET5.0, and future versions will fallback to this as the closest target.-->
+      <PropertyGroup>
+        <DefineConstants>$(DefineConstants);SUPPORTS_BITOPERATIONS</DefineConstants>
```
JimBobSquarePants, line 59: I forgot this was already defined.

src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.CommonFactoryMethods.cs
```diff
@@ -8,69 +8,44 @@ namespace SixLabors.ImageSharp.Memory
...
-                DefaultMaxPooledBufferSizeInBytes,
-                DefaultBufferSelectorThresholdInBytes,
-                DefaultLargePoolBucketCount,
-                DefaultNormalPoolBucketCount,
-                DefaultBufferCapacityInBytes);
-        }
+            => new ArrayPoolMemoryAllocator(maxPooledArrayLengthInBytes: DefaultMaxArrayLengthInBytes);
 
         /// <summary>
-        /// For environments with very limited memory capabilities, only small buffers like image rows are pooled.
```
JimBobSquarePants, line 50: I want to remove these factory methods unless we are sure we can provide useful and accurate defaults

antonfirsov: We should even consider renaming the whole class TBH, I'd postpone figuring out these details until the very end.

saucecontrol:
Profiling results are promising so far.  Your total VirtualAlloc for the process is less than half, so the pooling is definitely doing something.  You can also see the switch from the classic GC sawtooth memory graph to something more chunky, reflecting the immediate free of your large unmanaged buffers.  And it looks like your GC pause times are shorter, although that's just eyeballing it -- PerfView would let you confirm.

More aggressive pooling will still lead to high peak usage since your baseline will be higher and you've still got to allocate large regions on top of that.  It'll be tough to find a balance there.  Could be that reducing the pooling works out better since you've made the large allocations cheaper now.  And even cheaper native allocations are coming in .NET 6 with https://github.com/dotnet/runtime/pull/54006 :)

saucecontrol:
BTW, as @antonfirsov pointed out, you can't compare results across machines, but just as a point of reference, the machine I ran my traces on was a quad-core (i7 6700K) with 32GB RAM.  More cores will stress more because it's using just a simple `Parallel.ForEach`, but having more RAM available also makes the GC lazier, which is likely a contributing factor in the higher peaks on my graph.  The really interesting question will be what happens in a memory-constrained environment (32-bit process or container or such).

antonfirsov:
> Your total VirtualAlloc for the process is less than half, so the pooling is definitely doing something. You can also see the switch from the classic GC sawtooth memory graph to something more chunky, reflecting the immediate free of your large unmanaged buffers

I'm afraid the only explanation for this is the utilization of `ArrayPool<byte>.Shared` for small arrays:
The large array pool is only used to allocate arrays of 2MB (we are building discontiguous buffers out of them), and there are only [16 arrays per-bucket](https://github.com/SixLabors/ImageSharp/blob/73bd60a23319826b08a59764b1a5e6d38c8bbf96/src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.cs#L30). As a result we are only pooling 32 MB for large buffers

When the pool is full, the current code actually defers to GC (my my https://github.com/SixLabors/ImageSharp/pull/1660#discussion_r651108437 was incorrect). This is not what we wanted, we should use `UnmanagedBuffer` instead of allocating a `new byte[]` when `GcAwareConfigurableArrayPool` is full.

Re benchmarking:
- We should also measure **throughput** (~total processing time for a bee heads run)
- We should systematically compare the metrics for different parameter sets (`maxArraysPerPoolBucket`, `maxContiguousArrayLengthInBytes`) to find the optimal balance of memory usage and throughput
- As @saucecontrol also mentioned, checking the behavior for memory constrained is key

@JimBobSquarePants I can help out in the weekend with code/benchmarks.

JimBobSquarePants:
>we should use `UnmanagedBuffer` instead of allocating a new `byte[]` when `GcAwareConfigurableArrayPool` is full.

@antonfirsov That is problematic for returning an `IManagedByteBuffer` since we would then not have access to the array  I'll have to toggle the behavior.

saucecontrol:
> As a result we are only pooling 32 MB for large buffers.
> When the pool is full, the current code actually defers to GC 

Yeah, just had a look through the code and I'm seeing that now.  So it could be that the reduced GC times as well as the higher steady-state commit size are a side effect of the fact the buffers are of uniform size.  That'll give you better reuse of heap space already committed by GC and reduced fragmentation.

So that's already a decent improvement but could be quite a bit better when unmanaged buffers are actually getting used 😄

JimBobSquarePants:
>So that's already a decent improvement but could be quite a bit better when unmanaged buffers are actually getting used 😄

Gimme 10 minutes 😙

JimBobSquarePants:
@saucecontrol Latest results.

image

I switched the code to use unmanaged memory whenever the pool was exhausted or the requested buffer was too large. 
Absolute peak is 1701.219MB now but I'm intrigued as to why the sum amount is so much bigger though?

saucecontrol:
Hubba hubba.

So I think what's happening here is that with GC allocating the 2MB buffers, you end up re-using the LOH segment holes, so GC holds those `VirtualAlloc` blocks, effectively giving you a LOH segment pool.  Since at any given time your ArrayPool bucket is probably holding some combination of old and young buffers, they end up scattered among several LOH segments, keeping those from being freed, and explaining your higher baseline memory in the PR graph from above.  With unmanaged, you'll be seeing the full allocation count.

I reckon once https://github.com/SixLabors/ImageSharp/issues/1597 is taken care of, you'll be closer to the neighborhood of what ImageMagick and FreeImage do memory-wise on that test.

tests/ImageSharp.Tests/Memory/Allocators/ArrayPoolMemoryAllocatorTests.cs
```diff
@@ -186,7 +186,15 @@ static void RunTest()
...
 
-                Assert.False(StaticFixture.CheckIsRentingPooledBuffer<SmallStruct>(2 * 4096 * 4096));
+                if (TestEnvironment.IsWindows)
+                {
+                    // TODO: We should have an attribute for this kind of stuff.
+                    // This test passes locally but not in the UNIX CI.
+                    // This could be due to the GC simply returning the same buffer
+                    // from unmanaged memory but this requires confirmation.
+                    Assert.False(StaticFixture.CheckIsRentingPooledBuffer<SmallStruct>(2 * 4096 * 4096));
+                }
```
br3aker: Can be easily done via https://github.com/SixLabors/ImageSharp/pull/1657
But not sure if `RuntimeFeatures.Windows` would not look ugly alongside things like `Avx` though

JimBobSquarePants:
>So I think what's happening here is that with GC allocating the 2MB buffers, you end up re-using the LOH segment holes, so GC holds those VirtualAlloc blocks, effectively giving you a LOH segment pool. 

I think that's why `LargeBuffersAreNotPooled_OfBigValueType` must be failing on non-Windows platforms actually. When we compare memory the location seems to be identical but they are not expected to be.

Seeing some weird issues in the build now. `RemoteExecutor` has suddenly decided to stop working in Ubuntu and I cannot see a reason why. Getting OOM in Windows also when doing a gif comparison. 2 images at 92 frames each but that used top be ok before.

tests/ImageSharp.Tests/Memory/Allocators/ArrayPoolMemoryAllocatorTests.cs
JimBobSquarePants: Arcade has tons of stuff like that actually we could steal.
https://github.com/dotnet/arcade/blob/12956a100f3d7897456f8d1b8572b2329478aa52/src/Microsoft.DotNet.XUnitExtensions/src/Attributes/PlatformSpecificAttribute.cs

antonfirsov:
> So I think what's happening here is that with GC allocating the 2MB buffers, you end up re-using the LOH segment holes

The code is still pooling only [16*2=32MB](https://github.com/SixLabors/ImageSharp/blob/e9e19483e33755ade68b989a6dec5b53d03fc986/src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.cs#L30), that's a tiny part of the 1.7GB peak.  What we see here must be the result of building disconitguos buffers out of 2MB **unmanaged** memory blocks.

@JimBobSquarePants can we run an experiment trying to pool significantly more? (128MB, 512MB, 1024MB => `maxArraysPerPoolBucket=64, 256, 512`)

If we get better results with unmanaged buffers, we may consider dropping the large pool entirely, though I still have concerns on the reliability SafeHandle finalizers. AFAIK a single unhandled exception in any user finalizer will prevent the finalizer queue to finish, leading to actual memory leaks.

JimBobSquarePants:
Marking this as ready to review. There's a lot of testing and configuration to be done but I think the functionality is pretty much where we need it to be.

antonfirsov:
> the functionality is pretty much where we need it to be.

We still need to figure out how much value is there from pooling, if it has no real effect, that's a game changer on the current implementation and API shape.

We also need metrics on throughput, the new memory access patterns may have impact on cache utilization.

JimBobSquarePants:
I’ll leave the memory profiling to the more proficient. I need to take a break, struggling with jet lag

tests/ImageSharp.Tests/Memory/Allocators/ArrayPoolMemoryAllocatorTests.cs
br3aker: Oh that's like the exact thing you wanted!
Strange that they didn't implement something like that for intrinsics.

antonfirsov:
> I need to take a break, struggling with jet lag

TBH it's also a very difficult time for me now. There's plenty of work left IMO, I recommend to slow down, and to declare it as a marathon rather than a sprint, there's no good in doing it in rush. Will see what I can do on Saturday.

saucecontrol:
> The code is still pooling only 16*2=32MB, that's a tiny part of the 1.7GB peak. What we see here must be the result of building disconitguos buffers out of 2MB unmanaged memory blocks.

@antonfirsov Correct.  I was attempting to explain why the profile of the initial PR showed much lower total VirtualAlloc numbers with much higher peak and baseline memory.  The switch to unmanaged makes every buffer allocation an actual allocation, giving a higher total VirtualAlloc.  But those allocations are released immediately, keeping the instantaneous committed memory lower.

When falling back to unmanaged allocations, keeping the 2MB discontiguous buffer strategy will be a negative perf-wise in that `AllocHGlobal` is comparatively expensive.  It would be better to keep those allocations contiguous, but that looks like that would take a pretty big refactor.  More tuning is needed in either case.

> I still have concerns on the reliability SafeHandle finalizers. AFAIK a single unhandled exception in any user finalizer will prevent the finalizer queue to finish

That's true, but only because an unhandled exception in a finalizer will crash the process. 😆 

Your finalizers in particular will only be calling `LocalFree` (by way of `Marshal.FreeHGlobal`), which will always succeed provided the handle you give it was created by `LocalAlloc`, and `GC.RemoveMemoryPressure`, which only throws for invalid args.

JimBobSquarePants:
>@JimBobSquarePants can we run an experiment trying to pool significantly more? (128MB, 512MB, 1024MB => maxArraysPerPoolBucket=64, 256, 512)

@antonfirsov I need to better understand what we want here. As I see it the larger the pool the more aggressively we will have to trim the pool as it will end up holding onto memory again. With our new logic we'd end up adding 1024MB arrays to that pool and almost nothing would go to unmanaged memory. 

>When falling back to unmanaged allocations, keeping the 2MB discontiguous buffer strategy will be a negative perf-wise in that AllocHGlobal is comparatively expensive. It would be better to keep those allocations contiguous, but that looks like that would take a pretty big refactor. More tuning is needed in either case. 

@saucecontrol Actually, it's pretty simple as I recall. I think if we simply adjusted `MaxContiguousArrayLengthInBytes` to a larger number then `MemoryGroup<T>Allocate` will request larger buffers.

saucecontrol:
> I think if we simply adjusted `MaxContiguousArrayLengthInBytes` to a larger number then `MemoryGroup<T>Allocate` will request larger buffers.

Yeah, but that would mean larger managed buffer requests too, which would then always fail, no?  Would be a quick way to test whether going all unmanaged would be ok, though.

One more note on perf around that...

In addition to the fact `AllocHGlobal` is expensive, allocating any finalizable object is also expensive.  The GC's finalization queue is global and, therefore, takes a lock to update it when you `new` a finalizable object and when you `SuppressFinalize`, because these don't happen during a GC pause.  That's in addition to it always taking the GC's slow allocation path rather than the fast bump allocator usually used for small objects.  It's worth looking into whether those things are in part responsible for the more rough CPU usage graph in the latest trace.  Of course bigger chunks would mitigate that as well, so that might show up after a quick change to `MaxContiguousArrayLengthInBytes`

JimBobSquarePants:
>Yeah, but that would mean larger managed buffer requests too, which would then always fail, no? Would be a quick way to test whether going all unmanaged would be ok, though.

Ah no. If the requested amount is larger than the pool maximum of 2MB or if the pool is exhausted then we defer to unmanaged buffers. 

https://github.com/SixLabors/ImageSharp/blob/a22b0cb1aef2ecea37d902cf786d9980a61d00f2/src/ImageSharp/Memory/Allocators/Internals/GCAwareConfigurableArrayPool%7BT%7D.cs#L107-L157

https://github.com/SixLabors/ImageSharp/blob/a22b0cb1aef2ecea37d902cf786d9980a61d00f2/src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.cs#L133-L152

saucecontrol:
Yeah, that's what I meant.  `MemoryGroup<T>Allocate` would request chunks > 2MB, which can never be served by a managed bucket so would always fall through to managed.  As it is now, it will consume as many 2MB managed chunks as it can before falling through to unmanaged (this may or may not actually be a good thing -- more profiling is needed)

JimBobSquarePants:
>What I was picturing was something that requested 2MB (or whatever you pick for your max managed chunk size) at a time from the managed pool, and then when that returns null, request all the rest in one unmanaged allocation. That would require moving that abstraction out of the allocator, though.

Yeah, that get's a bit iffy. I'd really like to keep everything there. 

Here's what happens if I change the contiguous length to 24MB

image

Vs the current PR state
image

Looks like CPU takes a hit there.

saucecontrol:
Ouch.  Yeah, 24MB is too chunky.  You'd allocate 48MB when you 25MB is requested, which shows in your total VirtualAlloc number jumping way up again.  That's why it would be better for the MemoryGroup to know it's making an unmanaged 'rental' so it can size it exactly to what it needs.

There's a balance there somewhere.  Unfortunately it'll be a lot of testing.

JimBobSquarePants:
>That's why it would be better for the MemoryGroup to know it's making an unmanaged 'rental' so it can size it exactly to what it needs.

I think we can expose the required properties via the allocator interface easily enough. We're breaking it anyway.

src/ImageSharp/Memory/Allocators/Internals/GCAwareConfigurableArrayPool{T}.cs
```diff
@@ -0,0 +1,504 @@
...
+                {
+                    this.spinLock.Enter(ref lockTaken);
+
+                    if (this.index < buffers.Length)
+                    {
+                        buffer = buffers[this.index];
+                        buffers[this.index++] = null;
+                        allocateBuffer = buffer == null;
+                    }
+                }
```
br3aker, line 341: Calculating if buffer was null can be done outside of lock scope: `if(buffer == null) {}`
```suggestion
                bool lockTaken = false;
                try
                {
                    this.spinLock.Enter(ref lockTaken);

                    if (this.index < buffers.Length)
                    {
                        buffer = buffers[this.index];
                        buffers[this.index++] = null;
                    }
                }
```

JimBobSquarePants: That's a straight copy from the runtime 

https://github.com/dotnet/runtime/blob/a02b668214e9711cb3d5f426f085601698a24b92/src/libraries/System.Private.CoreLib/src/System/Buffers/ConfigurableArrayPool.cs#L198-L217

br3aker: Note: I don't know if you can suggest changes from different places so **do not** commit this, line 353 must be changed  to `buffer == null` for this to work.

br3aker: @JimBobSquarePants hm, I may be wrong of course but we can easily do that outside of the lock scope, no? That comparison is nothing but with spinlocks better be fast than asleep :P

src/ImageSharp/Memory/Allocators/Internals/GCAwareConfigurableArrayPool{T}.cs
```diff
@@ -0,0 +1,504 @@
...
+                        this.spinLock.Exit(false);
+                    }
+                }
+
+                // While we were holding the lock, we grabbed whatever was at the next available index, if
+                // there was one. If we tried and if we got back null, that means we hadn't yet allocated
+                // for that slot, in which case we should do so now.
+                if (allocateBuffer)
+                {
+                    if (this.index == 0)
```
br3aker, line 355: First of all, this is not thread safe, `this.index` variable is a shared state that must be altered only in locked scope.

`this.index` would never be 0 in this piece of code. If bucket is full (i.e. all buffers are not rented), `this.index` would be equal to zero. After exactly one rent in locked scope `this.index` would be incremented thus at `if(this.index == 0)` line index would always be >= 1.

br3aker: This can be fixed like this:
```csharp
bool lockTaken = false;
int takenIndex = -1;
try
{
    this.spinLock.Enter(ref lockTaken);

    if (this.index < buffers.Length)
    {
        buffer = buffers[this.index];
        buffers[this.index] = null;
        takenIndex = this.index++;
    }
}
finally
{
    if (lockTaken)
    {
        this.spinLock.Exit(false);
    }
}

if (buffer == null)
{
    if (takenIndex == 0)
    {
        // Stash the time the first item was added.
        this.firstItemMS = (uint)Environment.TickCount;
    }
// ...
```

br3aker:
2D buffer renting logic can be optimized.

From `Buffer2D<T>.Allocate`:
```csharp
// ...
var buffers = new IMemoryOwner<T>[bufferCount];
for (int i = 0; i < buffers.Length - 1; i++)
{
    buffers[i] = allocator.Allocate<T>(bufferLength, options);
}

if (bufferCount > 0)
{
    buffers[buffers.Length - 1] = allocator.Allocate<T>(sizeOfLastBuffer, options);
}
// ....
```

If we try to allocate a 2D buffer for some images from separate threads at the same time - those would be paused on spinlock with extremely fast response times which can lead to LOH fragmentation as we don't know which image/resource would be freed in what order and in what times:
```
Bucket:
[0]: free
[1]: free
[2]: free
[3]: free
[4]: free
...
```

Thread1 requests 2 buffers
Thread2 requests 3 buffers, let's say thread2 wins:
```
time quant 1
Bucket:
[0]: thread2 taken
[1]: free
[2]: free
[3]: free
[4]: free
...
```
```
// spin locks can be very fast considering nothing big happens inside locked code
// while thread2 doing some bureaucracy returning from allocator method & and calling it again with some checks again
// thread1 could easily have taken the lock which can lead to allocator fragmentation
time quant 2
Bucket:
[0]: thread2 taken
[1]: thread1 taken
[2]: free
[3]: free
[4]: free
...
```
```
// possible example
time quant 3-5
Bucket:
[0]: thread2 taken
[1]: thread1 taken
[2]: thread2 taken
[3]: thread1 taken
[4]: thread2 taken
...
```

And that's only 2 threads, add 8 or even 16 cores, async/await - boom.

Solution proposal: ability to rent a number of buffers at a single call. This can be easily implemented in the custom `GCAwareConfigurableArrayPool` but .net TLS pool isn't designed to support this so it will operate the same way as it is now, unfortunately. This would still be very fast & spin locks won't be hurt. Renting logic can be simplified to bucket buffer index increase, with multiple buffers rent this can be can as simple as `bucket.nextBufferIndex += numberOfBuffersToRent` with some checks. If we requested 10 buffers but got only 5 as buffer is exhausted during the process - we can allocat remaining slots on the fly and these buffers will participate at the returning stage. This might not be a good thing - out buffer is jagged with 2 halves from pool and some random LOH memory, returning them as a full chunk could create a fragmented space in the bucket. 

This can also possibly decrease fragmention at the bucket first allocation. Current implementation allocates new buffers if they were rented as null, batch renting would lead to 3 subsequent allocations which have a higher chance for being a continuous memory which also should help with fragmentation.

--------

This can be proofed via some random 4k image:
Custom log from `MemoryGroup.Allocate`:
```
Allocating 7478016 bytes via 4 buffers
Allocating 14755840 bytes via 8 buffers
```
Even 2 threads with same pictures would mess up the ordering by a lot:
image
Note: this is an array of which buffer was taken by which thread via `Thread.ManagedThreadId`. 
Proposed solution should chnage this behaviour to something like: `5 5 5 5 6 6 6 6 5 5 5 5 5 5 5 5 6 6 6 6`. But the game changer is that they are returned in batches.

P.S.
Sorry for a giant wall of text, couldn't come up with something more compact :(

src/ImageSharp/Memory/Allocators/Internals/GCAwareConfigurableArrayPool{T}.cs
JimBobSquarePants: Good catch. It's also in the wrong method. Will update.

JimBobSquarePants:
I definitely broke something in NET FX when I removed `IManagedByteBuffer`. The tests freeze up and never report completion. Will revert to the commit before that and try again.

antonfirsov (changes requested):
Don't get me wrong, this is amazing work, it's great to see a proof that unmanaged allocations can reduce memory peaks by more than 50%. However, this is a massive change in the core engine, and the current code and the status of validation is still very far from a state I would consider mergeable.

Here is the list of major problems we need to solve:

1. As a start, we need to setup systematic benchmarking, which can help us getting the following metrics for any configuration in a quantitative/comparable manner enabling **data-driven decision making** to shape the design and fine-tune the final perf parameters:
  - Peak memory usage
  - Average memory usage
  - CPU utilization
  - Throughput (~time taken to process all the bee heads)

We should be able to put these values into a table so we can easily compare benchmark results for different parameter sets.

2. With the current tuning, there is almost no pooling happening when it comes to large buffers, meaning that the current PR behavior does not justify the presence of the complex array pooling & trimming logic and the `ArrayPoolMemoryAllocator` name / API shape. We need to see how this works with pooling disabled + several more aggressive pooling behaviors.

> With our new logic we'd end up adding 1024MB arrays to that pool and almost nothing would go to unmanaged memory.

I did not propose 1024 as a final value. It is an extreme config in general, but not for the bee heads benchmark, where the memory usage [seems to fluctuate around 1.2GB](https://user-images.githubusercontent.com/385879/122156629-42a21480-ce61-11eb-8052-3949fef0b33f.png). I'm also more curious less aggressive, more reasonable pooling settings, but I see no reason for not getting all the data points that help us understanding the whole picture and make better decisions.

3. As pointed out in https://github.com/SixLabors/ImageSharp/pull/1660#issuecomment-862839513, we need different sizes for the contiguous blocks coming pooled arrays VS unmanaged buffers. There are several ways to implement this, probably the most straightforward thing is to move the code in [`MemoryGroup.Allocate`](https://github.com/SixLabors/ImageSharp/blob/0804c2993041035a71a3805ca9f6fc14aa51b833/src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroup%7BT%7D.cs#L75-L130) to a virtual method on `MemoryAllocator`, that could be overridden.

4. Using the `ArrayPool<T>` abstraction and the entire `GCAwareConfigurableArrayPool<T>` class to pool 2MB buffers doesn't bring us value, since we are utilizing only one bucket of that pool. We should implement a custom pool class dedicated to the concern of pooling uniform arrays. It should be relatively easy to refactor it from `GCAwareConfigurableArrayPool<T>.Bucket`.

5. We should periodically trim the pool by by a certain factor even when there is low memory pressure. This would address the concern of retaining memory unnecessarily (pointed out in #1590 and other user complaints), and also **enable us to pool much more** when there is high load.

6. We need extensive test coverage to validate our assumptions regarding the utilization of the pools and unmanaged buffers. We also need to test trimming. Unfortunately, it would be too expensive to do it with regular Xunit runs, but we can define a local-only test, that delivers & validates the logs/metrics proving the trimming is happening in the way we expect.

I understand this is enormous amount of work, but it was always the case for all the previous PR-s refactoring the memory management engine. I don't see a reason to lower our quality criterias in this case, and omit any of the points above, especially that we are about to introduce a breaking change, and we want to prevent future breaking changes. Personally, I want to start working on point 1., but even that alone may take several evenings, that makes it very hard to give an ETA. If we feel like the allocator work may block 1.1 for too long, we should focus on #1597 first, since it will bring even bigger improvement with a lower development cost.

JimBobSquarePants:
@antonfirsov I actually agree with all the above. I'm running before we can walk without the relevant experience and suffering as a result. #1597 should immediately benefit us for V1.1.

What I'm actually going to do is close this and instead introduce a few smaller PRs to do some sanitation work which will allow the allocator changes to be made more easily.

- Rename all `GetSpan`, `GetMemory` that require a single span to add a `Dangerous` prefix and remove all usage that is not required.
- Multiple PRs to Migrate any calls that use `AllocateManageByteBuffer` to use `Allocate<byte>` (_Turns out it isn't that that is breaking thing, the cause is still undetermined_)
- Optimize the Gif encoder to allow us to clear and reuse a cache to save memory churn in multiframe images.

antonfirsov:
I spent some time today trying to figure out how to get the desired metrics out of .ETL files, but I realized there is no easy way that is worth the efforts. This should not block systematic comparison, but will make it even more of a chore work 😞

PR #1663, opened 2021-06-15
Numerics.cs Log2 implementation

br3aker:
### Prerequisites

### Description
A rather small PR with Log2 implementation for `Numerics` class. Some internals (deBruijn sequence) is shared by both new allocator from https://github.com/SixLabors/ImageSharp/pull/1660 and existing jpeg encoder. Also added tests for Log2 as this is quiet important for allocator not to fail.


JimBobSquarePants:
Looks like the software fallback is incorrect.
https://github.com/SixLabors/ImageSharp/pull/1663/checks?check_run_id=2825926233#step:10:278

src/ImageSharp/Formats/Jpeg/Components/Encoder/HuffmanScanEncoder.cs
```diff
@@ -388,5 +388,34 @@ private void FlushInternalBuffer()
...
+        /// Calculates how many minimum bits needed to store given value for Huffman jpeg encoding.
+        /// This method does not follow the standard convention - it does not support input value of zero.
+        /// </summary>
+        /// <remarks>
+        /// Passing zero as input value would result in an undefined behaviour.
+        /// This is done for performance reasons as Huffman encoding code checks for zero value, second identical check would degrade the performance in the hot path.
+        /// </remarks>
+        /// <param name="value">The value.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static int GetHuffmanEncodingLegth(uint value)
```
JimBobSquarePants, line 402: Length

src/ImageSharp/Formats/Jpeg/Components/Encoder/HuffmanScanEncoder.cs
```diff
@@ -388,5 +388,34 @@ private void FlushInternalBuffer()
...
+
+            // BitOperations.Log2 implementation also checks if input value is zero for the convention
+            // As this is a very specific method for a very specific Huffman encoding code
+            // We can omit zero check as this should not be invoked with value == 0 and guarded in debug builds
+            return 32 - System.Numerics.BitOperations.LeadingZeroCount(value);
+#else
+            // On the contrary to BitOperations implementation this does follow the convention and supports value == 0 case
+            // Although it's still won't be called with value == 0
+            // As these implementations behave differently for the value == 0 case it's documented as undefined behaviour
+            return Numerics.Log2SoftwareFallback(value) + 1;
```
JimBobSquarePants, line 417: I don't think we should expose that as a separate method. 

I'd stick compiler conditionals in a `Numerics.Log2` method instead.

src/ImageSharp/Formats/Jpeg/Components/Encoder/HuffmanScanEncoder.cs
br3aker: Yes, fixed in upcoming push.

src/ImageSharp/Formats/Jpeg/Components/Encoder/HuffmanScanEncoder.cs
br3aker: I completely messed up with those comments as I completely forgot that huffman encoder can encounter a zero value during dc part encoding. Realistically, log2 is the solution for this method but `32 - lzcnt` is faster hardware-wise as there no if-check if lzcnt is supported. Or maybe I didn't understood correctly what you said about polyfill.

br3aker: @JimBobSquarePants sorry for namedropping, I'm not sure github sends notifications without tagging, at least I've seen your review by a chance.

You meant borrowing code from runtime source code? `BitOperations.LeadingZeroCount` is a polyfill itself, it has fallback code if hardware doesn't support `lzcnt`. It's really specific to jpeg right now plus fallback code assumes that input value is constrained in certain way, placing it in Numerics might not be good considering that.

JimBobSquarePants: What I meant was implement a `Numerics.LeadingZeroCount(uint value)` method so that your calling code would be simplified.

Something like this. (_I think this is correct, patched from runtime source_).

```c#
public static int LeadingZeroCount(uint value)
{
#if SUPPORTS_BITOPERATIONS
    return BitOperations.LeadingZeroCount(value);
#else
    // Software fallback
    // Unguarded fallback contract is 0->31, BSR contract is 0->undefined
    if (value == 0)
    {
        return 32;
    }

    return 31 ^ Log2SoftwareFallback(value);
#endif
```

br3aker: Oh, now I got it, thanks!
That would lead to unnecesarry operations in jpeg hot path code - I've updated correct implementation, all tests passed (build failed on some strange decoding test). Worth doing only if it's needed somewhere else in the lib,

src/ImageSharp/Formats/Jpeg/Components/Encoder/HuffmanScanEncoder.cs
```diff
@@ -388,5 +388,40 @@ private void FlushInternalBuffer()
...
+            // if 0 - return 0 in this case
+            // else - return log2(value) + 1
+            //
+            // Hack based on input value constaint:
+            // We know that input values are guaranteed to be maximum 16 bit large for huffman encoding
+            // We can safely shift input value for one bit -> log2(value << 1)
+            // Because of the 16 bit value constraint it won't overflow
+            // With that input value change we no longer need to add 1 before returning
+            // And this eliminates need to check if input value is zero - it is a standard convention which Log2SoftwareFallback adheres to
+            return Numerics.Log2SoftwareFallback(value << 1);
```
JimBobSquarePants, line 423:
```suggestion
            return Numerics.Log2(value << 1);
```

JimBobSquarePants: `Log2SoftwareFallback` should be private.

src/ImageSharp/Formats/Jpeg/Components/Encoder/HuffmanScanEncoder.cs
JimBobSquarePants: Ah ok, so you're concerned about the branching in older target frameworks. Ok, make sense.

JimBobSquarePants:
Just to note. I've fixed the StyleCop issues in master so when you update from there your warnings will now be build errors.

br3aker:
Yeah I've already encountered one of those, good to have it now!

src/ImageSharp/Formats/Jpeg/Components/Encoder/HuffmanScanEncoder.cs
br3aker: LGTM!
I'll also mark Log2SoftwareFallback as private cause jpeg encoder is the only thing that needed access to it, everything else should call Log2 polyfill.

br3aker:
There's definately something going on with DecodeBaselineJpeg & DecodeProgressiveJpeg tests. There are two different fails before and after the commit that didn't change anything in code-wise:
Before: https://github.com/SixLabors/ImageSharp/pull/1663/checks?check_run_id=2826376023#step:10:758
After: https://github.com/SixLabors/ImageSharp/pull/1663/checks?check_run_id=2826708693#step:10:25

Furthermore, both runs had exactly 1 failing test.

@JimBobSquarePants is it possible to re-run exactly one target? Need to be sure if next time it'll be same failing test or not.

JimBobSquarePants:
@br3aker The new fail is an issue we keep having with our Magick reference decoder behaving badly. I think it's due to memory pressure and I'm hoping that once we get our allocator fixes in they'll disappear. Frustratingly you cannot currently re-trigger an individual job.

P.S your PR branch needs to be updated from master.

br3aker:
@JimBobSquarePants as I recall correctly, there was a problem with Magick decoder during jpeg optimization PR testing but it was throwing `OutOfMemory`. Anyways, previous fail is also very strange, I checked stacktrace, there's no way it was caused by either this pr or memory issues. Hopefully this would pass and get merged so I'll start working on jpeg decoder stuff & some more encoder optimizations, I'll possibly take a look at those tests.

JimBobSquarePants (approved):
Great stuff!

PR #1662, opened 2021-06-14
TiffEncoder: Avoid buffer2D.GetSingleSpan() and use GetPixelRowSpan instead

brianpopow:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

As made clear with #1660 the usage of `buffer2D.GetSingleSpan()` is problematic and not always guaranteed to return a complete span. This **PR** will change the Tiff Encoder to avoid the usage of `buffer2D.GetSingleSpan()` and use `GetPixelRowSpan` instead.

Issue #1625, opened 2021-05-11
Resizing operation throws InvalidOperationException with exotic image resolution


br3aker:
- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Custom debug logs from `MemoryGroup<T>.Allocate(...)` call during steps stated below:
```
// Loading & decoding initial image from disk, 10x10=100 pixels - ok
// Minimal contiguous memory line 40 bytes - ok
Requesting Rgba32[100] (400 bytes) with stride of 40/1000 bytes
Allocation of 1 buffers 100 elements/400 bytes each

// Image cloning with bigger size for resizing, 15x15=225 pixels - ok
// Minimal contiguous memory line for 60 bytes - ok
Requesting Rgba32[225] (900 bytes) with stride of 60/1000 bytes
Allocation of 1 buffers 225 elements/900 bytes each

// Something related to resizing I guess - ok
Requesting Single[35] (140 bytes) with stride of 20/1000 bytes
Allocation of 1 buffers 35 elements/140 bytes each

// Something related to resizing I guess - ok
Requesting Single[35] (140 bytes) with stride of 20/1000 bytes
Allocation of 1 buffers 35 elements/140 bytes each

// That's where everything breaks due to non-contiguous memory allocation
Requesting Vector4[150] (2400 bytes) with stride of 160/1000 bytes
Allocation of 3 buffers 60 elements/960 bytes each
<-- After this call exception occurs -->
```

Last allocation from log allocates 150 Vector4 elements with a stride of 10 elements which is done by allocating 3 separate buffers of 60 elements each which fullfils requested "contiguous chunks of 10 elements each", so far so good.

Real problem lies inside [ResizeWorker](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeWorker.cs) class:
https://github.com/SixLabors/ImageSharp/blob/993f96eb7d90e05dbf2039c116d44be8fb081cd8/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeWorker.cs#L117-L118

[...]

PR #1674, opened 2021-06-19
Rename unsafe memory methods to use Dangerous prefix.

JimBobSquarePants:
### Prerequisites

### Description

We need to ensure we don't incorrectly use the single memory/span methods and prevent having to do work like #1662 
Should also help with refactoring #1625


antonfirsov (approved):
LGTM.

Issue #1675, opened 2021-06-20
Remove all uses of IManagedByteBuffer


JimBobSquarePants:
`IManagedByteBuffer` blocks any potential plans to use unmanaged memory as an optimization for our memory allocators. It's a legacy API used for older target frameworks but all use cases can now be replaced with `IMemoryOwner<byte>` since we have polyfilled the span APIs.

We should do so then remove the API.

PR #1676, opened 2021-06-20
Update AllocatePaddedPixelRowBuffer to use IMemoryOwner<byte>

JimBobSquarePants:
### Prerequisites

### Description

Refactors `MemeoryAllocatorExtensions.AllocatePaddedPixelRowBuffer` to return `IMemoryOwner<byte>`.

Touches #1675

### DIFF
--- a/src/ImageSharp/Common/Extensions/StreamExtensions.cs
+++ b/src/ImageSharp/Common/Extensions/StreamExtensions.cs
@@ -72,12 +72,6 @@ namespace SixLabors.ImageSharp
             }
         }
 
-        public static void Read(this Stream stream, IManagedByteBuffer buffer)
-            => stream.Read(buffer.Array, 0, buffer.Length());
-
-        public static void Write(this Stream stream, IManagedByteBuffer buffer)
-            => stream.Write(buffer.Array, 0, buffer.Length());
-
 #if !SUPPORTS_SPAN_STREAM
         // This is a port of the CoreFX implementation and is MIT Licensed:
         // https://github.com/dotnet/corefx/blob/17300169760c61a90cab8d913636c1058a30a8c1/src/Common/src/CoreLib/System/IO/Stream.cs#L742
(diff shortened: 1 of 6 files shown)


brianpopow (approved):

antonfirsov (approved):
Note that the whole purpose of `IManagedByteBuffer` is to avoid perf penalty on `<=netstandard2.1` coming from the buffer copy in the polyfilling methods. The idea is that any `MemoryAllocator` can implement `AllocateManagedByteBuffer` regardless of it's core logic, so strictly speaking it does not block switching to unmanaged allocators.

It's 2021 though, simplification of our codebase has higher priority than Framework performance, so I'm fine with the proposed change.

JimBobSquarePants:
@antonfirsov how would you access the Array property of an unmanaged buffer? That was the original trigger for me making the change.

antonfirsov:
@JimBobSquarePants you don't need to, because the returned buffer wouldn't be unmanaged. I mean: any allocator can return - for example - an array from `ArrayPool<byte>.Shared` in `.AllocateManagedByteBuffer`, even if the rest of it's allocators logic deals with unmanaged stuff. I know this is a bit counter-intuitive and ugly, but was the price paid of keeping Framework code optimal.

JimBobSquarePants:
I think our wires are a little crossed here. What I meant was that returning an `IManagedByteBuffer<T>` for a type like like [`UnmanagedBuffer<T>`](https://github.com/SixLabors/ImageSharp/blob/547614acb19a9d3e2e2ab718844fb82f2aae3d8a/src/ImageSharp/Memory/Allocators/Internals/UnmanagedBuffer%7BT%7D.cs) was impossible (_I think_) since there would be nothing to populate the `IManagedByteBuffer<T>.Array` property. This would mean we would be maintaining two separate memory allocation processes. 

I'm happy to make the performance sacrifice for Framework since the overarching benefit of an optimized allocator will outweigh a few ms in the codecs.

PR #1677, opened 2021-06-22
Remove IManagedByteBuffer Part 2

JimBobSquarePants:
### Prerequisites

### Description

More refactoring to use `IMemoryOwner<byte>`.

### DIFF
--- a/tests/ImageSharp.Tests/Formats/Bmp/BmpEncoderTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Bmp/BmpEncoderTests.cs
@@ -184,7 +184,13 @@ namespace SixLabors.ImageSharp.Tests.Formats.Bmp
             // The Magick Reference Decoder can not decode 4-Bit bitmaps, so only execute this on windows.
             if (TestEnvironment.IsWindows)
             {
-                TestBmpEncoderCore(provider, bitsPerPixel, supportTransparency: false);
+                // Oddly the difference only happens locally but we'll not test for that.
+                // I suspect the issue is with the reference codec.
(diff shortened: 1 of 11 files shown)


tests/ImageSharp.Tests/Formats/Bmp/BmpEncoderTests.cs
```diff
@@ -184,7 +184,13 @@ public void Encode_8BitGray_WithV3Header_Works<TPixel>(TestImageProvider<TPixel>
             // The Magick Reference Decoder can not decode 4-Bit bitmaps, so only execute this on windows.
             if (TestEnvironment.IsWindows)
             {
-                TestBmpEncoderCore(provider, bitsPerPixel, supportTransparency: false);
+                // Oddly the difference only happens locally but we'll not test for that.
```
antonfirsov, line 187: This is super weird. The reference decoder is [System.Drawing](https://github.com/SixLabors/ImageSharp/blob/5b5bb110ca0271f75eda400fdf7d22b919a8ab70/tests/ImageSharp.Tests/TestUtilities/TestEnvironment.Formats.cs#L72-L76), which should be the same thing on Framework and Core. Is it deterministic, failing for a single test run? What happens if you compare ImageSharp Framework vs Core output with Beyond Compare? If they are not identical, it might be worth to open an issue.

antonfirsov (approved):
LGTM.

tests/ImageSharp.Tests/Formats/Bmp/BmpEncoderTests.cs
JimBobSquarePants: Yep, deterministic. Super odd.

PR #1680, opened 2021-06-29
Added Image<TPixel> constructor from single ImageFrame

br3aker:
### Prerequisites

### Description
Added contructor from single frame used in some cases (and will be used in new version of jpeg decoder). Replaced multi-frame ctors with single-frame ctors where applicable.

### DIFF
--- a/src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter{TPixel}.cs
+++ b/src/ImageSharp/Formats/Tiff/Writers/TiffBiColorWriter{TPixel}.cs
@@ -26,7 +26,7 @@ namespace SixLabors.ImageSharp.Formats.Tiff.Writers
             : base(image, memoryAllocator, configuration, entriesCollector)
         {
             // Convert image to black and white.
-            this.imageBlackWhite = new Image<TPixel>(configuration, new ImageMetadata(), new[] { image.Clone() });
+            this.imageBlackWhite = new Image<TPixel>(configuration, new ImageMetadata(), image.Clone());
             this.imageBlackWhite.Mutate(img => img.BinaryDither(KnownDitherings.FloydSteinberg));
         }
 
--- a/src/ImageSharp/ImageFrameCollection{TPixel}.cs
+++ b/src/ImageSharp/ImageFrameCollection{TPixel}.cs
@@ -285,7 +285,7 @@ namespace SixLabors.ImageSharp
 
             this.frames.Remove(frame);
 
-            return new Image<TPixel>(this.parent.GetConfiguration(), this.parent.Metadata.DeepClone(), new[] { frame });
+            return new Image<TPixel>(this.parent.GetConfiguration(), this.parent.Metadata.DeepClone(), frame);
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


antonfirsov:
@br3aker I've seen your https://github.com/SixLabors/ImageSharp/issues/1597#issuecomment-870655388, just didn't have chance to react yet. I don't see the point in detaching `ImageFrame<T>` and `Image<T>` this much, the existing `ImageFrame[]` constructor is intended to support `image.Clone()` functionality AFAIK.

Is there anything preventing you to create the `Image<T>` with `Image.CreateUninitialized<T>()` in `SpectralToImageConverter<T>` instead of creating an `ImageFrame<T>`?

br3aker:
@antonfirsov I didn't remove already existing API for multi-frame images just added yet another *internal* contructor which already fixes several `new[] { singleFrame }`. While it's only a couple of bytes wasted it's still a waste which can be easily avoided without breaking semantics of the constructors.

You are right, `ImageFrame[]` is used for Cloning but is not limited by it. For example, Tiff decoder initializes image with ctor dependent on provided `ImageFrame` collection. Jpeg can't have more than one frame and it's 'natural' to create jpeg image from single frame, it's semantically correct for this format. And it's internal anyway, end user won't be able to break anything.

I prefer creating `ImageFrame<T>` instead of final `Image<T>` for 3 reasons:
1. `JpegDecoderCore` produces `Image<T>`, including pixel buffer itself and related metadata. Practically 'JpegDecoderCore' produces metadata and `ScanDecoder` + `SpectralConverter` produces pixel buffer. And only then everything is connected together to return `Image<T>`.
2. The only thing preventing `Image<T>` creation is `ImageMetadata` instance. While it can be easily passed to the converter ctor, it won't be used at all just passed to actual `Image<T>` ctor. And we don't even know if metadata is 'initialized', yes, it's a class and we only want the reference but it looks clunky to be honest. "Why would pixel converter use metadata on top of already injected decoding settings?"
3. As I said, semantics. This jpeg is single frame image and you even have `Start of Frame`.

Furthermore, as in converter we work with pixel buffer directly in hot path, it's usually better to save extra reference to pixel buffer itself which would look somewhat like this:

```csharp
class SpectralConverter<T>
{
    // exists here only to return after entire conversion chain
    public Image<T> Image { get; private set; }
    
    // actually used field
    private ImageFrame<T> frame;

    // ...
}
```

antonfirsov:
Sorry for disappearing, stuck in analysis paralysis with this a bit. The justification makes sense, but in long term we need to aim for decoder techniques, that can be refactored to work with public API-s. `ImageFrame<T>` has been designed to be created nd managed with `ImageFrameCollection<T>`, it's constructor will be never public, same true for the proposed new `Image<T>` constructor.

Maybe we should define public `Image<T>` constructors which consume, and take ownership of a `Buffer2D<T>` instead? You can fill the `Buffer2D<T>` in `SpectralToImageConverter<T>` then.

@JimBobSquarePants thoughts?

JimBobSquarePants:
I'll have to review how that works with our existing APIs. I'm loath to expose more public constructors unless we absolutely need them. All the wrapping ones we have already do make me uneasy.

br3aker:
Sorry for no response, was fixing cooling on my pc.

@antonfirsov `Buffer2D` constructor is even more suitable for converters, I like that!

@JimBobSquarePants maybe we can expose internal `Image<T>` constructor which accepts existing `ImageFrameCollection<T>`? That way constructors for single and many frames would go away.

br3aker:
As new Jpeg decoding pipeline have been merged, this is no longer needed, closing this for now.

PR #1694, opened 2021-07-12
JpegDecoder: post-process baseline spectral data per MCU-row 

### LINKED ISSUE #1597: JpegDecoder: post-process baseline spectral data per MCU-row

### LINKED ISSUE #1692: Jpeg colorspace is undefined after metadata only (Identify call) pass
### Prerequisites

### Description
Jpeg color space is undefined after `metadataOnly` decoding pass. This is happening because colorspace deduction (and metadata population) logic is locked behind `if(!metadataOnly)` check in `SOF` marker parsing (line 916):

https://github.com/SixLabors/ImageSharp/blob/fe65a387a953c8b2e41ccb5d6a9c915585e90d8f/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs#L868-L918

### Steps to Reproduce
```csharp
var info = Image.Identify("image_path");
Console.WriteLine(info.Metadata.GetFormatMetadata(JpegFormat.Instance).ColorType.HasValue); // prints false
```

```csharp
using var image = Image.Load("image_path");
Console.WriteLine(image.Metadata.GetFormatMetadata(JpegFormat.Instance).ColorType.HasValue); // prints true
```

br3aker:
### Prerequisites

### Description
Fixes #1597.

This also fixes #1692 simply because tests were failing with new architecture.

### Performance
No change. Baseline images should obviously be faster on the global scale due to lower memory footprint but in local benchmarks there's no change. And it's easy to explain: actual decoding code wasn't changed, only memory allocation and temporal buffers management.

_Note: PR benchmark is a bit faster than master branch (it's even faster than that because of a virtual call per each image stride). But I recall this false positive due to a really small image size. Most likely it's due to lesser memory footprint of spectral blocks allocation. I assume that at 4k resolutions there should be a little to no difference between them as actual code behind parsing/decoding wasn't touched._

|                             Method |            TestImage |     Mean |     Error |    StdDev | Ratio |
|----------------------------------- |--------------------- |---------:|----------:|----------:|------:|
| JpegDecoderCore.ParseStream master | Jpg/b(...)e.jpg [21] | 3.680 ms | 0.0152 ms | 0.0143 ms |  1.00 |
| JpegDecoderCore.ParseStream PR     | Jpg/b(...)e.jpg [21] | 3.474 ms | 0.0196 ms | 0.0174 ms |  0.94 |

Unfortunately, I don't have access to jetbrains memory profiler so I had a lot simplier check - custom debug allocator which simply prints all allocations. Here are dumps from the same image with same subsampling but different save methods - baseline and progressive:

*All small `byte` allocations are removed for clarity*

Progressive
```
Allocation of Rgba32[311400]
Allocation of Single[8448]
Allocation of Single[8448]
Allocation of Single[8448]
Allocation of Vector4[519]
Allocation of Block8x8[5016]
Allocation of Block8x8[1254]
Allocation of Block8x8[1254]
```

Baseline
```
Allocation of Rgba32[311400]
Allocation of Single[8448]
Allocation of Single[8448]
Allocation of Single[8448]
Allocation of Vector4[519]
Allocation of Block8x8[132] <- yay!
Allocation of Block8x8[33]  <- yay!
Allocation of Block8x8[33]  <- yay!
```

P.S. If somebody has some free time and access to jetbrains memory profiler - this is super welcome for the full picture!

### Tests & Legacy code
Everything is up and running. Old pipeline is removed.

### TODO
- [x] Fix existing tests
- [x] Tests for `SpectralConverter<TPixel>`
- [x] Remove legacy post processing code


JimBobSquarePants:
```
Allocation of Block8x8[132] <- yay!
Allocation of Block8x8[33]  <- yay!
Allocation of Block8x8[33]  <- yay!
```

@br3aker That is so nice to see! This should make a huge difference in our memory allocation benchmarks!

src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegFrame.cs
```diff
@@ -95,14 +95,23 @@ public void Dispose()
         /// </summary>
         public void InitComponents()
         {
-            this.McusPerLine = (int)MathF.Ceiling(this.SamplesPerLine / 8F / this.MaxHorizontalFactor);
-            this.McusPerColumn = (int)MathF.Ceiling(this.Scanlines / 8F / this.MaxVerticalFactor);
+            this.McusPerLine = (int)MathF.Ceiling(this.PixelWidth / 8F / this.MaxHorizontalFactor);
```
gfoidl, line 98: The values are all (positive) `int`s, so this can be written to avoid `(int)MathF.Ceiling` as
```c#
uint tmp = (uint)this.MaxHorizontalFactor * 8;
this.McusPerLine = (int)(((uint)this.PixelWidth + tmp - 1) / tmp);
```
[Sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKBuIGYACMxgYUZoG9GePrfmTYikYBJAHYBLDOQAUASj4DONAQIwALSbgB0AWTABXXAAUYUADKTxMRgF5Gs6xnl7smgGI7WMSQBtrAHNZTW0dE0kEGD8AdUkAE01GAHpGAA4PFMZQ3TcEAAloSQAvCHEMbD8PbDAMaHkAblVeAF9mnnbBZhEJaVIFTpV+NR5nbPwAB3tsrVzsAqLS8srq2uhGACp0puGRnP0jU3MrG2mQ2fDI6LjEjUYAanGpuEZyRVSMSZ2BNt2lXgY3TEUgw9AGfyGIx4hjGnymDlkMPK8n2eUKUBKZQqVRqdSgm22nXUFwMxjMlmstgRznksjpSJc+wiUViCSSjzhjBebyycManV+Ak6gLGzOubLu3ECMAwDUYuBlcsFAKYYzRiyxK1x6ylivleuVPBF5UYpKOFNOutl+utvxaQA==)

br3aker: @JimBobSquarePants can we get this as some `int DivideCeil(int number, int divisor)` method to `Numerics.cs`? Placing it somewhere in the `JpegDecoderCore` or even in a separate `JpegUtils` file seems odd.

br3aker:
**Update**
Existing spectral libjpeg comparison tests are now working ccrretly (actually found a bug in decoding code thanks to it!)

src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegFrame.cs
JimBobSquarePants: Sounds good to me!

br3aker:
@JimBobSquarePants found an error in jpeg decoder reference output files which was even marked as a bug:
```
// BUG: The following image has a high difference compared to the expected output: 1.0096%
TestImages.Jpeg.Baseline.Jpeg420Small,
```

I've re-saved actual jpeg to png using photoshop and got this difference from my file and existing png file:
image

With this fix image passes tolerance test:
```
*** Jpg/baseline/jpeg420small.jpg ***
Difference: 0,2863%
```

I'm not sure how I can update it, can you elaborate please?

JimBobSquarePants:
Awesome! It should be a case of replacing the reference image in this folder with the actual correct output from your tests as long as you have got Git LFS installed (see the readme for instructions)

https://github.com/SixLabors/ImageSharp/tree/master/tests/Images/External/ReferenceOutput/JpegDecoderTests

br3aker:
Thanks!

src/ImageSharp/Formats/Jpeg/Components/Decoder/SpectralConverter.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
+{
+    internal abstract class SpectralConverter
+    {
+        public abstract void InjectFrameData(JpegFrame frame, IRawJpegData jpegData);
```
JimBobSquarePants, line 8: Let's document these methods since they are inheritable.

tests/ImageSharp.Benchmarks/Codecs/Jpeg/DecodeJpegParseStreamOnly.cs
```diff
@@ -1,59 +1,59 @@
...
+//// Licensed under the Apache License, Version 2.0.
+
+//using System.IO;
+//using BenchmarkDotNet.Attributes;
+//using SixLabors.ImageSharp.Formats.Jpeg;
+//using SixLabors.ImageSharp.IO;
+//using SixLabors.ImageSharp.Tests;
+//using SDSize = System.Drawing.Size;
+
+//namespace SixLabors.ImageSharp.Benchmarks.Codecs.Jpeg
```
JimBobSquarePants, line 11: What's happening here?

tests/ImageSharp.Tests.ProfilingSandbox/Program.cs
```diff
@@ -32,14 +31,13 @@ private class ConsoleOutput : ITestOutputHelper
         /// </param>
         public static void Main(string[] args)
         {
-            LoadResizeSaveParallelMemoryStress.Run();
```
JimBobSquarePants, line 35: What's happening here? This is how we're currently comparing the memory usage between libraries.

JimBobSquarePants (commented):
This is looking really great so far!

tests/ImageSharp.Tests.ProfilingSandbox/Program.cs
br3aker: I was playing around in the sandbox while doing this PR, most likely it's a side effect, I will roll back to initial state.

tests/ImageSharp.Benchmarks/Codecs/Jpeg/DecodeJpegParseStreamOnly.cs
br3aker: Due to new architecture there was no stream parsing without metadata parsing. It's fixed now, need to uncomment this and alter a bit.

br3aker:
**Update**
- Added integer ceiled division (tests included)
- Fixed parse stream only benchmark and updated results for PR imlementation (see top comment for comparison)
- Fixed (last) invalidated test which saves encoded
- Added docs to the `SpectralConverter` class
- Restored Sandbox code

JimBobSquarePants (approved):
Really well done! This is an excellent piece of refactoring!

I can't believe you managed to solve that longstanding issue also! 😍

antonfirsov:
I did some WPA profiling on this on my 10 Core `i9-10900X`, and the results are quite impressive, although for some reason we didn't get the 8x memory footprint drop I expected. (Miscalculated the memory footprint of blocks in comparison to the image data?)

## 1. `LoadResizeSaveParallelMemoryStress` with `MaxDegreeOfParallelism = 8` and `Filter = JpegKind.Baseline`

#### Before the PR

Peak around 6.4 GB, total processing time `15.03` seconds.
image

#### Current master

Peak around 3.2 GB, total processing time `13.73` seconds.

image

## 2. `LoadResizeSaveParallelMemoryStress` with `MaxDegreeOfParallelism = 20` and `Filter = JpegKind.Any`

#### Before the PR

Peak around 13 GB.
image

#### Current master

Peak around 4.4 GB.
image

Great job @br3aker, thanks for the contribution!
/cc @JimBobSquarePants

JimBobSquarePants:
Those numbers 😘

Issue #845, opened 2019-03-09
Saving JPEG Quality defaults to below 5 - ignoring default of 75


Marcel0024:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Saving certain Jpeg images without defining the JpegEnconder quality defaults to a quality below 5.
There's a few jpeg I've seem with this behavior, can't pin point an common trait. Examples further below.

### Steps to Reproduce
```       
using (var img = Image.Load("original.jpg"))
{
    using (FileStream outImage = File.Create("result.jpg"))
    {
        img.SaveAsJpeg(outImage);
    }
}
```

Results: 

![image](https://user-images.githubusercontent.com/14852157/54068210-72a5ea80-424a-11e9-8897-92e2bdd9ae4e.png)

**With Quality at 50:**
```       
using (var img = Image.Load("original.jpg"))
{
    using (FileStream outImage = File.Create("result.jpg"))
    {
        img.SaveAsJpeg(outImage, new JpegEncoder { Quality = 50 } );
    }
}
```

Results:

![image](https://user-images.githubusercontent.com/14852157/54068226-c6183880-424a-11e9-8601-82e355bef9af.png)

As you can see quality 50 produces a higher quality image.
Here are 3 examples images:
[quality-loss-example.zip](https://github.com/SixLabors/ImageSharp/files/2948335/quality-loss-example.zip)

I've managed to track it down to [SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder.QualityEvaluator#EstimateQuality](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Formats/Jpeg/Components/Decoder/QualityEvaluator.cs#L79) method. 
'Normal' jpeg this method returns a 75 default quality. My example images the method returns 3 or 2.
![image](https://user-images.githubusercontent.com/14852157/54068534-062cea80-424e-11e9-93fe-fa00b2c727be.png)

### System Configuration

- ImageSharp version: 1.0.0-beta0006
- Environment (Operating system, version and so on): Windows 10 pro -Windows Server 2016
[...]

PR #1706, opened 2021-07-17
Jpeg quantization metadata

### LINKED ISSUE #1705: Use decoded quantization tables for Jpeg encoding, store them in JpegMetadata
### Prerequisites

### Description
https://user-images.githubusercontent.com/20967409/126037087-16a8bd51-8cc5-46c8-ab8a-e66d625a8595.jpg
This image gains a lot of noise after `decode-encode` scenario (you better download it and magnify with some app or use a histogram):

9483_comparison

### Steps to Reproduce
Image above with:
```csharp
using var image = Image.Load(loadPath);
JpegEncoder encoder = new JpegEncoder { };
image.SaveAsJpeg(savePath, encoder);
```

### Cause
Given photo uses some really strange quantization tables:

image

Non-standard tables are not a problem but current `QualityEvaluator` misses this one quiet drastically:
```
// DQT #1
Quality estimation:
        Standard: False
        Quality: 70

// DQT #2
Quality estimation:
        Standard: False
        Quality: 52

// image.Metadata.GetJpegMetadata().Quality
Loaded image quality: 52

// JPEGsnoop output:
Approx quality factor = 71.19 (scaling=57.62 variance=593.35) - Luminance
Approx quality factor = 80.24 (scaling=39.51 variance=961.47) - Chrominance
```

So now we have 2 problems:
1. Huge miss on quality
2. Deducting quality twice as current code does not differ 2+ DQT markers

### Proposal
Save decoded jpeg quantization tables to the jpeg metadata and use it as backup in the encoder:

```csharp
// somewhere in the jpeg encoder

// always prefer explicit quality
if(options.Quality != null) 
{
    return GetStandardQuantizationTablesFromQuality(options.Quality);
}
if(metadata.UsesStandardQuantizationTables)
{
    return GetStandardQuantizationTablesFromQuality(metadata.Quality);
}
return metadata.QuantizationTables;
```

br3aker:
### Prerequisites

### Description
fixes #1705 
Color deduction is improved, tested on some images - results are promising!
Moved everything quantization table-wise to the `Quantization.cs`

Image at #845 had invalid test 'quality', it should have been rounded, not ceiled, set it from 99 to 98 for now.

### TODO
- [x] encoder should select quantization tables based on new API
- [x] docs
- [x] tests
~~`nint` usage~~
- [x] tests to calculate `is this variance standard enough?' in practice because we actually need to determine if table is standard


src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
```diff
@@ -830,5 +830,46 @@ public void TransposeInto(ref Block8x8F d)
                 d.V7R.W = this.V7R.W;
             }
         }
+
+        /// <summary>
+        /// Compares entire 8x8 block to a single scalar value.
+        /// </summary>
+        /// <param name="value">Value to compare to.</param>
+        public bool EqualsToScalar(int value)
```
br3aker, line 838: I don't like this method but it's a must. Current pipeline converts encoded tables directly to `Block8x8F` and we should check if table contains all 1's...

src/ImageSharp/Formats/Jpeg/Components/Quantization.cs
```diff
@@ -0,0 +1,131 @@
...
+    internal static class Quantization
+    {
+        /// <summary>
+        /// Threshold at which given luminance quantization table should not be considered 'standard'.
+        /// </summary>
+        /// <remarks>
+        /// Jpeg does not define either 'quality' nor 'standard quantization table' properties
+        /// so this is purely a practical value derived from tests.
+        /// </remarks>
+        public const double StandardLuminanceTableVarianceThreshold = 10.0;
```
br3aker, line 22: This a WIP value, I will write tests which output variances with quality 1-100 for standard tables so we can use better value as threshold. Same for chroma threshold.

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
```diff
@@ -23,12 +46,26 @@ private JpegMetadata(JpegMetadata other)
...
+            this.ChromaQuality = other.ChromaQuality;
         }
 
         /// <summary>
         /// Gets or sets the encoded quality.
         /// </summary>
-        public int Quality { get; set; } = 75;
+        public int Quality
+        {
+            get => (int)Math.Round(this.LumaQuality);
```
br3aker, line 61: Current implementation assumes quality is based on luminance channel only, while this would work just fine as luma is more important than chroma, maybe we should calculate this some other way?

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
```diff
@@ -1,13 +1,36 @@
...
+
+        /// <summary>
+        /// Luminance qunatization table derived from jpeg image.
+        /// </summary>
+        /// <remarks>
+        /// Would be null if jpeg was encoded using table from ITU spec
+        /// </remarks>
+        internal Block8x8? chromaQuantizationTable;
+
+        internal double LumaQuality;
```
br3aker, line 30: According to JPEGsnoop implementation, we have 2 quality multipliers - luma & chroma. Two questions:
1. should we expose both as public properties?
2. do we even need quality for separate channels so we can encode using different multipliers?

tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.Metadata.cs
```diff
@@ -55,7 +55,7 @@ public partial class JpegDecoderTests
         {
             { TestImages.Jpeg.Baseline.Calliphora, 80 },
             { TestImages.Jpeg.Progressive.Fb, 75 },
-            { TestImages.Jpeg.Issues.IncorrectQuality845, 99 }
+            { TestImages.Jpeg.Issues.IncorrectQuality845, 98 }
```
br3aker, line 58: JPEGsnoop outputs `98.25`, not sure why it was ceiled.

br3aker (commented):
Have some questions about internal API

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
JimBobSquarePants: I think we should expose both.

The original Quality parameter could be read as an average of both when we have tables from the decoder and written to both when encoding against the standard tables. We can use the specific properties to apply more granular quality. 

I think we could also obsolete Quality

br3aker: Does current stylecop setup forces to obsolete usage to be an compilation error? Marking anything with `Obsolete(msg, error: false)` breaks local compilation on release setup.

JimBobSquarePants: Yeah must do. We'll have to ensure that we refactor tests to use the obsolete method as little as possible and suppress the warning in the source where we need to test it.

br3aker:
@JimBobSquarePants @antonfirsov 

Did some research on estimator and it works very well actually predicting matrices produced by the encoder:
<details>
  <summary>Luminance</summary>

```
Variances for Luminance table.
Quality levels 25-100:
q=25	0	est. q: 25
q=26	0,6900486702215858	est. q: 26,024412840373362
q=27	1,4706823623128003	est. q: 27,024602491459646
q=28	1,2785469780174026	est. q: 28,112191569091046
q=29	1,1775001233145304	est. q: 29,040546617422844
q=30	1,4174248602284933	est. q: 30,07976440516557
q=31	1,4646344796710764	est. q: 31,02076929396441
q=32	1,2118263731099432	est. q: 32,02205643673543
q=33	0,8141364518887713	est. q: 33,11632561401876
q=34	1,6665811662714987	est. q: 33,96628505199738
q=35	0,9232605200741091	est. q: 35,177795215278366
q=36	1,1127103053840983	est. q: 36,27815962938035
q=37	1,378856307335809	est. q: 37,002086926012055
q=38	1,0856774119383772	est. q: 38,22325158100902
q=39	1,1775001233163493	est. q: 39,115387591973736
q=40	1,4400911343473126	est. q: 39,86247084081212
q=41	1,3767202763010573	est. q: 41,29910411058831
q=42	1,234173531009219	est. q: 41,97338990890838
q=43	1,1282963146550173	est. q: 43,00023311570005
q=44	1,3585401036325493	est. q: 44,26381812775582
q=45	1,458902856309578	est. q: 45,04201720830167
q=46	0,6900486702379567	est. q: 46,350927451413654
q=47	1,088109967751734	est. q: 47,121053433942755
q=48	2,2080337819988927	est. q: 48,040862469704074
q=49	1,0528940597578185	est. q: 49,28736107087304
q=50	0	est. q: 50
q=51	1,0528931625449331	est. q: 50,722942454740405
q=52	2,2080339720541815	est. q: 52,03903312794864
q=53	1,0881065990797651	est. q: 53,05484114214778
q=54	0,6900482721739536	est. q: 53,93635290674865
q=55	1,220181547147149	est. q: 54,903984162956476
q=56	1,3714545976245063	est. q: 55,983932269737124
q=57	1,365342979261186	est. q: 57,04243411310017
q=58	1,1282981840340653	est. q: 58,139219880104065
q=59	1,1410138193514285	est. q: 59,037849912419915
q=60	1,1407521721157536	est. q: 59,940363466739655
q=61	1,2785477334891766	est. q: 61,07061146758497
q=62	1,2103239791313172	est. q: 61,95206558331847
q=63	1,2172928992104062	est. q: 63,06559508666396
q=64	1,1775002723243233	est. q: 63,91346617601812
q=65	1,0665629485201862	est. q: 65,00663976185024
q=66	1,3845724134389457	est. q: 65,94032463617623
q=67	1,4174274697306828	est. q: 66,8876466806978
q=68	1,1324866958821076	est. q: 68,01257678307593
q=69	1,1127106472777086	est. q: 68,91198516823351
q=70	1,4293262576607049	est. q: 70,06213702261448
q=71	0,9232593122746948	est. q: 71,06755869463086
q=72	1,2118242887463566	est. q: 71,92881139926612
q=73	1,7028581219874468	est. q: 72,91174442507327
q=74	1,2061910146539958	est. q: 74,16409873403609
q=75	0,8058072853714293	est. q: 74,74685530178249
q=76	1,2061902484078928	est. q: 75,83590135909617
q=77	1,7028572779545357	est. q: 77,08825510926545
q=78	1,2118248299298102	est. q: 78,0711883213371
q=79	0,923259961500662	est. q: 78,93244128208607
q=80	1,4293257435849682	est. q: 79,93786276783794
q=81	1,112709965312888	est. q: 81,08801490161568
q=82	1,1324864905966479	est. q: 81,98742298409343
q=83	1,4174264429316281	est. q: 83,11235310975462
q=84	1,3845713790170748	est. q: 84,05967517755926
q=85	1,049529159115309	est. q: 84,93440570309758
q=86	1,1775005830131704	est. q: 86,08653354458511
q=87	1,2172932673361174	est. q: 86,93440470378846
q=88	1,2103243191085085	est. q: 88,04793453309685
q=89	1,2785471962706083	est. q: 88,92938854405656
q=90	1,1407528967250187	est. q: 90,05963645176962
q=91	1,1410135141355227	est. q: 90,96215004101396
q=92	1,1282975903178567	est. q: 91,86078009661287
q=93	1,365343046062577	est. q: 92,95756578212604
q=94	1,3714546697582364	est. q: 94,0160675381776
q=95	1,2515629371297763	est. q: 95,03706148243509
q=96	0,6900487907659567	est. q: 96,06364700011909
q=97	1,0881073654912399	est. q: 96,94515896844678
q=98	2,3629059983706604	est. q: 97,68161100219004
q=99	5,001353712778048	est. q: 98,32455397117883
q=100	0	est. q: 100
Min variance: 0
Max variance: 5,001353712778048
```
</details>
<details>
  <summary>Chrominance</summary>


```
Variances for Chrominance table.
Quality levels 25-100:
q=25	0	est. q: 25
q=26	0,3289928270824021	est. q: 26,03743994664574
q=27	0,2567842723437934	est. q: 27,064679348888934
q=28	0,19492864389758324	est. q: 28,125373916432874
q=29	0,15367341504315846	est. q: 29,10831780075463
q=30	0,12293467960626003	est. q: 30,16416949602972
q=31	0,24340592175576603	est. q: 31,106414456925577
q=32	0,4086034253741673	est. q: 32,12127202644116
q=33	0,23096369710401632	est. q: 33,20182788235059
q=34	0,40788787931160186	est. q: 33,960405383481834
q=35	0,1754530939288088	est. q: 35,09803580540892
q=36	0,19451782029500464	est. q: 36,156069992036116
q=37	0,3453769973530143	est. q: 37,003982164021295
q=38	0,456310679546732	est. q: 38,09049339964678
q=39	0,15367341504315846	est. q: 38,99310668528405
q=40	0,36991917603336333	est. q: 39,88931618100753
q=41	0,3158340733843943	est. q: 41,2617105814941
q=42	0,33136279402606306	est. q: 41,96017561784348
q=43	0,15366260785413033	est. q: 43,02776360067011
q=44	0,2676441431249259	est. q: 44,25175909324916
q=45	0,1294194163674547	est. q: 44,99780749316061
q=46	0,3289928270805831	est. q: 46,30966084423982
q=47	0,23879525277334324	est. q: 47,1641921199368
q=48	0,13732431547396118	est. q: 48,01189636312582
q=49	0,47396762017160654	est. q: 49,08130488693843
q=50	0	est. q: 50
q=51	0,4739676361987222	est. q: 50,93589103780687
q=52	0,13732498997160292	est. q: 52,07042586989701
q=53	0,23879479229617573	est. q: 53,00631681457162
q=54	0,3289942392184457	est. q: 53,984418185427785
q=55	0,2589643885685291	est. q: 55,09069343097508
q=56	0,13713482134426158	est. q: 56,068794848397374
q=57	0,4446088114455051	est. q: 57,091491390019655
q=58	0,1536636163955336	est. q: 58,10201824642718
q=59	0,2281612571041478	est. q: 59,041866986081004
q=60	0,3358503991194084	est. q: 60,069571482017636
q=61	0,1949285494411015	est. q: 61,11229262314737
q=62	0,21875850971719046	est. q: 62,05319166183472
q=63	0,3679954791741693	est. q: 63,08983927592635
q=64	0,15367232371136197	est. q: 64,11389419808984
q=65	0,28967497732719494	est. q: 65,07345945574343
q=66	0,2991529692144468	est. q: 66,13836633041501
q=67	0,12293578855769738	est. q: 67,12021087296307
q=68	0,4809884522892389	est. q: 68,12198660336435
q=69	0,19451708518090527	est. q: 69,14468077011406
q=70	0,36365722516802634	est. q: 70,13153568841517
q=71	0,175452534112992	est. q: 71,22905980795622
q=72	0,40860331005569606	est. q: 72,16995642520487
q=73	0,1974451000164663	est. q: 73,18900562822819
q=74	0,364259353626494	est. q: 74,24038457684219
q=75	0,1931513081635785	est. q: 74,74056840874255
q=76	0,3642587075200936	est. q: 75,75961533002555
q=77	0,1974445989171727	est. q: 76,81099416222423
q=78	0,4086036432572655	est. q: 77,83004227094352
q=79	0,17545300647134354	est. q: 78,77094012219459
q=80	0,3636563494492293	est. q: 79,86846421845257
q=81	0,19451726894635613	est. q: 80,8553179493174
q=82	0,4809880078362312	est. q: 81,87801335006952
q=83	0,12293536124298043	est. q: 82,87978901062161
q=84	0,2991527726039749	est. q: 83,86163474060595
q=85	0,28967514720920917	est. q: 84,9265405209735
q=86	0,1536722510470554	est. q: 85,8861057786271
q=87	0,3679949720606146	est. q: 86,910161934793
q=88	0,21875810842482224	est. q: 87,9468082333915
q=89	0,1949284416592718	est. q: 88,8877066783607
q=90	0,33585009744371064	est. q: 89,93042961228639
q=91	0,22816131872224332	est. q: 90,958133013919
q=92	0,15366335787598473	est. q: 91,89798114821315
q=93	0,4446088580157834	est. q: 92,90850972174667
q=94	0,1371350071192694	est. q: 93,93120513414033
q=95	0,258963805380489	est. q: 94,90930595784448
q=96	0,3289943299256919	est. q: 96,01558267313521
q=97	0,2387951076492314	est. q: 96,99368320580106
q=98	0,13732446557704137	est. q: 97,92957349563949
q=99	0,8949631033036098	est. q: 98,82904179685283
q=100	0	est. q: 100
Min variance: 0
Max variance: 0,8949631033036098
```

</details>

Those are implemented as disabled tests which outputs exactly these results so it can be checked in the future if for whatever reason estimator would be changed or when transition to 16-bit ints is done.

**Summary**
Max variances:
Luma - `5,001353712778048`
Chroma - `0,8949631033036098`

While Luma has max variance of `~5` I would argue that we should use it as threshold. It's way too large compared to other results where max value is `~2.4` and actual average is `~1.4`. Especially considering 5 and 2.3 came from 99 and 98 quality settings respectively.

So my proposal for variance thresholds is:

```
Luma: 2.37
Chroma 0.9
```


**Why quality < 25 isn't present in the tests?**

Tests show that `quality=25` is the lower confidence threshold, everything below it results is a >+100 variance for both luma and chroma tables. This quality setting range should be considered highly compressed with a huge loss of details so we should threat them very carefully at re-encoding so proposed variance thresholds above would force them to use original tables from decoded file (only if user didn't manually set quality setting in the encoder/metadata).

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
```diff
@@ -830,5 +830,46 @@ public void TransposeInto(ref Block8x8F d)
...
+            if (Avx2.IsSupported)
+            {
+                const int equalityMask = unchecked((int)0b1111_1111_1111_1111_1111_1111_1111_1111);
+
+                var targetVector = Vector256.Create(value);
+                ref Vector256<float> blockStride = ref this.V0;
+
+                for (int i = 0; i < RowCount; i++)
+                {
+                    Vector256<int> areEqual = Avx2.CompareEqual(Avx.ConvertToVector256Int32WithTruncation(Unsafe.Add(ref this.V0, i)), targetVector);
```
gfoidl, line 850:
```suggestion
                    Vector256<int> areEqual = Avx2.CompareEqual(Avx.ConvertToVector256Int32WithTruncation(Unsafe.Add(ref this.V0, (nint)(uint)i)), targetVector);
```
Doesn't look that nice, but avoids the `movsxd` instruction (cf. [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYmpBkIYBvGgzNNdwAJ4YYDALLkAFLABmDa7YYATbBmxoDLw8Qd4IAJQMALwAfAwAqly42K6CAILe3i4w7r7+gbxh4RrU5hYeNnb2pNnunnZ5AUEhhRHRcYnJqSwZWW4+fk1OXMEY4U4co+GtxTQAvkA)).

Alternatively type the loop variable as `nint` and cast `(nint)(uint)RowCount`.

Same on other places where Unsafe.Add with an int-offset is used.

gfoidl (commented):
A micro-optimization hint.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
JimBobSquarePants: I can’t remember what language version we’re targeting. We may have to upgrade in the shared infrastructure props

br3aker: Not 9.0 for afaik, I remember trying to use `nint` and failed to do so.

JimBobSquarePants: I can upgrade but I'm not convinced we should bother with the optimization here or elsewhere. It's damn ugly and I'm sure also fixed in new target frameworks.

br3aker: We can use `nint` directly for indices so we won't need to cast `(nint)(uint)`.

JimBobSquarePants: LangVersion updated upstream!

`git submodule foreach git pull origin master`

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
```diff
@@ -23,12 +26,46 @@ private JpegMetadata(JpegMetadata other)
...
+        public int Quality
+        {
+            get => (int)Math.Round(this.LumaQuality);
+            set => this.LumaQuality = value;
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether jpeg was encoded using ITU section spec K.1 quantization tables
+        /// </summary>
+        public bool ItuSpecQuantization => !this.LumaQuantizationTable.HasValue && !this.ChromaQuantizationTable.HasValue;
```
antonfirsov: Naming suggestion:

```suggestion
        public bool UsesItuSpecQuantization => !this.LumaQuantizationTable.HasValue && !this.ChromaQuantizationTable.HasValue;
```

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
```diff
@@ -23,12 +26,46 @@ private JpegMetadata(JpegMetadata other)
...
+        internal double ChromaQuality { get; set; }
+
         /// <summary>
         /// Gets or sets the encoded quality.
         /// </summary>
-        public int Quality { get; set; } = 75;
+        public int Quality
+        {
+            get => (int)Math.Round(this.LumaQuality);
+            set => this.LumaQuality = value;
```
antonfirsov: How does luma vs chroma quality correalate in practical cases? Wouldn't it be better to drop both non-standard tables if user touches `Quality` to make the behavior symmetric? (see also my other comment)

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -753,81 +753,107 @@ private void ProcessApp14Marker(BufferedReadStream stream, int remaining)
...
+                        break;
+                    }
 
-            this.Metadata.GetFormatMetadata(JpegFormat.Instance).Quality = QualityEvaluator.EstimateQuality(this.QuantizationTables);
+                    // chrominance table
+                    case 1:
+                    {
+                        Quantization.EstimateQuality(ref table, Quantization.UnscaledQuant_Chrominance, out double quality, out double variance);
+                        jpegMetadata.ChromaQuality = quality;
+                        if (variance <= Quantization.StandardChrominanceTableVarianceThreshold)
```
antonfirsov: Is it common that luma table is non-standard while chroma table is standard, or vice-versa?

antonfirsov (commented):
To determine the variance thresholds, it seems worth to check how far do the variance values go with non-standard tables (I guess they are not slightly altered standard ones?). Maybe it's worth to keep the thresholds higher?

antonfirsov:
Looks like I was commenting on outdated code. Exposing `LuminanceQuality` + `ChromaQuality` + `Quality` is too much for a regular user IMO, can't we do some trick to avoid that?

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
br3aker: I decided to drop it in favour of internal check if tables are carried from decoding routine, this is a very specific check which I believe 99.99% users won't bother with.

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
br3aker: Recent commits made quality as average between luma and chroma but I don't think it's a very good idea (will post a more detailed answer below).

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
br3aker: It can be within some variance threshold but not that I'm aware of.

br3aker:
> can't we do some trick to avoid that?

Problem with separate qualities within jpeg metadata is that it shouldn't be 'averaged'. For example let's image from related issue:
```
Luma quality: 72
Chroma quality: 80
Average: 76
```

Simply re-encoding this image would result in more binary size of luma component as important details are already lost due to quality 72 and we are adding some noise from DCT transform rounding errors with higher quality. On the other hand, chroma would lost quite some details due to lower quality (remember, we are simply re-encoding it after changing exif, for example, so it's fine to assume quality shouldn't alter that much).

I am all for exposing only `Quality` property for `JpegEncoder` so regular users can simply write `Quality = 80` and expect everything to work. While `JpegMetadata` would expose `LumaQuality` and `ChromaQuality` for specialized usage. I'm not sure what to with `Quality` in metadata. Averaging is not always correct but removing it at all doesn't feel right...

br3aker:
@antonfirsov hi!

Did a final (to my taste) public API improvements:

`JpegMetadata` now provides 3 quality-related properties:
```
public int? LuminanceQuality; // this can be null only in image was created via ImageSharp, not decoded
public int? ChrominanceQuality; // null if grayscale
public int Quality; // average for luma/chroma qualities
```

`JpegEncoder` provided only general `Quality` property which sets both luma and chroma qualities to the same value. Which means there's no obsolete/removed API - only 2 new quality settings for metadata.

Internal implementation is still far from being final and I will work on it. Question right now is: is public API good enough?

antonfirsov:
@br3aker keeping `LuminanceQuality` and `ChrominanceQuality` `JpegMetadata`-only is much better, however I wonder if there is any end user use-case when a user is interested querying or altering them separately. If not, maybe we can turn them internal just like the quantization tables?

br3aker:
@antonfirsov I would like to say there is but I can't. There were no requests of such functionality. We could mimic JPEGsnoop functionality but JPEGsnoop already exists so yeah, it's better to hide everything and expose general `Quality` property for both metadata and encoder.

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
```diff
@@ -21,18 +27,94 @@ public JpegMetadata()
...
+        /// Note that jpeg image can have different quality for luminance and chrominance components.
+        /// This property return average for both qualities and sets both qualities to the given value.
+        /// </remarks>
+        public int Quality
+        {
+            get
+            {
+                int lumaQuality = this.LuminanceQuality ?? Quantization.DefaultQualityFactor;
+                int chromaQuality = this.ChrominanceQuality ?? lumaQuality;
+                return (int)Math.Round((lumaQuality + chromaQuality) / 2f);
```
antonfirsov, line 105: Does JPEGsnoop also report the average of the two values in `Approx quality factor`?

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
```diff
@@ -21,18 +27,94 @@ public JpegMetadata()
...
+        public int? ChrominanceQuality { get; set; }
 
         /// <summary>
         /// Gets or sets the encoded quality.
         /// </summary>
+        /// <remarks>
+        /// Note that jpeg image can have different quality for luminance and chrominance components.
+        /// This property return average for both qualities and sets both qualities to the given value.
+        /// </remarks>
+        public int Quality
```
antonfirsov, line 99: Would be a breaking change, but wouldn't it be better to define this as `int?` to better communicate the `LuminanceQuality  == null && LuminanceQuality == null` case?

br3aker: Oops, forgot about it being `int?` in current state. Will work on it tomorrow with no public luma/chroma in mind.

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
br3aker: JPEGsnoop has no 'quality' parameter, only separate luma/chroma. That's the main pain in the neck to define `Quality` alongside luma/chroma separation.

br3aker: It shouldn't be that bad for general images which have same quality levels for both tables. Jpegs with different quality levels for luma/chroma was a surprise to me and shouldn't be really popular.

antonfirsov: Are there any other libraries/tools which report a quality estimation after import?

br3aker: [Forensically](https://29a.ch/photo-forensics/#forensic-magnifier)
General quality
On different `luma(71.19)`/`chroma(80.24)` quality it returns `quality=81` - `ceil(max(luma, chroma))`?.

JPEGsnoop
Separate luma/chroma qualities
No general quality

ImageMagick
General quality

That's all I know, photoshop or any other tools for image manipulations do not expose such info afaik.
I guess not changing anything in public interface is the solution. We can always expose it later as internally metadata would still store separate qualities.

The only question is:
1. `quality => avg(luma, chroma)`
2. `quality => ceil(max(luma, chroma))`

This public property won't affect encoding process anyway but for user IMO it's better to expose max quality factor than average for reasons I stated in the comments above (loss of detail + extra noise and filesize).

antonfirsov: Great research, thanks! I also think that taking the maximum would be better.

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
antonfirsov: No it's `public int Quality` on the current master. I think we should consider changing it to `int?`, I doubt there are many users we would break.

JimBobSquarePants: Yep. Break it. We cannot afford to hold onto properties that cause complexity. We simply don't have the maintainer numbers to manage continuously backwards compatible codebase. 

I also vote max or we never really fix anything due to loss of detail/extra noise.

br3aker: Got ya, thanks for your thoughts!

br3aker: While we are here breaking API: isn't it better to remove setter for metadata quality? There's no usage for it anywhere, jpegs are encoded using quality from `JpegEncoder`.

JimBobSquarePants: Quality should be read from Meta and overwritten using the encoder however metadata is always writable for all formats. That means you should be able to set quality without spinning up a new encoder.

br3aker: Jpeg uses two qualities (luma/chroma) separately. `Quality` from metadata is just a public accessor for end user.

br3aker: I actually missed point for encoder, it's quality can be changed before each image encoding routine. Problem with metadata quality is that encoder first checks if its quality has value and only then takes quality from metadata (if it doesn't exists - take default value). So in a situation like this:

```csharp
var img = Image.Load(path);
img.Metadata.GetJpegMetadata().Quality = 100;

var encoder = new JpegEncoder { Quality = 80 };
img.SaveAsJpeg(path, encoder); // this would be saved quality=80 as encoder settings have higher priority
```

Metadata quality won't matter. It can affect encoding like this:

```csharp
var img = Image.Load(path);
img.Metadata.GetJpegMetadata().Quality = 100;

var encoder = new JpegEncoder {}; // we can even use as shared resource over multiple threads (when color stuff is fixed)
img.SaveAsJpeg(path, encoder); // this would be saved quality=100
```

Not sure if it was intended.

JimBobSquarePants: Those examples you are providing are by design.

Upon review `Quality` in `JpegMetadata` should definitely not be nullable. It has a default value of `75` for when we create an image from scratch or change format. You can override that during decoding by taking the maximum of your calculated values.

The `Quality` property in the `JpegEncoder` is nullable and that's by design. This means we can always use the `JpegMetadata` property as a fallback but allow overriding it by choosing to use a new `JpegEncoder`. Choosing an explicit  `JpegEncoder` is a concious decision that falls outside of the normal workflow. Normally people will normally use the JpegEncoder in the `Configuration` instance.

br3aker: Can't agree with non-nullable quality in metadata. What's the usage of it for the user? `Quality` property is not used in the encoder with this PR. At the decoding stage separate luma/chroma qualities are stored (or their unique tables if those are not standard). At the encoding stage encoder tries to take quality from following sources in this exact order:

1. Encoder.Quality if is not null
2. Metadata.LuminanceQuality if is not null
3. Metadata.LuminanceQuantizationTable if is not null
4. DefaultQuality (75) <- this is a constant, it's not gathered from metadata

Same for Chrominance. We must save separate qualities because there are images with different qualities. `JpegMetadata.Quality` is simply a metadata for end user, it's not really used for encoding (only if user set `Quality` to some value which internally sets luma and chroma qualities to a given value). There's no uniform quality value anymore.

JimBobSquarePants: >1. Encoder.Quality if is not null
>2. Metadata.LuminanceQuantizationTable if is not null
>3. Metadata.LuminanceQuality if is not null
>4. DefaultQuality (75) <- this is a constant in encoder code, it's not gathered from metadata


If you're doing it in that order what's the point of having the Quality property in the metadata at all?

I'm not convinced we should preserve the custom tables. Any manipulation of the incoming image will render any optimization performed that customization pointless.

br3aker: > If you're doing it in that order what's the point of having the Quality property in the metadata at all?

So we don't need to carry standard tables in the metadata, we can safely calculate them in the encoding routine using quality from decoder. And for actual metadata purposes if user wants it.

> I'm not convinced we should preserve the custom tables. Any manipulation of the incoming image will render any optimization performed that customization pointless.

Preserving custom tables is the least intrusive way of re-encoding the image. Does it affect encoded image? Yes. Does it affect it that much? No. Tested for 2x downscaling image. There is a difference, it ain't much but it does exist.

Any mid-priced camera tends to use custom tables for at least chrominance components so I don't see why we shouldn't preserve custom tables especially considering it is already done and is working correctly with existing tests. It doesnt really change anything internally except one method in the `JpegEncoder` and it doesn't affect performance. The only question left is `int Quality` or `int? Quality`.

Final word is yours so just yes/no on:
1. Custom tables
2. `int` or `int?`

And I will push final changes.

br3aker:
**Update**
- Removed quantization table carry-over in the metadata
- Added tests where applicable
- Used `nint`

I guess it's ready for final review.

br3aker:
@JimBobSquarePants as I suspected, langver 9.0 broke something (most likely in Magick.Net), I'm just too newbie to determine what. At least we know for sure that we need to rollback to 8.0 or whatever version was used.

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
antonfirsov: > 1. Custom tables

We need to optimize to our most popular use case, which is thumbnail making. Assuming the user operation is a significant downscale, is the re-encoded thumbnail Jpeg "more correct" with the original tables or the default ones? I can't decide, such a downscale changes the frequency characteristics of the image a lot.

Sidenote: If I was to develop a thumbnail app, I would always set the Quality of the thumbnail to a higher value, even if it was low originally. Does ImageSharp.Web do anything like this @JimBobSquarePants?

> 2. `int` or `int?`

To me `metadata.Quality == null` would represent the case that the decoder was unable to estimate the quality of the input image nicely, while returning `metadata.Quality == 75` is quite an arbitrary choice. I vote for the breaking change (just let's make sure to apply the label).

br3aker: > We need to optimize to our most popular use case, which is thumbnail making. Assuming the user operation is a significant downscale, is the re-encoded thumbnail Jpeg "more correct" with the original tables or the default ones?

Barely noticable for 2x downsampling. So we have following situation:
Default tables:
**good** re-encoding with minimal change
**good** encoding of a heavily changed image

Original tables:
**even better** re-encoding with mininal change
**good** encoding of a heavily changed image

> To me `metadata.Quality == null` would represent the case that the decoder was unable to estimate the quality of the input image nicely, while returning `metadata.Quality == 75` is quite an arbitrary choice. I vote for the breaking change (just let's make sure to apply the label).

Not sure if we are talking about the same thing. `metadata.Quality` **is not used** during encoding process. Internal `metadata.LuminanceQuality` & `metadata.ChrominanceQuality` fields are used. Using `quality = max(luma, chroma)` would kill the quality if we are not talking about something extreme like thumbnails. Public quality is strictly a user way of knowing what was the decoded image quality. `metadata.Quality` setter sets luma & chroma to the same value. In other words, internal implementation won't really change and we either do (simplified):
`int quality = encoder.Quality ?? metadata.Quality ?? 75;` or `int quality = encoder.Quality ?? metadata.Quality;`

JimBobSquarePants:
@br3aker Sorry about all this flip-flopping on this PR. I'll do a proper sit-down review ASAP for you!

That `nint` thing is odd. I'll see what I can figure out!

br3aker:
@JimBobSquarePants having my code reviewed by .net professionals is worth the wait. I understand that you have a lot of work aside this project so no worries, take your time!

About that `nint` thingy - [only x86 net472 failed, x64 and other runtimes/platforms are all good](https://github.com/SixLabors/ImageSharp/pull/1706/checks?check_run_id=3177171815), I couldn't really find any relevant info about 9.0 langver being supported by it so maybe it's not. As we have a lot places with `Unsafe.Add` which all can be 'optimized' with `nint` usage I suggest to just forget about native int in this PR and open separate issue/PR for nint stuff and experiment with it there.

JimBobSquarePants:
> .net professionals

Haha. No pressure!

I'll get on this tomorrow night.

JimBobSquarePants (approved):
@br3aker If you're happy with this then I'm happy. Thanks again for your help and patience!

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
JimBobSquarePants: @br3aker Ah sorry, my bad! I could have sworn I'd seen a change but I was wrong!

Thanks for doing the extra research. I agree with you on quality < 25. I really cannot see anyone doing that for anything other than experimental purpose.

Issue #1704, opened 2021-07-17
Saving a PNG as Jpeg only processes a part of the image on .NET 6


kerams:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Loading a PNG file and immediately saving it as Jpeg works as expected on .NET 5. However when I switch to `net6.0` in the csproj, only the top part of the image appears in the output jpeg file, the rest is black. The dimensions are preserved. Feel free to close this if you believe it's a bug in .NET and not anything to do with ImageSharp.

### Steps to Reproduce
```csharp
using var img = Image.Load("unknown.png");
var fs = File.OpenWrite("unknown.jpeg");
img.SaveAsJpeg(fs);
fs.Flush();
fs.Close();
```

Use this image
https://user-images.githubusercontent.com/5063478/126033585-4efb31ce-c231-4229-8412-68fb73950ffc.png

The result ont .NET 6 Preview 6 looks like this
https://user-images.githubusercontent.com/5063478/126033623-0f886133-bc0d-4449-8048-d5d29a0e377c.jpeg

### System Configuration
- ImageSharp version: 1.0.3
- Environment (Operating system, version and so on): Win 10
- .NET Framework version:
  - 5.0.400-preview.21277.10
  - 6.0.100-preview.6.21355.2

JimBobSquarePants:
You should raise this issue in the .NET runtime.

kerams:
I could do that, but I'm not sure how they're going to investigate what from their point of view would appear to be a corrupted image generation in a third party library.

JimBobSquarePants:
It’ll be treated as a regression since it works in all previous frameworks. Our code hasn’t changed. 

We had similar issues with the .NET 5 previews where the runtime was emitting incorrect values for some intrinsics instructions.

br3aker:
Image is corrupted in a very funny way:
1. Jpeg header
2. Pixels (visible rows at the top)
3. A lot of black pixels (visible bottom rows)
4. EOI terminator (as expected)
5. **Actual color data which should have been written in place of black pixels** ending with second EOI marker

[...]

PR #1707, opened 2021-07-19
Fix a few uses of DeflateStream.Read

stephentoub:
### Prerequisites

(68 of the 71 _existing_ PngDecoder tests already fail when run against `master`, before and after this change, and I'm not sure if that's expected... presumably not?  Maybe I'm just doing something wrong?  I'm running them in test explorer in VS2022.)

### Description

The code is assuming Stream.Read always returns the requested number of bytes unless it hits EOF, but that's not guaranteed; the Stream.Read contract only guarantees at least 1 byte will be returned until EOF even if more are requested.

Fixes https://github.com/SixLabors/ImageSharp/issues/1704
https://github.com/dotnet/runtime/issues/55866

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -506,11 +506,15 @@ namespace SixLabors.ImageSharp.Formats.Png
             while (this.currentRow < this.header.Height)
             {
                 Span<byte> scanlineSpan = this.scanline.GetSpan();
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Ah fantastic! Thanks @stephentoub 

I'm not sure why you were getting test failures locally, (_We have a few quirks in our setup (git submodules, LFS_), were any specific errors being reported? 

CI seems to be working well anyway, as does the test run on my machine.

stephentoub:
> We have a few quirks in our setup (git submodules, LFS

Ah, I bet it's LFS. I'll try to fix it tomorrow.  Thanks.

JimBobSquarePants (approved):

PR #1709, opened 2021-07-20
Update Discussions link in README

smashwilson:
### Prerequisites

- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:. _(N/A)_
- [x] I have provided test coverage for my change (where applicable) _(N/A)_

### Description

👋🏻 Hello! The `.../discussions/category_choices` URLs for GitHub Discussions is going to stop working in a day or so. We noticed that you're linking there from your README, so I thought I'd send you a PR to update it before your link goes dead. 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Support the efforts of the development of the Six Labors projects.
 
 ## Questions
 
-- Do you have questions? We are happy to help! Please [join our Discussions Forum](https://github.com/SixLabors/ImageSharp/discussions/category_choices), or ask them on [Stack Overflow](https://stackoverflow.com) using the `ImageSharp` tag. Please do not open issues for questions.
+- Do you have questions? We are happy to help! Please [join our Discussions Forum](https://github.com/SixLabors/ImageSharp/discussions), or ask them on [Stack Overflow](https://stackoverflow.com) using the `ImageSharp` tag. Please do not open issues for questions.
 - Please read our [Contribution Guide](https://github.com/SixLabors/ImageSharp/blob/master/.github/CONTRIBUTING.md) before opening issues or pull requests!
 
 ## Code of Conduct  


antonfirsov:
@smashwilson thank you! Any chance you can sign the CLA, so we can merge the PR as-is?

JimBobSquarePants:
What? Why in all holy hell would someone introduce a breaking change for Discussions URLs?

You'll break all our repositories and many others. There's a golden rule for the web. Don't break links. 

You should provide redirects.

smashwilson:
> Why in all holy hell would someone introduce a breaking change for Discussions URLs?

This was an intermediate page from an old iteration of the discussion creation flow. It was overlooked when we changed the flow, but it's already out of date with reference to new behavior, so we're cleaning it up rather than risk it lagging farther behind.

This page has also had less than a hundred hits in the last week across all of GitHub. That's low enough that I _thought_ we could get away with cleaning it up for good. But...

> This change in GitHub's URLs will break all our issue templates for all our repositories.

🤔 When I looked before, I only found a small handful of READMEs that were linking directly to `.../discussions/category_choices` URLs, like this one, which seemed like a reasonable source of the lingering traffic. Now that I'm searching issue templates specifically I do see more `category_choices` links! It's weird that (a) these didn't show up in a broader search, and (b) they're apparently getting such a minuscule amount of clicks?

In any case that means we'll certainly leave a redirect around so that we don't break those templates. Sorry for the heart attack there 😅 

You may or may not want to change the README link here just the same? I'll have to double-check if it's okay for me to sign the CLA with my work email.

AraHaan:
> > Why in all holy hell would someone introduce a breaking change for Discussions URLs?
> 
> This was an intermediate page from an old iteration of the discussion creation flow. It was overlooked when we changed the flow, but it's already out of date with reference to new behavior, so we're cleaning it up rather than risk it lagging farther behind.
> 
> This page has also had less than a hundred hits in the last week across all of GitHub. That's low enough that I _thought_ we could get away with cleaning it up for good. But...
> 
> > This change in GitHub's URLs will break all our issue templates for all our repositories.
> 
> 🤔 When I looked before, I only found a small handful of READMEs that were linking directly to `.../discussions/category_choices` URLs, like this one, which seemed like a reasonable source of the lingering traffic. Now that I'm searching issue templates specifically I do see more `category_choices` links! It's weird that (a) these didn't show up in a broader search, and (b) they're apparently getting such a minuscule amount of clicks?

It is known by the .NET Foundation that doing broad searches through all of github does not yield every single possible line of code where one could use that type of links. Try for example searching the usage of ``HandleRef`` in https://github.com/dotnet/ you will for sure not get every single usage of that deprecated .NET API's. I think what needs to happen is the search feature to be enhanced to actually show all occurrence. It does get pretty bad when one has to ``git clone`` every repository in an organization and then manually recursively search on their PC for every real occurrence of the very thing they are looking for.

PR #1719, opened 2021-07-30
Use same byte order as IFD directory to decode pixels for 16 bit per channel images

### LINKED ISSUE #1716: Issues reading 48bit Tiff
### Discussed in https://github.com/SixLabors/ImageSharp/discussions/1715

<sup>Originally posted by **ehardesty** July 27, 2021</sup>
I am using SixLabors.Imagesharp 1.0.4-alpha.0.50 to work with a 48 bit Tiff image that came from a scanner. However, I am only getting a garbled imaged

Source image looks like this: 
image

For debug purposes I stripped the code down to simply loading the image and saving it back to disk as a PNG
```
using (var image = await Image.LoadAsync(file))
{
    await image.SaveAsPngAsync(_debugOutput + fileName + ".png");
}
```

Output image:
image

Note that the black rectangle was added to anonymize the image but the original pre-edit has the same result.

Am I doing something wrong here? (wouldn't be surprising) If not, should I create an issue?

[TiffIssue.zip](https://github.com/SixLabors/ImageSharp/files/6886984/TiffIssue.zip)
</div>

brianpopow:
…channel data, fixes #1716

### Prerequisites

### Description

Use same byte order as IFD directory to decode pixels for 16 bit per channel data, fixes #1716.

### DIFF
--- a/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/Rgb161616TiffColor{TPixel}.cs
+++ b/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/Rgb161616TiffColor{TPixel}.cs
@@ -1,0 +1,58 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System;
+using System.Buffers.Binary;
+using SixLabors.ImageSharp.Memory;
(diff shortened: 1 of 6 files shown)


JimBobSquarePants (approved):
Lovely stuff!

PR #1720, opened 2021-08-02
Respect byte order with 16 bit tiff images

brianpopow:
### Prerequisites

### Description

This **PR** changes the tiff decoder, so it will respect the byte order for 16 bit gray tiff images and 16 bit rgb planar images.

### DIFF
--- a/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/BlackIsZero16TiffColor{TPixel}.cs
+++ b/src/ImageSharp/Formats/Tiff/PhotometricInterpretation/BlackIsZero16TiffColor{TPixel}.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0.
 
 using System;
-using System.Buffers.Binary;
 using SixLabors.ImageSharp.Formats.Tiff.Utils;
 using SixLabors.ImageSharp.Memory;
 using SixLabors.ImageSharp.PixelFormats;
@@ -10,18 +9,18 @@ using SixLabors.ImageSharp.PixelFormats;
 namespace SixLabors.ImageSharp.Formats.Tiff.PhotometricInterpretation
 {
     /// <summary>
-    /// Implements the 'RGB' photometric interpretation with 16 bits for each channel.
+    /// Implements the 'BlackIsZero' photometric interpretation for 16-bit grayscale images.
     /// </summary>
-    internal class Rgb161616TiffColor<TPixel> : TiffBaseColorDecoder<TPixel>
+    internal class BlackIsZero16TiffColor<TPixel> : TiffBaseColorDecoder<TPixel>
         where TPixel : unmanaged, IPixel<TPixel>
     {
         private readonly bool isBigEndian;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="Rgb161616TiffColor{TPixel}" /> class.
+        /// Initializes a new instance of the <see cref="BlackIsZero16TiffColor{TPixel}" /> class.
         /// </summary>
         /// <param name="isBigEndian">if set to <c>true</c> decodes the pixel data as big endian, otherwise as little endian.</param>
-        public Rgb161616TiffColor(bool isBigEndian) => this.isBigEndian = isBigEndian;
+        public BlackIsZero16TiffColor(bool isBigEndian) => this.isBigEndian = isBigEndian;
 
         /// <inheritdoc/>
         public override void Decode(ReadOnlySpan<byte> data, Buffer2D<TPixel> pixels, int left, int top, int width, int height)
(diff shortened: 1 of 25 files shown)


src/ImageSharp/Formats/Tiff/PhotometricInterpretation/BlackIsZero16TiffColor{TPixel}.cs
```diff
@@ -0,0 +1,48 @@
...
+            var color = default(TPixel);
+
+            int offset = 0;
+
+            var l16 = default(L16);
+            for (int y = top; y < top + height; y++)
+            {
+                for (int x = left; x < left + width; x++)
+                {
+                    ushort intensity = TiffUtils.ConvertToShort(data.Slice(offset, 2), this.isBigEndian);
```
IldarKhayrutdinov, line 37: Great job! 
If there is insufficient performance, you can make separate loops for isBigEndian and !isBigEndian

brianpopow: good idea, changed with 6b538f89

src/ImageSharp/Formats/Tiff/PhotometricInterpretation/BlackIsZero16TiffColor{TPixel}.cs
```diff
@@ -32,15 +32,25 @@ public override void Decode(ReadOnlySpan<byte> data, Buffer2D<TPixel> pixels, in
...
-                    ushort intensity = TiffUtils.ConvertToShort(data.Slice(offset, 2), this.isBigEndian);
-                    offset += 2;
+                    for (int x = left; x < left + width; x++)
+                    {
+                        ushort intensity = TiffUtils.ConvertToShortBigEndian(data.Slice(offset, 2));
+                        offset += 2;
 
-                    l16.PackedValue = intensity;
-                    color.FromL16(l16);
+                        pixels[x, y] = TiffUtils.ColorFromL16(l16, intensity, color);
```
JimBobSquarePants, line 42: Are we using the `[x, y]` indexer like this elsewhere? We should process per row span to avoid the additional per-pixel multiplication.

brianpopow: yeah your right. There was other places used the indexer which i have changed, too. See 0ae95f17

brianpopow: weird, now a test fails only on `netcore2.1`: TiffDecoder_CanDecode_48Bit

brianpopow: it only fails for `netcore2.1` and only in Release mode, not in Debug mode.
output looks like this:
output

brianpopow: its again the bug with the defaults i reported recently: [netcore2.1 default bug](https://github.com/dotnet/runtime/issues/55623) 

I will try to work around it, but there are several places were defaults are used in the TiffColorDecoders

src/ImageSharp/Formats/Tiff/PhotometricInterpretation/BlackIsZero16TiffColor{TPixel}.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            int offset = 0;
+
+            var l16 = default(L16);
+            for (int y = top; y < top + height; y++)
+            {
+                Span<TPixel> pixelRow = pixels.GetRowSpan(y);
+                if (this.isBigEndian)
+                {
+                    for (int x = left; x < left + width; x++)
```
JimBobSquarePants, line 38: You can go one better for performance now you have the span and avoid the bounds checks.. 

```c#
Span<TPixel> pixelRow = pixels.GetRowSpan(y).Slice(left, left + width);
```

brianpopow: Im sorry, but I dont understand. Can you explain further? How would the loop change then?

src/ImageSharp/Formats/Tiff/PhotometricInterpretation/BlackIsZero16TiffColor{TPixel}.cs
brianpopow: I have changed the default usages introduced with this PR with values defined in `TiffUtils` now to workaround the issue with netcore2.1

JimBobSquarePants: Man, that bug is annoying!

src/ImageSharp/Formats/Tiff/PhotometricInterpretation/BlackIsZero16TiffColor{TPixel}.cs
JimBobSquarePants: Actually sorry, I misread.

The span declaration would be:

Span<TPixel> pixelRow = pixels.GetRowSpan(y).Slice(left, width);

Then the inner loop would become:

```c#
for (x = 0; x < pixelRow.Length; x++)
```

brianpopow: yeah ok, makes sense now, changed with f1834c78

JimBobSquarePants (approved):
Nice work!

Issue #1583, opened 2021-03-25
System.DivideByZeroException


3400442579:
### Prerequisites

- [ ] I have written a descriptive issue title
- [ ] I have verified that I am running the latest version of ImageSharp
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [ ] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
```
 System.AggregateException: 发生一个或多个错误。 ---> System.DivideByZeroException: 尝试除以零。
   在 SixLabors.ImageSharp.Processing.Processors.Dithering.OrderedDither.Dither[TPixel](TPixel source, Int32 x, Int32 y, Int32 bitDepth, Single scale)
   在 SixLabors.ImageSharp.Processing.Processors.Dithering.OrderedDither.QuantizeDitherRowOperation`2.Invoke(Int32 y)
   在 SixLabors.ImageSharp.Advanced.ParallelRowIterator.RowOperationWrapper`1.Invoke(Int32 i)
   在 System.Threading.Tasks.Parallel.<>c__DisplayClass17_0`1.<ForWorker>b__1()
   在 System.Threading.Tasks.Task.InnerInvoke()
   在 System.Threading.Tasks.Task.InnerInvokeWithArg(Task childTask)
   在 System.Threading.Tasks.Task.<>c__DisplayClass176_0.<ExecuteSelfReplicating>b__0(Object <p0>)
   --- 内部异常堆栈跟踪的结尾 ---
   在 System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   在 System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   在 System.Threading.Tasks.Task.Wait()
   在 System.Threading.Tasks.Parallel.ForWorker[TLocal](Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Func`4 bodyWithLocal, Func`1 localInit, Action`1 localFinally)
   在 System.Threading.Tasks.Parallel.For(Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body)
[...]

PR #1721, opened 2021-08-03
Do not use static options for default quantizer constructors.

JimBobSquarePants:
### Prerequisites

### Description

While investigating #1583 I realized that the static default options in all the quantizer instances could be overwritten which we should not allow. 

I've also ensured we throw a better exception should someone use the default current struct instances to dither.


antonfirsov (approved):
LGTM, thanks!

antonfirsov:
Wait ... should we do the same checks in `ApplyPaletteDither` ?

JimBobSquarePants:
>Wait ... should we do the same checks in `ApplyPaletteDither` ?

Yuuuuuuuup. Missed that, good catch! Done.

PR #1727, opened 2021-08-06
Add support for decoding tiff's with float pixel data

brianpopow:
### Prerequisites

### Description

This **PR** adds support for decoding Tiff images with float pixel data.

- Add support for decoding tiff's with 32bit float rgb pixel data
- Add support for decoding tiff's with 32bit float gray pixel data with min is black
- Add support for decoding tiff's with 32bit float gray pixel data with min is white


src/ImageSharp/Formats/Tiff/PhotometricInterpretation/BlackIsZero32FloatTiffColor{TPixel}.cs
```diff
@@ -0,0 +1,69 @@
...
+            int offset = 0;
+            for (int y = top; y < top + height; y++)
+            {
+                Span<TPixel> pixelRow = pixels.GetRowSpan(y).Slice(left, width);
+                if (this.isBigEndian)
+                {
+                    for (int x = 0; x < pixelRow.Length; x++)
+                    {
+                        data.Slice(offset, 4).CopyTo(buffer);
+                        Array.Reverse(buffer);
```
JimBobSquarePants, line 44: Can we not use `BinaryPrimitives.ReadSingleBigEndian` here and elsewhere?

brianpopow: `ReadSingleBigEndian` is only available with `net5.0` and above

JimBobSquarePants: Didn't know that. Looking at the source we could only polyfill back to .NET Core 2.0 also. We'll leave it like this for now.

JimBobSquarePants (approved):
Great stuff 👍

PR #1730, opened 2021-08-08
Unmanaged pooling MemoryAllocator

### LINKED ISSUE #1596: More efficient MemoryAllocator

### LINKED ISSUE #1675: Remove all uses of IManagedByteBuffer

### LINKED ISSUE #1739: Pixel processing breaking changes & API discussion
# Why breaking changes again?

Currently the library suffers from serious [memory management issues under high parallel load](https://user-images.githubusercontent.com/6835152/128640734-4844ade8-25a8-4670-b111-92391883cd80.png).

#1730 is here to deal with this, but we realized **we cannot guarantee the safety of `.GetPixelRowSpan(y)` API-s** after switching to unmanaged memory.

Although well written written user code is not at risk, missing or buggy `Image<T>` lifetime management can lead to memory corruption under certain circumstances. Consider the following code:

```C#
var image = new Image<Rgba32>(w, h); // missing using statement
Span<Rgba32> span = image.GetPixelRowSpan(0); // last use of image, GC is free to collect it

// GC happens here and finalizers have enough time to finish
// and potentially return image's underlying memory to the OS

span[0] = default;  // memory corruption
```

We cannot let a silent package update to turn performance issues and minor bugs into serious security problems, and we don't want to maintain safe-looking API-s which are in fact inherently dangerous, therefore we have to replace existing `Span<TPixel>` methods with delegate-based variants, which can be implemented safely.

We are about to deliver tons of **bugfixes and performance improvements** in our next NuGet release. We believe this will justify the breaking changes, so we decided to bump our major version number.

**ImageSharp 2.0 is coming!**

# Proposed API changes

## Processing pixel rows using `Span<T>`

The following methods are about to be deleted:

```diff
public class Image<TPixel>
{
-	public Span<TPixel> GetPixelRowSpan(int rowIndex);
}

public class ImageFrame<TPixel>
{
-	public Span<TPixel> GetPixelRowSpan(int rowIndex);	
}

public class Buffer2D<TPixel>
{
-	public Span<TPixel> GetRowSpan(int rowIndex);	
}
```

... and replaced by delegate based variants:

```C#
public readonly ref struct PixelAccessor<TPixel> where TPixel : unmanaged, IPixel<TPixel>
{
    // We expose the dimensions of the backing image
    // to make it easy to use PixelAccessor<TPixel> with static delegates.
    public int Width { get; }
    public int Height { get; }
    public Span<TPixel> GetRowSpan(int rowIndex);
}

public delegate void PixelAccessorAction<TPixel>(PixelAccessor<TPixel> pixelAccessor) where TPixel : unmanaged, IPixel<TPixel>;

public class Image<TPixel> where TPixel : unmanaged, IPixel<TPixel>
{
	public void ProcessPixelRows(PixelAccessorAction<TPixel> processPixels);
}

public class ImageFrame<TPixel> where TPixel : unmanaged, IPixel<TPixel>
{
	public void ProcessPixelRows(PixelAccessorAction<TPixel> processPixels);
}

public class Buffer2D<T> where T : unmanaged
{
	// This is an advanced API, mostly used internally.
	// We don't expect many users to interact with it.
	// Regardless, adding the Dangerous prefix seems reasonable.
 	public Span<T> DangerousGetRowSpan(int rowIndex);	
}
```

The delegate is responsible to make sure, that necessary bookkeeping is invoked around pixel processing (increasing `SafeHandle` ref counts). `PixelAccessor<TPixel>` has to be a `ref struct` to make sure that no-one saves it to the heap escaping the bookkeeping mechanisms.

#### Sample code

```C#
using var image = new Image<Rgb24>(256, 256);
image.ProcessPixelRows(static pixelAccessor =>
{
    for (int y = 0; y < pixelAccessor.Height; y++)
    {
        Span<Rgb24> row = pixelAccessor.GetRowSpan(y);

        // Using row.Length helps JIT to eliminate bounds checks when accessing row[x].
        for (int x = 0; x < row.Length; x++)
        {
            row[x] = new Rgb24((byte)x, (byte)y, 0);
        }
    }
});
```

#### Processing multiple images simultaneously

Things get tricky with use cases like #1666. The following code using nested delegates won't work, because `PixelAccessor<TPixel>` is a `ref struct` and can not be captured by a delegate:

```diff
- static Image<Rgba32> Extract(Image<Rgba32> sourceImage, Rectangle sourceArea)
- {
-     Image<Rgba32> targetImage = new (sourceArea.Width, sourceArea.Height);
-     int height = sourceArea.Height;
-     sourceImage.ProcessPixelRows(sourceAccessor =>
-     {
-         targetImage.ProcessPixelRows(targetAccessor =>
-         {
-             for (int i = 0; i < height; i++)
-             {
-                 // [CS1628] Cannot use ref, out, or in parameter 'sourceAccessor' inside an anonymous method, lambda expression, query expression, or local function
-                 Span<Rgba32> sourceRow = sourceAccessor.GetRowSpan(sourceArea.Y + i);
-                 Span<Rgba32> targetRow = targetAccessor.GetRowSpan(i);
- 
-                 sourceRow.Slice(sourceArea.X, sourceArea.Width).CopyTo(targetRow);
-             }
-         });
-     });
- 
-     return targetImage;
- }
```

To deal with this, we need to expose additional delegates and `ProcessPixelRows` overloads that work with multiple `PixelProcessors<TPixel>`-s simultaneously.

```C#
public delegate void PixelAccessorAction<TPixel1, TPixel2>(PixelAccessor<TPixel1> pixelAccessor1, PixelAccessor<TPixel2> pixelAccessor2)
        where TPixel1 : unmanaged, IPixel<TPixel1>
        where TPixel2 : unmanaged, IPixel<TPixel2>;

public delegate void PixelAccessorAction<TPixel1, TPixel2, TPixel3>(PixelAccessor<TPixel1> pixelAccessor1, PixelAccessor<TPixel2> pixelAccessor2, PixelAccessor<TPixel2> pixelAccessor3)
        where TPixel1 : unmanaged, IPixel<TPixel1>
        where TPixel2 : unmanaged, IPixel<TPixel2>
        where TPixel3 : unmanaged, IPixel<TPixel3>;

public class Image<TPixel> where TPixel : unmanaged, IPixel<TPixel>
{
	public void ProcessPixelRows<TPixel2>(Image<TPixel2> image2,
		PixelAccessorAction<TPixel> processPixels1, PixelAccessorAction<TPixel> processPixels2)
		where TPixel2 : unmanaged, IPixel<TPixel2>;

	public void ProcessPixelRows<TPixel2, TPixel3>(Image<TPixel2> image2, Image<TPixel3> image3,
		PixelAccessorAction<TPixel> processPixels1, PixelAccessorAction<TPixel> processPixels2, PixelAccessorAction<TPixel> processPixels3)
		where TPixel2 : unmanaged, IPixel<TPixel2>
	    where TPixel3 : unmanaged, IPixel<TPixel3>;	
}

// Also expose the same overloads on ImageFrame<TPixel>

```

Now we can implement the `Extract` method:

```C#
private static Image<Rgba32> Extract(Image<Rgba32> sourceImage, Rectangle sourceArea)
{
    Image<Rgba32> targetImage = new (sourceArea.Width, sourceArea.Height);
    int height = sourceArea.Height;
    sourceImage.ProcessPixelRows(targetImage, (sourceAccessor, targetAccessor) =>
    {
        for (int i = 0; i < height; i++)
        {
            Span<Rgba32> sourceRow = sourceAccessor.GetRowSpan(sourceArea.Y + i);
            Span<Rgba32> targetRow = targetAccessor.GetRowSpan(i);

            sourceRow.Slice(sourceArea.X, sourceArea.Width).CopyTo(targetRow);
        }
    });

    return targetImage;
}
```

## Processing pixel rows using `Memory<T>`

There are some less-known overloads in `SixLabors.ImageSharp.Advanced.AdvancedImageExtensions` that allow accessing image rows through `Memory<T>`. We should consider adding the `Dangerous` prefix to them because accessing `Span<TPixel>` through `image.GetPixelRowMemory(y).Span` suffers from the same issues as `image.GetPixelRowSpan(y)`.

```diff
public static class AdvancedImageExtensions
{
-	public static Memory<TPixel> GetPixelRowMemory<TPixel>(this Image<TPixel> source, int rowIndex);
-	public static Memory<TPixel> GetPixelRowMemory<TPixel>(this ImageFrame<TPixel> source, int rowIndex)
+	public static Memory<TPixel> DangerousGetPixelRowMemory<TPixel>(this Image<TPixel> source, int rowIndex);
+	public static Memory<TPixel> DangerousGetPixelRowMemory<TPixel>(this ImageFrame<TPixel> source, int rowIndex)
}
```

Although these methods are alredy hidden behind the `*.Advanced` namespace, I'm in favor of this change, since users can abuse them to acquire a pixel row span with `image.GetPixelRowMemory(y).Span` instead of the `ProcessPixelRows` delegate API-s.

## Getting a single pixel span to `Image<TPixel>`

The method `Image.TryGetSinglePixelSpan(out Span<TPixel>)` is mostly used for interop scenarios, to acquire a `Span<TPixel>` that could be pinned so users can pass the image's pointer to the GPU or other image processing libraries. Not surprisingly, we have to replace this method:

```diff
public class Image<TPixel>
{
-	public bool TryGetSinglePixelSpan(out Span<TPixel> span)
+   public bool DangerousTryGetSinglePixelMemory(out Memory<TPixel> memory);
}
```

#### API usage

```C#
using var image = new Image<Rgba32>(4096, 4096);

if (!image.DangerousTryGetSinglePixelMemory(out Memory<Rgba32> memory))
{
    throw new Exception("Make sure to initialize MemoryAllocator.Default!");
}

using (MemoryHandle imageMemoryHandle = memory.Pin())
{
    IntPtr resourceHandle = Interop.AllocateSomeUnmanagedResourceToWorkWithPixelData(
        (IntPtr)imageMemoryHandle.Pointer,
        image.Width * image.Height);
    Interop.WorkWithUnmanagedResource(resourceHandle);
    Interop.FreeUnmanagedResource(resourceHandle);
}
```

#### Initializing `MemoryAllocator`

**Update:** The proposal to guarantee contiguous memory at alloactor level has been replaced by `Configuration.PreferContiguousImageBuffers`: https://github.com/SixLabors/ImageSharp/issues/1739#issuecomment-954078801

~After #1730, an image of `4096 x 4096` pixels will no longer fit into a single contiguous memory buffer by default.
The following initialization code has to be executed **once** during startup, to make sure that an image of this size is contiguous:~

##### Limiting pool size

```C#
MemoryAllocator.Default = MemoryAllocator.Create(
	new MemoryAllocatorOptions()
    {
        MaximumPoolSizeMegabytes= 8
    });
```

# Feedback is welcome!

/cc @saucecontrol @Sergio0694 @br3aker @alistaircarscadden @ThomasMiz @Veriara @DaZombieKiller @TanukiSharp @ptasev

antonfirsov:
### UPDATE 2: Ready for review!
- [x] Although later than used to, but 32 bit still fails with `OutOfMemoryException`. Consider retrying `Marshal.AllocHGlobal` on `OutOfMemoryException` after a short wait. *DONE: We are blocking the thread on OOM to retry allocations. 32 bit is 2x slower with 20 Threads than 64 bit, but doesn't OOM. The retries alone are not responsible for the 2x slowdown, 32bit runtime seems to work 1.5x slower also with 10 threads, when there are no OOMs.*
- [x] Implement `PreferContiguousImageBuffers`, remove `MemoryAllocator.MinimumContiguousBlockSizeBytes`.
- [x] Implement `PixelAccessor<T>` and other #1739 stuff
- [x] Remove `ArrayPoolMemoryAllocator`
- [x] Implement some other minor API changes, eg. `MemoryAllocator.Default`. -- Need to change `MemoryAllocator.Default`, it should be get-only.
- [x] Fix memory corruption / resource leak bugs

### Prerequisites

### Description

This PR introduces `UniformUnmanagedMemoryPoolMemoryAllocator` and sets it as default to fix #1596.

### UniformUnmanagedMemoryPoolMemoryAllocator functional characteristics

- For allocations below 1MB use `ArrayPool<byte>.Shared`
- For allocations above 1MB use `UniformUnmanagedMemoryPool` to allocate **4 MB** blocks of discontiguous unmanaged memory, of up to the pool's limit
- Over pool limit, allocate unmanaged buffers split to discontiguous blocks of 32 MB. These will be freed immediately

#### Pool size

According to my benchmaks, the pool should scale to the maximum desired size to achieve the best througput. There is no point placing an artificial pool limit, unless there is a physical limitation. I decided to set the maximum pool size to 1/8 th of the available physical memory in 64 bit .NET Core processes. This means that on a 16GB machine the pool can grow as large as 2 GB.

On 32 bit, and other (non-testable) platforms the pool limit is 128 MB.

#### Trimming

The trimming of the pools is triggered by both Gen 2 GC collect and a timer. (We need the timer since unmanaged allocations don't trigger GC)  On high load we trim the entire pool, on low load we trim 50% of the pool every minute.

#### Finalizers

With `ArrayPoolMemoryAllocator`, if an image is GC-d without being disposed, buffers are never returned to the pool. This means no hard memory leak, but the pools will be eventually exhausted, because the [bucket's running index](https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Private.CoreLib/src/System/Buffers/ConfigurableArrayPool.cs#L207) hitting the bucket limit.

To avoid this, `MemoryGroup<T>.Owned` and `UniformUnmanagedMemoryPool.FinalizableBuffer<T>` have finalizers returning the `UnmanagedMemoryHandle` to the pool. This can get tricky, since `UnmanagedMemoryHandle` is also finalizable:

https://github.com/SixLabors/ImageSharp/blob/1a41aaa10b7f7e3f7cd078507f500a0c4b47a010/src/ImageSharp/Memory/Allocators/Internals/UnmanagedMemoryHandle.cs#L58-L78

I'm moderately concerned about [CA2015](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2015), but I don't think it applies to us. Dispose will also free the memory used by a span. Touching a span or a pointer to `SkiaSharp` image's memory would be also a bug if the image is finalized.

### API changes

Resolves #1739
Fixes #1675

**Edit:** API changes implemented according to #1739.

### Benchmarking methodology

To determine these defaults I compared results of LoadResizeSaveParallelMemoryStress runs systematically, typically running them for a varying parameter a couple of times, while fixing all other parameters. I have a bunch of Excel documents comparing the tables, including all of them would be TLDR, but I can present information on request.

### Benchmark results

I was benchmarking on a 10 Core (20 Thread) 19-10900X with 64 GB RAM. This means I was able to stress highly parallel workload very extensive allocation pressure.

Here is an median processing time (seconds) of 40 runs of [`LoadResizeSaveParallelMemoryStress`](https://github.com/SixLabors/ImageSharp/blob/1a41aaa10b7f7e3f7cd078507f500a0c4b47a010/tests/ImageSharp.Tests.ProfilingSandbox/LoadResizeSaveParallelMemoryStress.cs) ("Classic" means `ArrayPoolMemoryAllocator`):

| Classic | Nopool  | Unmanaged Pool |
|---------|---------|----------------|
| 9.038   | 8.988   | 8.306          |
| 100.56% | 100.00% | 92.41%         |

ImageSharp is about 8% faster with the new default memory allocator.

Results of `LoadResizeSaveStressBenchmarks` BDN benchmark also show 7.5% improvement:

|                              Method | maxDegreeOfParallelism |       Mean |    Error |   StdDev | Ratio |     Gen 0 |     Gen 1 |     Gen 2 |    Allocated |
|------------------------------------ |----------------------- |-----------:|---------:|---------:|------:|----------:|----------:|----------:|-------------:|
|                       SystemDrawing |                     20 | 3,007.7 ms | 151.1 ms |  8.28 ms |  1.00 |         - |         - |         - |        30 KB |
|   ImageSharp_DefaultMemoryAllocator |                     20 |   682.6 ms | 395.6 ms | 21.69 ms |  0.23 |         - |         - |         - |     2,413 KB |
| ImageSharp_ArrayPoolMemoryAllocator |                     20 |   734.9 ms | 174.4 ms |  9.56 ms |  0.24 | 1000.0000 | 1000.0000 | 1000.0000 | 1,237,298 KB |
|                              Magick |                     20 | 1,545.1 ms | 624.1 ms | 34.21 ms |  0.51 |         - |         - |         - |       106 KB |
|                         MagicScaler |                     20 |   389.5 ms | 163.7 ms |  8.97 ms |  0.13 |         - |         - |         - |       101 KB |
|                          SkiaBitmap |                     20 |   889.6 ms | 182.9 ms | 10.03 ms |  0.30 |         - |         - |         - |       118 KB |
|                             NetVips |                     20 |   441.4 ms | 686.1 ms | 37.61 ms |  0.15 |         - |         - |         - |       102 KB |

#### VirtualAlloc commit lifetimes graph with ArrayPoolMemoryAllocator

image

#### VirtualAlloc commit lifetimes graph with the new allocator, demonstrating the trimming

image

#### VirtualAlloc commit lifetimes graph with pool size set to zero

image

I would be happy to see some expert feedback on this solution, especially for the finalizer tricks.

/cc @Sergio0694 @saucecontrol @br3aker


br3aker:
> (We need the timer since unmanaged allocations don't trigger GC)

Might be worth testing trimming solely via [GC.AddMemoryPressure](https://docs.microsoft.com/en-us/dotnet/api/system.gc.addmemorypressure?view=net-5.0) & gen2 callback?

antonfirsov:
@br3aker I'm not sure if AddMemoryPressure / RemoveMemoryPressure contributes to Gen2 allocation budget or not. Might make sense to try out, but I think the timer is not that expensive, and trimming is configured to be done every 1 minute (mimicking ArrayPool<T>.shared) anyways.

Directory.Build.props
```diff
@@ -26,4 +26,7 @@
     <Optimize>true</Optimize>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <DefineConstants Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','netcoreapp3.1'))">$(DefineConstants);NETCORE31COMPATIBLE</DefineConstants>
```
antonfirsov, line 30: @JimBobSquarePants what we need here is really to tell if we are building against .NET Core 3.1+
It's not only about particular API features, I don't have confidence if `Gen2GcCallback` would reliably work on other targets (which are also poorly covered by tests), so I prefer to exclude the whole thing unless we on .NET Core.

JimBobSquarePants:
Cannot wait to get stuck into reading this! 🤩

antonfirsov:
@JimBobSquarePants been thinking a lot & chatting with folks on the C#/lowlevel discord channel, I'm no longer sure if using unsafe memory is the right thing to do.

The problem is that an update to 1.1 may turn minor bugs and performance issues into security errors for users of `GetRowSpan(y)` and `TryGetSinglePixelSpan()`. These are methods which look completely safe, but will become inherently unsafe with this PR. (See the warnings added to docs in c9d13965e3badd7ecc228cf69359ffe1838c2149)

On the other hand, SkiaSharp merged a similar API without spending a single minute on security concerns: https://github.com/mono/SkiaSharp/pull/1242

JimBobSquarePants:
@antonfirsov That's the `IMemory<T>` finalizer warning yeah? That would require some impressively bad code to trigger but you're right we should do a careful review.

I need to do some reading there and maybe see if we can get someone from the runtime team who worked in that area to comment.

antonfirsov:
Strictly speaking, the finalizer warning doesn't apply to us since the `SafeHandle` is finalizable and returns memory anyways.

> That would require some impressively bad code to trigger 

First I also thought so, but in fact it can be as simple as:

```
var image = new Image<Rgba32>(w, h); // no using
Span<Rgba32> span = image.GetPixelRowSpan(0); // last use of the object `image`, finalizers may run after this point

// some relatively long running code here to allow the finalizers to finish

span[0] = default;  // memory corruption
```

JimBobSquarePants:
I wonder if we could write an analyzer?

antonfirsov:
> I need to do some reading there and maybe see if we can get someone from the runtime team who worked in that area to comment.

That would be awesome, but I'm afraid the would share the concerns around unsafe memory and push back.

> I wonder if we could write an analyzer?

Would it work out of the box just by using the library? Note that it won't help existing users doing a package update without recompilation, and then running into a potential security issue.

JimBobSquarePants:
The trick, I think, would be to make the analyzer a dependency of the main library Like [Xunit ](https://www.nuget.org/packages/xunit/) do.

Have we made any breaking changes that require recompilation? Maybe we should just to ensure people should rebuild. 👿

antonfirsov:
> **GrabYourPitchforks** — Yesterday at 6:09 PM
> The original design for Memory didn't have a span accessor; you had to provide a delegate and the span would be passed as a parameter to the delegate. This solved many (but not all) ownership problems.


There is a safe, breaking way to re-implement span accessors by using delegates, inspired by the comment above:

```diff
  public class Image<T>
  {
-     public Span<T> GetPixelRowSpan(int y);
-      public bool TryGetSinglePixelSpan(out Span<TPixel> span);
+     public void ProcessPixels(Action<PixelAccessor<T>> rowProcessor);
+     public bool TryProcessSinglePixelSpan(SpanAction<T> pixelProcessor);
  }

+ public struct PixelAcceessor<T>
+ {
+     Span<T> GetRowSpan(int y);
+ }
```

The simplest thing would be to go ahead with this breaking change and **bump ImageSharp version number to 2.0**. The improvements will justify the change.

JimBobSquarePants:
> bump ImageSharp version number to 2.0

2.0 was to be my kill all old target frameworks release. I want to ship a working V1 of Fonts and Drawing before starting work on it.

antonfirsov:
That can be 3.0 then. We follow semantic versioning more or less no, no point to be afraid of major version jumps as breaking changes land IMO.

antonfirsov:
But I'm also fine with a hard-breaking 1.1, this is more about PR and communication than anything else,

Renaming the milestones seems to be better thing to do for me, we can even benefit out of it.

JimBobSquarePants:
My only issue with jumping from 2.0 to 3.0 would that in real terms it would probably occur over a short timespan which, in my opinion does reflect well on the quality. V1.1 be, by far, my next desired target.

This is a massive breaking change though so I'm deeply conflicted. 🙁

Sergio0694:
> Disclaimer: I'm not home and with barely functioning internet, so might've missed some bits here.

I have a question regarding `PixelAcceessor<T>`, isn't that quite similar to what we already offer with the `ProcessPixelRows___` APIs already? Given that those also allow working on arbitrary pixel types and offer batched and vectorized pixel format conversion, should we consider just pushing devs currently accessing individual pixels towards those APIs instead? If not and if we still want to offer a way to access individual pixel specific values, should we align the design of these new APIs to follow those, for consistency? Just some random thoughts here 😄

Btw amazing work on all this @antonfirsov, I'll also need to find some time to carefully go through all this like James said and have a proper read, as the whole investigation seems super interesting! 🚀

JimBobSquarePants:
After careful consideration. I'm up for a V2 release. It's good opportunity to fix a few things plus we are already adding a significant amount of fixes/functionality to the release so let's make a show of it.

antonfirsov:
@Sergio0694 yeah the shape of [PixelRowDelegateExtensions](https://github.com/SixLabors/ImageSharp/blob/97dde7fb2f0ac97b5b1f9e6b0555a53d2e270c46/src/ImageSharp/Processing/Extensions/Effects/PixelRowDelegateExtensions.cs) is very similar to `PixelAcceessor<T>`, but we need `PixelAcceessor<T>` to work with `Image` directly to make it easier for today's `GetRowSpan` users. I'm open to ideas that may reduce API bloat, but I don't have the capacity to figure it out myself :(

Directory.Build.props
JimBobSquarePants: We should have access to `NETCOREAPP3_1_OR_GREATER` now I believe.

src/ImageSharp/ImageSharp.csproj
```diff
@@ -13,6 +13,7 @@
     <PackageTags>Image Resize Crop Gif Jpg Jpeg Bitmap Png Tga NetCore</PackageTags>
     <Description>A new, fully featured, fully managed, cross-platform, 2D graphics API for .NET</Description>
     <Configurations>Debug;Release;Debug-InnerLoop;Release-InnerLoop</Configurations>
+    <DefineConstants Condition="'$(Configuration)' == 'Debug-InnerLoop'">$(DefineConstants);DEBUG</DefineConstants>
```
JimBobSquarePants, line 16: I'm surprised this is required, however I had to do something similar to ensure release builds are optimized. You'll want to move it up to the root `Directory.Build.props` to live with that command.

antonfirsov: Yeah, I noticed this by realizing that `DebugGuard` doesn't work in `Debug-Innerloop`. Will promote to `Directory.Builld.props`, but might be something to look into at SharedInfrastructure level even.

antonfirsov:
I'm marking this ready for review, since I fixed the memory corruption issues in 77e7700 and finalized the API changes.

Thanks to further recent improvements like #1694 and #1773, we now match (and slightly beat) skia when used without target size decoding.

### Skia - Regular
image

### ImageSharp

image

### Skia - Decode to target size

image

tests/ImageSharp.Tests/RunTestsInLoop.ps1
```diff
@@ -0,0 +1,22 @@
+﻿# This script can be used to collect logs from sporadic bugs
+Param(
```
antonfirsov, line 2: This started as personal sandboxing, but can be useful for stressing local runs in the future, so decided to keep it.

Directory.Build.props
antonfirsov: Depends if we are using [5.0.2xx SDK](https://github.com/dotnet/sdk/pull/14798) DK on CI. Likely yes, but we need to be 100% sure to a void building the wrong thing. I can't decide based on the build logs.

src/ImageSharp/Memory/Allocators/MemoryAllocatorSettings.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Memory
+{
+    /// <summary>
+    /// Defines options for creating the default <see cref="MemoryAllocator"/>.
+    /// </summary>
+    public struct MemoryAllocatorSettings
```
antonfirsov, line 9: I named this `MemoryAllocatorSettings` instead of a more conventional name candidate `MemoryAllocatorOptions`, so it's not confused with `MemoryAllocationOptions`.

Directory.Build.props
JimBobSquarePants: Looks like they're all using 5.0.403. You can see what's installed by following each link in the readme.

https://github.com/actions/virtual-environments

src/ImageSharp/Common/Helpers/SimdUtils.Shuffle.cs
```diff
@@ -182,9 +183,9 @@ private static void VerifyShuffle3SpanInput<T>(ReadOnlySpan<T> source, Span<T> d
             where T : struct
         {
             DebugGuard.IsTrue(
-                source.Length == dest.Length,
+                source.Length <= dest.Length,
                 nameof(source),
-                "Input spans must be of same length!");
+                "Source should fit into dest!");
```
JimBobSquarePants, line 188: Did we have a scenario where they weren't the same length? When I wrote this `IShuffle3` and `IShuffle4` were mirrored implementations designed for working against specific pixel formats. `VerifyShuffle4SpanInput` still checks for equality.

antonfirsov: The original version fails [`EncodeBaseline_WorksWithDiscontiguousBuffers<Bgr24>`](https://github.com/SixLabors/ImageSharp/blob/424d4f94357f6efa66b436e59be367fe6b0c4dc6/tests/ImageSharp.Tests/Formats/Jpg/JpegEncoderTests.cs#L209) on current master, we just did not notice it since CI doesn't run tests in Debug (maybe it should?).

`this.rgbSpan` is longer than `this.pixelSpan` here:
https://github.com/SixLabors/ImageSharp/blob/659daffad916a20a5b62399089c212033c7494e4/src/ImageSharp/Formats/Jpeg/Components/Encoder/YCbCrForwardConverter420%7BTPixel%7D.cs#L109

because of the padding added by @br3aker:
https://github.com/SixLabors/ImageSharp/blob/659daffad916a20a5b62399089c212033c7494e4/src/ImageSharp/Formats/Jpeg/Components/Encoder/YCbCrForwardConverter420%7BTPixel%7D.cs#L88

I believe we should follow `Span.CopyTo(source, dest)` semantics in this family of methods, which allows `dest` to be longer and processes `source.Length` items.

src/ImageSharp/Processing/Processors/ImageProcessor{TPixel}.cs
```diff
@@ -45,27 +45,15 @@ protected ImageProcessor(Configuration configuration, Image<TPixel> source, Rect
         /// <inheritdoc/>
         void IImageProcessor<TPixel>.Execute()
         {
-            try
-            {
-                this.BeforeImageApply();
+            // TODO: Try-catch logic temporarily removed, put it back.
```
JimBobSquarePants, line 48: Can remove the TODO here since we're ditching the wrapper

tests/ImageSharp.Tests/Memory/Allocators/UniformUnmanagedMemoryPoolTests.Trim.cs
```diff
@@ -0,0 +1,150 @@
...
+                    Assert.True(
+                        UnmanagedMemoryHandle.TotalOutstandingHandles <= 80,
+                        $"UnmanagedMemoryHandle.TotalOutstandingHandles={UnmanagedMemoryHandle.TotalOutstandingHandles} > 80");
+                    pool.Return(b);
+                }
+            }
+
+            public static readonly bool IsNotMacOs = !TestEnvironment.IsOSX;
+
+            // TODO: Investigate failures on MacOS. All handles are released after GC.
```
JimBobSquarePants, line 52: Maybe worth raising an issue to track.

tests/ImageSharp.Tests/Memory/Allocators/UniformUnmanagedMemoryPoolTests.cs
```diff
@@ -0,0 +1,379 @@
...
+            Assert.NotNull(b0);
+            cleanup.Register(b0);
+            UnmanagedMemoryHandle[] b1 = pool.Rent(attempt);
+            Assert.NotNull(b1);
+            cleanup.Register(b1);
+        }
+
+        public static readonly bool IsNotMacOS = !TestEnvironment.IsOSX;
+
+        // TODO: Investigate MacOS failures
```
JimBobSquarePants, line 250: Are these intermittent? Restricted to certain target frameworks?

JimBobSquarePants (commented):
This is an extraordinary body of work. Surprisingly simply to follow though giving the subject matter. Well done!

src/ImageSharp/Common/Helpers/SimdUtils.Shuffle.cs
JimBobSquarePants: >CI doesn't run tests in Debug (maybe it should?).

Oh please no! They take long enough!!

>I believe we should follow `Span.CopyTo(source, dest)` semantics in this family of methods, which allows dest to be longer and processes `source.Length` items.

I agree then and we should update `VerifyShuffle4SpanInput ` to match.

src/ImageSharp/Processing/Processors/Dithering/ErrorDither.cs
```diff
@@ -105,16 +106,23 @@ public ErrorDither(in DenseMatrix<float> matrix, int offset)
...
 
             for (int y = bounds.Top; y < bounds.Bottom; y++)
             {
-                ref TPixel sourceRowRef = ref MemoryMarshal.GetReference(source.GetPixelRowSpan(y));
-                ref byte destinationRowRef = ref MemoryMarshal.GetReference(destination.GetWritablePixelRowSpanUnsafe(y - offsetY));
+                // Unsafe optimizations undone temporarily.
+                // Sporadic local AccessViolationException indicates possible indexing bug.
+                // ref TPixel sourceRowRef = ref MemoryMarshal.GetReference(source.DangerousGetRowSpan(y));
+                // ref byte destinationRowRef = ref MemoryMarshal.GetReference(destination.GetWritablePixelRowSpanUnsafe(y - offsetY));
+                Span<TPixel> sourceSpan = sourceBuffer.DangerousGetRowSpan(y);
```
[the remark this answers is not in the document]
antonfirsov: Oh good catch, it was actually my finalization bug which for some reason happened more often around quantization, we are safe to change this back now.

tests/ImageSharp.Tests/Memory/Allocators/UniformUnmanagedMemoryPoolTests.Trim.cs
antonfirsov: Yeah will do after the merge, when it officially becomes the product code.

tests/ImageSharp.Tests/Memory/Allocators/UniformUnmanagedMemoryPoolTests.cs
antonfirsov: This one was intermittent, the one above *seemed* to occur consistently on .NET 6, but nothing is proven without having a Mac ...
Here it's probably just a test issue because of the OS returning identical addresses in a follow-up allocation, despite my attempt to push them out by dummy allocations:

https://github.com/SixLabors/ImageSharp/blob/7ec7447808725f761dc7cba83fea676c4420158f/tests/ImageSharp.Tests/Memory/Allocators/UniformUnmanagedMemoryPoolTests.cs#L266-L267

src/ImageSharp/Common/Helpers/SimdUtils.Shuffle.cs
antonfirsov: We can introduce a new `Checked` build configuration, like many C/C++ projects do. It's basically Release build but with all the checks/guards enabled.

src/ImageSharp/Memory/Allocators/MemoryAllocatorSettings.cs
antonfirsov: > You mean `AllocationOptions` yeah ?

Yeah, I meant that. It would be still somewhat confusing.

I'm not sure if `BufferInitializationBehavior` is a future-proof name. We can potentially introduce flags for alignment constraints or similar stuff.

brianpopow:
Maybe its worth a try to get rid of `[Collection("RunSerial")]` workaround from the Encoder/Decoder test?

antonfirsov:
> Maybe its worth a try to get rid of `[Collection("RunSerial")]` workaround from the Encoder/Decoder test?

Yeah we should try it, might save some execution time, and see if 32 bit OOMs (it should, at least not from ImageSharp code). Would be better to do it in a separate PR though.

src/ImageSharp/Memory/Allocators/Internals/UnmanagedMemoryHandle.cs
```diff
@@ -0,0 +1,140 @@
...
+                {
+                    // We are low on memory, but expect some memory to be freed soon.
+                    // Block the thread & retry to avoid OOM.
+                    if (counter < MaxAllocationAttempts)
+                    {
+                        counter++;
+                        Interlocked.Increment(ref totalOomRetries);
+
+                        Interlocked.CompareExchange(ref lowMemoryMonitor, new object(), null);
+                        Monitor.Enter(lowMemoryMonitor);
```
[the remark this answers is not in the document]
antonfirsov: Also took me a while to find out that `Monitor.PulseAll` is the synchronization primitive I'm looking for. Still wondering if it's possible to do it more efficiently, but hopefully this code path will be hit only in 0.001% of the cases. It's is not covered by unit tests. After merging the latest master I had to emulate leaks and bump the pool size to see any first chance OOM-s.

src/ImageSharp/Memory/Allocators/UniformUnmanagedMemoryPoolMemoryAllocator.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        public override void ReleaseRetainedResources() => this.pool.Release();
+
+        private static long GetDefaultMaxPoolSizeBytes()
+        {
+#if NETCORE31COMPATIBLE
+            // On 64 bit .NET Core 3.1+, set the pool size to a portion of the total available memory.
+            // There is a bug in GC.GetGCMemoryInfo() on .NET 5 + 32 bit, making TotalAvailableMemoryBytes unreliable:
+            // https://github.com/dotnet/runtime/issues/55126#issuecomment-876779327
+            if (Environment.Is64BitProcess || !RuntimeInformation.FrameworkDescription.StartsWith(".NET 5.0"))
```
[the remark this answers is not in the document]
antonfirsov: Yeah, I'm very happy the runtime folks fixed it.

src/ImageSharp/Memory/Allocators/Internals/UnmanagedMemoryHandle.cs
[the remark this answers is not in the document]
JimBobSquarePants: Yeah this will be very rare. I think this is a good solution though.

src/ImageSharp/Memory/Allocators/MemoryAllocatorSettings.cs
JimBobSquarePants: `BufferInitializationOptions` then. I think that would be enough to allow us to keep `MemoryAllocationOptions` and allow future-proofing. I'd really like to maintain the convention on upper API terminology.

JimBobSquarePants: If we update the naming I'm more than happy for this to get merged in.

antonfirsov: I don't like `BufferInitializationOptions` since it's really not only about initialization. I think `AllocationOptions` describes the purpose perfectly, I'd prefer to keep that name.

I would rather rename `MemoryAllocatorSettings` to `MemoryAllocatorCreationOptions`. A bit verbose, but I can't come up with a better compromise.

JimBobSquarePants: I'm curious about these potential future flags? We'd have to change the type from an enum to accommodate them. 

I think we should go with `MemoryAllocationOptions` at the upper level and make any compromise with  `AllocationOptions`. I don't actually agreed that the latter describes the purpose well. It's very abstract.

antonfirsov: > I'm curious about these potential future flags? 

For example

```C#
using var mem = memoryAllocator.Allocate<byte>(AllocationOptions.Aligned32);
```

would allocate a buffer where the starting address is divisible by 32 (aligned to 32 byte boundaries). This results in faster AVX loads.

> I don't actually agreed that the latter describes the purpose well. It's very abstract.

I don't see why, it defines flags customizing the `Allocate` call. I think `MemoryAllocatorOptions` / `MemoryAllocatorCreationOptions` is a very narrow customer-facing use case, while `AllocationOptions` is everywhere in our codebase, heavily used by community contributors, I don't like the idea of forcing an unnatural name there.

JimBobSquarePants: Ahhhh right! You meant literal flags! Your example clears everything up there. `AllocationOptions` is fine. Sorry 🙏

`Creation` feels like an odd word then, not a computing term. I think we should just go with `MemoryAllocatorOptions` if the use case is so narrow. You'd have to be coding with Notepad not to see the difference.

antonfirsov: > `Creation` feels like an odd word then, not a computing term.

I think [I get what you mean](https://www.google.com/search?q=creation&tbm=isch) 😆. Let's keep it `MemoryAllocatorOptions` then!

JimBobSquarePants:
@antonfirsov Just merged master into this branch locally and ran the tests. 

The following tests failed for me when running all tests at one with Release-InnerLoop and .NETCore 3.1 x64
MemoryGroupFinalizer_ReturnsToPool
MultiplePoolInstances_TrimPeriodElapsed_AllAreTrimmed
GC_Collect_OnHighLoad_TrimsEntirePool

The first two pass when ran individually but I cannot get the last to pass at all.

antonfirsov:
> MemoryGroupFinalizer_ReturnsToPool

There are 3 cases (300, 600, 1200), I wonder which one fails?

> MultiplePoolInstances_TrimPeriodElapsed_AllAreTrimmed

This test is timing dependent, if the CPU is under high load, it can happen that the timer callbacks fire late or timings misalign in other means. This can explain the MacOS failures too, I will try to move this test to a separate serial collection.

> GC_Collect_OnHighLoad_TrimsEntirePool

This one is only supposed to run on 32 bit, VS test runner has trouble understanding `[ConditionalFact]`, but `dotnet test` should skip it on x64.

JimBobSquarePants:
Some more detail on the failures on my machine. They are repeatable by running each theory group.

`MemoryOwnerFinalizer_ReturnsToPool` fails for 600.
```
Message: 
Microsoft.DotNet.RemoteExecutor.RemoteExecutionException : Remote process failed with an unhandled exception.

  Stack Trace: 
Child exception:
  Xunit.Sdk.EqualException: Assert.Equal() Failure
Expected: 42
Actual:   240
UniformUnmanagedPoolMemoryAllocatorTests.AllocateSingleAndForget(UniformUnmanagedMemoryPoolMemoryAllocator allocator, Int32 length, Boolean check) line 346
UniformUnmanagedPoolMemoryAllocatorTests.<MemoryOwnerFinalizer_ReturnsToPool>g__RunTest|14_0(String lengthStr) line 326
Child process:
  SixLabors.ImageSharp.Tests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13 SixLabors.ImageSharp.Tests.Memory.Allocators.UniformUnmanagedPoolMemoryAllocatorTests Void <MemoryOwnerFinalizer_ReturnsToPool>g__RunTest|14_0(System.String)
Child arguments:
  600
```


`MemoryGroupFinalizer_ReturnsToPool` fails for 600 and 1200.

```bash
Message: 
Microsoft.DotNet.RemoteExecutor.RemoteExecutionException : Remote process failed with an unhandled exception.

  Stack Trace: 
Child exception:
  Xunit.Sdk.EqualException: Assert.Equal() Failure
Expected: 42
Actual:   224
UniformUnmanagedPoolMemoryAllocatorTests.AllocateGroupAndForget(UniformUnmanagedMemoryPoolMemoryAllocator allocator, Int32 length, Boolean check) line 295
UniformUnmanagedPoolMemoryAllocatorTests.<MemoryGroupFinalizer_ReturnsToPool>g__RunTest|12_0(String lengthStr) line 277
Child process:
  SixLabors.ImageSharp.Tests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13 SixLabors.ImageSharp.Tests.Memory.Allocators.UniformUnmanagedPoolMemoryAllocatorTests Void <MemoryGroupFinalizer_ReturnsToPool>g__RunTest|12_0(System.String)
Child arguments:
  600
```

```bash
Message: 
Microsoft.DotNet.RemoteExecutor.RemoteExecutionException : Remote process failed with an unhandled exception.

  Stack Trace: 
Child exception:
  Xunit.Sdk.EqualException: Assert.Equal() Failure
Expected: 42
Actual:   0
UniformUnmanagedPoolMemoryAllocatorTests.AllocateGroupAndForget(UniformUnmanagedMemoryPoolMemoryAllocator allocator, Int32 length, Boolean check) line 295
UniformUnmanagedPoolMemoryAllocatorTests.<MemoryGroupFinalizer_ReturnsToPool>g__RunTest|12_0(String lengthStr) line 277
Child process:
  SixLabors.ImageSharp.Tests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13 SixLabors.ImageSharp.Tests.Memory.Allocators.UniformUnmanagedPoolMemoryAllocatorTests Void <MemoryGroupFinalizer_ReturnsToPool>g__RunTest|12_0(System.String)
Child arguments:
  1200

```

`MultiplePoolInstances_TrimPeriodElapsed_AllAreTrimmed` is also failing for me. 

```bash
Message: 
Microsoft.DotNet.RemoteExecutor.RemoteExecutionException : Remote process failed with an unhandled exception.

  Stack Trace: 
Child exception:
  Xunit.Sdk.EqualException: Assert.Equal() Failure
Expected: 128
Actual:   0
NonParallel.<MultiplePoolInstances_TrimPeriodElapsed_AllAreTrimmed>g__RunTest|0_0() line 84
Child process:
  SixLabors.ImageSharp.Tests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13 SixLabors.ImageSharp.Tests.Memory.Allocators.UniformUnmanagedMemoryPoolTests+Trim+NonParallel Void <MultiplePoolInstances_TrimPeriodElapsed_AllAreTrimmed>g__RunTest|0_0()
```

antonfirsov:
@JimBobSquarePants the naming reflects the old buggy design that was depending on finalization order, will change it when everything else is fixed.

IMemoryOwner/MemoryGuard doesn't have a finalizer anymore, it's the associated Lifetime Guards that return things to pools:
[UniformUnmanagedMemoryPool.LifetimeGuards.cs](https://github.com/SixLabors/ImageSharp/blob/7eaa5ee1a2c3536041ff960050e15053b479511b/src/ImageSharp/Memory/Allocators/Internals/UniformUnmanagedMemoryPool.LifetimeGuards.cs)
[SharedArrayPoolBuffer.LifetimeGuard](https://github.com/SixLabors/ImageSharp/blob/7eaa5ee1a2c3536041ff960050e15053b479511b/src/ImageSharp/Memory/Allocators/Internals/SharedArrayPoolBuffer%7BT%7D.cs#L62)

The parameters of the tests are exercising different cases:
- 300 : Rented from / returned to `ArrayPool<byte>.Shared`
- 600: Rented from / returned to `UniformUnmanagedMemoryPool`
- 1200: over pool limit, guard frees the handle

JimBobSquarePants:
@antonfirsov The breaking behavior I'm seeing is the memory pool being trimmed after the first allocation because is thinks it's under high pressure (_92% of my memory appears to be in use!!_).

antonfirsov:
@JimBobSquarePants thanks that's super valuable info! I disabled these tests for local runs in 4986c52, that's the best idea I was able to come up with.

JimBobSquarePants (approved):
@anton I'm happy with this if you are. My debugging suggested that everything was working as expected and I cannot see any issues.

antonfirsov:
> I wonder whether we'd be better moving our high memory detection code into an internal utility class like the runtime does and use that as a condition instead?

I haven't seen this being done in test code, wanted to avoid complicating tests further. I don't expect them to fail unless someone touches the allocator logic.

> I'm happy with this if you are.

Yeah. let's merge this and start dogfeeding it, that should detect potential issues faster than sitting further on the PR :)

PR #1731, opened 2021-08-09
Add support for decoding tiff's with YCbCr PhotometricInterpretation

brianpopow:
### Prerequisites

- [ ] I have provided test coverage for my change (where applicable) **note: i could not provide test images for planar ycbcr with subsampling > 1. No tool I know of, could write those.**

### Description

Add support for decoding YCbCr tiff images with contiguous and planar pixel layout.


src/ImageSharp/Formats/Tiff/PhotometricInterpretation/YCbCrConverter.cs
```diff
@@ -0,0 +1,121 @@
...
+            Rgba32 rgba = this.converter.Convert(yExpanded, cbExpanded, crExpanded);
+
+            return rgba;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static byte RoundAndClampTo8Bit(float value)
+        {
+            int input = (int)MathF.Round(value);
+            return (byte)Math.Min(Math.Max(input, 0), 255);
```
gfoidl, line 71:
```suggestion
            return (byte)Math.Clamp(input, 0, 255);
```
? (and if it's available for all TFMs).

Right now codegen is pretty much equal, but in a future JIT there could be some improvements in that area.

gfoidl (commented):
A "nit-suggestion".

src/ImageSharp/Formats/Tiff/PhotometricInterpretation/YCbCrConverter.cs
brianpopow: yes makes sense, thanks

brianpopow: unfortunately not available with `net472`, using `Numerics.Clamp` now.

JimBobSquarePants: Numerics version is faster than the runtime one I think

gfoidl: > Numerics version is faster than the runtime one I think

The runtime one has an additional check for min > max which isn't needed here as the limits are constants (0 and 255). After inlineing the JIT is able to remove that min > max check, so codegen is quite similar. The difference is that the Numerics-version checks value > max first, while the runtime-version checks value < min first. So which version is faster depends on the likeness of value < min vs value > max and the branch-predictor's history (and potentially on dynamic profile guided optimization).

`Numerics.Clamp` is good to use here. Maybe just add a `Debug.Assert(min < max)` there to catch "typos" early?

JimBobSquarePants (approved):
LGTM.

Something to look at in the future would be vectorizing the planar conversion. I'm confident that we can do some optimization there.

PR #1732, opened 2021-08-10
If component id's are R, G, B in ASCII the color space should be RGB

brianpopow:
### Prerequisites

### Description

I have noticed that the colorspace is determined wrongly for some jpeg. For example:
82846232-1929ea00-9ee8-11ea-8f25-53a72b14160c

is decoded as:
wrong_color

When the frame component Id's are RGB in ASCII, the color space should be RGB not YCbCr.


JimBobSquarePants:
TIL! I did not know about this feature at all!

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -362,6 +362,12 @@ private JpegColorSpace DeduceJpegColorSpace(byte componentCount)
                     return JpegColorSpace.RGB;
                 }
 
+                // If the component Id's are R, G, B in ASCII the colorspace is RGB and not YCbCr.
+                if (components[0].Id == 82 && components[1].Id == 71 && components[2].Id == 66)
```
JimBobSquarePants, line 366: If you switch the order of these checks I believe we avoid the bounds checks for 2 out of 3 of them.

JimBobSquarePants (approved):

PR #1734, opened 2021-08-12
WIP: Adds support for Tiff's with jpeg compression

brianpopow:
### Prerequisites

### Description

This **PR** adds support for decoding and encoding of tiff's with jpeg compression. See adobe tech note 2 for details, spec pdf is in the tiff folder: ff8dfde1

For this to work, I need to add a new method `LoadTables` to the JpegDecoderCore, which only loads quantization and/or Huffman tables. Those are only stored once in  at tiff with jpeg compression, if `JPEGTables` are present in a Striped tiff (Tiff with pixel data is stored in multiple segments).

Also the JpegEncoder needs to store the data as RGB instead of YCbCr. There is now a new option for `JpegSubsample` to do that.

*note: #1732 is prerequisite for this PR*


br3aker:
Is placing rgb as option to `SubSampling` enum a good idea? Maybe it's the time to separate subsampling from color space option?

brianpopow:
> Is placing rgb as option to `SubSampling` enum a good idea? Maybe it's the time to separate subsampling from color space option?

yeah separating subsampling and color space us a good suggestion

br3aker:
Or we can just rename `SubSampling` to `ColorSpace` with following options:
1. Grayscale
2. YCbCr444
3. YCbCr420
4. Rgb

[As it was decided (?) that the next release would be a major one](https://github.com/SixLabors/ImageSharp/pull/1730#issuecomment-896063192) I assume we can break it.

YCbCr can have other subsampling stuff like 422, 411, 410 even but I think it would be better to add them as options to 'ColorSpace' instead of 2 different options one of which would work only with YCbCr.

brianpopow:
> Or we can just rename `SubSampling` to `ColorSpace` with following options:
> 
> 1. Grayscale
> 2. YCbCr444
> 3. YCbCr420
> 4. Rgb
> 
> [As it was decided (?) that the next release would be a major one](https://github.com/SixLabors/ImageSharp/pull/1730#issuecomment-896063192) I assume we can break it.
> 
> YCbCr can have other subsampling stuff like 422, 411, 410 even but I think it would be better to add them as options to 'ColorSpace' instead of 2 different options one of which would work only with YCbCr.

That really sounds like the better solution, but I am undecided, if we really should introduce a breaking change for a Tiff related feature.

br3aker:
> That really sounds like the better solution, but I am undecided, if we really should introduce a breaking change for a Tiff related feature.

Different color spaces for jpeg were in plans for quite a while now https://github.com/SixLabors/ImageSharp/issues/808 so it's good I think :)

This can be introduced as a new property in non-breaking manner:
1. Existing `SubSample` with Grayscale/YCbCr444/YCbCr420
2. New `ColorSpace` with Grayscale/YCbCr444/YCbCr420/Rgb

Internal code would use new `ColorSpace` and `SubSample` would map to it. Marking `Subsample` obsolete and removing it somewhen in the next 2.X minor release.

JimBobSquarePants:
We don’t need to worry about breaking changes with Tiff at all. We haven’t shipped anything other than dev builds with it (and won’t) until 2.0

br3aker:
@JimBobSquarePants renaming/new `ColorSpace` property would alter public `JpegEncoder` interface, tiff just uses it internally via this commit.

JimBobSquarePants:
@br3aker  I don't follow. We already have `IJpegOptions.JpegColorType`. Adding the new types as an when we introduce them would be fine.

br3aker:
@JimBobSquarePants my latest proposal was to remove the `SubSampling` parameter at all as it's only used for 1/3 of the existing colors types (grayscale, ycbcr and rgb). As for my understanding `SubSampling` is more of a legacy thingy which was introduced when no other color spaces were implemented.

IMO it's more understandable to have a single parameter `ColorSpace` or existing `JpegColorType` with following options:
1. Grayscale/Luminance
2. Rgb
3. YCbCr444
4. YCbCr420
5. (not implemented) Cmyk
6. (not implemented) YccK
7. (not implemented) YCbCr422
8. (not implemented) YCbCr411
9. Whatever else jpeg standard supports

While not removing parameter (marking it obsolete though) itself so it won't break anyone.

JimBobSquarePants:
I guess.... They're separate concerns but combining them makes sense in this instance. 

>While not removing parameter (marking it obsolete though) itself so it won't break anyone.

We'd just make the breaking change. It's 2.0, it's also free software - People can recompile.

br3aker:
Making breaking changes is not fun, I understand. But I believe it's for good here, possibility of Rgb with Ratio420 would be 'strange', to say the least :)

JimBobSquarePants:
Do you have any docs talking about subsampling with Rgb? I can't find anything to say anything other than YCbCr does it.

br3aker:
> Do you have any docs talking about subsampling with Rgb? I can't find anything to say anything other than YCbCr does it.

Oh no, no, sorry for inconvenience. I meant right now it's possible to write something like:

```csharp
var encoder1 = new JpegEncoder 
{
    JpegColorType = Luminance,
    JpegSubsample = Ratio420
};

// With this commit adding Rgb as option
var encoder2 = new JpegEncoder 
{
    JpegColorType = Rgb,
    JpegSubsample = Ratio420
};
```

Both of which makes no sense as chroma sub-sampling only works with YCbCr as it clearly divides luminance from chrominance. Providing only color type would solve these.

JimBobSquarePants:
Ah right yeah, that all makes sense and I concur. Let’s fix it properly

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbForwardConverter{TPixel}.cs
```diff
@@ -0,0 +1,114 @@
...
+            ref Block8x8F redBlock = ref this.R;
+            ref Block8x8F greenBlock = ref this.G;
+            ref Block8x8F blueBlock = ref this.B;
+
+            CopyToBlock(this.rgbSpan, ref redBlock, ref greenBlock, ref blueBlock);
+        }
+
+        private static void CopyToBlock(Span<Rgb24> rgbSpan, ref Block8x8F redBlock, ref Block8x8F greenBlock, ref Block8x8F blueBlock)
+        {
+            ref Rgb24 rgbStart = ref rgbSpan[0];
```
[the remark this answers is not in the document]
brianpopow: @gfoidl thanks for your suggestions

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -141,12 +145,27 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream, CancellationToken
...
-        /// <param name="quant">The quantization table to copy data from</param>
+        /// <returns>The component Ids.</returns>
+        private byte[] GetComponentIds()
+        {
+            if (this.subsample == JpegSubsample.Rgb)
+            {
+                return new byte[] { 82, 71, 66 };
+            }
+
+            return new byte[] { 1, 2, 3 };
```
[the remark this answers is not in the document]
brianpopow: changed with 96edc48f and 77cdbbb8

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -34,19 +34,24 @@ internal sealed unsafe class JpegEncoderCore : IImageEncoderInternals
         private readonly byte[] buffer = new byte[20];
 
         /// <summary>
-        /// Gets or sets the subsampling method to use.
+        /// The default component id's.
         /// </summary>
-        private JpegSubsample? subsample;
+        private static readonly byte[] DefaultComponentIds = { 1, 2, 3 };
```
gfoidl, line 39: That way it's a "constant", but it still needs some effort at runtime to use it. See e.g. https://vcsjones.dev/csharp-readonly-span-bytes-static/ for a write-up for this.

These component-id constants are used only at one single place, so I'll write it as suggested in https://github.com/SixLabors/ImageSharp/pull/1734#discussion_r687906455. Yields readable code with good locality, and has best codegen.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -545,8 +571,9 @@ private void WriteStartOfFrame(int width, int height, int componentCount)
             for (int i = 0; i < componentCount; i++)
             {
                 int i3 = 3 * i;
-                this.buffer[i3 + 6] = (byte)(i + 1);
 
+                // Component ID.
+                this.buffer[i3 + 6] = componentIds[i];
```
gfoidl, line 576: To elide some bound checks slice `buffer` to a Span so that indexing can start at `0` (and write in reverse order, so start at `[2]`, `[1]`, then `[0]`)?

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -557,19 +584,10 @@ private void WriteStartOfFrame(int width, int height, int componentCount)
...
+        /// <param name="componentIds">The componentId's.</param>
+        private void WriteStartOfScan(int componentCount, ReadOnlySpan<byte> componentIds)
         {
-            Span<byte> componentId = stackalloc byte[]
-            {
-                0x01,
-                0x02,
-                0x03
-            };
             Span<byte> huffmanId = stackalloc byte[]
```
gfoidl, line 591: This could use the same ROSpan<byte> compiler optimization.
```suggestion
            // This uses a C#'s compiler optimization that refers to the static data segment of the assembly,
            // and doesn't incur any allocation at all.
            ReadOnlySpan<byte> huffmanId = new byte[]
```

src/ImageSharp/Formats/Jpeg/Components/Encoder/HuffmanScanEncoder.cs
```diff
@@ -234,6 +231,65 @@ public void EncodeGrayscale<TPixel>(Image<TPixel> pixels, ref Block8x8F luminanc
...
+                for (int x = 0; x < pixels.Width; x += 8)
+                {
+                    pixelConverter.Convert(x, y, ref currentRows);
+
+                    prevDCY = this.WriteBlock(
+                        QuantIndex.Luminance,
+                        prevDCY,
+                        ref pixelConverter.R,
+                        ref luminanceQuantTable,
+                        ref unzig);
```
br3aker, line 272: Not sure about actual specs for rgb encoding but I guess we should use same quantization table for all rgb channels. If this is a thing - we should also fix quantization table index for all channels set to either 0 (luminance) or 1 (chrominance).

brianpopow: yeah I think you are right. While the image looks correct when decoding with ImageSharp, it looks like other tools like IrfanView interpret the data still as YCbCr. Probably because there are more then one quantization table present.

output

br3aker: That's strange, decoders should not assume color space via qtable count. Primary 'tag' for rgb should be the app14 marker which you've already implemented and it is working correctly.

What's even funnier - Photoshop opens it correctly.

brianpopow: > That's strange, decoders should not assume color space via qtable count. Primary 'tag' for rgb should be the app14 marker which you've already implemented and it is working correctly.

This was just a very wild guess from my side. ImageMagick does also read the image as above and im pretty sure it can handle jpeg with RGB colorspace. If i encode an image with `libjpeg` with RGB colorspace, imagemagick can read it correctly. This means im doing something wrong here, but i can not see at the moment what that is. Any idea's are welcome, I am a bit stuck at the moment.

brianpopow: btw. the tiff images which use jpeg compression look correct.

br3aker: I'll take a closer look at the code tomorrow. If photoshop says file is not corrupted - we are somewhat bytes away from cracking it so it shouldn't take long.

brianpopow: thank you @br3aker, I appreciate your help.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
JimBobSquarePants: BT @gfoidl Love these optimization tips, keep them coming! 👍

br3aker:
@brianpopow did a bit of testing with rgb stuff and actually solved it:

**First of all, actual solution**: `APP0` marker is conflicting with `APP14` marker. Simply not writing it for rgb encoded data solves this issue. 

Explanation (?): RGB is not JFIF compatible 'format' - there's no rgb mentions in the itu specs. Standard JPEG with YCbCr is a JPEG FIF file - it must contain `SOI -> APP0 -> etc` marker sequence. RGB (and I suppose Cmyk/YccK also) are not **JPEG FIF** compatible files, they are **JPEG IF** compatible files - they must contain `SOI -> etc` marker sequence. 

As far as I know rgb jpegs are more of a Adobe experiment over the actual jpeg format so it explains a lot. Some software (if not most) uses APP0 to determine if image is a standard jpeg and I would dare to say it is kind of wrong - while it's technically correct as there's no format specification for anything `APPX` related, `APP0` check limits actual format as only `APP0` marker can contain image thumbnail.

**tl;dr** drop APP0 marker, we should separate `private void WriteApplicationHeader(ImageMetadata meta) {}` call into:
`WriteStartOfImage` - Writes SOI marker
`WriteJfifApplicationHeader` - Writes APP0 **only for YCbCr encoding**

Second interesting thing I found - libjpeg-turbo (and old libjpeg I guess) uses standard jpeg luminance quantization table and luminance huffman table for all rgb channels. While writing different tables/huffman for each channel is okay as long as correct indices for them are provided - we already do that so it's fine. It' still better to use same quantization table for each channel though :P

brianpopow:
> @brianpopow did a bit of testing with rgb stuff and actually solved it:
> 
> **First of all, actual solution**: `APP0` marker is conflicting with `APP14` marker. Simply not writing it for rgb encoded data solves this issue. 
> ...

@br3aker thank you a lot for investigating this issue, its very helpful. I will try to change the encoder as you suggested.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
brianpopow: changed with 49d86e1b

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
brianpopow: changed with 66a9d1ae

src/ImageSharp/Formats/Jpeg/Components/Encoder/RgbForwardConverter{TPixel}.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            for (int i = 0; i < Block8x8F.Size; i++)
+            {
+                Rgb24 c = Unsafe.Add(ref rgbStart, i);
+
+                redBlock[i] = c.R;
+                greenBlock[i] = c.G;
+                blueBlock[i] = c.B;
```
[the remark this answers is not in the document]
JimBobSquarePants: I think we can support `nint`, we just need to change our build process. For reasons I cannot recall we're building the code with explicit target frameworks which I can't see a reason to do since nobody ever does this. We only need to test with individual frameworks and build things normally.

PR #1742, opened 2021-08-17
Throw NotSupportedException for arithmetic coding and lossless jpeg's

brianpopow:
### Prerequisites

### Description

Currently the jpeg decoder will fail with an object null exception with jpegs which are not supported yet.
This **PR** changes the jpeg decoder to throw a NotSupportedException instead.


src/ImageSharp/Formats/Jpeg/JpegConstants.cs
```diff
@@ -173,6 +178,46 @@ internal static class Markers
             /// </summary>
             public const byte SOF2 = 0xC2;
 
+            /// <summary>
+            /// Start of Frame marker, non differential lossless, Huffman coding.
+            /// </summary>
+            public const byte SOF3 = 0xC3;
+
```
brianpopow, line 185: note: there are also 0xC5 and 0xC6 StartOfFrame markers for Differential progressive DCT and Differential sequential DCT. I could not find any real life examples of those, so I was unsure if I should add those here too.

src/ImageSharp/Formats/Jpeg/JpegThrowHelper.cs
```diff
@@ -8,6 +8,13 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
 {
     internal static class JpegThrowHelper
     {
+        /// <summary>
+        /// Cold path optimization for throwing <see cref="NotSupportedException"/>'s.
+        /// </summary>
+        /// <param name="errorMessage">The error message for the exception.</param>
+        [MethodImpl(InliningOptions.ColdPath)]
+        public static void ThrowNotSupportedException(string errorMessage) => throw new NotSupportedException(errorMessage);
```
antonfirsov, line 16: Do we throw `ImageFormatException` or `NotSupportedException` in other cases we don't support decoding a valid input?

Ideally, we need to make sure to document all exceptions we throw on `Image.Load` overloads, although I'm less strict on this since I know that the BCL is also full of undocumented exceptions :)

brianpopow: we throw `ImageFormatException` in cases when the image data is not as expected, e.g. not as the specification suggest.
Those cases here the jpeg files are valid, so i think `NotSupportedException` is the correct Exception to throw.

br3aker: +1 for `NotSupportedException` because corrupted image and not supported image are very different scenarios :)

src/ImageSharp/Formats/Jpeg/JpegConstants.cs
br3aker: Afaik arithmetic coding is not very common (at least for general internet use case) as it's harder to compute and storage gain is not really better than DCT one. Adding this just for 'why-not-ness' won't break anything as it's internal or kill the performance. So, why not?

tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.cs
```diff
@@ -174,6 +174,19 @@ public async Task Identify_IsCancellable()
             await Assert.ThrowsAsync<TaskCanceledException>(async () => await Image.IdentifyAsync(config, "someFakeFile", cts.Token));
         }
 
+        [Theory]
+        [WithFile(TestImages.Jpeg.Baseline.ArithmeticCoding, PixelTypes.Rgba32)]
+        [WithFile(TestImages.Jpeg.Baseline.ArithmeticCodingProgressive, PixelTypes.Rgba32)]
```
br3aker, line 179: Progressive image should be placed to `Jpeg/Progressive` folder.

src/ImageSharp/Formats/Jpeg/JpegConstants.cs
JimBobSquarePants: If anyone gets bored they could copy implementations of the missing decoder functionality from here. 

https://github.com/yigolden/JpegLibrary

They copied a bunch of stuff from us so all fair.

src/ImageSharp/Formats/Jpeg/JpegThrowHelper.cs
antonfirsov: Great to see the docs fixed in `Image.FromFile.cs` and `Image.FromBytes.cs` ! 👍

I would also change the docs in `Image.FromStream`:

```C#
/// <exception cref="NotSupportedException">The stream is not readable. -or- The image format is not supported.</exception>
```

JimBobSquarePants (approved):
Great stuff!

PR #1745, opened 2021-08-19
Enable .NET 6 Preview Builds

JimBobSquarePants:
### Prerequisites

### Description

<strike>Just use the environment compiler to build things. When consumers build their applications they're already consuming a prebuilt binary. We only need to test with individual frameworks.</strike>

Added .NET 6 Preview builds to the matrix and fixed up the solution to handle that scenario. Also fixed the langversion to C#9

This allows us to use features like `nint` and also build/test against .NET 6 previews which we sorely need to do.

### DIFF
--- a/ci-build.ps1
+++ b/ci-build.ps1
@@ -1,11 +1,6 @@
-param(
-  [Parameter(Mandatory = $true, Position = 0)]
-  [string]$targetFramework
-)
-
 dotnet clean -c Release
 
 $repositoryUrl = "https://github.com/$env:GITHUB_REPOSITORY"
 
 # Building for a specific framework.
-dotnet build -c Release -f $targetFramework /p:RepositoryUrl=$repositoryUrl
+dotnet build -c Release /p:RepositoryUrl=$repositoryUrl
--- a/ci-build.ps1
+++ b/ci-build.ps1
@@ -1,6 +1,11 @@
+param(
+  [Parameter(Mandatory = $true, Position = 0)]
+  [string]$targetFramework
+)
+
 dotnet clean -c Release
 
 $repositoryUrl = "https://github.com/$env:GITHUB_REPOSITORY"
 
 # Building for a specific framework.
-dotnet build -c Release /p:RepositoryUrl=$repositoryUrl
+dotnet build -c Release -f $targetFramework /p:RepositoryUrl=$repositoryUrl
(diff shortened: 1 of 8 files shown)


ci-build.ps1
```diff
@@ -1,11 +1,6 @@
...
-  [Parameter(Mandatory = $true, Position = 0)]
-  [string]$targetFramework
-)
-
 dotnet clean -c Release
 
 $repositoryUrl = "https://github.com/$env:GITHUB_REPOSITORY"
 
 # Building for a specific framework.
-dotnet build -c Release -f $targetFramework /p:RepositoryUrl=$repositoryUrl
```
antonfirsov, line 11: This will build every target making the builds longer, won't this?

Does the target framework impact the compiler version, isn't it purely an SDK thing?

JimBobSquarePants: Yep so I had this badly wrong. My understanding was that specifying the target framework moniker causes the underlying MsBuild call to look for and use that specific SDK to compile against. That's obviously not true. 

I've fixed up the project now so we can build against the preview in CI and everything else is unaffected. I got the langversion working properly also so we can use the C#9 features.

JimBobSquarePants:
Think I might just flatten this one when I merge it 😝

iamcarbon:
We should consider dropping netcoreapp2.1 for the 2.0 release. It's going out of support this month.

JimBobSquarePants:
@iamcarbon It’s only V2.0 because we have to introduce the breaking changes in #1730

My plan following this release is to immediately start work on V3.0 which will reduce the target frameworks to a minimum of .NET Core 3.1

As such this will act as the last stable release for legacy platforms.

src/ImageSharp/Processing/Processors/Normalization/AdaptiveHistogramEqualizationSlidingWindowProcessor{TPixel}.cs
```diff
@@ -257,7 +257,7 @@ protected override void OnFrameApply(ImageFrame<TPixel> source)
         [MethodImpl(InliningOptions.ShortMethod)]
         private void AddPixelsToHistogram(ref Vector4 greyValuesBase, ref int histogramBase, int luminanceLevels, int length)
         {
-            for (int idx = 0; idx < length; idx++)
+            for (nint idx = 0; idx < length; idx++)
```
brianpopow, line 260: @JimBobSquarePants could you explain a bit what the advantages of using `nint` is? This is kind of new to me.

JimBobSquarePants: Basically better codegen. 

https://github.com/SixLabors/ImageSharp/pull/1734#discussion_r687902047

brianpopow (approved):
looks good

PR #1760, opened 2021-09-11
Support BigTiff decoding 

IldarKhayrutdinov:
### Prerequisites

### Description

Support decoding BigTIFF file format: https://www.awaresystems.be/imaging/tiff/bigtiff.html

### Todo
- [x] ~~Add more test files~~ Add unti-tests for uncovered Exif tags
- [x] Revise reading 64-bit values of ExifReader
- [x] Revise work with `StripOffsets, StripByteCounts` tags 
- [x] Revise work with `SubIFD, GPSIfd` tags


src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
```diff
@@ -349,6 +394,95 @@ private void ReadValue(List<IExifValue> values)
...
+                ulong newOffset = this.ConvertToUInt64(this.offsetBuffer8);
+                if (newOffset > ulong.MaxValue || (newOffset + size) > (ulong)this.data.Length)
+                {
+                    this.AddInvalidTag(new UnkownExifTag(tag));
+                    return;
+                }
+
+                this.RegisterExtLoader(newOffset, () =>
+                {
+                    byte[] dataBuffer = new byte[size];
```
gfoidl, line 470: Can the allocation for dataBuffer be avoided?
As well as the closure for size?

For dataBuffer stack-alloc or use a buffer from a pool?
To avoid the closure `RegisterExtLoader` should have a generic state argument, which allows to pass in the state.

But the question is: This adds some more complexity, so is it worth?

IldarKhayrutdinov: @gfoidl Thanks for your thoughtful review. This code confuses me too and definitely needs to be improved. BufferPool would be more suitable for this code, but so far it isn't used in Exif. Usually Exif contains a little data, so this is not critical. Closing will not happen and the state will not be damaged either, this is an internal class and the extLoaders are guaranteed called before the stream closing. But from the outside perspective the code really looks questionable, I will try to improve this code.

src/ImageSharp/Formats/Tiff/Ifd/DirectoryReader.cs
```diff
@@ -56,13 +63,13 @@ private static ByteOrder ReadByteOrder(Stream stream)
             throw TiffThrowHelper.ThrowInvalidHeader();
         }
 
-        private IEnumerable<ExifProfile> ReadIfds()
+        private IEnumerable<ExifProfile> ReadIfds(bool isBigTiff)
         {
             var readers = new List<EntryReader>();
```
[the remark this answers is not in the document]
IldarKhayrutdinov: no, frames count isn't known in advance in Tiff format

src/ImageSharp/Formats/Tiff/Ifd/DirectoryReader.cs
```diff
@@ -49,8 +49,8 @@ public IEnumerable<ExifProfile> Read()
 
         private static ByteOrder ReadByteOrder(Stream stream)
         {
-            byte[] headerBytes = new byte[2];
-            stream.Read(headerBytes, 0, 2);
+            Span<byte> headerBytes = stackalloc byte[2];
```
gfoidl, line 48: @JimBobSquarePants we use a lot of `stackalloc`s for local buffers. By default the JIT must emit code to zero the buffer contents. `[SkipLocalsInit]` tells the JIT that this zeroing shouldn't be done, hence the produced machinecode is more compact, and yields better perf. This attribute can be applied to individual methods, types, modules, ...
I'd like to enable it globally (i.e. per module), so that all occurences of stackalloc will benefit from this.

Is there any policy for this? In regards of scratch buffers not being zeroed?

For older targets, where this attribute isn't available, one can create a dummy attribute with same typename, to make it work.

JimBobSquarePants: We’ve been very careful to ensure we don’t use what we don’t write so we should be fine to enable it.

src/ImageSharp/Metadata/Profiles/Exif/Values/ExifNumberArray.cs
```diff
@@ -54,13 +54,25 @@ public override bool TrySetValue(object value)
                 case ushort val:
                     return this.SetSingle(val);
                 case int[] array:
+                {
+                    // workaround for inconsistent covariance of value-typed arrays
+                    if (value.GetType().Equals(typeof(uint[])))
```
gfoidl, line 59:
```suggestion
                    if (value.GetType() == typeof(uint[]))
```
Produces [less code](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunGBLAHYYGAFSgBPAMowMANWwAbAK4wAFBGAArGGGEA3RSoCUDLj07UeVhrgDufDGAAWDVQeUwjZ6x28+eYHgwDIL6iiBMAOwM5ADcfv4MgbjBoQDaALoM2FBQ2OIgCYnmRcXWfABmru4qAHQA4jIi4gAOaka1AKIAjkqKuKoYrTAQFapK6RlG06VlJZZzi8TRpPELiz4AvrOLO3PLDHRrGzzb68UAJjAV2EoKWHs+B3Bxs2dW7zx+9CFCohLSOSGGDkdRaHRhDwmPwWRJ2BzOarArznKywubJVJ/dwRA6vVH+TG/DCZbK5fKFAlldEnKyVJEeBpNYaqEwAXjZDCGbVG40mMypGxptKeK2OItOj2KUsSByOMslgqsVxudweSv8z3xiU+3DOmyAA). I believe the JIT is able to play some optimizations in that case.

src/ImageSharp/Metadata/Profiles/Exif/Values/ExifNumberArray.cs
```diff
@@ -54,13 +54,25 @@ public override bool TrySetValue(object value)
...
+                    }
+
                     return this.SetArray(array);
-                case uint[] array:
-                    return this.SetArray(array);
+                }
+
                 case short[] array:
+                {
+                    if (value.GetType().Equals(typeof(ushort[])))
```
gfoidl, line 69:
```suggestion
                    if (value.GetType() == typeof(ushort[]))
```

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
```diff
@@ -119,7 +110,30 @@ internal abstract class BaseExifReader
...
+            foreach ((ulong offset, ExifDataType dataType, ulong numberOfComponents, ExifValue exif) in this.BigValues)
+            {
+                ulong size = numberOfComponents * ExifDataTypes.GetSize(dataType);
+                if (size > maxSize)
+                {
+                    maxSize = size;
+                }
+            }
+
+            Span<byte> buf = new byte[maxSize];
```
gfoidl, line 127:
```suggestion
            byte[] bufferToReturnToPool = null;
            Span<byte> buffer = (maxSize <= 256    // better to use a defined constant for this
                ? stackalloc byte[256]
                : bufferToReturnToPool = ArrayPool<byte>.Shared.Rent(maxSize)
            ).Slice(maxSize);

            // ...

            if (bufferToReturnToPool is not null)
            {
                ArrayPool<byte>.Shared.Return(bufferToReturnToPool);
            }
```
To avoid the allocation, and try to use stack allocated buffer.

Or can we assume that `maxSize <= 256` will be true for all inputs? If so, the code can be simplified to just stack allocated and then slice the buffer*.

\* `stackalloc byte[ConstantWithPowerOf2]` is cheaper than `stackalloc byte[variable]`.

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
```diff
@@ -136,22 +150,60 @@ protected void ReadValues(List<IExifValue> values, uint offset)
             this.Seek(offset);
             int count = this.ReadUInt16();
 
+            Span<byte> offsetBuffer = new byte[4];
```
gfoidl, line 153:
```suggestion
            Span<byte> offsetBuffer = stackalloc byte[4];
```

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
```diff
@@ -136,22 +150,60 @@ protected void ReadValues(List<IExifValue> values, uint offset)
...
+
+        protected void ReadValues64(List<IExifValue> values, ulong offset)
+        {
+            DebugGuard.MustBeLessThanOrEqualTo(offset, (ulong)this.data.Length, "By spec UInt64.MaxValue is supported, but .Net Stream.Length can Int64.MaxValue.");
 
-            if (this.gpsOffset != 0)
+            this.Seek(offset);
+            ulong count = this.ReadUInt64();
+
+            Span<byte> offsetBuffer = new byte[8];
```
gfoidl, line 180:
```suggestion
            Span<byte> offsetBuffer = stackalloc byte[8];
```

brianpopow:
@IldarKhayrutdinov hey Ildar, nice to see you back with more good tiff stuff! You have still "add more tests" in your TODO list. I have seen there are already a bunch of big tiff tests. What kind of tests do you think are missing? Maybe I can help out with that.

src/ImageSharp/Metadata/Profiles/Exif/Values/ExifLong8Array.cs
```diff
@@ -0,0 +1,173 @@
...
+            get
+            {
+                if (this.Value is null)
+                {
+                    return ExifDataType.Long;
+                }
+
+                for (int i = 0; i < this.Value.Length; i++)
+                {
+                    if (this.Value[i] > uint.MaxValue)
```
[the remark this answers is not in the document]
IldarKhayrutdinov: Thanks to @gfoidl! I missed such simple optimizations (copied from similar classes, also need to optimize)

src/ImageSharp/Metadata/Profiles/Exif/Values/ExifLong8Array.cs
```diff
@@ -0,0 +1,173 @@
...
+            return true;
+        }
+
+        private bool SetArray(long[] values)
+        {
+            var numbers = new ulong[values.Length];
+            for (int i = 0; i < values.Length; i++)
+            {
+                numbers[i] = (ulong)values[i];
+            }
```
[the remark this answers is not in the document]
IldarKhayrutdinov: basically there will a small arrays

IldarKhayrutdinov:
> @IldarKhayrutdinov hey Ildar, nice to see you back with more good tiff stuff! You have still "add more tests" in your TODO list. I have seen there are already a bunch of big tiff tests. What kind of tests do you think are missing? Maybe I can help out with that.

@brianpopow I'm happy to contribute to the project! Yes, there is not enough a coverage, so I wanted more test files.

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
IldarKhayrutdinov: @JimBobSquarePants what should we do in such cases?
`MemoryAllocator` dragging to here is probably impractical.

src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
```diff
@@ -211,8 +218,9 @@ private ImageFrame<TPixel> DecodeFrame<TPixel>(ExifProfile tags, CancellationTok
             var frame = new ImageFrame<TPixel>(this.Configuration, width, height, imageFrameMetaData);
 
             int rowsPerStrip = tags.GetValue(ExifTag.RowsPerStrip) != null ? (int)tags.GetValue(ExifTag.RowsPerStrip).Value : TiffConstants.RowsPerStripInfinity;
-            Number[] stripOffsets = tags.GetValue(ExifTag.StripOffsets)?.Value;
-            Number[] stripByteCounts = tags.GetValue(ExifTag.StripByteCounts)?.Value;
+
+            var stripOffsets = (Array)tags.GetValueInternal(ExifTag.StripOffsets)?.GetValue();
```
IldarKhayrutdinov, line 222: `stripOffsets` can be a `Number` array or `ulong` array

src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
```diff
@@ -413,7 +423,14 @@ private void DecodeStripsChunky<TPixel>(ImageFrame<TPixel> frame, int rowsPerStr
                     break;
                 }
 
-                decompressor.Decompress(this.inputStream, (uint)stripOffsets[stripIndex], (uint)stripByteCounts[stripIndex], stripHeight, stripBufferSpan);
+                ulong offset = stripOffsets.GetValue(stripIndex) switch { ulong val => val, Number val => (uint)val };
```
IldarKhayrutdinov, line 426: This conversion will be performed once for each strip, performance will not be affected. Not beautiful, but so far this is the best solution for converting `Number` and `ulong` arrays to same variable (without allocating additional memory, these arrays can be large).

JimBobSquarePants: @IldarKhayrutdinov Is this comment still relevant? The UI is marked as outdated.

I have no issue extending the type if the extensions can be kept internal?

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
JimBobSquarePants: Do we have any numbers for how often `maxSize` is over 256?

I would always fallback to the MemoryAllocator rather than the `ArrayPool`.

IldarKhayrutdinov: A few, mostly it's relevant for `StripOffsets`, `StripByteCounts`. Examples, for ~4Mb file and 8Kb (recommended) strip size, required about 512 strips * 4 (long size) = 2Kb bytes for stripOffsets tag value, 16Mb - 8Kb, extreme 1Gb (BigTiff can greater than 4Gb) - 512Kb

src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
IldarKhayrutdinov: Yes relevant, there was just text formatting. 
I seem to be incorrectly expressed. I meant about changing struct Number, which is a public:
 https://github.com/IldarKhayrutdinov/ImageSharp/blob/978757fb51a3cdc8bbd0a2d8d3df25988ff0e7c4/src/ImageSharp/Primitives/Number.cs#L16-L23

instead
```
private readonly int signedValue;
private readonly uint unsignedValue;
```

required
```
private readonly long signedValue;
private readonly ulong unsignedValue;
```

and add cast operators

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
JimBobSquarePants: If I'm reading this correctly we should just use the MemoryAllocator. We'll be switching out for the new one soon enough anyway.

src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
JimBobSquarePants: Since our casting is explicit I would think the expansion is ok.

IldarKhayrutdinov:
@JimBobSquarePants  @brianpopow Please review :-) I added unit tests for uncovered tags (it wasn't possible to find BigTiff files that covering all tags).

src/ImageSharp/Formats/Tiff/TiffImageFormatDetector.cs
```diff
@@ -26,9 +34,49 @@ public IImageFormat DetectFormat(ReadOnlySpan<byte> header)
...
         {
-            return header.Length >= this.HeaderSize &&
-                   ((header[0] == 0x49 && header[1] == 0x49 && header[2] == 0x2A && header[3] == 0x00) || // Little-endian
-                    (header[0] == 0x4D && header[1] == 0x4D && header[2] == 0x00 && header[3] == 0x2A)); // Big-endian
+            if (header.Length >= this.HeaderSize)
+            {
+                if (header[0] == 0x49 && header[1] == 0x49)
+                {
+                    // Little-endian
+                    if (this.isBigTiff is false)
```
JimBobSquarePants, line 42: That's an odd way to say `!this.isBigTiff`

src/ImageSharp/Formats/Tiff/TiffConfigurationModule.cs
```diff
@@ -13,7 +13,9 @@ public void Configure(Configuration configuration)
         {
             configuration.ImageFormatsManager.SetEncoder(TiffFormat.Instance, new TiffEncoder());
             configuration.ImageFormatsManager.SetDecoder(TiffFormat.Instance, new TiffDecoder());
-            configuration.ImageFormatsManager.AddImageFormatDetector(new TiffImageFormatDetector());
+            configuration.ImageFormatsManager.AddImageFormatDetector(new TiffImageFormatDetector(false));
```
JimBobSquarePants, line 16: Is there a way to do this without multiple `TiffImageFormatDetector` instances? This feels like an abuse of the API.

src/ImageSharp/Formats/Tiff/TiffImageFormatDetector.cs
```diff
@@ -10,8 +10,16 @@ namespace SixLabors.ImageSharp.Formats.Tiff
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TiffImageFormatDetector"/> class.
+        /// </summary>
+        /// <param name="isBigTiff">if set to <c>true</c> is BigTiff.</param>
+        public TiffImageFormatDetector(bool isBigTiff) => this.isBigTiff = isBigTiff;
+
         /// <inheritdoc/>
-        public int HeaderSize => 4;
+        public int HeaderSize => this.isBigTiff ? 8 : 4;
```
JimBobSquarePants, line 22: Why not always use 8?

src/ImageSharp/Formats/Tiff/TiffMetadata.cs
```diff
@@ -26,6 +26,14 @@ public TiffMetadata()
         /// </summary>
         public ByteOrder ByteOrder { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether this instance is BigTiff format.
+        /// </summary>
+        /// <value>
+        ///   <c>true</c> if this instance BigTiff format; otherwise, <c>false</c>.
+        /// </value>
+        public bool IsBigTiff { get; set; }
```
JimBobSquarePants, line 35: I don't think we should do it like this. I'm always wary of booleans as they limit future expansion. An enum would allow us to expand into future things like GeoTIFF

```c#
public enum TiffFormatType
{
    Default,
    BigTIFF
}
```

src/ImageSharp/Formats/Tiff/TiffImageFormatDetector.cs
JimBobSquarePants: I would actually use a single instance of the detector and create two separate private methods called from here and test against both.

IldarKhayrutdinov: done

src/ImageSharp/Formats/Tiff/TiffConfigurationModule.cs
IldarKhayrutdinov: really, it's easier

src/ImageSharp/Formats/Tiff/TiffImageFormatDetector.cs
IldarKhayrutdinov: done

src/ImageSharp/Formats/Tiff/TiffMetadata.cs
IldarKhayrutdinov: yeah, done

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
```diff
@@ -136,25 +182,59 @@ protected void ReadValues(List<IExifValue> values, uint offset)
...
+        {
+            this.Seek(tag.Offset);
+            if (this.TryReadSpan(buffer))
+            {
+                object value = this.ConvertValue(tag.DataType, buffer, tag.NumberOfComponents > 1 || tag.Exif.IsArray);
+                this.Add(values, tag.Exif, value);
+            }
+        }
+
+        ////protected void ReadSubIfd64(List<IExifValue> values)
```
JimBobSquarePants, line 227: Let's delete this if we do not need it.

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
```diff
@@ -330,21 +426,94 @@ private void ReadValue(List<IExifValue> values)
...
+                case ExifTagValue.StripByteCounts:
+                    exifValue = new ExifLong8Array(ExifTagValue.StripByteCounts);
+                    break;
+                case ExifTagValue.TileOffsets:
+                    exifValue = new ExifLong8Array(ExifTagValue.TileOffsets);
+                    break;
+                case ExifTagValue.TileByteCounts:
+                    exifValue = new ExifLong8Array(ExifTagValue.TileByteCounts);
+                    break;
+                ////case ExifTagValue.SubIFDOffset:
```
JimBobSquarePants, line 482: Same as with other commented code.

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
```diff
@@ -368,11 +537,19 @@ private void Add(IList<IExifValue> values, IExifValue exif, object value)
...
+                this.AddSubIfd(value);
             }
             else if (exif.Tag == ExifTag.GPSIFDOffset)
             {
-                this.gpsOffset = (uint)value;
+                this.AddSubIfd(value);
+            }
+            else if (exif.Tag == ExifTag.SubIFDs)
+            {
+                //// didn't find any useful data in SubIFDs
```
JimBobSquarePants, line 548: As above

tests/ImageSharp.Tests/Formats/Tiff/BigTiffMetadataTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        [Fact]
+        public void NotCoveredTags64bit()
+        {
+            var testTags = new Dictionary<ExifTag, (ExifDataType DataType, object Value)>
+            {
+                { new ExifTag<ulong>((ExifTagValue)0xdd11), (ExifDataType.Long8, ulong.MaxValue) },
+                { new ExifTag<long>((ExifTagValue)0xdd12), (ExifDataType.SignedLong8, long.MaxValue) },
+                //// WriteIfdTags64Bit: arrays aren't support
```
JimBobSquarePants, line 169: How come? In the spec or our code?

tests/Images/Input/Tiff/BigTiff/BigTIFFSamples.html
```diff
@@ -0,0 +1,239 @@
+<html>
```
JimBobSquarePants, line 1: I'd much prefer Markdown if this file is absolutely required. Makes it readable in the repository then.

JimBobSquarePants:
Looking good @IldarKhayrutdinov Just a few minor bits of cleanup 👍

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
IldarKhayrutdinov: done

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
IldarKhayrutdinov: done

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
IldarKhayrutdinov: done

tests/ImageSharp.Tests/Formats/Tiff/BigTiffMetadataTests.cs
IldarKhayrutdinov: Our code. No test files for coverage (64-bit arrays).

tests/Images/Input/Tiff/BigTiff/BigTIFFSamples.html
IldarKhayrutdinov: Done. Converted to MD.

src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
IldarKhayrutdinov: 64-bit tags will rarely be used, in Exif it's not used at all, I think it would be better not to change the Number type

JimBobSquarePants (approved):
Really nice bit of work @IldarKhayrutdinov 👍 Thanks again for your efforts!

PR #1398, opened 2020-10-21
Add SimdUtils.HwIntrinsics

JimBobSquarePants:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description
Replaces the `SimdUtilsAxv2Intrinsics` with `SimdUtils.HwIntrinsics` containing Avx2 and Sse2 implementations of `ByteToNormalizedFloatReduce` and `NormalizedFloatToByteSaturateReduce`. This gives us fully accelerated conversion in both directions.

# Benchmarks

_Warmup and iteration count is probably a little low for these as demonstrated by the differences between `PixelOperations_Specialized` and `HwIntrinsics`. One calls the other!_

**ToVector4_Rgba32**

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19041.572 (2004/?/20H1)
Intel Core i7-8650U CPU 1.90GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.100-rc.2.20479.15
  [Host]     : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  Job-EFOYPA : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT

Runtime=.NET Core 3.1  IterationCount=3  LaunchCount=1
WarmupCount=3
```

|                      Method | Count |        Mean |         Error |     StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------------- |------ |------------:|--------------:|-----------:|------:|--------:|-------:|------:|------:|----------:|
|       FallbackIntrinsics128 |    64 |   153.00 ns |      9.491 ns |   0.520 ns |  1.34 |    0.01 |      - |     - |     - |         - |
|        PixelOperations_Base |    64 |   206.14 ns |      4.369 ns |   0.240 ns |  1.81 |    0.01 | 0.0057 |     - |     - |      24 B |
[...]

PR #1770, opened 2021-10-01
Pin sourceBase in HwIntrinsics ByteToNormalizedFloat

antonfirsov:
### Prerequisites

### Description
This fixes an overlook in #1398 (which I was reviewing ...), that exposes a potential memory corruption.

`source` can point to a managed array here, meaning if we don't ask for pinning explicitly, the GC can move the backing array, invalidating the destination address pointed by `sourceBase`.

In practice this almost never happens since big image arrays are allocated on the LOH today, however I still can imagine an unlucky situation with small images where this causes an error.

`AsPointer` should be used with care, only in cases when we know the destination reference is part of a pinned, native or stack object.


JimBobSquarePants (approved):
Good catch!

br3aker:
Isn't it better to use `ref`'s and `Unsafe.Add` for pointer arithmetic? This would eliminate the need to pin input span. But I'm not so sure about performance in comparison to raw pointers here.

antonfirsov:
> Isn't it better to use `ref`'s and `Unsafe.Add` instead of pointer arithmetic? 

Yes, JIT should generate the same machine code for `ref` -> value loading and it's generally better to avoid pinning, but I wanted a *very* quick 5 minute fix for now. Did a quick run of our `Rgba32 -> Vector4` benchmarks, and nothing changed, so decided to not bother anymore. May push a better fix later when I have more time dedicated to code around SIMD stuff.

JimBobSquarePants:
We can deal with that kind of improvement for V3. I’d like to focus now on simply shipping something so we can finally ditch the legacy and build what we want to.

br3aker:
> it's generally better to avoid pinning

As you already said, this code usually works with LOH memory so pinning should be almost free. My main point was about more 'C#'ish code :) 
But you're right, it's better to fix problems within a quick timespan than spend weeks writing 'best of best' code.

antonfirsov:
@JimBobSquarePants codecov here fails with a `-0.01%` change, not sure what is the reason. Can't we at least set a threshold different than `0.0 %`?

JimBobSquarePants:
>Can't we at least set a threshold different than 0.0 %?

The only thing I can see that might allow us to toggle this is the precision field. We could try updating that in the repo codecov.yml file and see what happens?
https://docs.codecov.com/docs/codecovyml-reference#coverageprecision

PR #1771, opened 2021-10-02
Use shared ArrayPoolMemoryAllocator in tests to avoid OOM

antonfirsov:
### Prerequisites

### Description

I realized that the main reason behind the 32 bit test OOM-s is that every `TestImageProvider` creates a new `ArrayPoolMemoryAllocator` which practically disables pooling for the entire test run. Except for a few cases using `LimitBufferCapacity` which the PR addresses, we should be safe to use the shared allocator.

#1730 is also contains this fix, but it makes sense to get rid of the OOM pain on master now.


JimBobSquarePants:
You're smashing out the fixes today! 😍

JimBobSquarePants (approved):

PR #1772, opened 2021-10-02
Fix a bug in MagickReferenceDecoder

antonfirsov:
### Prerequisites

### Description

This is causing intermittent failures on 32 bit. If `destinationGroup` contains of 2 `Memory<TPixel>` blocks, `rgbaBytes` is double of `destBuffer`, which is an invalid call to `PixelOperations.FromRgba32(Bytes)`, which wants to convert `source.Length` amount of pixels.


JimBobSquarePants (approved):
Great catch!!

PR #1773, opened 2021-10-03
Decode Jpeg into Rgb24 when pixel type not specified

### LINKED ISSUE #1121: Speed up JPEG Decoder color conversion

### LINKED ISSUE #1410: Non-generic Image.Load should decode Jpeg into Image<Rgb24>

antonfirsov:
### Prerequisites

### Description
Finish the plan defined in #1410:
- Remove packing from JpegColorConverters and utilize #1462 `PixelOperations.PackFromRgbPlanes` instead
- Return `Image<Rgb24>` from non-generic Decode
- `ResizeProcessor<T>`:  omit alpha premultiplication if the pixel type has no alpha channel

Fixes #1410, fixes #1121.

### Performance

Did usual run of `LoadResizeSaveStressBenchmarks` on 10 core desktop. The overall effect on speed is within noise range, but there might be some 1%-ish improvement. I assume the reason is that #1462's packing actually doesn't make big difference to the packing code introduced in #1411.

But memory usage dropped significantly, which is the main goal here.

#### Before
```
|        Method | maxDegreeOfParallelism |       Mean |       Error |   StdDev | Ratio | RatioSD |     Gen 0 |     Gen 1 |     Gen 2 |    Allocated |
|-------------- |----------------------- |-----------:|------------:|---------:|------:|--------:|----------:|----------:|----------:|-------------:|
| SystemDrawing |                      1 | 8,711.9 ms |   862.64 ms | 47.28 ms |  1.00 |    0.00 |         - |         - |         - |        25 KB |
|    ImageSharp |                      1 | 5,218.4 ms |   135.62 ms |  7.43 ms |  0.60 |    0.00 | 3000.0000 | 3000.0000 | 3000.0000 | 1,226,257 KB |
|               |                        |            |             |          |       |         |           |           |           |              |
| SystemDrawing |                      5 | 3,128.0 ms | 1,188.61 ms | 65.15 ms |  1.00 |    0.00 |         - |         - |         - |        26 KB |
|    ImageSharp |                      5 | 1,516.9 ms |   782.02 ms | 42.86 ms |  0.49 |    0.02 | 2000.0000 | 2000.0000 | 2000.0000 | 1,226,402 KB |
|               |                        |            |             |          |       |         |           |           |           |              |
| SystemDrawing |                     10 | 2,828.7 ms |   786.61 ms | 43.12 ms |  1.00 |    0.00 |         - |         - |         - |        29 KB |
|    ImageSharp |                     10 |   823.4 ms |    80.49 ms |  4.41 ms |  0.29 |    0.01 | 1000.0000 | 1000.0000 | 1000.0000 | 1,229,766 KB |
|               |                        |            |             |          |       |         |           |           |           |              |
| SystemDrawing |                     20 | 2,793.0 ms |   735.65 ms | 40.32 ms |  1.00 |    0.00 |         - |         - |         - |        30 KB |
|    ImageSharp |                     20 |   686.0 ms |   570.78 ms | 31.29 ms |  0.25 |    0.01 | 1000.0000 | 1000.0000 | 1000.0000 | 1,237,201 KB |
```

#### After
```
|        Method | maxDegreeOfParallelism |       Mean |      Error |   StdDev | Ratio | RatioSD |     Gen 0 |     Gen 1 |     Gen 2 |  Allocated |
|-------------- |----------------------- |-----------:|-----------:|---------:|------:|--------:|----------:|----------:|----------:|-----------:|
| SystemDrawing |                      1 | 8,757.5 ms | 1,109.5 ms | 60.81 ms |  1.00 |    0.00 |         - |         - |         - |      25 KB |
|    ImageSharp |                      1 | 4,868.9 ms |   258.8 ms | 14.18 ms |  0.56 |    0.00 | 2000.0000 | 2000.0000 | 2000.0000 | 776,094 KB |
|               |                        |            |            |          |       |         |           |           |           |            |
| SystemDrawing |                      5 | 3,222.5 ms |   913.8 ms | 50.09 ms |  1.00 |    0.00 |         - |         - |         - |      27 KB |
|    ImageSharp |                      5 | 1,397.8 ms |   624.2 ms | 34.21 ms |  0.43 |    0.02 |         - |         - |         - | 776,093 KB |
|               |                        |            |            |          |       |         |           |           |           |            |
| SystemDrawing |                     10 | 2,849.4 ms | 1,590.1 ms | 87.16 ms |  1.00 |    0.00 |         - |         - |         - |      29 KB |
|    ImageSharp |                     10 |   852.6 ms |   576.0 ms | 31.57 ms |  0.30 |    0.02 | 1000.0000 | 1000.0000 | 1000.0000 | 843,968 KB |
|               |                        |            |            |          |       |         |           |           |           |            |
| SystemDrawing |                     20 | 2,764.5 ms | 1,152.1 ms | 63.15 ms |  1.00 |    0.00 |         - |         - |         - |      30 KB |
|    ImageSharp |                     20 |   622.7 ms |   475.5 ms | 26.07 ms |  0.23 |    0.01 | 1000.0000 | 1000.0000 | 1000.0000 | 949,029 KB |
```

/cc @br3aker @JimBobSquarePants @saucecontrol


src/ImageSharp/Formats/Jpeg/Components/Decoder/SpectralConverter{TPixel}.cs
```diff
@@ -40,23 +43,20 @@ public SpectralConverter(Configuration configuration, CancellationToken cancella
 
         private bool Converted => this.pixelRowCounter >= this.pixelBuffer.Height;
 
-        public Buffer2D<TPixel> PixelBuffer
+        public Buffer2D<TPixel> GetPixelBuffer()
```
antonfirsov, line 46: There is heavy lazy initialization, with side effects like `CancellationToken` firing. It's better to expose such stuff from a method than a property.

br3aker: I didn't like this property approach with a lot of lazy stuff behind from the start but couldn't come up with anything better. There's a lot of confusing bits of code here and there from my pr, I will work on decoder performance/refactoring next as I'm done with the encoder for now.

JimBobSquarePants:
A lot of reading here. Will try to review tomorrow night

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromRgbVector8.cs
```diff
@@ -18,50 +18,32 @@ public FromRgbVector8(int precision)
...
-                ref Vector<float> ggRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref gg);
-                ref Vector<float> bbRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref bb);
+                    ref Unsafe.As<float, Vector<float>>(ref MemoryMarshal.GetReference(values.Component2));
 
                 var scale = new Vector<float>(1 / this.MaximumValue);
 
                 // Walking 8 elements at one step:
-                int n = result.Length / 8;
+                int n = values.Component0.Length / 8;
                 for (int i = 0; i < n; i++)
```
gfoidl, line 34:
```suggestion
                for (nint i = 0; i < n; i++)
```

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromRgbVector8.cs
```diff
@@ -18,50 +18,32 @@ public FromRgbVector8(int precision)
...
-                ref Vector<float> rrRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref rr);
-                ref Vector<float> ggRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref gg);
-                ref Vector<float> bbRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref bb);
+                    ref Unsafe.As<float, Vector<float>>(ref MemoryMarshal.GetReference(values.Component2));
 
                 var scale = new Vector<float>(1 / this.MaximumValue);
 
                 // Walking 8 elements at one step:
-                int n = result.Length / 8;
+                int n = values.Component0.Length / 8;
```
gfoidl, line 33:
```suggestion
                int n = (int)((uint)values.Component0.Length / 8);
```

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrAvx2.cs
```diff
@@ -50,19 +47,19 @@ protected override void ConvertCoreVectorized(in ComponentValues values, Span<Ve
                 Vector256<int> vcontrol = Unsafe.As<byte, Vector256<int>>(ref control);
 
                 // Walking 8 elements at one step:
-                int n = result.Length / 8;
+                int n = values.Component0.Length / 8;
```
gfoidl, line 50:
```suggestion
                int n = (int)((uint)(values.Component0.Length / 8);
```

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrAvx2.cs
```diff
@@ -50,19 +47,19 @@ protected override void ConvertCoreVectorized(in ComponentValues values, Span<Ve
...
                 for (int i = 0; i < n; i++)
```
gfoidl, line 51:
```suggestion
                for (nint i = 0; i < n; i++)
```

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector4.cs
```diff
@@ -20,73 +20,69 @@ public FromYCbCrVector4(int precision)
...
 
-                ref Vector4Octet resultBase =
-                    ref Unsafe.As<Vector4, Vector4Octet>(ref MemoryMarshal.GetReference(result));
-
                 var chromaOffset = new Vector4(-this.HalfValue);
                 var maxValue = this.MaximumValue;
 
                 // Walking 8 elements at one step:
-                int n = result.Length / 8;
+                int n = values.Component0.Length / 8;
```
gfoidl, line 38:
```suggestion
                int n = (int)((uint)values.Component0.Length / 8);
```

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector4.cs
```diff
@@ -20,73 +20,69 @@ public FromYCbCrVector4(int precision)
...
 
                 for (int i = 0; i < n; i++)
```
gfoidl, line 40:
```suggestion
                for (nint i = 0; i < n; i++)
```

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector8.cs
```diff
@@ -19,41 +19,32 @@ public FromYCbCrVector8(int precision)
...
-
-                Vector4Pair rr = default;
-                Vector4Pair gg = default;
-                Vector4Pair bb = default;
-
-                ref Vector<float> rrRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref rr);
-                ref Vector<float> ggRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref gg);
-                ref Vector<float> bbRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref bb);
-
+                int n = values.Component0.Length / 8;
```
gfoidl, line 34:
```suggestion
                int n = (int)((uint)values.Component0.Length / 8);
```
or maybe create a helper for this? The casts aren't very intuitive, but codegen is quite a difference.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector8.cs
```diff
@@ -19,41 +19,32 @@ public FromYCbCrVector8(int precision)
...
                 var scale = new Vector<float>(1 / this.MaximumValue);
 
                 for (int i = 0; i < n; i++)
```
gfoidl, line 37:
```suggestion
                for (nint i = 0; i < n; i++)
```

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYccKAvx2.cs
```diff
@@ -22,52 +22,42 @@ public FromYccKAvx2(int precision)
...
 
-                // Used for packing.
-                var va = Vector256.Create(1F);
-                ref byte control = ref MemoryMarshal.GetReference(HwIntrinsics.PermuteMaskEvenOdd8x32);
-                Vector256<int> vcontrol = Unsafe.As<byte, Vector256<int>>(ref control);
-
                 // Walking 8 elements at one step:
-                int n = result.Length / 8;
+                int n = values.Component0.Length / 8;
                 for (int i = 0; i < n; i++)
```
gfoidl, line 48:
```suggestion
                int n = (int)((uint)values.Component0.Length / 8);
                for (nint i = 0; i < n; i++)
```

Now I'll stop commenting on these -- otherwise it's too much noise generated by me 😉

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYccKVector8.cs
```diff
@@ -18,34 +18,23 @@ public FromYccKVector8(int precision)
...
-                Vector4Pair bb = default;
-
-                ref Vector<float> rrRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref rr);
-                ref Vector<float> ggRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref gg);
-                ref Vector<float> bbRefAsVector = ref Unsafe.As<Vector4Pair, Vector<float>>(ref bb);
+                int n = values.Component0.Length / 8;
 
-                var scale = new Vector<float>(1 / this.MaximumValue);
+                var scale = new Vector<float>(1 / (this.MaximumValue * this.MaximumValue));
                 var max = new Vector<float>(this.MaximumValue);
```
gfoidl, line 38:
```suggestion
                var max = new Vector<float>(this.MaximumValue);
                var scale = new Vector<float>(1f) / (max * max);
```
is [less ceremony](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4a9JqwByHAVGXqamgMxNSDIQwDeNBg6bmAZpIjYMDALLYE/fQBq2JIcMLYMAOYwGGoMuNGxAL72jikOYoYKHmEBMGAY0AA8ru4YAHwMAPq4YMEwGtSODBm8WRg5eQVQxW4eFZX4vg1NaU5MKAwAggAUAJSjdo1NjtW10gwAvAxcMADuDLn5RSV90+QMAPQM0xgAFry4LD5++IHBoQwAVAx3D0++/nwQRCMFms2GywcA18m22ewOnWOvXKN3uj2egOBoXBo2SSwco2I5mIEwAQnMFqMmlkoAxBghYTt9ocuj1SmVUX8Ma8ge9QRDIdCGVt6QLllTHDS4mswlsmQijt0TijyM4cfjlqs6rCatqrtN6V86b51U08YkgA).

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroupExtensions.cs
```diff
@@ -61,6 +61,35 @@ internal static Memory<T> GetBoundedSlice<T>(this IMemoryGroup<T> group, long st
...
+        internal static Memory<T> GetRemainingSliceOfBuffer<T>(this IMemoryGroup<T> group, long start)
+            where T : struct
+        {
+            Guard.NotNull(group, nameof(group));
+            Guard.IsTrue(group.IsValid, nameof(group), "Group must be valid!");
+            Guard.MustBeLessThan(start, group.TotalLength, nameof(start));
+
+            int bufferIdx = (int)(start / group.BufferLength);
+
+            if (bufferIdx < 0)
```
gfoidl, line 76: Can combine with the check below by using
```c#
if ((uint)bufferIdx >= group.Count)
{
   ...
}
```
A negative value for `bufferIdx` is after the `uint` cast in the range > int.MaxValue, so definitely >= group.Count too.

gfoidl (commented):
Just a quick look over the code, not detailed review.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromRgbVector8.cs
antonfirsov: I think this trick is only worth it when we are inside a tight loop or, or in a method that is expected to be called from a loop. Otherwise I prefer readability.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector4.cs
antonfirsov: I'm not sure it's worth it, see https://github.com/SixLabors/ImageSharp/pull/1773#discussion_r720905777

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromYCbCrVector8.cs
antonfirsov: I'm not sure it's worth it, see https://github.com/SixLabors/ImageSharp/pull/1773#discussion_r720905777

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromRgbVector8.cs
antonfirsov: Or is it possible to achieve the same in a simpler way `nint`? I changed my code everywhere to `nint n = values.Component0.Length / 8;` because `n` is only used for the `i < n` comparison.

gfoidl: We should go with readability, as you did now with `nint`. Codegen isn't ideal, but this pattern `{array, span}.Length / ConstantOfPowerOfTwo` is so common, that the JIT should be able to optimize this, so we don't need any hacks.
I'll have a look if there's a JIT-issue, otherwise create one.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverter.FromGrayScaleBasic.cs
```diff
@@ -17,25 +17,35 @@ public FromGrayscaleBasic(int precision)
...
+                if (!values.IsEmpty)
+                {
+                    float scaleValue = 1f / maxValue;
+                    values[0] *= scaleValue;
+
+                    if (values.Length > 1)
+                    {
+                        values[1] *= scaleValue;
+
+                        if (values.Length > 2)
```
gfoidl, line 44:
```suggestion
                    if ((uint)values.Length > 1)
                    {
                        values[1] *= scaleValue;

                        if ((uint)values.Length > 2)
```
Elides further bound checks.
(Sorry that I didn't comment on that yesterday)

JimBobSquarePants (approved):
I can't believe how much easier the code is to read now! I did a few readthroughs just to make sure I wasn't missing something 😄 

Just @gfoidl last suggestion and it's good to go. 👍

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeProcessor{TPixel}.cs
```diff
@@ -187,6 +188,8 @@ private static PixelConversionModifiers GetModifiers(bool compand, bool premulti
             bool compand,
             bool premultiplyAlpha)
         {
+            bool pixelHasNoAlpha = PixelOperations<TPixel>.Instance.GetPixelTypeInfo()?.AlphaRepresentation == PixelAlphaRepresentation.None;
+            premultiplyAlpha &= !pixelHasNoAlpha;
```
antonfirsov, line 192: @JimBobSquarePants I couldn't figure out how should this *really* work with our existing pixel types, so went with the safest option here (only omit premultiplication when we are 100% sure that the alpha channel is missing).

However, if I understand it correctly normally we should only premultiply in case of `PixelAlphaRepresentation.Associated`. What confuses me is that `Rgba32` and friends are configured to `Unassociated`:

https://github.com/SixLabors/ImageSharp/blob/97dde7fb2f0ac97b5b1f9e6b0555a53d2e270c46/src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Rgba32.PixelOperations.cs#L22-L23

JimBobSquarePants: Ah no. Associated alpha means premultiplied. The only occasion we should premultiply is `PixelAlphaRepresentation.Unassociated`
https://en.wikipedia.org/wiki/Alpha_compositing#Straight_versus_premultiplied

antonfirsov: Thanks, I see now!

Would be the following code better here then?
```suggestion
            PixelAlphaRepresentation? alphaRepresentation = PixelOperations<TPixel>.Instance.GetPixelTypeInfo()?.PixelAlphaRepresentation;

            // Premultiply only if alpha representation is unknown or Unassociated:
            bool needsPremultiplication = alphaRepresentation == null || alphaRepresentation.Value == PixelAlphaRepresentation.Unassociated;
            premultiplyAlpha &= needsPremultiplication;
```

JimBobSquarePants: Yep. That looks good to me. 👍

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeProcessor{TPixel}.cs
```diff
@@ -187,6 +188,11 @@ private static PixelConversionModifiers GetModifiers(bool compand, bool premulti
             bool compand,
             bool premultiplyAlpha)
         {
+            PixelAlphaRepresentation? alphaRepresentation = PixelOperations<TPixel>.Instance.GetPixelTypeInfo()?.PixelAlphaRepresentation;
```
JimBobSquarePants, line 191:
```suggestion
            PixelAlphaRepresentation? alphaRepresentation = PixelOperations<TPixel>.Instance.GetPixelTypeInfo()?.AlphaRepresentation;
```

PR #1794, opened 2021-10-27
Support executing benchmarks on arm64

kunalspathak:
### Prerequisites

### Description

Because of existing roslyn issue https://github.com/dotnet/roslyn/issues/42393, building and executing the benchmarks fail on arm64. In the past, we fixed this issue in dotnet/performance on arm64 in https://github.com/dotnet/performance/pull/1243 and related change in https://github.com/dotnet/BenchmarkDotNet/issues/1395. This PR does similar changes to make sure the benchmarks can be executed on arm64 platform.

### DIFF
--- a/src/ImageSharp/Formats/Webp/Lossless/WebpLosslessDecoder.cs
+++ b/src/ImageSharp/Formats/Webp/Lossless/WebpLosslessDecoder.cs
@@ -218,7 +218,7 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossless
             ColorCache colorCache = decoder.Metadata.ColorCache;
             int colorCacheLimit = lenCodeLimit + colorCacheSize;
             int mask = decoder.Metadata.HuffmanMask;
-            HTreeGroup[] hTreeGroup = GetHTreeGroupForPos(decoder.Metadata, col, row);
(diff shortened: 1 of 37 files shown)


kesiasimon:
Hi . I am trying to test the performance (both memory and cpu usage for file read &write ,Network IO) of Arm64 machine by running a simple performance testing console application using benchmarkdotnet. (I need to test it for both x64 as well as ARM64 platform in .net4.7.2 and .net6) .Now i've build a benchmark application with .net6 and platform as ARM64. but m not able to run it on ARM64 machine. Could you plz help to solve this issue.

kunalspathak:
> but m not able to run it on ARM64 machine. Could you plz help to solve this issue.

Can I give details of steps you are using to run benchmarks and what errors are you getting?

kesiasimon:
I've created a simple benchmark application and it is running fine on visual studio.but if i take the release folder out and try to test the same then the application is not running. it is showing error like unable to find the .sln or .csproj file. and i try to publish the application and tried with the setup .which is also not working and giving the same error.  now i've tried it within my pc (instead of ARM pc) that is also giving the same error. the application is only running inside the visual studio. but for testing i need a build file.

PR #1797, opened 2021-10-29
Support High Precision TPixel formats via Color

JimBobSquarePants:
### Prerequisites

### Description

Switches out the `Rgba64` color backing type for `RgbaVector` in `Color`. This allows supporting all the pixel formats in the library without loss of precision. 

A new `Color.FromPixel<TPixel>(TPixel pixel)` method has been provided to support this.

See https://github.com/SixLabors/ImageSharp.Drawing/discussions/165


JimBobSquarePants:
@antonfirsov one thing I’m struggling with is how bulk operations would work with the boxed value. From a safety perspective you’d have to check each item in the input Color span to ensure they conform to the correct type.

tocsoft:
> @antonfirsov one thing I’m struggling with is how bulk operations would work with the boxed value. From a safety perspective you’d have to check each item in the input Color span to ensure they conform to the correct type.

We don't/shouldn't be using `Color` in any bulk operations, we always translate to a real `IPixel` first before any sort of bulk operations happen, also we should never be convert between a `Color` to `IPixel` in any sort of loop.

Also do we even have any `Span<Color>` usage? I didn't think we did. `Color` isn't an `IPixel` after all, its just there to provide a way of providing colors in an `IPixel` agnostic format until we have access to the generic type and as soon as we have the access to the generic `TPixel` type we immediately convert it over.

antonfirsov:
I agree with @tocsoft, we don't do (and should never do) bulk conversion of `Color` on hot path.

> Also do we even have any `Span<Color>` usage? 

When we initialize a pixel-specific operation executor, that uses a series of colors, we need to convert `Span<Color>` to `TPixel[]` like in case of `PaletteQuantizer`:
https://github.com/SixLabors/ImageSharp/blob/4638965c3748a916f4573164921dd0f101093cdc/src/ImageSharp/Processing/Processors/Quantization/PaletteQuantizer.cs#L56-L60

This happens on a very small batch of data, so the cost is negligible compared to the quantization.

JimBobSquarePants:
We already expose a bulk method on the Color type.

https://github.com/SixLabors/ImageSharp/pull/1797/files#diff-03af69fc3b4de3f01abdac6b07435fea4a80e74d6d15f4bb3223c85d073466a9R242

antonfirsov:
@JimBobSquarePants yes, we use that method in `PaletteQuantizer`.

It exists because of an effort to optimize as much as possible, but that doesn't mean the usages are on a hot path, and the conversion efficiency has measurable impact.

JimBobSquarePants:
I’m just pointing out that we expose is and as such we’d have to lose any optimisation to use the boxes value. We wouldn’t use it on a hot path nor mix color precision but who knows what madness lies beyond our walls!

I’m happy to experiment and convert it to a loop with check but just wanted everyone to be aware.

JimBobSquarePants:
Closed in favor of #1801

Issue #1786, opened 2021-10-23
WebP - Improve performance 


JimBobSquarePants:
![TomCruiseNeedForSpeedGIF](https://user-images.githubusercontent.com/385879/138555482-5f10fd3d-f49d-4eec-8b59-450228766449.gif)

Now that WebP has been merged #1552 into master we need to polish the performance to bring it up to speed for the release of V2. 

We're looking for community contribution here while we focus on bring the Drawing/Fonts libraries up to RC status and also work on the new Unamanaged Memory allocator #1730 so anyone who like to make things go fast please have a crack.

### Benchmarks

The last benchmarks (June 2021) indicate that we're a good bit slower than the Magick.NET implementation. (_There's a positive here, potentially lots of low hanging fruit_)

**Decode**

Big improvements in .NET Core with some SIMD which is positive. Definitely some allocations to deal with but will require profiling to determine if something else is in play.

```
Intel Core i7-6700K CPU 4.00GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.202
    [Host]     : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT
    Job-AQFZAV : .NET Framework 4.8 (4.8.4180.0), X64 RyuJIT
    Job-YCDAPQ : .NET Core 2.1.18 (CoreCLR 4.6.28801.04, CoreFX 4.6.28802.05), X64 RyuJIT
    Job-WMTYOZ : .NET Core 3.1.4 (CoreCLR 4.700.20.20201, CoreFX 4.700.20.22101), X64 RyuJIT

IterationCount=3  LaunchCount=1  WarmupCount=3
|                     Method |        Job |       Runtime |        TestImageLossy |        TestImageLossless |       Mean |     Error |   StdDev |     Gen 0 |     Gen 1 | Gen 2 |   Allocated |
|--------------------------- |----------- |-------------- |---------------------- |------------------------- |-----------:|----------:|---------:|----------:|----------:|------:|------------:|
|        'Magick Lossy Webp' | Job-IERNAB |    .NET 4.7.2 | Webp/earth_lossy.webp | Webp/earth_lossless.webp |   105.8 ms |   6.28 ms |  0.34 ms |         - |         - |     - |    17.65 KB |
[...]

PR #1799, opened 2021-10-31
Webp encoding: Reduce allocations.

brianpopow:
### Prerequisites

### Description

This PR is meant to reduce the allocations during webp encoding significantly. 

Related to #1786

Before:

```
BenchmarkDotNet=v0.13.0, OS=Windows 10.0.19043.1288 (21H1/May2021Update)
.NET SDK=6.0.100-rc.2.21505.57
  [Host]     : .NET 5.0.11 (5.0.1121.47308), X64 RyuJIT
  Job-OKVXPJ : .NET 5.0.11 (5.0.1121.47308), X64 RyuJIT
  Job-PBVPYV : .NET Core 3.1.20 (CoreCLR 4.700.21.47003, CoreFX 4.700.21.47101), X64 RyuJIT
  Job-SLAEGF : .NET Framework 4.8 (4.8.4420.0), X64 RyuJIT

|                     Method |        Job |              Runtime |             Arguments |    TestImage |      Mean |      Error |    StdDev | Ratio | RatioSD |       Gen 0 |     Gen 1 |     Gen 2 |  Allocated |
|--------------------------- |----------- |--------------------- |---------------------- |------------- |----------:|-----------:|----------:|------:|--------:|------------:|----------:|----------:|-----------:|
|        'Magick Webp Lossy' | Job-OKVXPJ |             .NET 5.0 | /p:DebugType=portable | Png/Bike.png |  26.78 ms |   4.110 ms |  0.225 ms |  0.15 |    0.00 |           - |         - |         - |      68 KB |
|    'ImageSharp Webp Lossy' | Job-OKVXPJ |             .NET 5.0 | /p:DebugType=portable | Png/Bike.png | 285.95 ms |  33.692 ms |  1.847 ms |  1.56 |    0.03 | 135000.0000 |         - |         - | 552,714 KB |
|     'Magick Webp Lossless' | Job-OKVXPJ |             .NET 5.0 | /p:DebugType=portable | Png/Bike.png | 183.18 ms |  34.303 ms |  1.880 ms |  1.00 |    0.00 |           - |         - |         - |     520 KB |
| 'ImageSharp Webp Lossless' | Job-OKVXPJ |             .NET 5.0 | /p:DebugType=portable | Png/Bike.png | 347.68 ms |  67.105 ms |  3.678 ms |  1.90 |    0.02 |  34000.0000 | 5000.0000 | 2000.0000 | 161,675 KB |
|                            |            |                      |                       |              |           |            |           |       |         |             |           |           |            |
|        'Magick Webp Lossy' | Job-PBVPYV |        .NET Core 3.1 |               Default | Png/Bike.png |  26.77 ms |   8.193 ms |  0.449 ms |  0.15 |    0.00 |           - |         - |         - |      68 KB |
|    'ImageSharp Webp Lossy' | Job-PBVPYV |        .NET Core 3.1 |               Default | Png/Bike.png | 294.26 ms | 156.747 ms |  8.592 ms |  1.62 |    0.08 | 135000.0000 |         - |         - | 552,713 KB |
|     'Magick Webp Lossless' | Job-PBVPYV |        .NET Core 3.1 |               Default | Png/Bike.png | 181.80 ms |  69.170 ms |  3.791 ms |  1.00 |    0.00 |           - |         - |         - |     520 KB |
| 'ImageSharp Webp Lossless' | Job-PBVPYV |        .NET Core 3.1 |               Default | Png/Bike.png | 357.09 ms | 433.254 ms | 23.748 ms |  1.97 |    0.16 |  34000.0000 | 5000.0000 | 2000.0000 | 161,668 KB |
|                            |            |                      |                       |              |           |            |           |       |         |             |           |           |            |
|        'Magick Webp Lossy' | Job-SLAEGF | .NET Framework 4.7.2 |               Default | Png/Bike.png |  26.63 ms |   2.530 ms |  0.139 ms |  0.14 |    0.00 |           - |         - |         - |      68 KB |
|    'ImageSharp Webp Lossy' | Job-SLAEGF | .NET Framework 4.7.2 |               Default | Png/Bike.png | 429.41 ms |  38.772 ms |  2.125 ms |  2.33 |    0.01 | 135000.0000 |         - |         - | 554,351 KB |
|     'Magick Webp Lossless' | Job-SLAEGF | .NET Framework 4.7.2 |               Default | Png/Bike.png | 183.96 ms |  33.354 ms |  1.828 ms |  1.00 |    0.00 |           - |         - |         - |     523 KB |
| 'ImageSharp Webp Lossless' | Job-SLAEGF | .NET Framework 4.7.2 |               Default | Png/Bike.png | 428.93 ms |  60.292 ms |  3.305 ms |  2.33 |    0.04 |  34000.0000 | 5000.0000 | 2000.0000 | 162,125 KB |
```

After

```
|                     Method |        Job |              Runtime |    TestImage |      Mean |      Error |    StdDev | Ratio | RatioSD |     Gen 0 |     Gen 1 |     Gen 2 | Allocated |
|--------------------------- |----------- |--------------------- |------------- |----------:|-----------:|----------:|------:|--------:|----------:|----------:|----------:|----------:|
|        'Magick Webp Lossy' | Job-TRNEBU |             .NET 5.0 | Png/Bike.png |  26.38 ms |   4.031 ms |  0.221 ms |  0.14 |    0.00 |         - |         - |         - |     68 KB |
|    'ImageSharp Webp Lossy' | Job-TRNEBU |             .NET 5.0 | Png/Bike.png | 232.16 ms |  50.122 ms |  2.747 ms |  1.27 |    0.02 | 4000.0000 |         - |         - | 16,704 KB |
|     'Magick Webp Lossless' | Job-TRNEBU |             .NET 5.0 | Png/Bike.png | 182.79 ms |   8.574 ms |  0.470 ms |  1.00 |    0.00 |         - |         - |         - |    520 KB |
| 'ImageSharp Webp Lossless' | Job-TRNEBU |             .NET 5.0 | Png/Bike.png | 350.50 ms |  66.007 ms |  3.618 ms |  1.92 |    0.02 | 8000.0000 | 3000.0000 | 1000.0000 | 61,556 KB |
|                            |            |                      |              |           |            |           |       |         |           |           |           |           |
|        'Magick Webp Lossy' | Job-XLXDYZ |        .NET Core 3.1 | Png/Bike.png |  25.11 ms |   8.152 ms |  0.447 ms |  0.14 |    0.01 |         - |         - |         - |     68 KB |
|    'ImageSharp Webp Lossy' | Job-XLXDYZ |        .NET Core 3.1 | Png/Bike.png | 241.75 ms | 245.192 ms | 13.440 ms |  1.36 |    0.11 | 4000.0000 |         - |         - | 16,703 KB |
|     'Magick Webp Lossless' | Job-XLXDYZ |        .NET Core 3.1 | Png/Bike.png | 178.20 ms | 102.407 ms |  5.613 ms |  1.00 |    0.00 |         - |         - |         - |    520 KB |
| 'ImageSharp Webp Lossless' | Job-XLXDYZ |        .NET Core 3.1 | Png/Bike.png | 332.18 ms | 112.166 ms |  6.148 ms |  1.87 |    0.07 | 8000.0000 | 3000.0000 | 1000.0000 | 61,555 KB |
|                            |            |                      |              |           |            |           |       |         |           |           |           |           |
|        'Magick Webp Lossy' | Job-XXCPPX | .NET Framework 4.7.2 | Png/Bike.png |  26.47 ms |   1.311 ms |  0.072 ms |  0.14 |    0.00 |         - |         - |         - |     68 KB |
|    'ImageSharp Webp Lossy' | Job-XXCPPX | .NET Framework 4.7.2 | Png/Bike.png | 473.98 ms |  87.433 ms |  4.793 ms |  2.59 |    0.02 | 4000.0000 |         - |         - | 16,753 KB |
|     'Magick Webp Lossless' | Job-XXCPPX | .NET Framework 4.7.2 | Png/Bike.png | 182.86 ms |  26.630 ms |  1.460 ms |  1.00 |    0.00 |         - |         - |         - |    523 KB |
| 'ImageSharp Webp Lossless' | Job-XXCPPX | .NET Framework 4.7.2 | Png/Bike.png | 518.88 ms | 128.756 ms |  7.058 ms |  2.84 |    0.05 | 8000.0000 | 3000.0000 | 1000.0000 | 61,697 KB |
```


JimBobSquarePants (approved):
Nice improvement :+1:

src/ImageSharp/Formats/Webp/Lossless/HuffmanUtils.cs
```diff
@@ -202,9 +202,14 @@ public static void GenerateOptimalTree(HuffmanTree[] tree, uint[] histogram, int
                 }
 
                 // Build the Huffman tree.
+#if NET5_0_OR_GREATER
+                Span<HuffmanTree> treeSlice = tree.AsSpan().Slice(0, treeSize);
+                treeSlice.Sort(HuffmanTree.Compare);
+#else
                 HuffmanTree[] treeCopy = tree.AsSpan().Slice(0, treeSize).ToArray();
```
JimBobSquarePants, line 209: It's unfortunate that .NET Core 3.1 doesn't give us sort here. We could copy [`ArraySortHelper`](https://source.dot.net/#System.Private.CoreLib/ArraySortHelper.cs,63a9955a91f2b37b) or even use something like [`SortUtility`](https://github.com/SixLabors/ImageSharp.Drawing/blob/e987bd8c428792c8a00cce0bc50b59b6401a280c/src/ImageSharp.Drawing/Utilities/SortUtility.cs) from Drawing but I don't know if it's worth the effort. @antonfirsov what do you think?

src/ImageSharp/Formats/Webp/Lossless/PredictorEncoder.cs
```diff
@@ -17,6 +17,13 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossless
     /// </summary>
     internal static unsafe class PredictorEncoder
     {
+        private static readonly sbyte[] DeltaLut = { 16, 16, 8, 4, 2, 2, 2 };
```
gfoidl, line 20:
```suggestion
        // This uses C#'s compiler optimization to refer to assembly's static data directly.
        private static ReadOnlySpan<sbyte> DeltaLut => new sbyte[] { 16, 16, 8, 4, 2, 2, 2 };
```

src/ImageSharp/Formats/Webp/Lossless/Vp8LEncoder.cs
```diff
@@ -76,6 +85,8 @@ internal class Vp8LEncoder : IDisposable
 
         private const int PaletteInvSize = 1 << PaletteInvSizeBits;
 
+        private static readonly byte[] Order = { 1, 2, 0, 3 };
```
gfoidl, line 88:
```suggestion
        // This uses C#'s compiler optimization to refer to assembly's static data directly.
        private static ReadOnlySpan<byte> Order => new byte[] { 1, 2, 0, 3 };
```

src/ImageSharp/Formats/Webp/Lossless/Vp8LEncoder.cs
```diff
@@ -19,6 +19,15 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossless
     /// </summary>
     internal class Vp8LEncoder : IDisposable
     {
+        /// <summary>
+        /// Scratch buffer to reduce allocations.
+        /// </summary>
+        private readonly int[] scratch = new int[256];
```
gfoidl, line 25: Instead of the `int[]` one could use fixed buffer wrapped in a private struct. See [this example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQGADAAQY4B0ASgK4B2ALgJYC2MlAkkzFBAA4BlXgDdmYGAGcA3IUIYAzOSykAwqQDehUjtLMeUegEMANqQaSjAMxjLSANX4AOADIBRepAAmvbbq0EukGk/FDMIkaMtoJgUJFgABakkrHxSQC8pPQwAO4AFACUsoHBOn6liuQopMKMefqMevQ+CAA0TY0RJrQwBaTpAHzJqYyJANr6LQC6/aRdPcWlOpUY1aomMEZQ9UzJjFuM7Q2kG/QA5owJfYPDcaMJlACCkoL8RvR5kvtQhycw55cCpR1pttkVysEIUFQuFIrYvlBaGBGjE7okof4MaUYREoqQrMwEDAvB1SMBaFYbFAxlgAKxIKaLJZBLHBSqwKyky7MSQTXaTGAIGY3DlkilUvnTJnMsolGXkJSvd4AHgaQ2eSo+xy+ByOu1OFyurKCAXlpVF2u+jGoME5mVF3N5Op+jONSwwAHZSABZGCsaAAT29W0kCVMwNgcM1eVFzuttvaBsB0plAF9Wem5aRM6mgA=).
(Note: the encapsulating class must be `unsafe` in order to make it work)

So `scratch` becomes a strong type and can have methods too 😄 
It avoids the extra array allocation, but at the cost of a bigger `Vp8LEncoder` -- I don't know what is better.

For code elegance I'd prefer the fixed buffer / strong type approach, as there could be a method like `Span<int> AsSpan(int start, int length, bool clear = true)` and that like.

src/ImageSharp/Formats/Webp/Lossless/Vp8LEncoder.cs
```diff
@@ -780,7 +793,8 @@ private void EncodeImageNoHuffman(Span<uint> bgra, Vp8LHashChain hashChain, Vp8L
         private void StoreHuffmanCode(HuffmanTree[] huffTree, HuffmanTreeToken[] tokens, HuffmanTreeCode huffmanCode)
         {
             int count = 0;
-            int[] symbols = { 0, 0 };
+            Span<int> symbols = this.scratch.AsSpan(0, 2);
+            symbols.Clear();
```
gfoidl, line 797: E.g. with the comment above this could be 
```c#
Span<int> symbols = this.scratch.AsSpan(0, 2, clear: true);
```

src/ImageSharp/Formats/Webp/Lossless/Vp8LEncoder.cs
gfoidl: Or rent this array from an array pool / memory pool?
(Maybe that overkill here?)

src/ImageSharp/Formats/Webp/Lossless/WebpLosslessDecoder.cs
```diff
@@ -418,14 +418,15 @@ private void ReadHuffmanCodes(Vp8LDecoder decoder, int xSize, int ySize, int col
             var huffmanTables = new HuffmanCode[numHTreeGroups * tableSize];
             var hTreeGroups = new HTreeGroup[numHTreeGroups];
             Span<HuffmanCode> huffmanTable = huffmanTables.AsSpan();
+            int[] codeLengths = new int[maxAlphabetSize];
```
gfoidl, line 421: Rent from pool?

src/ImageSharp/Formats/Webp/Lossy/Vp8EncIterator.cs
```diff
@@ -81,6 +81,8 @@ public Vp8EncIterator(byte[] yTop, byte[] uvTop, uint[] nz, Vp8MacroBlockInfo[]
             this.I4Boundary = new byte[37];
             this.BitCount = new long[4, 3];
             this.Scratch = new byte[WebpConstants.Bps * 16];
+            this.Scratch2 = new short[17 * 16];
+            this.Scratch3 = new int[16];
```
gfoidl, line 85: I'm kind of fan a fixed sized buffers -- so would try to use these here too.
Yeah they make the object bigger (by the size of the buffer), but it avoids further objects that the GC needs to track, so most of the time it's advantageous.

**Note**: it's unsafe, so there are no bound checks emitted!

src/ImageSharp/Formats/Webp/Lossy/Vp8Encoder.cs
```diff
@@ -70,6 +70,11 @@ internal class Vp8Encoder : IDisposable
         /// </summary>
         private int uvAlpha;
 
+        /// <summary>
+        /// Scratch buffer to reduce allocations.
+        /// </summary>
+        private readonly int[] scratch = new int[16];
```
gfoidl, line 76: Here too.
(I'll stop commenting on these buffer for the rest of the code).

src/ImageSharp/Formats/Webp/Lossless/Vp8LEncoder.cs
brianpopow: @gfoidl would you think the same trick could be applied to the BitMask in `Vp8LBitReader`?

gfoidl: You mean https://github.com/SixLabors/ImageSharp/blob/d0212221622b015706ca4d0728c269ec3caa5c7a/src/ImageSharp/Formats/Webp/BitReader/Vp8LBitReader.cs#L31-L40 ?

Then this C# optimization won't work, as it only works for single byte-size types (`byte`, `sbyte`, `bool`) due to endianess concerns. There is https://github.com/dotnet/runtime/issues/60948 to allow other (primitive and simple) types too.

One could type it a sequence of bytes in order to the ROS<byte>-trick, then cast that span to a ROS<uint> one. In that case a) code becomes ugly and b) same endianess-trouble occurs. I wouldn't do this -- once the aforementioned runtime-issue is resolved we may do a pass over the codebase to see where to apply this trick for other types than single byte-sized ones.

---

But this field should be made `static readonly` so that the JIT can treat it as "constant" (and for Tier-1 can skip the beforefield init check).

brianpopow: yes that's what I meant. Thanks for the clarification.

Another question i have: For what sizes of array's would you recommend using this trick. I guess it should be rather small array's only? We have alot of `byte`, `sbyte` lookups in `WebpLookupTables`. Would it make sense to apply this there also or should we not go to crazy with this?

gfoidl: I'd do it for all static data of type byte, sbyte, bool -- no matter what size.

Look at this [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQGAzAATakDCphA3oaY6QA5QCWAbgIYAuM5OJKVhcAJhAB2AGwCepYDL4BtALqkAzgH0uUKFzkBeUhJgB3eYpirStUjgA0pLI+KOUjgKyOkjgOyOADkcATlIAXwBuBiZoxgB6ONIlAAY1NnVSUTYAM2yYKAkeRx4IUgALLg5+HlNSmEL2GAy2CVIeMv4ASQAZWNIE0gAVKDkS0jAKiQBzfhS1IwcNUvUuAFt+LgyuYAgq0i4JUQ0YfjKIc3VVnR5SACUIdVlW9NIQOABKPtZOXn4MQTuMDEAHlpDIAMrMA4AHgAQpYAHykSEHUgGJEmcwKZRqWzJRzOUiuUjuUheUg+Uj+UhBUihSKEMJAA==). You can see that for the array and for the ROS there's data in the static segment of the assembly. So the size of the DLL won't increase by using the ROS-trick extensively. 

The big advantage is that the ROS refers to the static data directly, no need for the extra array allocation (even if static initialized, so as singleton), and that with the ROS the data is truly read-only (for the array you can write to the elements even if it is declared as readonly, as this is just for the reference to the array itself).

src/ImageSharp/Formats/Webp/Lossless/Vp8LEncoder.cs
brianpopow: I think the array is too small to utilize the `MemoryAllocator` here.

On the fixed buffer wrapped in a private struct idea, I am undecided. @JimBobSquarePants what do you think should we go that way or should i leave it as it is?

JimBobSquarePants: We use that trick in the jpeg decoder so there's precedence. The input to the enclosing type is always sanitized since your encoding so I'd go for it if you have the time. 

https://github.com/SixLabors/ImageSharp/blob/d0212221622b015706ca4d0728c269ec3caa5c7a/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs#L13-L56

src/ImageSharp/Formats/Webp/Lossless/HuffmanUtils.cs
```diff
@@ -202,9 +202,14 @@ public static void GenerateOptimalTree(HuffmanTree[] tree, uint[] histogram, int
                 }
 
                 // Build the Huffman tree.
+#if NET5_0_OR_GREATER
+                Span<HuffmanTree> treeSlice = tree.AsSpan().Slice(0, treeSize);
```
turbedi, line 206:
```suggestion
                Span<HuffmanTree> treeSlice = tree.AsSpan(0, treeSize);
```
Also applicable on line 209

src/ImageSharp/Formats/Webp/Lossless/Vp8LEncoder.cs
brianpopow: mhm `MemoryMarshal.CreateSpan` does not exist in `net472`, `netstandard1.3` and `netstandard2.0`. What would be the right way to get the Span there? I think, i miss something obvious.

gfoidl: Ah, that's too bad. One would need to create the span via the pointer overload, but that needs a fixed-block which isn't feasible here (? or is it).

But: do we need to create a span at all? The span is needed to call Clear, and use indexed access to the elements. This can be done in the `Scratch`-type / buffer directly. So the buffer could be passed around instead of the span.
Note: the buffer is now a type, so can have methods too.

In order to clear effeciently for .NET Core+ one could create a span and call clear (as it's vectorized), for older runtimes just use a manual loop (maybe unrolled).
Or even better just rely on the runtime to zero-init (iif the `[SkipLocalsInit]` isn't applied, which isn't at the moment) the fixed sized buffer, so instead of clearing create a new buffer.

I'll think about this a bit more, if there is maybe any better approach to that.

src/ImageSharp/Formats/Webp/Lossless/HuffmanUtils.cs
brianpopow: changed with e97c364b

src/ImageSharp/Formats/Webp/Lossless/Vp8LEncoder.cs
brianpopow: I tried to change it to a struct, but without being able to create a Span from `MemoryMarshal.CreateSpan` it seems to make things more complicated then before. For example all methods in `PredictorEncoder` operate on Spans.
Maybe there is a way of doing it without overcomplicating things, but i dont see it.

I would like to keep it as it is for now.

brianpopow:
@JimBobSquarePants any objections that we merge this PR as it is? I really would like to see how it does in combination with the other optimizations. I think the reduced allocations can be beneficial to performance, too.

JimBobSquarePants:
@brianpopow no objections, rock on🤘

PR #1801, opened 2021-11-01
Support High Precision TPixel formats via Color - Boxed

JimBobSquarePants:
### Prerequisites

### Description

An alternative to #1797 that uses a boxed `IPixel` representation to hold any Color created via new `Color.FromPixel<TPixel>(TPixel pixel)` method or via `Color(Vector4)`. This allows supporting all the pixel formats in the library without loss of precision. 

See https://github.com/SixLabors/ImageSharp.Drawing/discussions/165

CC. @tocsoft @antonfirsov


src/ImageSharp/Color/Color.cs
```diff
@@ -91,7 +96,18 @@ private Color(byte r, byte g, byte b)
...
+        /// <summary>
+        /// Creates a <see cref="Color"/> from the given <typeparamref name="TPixel"/>.
+        /// </summary>
+        /// <param name="pixel">The pixel to convert from.</param>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <returns>The <see cref="Color"/>.</returns>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public static Color FromPixel<TPixel>(TPixel pixel)
+            where TPixel : unmanaged, IPixel<TPixel>
+            => new(pixel);
```
antonfirsov, line 110: I'm surprised we didn't have this method before. We should avoid boxing in case user provides a typical pixel like `Rgba32`, which I would expect to be the more common than advanced scenarios like https://github.com/SixLabors/ImageSharp.Drawing/discussions/165

I realized that we don't need to add new stuff to `PixelOperations<T>` unless we see users complaining about the default heuristic, which is very unlikely.

```suggestion
            => Unsafe.SizeOf<TPixel> > sizeof(Rgba64)  ? new(pixel) : new(pixel.ToRgba64());
```

src/ImageSharp/Color/Color.cs
```diff
@@ -240,28 +261,38 @@ public TPixel ToPixel<TPixel>()
             Span<TPixel> destination)
             where TPixel : unmanaged, IPixel<TPixel>
         {
-            ReadOnlySpan<Rgba64> rgba64Span = MemoryMarshal.Cast<Color, Rgba64>(source);
-            PixelOperations<TPixel>.Instance.FromRgba64(configuration, rgba64Span, destination);
+            Guard.DestinationShouldNotBeTooShort(source, destination, nameof(destination));
+            for (int i = 0; i < source.Length; i++)
+            {
+                destination[i] = source[i].ToPixel<TPixel>();
+            }
```
antonfirsov, line 268: Just a note that it's possible to detect the common `boxedHighPrecisionPixel  == null` case with SIMD and bulk convert, if we really want to.

src/ImageSharp/Color/Color.cs
```diff
@@ -240,28 +261,38 @@ public TPixel ToPixel<TPixel>()
...
         [MethodImpl(InliningOptions.ShortMethod)]
         public bool Equals(Color other)
         {
-            return this.data.PackedValue == other.data.PackedValue;
+            if (this.boxedHighPrecisionPixel is null && other.boxedHighPrecisionPixel is null)
+            {
+                return this.data.PackedValue == other.data.PackedValue;
+            }
+
+            return this.ToVector4().Equals(other.ToVector4());
```
antonfirsov, line 280: This implementation is not correct when the pixel is really a high precision one. I think we should expect custom `IPixel` implementations to override `object.Equals(object)`, like our pixels do. This is a practice `struct` equality implementers should follow IMO.

antonfirsov (commented):
We should avoid boxing when not necessary, and ensure equality correctness, otherwise looks good!

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Yeah wasn’t sure about that.

tests/ImageSharp.Tests/Color/ColorTests.CastTo.cs
```diff
@@ -88,6 +88,19 @@ public void Bgr24()
                 Bgr24 data = color;
                 Assert.Equal(source, data);
             }
+
+            [Fact]
+            public void TPixel()
+            {
+                var source = new RgbaVector(1, .1F, .133F, .864F);
```
antonfirsov, line 95: I recommend to stress the precision in this method:

```suggestion
                var source = new RgbaVector(float.Epsilon, 2 * float.Epsilon, float.MaxValue, float.MinValue);
```

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Does sizeof work here? Does the compiler see that as a constant?

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Wasn’t aware of that. SIMD is still very much a Google event for me

src/ImageSharp/Color/Color.cs
antonfirsov: >  Does the compiler see that as a constant?

As far as I'm aware it should, since it's not a generic argument, however we need to mark the method unsafe too which is not in my recommendation.

src/ImageSharp/Color/Color.cs
antonfirsov: `object` is basically a pointer so it's unsafe convertible to `long` on 64 bits. We can check 4 `Color`s with bitwise masks, if all of them are NULL we can shuffle stuff so the `data` fields are together, then bulk convert them.

Super unsure if it's worth the trouble.

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Just remembered why I wrote it this way. `ToRgba64` doesn't exist on `IPixel<TSelf>`;

antonfirsov: > `ToRgba64` doesn't exist on `IPixel<TSelf>`;

I missed that. We can implement this as following:

```C#
public unsafe static Color FromPixel<TPixel>(TPixel pixel)
{
    // Avoid boxing in case we can convert to Rgba64 safely and efficently
    if (typeof(TPixel) == typeof(Rgba64))
    {
        Rgba64 p = (Rgba64)(object)pixel;
        return new(p);
    }
    else if (typeof(TPixel) == typeof(Rgb48))
    {
        Rgb48 p = (Rgb48)(object)pixel;
        return new(new Rgba64(p.R, p.G, p.B, ushort.MaxValue)); // or add a new Color(Rgb48) constructor
    }
    else if (Unsafe.SizeOf(TPixel) <= sizeof(Rgba32))
    {
        return new(pixel.ToRgba32());
    }
    else
    {
        return new(pixel);
    }
}
```

The JIT will optimize away the unnecessary branches and the boxing from the `(Rgba64)(object)pixel` conversion.

src/ImageSharp/Color/Color.cs
JimBobSquarePants: >Super unsure if it's worth the trouble.

Yeah, not attempting that just now. Not worth the trouble.

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Cool, will do.

src/ImageSharp/Color/Color.cs
```diff
@@ -91,7 +96,38 @@ private Color(byte r, byte g, byte b)
...
+        {
+            // Avoid boxing in case we can convert to Rgba64 safely and efficently
+            if (typeof(TPixel) == typeof(Rgba64))
+            {
+                return new((Rgba64)(object)pixel);
+            }
+            else if (typeof(TPixel) == typeof(Rgb48))
+            {
+                return new((Rgb48)(object)pixel);
+            }
```
antonfirsov, line 119: Forgot that we also need to special case L16:

```suggestion
            }
            else if (typeof(TPixel) == typeof(L16))
            {
                Rgb48 p = (L16)(object)pixel;
                return new(new Rgba64(p.PackedValue, p.PackedValue, p.PackedValue, ushort.MaxValue));
            }
```

src/ImageSharp/Color/Color.cs
JimBobSquarePants: La32 also.

tests/ImageSharp.Tests/Color/ColorTests.CastTo.cs
```diff
@@ -88,6 +88,19 @@ public void Bgr24()
...
+            {
+                var source = new RgbaVector(float.Epsilon, 2 * float.Epsilon, float.MaxValue, float.MinValue);
+
+                // Act:
+                var color = Color.FromPixel(source);
+
+                // Assert:
+                RgbaVector data = color.ToPixel<RgbaVector>();
+                Assert.Equal(source, data);
+            }
```
antonfirsov, line 103:
```suggestion
            public static TheoryData<IPixel> GenericPixel_Data = new TheoryData<IPixel>
            {
                { new new RgbaVector(float.Epsilon, 2 * float.Epsilon, float.MaxValue, float.MinValue) },
                { new Rgba64((ushort)1, (ushort)2, ushort.MaxValue, (ushort)(ushort.MaxValue - 1)) },
                { new Rgb48((ushort)1, (ushort)2, (ushort)(ushort.MaxValue - 1))) },
                { new La32((ushort)1, (ushort)(ushort.MaxValue - 1))) },
                { new L16((ushort)(ushort.MaxValue - 1))) },
                { new Rgba32(1, 2, 255, 254) },
            }

            [Theory]
            [MemberData(nameof(GenericPixel_Data))]
            public void GenericPixel<TPixel>(TPixel source)
                where TPixel : unmanaged, IPixel<TPixel>
            {
                // Act:
                var color = Color.FromPixel(source);

                // Assert:
                TPixel actual = color.ToPixel<TPixel>();
                Assert.Equal(source, actual);
            }
```

Haven't checked if it compiles ..

antonfirsov (approved):
Product code LGTM. Let's extend the coverage a bit, and it is good to merge!

We also need tests for the IPixel path for `Equality_WhenTrue` and `Equality_WhenFalse`.

tests/ImageSharp.Tests/Color/ColorTests.CastTo.cs
JimBobSquarePants: This is doing my head in. Tried to hack around it but not getting very far.
https://github.com/xunit/xunit/issues/2061

JimBobSquarePants: Hacked it, those implicit casts are a nightmare.

calrsom:
Thanks to everyone who helped with this!

PR #1800, opened 2021-10-31
Webp: Use byte arrays instead of Dictionary's for lookups

brianpopow:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

During profiling I noticed alot of time with lossy decoding is spend in System.Collections. Turns out using Dictionary's for lookup tables was not the best idea. I have changed that now to byte array's. it reduces the lossy decoding time from **645.7 ms** to **280.8 ms**.

Before:
```
|                     Method |        Job |              Runtime |             Arguments |        TestImageLossy |        TestImageLossless |       Mean |     Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------- |----------- |--------------------- |---------------------- |---------------------- |------------------------- |-----------:|----------:|--------:|------:|------:|------:|----------:|
|        'Magick Lossy Webp' | Job-OJJMOX |             .NET 5.0 | /p:DebugType=portable | Webp/earth_lossy.webp | Webp/earth_lossless.webp |   106.3 ms |   8.05 ms | 0.44 ms |     - |     - |     - |     25 KB |
|    'ImageSharp Lossy Webp' | Job-OJJMOX |             .NET 5.0 | /p:DebugType=portable | Webp/earth_lossy.webp | Webp/earth_lossless.webp |   645.7 ms | 148.77 ms | 8.15 ms |     - |     - |     - |  2,430 KB |
|     'Magick Lossless Webp' | Job-OJJMOX |             .NET 5.0 | /p:DebugType=portable | Webp/earth_lossy.webp | Webp/earth_lossless.webp |   107.3 ms |   4.08 ms | 0.22 ms |     - |     - |     - |     16 KB |
| 'ImageSharp Lossless Webp' | Job-OJJMOX |             .NET 5.0 | /p:DebugType=portable | Webp/earth_lossy.webp | Webp/earth_lossless.webp |   303.3 ms |  42.16 ms | 2.31 ms |     - |     - |     - |  2,092 KB |
[...]

PR #1803, opened 2021-11-01
Use helper methods to access clip tables

brianpopow:
### Prerequisites

### Description

This is a follow up PR to #1800: As suggested by @antonfirsov, the clip tables are now accessed via methods.

Also adds some minor improvements to `Vp8LBitReader`


src/ImageSharp/Formats/Webp/WebpLookupTables.cs
```diff
@@ -1438,6 +1241,216 @@ static WebpLookupTables()
             InitializeFixedCostsI4();
         }
 
+        private static readonly byte[] Abs0Table =
```
gfoidl, line 1244: Make these use the ReadOnlySpan-trick?
Cf. https://github.com/SixLabors/ImageSharp/pull/1799#discussion_r740109460

It's already used in this class e.g. in `LogTable8Bit` (above).

brianpopow: Before I make this change, i would like to ask for other opinions on that matter. Should we really go all in with that trick and change all `byte`,  `sbyte` arrays to ReadOnlySpan here?

JimBobSquarePants: I would say yes.

PR #1804, opened 2021-11-01
Add sse2 version of select

brianpopow:
### Prerequisites

### Description

This adds a SSE2 version of the method Select(), which is used during lossless encoding. Its a bit faster, but its very specific to the image which is encoded, since this only gets used in `Predictor11`

TODO:
- [x] Add tests

Before:
select

After with sse2:
select_sse

### DIFF
--- a/src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
+++ b/src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
@@ -27,10 +27,6 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossless
 
         private const double Log2Reciprocal = 1.44269504088896338700465094007086;
 
-#if SUPPORTS_RUNTIME_INTRINSICS
-        private static readonly Vector128<byte> Zero = Vector128.Create(0).AsByte();
-#endif
-
         /// <summary>
         /// Returns the exact index where array1 and array2 are different. For an index
         /// inferior or equal to bestLenMatch, the return value just has to be strictly
@@ -1262,8 +1258,8 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossless
                     Vector128<byte> cb0 = Sse2.SubtractSaturate(c0, b0);
                     Vector128<byte> ac = Sse2.Or(ac0, ca0);
                     Vector128<byte> bc = Sse2.Or(bc0, cb0);
-                    Vector128<byte> pa = Sse2.UnpackLow(ac, Zero); // |a - c|
-                    Vector128<byte> pb = Sse2.UnpackLow(bc, Zero); // |b - c|
+                    Vector128<byte> pa = Sse2.UnpackLow(ac, Vector128<byte>.Zero); // |a - c|
+                    Vector128<byte> pb = Sse2.UnpackLow(bc, Vector128<byte>.Zero); // |b - c|
                     Vector128<ushort> diff = Sse2.Subtract(pb.AsUInt16(), pa.AsUInt16());
                     Sse2.Store((ushort*)p, diff);
                 }
--- a/src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
+++ b/src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
@@ -27,6 +27,10 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossless
 
         private const double Log2Reciprocal = 1.44269504088896338700465094007086;
 
+#if SUPPORTS_RUNTIME_INTRINSICS
+        private static readonly Vector128<byte> Zero = Vector128.Create(0).AsByte();
+#endif
+
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
```diff
@@ -1240,14 +1245,43 @@ private static uint ClampedAddSubtractHalf(uint c0, uint c1, uint c2)
...
+                    Vector128<byte> a0 = Sse2.ConvertScalarToVector128UInt32(a).AsByte();
+                    Vector128<byte> b0 = Sse2.ConvertScalarToVector128UInt32(b).AsByte();
+                    Vector128<byte> c0 = Sse2.ConvertScalarToVector128UInt32(c).AsByte();
+                    Vector128<byte> ac0 = Sse2.SubtractSaturate(a0, c0);
+                    Vector128<byte> ca0 = Sse2.SubtractSaturate(c0, a0);
+                    Vector128<byte> bc0 = Sse2.SubtractSaturate(b0, c0);
+                    Vector128<byte> cb0 = Sse2.SubtractSaturate(c0, b0);
+                    Vector128<byte> ac = Sse2.Or(ac0, ca0);
+                    Vector128<byte> bc = Sse2.Or(bc0, cb0);
+                    Vector128<byte> pa = Sse2.UnpackLow(ac, Zero); // |a - c|
```
br3aker, line 1265: While vector creation on pre .net6 is compiled to a really bad sequence of scalar sets, explicit `Vector128<T>.Zero` [compiles to a xor command ](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8ovLrl5hcAbho1iAZiakGAYQYBvGg11NlANRhgM0cqQAcAHgBmAGwjYMAPgYBZABQBKBgF5XRiZmlrYOTs4sAFowUBBy1AC+QA==)which would be a little clearer imo. Plus it might be a very tiny bit faster than static variable read.

brianpopow: thanks @br3aker

JimBobSquarePants (approved):
LGTM 👍

src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
gfoidl: FYI: `default` for the vector creates the same code as [`Vector128<T>.Zero`](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8ovLrl5hcAbho1iAZiakGAYQYBvGg11NlANRhgM0cqQAcAHmABPDDAB8DALIAKAJQMAvM6MmzSxt7JxYALRgoCDlqPX0Gf1Moc2s7B2cXck8fZwATGAAzbA4AGwwYgF8gA==)

(I prefer Vector128<T>.Zero).

src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
```diff
@@ -1240,14 +1241,43 @@ private static uint ClampedAddSubtractHalf(uint c0, uint c1, uint c2)
...
+                    Vector128<byte> cb0 = Sse2.SubtractSaturate(c0, b0);
+                    Vector128<byte> ac = Sse2.Or(ac0, ca0);
+                    Vector128<byte> bc = Sse2.Or(bc0, cb0);
+                    Vector128<byte> pa = Sse2.UnpackLow(ac, Vector128<byte>.Zero); // |a - c|
+                    Vector128<byte> pb = Sse2.UnpackLow(bc, Vector128<byte>.Zero); // |b - c|
+                    Vector128<ushort> diff = Sse2.Subtract(pb.AsUInt16(), pa.AsUInt16());
+                    Sse2.Store((ushort*)p, diff);
+                }
+
+                int paMinusPb = output[0] + output[1] + output[2] + output[3];
```
gfoidl, line 1267: You can put this into the `fixed`-block and access it via the pointer to avoid bound checks*.
If `output` would be too small, then there's a bug somewhere else 😉 (fortunately there's none).

\* or reverse the order to read `output[3]` first, then `[2]`, ... thant there's only one bound-check.

gfoidl: Ahhh, too late....

brianpopow: > or reverse the order to read output[3] first, then [2], ... thant there's only one bound-check.

ah yeah, always forget about that trick, thx. Will do with a follow up PR.

gfoidl: I'll put it (incl. the return) within the fixed block here.

PR #1810, opened 2021-11-07
SSE41 version of Hadamard transform

brianpopow:
### Prerequisites

### Description

This adds a SSE41 version of the Hadamard transform which is used during lossy webp encoding.
Relates to #1786

Profiling results:

Before:
hadamard

After:
hadamard_see

**note the number of calls is halfed, because the sse version does two transformas at once**


antonfirsov (approved):
LGTM.

PR #1814, opened 2021-11-07
Add SSE2 version of Mean16x4

brianpopow:
### Prerequisites

### Description
 This PR add SSE2 version of Mean16x4 which is used during lossy encoding with mode 1.
Related to #1786

### Profile Results

Before:
Mean

After:
mean_sse

The overall encoding time was 9002 ms. 
Its an improvement, but only a small one. I guess every bit counts.

### DIFF
--- a/src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
+++ b/src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
@@ -13,7 +13,7 @@ using System.Runtime.Intrinsics.X86;
 // ReSharper disable InconsistentNaming
 namespace SixLabors.ImageSharp.Formats.Webp.Lossy
 {
-    internal static unsafe class LossyUtils
+    internal static class LossyUtils
     {
 #if SUPPORTS_RUNTIME_INTRINSICS
         private static readonly Vector128<byte> Mean16x4Mask = Vector128.Create((short)0x00ff).AsByte();
@@ -967,11 +967,10 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossy
                 Vector128<int> e1 = Sse2.Add(d2, d3);
                 Vector128<int> f0 = Sse2.Add(e0, e1);
                 Vector128<short> hadd = Ssse3.HorizontalAdd(f0.AsInt16(), f0.AsInt16());
+                Vector128<uint> wide = Sse2.UnpackLow(hadd, Vector128<short>.Zero).AsUInt32();
 
-                dc[3] = (uint)hadd.GetElement(3);
-                dc[2] = (uint)hadd.GetElement(2);
-                dc[1] = (uint)hadd.GetElement(1);
-                dc[0] = (uint)hadd.GetElement(0);
+                ref uint outputRef = ref MemoryMarshal.GetReference(dc);
+                Unsafe.As<uint, Vector128<uint>>(ref outputRef) = wide;
             }
             else
 #endif
--- a/src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
+++ b/src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
@@ -4,12 +4,20 @@
 using System;
 using System.Buffers.Binary;
 using System.Runtime.CompilerServices;
+#if SUPPORTS_RUNTIME_INTRINSICS
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.X86;
+#endif
 
 // ReSharper disable InconsistentNaming
 namespace SixLabors.ImageSharp.Formats.Webp.Lossy
 {
-    internal static class LossyUtils
+    internal static unsafe class LossyUtils
     {
+#if SUPPORTS_RUNTIME_INTRINSICS
+        private static readonly Vector128<byte> Mean16x4Mask = Vector128.Create(0x00ff).AsByte();
+#endif
+
         [MethodImpl(InliningOptions.ShortMethod)]
         public static int Vp8Sse16X16(Span<byte> a, Span<byte> b) => GetSse(a, b, 16, 16);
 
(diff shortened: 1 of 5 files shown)


src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
```diff
@@ -801,26 +809,63 @@ public static void HFilter8i(Span<byte> u, Span<byte> v, int offset, int stride,
...
-
-        [MethodImpl(InliningOptions.ShortMethod)]
-        public static int YuvToB(int y, int u) => Clip8(MultHi(y, 19077) + MultHi(u, 33050) - 17685);
+#if SUPPORTS_RUNTIME_INTRINSICS
+            if (Sse2.IsSupported)
+            {
+#pragma warning disable SA1503 // Braces should not be omitted
+                tmp.Clear();
+                fixed (byte* inputPtr = input)
+                fixed (ushort* tmpPtr = tmp)
```
antonfirsov, line 820: Same recommendations to avoid pinning. Alternatively you can pin `YuvIn` before the `for (k = 0; k < 16; k += 4)` loop, and pass pointers to the method.

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
```diff
@@ -938,26 +942,58 @@ public static void HFilter8i(Span<byte> u, Span<byte> v, int offset, int stride,
...
+                Vector128<int> d1 = Sse2.Add(b1.AsInt32(), c1.AsInt32());
+                Vector128<int> d2 = Sse2.Add(b2.AsInt32(), c2.AsInt32());
+                Vector128<int> d3 = Sse2.Add(b3.AsInt32(), c3.AsInt32());
+                Vector128<int> e0 = Sse2.Add(d0, d1);
+                Vector128<int> e1 = Sse2.Add(d2, d3);
+                Vector128<int> f0 = Sse2.Add(e0, e1);
+                ref ushort outputRef = ref MemoryMarshal.GetReference(tmp);
+                Unsafe.As<ushort, Vector128<ushort>>(ref outputRef) = f0.AsUInt16();
+
+                dc[0] = (uint)(tmp[1] + tmp[0]);
```
JimBobSquarePants, line 973: It looks to me like if you reverse these span assignments you'll cut out 3 of 4 bounds checks.

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
```diff
@@ -938,26 +942,58 @@ public static void HFilter8i(Span<byte> u, Span<byte> v, int offset, int stride,
...
+                dc[1] = (uint)(tmp[3] + tmp[2]);
+                dc[2] = (uint)(tmp[5] + tmp[4]);
+                dc[3] = (uint)(tmp[7] + tmp[6]);
```
antonfirsov, line 976: Isn't this the same as  [`_mm_hadd_epi16`](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_hadd_epi16&ig_expand=3805) aka. `Ssse3.HorizontalAdd`?

I'm afraid 12 span indexer bound checks have measureable impact here. All of them seem unnecessary, since is `tmp` is always of 16 size and `dc` is always of 4 size. If we can't find any matching HorizontalAdd for this, maybe we should consider passing `tmp` as a pointer and and indexing `dc` with `Unsafe.*` stuff.

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
```diff
@@ -938,26 +942,58 @@ public static void HFilter8i(Span<byte> u, Span<byte> v, int offset, int stride,
...
-
-        [MethodImpl(InliningOptions.ShortMethod)]
-        public static int YuvToB(int y, int u) => Clip8(MultHi(y, 19077) + MultHi(u, 33050) - 17685);
-
-        [MethodImpl(InliningOptions.ShortMethod)]
-        public static int YuvToG(int y, int u, int v) => Clip8(MultHi(y, 19077) - MultHi(u, 6419) - MultHi(v, 13320) + 8708);
+#if SUPPORTS_RUNTIME_INTRINSICS
+            if (Sse2.IsSupported)
+            {
+                tmp.Clear();
```
antonfirsov, line 950: Is this really needed? We override the contents in the end.

brianpopow: yeah i think you are right, its not needed

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
brianpopow: yes it is the same as `Ssse3.HorizontalAdd`, good catch

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
```diff
@@ -938,26 +942,57 @@ public static void HFilter8i(Span<byte> u, Span<byte> v, int offset, int stride,
...
+                Vector128<int> e0 = Sse2.Add(d0, d1);
+                Vector128<int> e1 = Sse2.Add(d2, d3);
+                Vector128<int> f0 = Sse2.Add(e0, e1);
+                Vector128<short> hadd = Ssse3.HorizontalAdd(f0.AsInt16(), f0.AsInt16());
+                Vector64<short> lower = hadd.GetLower();
+
+                dc[0] = (uint)lower.GetElement(0);
+                dc[1] = (uint)lower.GetElement(1);
+                dc[2] = (uint)lower.GetElement(2);
+                dc[3] = (uint)lower.GetElement(3);
```
JimBobSquarePants, line 975:
```suggestion
                dc[3] = (uint)lower.GetElement(3);
                dc[2] = (uint)lower.GetElement(2);
                dc[1] = (uint)lower.GetElement(1);
                dc[0] = (uint)lower.GetElement(0);
```

JimBobSquarePants (approved):
LGTM 👍

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
```diff
@@ -938,26 +942,57 @@ public static void HFilter8i(Span<byte> u, Span<byte> v, int offset, int stride,
...
+                Vector128<int> e0 = Sse2.Add(d0, d1);
+                Vector128<int> e1 = Sse2.Add(d2, d3);
+                Vector128<int> f0 = Sse2.Add(e0, e1);
+                Vector128<short> hadd = Ssse3.HorizontalAdd(f0.AsInt16(), f0.AsInt16());
+                Vector64<short> lower = hadd.GetLower();
+
+                dc[3] = (uint)lower.GetElement(3);
+                dc[2] = (uint)lower.GetElement(2);
+                dc[1] = (uint)lower.GetElement(1);
+                dc[0] = (uint)lower.GetElement(0);
```
antonfirsov, line 975: I'm not sure what does `GetElement` compile to but it can be terribly inefficient. `dc[0...3]` is 64 bits if I'm not mistaken:

```suggestion
                Unsafe.As<uint, Vector64<short>>(ref MemoryMarshal.GetReference(dc)) = lower;
```

antonfirsov: No sorry, it's actually 128 bits. It needs some widening with intrinsics then.

brianpopow: I think this GetLower is not needed at all, now that i think again about it.

What about this: 1452ba00

brianpopow: @antonfirsov so you think we should try to avoid `GetElement` entirely?

antonfirsov: I think it!s a good practice to avoid it whenever possible. It looks like it compiles to [VPEXTRW](https://www.felixcloutier.com/x86/pextrw) extracting scalar registers from the SIMD register, but then you do scalar conversion for each value. We can get much better results with shuffling:
[SHARPLAB](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwEYCwAUJgAwAEmuAdAEoCuMCUYAplQMID2YADlADYsATgGVhANygBjFigDcRUhWr1GzNgElGwzjzFDJM+YvKVaDJqypaEQ2CmnHiplRfXXGdmA6koqADQAOADYFQkUAZmVgimwydjIAbyIyVOiyBhQIADMWCgAWMgBxFgQAUUFWRgAKADUWKQROIVxsQIAeFAALZoQAPjJxOAzYBAAqMiEAShS05MI0xcmAbQiAXTIAXjJqulGp8SoS8sqWGuwpsKW0oWXsDe3d/cPjipYqhGqLq+uV3AedntGAcjqU3h9qrhLrNrrcSACnsCXmDTjUSNCFmkAL4wsi4ygxTI5PKYQqvVEIAD6Ih4EBgdQaTRabU6PSE/UGwxpdPaQI501x81+K3WW0Bz1BJ3eZ0+31xi1u9zFiIQIPJ0vOGOFt3+yr5apRGs+UJ+sOW8L1EvVEPRprIOPCmNSBIy3mJBTIIiZLAZjWarQ63V6AyGI0YEwFTqS8rS9T9zI6fIGAHcoAATPLbEQoFjYKgAVRgtKkAGsADKcZPVUNxpkB1nBqgALR0UyoAEEUPmbBFsNUtb9s7mqF7mj6hMNUxmB6kHUR8bgYqTPd7qbT6bX/Syg+yQ1z17zRgNI4shb9NwnD4wU+nM56c3nC8Xy5Xq8ML/Wd/1m62O12e32M5LIWWS5H+V4IO+jJbomR59NUQgsNkZAALLvM0ACeKEQEI3QQPwko0EhwhnDICFTFMYpTiwdoOliQA==)

I think we are looking for [_mm_unpacklo_epi16](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_unpacklo_epi16&ig_expand=732) / `Sse2.UnpackLow` interleaving `lower` with zeros to get equivalent `uint` representation, but not sure how would this act with negative `short`-s. Can `hadd` end up having negative values?

```C#
Sse2.UnpackLow(lower, Vector128<short>.Zero)
```

brianpopow: That looks indeed much better. I dont think there can be negative values.

Thanks for your help!

brianpopow: using `Sse2.Store` looks even a tiny bit better

antonfirsov: `dc` is stackallocked, so you can pass it as a pointer easily if you want to, though I'm not sure if it's worth the trouble.
The main benefit would be the removal of one `Slice()` on [the call site](https://github.com/SixLabors/ImageSharp/blob/7d8225b59a633b08b51e74bbb960d4d52b420a84/src/ImageSharp/Formats/Webp/Lossy/Vp8EncIterator.cs#L366), the disadvantage is the inconsistency of the method signature (`input` is a span, `dc` is a pointer).

brianpopow: I think its not worth it, lets keep the method signature consistent and keep it as it is.

PR #1817, opened 2021-11-09
Add SSE2 version of Vp8Sse4X4

brianpopow:
### Prerequisites

### Description

This adds a SSE2 version of the method `Vp8Sse4X4`, which is used during lossy encoding of webp images.
Related to #1786

#### Master Branch:
Vp8Sse4X4_master

#### PR:
Vp8Sse4X4_sse

P.S. im not sure again why the call count is different, it was the same image

### DIFF
--- a/src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
+++ b/src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
@@ -19,14 +19,17 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossy
         private static readonly Vector128<byte> Mean16x4Mask = Vector128.Create((short)0x00ff).AsByte();
 #endif
 
+        // Note: method name in libwebp reference implementation is called VP8SSE16x16.
         [MethodImpl(InliningOptions.ShortMethod)]
-        public static int Vp8Sse16X16(Span<byte> a, Span<byte> b) => GetSse(a, b, 16, 16);
+        public static int Vp8_Sse16X16(Span<byte> a, Span<byte> b) => Vp8_SseNxN(a, b, 16, 16);
 
+        // Note: method name in libwebp reference implementation is called VP8SSE16x8.
         [MethodImpl(InliningOptions.ShortMethod)]
-        public static int Vp8Sse16X8(Span<byte> a, Span<byte> b) => GetSse(a, b, 16, 8);
+        public static int Vp8_Sse16X8(Span<byte> a, Span<byte> b) => Vp8_SseNxN(a, b, 16, 8);
 
+        // Note: method name in libwebp reference implementation is called VP8SSE4x4.
         [MethodImpl(InliningOptions.ShortMethod)]
-        public static int Vp8Sse4X4(Span<byte> a, Span<byte> b)
+        public static int Vp8_Sse4X4(Span<byte> a, Span<byte> b)
         {
 #if SUPPORTS_RUNTIME_INTRINSICS
             if (Sse2.IsSupported)
@@ -67,12 +70,12 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossy
             else
 #endif
             {
-                return GetSse(a, b, 4, 4);
+                return Vp8_SseNxN(a, b, 4, 4);
             }
         }
 
         [MethodImpl(InliningOptions.ShortMethod)]
-        public static int GetSse(Span<byte> a, Span<byte> b, int w, int h)
+        public static int Vp8_SseNxN(Span<byte> a, Span<byte> b, int w, int h)
         {
             int count = 0;
             int aOffset = 0;
--- a/src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
+++ b/src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
@@ -28,14 +28,16 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossy
             if (Sse2.IsSupported)
             {
                 // Load values.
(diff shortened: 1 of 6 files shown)


src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
```diff
@@ -22,7 +22,48 @@ internal static unsafe class LossyUtils
...
+        public static int Vp8Sse4X4(Span<byte> a, Span<byte> b)
+        {
+#if SUPPORTS_RUNTIME_INTRINSICS
+            if (Sse2.IsSupported)
+            {
+                // Load values.
+                Vector128<byte> a0 = Unsafe.As<byte, Vector128<byte>>(ref MemoryMarshal.GetReference(a));
+                Vector128<byte> a1 = Unsafe.As<byte, Vector128<byte>>(ref MemoryMarshal.GetReference(a.Slice(WebpConstants.Bps, 8)));
+                Vector128<byte> a2 = Unsafe.As<byte, Vector128<byte>>(ref MemoryMarshal.GetReference(a.Slice(WebpConstants.Bps * 2, 8)));
+                Vector128<byte> a3 = Unsafe.As<byte, Vector128<byte>>(ref MemoryMarshal.GetReference(a.Slice(WebpConstants.Bps * 3, 8)));
```
antonfirsov, line 34:
```suggestion
                ref byte aRef = ref MemoryMarshal.GetReference(a);
                Vector128<byte> a0 = Unsafe.As<byte, Vector128<byte>>(ref aRef);
                Vector128<byte> a1 = Unsafe.As<byte, Vector128<byte>>(ref Unsafe.Add(ref aRef, WebpConstants.Bps));
                Vector128<byte> a2 = Unsafe.As<byte, Vector128<byte>>(ref Unsafe.Add(ref aRef, WebpConstants.Bps * 2));
                Vector128<byte> a3 = Unsafe.As<byte, Vector128<byte>>(ref Unsafe.Add(ref aRef, WebpConstants.Bps * 3));
```

Same for b.

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
```diff
@@ -26,7 +26,50 @@ internal static class LossyUtils
...
+            }
+            else
+#endif
+            {
+                return GetSse(a, b, 4, 4);
+            }
+        }
 
         [MethodImpl(InliningOptions.ShortMethod)]
         public static int GetSse(Span<byte> a, Span<byte> b, int w, int h)
```
JimBobSquarePants, line 75: This method name confuses me shouldn't it be `Vp84X4`?

antonfirsov: `Vp84X4` is confusing because of `84`. Parent method can be `Vp8_4x4` while this one can be `Vp8_4X4Scalar`. Or maybe `Vp8_4x4` could be inlined if it's not used anywhere else?

src/ImageSharp/Formats/Webp/Lossy/Vp8ModeScore.cs
```diff
@@ -97,18 +97,11 @@ public Vp8ModeScore()
...
-                for (int j = 0; j < 3; j++)
-                {
-                    this.Derr[i, j] = 0;
-                }
-            }
+            Array.Clear(this.YDcLevels, 0, this.YDcLevels.Length);
+            Array.Clear(this.YAcLevels, 0, this.YAcLevels.Length);
+            Array.Clear(this.UvLevels, 0, this.UvLevels.Length);
+            Array.Clear(this.ModesI4, 0, this.ModesI4.Length);
+            Array.Clear(this.Derr, 0, this.Derr.Length);
```
antonfirsov, line 104: Unrelated to the main thing in the PR, but `Vp8ModeScore` is also a good candidate to be made a value type with fixed buffers.

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
brianpopow: I agree that the name is not ideal, but its called `VP8SSE4x4` in the original libwebp source and I wanted to keep the method names to be as close to the original as possible to make it easier to follow/commpare the original code.
A better name might have been `CalculateDistortion4x4`, but again I am still in favor of keeping the names similar to the original.

src/ImageSharp/Formats/Webp/Lossy/Vp8ModeScore.cs
brianpopow: yes seems reasonable, `Vp8ModeScore` makes up the most of the allocations during encoding, but I might try that in a different PR.

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
brianpopow: > This method name confuses me shouldn't it be `Vp84X4`?

`GetSse` is not called `Vp84X4` because of the variable parameters w and h.

I tried to rename the methods: 7e20c5da
Is that better (or worse)?

JimBobSquarePants: Much better yes 👍

src/ImageSharp/Formats/Webp/Lossy/Vp8ModeScore.cs
JimBobSquarePants: Definitely worth a look after.

JimBobSquarePants (approved):
LGTM so far.

PR #1818, opened 2021-11-10
Hoist some of the calculations from loops of 3DMoments()

kunalspathak:
### Prerequisites

### Description

The 4-level nested for loop in 3DMoments is very expensive, the inner most loop executes almost 1.1M times. We call `GetPaletteIndex()` method that does bunch of calculation using index variables of all the 4-loops. 

https://github.com/SixLabors/ImageSharp/blob/255226b03c8350f88e641bdb58c9450b68729ef7/src/ImageSharp/Processing/Processors/Quantization/WuQuantizer%7BTPixel%7D.cs#L418-L444

There is a potential to hoist some of the invariants  calculation outside the loop. I verified the benchmark and see around 2-3% improvement.

Here is the assembly difference: https://www.diffchecker.com/I4uqCtBO


JimBobSquarePants:
@kunalspathak Thanks for this it is very interesting. I've just read your linked runtime issue also. 

If we can add some supplementary comments to explain the optimization and refactor some of the variable naming (prefer ` ind1R` over ` ind1_r` ) then that would be a great addition to the codebase! 👍

kunalspathak:
I have also noticed other places where similar technique could be use although I didn't measure its performance. 
For eg, if I just search for all the references of `GetPaletteIndex`: https://github.com/SixLabors/ImageSharp/blob/7d74c4c95a8fb9c2df614c9dab708c4cf4b9181c/src/ImageSharp/Processing/Processors/Quantization/WuQuantizer%7BTPixel%7D.cs#L217 , once of the reference is : 
https://github.com/SixLabors/ImageSharp/blob/7d74c4c95a8fb9c2df614c9dab708c4cf4b9181c/src/ImageSharp/Processing/Processors/Quantization/WuQuantizer%7BTPixel%7D.cs#L233-L249

Here, `r` related calculation for `cube.RMax` and `cube.RMin` can be precalculated and reused.

Another interesting method is `Mark()`. What is the common range of `RMin ~ RMax`, `GMin ~ GMax`, etc. If it is `>= 32`, we are already calculating `GetPaletteIndex()` million times and some of the calculation can be hoisted out the way its been done in this PR. Let me know and I can do for `Mark()` method as well.

https://github.com/SixLabors/ImageSharp/blob/7d74c4c95a8fb9c2df614c9dab708c4cf4b9181c/src/ImageSharp/Processing/Processors/Quantization/WuQuantizer%7BTPixel%7D.cs#L629-L641

tests/ImageSharp.Benchmarks/Codecs/EncodeIndexedPng.cs
```diff
@@ -85,7 +85,7 @@ public void PngCoreWu()
         public void PngCoreWuNoDither()
         {
             using var memoryStream = new MemoryStream();
-            var options = new PngEncoder { Quantizer = new WuQuantizer(new QuantizerOptions { Dither = null }) };
+            var options = new PngEncoder { Quantizer = new WuQuantizer(new QuantizerOptions { Dither = null }), ColorType = PngColorType.Palette };
```
kunalspathak, line 88: I made this change so I can get perf coverage for it. Should I keep it or revert it?

JimBobSquarePants: Yep. That's an oversight from us. 

Very happy to see what other optimizations you can bring as long as they are well commented. Thanks! 👍

kunalspathak: Sure, happy to contribute. I think it will be more productive for us to contribute or investigate the performance issues if these benchmarks are in our system. Any progress on https://github.com/SixLabors/ImageSharp/discussions/1795?

kunalspathak:
This should be ready for review now.

tests/ImageSharp.Benchmarks/Codecs/EncodeIndexedPng.cs
JimBobSquarePants: Leaving that one for @antonfirsov Likely after we complete #1730 and release V2.

JimBobSquarePants (approved):
LGTM, thanks! 👍

tests/ImageSharp.Benchmarks/Codecs/EncodeIndexedPng.cs
antonfirsov: Answered:
https://github.com/SixLabors/ImageSharp/discussions/1795#discussioncomment-1635798

kunalspathak:
This workaround will not be needed anymore because codegen will take care of it as part of https://github.com/dotnet/runtime/pull/68061. Here is the diff for the earlier version of C# code:

image

Full diff: https://www.diffchecker.com/DKGAZKq0

JimBobSquarePants:
That’s .NET 7 though yeah? We’ll be sticking to LTS release targets from now on so just 6 for now. 

I am planning on revising many parts of the codebase though to simplify things where possible now we have a much more modern single target.

kunalspathak:
> That’s .NET 7 though yeah? 

Yes.

> We’ll be sticking to LTS release targets from now on so just 6 for now.

Sure, this was just FYI :)

JimBobSquarePants:
It’s really good to know!!

PR #1819, opened 2021-11-10
Add sse2 version of inverse transform

brianpopow:
### Prerequisites

### Description

This PR introduces a SSE2 version of the method `ITransform`, which is used during lossy webp encoding.
Related to #1786

The profiling results look good:

### master branch:
ITransform

### PR
ITransform_sse

Note: the different call counts are due to the SSE version is doing two transforms at once.


src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
```diff
@@ -68,12 +82,197 @@ static Vp8Encoding()
...
+                var in0 = Vector128.Create(Unsafe.As<short, long>(ref inputRef), 0);
+                var in1 = Vector128.Create(Unsafe.As<short, long>(ref Unsafe.Add(ref inputRef, 4)), 0);
+                var in2 = Vector128.Create(Unsafe.As<short, long>(ref Unsafe.Add(ref inputRef, 8)), 0);
+                var in3 = Vector128.Create(Unsafe.As<short, long>(ref Unsafe.Add(ref inputRef, 12)), 0);
+
+                // a00 a10 a20 a30   x x x x
+                // a01 a11 a21 a31   x x x x
+                // a02 a12 a22 a32   x x x x
+                // a03 a13 a23 a33   x x x x
+                if (doTwo)
```
antonfirsov, line 123: It's better to avoid giving unnecessary extra work to the branch predictor, and refactor `ITransform` into two separate methods.

One hidden cost is that assembly code having branches on a runtime constant will be longer, and needs more cache lines, which are expensive to load.

Alternative "fun" solution, sort of template metaprogramming in C#, that will eliminate branches at JIT time:

```C#
ITransform<TDoTwo>(...) {
    ...
    if (typeof(TDoTwo) == typeof(DoTwo_True)) {
        ...
    }
    else {
       ...
    }
}
```

Can't decide if it's worth it or better to just duplicate the code, and maybe create reusable helper methods for shared bits.

src/ImageSharp/Formats/Webp/Lossy/LossyUtils.cs
```diff
@@ -769,6 +748,43 @@ public static int TTransformSse41(Span<byte> inputA, Span<byte> inputB, Span<ush
 
             return Numerics.ReduceSum(result);
         }
+
+        // Transpose two 4x4 16b matrices horizontally stored in registers.
+        public static void Vp8Transpose_2_4x4_16b(Vector128<short> b0, Vector128<short> b1, Vector128<short> b2, Vector128<short> b3, out Vector128<long> output0, out Vector128<long> output1, out Vector128<long> output2, out Vector128<long> output3)
```
antonfirsov, line 753:
```suggestion
        [MethodImpl(InliningOptions.ShortMethod)]
        public static void Vp8Transpose_2_4x4_16b(Vector128<short> b0, Vector128<short> b1, Vector128<short> b2, Vector128<short> b3, out Vector128<long> output0, out Vector128<long> output1, out Vector128<long> output2, out Vector128<long> output3)
```

src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
```diff
@@ -68,51 +81,299 @@ static Vp8Encoding()
...
+                ref2 = Sse2.PackUnsignedSaturate(ref2InvAdded, ref2InvAdded);
+                ref3 = Sse2.PackUnsignedSaturate(ref3InvAdded, ref3InvAdded);
+
+                // Unsigned saturate to 8b.
+                ref byte outputRef = ref MemoryMarshal.GetReference(dst);
+
+                // Store eight bytes/pixels per line.
+                Unsafe.As<byte, Vector64<byte>>(ref outputRef) = ref0.GetLower();
+                Unsafe.As<byte, Vector64<byte>>(ref Unsafe.Add(ref outputRef, WebpConstants.Bps)) = ref1.GetLower();
+                Unsafe.As<byte, Vector64<byte>>(ref Unsafe.Add(ref outputRef, WebpConstants.Bps * 2)) = ref2.GetLower();
```
antonfirsov, line 189: In the first 3 calls, it's possible and better to avoid `GetLower`. The second store will overwrite the upper 8 bit written in the first store and so on:

```suggestion
                Unsafe.As<byte, Vector128<byte>>(ref outputRef) = ref0;
                Unsafe.As<byte, Vector128<byte>>(ref Unsafe.Add(ref outputRef, WebpConstants.Bps)) = ref1;
                Unsafe.As<byte, Vector128<byte>>(ref Unsafe.Add(ref outputRef, WebpConstants.Bps * 2)) = ref2;
```

If you guarantee at the call site that `dst` has a patting of 8 bytes, you can also avoid it at the last call.

brianpopow: I dont think that will work. I only want to write 8 bytes from each ref vector. 
The first write will be at position 0, the second at 32 (note the Unsafe.Add() with `WebpConstants.Bps`, WebpConstants.Bps is 32), the third at 64 and the last at 96.

antonfirsov: Ah haven't noticed that `Bps` is 32, nevermind then.

antonfirsov (approved):
Looks good!

PR #1824, opened 2021-11-12
Add AVX version of CollectColorBlueTransforms and CollectColorRedTransforms

brianpopow:
### Prerequisites

### Description

This PR adds a AVX version of CollectColorBlueTransforms and CollectColorRedTransforms, which is used during encoding of lossless webp images.
Related to: #1786

TODO:
~- There are still 2 failing tests.~ Fixed by @JimBobSquarePants with 55f04f6

Profiling results with test image: `Images\Input\Webp\earth_lossless.webp` :

### Master
CollectColor

### PR
ColectColorAvx

### DIFF
--- a/src/ImageSharp/Formats/Webp/Lossless/ColorSpaceTransformUtils.cs
+++ b/src/ImageSharp/Formats/Webp/Lossless/ColorSpaceTransformUtils.cs
@@ -1,0 +1,268 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+#if SUPPORTS_RUNTIME_INTRINSICS
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.X86;
+#endif
+
+namespace SixLabors.ImageSharp.Formats.Webp.Lossless
+{
+    internal static class ColorSpaceTransformUtils
+    {
+#if SUPPORTS_RUNTIME_INTRINSICS
+        private static readonly Vector128<byte> CollectColorRedTransformsGreenMask = Vector128.Create(0x00ff00).AsByte();
+
+        private static readonly Vector128<byte> CollectColorRedTransformsAndMask = Vector128.Create((short)0xff).AsByte();
+
+        private static readonly Vector256<byte> CollectColorRedTransformsGreenMask256 = Vector256.Create(0x00ff00).AsByte();
+
+        private static readonly Vector256<byte> CollectColorRedTransformsAndMask256 = Vector256.Create((short)0xff).AsByte();
+
+        private static readonly Vector128<byte> CollectColorBlueTransformsGreenMask = Vector128.Create(0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255);
+
+        private static readonly Vector128<byte> CollectColorBlueTransformsGreenBlueMask = Vector128.Create(255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 0, 0);
+
+        private static readonly Vector128<byte> CollectColorBlueTransformsBlueMask = Vector128.Create(255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0);
+
+        private static readonly Vector128<byte> CollectColorBlueTransformsShuffleLowMask = Vector128.Create(255, 2, 255, 6, 255, 10, 255, 14, 255, 255, 255, 255, 255, 255, 255, 255);
+
+        private static readonly Vector128<byte> CollectColorBlueTransformsShuffleHighMask = Vector128.Create(255, 255, 255, 255, 255, 255, 255, 255, 255, 2, 255, 6, 255, 10, 255, 14);
+
+        private static readonly Vector256<byte> CollectColorBlueTransformsShuffleLowMask256 = Vector256.Create(255, 2, 255, 6, 255, 10, 255, 14, 255, 255, 255, 255, 255, 255, 255, 255, 255, 18, 255, 22, 255, 26, 255, 30, 255, 255, 255, 255, 255, 255, 255, 255);
+
+        private static readonly Vector256<byte> CollectColorBlueTransformsShuffleHighMask256 = Vector256.Create(255, 255, 255, 255, 255, 255, 255, 255, 255, 2, 255, 6, 255, 10, 255, 14, 255, 255, 255, 255, 255, 255, 255, 255, 255, 18, 255, 22, 255, 26, 255, 30);
+
+        private static readonly Vector256<byte> CollectColorBlueTransformsGreenBlueMask256 = Vector256.Create(255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 0, 0, 255, 255, 0, 0);
+
+        private static readonly Vector256<byte> CollectColorBlueTransformsBlueMask256 = Vector256.Create(255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0);
+
+        private static readonly Vector256<byte> CollectColorBlueTransformsGreenMask256 = Vector256.Create(0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255);
+#endif
+
+        public static void CollectColorBlueTransforms(Span<uint> bgra, int stride, int tileWidth, int tileHeight, int greenToBlue, int redToBlue, Span<int> histo)
+        {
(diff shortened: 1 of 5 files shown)


brianpopow:
The AVX version does not produce the same results as the SSE41 version. Can't tell why that is so far, both versions look the same.

JimBobSquarePants:
@brianpopow I’ll try and have a look later. Are the masks a byte? If so there’s a method in the source MMShuffle you could use to compare to.

brianpopow:
> @brianpopow I’ll try and have a look later. Are the masks a byte? If so there’s a method in the source MMShuffle you could use to compare to.

The masks are explicitly defined following the pattern from SSE41: [Mask](https://github.com/SixLabors/ImageSharp/pull/1824/files#diff-9729d8660a96ce14c5ad80cc189ae4f2c9ee1882d63a4ad242de201548a4d866R56)

Maybe it was naive from me to assume, I could just use the same pattern as in SSE41. This time there is no reference in `libwebp`, I could compare it to.

I was testing it with simple test pattern images, like only red, only blue, only green images and it seems to work, but not with more complex examples: See the failing .

antonfirsov:
@brianpopow my recommendation is to extract the AVX, the SSE and the scalar path to separate unit-testable methods and debug them side-by side to spot the issue.

JimBobSquarePants:
Are you sure your Sse version is correct?

image

brianpopow:
> Are you sure your Sse version is correct?
> 

I am pretty sure that the SSE version produces the same results as the none vectorized version `CollectColorBlueTransformsNoneVectorized`, yes.

Here is an example with all pixels red (with alpha at 255):
red_example

```
uint[] bgrRed =
{
    4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760,
    4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760,
    4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760,
    4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760, 4294901760,
    4294901760
};
```

JimBobSquarePants:
Ah right, my bad. My input was in byte ranges. Looks to me like your'e not crossing the boundaries properly. 

`CollectColorBlueTransformsShuffleLowMask256 ` should be something like.

```c#
private static readonly Vector256<byte> CollectColorBlueTransformsShuffleLowMask256 = Vector256.Create(255, 2, 255, 6, 255, 10, 255, 14, 255, 255, 255, 255, 255, 255, 255, 255, 255, 18, 255, 22, 255, 26, 255, 30, 255, 255, 255, 255, 255, 255, 255, 255);
```

JimBobSquarePants:
@brianpopow I updated the masks for you and everything is passing now. I won't approve/merge yet until you're happy that you've covered everything you need to.

brianpopow:
> @brianpopow I updated the masks for you and everything is passing now. I won't approve/merge yet until you're happy that you've covered everything you need to.

Awesome, thanks for your help!

I think Anton is right. The color transform methods are not easily unit testable. I want to do a favor to "future me" and separate them into a own class.

brianpopow:
I think this is now ready for a final review

src/ImageSharp/Formats/Webp/Lossless/PredictorEncoder.cs
```diff
@@ -572,19 +551,17 @@ private static byte NearLosslessComponent(byte value, byte predict, byte boundar
...
+            // upper is closer to residual than lower.
+            if (residual <= boundaryResidual && upper > boundaryResidual)
+            {
+                // Halve quantization step to avoid crossing boundary. This midpoint is
+                // on the same side of boundary as residual because midpoint <= residual
+                // (since upper is closer than lower) and residual is below the boundary.
+                return (byte)(lower + (quantization >> 1));
             }
+
+            return (byte)(upper & 0xff);
```
gfoidl, line 564:
```suggestion
            return (byte)upper;
```
With the cast to `byte` to mask isn't necessary. But modern JIT will optimize this away anyway. It's solely about less IL-size (if that counts).

src/ImageSharp/Formats/Webp/Lossless/ColorSpaceTransformUtils.cs
```diff
@@ -0,0 +1,268 @@
...
+            {
+                const int span = 16;
+                Span<ushort> values = stackalloc ushort[span];
+                var multsr = Vector256.Create(LosslessUtils.Cst5b(redToBlue));
+                var multsg = Vector256.Create(LosslessUtils.Cst5b(greenToBlue));
+                for (int y = 0; y < tileHeight; y++)
+                {
+                    Span<uint> srcSpan = bgra.Slice(y * stride);
+                    ref uint inputRef = ref MemoryMarshal.GetReference(srcSpan);
+                    for (int x = 0; x + span <= tileWidth; x += span)
```
gfoidl, line 59:
```suggestion
                    for (int x = 0; x <= tileWidth - span; x += span)
```
So `tileWidth - span` can be kept in a register, and doesn't need to be re-calculated for each iteration.

src/ImageSharp/Formats/Webp/Lossless/ColorSpaceTransformUtils.cs
```diff
@@ -0,0 +1,268 @@
...
+            {
+                const int span = 8;
+                Span<ushort> values = stackalloc ushort[span];
+                var multsr = Vector128.Create(LosslessUtils.Cst5b(redToBlue));
+                var multsg = Vector128.Create(LosslessUtils.Cst5b(greenToBlue));
+                for (int y = 0; y < tileHeight; y++)
+                {
+                    Span<uint> srcSpan = bgra.Slice(y * stride);
+                    ref uint inputRef = ref MemoryMarshal.GetReference(srcSpan);
+                    for (int x = 0; x + span <= tileWidth; x += span)
```
gfoidl, line 104:
```suggestion
                    for (int x = 0; x <= tileWidth - span; x += span)
```
Same here.

src/ImageSharp/Formats/Webp/Lossless/ColorSpaceTransformUtils.cs
```diff
@@ -0,0 +1,268 @@
...
+            if (Avx2.IsSupported && tileWidth >= 16)
+            {
+                var multsg = Vector256.Create(LosslessUtils.Cst5b(greenToRed));
+                const int span = 16;
+                Span<ushort> values = stackalloc ushort[span];
+                for (int y = 0; y < tileHeight; y++)
+                {
+                    Span<uint> srcSpan = bgra.Slice(y * stride);
+                    ref uint inputRef = ref MemoryMarshal.GetReference(srcSpan);
+                    for (int x = 0; x + span <= tileWidth; x += span)
```
gfoidl, line 173:
```suggestion
                    for (int x = 0; x <= tileWidth - span; x += span)
```

src/ImageSharp/Formats/Webp/Lossless/ColorSpaceTransformUtils.cs
```diff
@@ -0,0 +1,268 @@
...
+            else if (Sse41.IsSupported)
+            {
+                var multsg = Vector128.Create(LosslessUtils.Cst5b(greenToRed));
+                const int span = 8;
+                Span<ushort> values = stackalloc ushort[span];
+                for (int y = 0; y < tileHeight; y++)
+                {
+                    Span<uint> srcSpan = bgra.Slice(y * stride);
+                    ref uint inputRef = ref MemoryMarshal.GetReference(srcSpan);
+                    for (int x = 0; x + span <= tileWidth; x += span)
```
gfoidl, line 214:
```suggestion
                    for (int x = 0; x <= tileWidth - span; x += span)
```

src/ImageSharp/Formats/Webp/Lossless/ColorSpaceTransformUtils.cs
gfoidl: BTW: can the loop variable be of type `nint` to avoid the `movsxd` machine-instruction?

brianpopow: @gfoidl thanks again for your suggestions.

About the `nint`. I am not familiar with nint, so i dont know scenarios where it should be avoided to be used. I think that we can use it here, but i feel that i cannot really judge it.

JimBobSquarePants (approved):
BoratVeryGIF

src/ImageSharp/Formats/Webp/Lossless/ColorSpaceTransformUtils.cs
```diff
@@ -0,0 +1,268 @@
...
+                        {
+                            ++histo[values[i]];
+                        }
+                    }
+                }
+
+                int leftOver = tileWidth & (span - 1);
+                if (leftOver > 0)
+                {
+                    CollectColorBlueTransformsNoneVectorized(bgra.Slice(tileWidth - leftOver), stride, leftOver, tileHeight, greenToBlue, redToBlue, histo);
```
antonfirsov, line 136: Typo: `NoneVectorized` -> `NonVectorized`. Or even better: `CollectColorBlueTransformsScalar`.

Don't bother to raise a separate PR, just include it to some other one if you can.

brianpopow: sure, will do that.

JimBobSquarePants: I've already changed it in #1830

PR #1831, opened 2021-11-16
WIP: Add SSE2 versions of Average2/3/4

brianpopow:
### Prerequisites

### Description

I though it would be a good idea to use SSE2 in `Average2`, `Average2` and `Average3`, which is used `Predictor5` to `Predictor10`.  Those Predictor methods are used during lossless webp encoding and decoding.

Unfortunately the profiling results do not look good. Its actually slower then the none vectorized versions.
I think it could be due to `_mm_cvtsi32_si128` having a latency of 2 for skylake, according to the [intel docs](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_cvtsi32_si128&ig_expand=5074,2283,7353,92,349,2283), which is used often here.

Maybe this will look better on newer CPU's. Mine is rather old now. Its a `i7 6700K`
Also maybe its possible to avoid using `_mm_cvtsi32_si128` here and replace it with other commands.

Related to #1786

Profile results:

### Master:

Average3

Average4

### PR:

Average3_sse

Average4_sse

The profiling was done by decoding a lossless webp and then encode it again.
```
using var image = Image.Load(@"Calliphora_lossless_originalsize.webp");
image.Save("output.webp", new WebpEncoder() { FileFormat = WebpFileFormatType.Lossless });
```

The following image was used:
[Calliphora_lossless_originalsize.zip](https://github.com/SixLabors/ImageSharp/files/7547527/Calliphora_lossless_originalsize.zip)

### DIFF
--- a/src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
+++ b/src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
@@ -49,6 +49,8 @@ namespace SixLabors.ImageSharp.Formats.Webp.Lossless
         private static readonly Vector128<byte> TransformColorInverseAlphaGreenMask = Vector128.Create(0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255);
 
         private static readonly byte TransformColorInverseShuffleMask = SimdUtils.Shuffle.MmShuffle(2, 2, 0, 0);
+
+        private static readonly Vector128<byte> Ones = Vector128.Create((byte)1);
 #endif
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


antonfirsov:
> Unfortunately the profiling results do not look good. Its actually slower then the none vectorized versions.

Maybe `Average4` makes sense, but most of the SSE variants actually do extra work, the inputs & outputs are all scalar `uint` values, I don't see why/how could be the current approach faster.

What you could attempt is to implement batched variants for the `Predictor*` methods, something like `Predictor10(uint* left, uint *top, nint count)` and do SSE/AVX batches of 4/8 values walking the buffer, but that's going to be a major refactor looking at the calling code.

You will have to do even more operations in the same SIMD batch, for example `AddPixels` in `PredictorAddX` methods. Not sure about `GetResidual` though. The it looks like the switch statement can be moved outside of the loop, but you may need to store `predict` results in a temporary buffer, and do a second loop for the [post-operations](https://github.com/SixLabors/ImageSharp/blob/2f048ca1d311b11047fc4be706ea4aaa1ef4c524/src/ImageSharp/Formats/Webp/Lossless/PredictorEncoder.cs#L421-L465).

brianpopow:
There is actually a `PredictBatch` methods which applies the same predictor method on 32 pixels each, but it will be a huge overhaul to make the Predictor methods work on many pixels.

I dont think its worth the effort. The Average methods only make a small part of the total consumed time.

I will close this PR then. Will keep the branch, maybe we come up with something in the future.

PR #1834, opened 2021-11-17
Fix gaussian blurring changing the blur colour when transperancy is used

Turnerj:
### Prerequisites

### Description

Firstly, all credit goes to @JimBobSquarePants for [identifying the fix](https://twitter.com/James_M_South/status/1460931447766749193).

Blurring an image that has a transparent background can lead to strange results depending on the colour.
For example, take this code:

```csharp
using var image = new Image<Rgba32>(400, 400);
image.Mutate(x => x.Fill(Color.ParseHex("48f25b"), new Rectangle(100, 100, 200, 200)).GaussianBlur(20));
```

Master generates this:

image

This PR generates:

image

I do want to add tests for it, just working out how to do it like how you have your Bokeh blur tests.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Convolution/Convolution2PassProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Convolution/Convolution2PassProcessor{TPixel}.cs
@@ -396,6 +396,8 @@ namespace SixLabors.ImageSharp.Processing.Processors.Convolution
 
                     PixelOperations<TPixel>.Instance.ToVector4(this.configuration, sourceRow, sourceBuffer);
 
+                    Numerics.Premultiply(sourceBuffer);
+
                     ref Vector4 sourceBase = ref MemoryMarshal.GetReference(sourceBuffer);
                     ref Vector4 sourceEnd = ref Unsafe.Add(ref sourceBase, sourceBuffer.Length);
                     ref Vector4 targetStart = ref targetBase;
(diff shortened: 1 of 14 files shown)


Turnerj:
Took a little while but I worked out how to do the reference images for tests (and yep, I did run the optimizer on them so the file size is smaller for the repo).

This is the source image:
blur

Running the test on master:
OnFullImage_Rgba32_blur_3
InBox_Rgba32_blur_5

Running the test on this PR:

OnFullImage_Rgba32_blur_3

InBox_Rgba32_blur_5

As you can see, the blurring was doing all sorts of weirdness with colours without that line.

JimBobSquarePants (approved):
HighFiveTopGunGIF

src/ImageSharp/Processing/Processors/Convolution/Convolution2PassProcessor{TPixel}.cs
```diff
@@ -396,6 +396,8 @@ private void Convolve4(int y, Span<Vector4> span)
 
                     PixelOperations<TPixel>.Instance.ToVector4(this.configuration, sourceRow, sourceBuffer);
 
+                    Numerics.Premultiply(sourceBuffer);
```
antonfirsov, line 399: This can be made conditional on `PixelTypeInfo.AlphaRepresentation`:

https://github.com/SixLabors/ImageSharp/blob/9777a3b84eb8a38adb09c13577be393980116b28/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeProcessor%7BTPixel%7D.cs#L191-L194

JimBobSquarePants: Yeah. Something to look into. I keep meaning to see what else we can add to `PixelTypeInfo` actually that would help us in other places.

PR #1835, opened 2021-11-17
Fix vanilla build on VS2019

ynse01:
### Prerequisites

- [X ] I have written a descriptive pull-request title
- [ X] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [ X] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.

### Description
VS2019 uses the C# 8.0 compiler by default, when using .NET Core 5 or lower as TargetFramework. ImageSharp code is using C# 9.0 features. Nowhere is the Language version specified. Apparently relying on the default C# compiler of .NET 6 TargetFramework, which is 9.0.

Fixed it by specifying the Language version, to 9.0.

Verified in VS2019 v16.11.7 and VS2022 v17.0.1.
Installed SDKs:
3.1.407
5.0.403
6.0.100


JimBobSquarePants:
The language version is specified in the shared git submodule. I’ve been able to build with VS2019 without any additional changes.

ynse01:
Indeed, shared-infrastructure contains this already. My bad.

PR #1836, opened 2021-11-17
Add SSE41 version of UpSample

brianpopow:
### Prerequisites

### Description

This PR adds a SSE41 version of UpSample, which is used during decoding of lossy webp images to convert from YUV to RGB.
This is where a significant portion of the time is spend (ca. 30%).

Related to #1786

This is still work in progress. This is still left to do:
- [x]  Implement converting last block.
- [x] ~The last column of each block seems to be incorrect.~ Fixed with 3f43883
- [x] Add tests
- [x] Add benchmarks once above issues are fixed

Example output:
output

Output looks correct now:
output


brianpopow:
Benchmark results look pretty good:

### master
upsample

### PR
upsample_sse

Testimage used:
[Calliphora_lossy.zip](https://github.com/SixLabors/ImageSharp/files/7563427/Calliphora_lossy.zip)

brianpopow:
When looking at the profiler measures, the improvements seem huge. 
I have started to look into improving the UpScale method, because the profiler was telling me 30% of the time is spent here.
Unfortunately when looking at the benchmark results, its not a big difference. Its barely noticeable.
I really was expecting a big impact on the lossy decoding time for this PR, but its not there.

I am not sure what im doing wrong here or if im reading the profiler output wrong. I am using dotTrace with `Line-by-Line` sampling, because I thought this is the most accurate, but it seems to be rather misleading.

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
```diff
@@ -18,6 +23,269 @@ internal static class YuvConversion
...
+            }
+
+            // For UpSample32Pixels, 17 u/v values must be read-able for each block.
+            int pos;
+            int uvPos;
+            for (pos = 1, uvPos = 0; pos + 32 + 1 <= len; pos += 32, uvPos += 16)
+            {
+                UpSample32Pixels(topU.Slice(uvPos), curU.Slice(uvPos), ru);
+                UpSample32Pixels(topV.Slice(uvPos), curV.Slice(uvPos), rv);
+                ConvertYuvToBgrSse41(topY, bottomY, topDst, bottomDst, ru, rv, pos, xStep);
```
antonfirsov, line 186: I would avoid slicing and passing spans to these private methods here, and iterate instead using reference references or pointers. (Pinning is fine with such complex code IMO if it helps keeping readability.)

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
```diff
@@ -312,6 +580,175 @@ public static void YuvToBgr(int y, int u, int v, Span<byte> bgr)
...
+
+            Vector128<ushort> r0 = Sse2.MultiplyHigh(v0.AsUInt16(), K26149.AsUInt16());
+            Vector128<ushort> r1 = Sse2.Subtract(y1.AsUInt16(), K14234.AsUInt16());
+            Vector128<ushort> r2 = Sse2.Add(r1, r0);
+
+            Vector128<ushort> g0 = Sse2.MultiplyHigh(u0.AsUInt16(), K6419.AsUInt16());
+            Vector128<ushort> g1 = Sse2.MultiplyHigh(v0.AsUInt16(), K13320.AsUInt16());
+            Vector128<ushort> g2 = Sse2.Add(y1.AsUInt16(), K8708.AsUInt16());
+            Vector128<ushort> g3 = Sse2.Add(g0, g1);
+            Vector128<ushort> g4 = Sse2.Subtract(g2, g3);
```
antonfirsov, line 731: You can try re-grouping the code to do identical operations in a row. It works better with CPU piplines. I would also try to  group the vector loads of `LoadHigh` (after inlining the method).

```suggestion
            Vector128<ushort> y1 = Sse2.MultiplyHigh(y0.AsUInt16(), K19077.AsUInt16());
            Vector128<ushort> r0 = Sse2.MultiplyHigh(v0.AsUInt16(), K26149.AsUInt16());
            Vector128<ushort> g0 = Sse2.MultiplyHigh(u0.AsUInt16(), K6419.AsUInt16());
            Vector128<ushort> g1 = Sse2.MultiplyHigh(v0.AsUInt16(), K13320.AsUInt16());
            
            Vector128<ushort> r1 = Sse2.Subtract(y1.AsUInt16(), K14234.AsUInt16());
            Vector128<ushort> r2 = Sse2.Add(r1, r0);

            Vector128<ushort> g2 = Sse2.Add(y1.AsUInt16(), K8708.AsUInt16());
            Vector128<ushort> g3 = Sse2.Add(g0, g1);
            Vector128<ushort> g4 = Sse2.Subtract(g2, g3);
```

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
```diff
@@ -312,6 +586,178 @@ public static void YuvToBgr(int y, int u, int v, Span<byte> bgr)
...
+            g = Sse2.ShiftRightArithmetic(g4.AsInt16(), 6); // range: [-10953, 27710]
+            b = Sse2.ShiftRightLogical(b2.AsInt16(), 6); // range: [0, 34238]
+        }
+
+        // Load the bytes into the *upper* part of 16b words. That's "<< 8", basically.
+        [MethodImpl(InliningOptions.ShortMethod)]
+        private static Vector128<byte> LoadHigh(Vector64<byte> src)
+        {
+            Vector128<byte> tmp = Unsafe.As<Vector64<byte>, Vector128<byte>>(ref src);
+            return Sse2.UnpackLow(Vector128<byte>.Zero, tmp);
```
antonfirsov, line 757: Can't you actually load to `Vector128<byte>`, then bitshift? Could be much cheaper, not sure what is `Unsafe.As<Vector64<byte>, Vector128<byte>>(ref src)` being compiled to.

brianpopow: Its just a `mov`, [SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCEBXAMy5ilwDcNek1YAlDgDsMAS3wwWASWlQZk3DLCDhjZiwnS5C5RlXrNuFgA0AHEiG1d4qbPksAwhHwAHGQBs+AGU+ADdNGG1aAGYGKVxsXiZSBncGAG8aBiymGOIUBgBZAAoANRgwDGhUAB5gAE8MGAA+BlwoMABKBkzsjOpsgYYyiuhyUhtahuaGDB8GAF4GAFV1BIUAQVxq4cqoGvrGprQh8t2xiYPmpqLYLlb2jocBgF8aZ6A=), this should be fine, right?

brianpopow: I cant seem to figure out how to do this differently/better.

JimBobSquarePants:
I'm happy for this to go in as-is but I'll leave it to @antonfirsov for a final review since he had questions/suggestions

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
antonfirsov: I thought this is a vector level shuffle/shift, now I see what you are doing. What I mean is that if you only unpack from the lower half of `tmp`, pushing this through `Vector64<byte>` can be avoided by adding an 8 byte padding to `y`, `u`, `v`:

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCEBXAMy5ilwDcNek1YAlDgDsMAS3wwWASWlQZk3DLCDhjZiwnS5C5RlXrNuFgA0AHEiG1d4qbPksAwhHwAHGQBs+AGU+ADdNGG1aAGYGKVxsXiZSBncGAG8aBiymGIA1GDAMaHJSGwAeYABPDBgAPgYAQQAKWC4GKpqGXCgwAEoM6myhhnzC6FQK6rqGDB9yBgBeBgBVdQSFBtxJmrQRgqKoCY662paYNu6+h2Hs0YOS8uP62e9kpdX43hZNsrvxlG2dV2fygD0BtVOrRmc161xuTAA7AxArgYKQWKtvNgwABrAAyEAA7k0QWCniwAFp8CC7F6kWGZbIAX0ZWVZOT2Y1BpXBDDYZzaxy6PX67KGpJ5bCmzx8ixWay+P2OwP2xR5T0h52FVzF2WISJRaIxkixuIJxIlj2llOptJ8DMGzJoTKAA===

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
```diff
@@ -312,6 +586,178 @@ public static void YuvToBgr(int y, int u, int v, Span<byte> bgr)
             bgr[0] = (byte)YuvToB(y, u);
         }
 
+#if SUPPORTS_RUNTIME_INTRINSICS
+
+        private static void ConvertYuvToBgrSse41(Span<byte> topY, Span<byte> bottomY, Span<byte> topDst, Span<byte> bottomDst, Span<byte> ru, Span<byte> rv, int curX, int step)
+        {
+            YuvToBgrSse41(topY.Slice(curX), ru, rv, topDst.Slice(curX * step));
+
+            if (bottomY != null)
```
antonfirsov, line 595: This if can be moved outside of the loop for happier pipelines / branch predictor.

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
```diff
@@ -312,6 +586,178 @@ public static void YuvToBgr(int y, int u, int v, Span<byte> bgr)
             bgr[0] = (byte)YuvToB(y, u);
         }
 
+#if SUPPORTS_RUNTIME_INTRINSICS
+
+        private static void ConvertYuvToBgrSse41(Span<byte> topY, Span<byte> bottomY, Span<byte> topDst, Span<byte> bottomDst, Span<byte> ru, Span<byte> rv, int curX, int step)
```
antonfirsov, line 591: These spans can be also converted to `ref`-s.

antonfirsov (approved):
There is still some space for improvement, but feel free to move on, if you'd better focus on something else.

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
brianpopow: I was not aware of this, looks much better now, thanks.

Changed with cc5f7af7

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
brianpopow: Moved the if outside of the loop with 65870b96

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
brianpopow: I feel a bit more comfortable with using spans here, not sure if its really worth changing to ref's

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
```diff
@@ -312,6 +602,175 @@ public static void YuvToBgr(int y, int u, int v, Span<byte> bgr)
...
+            output2 = Ssse3.Shuffle(input0, shuffle2);
+            output3 = Ssse3.Shuffle(input1, shuffle0);
+            output4 = Ssse3.Shuffle(input1, shuffle1);
+            output5 = Ssse3.Shuffle(input1, shuffle2);
+        }
+
+        // Convert 32 samples of YUV444 to R/G/B
+        private static void ConvertYuv444ToRgbSse41(Span<byte> y, Span<byte> u, Span<byte> v, out Vector128<short> r, out Vector128<short> g, out Vector128<short> b)
+        {
+            Vector128<byte> y0 = LoadHigh(ref MemoryMarshal.GetReference(y));
```
antonfirsov, line 739: Ignore me if you already double checked this, but what you really need to make sure now is that if `y`,`u`,`v` have 8 bytes even at the very end, otherwise it's a memory corruption.

The code that allocates these buffers seems to add padding at first glance, but I don't really follow it:

https://github.com/SixLabors/ImageSharp/blob/099a67605e36262601101c5ef51f7fcea3ba9962/src/ImageSharp/Formats/Webp/Lossy/Vp8Decoder.cs#L66-L73

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
antonfirsov: The main overhead seems to appear here:

https://github.com/SixLabors/ImageSharp/blob/984a72571db182b8a7c0eca83f9e9de6a9b162df/src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs#L620-L622

You can consider "de-spanning" only in this method and `ConvertYuv444ToRgbSse41` replacing `.Slice` with `Unsafe.Add` to y/u/v base references.

brianpopow: ok i am fine with that: 6293f72c

JimBobSquarePants (approved):

src/ImageSharp/Formats/Webp/Lossy/YuvConversion.cs
```diff
@@ -312,6 +602,178 @@ public static void YuvToBgr(int y, int u, int v, Span<byte> bgr)
...
+            output4 = Ssse3.Shuffle(input1, shuffle1);
+            output5 = Ssse3.Shuffle(input1, shuffle2);
+        }
+
+        // Convert 32 samples of YUV444 to B/G/R
+        private static void ConvertYuv444ToBgrSse41(ref byte y, ref byte u, ref byte v, out Vector128<short> r, out Vector128<short> g, out Vector128<short> b)
+        {
+            Vector128<byte> y0 = LoadHigh(ref y);
+            Vector128<byte> u0 = LoadHigh(ref u);
+            Vector128<byte> v0 = LoadHigh(ref v);
```
antonfirsov, line 744: Wanted to stop already but can't resist 😄 Pipelined loads may perform better:

```suggestion
            Vector128<byte> y0 = Unsafe.As<byte, Vector128<byte>>(ref y);
            Vector128<byte> u0 = Unsafe.As<byte, Vector128<byte>>(ref u);
            Vector128<byte> v0 = Unsafe.As<byte, Vector128<byte>>(ref v);
            y0 = Sse2.UnpackLow(Vector128<byte>.Zero, y0);
            u0 = Sse2.UnpackLow(Vector128<byte>.Zero, u0);
            v0 = Sse2.UnpackLow(Vector128<byte>.Zero, v0);
```

brianpopow: yes, makes sense. Changed with 7775c343

antonfirsov (approved):
Thanks for the patience!

PR #1840, opened 2021-11-19
Remove HistogramEqualizationOptions.Default

antonfirsov:
### Prerequisites

### Description
We should not expose an arbitrary mutable variable like this. The common practice in the library is to use constructor to new-up instances of similar `XyzOptions` objects.


tests/ImageSharp.Tests/Formats/WebP/LossyUtilsTests.cs
```diff
@@ -38,7 +38,7 @@ private static void RunVp8Sse4X4Test()
             int actual = LossyUtils.Vp8_Sse4X4(a, b);
 
             Assert.Equal(expected, actual);
-		}
+        }
```
antonfirsov, line 41: Unrelated tabs -> spaces change. I wonder why doesn't stylecop catch these?

JimBobSquarePants (approved):
Yeah, this shouldn’t exist. Good catch!

PR #1844, opened 2021-11-21
Unified jpeg benchmark

br3aker:
### Prerequisites

### Description
This PR adds 'unified' jpeg decoder benchmark with different kinds of jpeg images for somewhat baseline performance comparison of new code/optimizations. I've mainly used already existing images for this benchmark it by no means has all of different jpeg kinds - no progressive 4:4:4 for example. Benchmark image collection should naturally grow over time.

I've also restructured formats behcmarks folder to this instead of a one big folder of all formats benchmarks:
image


JimBobSquarePants (approved):
Good idea 👍

PR #1845, opened 2021-11-21
Jpeg decoder mcu size fix

br3aker:
### Prerequisites

### Description
This PR fixes a bug described in https://github.com/SixLabors/ImageSharp/pull/1816#issuecomment-965916098.

Basically, it removes dependency on the 'first component' during decoding because jpegs can have lower luma sampling than chroma. This bug can be spotted with image from current test image base - [jpeg422.jpg](https://github.com/SixLabors/ImageSharp/blob/master/tests/Images/Input/Jpg/baseline/jpeg422.jpg). This is a decoded image from current master code:
jpeg422_MASTER_BUG

Current PR produces almost ideal image - diff is 0.0013% compared to PhotoShop png reference:

```
MASTER
Total difference: 19,7988%
[Δ(0,-20560,0,0) @ (0,0)];
[Δ(0,-20560,0,0) @ (1,0)];
[Δ(0,-20303,-257,0) @ (2,0)];
[Δ(0,-20046,0,0) @ (3,0)];
[Δ(0,-19789,-257,0) @ (4,0)]...
```
```
PR
Total difference: 0,0013%
[Δ(0,-257,257,0) @ (18,2)];
[Δ(0,-257,0,0) @ (19,3)];
[Δ(0,-257,0,0) @ (23,7)];
[Δ(0,257,0,0) @ (32,8)];
[Δ(0,0,-257,0) @ (56,8)]...
```

I've also added this image to the decoder test suit.


src/ImageSharp/Formats/Jpeg/Components/Decoder/SpectralConverter{TPixel}.cs
```diff
@@ -65,26 +62,26 @@ public override void InjectFrameData(JpegFrame frame, IRawJpegData jpegData)
             MemoryAllocator allocator = this.configuration.MemoryAllocator;
 
             // iteration data
-            IJpegComponent c0 = frame.Components[0];
+            int majorBlockWidth = frame.Components.Max((component) => component.SizeInBlocks.Width);
```
JimBobSquarePants, line 65: How’s the perf using Max here? We’ve avoided Linq everywhere in the codecs AFAIK

src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanScanDecoder.cs
```diff
@@ -151,6 +151,7 @@ private void ParseBaselineData()
             if (this.componentsCount == this.frame.ComponentCount)
             {
                 this.ParseBaselineDataInterleaved();
+                this.spectralConverter.CommitConversion();
```
br3aker, line 154: Before this PR spectral converter did some magic to determine if image was converted based on number off processed rows. But someone wise once said: "explicit is better than implicit" so I've decided to create a separate method to 'seal' conversion. It doesn't look so OOPish but at least it's more readable IMO.

src/ImageSharp/Formats/Jpeg/Components/Decoder/SpectralConverter{TPixel}.cs
br3aker: Considering it's from 1 to 4 elements and it executes only once I don't think it will matter but I will run benchmarks just to be sure.

br3aker:
Fresh benchmarks

Master
|        'Baseline 4:4:4 Interleaved' | 11.712 ms | 0.0454 ms | 0.0355 ms |
|        'Baseline 4:2:0 Interleaved' |  8.931 ms | 0.0513 ms | 0.0455 ms |
|        'Baseline 4:0:0 (grayscale)' |  1.639 ms | 0.0045 ms | 0.0037 ms |
| 'Progressive 4:2:0 Non-Interleaved' | 13.554 ms | 0.0711 ms | 0.0665 ms |


PR
|        'Baseline 4:4:4 Interleaved' | 11.781 ms | 0.0737 ms | 0.0654 ms |
|        'Baseline 4:2:0 Interleaved' |  8.688 ms | 0.0345 ms | 0.0306 ms |
|        'Baseline 4:0:0 (grayscale)' |  1.643 ms | 0.0092 ms | 0.0086 ms |
| 'Progressive 4:2:0 Non-Interleaved' | 13.770 ms | 0.0928 ms | 0.0823 ms |

While progressive 420 is indeed slower by almost 1.5% baseline 420 is faster by 2.7% without any particular reasons. I believe it's a noise, performance shouldn't be affected by LINQ calls, especially by 4 element array.

JimBobSquarePants (approved):
LGTM. Your additional comments make sense

PR #1846, opened 2021-11-21
Webp improvements

brianpopow:
### Prerequisites

### Description

This PR changes the following:

- Change hashchain to use the memoryAllocator (OffsetLength array is as big as the pixel count)
- Change huffman code and HTreeGroup to a struct
- Avoid allocating too many `CostInterval` objects by pooling a small amount of them. The number of `CostInterval` objects can easily get into 100k+ for an 1 megapixel image.
- CostsManager now uses MemoryAllocator
- Remove not needed DeepClone of HuffmanTree (while it was changed to a struct, the DeepClone was not removed and is responible for alot of allocations)
- Change PixOrCopyMode, HistoIx and EntropyIx enums to be of byte type

Related to #1786

### Allocations master:

Allocations_master

### Allocations PR:

Allocations_pr

Test was done with lossless encoding of Calliphora image.

### DIFF
--- a/src/ImageSharp/Formats/Webp/Lossless/Vp8LHashChain.cs
+++ b/src/ImageSharp/Formats/Webp/Lossless/Vp8LHashChain.cs
@@ -8,7 +8,7 @@ using SixLabors.ImageSharp.Memory;
 
 namespace SixLabors.ImageSharp.Formats.Webp.Lossless
 {
-    internal class Vp8LHashChain
+    internal class Vp8LHashChain : IDisposable
     {
         private const uint HashMultiplierHi = 0xc6a4a793u;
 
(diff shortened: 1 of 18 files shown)


src/ImageSharp/Formats/Webp/Lossless/Vp8LHashChain.cs
```diff
@@ -280,5 +287,25 @@ private static int GetWindowSizeForHashChain(int quality, int xSize)
...
+
+                this.disposed = true;
+            }
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+            this.Dispose(disposing: true);
+            GC.SuppressFinalize(this);
```
JimBobSquarePants, line 308: If your types can be sealed I would do so and avoid `SuppressFinalize` where it is not necessary.

antonfirsov: Yeah I think the `Dispose(bool disposing)` pattern is only favorable when we actually have a finalizer or expect subclasses to have a finalizer.

JimBobSquarePants (approved):
Nice job 👍

PR #1847, opened 2021-11-23
Jpeg new IDCT & fused zigzag/transpose steps

br3aker:
### Prerequisites

### Description
This PR brings 2 new things:
1. New IDCT implementation base on Arai, Agui, and Nakajima works, it's a faster than master implementation with same level of overall quality, spectral results are even closer to libjpeg spectral dump
2. Fused matrix tranpose into zig-zag step during jpeg blocks decoding

### Benchmarks
Master
|        'Baseline 4:4:4 Interleaved' | 11.781 ms | 0.0737 ms | 0.0654 ms |
|        'Baseline 4:2:0 Interleaved' |  8.688 ms | 0.0345 ms | 0.0306 ms |
|        'Baseline 4:0:0 (grayscale)' |  1.643 ms | 0.0092 ms | 0.0086 ms |
| 'Progressive 4:2:0 Non-Interleaved' | 13.770 ms | 0.0928 ms | 0.0823 ms |

PR
|        'Baseline 4:4:4 Interleaved' | 11.127 ms | 0.0659 ms | 0.0550 ms |
|        'Baseline 4:2:0 Interleaved' |  8.458 ms | 0.0289 ms | 0.0256 ms |
|        'Baseline 4:0:0 (grayscale)' |  1.550 ms | 0.0050 ms | 0.0044 ms |
| 'Progressive 4:2:0 Non-Interleaved' | 13.220 ms | 0.0449 ms | 0.0398 ms |


src/ImageSharp/Formats/Jpeg/Components/FastFloatingPointDCT.cs
```diff
@@ -15,102 +15,196 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
...
         };
 
         /// <summary>
-        /// Adjusts given quantization table to be complient with FDCT implementation.
+        /// Adjusts given quantization table for usage with <see cref="TransformIDCT"/>.
+        /// </summary>
+        /// <param name="quantTable">Quantization table to adjust.</param>
+        public static void AdjustToIDCT(ref Block8x8F quantTable)
+        {
+            for (int i = 0; i < Block8x8F.Size; i++)
```
JimBobSquarePants, line 68: AdjustmentCoefficients is 64 elements yeah? If so, since `Block8x8F` indexer doesn't do bounds checks in release we could avoid the check here.

src/ImageSharp/Formats/Jpeg/Components/Block8x8.cs
```diff
@@ -337,6 +337,64 @@ public nint GetLastNonZeroIndex()
             }
         }
 
+        /// <summary>
+        /// Transpose the block inplace.
+        /// </summary>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public void TransposeInplace()
```
JimBobSquarePants, line 344: According to codecov we have no coverage for this? It looks correct to me but better to be safe than sorry.

br3aker: Done.

src/ImageSharp/Formats/Jpeg/Components/FastFloatingPointDCT.cs
br3aker: Fixed. Did it without indexer with pure `Unsafe.Add`. We need `nint` indexer for `Block8x8F` & `Block8x8` IMO.

JimBobSquarePants: Is there any reason we can’t have that now?

br3aker: There are a lot of places where indexer is used and which can use nint indexer. I'd rather do it as a separate commit. Code cleanup is the last part of my decoder optimization commit so I can do it there.

JimBobSquarePants: Yep. Makes sense

JimBobSquarePants (approved):
Thanks for working on this. Excellent as ever! 👍

PR #1848, opened 2021-11-23
Add AVX2 versions of CombinedShannonEntropy

brianpopow:
### Prerequisites

### Description

This PR adds a AVX2 version of `CombinedShannonEntropy`, which is used during lossless webp encoding.

Related to #1786

Profiling results:

### master
shannon_master

### PR
shannon_avx

Test image was `Callipohora.png`


src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
```diff
@@ -759,28 +760,184 @@ public static void BundleColorMap(Span<byte> row, int width, int xBits, Span<uin
...
+                        if (tmpRef != 0)
+                        {
+                            retVal -= FastSLog2((uint)tmpRef);
+                            if (Unsafe.Add(ref xRef, i) != 0)
+                            {
+                                retVal -= FastSLog2((uint)Unsafe.Add(ref xRef, i));
+                            }
+                        }
+
+                        if (Unsafe.Add(ref tmpRef, 1) != 0)
```
brianpopow, line 797: Note: I have tried to put those repeating if statements into own method calls, but profiling has shown that this makes it actually slower even with Aggressive Inlining.

JimBobSquarePants: Couldn't it be a loop? It looks incremental 0-7 to me. 

However, my money says there's something clever that can be done with masking here to determine if each element != 0 and apply the diff as a single operation. I'm sure I've seen similar before.

br3aker: Simd? You can check each element in a vector without if checks at all for `tmpRef` and `xRef` vectors.

You can also remove if checks with log2  precalculation for each case and simply multiplying that log2 vectors with comparison mask values. This may or may not be faster than if-checks - depends ong FastSLog2 implementation.

brianpopow: The original SSE2 version of this uses macros. I tried to keep it as similar as possible.
Because of that, I thought a loop might be also slower, but I will give it a try and test it.

`FastSLog2` is in most cases (>90% just a lookup table), but still most of the time is spend here.

br3aker: I guess loop can further degrade branch predictor history. Masking is still possible but that would require some serious code rewrite for AVX branch.

brianpopow: I have done profile tests now with loops several times and it makes it noticeable slower.

I am not really sure that it is worth trying to optimize it further, if it would take a huge amount of work to do so. As can be seen here:
shannon_avx2

The main time consumer is `FastSLog2`.

br3aker: I honestly have no idea how webp works but as far as I understand checked value is actually pretty random, adding for-loop on top of it may screw up branch predictor with yet another stable (always true for 8 iterations) if-check - that's what's most likely causing performance drop.

All of extra if-checks can be removed via simd masks but it's a question whether it'd be faster than current implementation.

brianpopow: I had a previous SSE2 version which used masks, if you look at commit: ed8bd615f2be3cafd1a23782e9f7d07c6375d967, but this was not better then the current approach. Maybe, if we could create a AVX version of that, it could be better, but I am also not sure if it really can beat the current one.

br3aker: It's actually easier to describe it in code than in human language. I'll probably try to implement it after this gets merged.

brianpopow: @br3aker thanks, but my advice would be, to only do it, if you think it would be easy for you and not to much work. As I said before, I am really unsure how much we can gain from this.

src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
```diff
@@ -759,28 +760,184 @@ public static void BundleColorMap(Span<byte> row, int width, int xBits, Span<uin
...
             {
-                uint xi = (uint)x[i];
-                if (xi != 0)
+                double retVal = 0.0d;
+                Span<int> tmp = stackalloc int[8];
+                ref int xRef = ref MemoryMarshal.GetReference(x);
+                ref int yRef = ref MemoryMarshal.GetReference(y);
+                Vector256<int> sumXY256 = Vector256<int>.Zero;
+                Vector256<int> sumX256 = Vector256<int>.Zero;
+                ref int tmpRef = ref MemoryMarshal.GetReference(tmp);
```
gfoidl, line 772: `Span<int> tmp` is only used to get `ref int tmpRef`, so we could just avoid the span at all, including the stack-cookie check the JIT needs to emit (to detect stack overruns).

```suggestion
                Vector256<int> tmp = Vector256<int>.Zero;    // has the size of the scratch space of sizeof(int) * 8
                ref int xRef = ref MemoryMarshal.GetReference(x);
                ref int yRef = ref MemoryMarshal.GetReference(y);
                Vector256<int> sumXY256 = Vector256<int>.Zero;
                Vector256<int> sumX256 = Vector256<int>.Zero;
                ref int tmpRef = ref Unsafe.As<Vector256<int>, int>(ref tmp);
```

For comparison see [sharplap](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4a9Jq049+ggJI8ZEMfKhKV6zY2btufASyMYovLrmXXqmgMxNSBiEGAG8aBgimf3cMBgBZOgAKAEoGcMiw6kjshlkxbC4AHhiAPgYMcQYAXgZcDGwwAGtsSUkIMAYYgG0ADgBdDSyciNgAM06ecvE2GHGasfiYfGgATzjsKFwACxaWAHEYDBnRmRguFUSKsWTB4Yj04eIAdgYAVQ9sE5YAQQATX8SCyuxzQDHINweDAAvpDIV1ZI1eGIADLtFq4Iy8DB9SHEaKTOLkFJpIYRTJ3XL5IqlKZiaq1epNFptDrdfq3O4LGK0470hZxJardabHaSfaHY6nc4wS7iCGknKQ7LPN4fL5/AFA6azUHgjkRGEK3H42JxUgpSHku4ANRgYAw0FIAFYkMUeGVfhx8PgVvTfrNsBxJBh9Tk8gU3Rgyld6QLllA1httrshLBsBgYOGuIDZqrcJ9BN9cIVbfbHS7IyVQaUc+NPd6VslQT0bhSlZEuZNgbn5rm40Kk6LxUdZlKLld5W2FcqXu98+r/rWeTqwZPIobssaJqbfBaFVbhqWHVBna6afWfX6A0GQ+2Rrnud25gwFnOCz9i0fy2f3dX3UuL0bUNN2nSIVTfBdNVzJ9dTXA0aChIA=) (method _M3_ is used here).

src/ImageSharp/Common/Helpers/Numerics.cs
```diff
@@ -820,6 +820,21 @@ public static int ReduceSum(Vector128<int> accumulator)
...
+        /// <summary>
+        /// Reduces elements of the vector into one sum.
+        /// </summary>
+        /// <param name="accumulator">The accumulator to reduce.</param>
+        /// <returns>The sum of all elements.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int ReduceSum(Vector256<int> accumulator)
+        {
+            Vector128<int> vec0 = Avx2.ExtractVector128(accumulator, 0);
+            Vector128<int> vec1 = Avx2.ExtractVector128(accumulator, 1);
```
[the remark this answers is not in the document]
brianpopow: That's a neat trick

src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
brianpopow: that's a neat trick

JimBobSquarePants (approved):

PR #1849, opened 2021-11-24
Add AVX2 version of AddVector

brianpopow:
### Prerequisites

### Description

This PR adds a AVX2 version of AddVector in `Vp8LHistogram`, which is used during lossless webp encoding.

Related to #1786

Profiling results:

### master
AddVector

### PR
AddVector_avx


src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
```diff
@@ -80,8 +80,10 @@ public static int FindMatchLength(ReadOnlySpan<uint> array1, ReadOnlySpan<uint>
         public static int VectorMismatch(ReadOnlySpan<uint> array1, ReadOnlySpan<uint> array2, int length)
```
br3aker, line 80: If array sizes are generally long enough - it can be easily vectorized.

src/ImageSharp/Formats/Webp/Lossless/Vp8LHistogram.cs
```diff
@@ -505,11 +511,48 @@ private static double ExtraCost(Span<uint> population, int length)
             return cost;
         }
 
-        private static void AddVector(uint[] a, uint[] b, uint[] output, int size)
+        private static void AddVector(Span<uint> a, Span<uint> b, Span<uint> output, int size)
```
antonfirsov, line 514: We should `DebugGuard` for count mismatch.

Nit: `count` would be better name than `size` here.

src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
brianpopow: If you look at the commit history, i tried this: 491b742

It actually was twice as fast with benchmark.net and synthetic data, but with real data it did not perform well.
Usually the loop is quit early on, so it was actually slower then the non vectorized version.

antonfirsov (approved):
This looks good all the way now. I keep wondering about the reusability of the method, but probably `uint` is very specific, making it less useful for other codecs / scenarios.

src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
br3aker: Hmmm, got it. Thanks for the explanation!

PR #1851, opened 2021-11-25
Add support for Portable Bitmap images

ynse01:
### Prerequisites

### Description
This implements the Portable Bitmap support [idea](https://github.com/SixLabors/ImageSharp/discussions/1842) I posted earlier.

This adds support for encoding and decoding of Portable Bitmap images. These are the following image types:
- Pbm - Black and White images
- Pgm - Grayscale images
- Ppm - Color images

All of these in both Plain and Binary excoding.
These image formats fall under the PNM family of images. See [The PNM format](http://netpbm.sourceforge.net/doc/pnm.html)
The somewhat related [PAM](http://netpbm.sourceforge.net/doc/pam.html) image format, is not implemented here.

The class structure of the code itself is largly based on the existing Tga encoder / decoder. 

I could find any other reference to black-and-white pixels. So, I mapped these pixels in a L8 pixel.  White is bit value 0 in Pbm, but maps to L8 value 255. Black is bit valaue 1, but maps to L8 value 0. This choice aligns these pixels better with the existing use of L8.

Test code:
I'm open for suggestion on how to improve the test coverage. Some of my struggle was with the RoundTripTest. I couldn't get the L8 pixel images to load properly. Image.Load always returns RGB pixels, which break the pixel comparison.
In PgmDecoderTest, there is an issue with differences in behavior of Magick. It treats black-and-white pixels differently. And in the same test, the RGB input images don't load in Magick, where they DO load in IrfanView, XnConvert and my code.


tests/ImageSharp.Tests/Formats/Pbm/PbmDecoderTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        [Theory]
+        [InlineData(BlackAndWhitePlain, PbmColorType.BlackAndWhite)]
+        [InlineData(BlackAndWhiteBinary, PbmColorType.BlackAndWhite)]
+        [InlineData(GrayscalePlain, PbmColorType.Grayscale)]
+        [InlineData(GrayscaleBinary, PbmColorType.Grayscale)]
+        [InlineData(GrayscaleBinaryWide, PbmColorType.Grayscale)]
+        [InlineData(RgbPlain, PbmColorType.Rgb)]
+        [InlineData(RgbBinary, PbmColorType.Rgb)]
+        public void PpmDecoder_CanDecode(string imagePath, PbmColorType expectedColorType)
```
antonfirsov, line 24: Any chance you can add pixel data validation to these tests?
Instead of relying on an ImageMagick reference decoder, you can push the expected output as a PNG in the `ReferenceOutput` folder, and do something similar `JpegDecoderTests` do:
- [Tests]( https://github.com/SixLabors/ImageSharp/blob/d81c511f068ac0fdb580310fe0d2328ca713e919/tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.Baseline.cs)
- [Expected output](https://github.com/SixLabors/ImageSharp/tree/master/tests/Images/External/ReferenceOutput/JpegDecoderTests)

If you are sure your current implementation is correct, you can just copy the output of [`DebugSave`](https://github.com/SixLabors/ImageSharp/blob/d81c511f068ac0fdb580310fe0d2328ca713e919/tests/ImageSharp.Tests/TestUtilities/TestImageExtensions.cs#L29) there without bothering to convert images with an external app.

We also need at least one test to validate `Image.Load<T>` works with multiple pixel types.

src/ImageSharp/Formats/Pbm/BinaryEncoder.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                WriteBlackAndWhite(configuration, stream, image);
+            }
+        }
+
+        private static void WriteGrayscale<TPixel>(Configuration configuration, Stream stream, ImageFrame<TPixel> image)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            int width = image.Size().Width;
+            int height = image.Size().Height;
```
antonfirsov, line 64:
```suggestion
            int width = image.Width;
            int height = image.Height;
```

Also on other places.

antonfirsov (commented):
Thanks a lot for taking the time for this! Implementation generally looks good, but it would be nice to see validation and debug output in the decoder tests. That would help us reasoning about a format we aren't familiar with.

tests/Images/Input/Pbm/00000_00000.ppm
```diff
@@ -0,0 +1,4 @@
+P6
```
JimBobSquarePants, line 1: Looks like we'll need to add the extensions for these file types to our .gitconfig and also ensure that the images are stored in GitLFS. I'll do the upstream work to allow that.

src/ImageSharp/Formats/Pbm/StreamExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace SixLabors.ImageSharp.Formats.Pbm
+{
+    internal static class StreamExtensions
+    {
+        public static int WriteDecimal(this Stream stream, int value)
+        {
+            string str = value.ToString();
+            byte[] bytes = Encoding.ASCII.GetBytes(str);
+            stream.Write(bytes);
+            return bytes.Length;
```
gfoidl, line 16: Make it allocation-free:
```suggestion
            Span<byte> buffer = stackalloc byte[128];
            Utf8Formatter.TryFormat(value, buffer, out int written);
            stream.Write(buffer);
            return written;
```

src/ImageSharp/Formats/Pbm/BufferedReadStreamExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        /// <summary>
+        /// Skip over any whitespace or any comments.
+        /// </summary>
+        public static void SkipWhitespaceAndComments(this BufferedReadStream stream)
+        {
+            bool isWhitespace;
+            do
+            {
+                int val = stream.ReadByte();
```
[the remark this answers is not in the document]
ynse01: For Plain encoding this method runs for every pixel, so occurence is high. However, most implementations only write a single whitespace character between pixels.

So, in most invocations you'll end up running the outer loop only twice.

tests/ImageSharp.Tests/Formats/Pbm/PbmDecoderTests.cs
ynse01: Sure, see PbmDecoderTests for this addition.

Also, I wrote a more eleborate description of the image formats in the code documentation of the PbmEncoder and PbmDecoder. Including a link to the specification.

src/ImageSharp/Formats/Pbm/BinaryEncoder.cs
ynse01: Indeed, I corrected all the Encoders. Decoders where OK.

src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
```diff
@@ -0,0 +1,193 @@
...
+                    int bit = startBit;
+                    startBit = 0;
+                    for (; bit < 8; bit++)
+                    {
+                        bool bitValue = (raw & (0x80 >> bit)) != 0;
+                        rowSpan[x] = bitValue ? black : white;
+                        x++;
+                        if (x == width)
+                        {
+                            startBit = (bit + 1) % 8;
```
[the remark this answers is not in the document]
ynse01: Thanks for the tip !

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        /// <summary>
+        /// Processes the ppm header.
+        /// </summary>
+        /// <param name="stream">The input stream.</param>
+        private void ProcessHeader(BufferedReadStream stream)
+        {
+            byte[] buffer = new byte[2];
+
+            int bytesRead = stream.Read(buffer, 0, 2);
```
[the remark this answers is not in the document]
ynse01: stream.Read is available everywhere, no compiler issues encountered.

src/ImageSharp/Formats/Pbm/PbmEncoderCore.cs
```diff
@@ -0,0 +1,176 @@
...
+                {
+                    signature = "P6";
+                }
+            }
+
+            return signature;
+        }
+
+        private void WriteHeader(Stream stream, string signature, Size pixelSize)
+        {
```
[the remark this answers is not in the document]
ynse01: I'm all in for allocation free code. In server side application you never know what will be a hot method. Applied your suggestion.

src/ImageSharp/Formats/Pbm/BinaryEncoder.cs
```diff
@@ -0,0 +1,203 @@
...
+                        if (rowSpan[x].PackedValue < 128)
+                        {
+                            value |= 0x80 >> i;
+                        }
+
+                        x++;
+                        if (x == width)
+                        {
+                            previousValue = value;
+                            startBit = (i + 1) % 8;
```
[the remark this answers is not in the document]
ynse01: Unsigned is not an option, as .NET did not provide bitshifting operators for unsigned.

src/ImageSharp/Formats/Pbm/PbmImageFormatDetector.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <inheritdoc/>
+        public IImageFormat DetectFormat(ReadOnlySpan<byte> header) => this.IsSupportedFileFormat(header) ? PbmFormat.Instance : null;
+
+        private bool IsSupportedFileFormat(ReadOnlySpan<byte> header)
+        {
+            if (header.Length >= this.HeaderSize)
+            {
+                return header[0] == P && header[1] > Zero && header[1] < Seven;
+            }
```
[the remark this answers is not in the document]
ynse01: Applied your suggestion

src/ImageSharp/Formats/Pbm/StreamExtensions.cs
ynse01: Applied your suggestion. Thanks for pointing all these optimization out to me !

src/ImageSharp/Formats/Pbm/BufferedReadStreamExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// Read a decimal text value.
+        /// </summary>
+        /// <returns>The integer value of the decimal.</returns>
+        public static int ReadDecimal(this BufferedReadStream stream)
+        {
+            int value = 0;
+            while (true)
+            {
+                int current = stream.ReadByte() - 0x30;
+                if (current < 0 || current > 9)
```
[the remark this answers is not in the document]
ynse01: Corrected

src/ImageSharp/Formats/Pbm/BinaryEncoder.cs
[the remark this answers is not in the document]
ynse01: Unsigned is not an option, as .NET did not provide bitshifting operators for unsigned.

src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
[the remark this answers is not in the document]
ynse01: Thanks for the tip !

src/ImageSharp/Formats/Pbm/BinaryEncoder.cs
ynse01: Indeed, I corrected all the Encoders. Decoders where OK.

tests/ImageSharp.Tests/Formats/Pbm/PbmDecoderTests.cs
ynse01: Sure, see PbmDecoderTests for this addition.

Also, I wrote a more eleborate description of the image formats in the code documentation of the PbmEncoder and PbmDecoder. Including a link to the specification.

src/ImageSharp/Formats/Pbm/BinaryEncoder.cs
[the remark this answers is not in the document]
br3aker: You meant bit rotation for unsigned? Bit shifting is available for unsigned numbers.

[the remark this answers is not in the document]
ynse01: Line 182: `value |= 0x80 >> i`
Doesn't compile if `i` is unsigned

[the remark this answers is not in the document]
br3aker: Oh I see, I thought you were talking about something like  `value <<= i` for unsigned `value`.

gjoidl meant that JIT automatically can do `& 7` bithack if value is unsigned but if you`re sure that your `int` number is always positive (which it is as `i` is between `startbit` and 8) - you can manually apply this bithack via `& 7` instead of `% 8`.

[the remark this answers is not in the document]
ynse01: Indeed @br3aker , so I changed line 189 into:
`startBit = (i + 1) & 7; // Round off to below 8.`

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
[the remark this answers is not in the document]
JimBobSquarePants: Yeah we've written polyfils for non supported platforms.

JimBobSquarePants:
@ynse01 Looks like you're way ahead of me dealing with the individual files but could you do me a favor and run the following command from the repo root to updated the shared infrastructure code and then build your solution to ensure the `.gitattributes` file is copied across containing new instructions for handling the image type.

```bash
```

ynse01:
The test images should now be in Git LFS.

I can't find why the test coverage dropped. Can it be a false positive, as it reports a delta of only -1 % ? Let me know if I need to take action on the coverage.

JimBobSquarePants:
>I can't find why the test coverage dropped. Can it be a false positive, as it reports a delta of only -1 % ? Let me know if I need to take action on the coverage.

@ynse01 I'm firmly convince the code coverage is hocus pocus. 9/10 times it seems to report nonsensical results. I only keep it around to be indicative and because Enterprise types love to see those kinds of metrics.

src/ImageSharp/Formats/Pbm/PlainEncoder.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        private static void WriteGrayscale<TPixel>(Configuration configuration, Stream stream, ImageFrame<TPixel> image)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            int width = image.Width;
+            int height = image.Height;
+            MemoryAllocator allocator = configuration.MemoryAllocator;
+            using IMemoryOwner<L8> row = allocator.Allocate<L8>(width);
+            Span<L8> rowSpan = row.GetSpan();
+            Span<byte> plainSpan = stackalloc byte[width * MaxCharsPerPixelGrayscale];
```
JimBobSquarePants, line 80: We shouldn't be using `stackalloc` for larger numbers like this here and elsewhere. You're actually allocating more memory here than you use the `MemoryAllocator` to allocate the `L8` span. Might work in some environments but stackoverflow in others.

Here's a great blog post on `stackalloc` usage. https://vcsjones.dev/

ynse01: Interesting read. Reverted back to Allocator again.

src/ImageSharp/Formats/Pbm/PlainEncoder.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        private static void WriteGrayscale<TPixel>(Configuration configuration, Stream stream, ImageFrame<TPixel> image)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            int width = image.Width;
+            int height = image.Height;
+            MemoryAllocator allocator = configuration.MemoryAllocator;
+            using IMemoryOwner<L8> row = allocator.Allocate<L8>(width);
+            Span<L8> rowSpan = row.GetSpan();
+            IMemoryOwner<byte> plainMemory = allocator.Allocate<byte>(width * MaxCharsPerPixelGrayscale);
```
antonfirsov, line 79: `using` is missing all cases of cf8e184.

tests/ImageSharp.Tests/Formats/Pbm/RoundTripTests.cs
```diff
@@ -0,0 +1,44 @@
...
+            ImageComparer.Exact.VerifySimilarity(originalImage, encodedImage);
+        }
+
+        private Image<TPixel> RoundTrip<TPixel>(Image<TPixel> originalImage)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            using var decodedStream = new MemoryStream();
+            originalImage.SaveAsPbm(decodedStream);
+            decodedStream.Seek(0, SeekOrigin.Begin);
+            var encodedImage = (Image<TPixel>)Image.Load(decodedStream);
```
antonfirsov, line 40:
```suggestion
            var encodedImage = Image.Load<TPixel>(decodedStream);
```

tests/ImageSharp.Tests/Formats/Pbm/PbmDecoderTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        [WithFile(GrayscaleBinary, PixelTypes.L8, true)]
+        [WithFile(GrayscaleBinaryWide, PixelTypes.L16, true)]
+        [WithFile(RgbPlainNormalized, PixelTypes.Rgb24, false)]
+        [WithFile(RgbBinary, PixelTypes.Rgb24, false)]
+        public void DecodeReferenceImage<TPixel>(TestImageProvider<TPixel> provider, bool isGrayscale)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            using Image<TPixel> image = provider.GetImage();
+
+            image.CompareToReferenceOutput(provider, grayscale: isGrayscale);
```
antonfirsov, line 88: Helps reasoning about the output when testing locally with `Debug` builds:

```suggestion
            image.DebugSave(provider);
            image.CompareToReferenceOutput(provider, grayscale: isGrayscale);
```

tests/ImageSharp.Tests/Formats/Pbm/PbmEncoderTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            where TPixel : unmanaged, IPixel<TPixel> => TestPbmEncoderCore(provider, PbmColorType.BlackAndWhite, PbmEncoding.Plain);
+
+        [Theory]
+        [WithFile(BlackAndWhiteBinary, PixelTypes.Rgb24)]
+        public void PbmEncoder_P4_Works<TPixel>(TestImageProvider<TPixel> provider)
+            where TPixel : unmanaged, IPixel<TPixel> => TestPbmEncoderCore(provider, PbmColorType.BlackAndWhite, PbmEncoding.Binary);
+
+        /* Disabled as Magick throws an error reading the input image
+        [Theory]
+        [WithFile(GrayscalePlain, PixelTypes.Rgb24)]
```
antonfirsov, line 98: Might worth to also feed this to `PbmColorImageCanRoundTrip` since normal don't work on it.

src/ImageSharp/Formats/Pbm/PbmConstants.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Contains PBM constant values defined in the specification.
+    /// </summary>
+    internal static class PbmConstants
+    {
+        /// <summary>
+        /// The maximum allowable pixel value of a ppm image.
+        /// </summary>
+        public const ushort MaxLength = 65535;
```
antonfirsov, line 16: This is an unused. Is it needed?

antonfirsov (commented):
Just a few more points and this good to merge!

@JimBobSquarePants can you take a look at the shared-infrastructure module update? Is that just a syncup with latest upstream?

JimBobSquarePants:
>@JimBobSquarePants can you take a look at the shared-infrastructure module update? Is that just a syncup with latest upstream?

@antonfirsov yeah. We needed to do so to get the GitLFS instructions for the new filetypes.

src/ImageSharp/Formats/Pbm/PbmConstants.cs
ynse01: It should have been used to cap the MaxPixelValue, good catch ! Used it now.

tests/ImageSharp.Tests/Formats/Pbm/PbmEncoderTests.cs
ynse01: Indeed, added the Grayscale images

tests/ImageSharp.Tests/Formats/Pbm/PbmDecoderTests.cs
ynse01: Sure, I interpreted this as left over debug code initially. Hence, I removed it before sending the pull request. Put them back now.

tests/ImageSharp.Tests/Formats/Pbm/RoundTripTests.cs
ynse01: OK, done

src/ImageSharp/Formats/Pbm/PlainEncoder.cs
ynse01: Oops, placed them back again.

src/ImageSharp/Formats/Pbm/PbmConstants.cs
JimBobSquarePants: I'm curious about this constant. It appears to me that it is used for clamping during encoding but why not make `MaxPixelValue` in the metadata a `ushort`?

We're also automatically the value to 255 in the `PbmMetadata` constructor. I'm assuming that is to set the default to the non-wide color type.

Looking at how  `MaxPixelValue` is used it appears that any value over 255 indicates to use wider bit depth and we simply read/write the value in the header. I would suggest that instead of using this property in the metadata we should expand the `PbmColorType` type to add two new entries. `GrayscaleWide` and `RgbWide`.

ynse01: MaxPixelValue can be any value. For example, in GrayscalePlain it is set to 15.
If the pixels are left untouched, as I do in my implementation, the image would be rather dark. I deliberately left the upscaling of the pixel values up to the user of ImageSharp. Most application though upscale the pixel values to the full range of the data structure, when reading the image.

Because I chose to leave the pixel values untouched, I need to indicate to the user what scaling factor they need to apply, if they wish to do so. Hence the addition of MaxPixelValue to the Metadata.

Let me know what behavior you guys prefer.

JimBobSquarePants: > Most application though upscale the pixel values to the full range of the data structure, when reading the image.

Wouldn't not doing this then mean that any application of a change to the image would render a different output to expected?

E.G

- Load Image
- Set [0,0] to color
- Save Image
- Load Image
- Read [0, 0]
- Compare to expected = Fail.

Would that be the correct summation? If so, we want to avoid people having to do any additional calculations when changing pixel values.

ynse01: As long as we do remain _consistently_ on the upscaling in both Encoding and Decoding, your example will work as expected.

As long as you stay within the range of [0..MaxPixelValue] the value will round trip, even in other applications (that do upscaling). 
If you go above the MaxPixelValue however, some other applications will refuse to load the image.

From this library I do not know what the intent of the user is. Some would want to retain the original pixel values. Some however, want the upscaling behavior they see in other applications.
By not upscaling automatically, I'm offering both alternatives (be it with an extra Mutate call for the second case).

Would you like to make this upscaling choice a property of IPbmEncoderOptions ?
And if so, what would be the default?

ynse01: I've implemented pixel value scaling. Didn't made it configurable, which aligns with other Formats. I users request configurability in the future, we can easily add it later.

tests/ImageSharp.Tests/Formats/Pbm/PbmDecoderTests.cs
antonfirsov: `DebugSave()` is no-op when running Release configuration, so it doesn't hurt to keep them in all tests, but helps to see the output immediately in case a regression happens.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
```diff
@@ -0,0 +1,199 @@
...
+                {
+                    if (this.MaxPixelValue is not 255 and not 65535)
+                    {
+                        needsUpscaling = true;
+                    }
+                }
+
+                return needsUpscaling;
+            }
+        }
```
antonfirsov, line 71:
```suggestion
        private bool NeedsUpscaling =>
            this.ColorType != PbmColorType.BlackAndWhite &&
            this.MaxPixelValue is not 255 and not 65535;
```

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
```diff
@@ -0,0 +1,199 @@
...
+                PlainDecoder.Process(this.Configuration, pixels, stream, this.ColorType, this.MaxPixelValue);
+            }
+        }
+
+        private void ProcessUpscaling<TPixel>(Image<TPixel> image)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            int maxAllocationValue = (this.MaxPixelValue > 255) ? 65535 : 255;
+            float factor = maxAllocationValue / this.MaxPixelValue;
+            image.Mutate(x => x.Brightness(factor));
```
antonfirsov, line 196: Do we have tests that exercising this logic? Is the scaled output consistent with other common implementations?

antonfirsov (commented):
Noticed a few more points after the recent changes.

antonfirsov:
After checking out the branch, the following diff automatically appears for me:

image

Isn't this the LFS update resulting from 9599efe5201dc516673000adecf1dd84b6976cf5 we should commit & push? Or is this some anomaly on my machine only?

src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
```diff
@@ -0,0 +1,193 @@
...
+                ProcessBlackAndWhite(configuration, pixels, stream);
+            }
+        }
+
+        private static void ProcessGrayscale<TPixel>(Configuration configuration, Buffer2D<TPixel> pixels, BufferedReadStream stream)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            int width = pixels.Width;
+            int height = pixels.Height;
+            int bytesPerPixel = 1;
```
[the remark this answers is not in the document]
ynse01: Indeed, done on all encoders and decoders.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
ynse01: Indeed, this looks better.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
ynse01: I have seen scaled output in xnConvert and Irfanview.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
```diff
@@ -0,0 +1,199 @@
...
+                case '6':
+                    // Binary PPM format: 3 components per pixel, in binary integers.
+                    this.ColorType = PbmColorType.Rgb;
+                    this.Encoding = PbmEncoding.Binary;
+                    break;
+                case '7':
+                // PAM image: sequence of images.
+                // Not implemented yet
+                default:
+                    throw new NotImplementedException("TODO");
```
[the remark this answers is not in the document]
ynse01: Yes, this was a leftover.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// <param name="stream">The input stream.</param>
+        private void ProcessHeader(BufferedReadStream stream)
+        {
+            Span<byte> buffer = stackalloc byte[2];
+
+            int bytesRead = stream.Read(buffer);
+            if (bytesRead != 2 || buffer[0] != 'P')
+            {
+                // Empty or not an PPM image.
+                throw new InvalidImageContentException("TODO");
```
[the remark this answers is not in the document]
ynse01: Yes it is, this was a leftover. Although, the PbmImageFormatDector should catch this already.

ynse01:
> After checking out the branch, the following diff automatically appears for me:
> 
> image
> 
> Isn't this the LFS update resulting from [9599efe](https://github.com/SixLabors/ImageSharp/commit/9599efe5201dc516673000adecf1dd84b6976cf5) we should commit & push? Or is this some anomaly on my machine only? @JimBobSquarePants @brianpopow?

@antonfirsov No, this is not specific to your machine. These files reside both in regualr GIT and in LFS now. I applied `git lfs migrate` on them. Please check if this helps on your machine also.

brianpopow:
> > After checking out the branch, the following diff automatically appears for me:
> > image
> > Isn't this the LFS update resulting from [9599efe](https://github.com/SixLabors/ImageSharp/commit/9599efe5201dc516673000adecf1dd84b6976cf5) we should commit & push? Or is this some anomaly on my machine only? @JimBobSquarePants @brianpopow?
> 
> @antonfirsov No, this is not specific to your machine. These files resided both in LFS and as regular GIT files. I applied `git lfs migrate` on them, which should remove them as regulat GIT files. Please check if this helps on your machine also.

I have just checked out the branch without any issue. @antonfirsov can you confirm this?

tests/ImageSharp.Tests/Formats/Pbm/PbmEncoderTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                var encoder = new PbmEncoder { ColorType = colorType, Encoding = encoding };
+
+                using (var memStream = new MemoryStream())
+                {
+                    image.Save(memStream, encoder);
+                    memStream.Position = 0;
+                    using (var encodedImage = (Image<TPixel>)Image.Load(memStream))
+                    {
+                        TgaTestUtils.CompareWithReferenceDecoder(provider, encodedImage, useExactComparer, compareTolerance);
```
brianpopow, line 138: maybe we should put this method in some common place for testing?

tests/ImageSharp.Tests/Formats/Pbm/PbmEncoderTests.cs
```diff
@@ -0,0 +1,144 @@
...
+        public void PbmEncoder_P2_Works<TPixel>(TestImageProvider<TPixel> provider)
+            where TPixel : unmanaged, IPixel<TPixel> => TestPbmEncoderCore(provider, PbmColorType.Grayscale, PbmEncoding.Plain);
+        */
+
+        [Theory]
+        [WithFile(GrayscaleBinary, PixelTypes.Rgb24)]
+        public void PbmEncoder_P5_Works<TPixel>(TestImageProvider<TPixel> provider)
+            where TPixel : unmanaged, IPixel<TPixel> => TestPbmEncoderCore(provider, PbmColorType.Grayscale, PbmEncoding.Binary);
+
+        /* Disabled as Magick throws an error reading the input image
```
brianpopow, line 108: I wonder why magick fails here, @ynse01 are you sure this image is valid? Maybe we should raise an issue in the magick repo then?

brianpopow: I think there is a missing line feed

ynse01: I took these examples from the offcial specification website and the linux man page of libnetpbm.
The specification makes no distiction btween different types of whitespace. So, line feeds are not significant.
I consider these failures a bug in Magick. I'll post an issue about it on their repo.

tests/ImageSharp.Tests/Formats/Pbm/PbmEncoderTests.cs
ynse01: Tiff has a very similar utility also. So, now we have three formats using similar code.
Please advise on a location of this utility class and I'll refactor all three of them to a single location.

brianpopow: I would put it in `ImageSharp.Tests\TestUtilities\ImageComparison`. Maybe call the class `ImageComparingUtils.cs`?

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
ynse01: Filed an issue to Magicks repo [here](https://github.com/ImageMagick/ImageMagick/issues/4546)

ynse01: While we wait for Magick to fix the bug, I added some Magick compatble versions. Now we have a reference test with Magick for every ColorType and Encoding combination..

tests/ImageSharp.Tests/Formats/Pbm/PbmEncoderTests.cs
ynse01: Done this, no code changes required to the TestUtils.

tests/ImageSharp.Tests/Formats/Pbm/PbmEncoderTests.cs
ynse01: As above, while we wait for Magick to fix the bug, I placed some Magick compatible input images.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
antonfirsov: Looks like they are pushing back and nominating Netpbm as a de facto reference implementation. Feels like we can't do anything else but conform to Netpbm for maximum compatibility.

ynse01: Just to be clear, I have enabled the reference tests with Magick for all image type. To do this, I added a Magick compatible variant of the offending images. 

Turns out Magick expects a closing space after the last pixel. My code doesn't care and accepts image with and without this space at the end.

So, my code can handle both variants. To proof this, I added both variants to the RoundTripTest and others tests as well.

I propose to keep this behavior, as I consider it more user-friendly then Magick and NetPbm.

brianpopow: > I propose to keep this behavior, as I consider it more user-friendly then Magick and NetPbm.

I also think this is a good idea. I tested those files with gimp, irfanview and xnview, which also dont complain decoding it.
@antonfirsov, @JimBobSquarePants  what are your opinions on this?

antonfirsov: Ah sorry got confused and thought we are producing non-compliant outputs, nevermind me.

JimBobSquarePants: I'm with keeping the more friendly approach. It's an odd stance to have to demand the trailing whitespace on the last input. Most decoders handle missing EOF indicators well enough (Including several in _ImageMagick_) .

src/ImageSharp/Formats/Pbm/PbmDecoder.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        /// <inheritdoc/>
+        public async Task<IImageInfo> IdentifyAsync(Configuration configuration, Stream stream, CancellationToken cancellationToken)
+        {
+            Guard.NotNull(stream, nameof(stream));
+
+            // The introduction of a local variable that refers to an object the implements
+            // IDisposable means you must use async/await, where the compiler generates the
+            // state machine and a continuation.
+            var decoder = new PbmDecoderCore(configuration);
```
JimBobSquarePants, line 78: But `PbmDecoderCore` doesn't implement `IDisposable`. You can pass the `Task` directly like other decoders.

https://github.com/SixLabors/ImageSharp/blob/4ee7830105afa04629a1d9261ae26786a08584ef/src/ImageSharp/Formats/Png/PngDecoder.cs#L31-L36

src/ImageSharp/Formats/Pbm/PbmConstants.cs
JimBobSquarePants: I'm still confused by this scaling stuff. 

We're reading the `MaxPixelValue` and then adjusting the brightness of the input pixels by that factor (_We're introducing parallelism and additional `TPixel/Vector4` conversion here which we've avoided elsewhere. It would be much better to operate on the byte row directly_).

We can now operate on the decoded pixels without additional per operation calculations - great!

However, when we encode, we pass the original `MaxPixelValue` back down to the encoder while leaving the original pixel values themselves unchanged. This means that subsequent decodes using our decoder would continue to run the upscaling each time which would be destructive. 

Please correct me if I'm wrong but as far as I can see `MaxPixelValue` has two functions. 
- To determine the bit depth of the encoded image
- To set the brightness of the image.

I still think the additional enum values are the best way to go with the correct scaling applied for decode/encode for each color/encoding type combination.

ynse01: My understanding of MaxPixelValue is the same as yours. I'll add some extra checks in the PbmRoundTripTest to make sure we're not messing with the MaxPixelValue.

The behavior I see in XnConvert and Irfanview, is what I would like to copy here. Decoded the pixels are upscaled to capacity for the pixel type. En during Encoding the scaled pixel values are written, together with a modified MaxPixelValue.

Now for the enums, I agree that enums are better then the raw MaxPixelValue form the image. Here, let's also look at [PAM](http://netpbm.sourceforge.net/doc/pam.html) images, as these are a likely addition lateron. PAM adds several pixel types, making a single enum a bit big. I therefore propose 2 enums:

1. ColorType, having the following values: BlackAndWhite, GrayScale, Rgb, Grayscale_Alpha, Rgb_Alpha.
2. PixelDataType (PAM calls these TypleTypes), having values: Bit, Byte, Short, Float and Double

Combining it into a single enum will give you 17 entries (B&W-Bit + 4 times 4 for the other combinations).

@JimBobSquarePants: Let me know your enum preference.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
ynse01: Let me double check if our Encoded images load into both Magick and NetPbm, which might not be the case now. Anyway, I'll add a check for the EOF indicator in the tests.

src/ImageSharp/Formats/Pbm/PbmConstants.cs
JimBobSquarePants: 17... Yeah best splitting them up. That's what we do with png.  We use [utilities](https://github.com/SixLabors/ImageSharp/blob/4ee7830105afa04629a1d9261ae26786a08584ef/src/ImageSharp/Formats/Png/PngEncoderOptionsHelpers.cs) to ensure valid combinations of the two enums are in use.

ynse01: @JimBobSquarePants  I have pushed the Enum changes, let me know your comments.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
ynse01: After adding the EOF indicator, all ColorTypes and Encodings open ciorrectly in both NetPbm and Magick.

src/ImageSharp/Formats/Pbm/PbmDecoder.cs
ynse01: Indeed, pushed the changes

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
ynse01: OK, checked Encoding all ColorTypes and Encodings and load them back into both NetPbm and Magick. All are excepted without warnings now.

JimBobSquarePants:
@ynse01 If we can get these conflicting tests fixed I'll be more than happy to merge this PR. We an review the brightness adjustment approach post-merge. 

Really solid bit of coding! Thanks for taking the time to submit it!

ynse01:
> @ynse01 If we can get these conflicting tests fixed I'll be more than happy to merge this PR. We an review the brightness adjustment approach post-merge.
> 
> Really solid bit of coding! Thanks for taking the time to submit it!

Thanks for taking in my changes!

I've merges latest master into this PR, that should resolve the conflicts.

PR #1853, opened 2021-11-27
Jpeg codebase cleanup

br3aker:
### Prerequisites

### Description

**Right now this PR exist purely to test crashes during testing, work is still in progress.**

This PR consists of a lot of small fixes to the jpeg codebase: general cleanup, redundant/unused code removal and small performance tweaks via removing unnecessary checks.


src/ImageSharp/Formats/Jpeg/Components/Encoder/LuminanceForwardConverter{TPixel}.cs
```diff
@@ -15,39 +16,63 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Encoder
...
     {
+        /// <summary>
+        /// Number of pixels processed per single <see cref="Convert(int, int, ref RowOctet{TPixel})"/> call
+        /// </summary>
+        private const int PixelsPerSample = 8 * 8;
+
+        /// <summary>
+        /// <see cref="Size"/> of sampling area from given frame pixel buffer.
+        /// </summary>
+        private static readonly Size SampleSize = new Size(8, 8);
```
gfoidl, line 27: `Size` is simple struct (just containing 2 ints), so it may be better to just use a property / factory here:
```suggestion
        private static Size SampleSize => new Size(8, 8);
```
so for each access a new struct is created, but as it's small the JIT could enregister and perform other optimizations, that aren't possible by reading the struct from memory.

In short with the property we
* can avoid a memory read
* allow the JIT to perform optimizations

The following benchmark is _not_ the most representative one, but (at least for .NET 6 and I believe starting from .NET Core 3.1) one can see where it goes by looking at the disassembly.
<details>
  <summary>Benchmark code</summary>

```c#

BenchmarkDotNet.Running.BenchmarkRunner.Run<Bench>();

[DisassemblyDiagnoser]
[ShortRunJob]
public class Bench
{
    private static readonly Size s_samplesize = new(8, 8);
    private static Size SampleSize => new(8, 8);

    [Benchmark(Baseline = true)]
    public int StaticReadonly() => s_samplesize.Width + s_samplesize.Height;

    [Benchmark]
    public int Property() => SampleSize.Width + SampleSize.Height;
}

public struct Size
{
    public int Width  { get; set; }
    public int Height { get; set; }

    public Size(int width, int height) => (this.Width, this.Height) = (width, height);
}
```
</details>

```assembly
; .NET 6.0.0 (6.0.21.52210), X64 RyuJIT

; Bench.StaticReadonly()
       mov       rax,25118002F10
       mov       rax,[rax]
       mov       edx,[rax+8]
       add       edx,[rax+0C]
       mov       eax,edx
       ret
; Total bytes of code 22

; Bench.Property()
       mov       eax,10
       ret
; Total bytes of code 6
```

With the property the applies all possible optimizations, while with the `static readonly` it must read from memory and can't do optimizations.

src/ImageSharp/Formats/Jpeg/Components/GenericBlock8x8.Generated.cs
```diff
@@ -1,23 +0,0 @@
-﻿// Copyright (c) Six Labors.
```
gfoidl, line 1: `GenericBlock8x8.Generated.tt` is an empty file -- should it be deleted?

src/ImageSharp/Formats/Jpeg/Components/Encoder/LuminanceForwardConverter{TPixel}.cs
br3aker: I didn't even know jit ca do this with small structs, thanks!

src/ImageSharp/Formats/Jpeg/Components/GenericBlock8x8.Generated.cs
br3aker: It should, looks like I forgot about it, thanks!

br3aker: This might be a github diff bug or something but I couldn't find `GenericBlock8x8.Generated.tt` file in my repo:

image

JimBobSquarePants: Should be under `Format\Jpeg\Components`

br3aker: Yeah but screen above is `Format\Jpeg\Components`, here's another one directly from VS:
image

Sorry for my agnorance but what those `.Generated.tt` files mean exactly? First time encountering those.

br3aker:
Current progress is big enough for review, there's still some work left - I will work on it as a separate PR.

src/ImageSharp/Formats/Jpeg/Components/GenericBlock8x8.Generated.cs
JimBobSquarePants: They're T4 templates generating partials of the named types. If you expand them you see the generated file. e.g. `Block8x8F.Generated.cs`. 

That empty one looks like a leftover from previous cleanup and is safe to delete.

JimBobSquarePants (approved):
Nice job! I've been reviewing as you've been working through. Looks much better.

br3aker:
Wow that was fast, thanks!

JimBobSquarePants:
I've been struggling to find something useful to do so have been focusing on keeping up with PRs.

src/ImageSharp/Formats/Jpeg/Components/GenericBlock8x8.Generated.cs
br3aker: Completely missed your response - thanks for clarifying .tt stuff! `GenericBlock8x8.Generated.tt` seems to be deleted now (I did nothing after this review, looks like a diff tool bug or whatever).

PR #1861, opened 2021-11-30
Png - Preserve Pixel Format for Non-generic decode.

### LINKED ISSUE #1813: Decode PNG to source stream's pixel type
PNG decoder is not CPU intensive compared to other codecs, but we can save some memory if there is no alpha or if the PNG is grayscale. We can instantiate the `Image<T>` after reading the header chunk depending on `header.ColorType` and `header.BitDepth`.

It would also help preserving the original images precision in case of `header.BitDepth == 16`.

JimBobSquarePants:
### Prerequisites

### Description

Fixes #1813 by determining the most appropriate pixel format to decode to based upon the detected color type and bit depth.

I could have avoided the identify/reset/decode route and gained a little more performance but `Identify` is so quick anyway it seemed like the simplest approach.


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -364,11 +371,10 @@ private void ReadPhysicalChunk(ImageMetadata metadata, ReadOnlySpan<byte> data)
         /// <param name="pngMetadata">The metadata to read to.</param>
         /// <param name="data">The data containing physical data.</param>
         private void ReadGammaChunk(PngMetadata pngMetadata, ReadOnlySpan<byte> data)
-        {
+
             // The value is encoded as a 4-byte unsigned integer, representing gamma times 100000.
             // For example, a gamma of 1/2.2 would be stored as 45455.
-            pngMetadata.Gamma = BinaryPrimitives.ReadUInt32BigEndian(data) / 100_000F;
-        }
+            => pngMetadata.Gamma = BinaryPrimitives.ReadUInt32BigEndian(data) / 100_000F;
```
gfoidl, line 377: Flip to multiplications, as it's cheaper than division?
```suggestion
            => pngMetadata.Gamma = BinaryPrimitives.ReadUInt32BigEndian(data) * 1e-5F;
```

antonfirsov (commented):
> but Identify is so quick anyway it seemed like the simplest approach.

I'm not sure if it's always true. There is some unnecessary IO cost reading the stream twice when the file is large and/or there are `*Text` chunks not being skipped on `ignoreMetadata`. This may hurt scalability, and it doesn't seem to be very hard to fix:

We can introduce a field in `PngDecoderCore` that asks to skip everything except the absolute minimum needed to determine the pixel type (Header + Transparency?), and **stop** when the essential chunks are processed or when a Data chunk is detected.

Later we can improve the `IImageDecoderInternals` infrastructure to support these scenarios, but that's far beyond the essential feature work we need for Png.

JimBobSquarePants:
>I'm not sure if it's always true. There is some unnecessary IO cost reading the stream twice when the file is large and/or there are *Text chunks not being skipped on ignoreMetadata. This may hurt scalability, and it doesn't seem to be very hard to fix:

I've changed the behavior of the chunk reader to avoid reading data from anything other than the header and transparency chunks when required. This also allows skipping the entire chunk. It's a little hacky but should be fast enough for now.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -241,29 +259,79 @@ public IImageInfo Identify(BufferedReadStream stream, CancellationToken cancella
...
+                                if (this.colorMetadataOnly)
+                                {
+                                    this.SkipChunkDataAndCrc(chunk);
+                                    break;
+                                }
+
                                 this.ReadGammaChunk(pngMetadata, chunk.Data.GetSpan());
                                 break;
                             case PngChunkType.Data:
                                 this.SkipChunkDataAndCrc(chunk);
```
antonfirsov, line 283: [Spec says](http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.tRNS):

> When present, the tRNS chunk must precede the first IDAT chunk, and must follow the PLTE chunk, if any.

It means that if we reach Data, we are free to quit the loop. Forwarding the file position beyond the data chunk may have IO overhead we can easily avoid here.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -241,29 +259,79 @@ public IImageInfo Identify(BufferedReadStream stream, CancellationToken cancella
...
+                                break;
+                            case PngChunkType.Transparency:
+                                byte[] alpha = new byte[chunk.Length];
+                                chunk.Data.GetSpan().CopyTo(alpha);
+                                this.paletteAlpha = alpha;
+                                this.AssignTransparentMarkers(alpha, pngMetadata);
+
+                                if (this.colorMetadataOnly && this.isHeaderChunkReached)
```
antonfirsov, line 291: According to the spec, IDAT is always the first, so - in theory - we don't need  to track if we reached it. I wonder if there are malformed PNG-s in the wild which do not follow this rule.

@brianpopow any thoughts?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -241,29 +259,79 @@ public IImageInfo Identify(BufferedReadStream stream, CancellationToken cancella
...
+                                {
+                                    // Quick exit
+                                    this.isEndChunkReached = true;
```
antonfirsov, line 294: I wonder if we should just `break` out of the loop instead of abusing this flag?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
brianpopow: You mean `IHDR` is always first I guess? I will go through my collection of "weird and broken images" to check if i can find any which has not `IHDR` first.

antonfirsov: > You mean IHDR is always first I guess? 

Yes, sorry for the typo.

JimBobSquarePants:
@antonfirsov @brianpopow I'll just follow the strict specification here and optimize accordingly. The worst that can happen is that they decode to the incorrect pixel format.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
brianpopow: I could not find any png which do not start with `IHDR`. Nonetheless I am still in favor of keeping it as it is, just to be sure,  because in theory such png can be created.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: `goto` baby; `goto`.

antonfirsov (approved):
LGTM. I think it's possible to achieve the same without `goto` will do it in a separate cleanup PR.

JimBobSquarePants:
>LGTM. I think it's possible to achieve the same without goto will do it in a separate cleanup PR.

Whole thing needs a cleanup. We should be skipping all chunks we don't currently read on decode but are instead reading the data.

PR #1864, opened 2021-12-02
Automatic SSE color converters based on Vector API

br3aker:
### Prerequisites

### Description
Automatic AVX/SSE implementation for jpeg color converters based on `Vector` API. This PR uses existing code but with variable vector size rather than explicit `8`.

Master
|                              Method |                Job |       Mean |     Error |    StdDev | Ratio |
|------------------------------------ |------------------- |-----------:|----------:|----------:|------:|
|        'Baseline 4:4:4 Interleaved' | 1. No HwIntrinsics | 161.405 ms | 1.0435 ms | 0.8714 ms |  1.00 |
|        'Baseline 4:4:4 Interleaved' |             2. SSE |  29.430 ms | 0.1136 ms | 0.0949 ms |  0.18 |
|        'Baseline 4:4:4 Interleaved' |             3. AVX |  11.229 ms | 0.0369 ms | 0.0346 ms |  0.07 |
|                                     |                    |            |           |           |       |
|        'Baseline 4:2:0 Interleaved' | 1. No HwIntrinsics |  69.386 ms | 0.4476 ms | 0.4187 ms |  1.00 |
|        'Baseline 4:2:0 Interleaved' |             2. SSE |  14.439 ms | 0.0582 ms | 0.0516 ms |  0.21 |
|        'Baseline 4:2:0 Interleaved' |             3. AVX |   8.542 ms | 0.0390 ms | 0.0365 ms |  0.12 |
|                                     |                    |            |           |           |       |
|        'Baseline 4:0:0 (grayscale)' | 1. No HwIntrinsics |  12.338 ms | 0.2449 ms | 0.2405 ms |  1.00 |
|        'Baseline 4:0:0 (grayscale)' |             2. SSE |   3.307 ms | 0.0115 ms | 0.0108 ms |  0.27 |
|        'Baseline 4:0:0 (grayscale)' |             3. AVX |   1.566 ms | 0.0063 ms | 0.0056 ms |  0.13 |
|                                     |                    |            |           |           |       |
| 'Progressive 4:2:0 Non-Interleaved' | 1. No HwIntrinsics | 118.681 ms | 0.5633 ms | 0.5269 ms |  1.00 |
| 'Progressive 4:2:0 Non-Interleaved' |             2. SSE |  24.615 ms | 0.1012 ms | 0.0947 ms |  0.21 |
| 'Progressive 4:2:0 Non-Interleaved' |             3. AVX |  13.206 ms | 0.0487 ms | 0.0407 ms |  0.11 |

PR
|                              Method |                Job |       Mean |     Error |    StdDev | Ratio |
|------------------------------------ |------------------- |-----------:|----------:|----------:|------:|
|        'Baseline 4:4:4 Interleaved' | 1. No HwIntrinsics | 219.807 ms | 1.1166 ms | 0.9898 ms |  1.00 |
|        'Baseline 4:4:4 Interleaved' |             2. SSE |  28.429 ms | 0.1111 ms | 0.0985 ms |  0.13 |
|        'Baseline 4:4:4 Interleaved' |             3. AVX |  11.344 ms | 0.0534 ms | 0.0474 ms |  0.05 |
|                                     |                    |            |           |           |       |
|        'Baseline 4:2:0 Interleaved' | 1. No HwIntrinsics |  91.371 ms | 0.1973 ms | 0.1749 ms |  1.00 |
|        'Baseline 4:2:0 Interleaved' |             2. SSE |  13.635 ms | 0.0617 ms | 0.0577 ms |  0.15 |
|        'Baseline 4:2:0 Interleaved' |             3. AVX |   8.729 ms | 0.1201 ms | 0.1065 ms |  0.10 |
|                                     |                    |            |           |           |       |
|        'Baseline 4:0:0 (grayscale)' | 1. No HwIntrinsics |  12.822 ms | 0.0957 ms | 0.0895 ms |  1.00 |
|        'Baseline 4:0:0 (grayscale)' |             2. SSE |   3.299 ms | 0.0114 ms | 0.0107 ms |  0.26 |
|        'Baseline 4:0:0 (grayscale)' |             3. AVX |   1.551 ms | 0.0086 ms | 0.0080 ms |  0.12 |
|                                     |                    |            |           |           |       |
| 'Progressive 4:2:0 Non-Interleaved' | 1. No HwIntrinsics | 176.093 ms | 0.4813 ms | 0.4502 ms |  1.00 |
| 'Progressive 4:2:0 Non-Interleaved' |             2. SSE |  22.979 ms | 0.0551 ms | 0.0430 ms |  0.13 |
| 'Progressive 4:2:0 Non-Interleaved' |             3. AVX |  13.356 ms | 0.0924 ms | 0.0864 ms |  0.08 |

SSE speedup might not be that big but it's a nice 1ms saving for each case.

Don't look at scalar implementation performance 'downgrade' - BenchmarkDotNet decided to do a 1 iteration run for those so it's irrelevant.

----------------------------------------------------------------

PR also includes tests and **very little** performance tweaks - I've removed some if checks from converters for simd size as it's guaranteed that JPEG images have a width divisible by 8 without a remainder as it consists of 8x8 blocks. Once again, speedup is not that big but it's always nice to be a little bit faster with less code.

Master
|        'Baseline 4:4:4 Interleaved' | 11.504 ms | 0.0719 ms | 0.0673 ms |
|        'Baseline 4:2:0 Interleaved' |  8.605 ms | 0.0372 ms | 0.0348 ms |
|        'Baseline 4:0:0 (grayscale)' |  1.575 ms | 0.0054 ms | 0.0042 ms |
| 'Progressive 4:2:0 Non-Interleaved' | 13.207 ms | 0.0472 ms | 0.0369 ms |

PR
|        'Baseline 4:4:4 Interleaved' | 11.545 ms | 0.0444 ms | 0.0394 ms |
|        'Baseline 4:2:0 Interleaved' |  8.451 ms | 0.0450 ms | 0.0421 ms |
|        'Baseline 4:0:0 (grayscale)' |  1.519 ms | 0.0017 ms | 0.0015 ms |
| 'Progressive 4:2:0 Non-Interleaved' | 13.031 ms | 0.0201 ms | 0.0168 ms |

For whatever reason first image benchmark is **very** unstable, it goes from 11.0ms up to 11.5ms on master and this PR. Other images have a stable speedup (at least on simple benchmark config, multi-intrinsic one is a bit unstable for all test cases). Maybe it's caused by the GC. Baseline 444 image allocates 6 times more bytes than Baseline 420 and Baseline 400.


brianpopow:
> Don't look at scalar implementation performance 'downgrade' - BenchmarkDotNet decided to do a 1 iteration run for those so it's irrelevant.

Maybe its worth changing the benchmark Config from `.WithLaunchCount(1)` to `.WithLaunchCount(3)` to get more stable results? It will take longer of course.

br3aker:
> Maybe its worth changing the benchmark Config from .WithLaunchCount(1) to .WithLaunchCount(3) to get more stable results? It will take longer of course.

Scalar code wasn't touched by this PR as it's purely about automatic SSE converters and AVX converters optimization. So it's actually irrelevant but we have a scalar/sse/avx config so I've decided  to leave it as is. AFAIK BenchmarkDotNet don't want to benchmark code with such high magnitude (and we don't need it either so....)

br3aker:
It's actually phenomenal how much performance was gained via intrinsics!

JimBobSquarePants:
@br3aker Looks like we've got some failing tests in NET472

https://github.com/SixLabors/ImageSharp/runs/4393171045?check_suite_focus=true#step:12:78

br3aker:
Will debug a bit later. Error is strange, all 3 tests with different seeds have same assert failure.

br3aker:
This is very strange, I've done nothing to the actual convertion code yet it have passed net472 tests.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ColorConverters/JpegColorConverterVector.cs
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            public sealed override bool IsAvailable => Vector.IsHardwareAccelerated && Vector<float>.Count % 4 == 0;
+
+            public override void ConvertToRgbInplace(in ComponentValues values)
+            {
+                DebugGuard.IsTrue(this.IsAvailable, $"{this.GetType().Name} converter is not supported on current hardware.");
+
+                int length = values.Component0.Length;
+                int remainder = length % Vector<float>.Count;
```
gfoidl, line 36:
```suggestion
                int remainder = (int)((uint)length % (uint)Vector<float>.Count);
```
results in [`& 7`](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgDkBXfGKASzFwG4BYAKD+IBmSgCZyAYXIBvPuTmVhrAHYZyAWSoAKZaoA2MJQHMMACwCU5ALwA+cvqOnyAUnIA1GGAzQAPADNdENgY1jTiEAwqPLzyCuQ66iLaKnYGxuZWtkkYZpqaDDpm9mnO5HkF7p4+/oHBoeEqZlEAvkA=) instead of the division.

The second cast is strange, as `Vector<float>.Count` should be treated as constant by the JIT.

gfoidl: Filed https://github.com/dotnet/runtime/issues/62326 to address this in the JIT too.

br3aker:
@JimBobSquarePants okay, I'm almost 100% sure there's some sort of a bug here. I can reproduce failing tests locally only on `core 2.1`. `net472`, `core 3.1` and `net5` are all passing. Furthermore, only remote executor tests are failing, `AllowAll` executes in the current process without remote execution and it passes for `Vector<T>` implementations, `EnableAVX 0` for `Vector<float>.Count == 4` aka auto SSE fails with some absolutely random values, it just can't happen - it's the same code with `Vector<float>`. Especially considering other passing runtimes.

What's our next step on this?

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
```diff
@@ -433,29 +436,29 @@ static JpegColorConverter.ComponentValues Copy(JpegColorConverter.ComponentValue
...
+            bool equal = ColorSpaceComparer.Equals(expected, actual);
+            Assert.True(equal, $"Colors {expected} and {actual} are not equal at index {i}");
         }
 
-        private static void ValidateCyyK(in JpegColorConverter.ComponentValues values, in JpegColorConverter.ComponentValues result, int i)
+        private static void ValidateCyyK(in JpegColorConverterBase.ComponentValues values, in JpegColorConverterBase.ComponentValues result, int i)
         {
-            var v = new Vector4(0, 0, 0, 1F);
-            var scale = new Vector4(1 / 255F, 1 / 255F, 1 / 255F, 1F);
+            Vector4 v = Vector4.Zero;
```
brianpopow, line 461: @br3aker could you try replacing this Vector4.Zero with the explicit `Vector4(0, 0, 0, 1F)`? I had a similar strange issue in the past when using Vector default values with `net472` and `netcoreapp2.0`. See: https://github.com/dotnet/runtime/issues/55623

br3aker: I remember that issue but unfortunately all tests with Vector code fail - even YCbCr which does not use Vector4 at all for comparison.

brianpopow: I bet its some similar strange issue.

br3aker:
Yet another detail: remote fails on 4th index - start of the second SSE vector.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
```diff
@@ -467,58 +470,60 @@ private static void ValidateCyyK(in JpegColorConverter.ComponentValues values, i
...
+            bool equal = ColorSpaceComparer.Equals(expected, actual);
+            Assert.True(equal, $"Colors {expected} and {actual} are not equal at index {i}");
         }
 
-        private static void ValidateCmyk(in JpegColorConverter.ComponentValues values, in JpegColorConverter.ComponentValues result, int i)
+        private static void ValidateCmyk(in JpegColorConverterBase.ComponentValues values, in JpegColorConverterBase.ComponentValues result, int i)
         {
-            var v = new Vector4(0, 0, 0, 1F);
-            var scale = new Vector4(1 / 255F, 1 / 255F, 1 / 255F, 1F);
+            Vector4 v = Vector4.Zero;
```
brianpopow, line 511: This `v` vector does not seem to be used and can be removed?

br3aker: I will fix all warning right after that nasty failing test, thanks!

br3aker:
@brianpopow I'm either dumb or don't understand how remote executor works. Code below **fails** on `core2.1` and `net472`, works on every other runtime:
image


```
[Fact]
public void FoolproofFact()
{
        RunTest,
        HwIntrinsics.DisableAVX);

    static void RunTest()
    {
        Assert.Equal(4, Vector<float>.Count);
    }
}
```

This test **does NOT fail** on any of the runtimes:
```
[Fact]
public void FoolproofFact()
{
        RunTest,
        HwIntrinsics.DisableSIMD);

    static void RunTest()
    {
        Assert.Equal(4, Vector<float>.Count);
    }
}
```

Looks like 4 is a minimum hardware vector size. Wondering if JIT could compile Vector<float> into SSE instructions with `Vector<float>.Count == 4` - that would explain index 4 error.

br3aker:
Another detail related to this problem, some code in repo uses `Vector<T>.Count` API in conjunction with the x86 intrinsics - they both are under `SUPPORTS_RUNTIME_INTRINSICS` conditional. Example: [png filters](https://github.com/SixLabors/ImageSharp/tree/master/src/ImageSharp/Formats/Png/Filters). That conditional is set for core 3.0 and up - platforms which do not suffer from this strange error I've encountered. That's why tests with existing code pass - they do not run remote executor for `Vector<T>.Count` on net472 and core20.

Marking this PR WIP until we decide how we should approach this.

brianpopow:
@br3aker I think the reason for the failing tests is, that they are not inside a `#if SUPPORTS_RUNTIME_INTRINSICS` block, e.g. `FromYccKVector`. The test runs with and without AVX.

br3aker:
@brianpopow `Vector<T>` implementation for 8 floats wasn't excluded from the net472/core21 via `#if` before this PR so it will be a regression. Which is actually strange, a lot of simd utils code with `Vector<T>` is covered by `#if`. It clearly works without remote executor at least on my machine and it did work on CI.

Honestly, if it was my call - I'd say it's good to go without tests with disabled avx for SSE check. If it works on other runtimes without a problem and work with avx hardware - sse should not be a problem, at least from out side.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
```diff
@@ -18,20 +18,17 @@ namespace SixLabors.ImageSharp.Tests.Formats.Jpg
     [Trait("Format", "Jpg")]
     public class JpegColorConverterTests
     {
+        private static readonly float MaxColorChannelValue = 255f;
+
         private const float Precision = 0.1F / 255;
 
-        private static readonly ApproximateColorSpaceComparer ColorSpaceComparer = new ApproximateColorSpaceComparer(Precision);
+        private const int TestBufferLength = 40;
```
antonfirsov, line 25: We need to test remainder paths, which doesn't happen with `TestBufferLength=40`.

antonfirsov (commented):
`FeatureTestRunner` uses `COMPlus_*` [environment variables](https://github.com/steveharter/dotnet_coreclr/blob/master/Documentation/project-docs/clr-configuration-knobs.md) to turn off individual instruction sets or SIMD entirely. The individual flags (eg. `COMPlus_EnableAVX`) are only recognized by .NET Core 3.0+ runtimes, where intrinsics have been introduced.

I'm not super sure if the `FeatureTestRunner` transition is worth the trouble in case of `JpegColorConverterTests`. Upon the previous problem, it turns very fast unit tests into slow out-of-process test runs, pressing us to reduce the number of test cases to save execution time.

If we somehow manage to resolve these challenge with `Vector<float>.Count` without loosing coverage I'm fine with it though.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
br3aker: Jpeg files always have strides of pixels divisible by 8, there will neven be a remainder problem. The only way remainder can exist - AVX-512 which has 16 floats per vector. `40 % 16 == 8`, size of 40 pixels covers possible remainder path.

br3aker:
> If we somehow manage to resolve the challenge of Vector<float>.Count without loosing coverage I'm fine with it though.

It doesn't affect coverage afaik. The only problem atm is that automatic SSE implementation by jit from `Vector<float>` code won't be tested if CI hardware supports avx. I actually not even sure that there's any CPUs right now which don't support at least minimal AVX set.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
```diff
@@ -40,323 +37,273 @@ public JpegColorConverterTests(ITestOutputHelper output)
...
-            ValidateConversion(
-                new JpegColorConverter.FromRgbAvx2(8),
-                3,
-                inputBufferLength,
-                resultBufferLength,
-                seed);
+            FeatureTestRunner.RunWithHwIntrinsicsFeature(
+                RunTest,
+                seed,
+                HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX);
```
antonfirsov, line 221: > It doesn't affect coverage afaik.

Makes sense, 3.1+ builds should cover the SSE cases, so it's fine if 2.1 and Framework are being executed against AVX only. Maybe you can do something like:

```C#
            HwIntrinsics intrinsicsConfig =  HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX;
#else
            HwIntrinsics intrinsicsConfig = HwIntrinsics.AllowAll;
#endif

                RunTest,
                seed,
                intrinsicsConfig);

```

JimBobSquarePants (approved):
Looks great. Nice and easy to read 👍

JimBobSquarePants:
@brianpopow weird... That test I thought we just disabled failed here. https://github.com/SixLabors/ImageSharp/runs/4497782950?check_suite_focus=true#step:13:277

brianpopow:
> @brianpopow weird... That test I thought we just disabled failed here. https://github.com/SixLabors/ImageSharp/runs/4497782950?check_suite_focus=true#step:13:277

I dont understand how this test gets executed `AllocateSingleMemoryOwner_Finalization_ReturnsToPool` on macOS.

Also it seems there are other tests failing, not just this one: `AllocateMemoryGroup_Finalization_ReturnsToPool`

brianpopow:
Should i add a `ActiveIssue` attribute to `AllocateMemoryGroup_Finalization_ReturnsToPool`?

JimBobSquarePants:
I think so for now. We're gonna end up with a test run backlog otherwise.

brianpopow:
The tests are still executed on MacOS (and failing)

JimBobSquarePants:
Maybe I read the docs wrong.... Let's use our own test utilities inside each method to skip. I won't be able to look at the tests on a Mac until tomorrow now.

PR #1866, opened 2021-12-03
Do not remove EXIF orientation in RotateProcessor

ronaldbarendse:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description
I noticed the RotateProcessor also removed the EXIF orientation, which should be an explicit operation (as mentioned here https://github.com/SixLabors/ImageSharp/issues/400#issuecomment-944202410).

PR #1868, opened 2021-12-03
Add AutoOrientMode to configure handling of EXIF orientation flag

ronaldbarendse:
### Prerequisites

### Description
This PR makes the `OrientationMode` enumeration public and introduces a `GetOrientation()` extension method on the image metadata, so you can easily get this value using e.g. `image.Metadata.GetOrientation()`.

The `AutoOrientProcessor` uses this extension method (instead of its own private `GetExifOrientation()` that also resets the value) and now also allows you to configure how to deal with the EXIF orientation flag after applying the rotating/flipping. The default is to reset the flag to 0/horizontal (normal)/top-left to keep the existing behavior.

Because this processor uses the `RotateProcessor`, PR https://github.com/SixLabors/ImageSharp/pull/1866 needs to be merged to be able to use `AutoOrientMode.KeepExifValue` on images that are rotated (otherwise the flag would still be removed).

_I've created this as draft, as it still needs test coverage to validate the new `AutoOrientMode` and also requires the above PR to be merged first._

### DIFF
--- a/src/ImageSharp/Processing/AutoOrientMode.cs
+++ b/src/ImageSharp/Processing/AutoOrientMode.cs
@@ -1,0 +1,26 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Processing
+{
+    /// <summary>
+    /// Provides an enumeration over how an image should be auto oriented.
+    /// </summary>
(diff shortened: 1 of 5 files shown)


JimBobSquarePants:
We have a feature request process. 

https://github.com/SixLabors/ImageSharp/issues/new/choose

PR #1869, opened 2021-12-03
Do not keep a reference to the FlipProcessor instance

ronaldbarendse:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description
`FlipProcessor<TPixel>` kept a reference to the `FlipProcessor`, which is inconsistent with all the other processors, so this is a simple change to only store the `FlipMode`.

PR #1870, opened 2021-12-03
Do not keep a reference to the EntropyCropProcessor instance

ronaldbarendse:
### Prerequisites

### Description
Similar to https://github.com/SixLabors/ImageSharp/pull/1869, this only stores the threshold value of the `EntropyCropProcessor`.


ronaldbarendse:
Searching on `this.definition` shows a lot more processors using the same approach, so I'm not sure that would be the 'correct' way and if it makes sense to refactor these: https://github.com/SixLabors/ImageSharp/search?q=this.definition+filename%3A*%7BTPixel%7D.cs

JimBobSquarePants:
@ronaldbarendse can you please open discussions before PRs. This change has a negligible effect on the library and zero affect on the output but now takes time for us to review. 

I appreciate you feel like you are being helpful but we are a very small team with limited resources.

PR #1873, opened 2021-12-04
Allocate clean buffer in AnalyzeEntropy

### LINKED ISSUE #1872: Lossless Webp encoded files does not match reference implementation
### Discussed in https://github.com/SixLabors/ImageSharp/discussions/1863

<sup>Originally posted by **jubilant-enigma** November 30, 2021</sup>
~~In my experience, the new WebP encoder (at least in lossless mode) doesn't seem to match the file sizes possible with the official cwebp encoder. I've seen scenarios where the ImageSharp WebP encoder is as much as 50% larger than the cwebp encoder.~~

~~Is it planned/in progress for the ImageSharp WebP encoder to be able to match cwebp?~~

See https://github.com/SixLabors/ImageSharp/discussions/1863#discussioncomment-1735073

I'm currently working with ImageSharp 2.0.0-alpha.0.119 from MyGet.</div>

The image linked in the discussion will produce a 10MB file, where the reference libwebp will produce ~6MB.

Settings:
```
var webpEncoder = new WebpEncoder()
{
    FileFormat = WebpFileFormatType.Lossless,
    Method = WebpEncodingMethod.BestQuality,
};
```

brianpopow:
### Prerequisites

### Description

This PR fixes a bug in `AnalyzeEntropy` which did not allocate a clean buffer from the memory allocator, resulting in possible unpredictable results.

Closes #1872

### DIFF
--- a/tests/ImageSharp.Tests/Formats/WebP/WebpEncoderTests.cs
+++ b/tests/ImageSharp.Tests/Formats/WebP/WebpEncoderTests.cs
@@ -102,6 +102,24 @@ namespace SixLabors.ImageSharp.Tests.Formats.Webp
         }
 
         [Theory]
+        [WithFile(Lossy.NoFilter06, PixelTypes.Rgba32, 15114)]
+        public void Encode_Lossless_WithBestQuality_HasExpectedSize<TPixel>(TestImageProvider<TPixel> provider, int expectedBytes)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            var encoder = new WebpEncoder()
+            {
+                FileFormat = WebpFileFormatType.Lossless,
+                Method = WebpEncodingMethod.BestQuality
+            };
+
+            using Image<TPixel> image = provider.GetImage();
+            using var memoryStream = new MemoryStream();
+            image.Save(memoryStream, encoder);
+
+            Assert.Equal(memoryStream.Length, expectedBytes);
+        }
+
+        [Theory]
         [WithFile(RgbTestPattern100x100, PixelTypes.Rgba32, 85)]
         [WithFile(RgbTestPattern100x100, PixelTypes.Rgba32, 60)]
         [WithFile(RgbTestPattern80x80, PixelTypes.Rgba32, 40)]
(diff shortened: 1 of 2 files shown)


antonfirsov (approved):
Any chance we can somehow regression test this? (Eg. check resulting stream size in some of the encoder tests?)

brianpopow:
> Any chance we can somehow regression test this? (Eg. check resulting stream size in some of the encoder tests?)

I have added a test to check for the expected file size, but I am not sure if that will trigger the actual issue. The issue #1872 will only happen, if the allocator will allocate unclean buffers, which i believe will not be the case when i create a new instance of `WebpEncoder`

tests/ImageSharp.Tests/Formats/WebP/WebpEncoderTests.cs
```diff
@@ -101,6 +101,24 @@ public void Encode_Lossless_WithDifferentMethodAndQuality_Works<TPixel>(TestImag
...
+            {
+                FileFormat = WebpFileFormatType.Lossless,
+                Method = WebpEncodingMethod.BestQuality
+            };
+
+            using Image<TPixel> image = provider.GetImage();
+            using var memoryStream = new MemoryStream();
+            image.Save(memoryStream, encoder);
+
+            Assert.Equal(memoryStream.Length, expectedBytes);
```
antonfirsov, line 119: Do we need strict equality? Doesn't the spec allow some space for the implementation?

brianpopow: No the spec is not that strict. The analyses step could have choosen different transformation, e.g. here the analyses step has choosen to use the following features PREDICTION CROSS-COLOR-TRANSFORM SUBTRACT-GREEN. If one of those transformations would not been applied, the image would be still valid, but a bit larger.

Nonetheless i think we should try to aim to be as close to the reference implementation as possible and this is what libwebp would have encoded to.

antonfirsov: Ok then, we can relax this restriction later if it will become necessary for some reason in the future.

antonfirsov (approved):

PR #1877, opened 2021-12-05
Added support for loading exif data from PNG "Raw profile type exif" text chunk

jubilant-enigma:
I don't really have much experience in C# outside of toy projects, so I don't really have a good idea of what best practices are for C# these days. 

Some more optimal functions (span related, etc) seemed available with some of the newer .net versions, however, then .netstandard1.3 would produce build errors. Yet that tests don't seem to be run on that target? Not sure what's going on there.

### Prerequisites

### Description
See https://github.com/SixLabors/ImageSharp/discussions/1875.


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -112,6 +112,11 @@ internal sealed class PngDecoderCore : IImageDecoderInternals
         /// </summary>
         private PngChunk? nextChunk;
 
+        /// <summary>
+        /// "Exif" and two zero bytes. Used for the legacy exif parsing.
+        /// </summary>
+        private static readonly byte[] ExifHeader = new byte[] { 0x45, 0x78, 0x69, 0x66, 0x00, 0x00 };
```
gfoidl, line 118:
```suggestion
        // This uses C# compiler's optimization to refer to the static data directly, no intermediate array allocations happen.
        private static ReadOnlySpan<byte> ExifHeader => new byte[] { 0x45, 0x78, 0x69, 0x66, 0x00, 0x00 };
```
For background info on that see e.g. https://vcsjones.dev/csharp-readonly-span-bytes-static/

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -971,6 +977,94 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
...
+            if (!dataSpan.Slice(0, 4).ToString().Equals("exif", StringComparison.OrdinalIgnoreCase))
+            {
+                // "exif" identifier is missing from the beginning of the text chunk
+                return;
+            }
+
+            // Skip to the data length
+            dataSpan = dataSpan.Slice(4).TrimStart();
+            int dataLengthEnd = dataSpan.IndexOf('\n');
+            int dataLength = int.Parse(dataSpan.Slice(0, dataSpan.IndexOf('\n')).ToString());
```
gfoidl, line 1008:
```suggestion
            int dataLength = int.Parse(dataSpan.Slice(0, dataSpan.IndexOf('\n')));
```
`int.Parse` has an overload for ROS<char>, so the intermediate string-allocation can be avoided.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -971,6 +977,94 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
...
+            dataSpan = dataSpan.Slice(dataLengthEnd).Trim();
+            string dataSpanString = dataSpan.ToString().Replace("\n", string.Empty);
+            if (dataSpanString.Length != (dataLength * 2))
+            {
+                // Invalid length
+                return;
+            }
+
+            // Parse the hex-encoded data into the byte array we are going to hand off to ExifProfile
+            byte[] dataBlob = new byte[dataLength - ExifHeader.Length];
```
gfoidl, line 1020: Is this array allocation OK?
(just perf-wise)

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -971,6 +977,94 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
...
+            for (int i = 0; i < dataLength; i++)
+            {
+                byte parsed = Convert.ToByte(dataSpanString.Substring(i * 2, 2), 16);
```
gfoidl, line 1023: Note: I'm not familiar with the algorithm, so maybe this comment may not be applicable.

Couldn't there be used `Encoding.UTF8.GetBytes` (ideally with span-overloads)?
The built-in method is highly optimized, and I doubt that the char-for-char loop will be faster -- code readability should be better with the .NET provided method too.

Or is this done manually with this loop because of the check in L1026?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -971,6 +977,94 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
...
+                if (i < ExifHeader.Length)
```
gfoidl, line 1024:
```suggestion
                if ((uint)i < (uint)ExifHeader.Length)
```
With the cast to `uint` the JIT can better prove that the values aren't negative, and in the correct relation (i < length), so it can elide the bound checks for `ExifHeader[i]`'s access.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
antonfirsov: It's [missing](https://apisof.net/catalog/71bfcfac-9ac2-5416-bb5c-b5c645090d9d) in some of our target frameworks. Unfortunately this looks like a candidate for code like this:

```C#
internal static int ParseInt32(ReadOnlySpan<char> span)
{
#if NETSTANDARD2_1 || NETCOREAPP2_1_OR_GREATER
    return int.Parse(span, CultureInfo.InvariantCulture);
#else
    return int.Parse(span.ToString(), CultureInfo.InvariantCulture);
#endif
}
```

src/ImageSharp/Formats/Png/PngDecoderCore.cs
antonfirsov:
```suggestion
            using IMemoryOwner<byte> dataBuffer = this.memoryAllocator.Allocate<byte>(dataLength);
            Span<byte> dataBlob = dataBuffer.GetSpan();
```

After #1730, I'm planning to create a utility similar to SixLabors/Font's [Buffer<T>](https://github.com/SixLabors/Fonts/blob/master/src/SixLabors.Fonts/Buffer%7BT%7D.cs) that avoids `IMemoryOwner` allocation when it's backed by `ArrayPool<byte>.Shared`, but for now this should be good enough.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
antonfirsov: This code looks like it's converting a hexadecimal substring eg `0F`, not an ASCII/utf8 value. The `*.Substring() -> string` allocation is very unfortunate inside a loop though. I don't see span overloads in `Convert.*` methods, wonder if there is a corresponding Span API. If not, implementing our own helper method should not be hard.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: @antonfirsov Is there some "Utility" static class etc where you would like that added? Or just stick it in `PngDecoderCore` for now?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: Correct, it is converting hexadecimal string values to a byte array. The only sort of applicable Convert method here would be https://docs.microsoft.com/en-us/dotnet/api/system.convert.fromhexstring?view=net-6.0, but that is only available with .NET 5+, and it would still require allocating many byte arrays (since we would only be able to parse one line at a time). I'll add a comment along these lines to the code, but the "raw profile type exif" chunk content looks like this:

```

exif
     194
45786966000049492a000800000003000e01020044000000320000001192040001000000
2a000000139202004600000076000000000000004120636f6c6f7266756c2074696c696e
67206f6620626c75652c207265642c2079656c6c6f772c20616e6420677265656e203478
3420706978656c20626c6f636b732e004475706c6963617465642066726f6d206261736e
337030322e706e672c207468656e20696d616765206d65746164617461206d6f64696669
6564207769746820657869763200

```

Like @antonfirsov mentioned, it's probably best to just do this via a helper method that doesn't allocate.

jubilant-enigma:
Additional design decision: should the `Raw profile type exif` text chunk be removed from the PngMetadata after successful parsing?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: @antonfirsov this suggestion would be ideal, however currently `ExifProfile` does not have a constructor that can accept `Memory`/`Span` objects, only `byte[]`.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
antonfirsov: Yes, I don't see other use cases for it ATM.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
antonfirsov: Ah I missed that it goes to `ExifProfile`, than it should be a `new byte[]` indeed.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: Very well. A similar function, `StringEquals` for comparing two `ReadOnlySpan<char>`s, was also added to `PngDecoderCore`.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: I have implemented `HexStringToBytes` that does not allocate. Mine is (IMO) a more readable and safer version of https://stackoverflow.com/a/9995303/871842 that also checks the input data is valid. I'm not sure if @antonfirsov or @gfoidl have something to say about exceptions being used in a hot loop like this. I benchmarked it against the exact code given in the stackoverflow answer above (which has no data validation or exceptions), and there was no difference in performance with 8MB of hex data.

jubilant-enigma: Forgot to post these benchmark results in case anyone is curious:

```
|                 Method |     Mean |    Error |   StdDev |
|----------------------- |---------:|---------:|---------:|
| StackOverflowConverter | 75.38 ms | 0.745 ms | 0.582 ms |
|            MyConverter | 71.99 ms | 1.373 ms | 1.737 ms |
|        NativeConverter | 14.39 ms | 0.283 ms | 0.474 ms |
```

NativeConverter is `Convert.FromHexBytes()` available in .NET 5 and newer.

JimBobSquarePants: You mean [`Convert.FromHexString`](https://source.dot.net/#System.Private.CoreLib/Convert.cs,c9e4fbeaca708991) no?

JimBobSquarePants:
>Additional design decision: should the Raw profile type exif text chunk be removed from the PngMetadata after successful parsing?

Once we have that profile, absolutely.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -971,6 +978,200 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
...
+                metadata.ExifProfile = newProfile;
+            }
+            else
+            {
+                // Try to merge existing keys with the ones from the new profile
+                foreach (IExifValue newKey in newProfile.Values)
+                {
+                    if (replaceExistingKeys || metadata.ExifProfile.GetValueInternal(newKey.Tag) is null)
+                    {
+                        metadata.ExifProfile.SetValueInternal(newKey.Tag, newKey.GetValue());
```
JimBobSquarePants, line 1171: Why are we using `Get/Set` internal methods here?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -971,6 +978,200 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
...
+            }
+            catch
+            {
+                return;
+            }
+
+            this.MergeOrSetExifProfile(metadata, new ExifProfile(exifBlob), replaceExistingKeys: false);
+        }
+
+        private static bool StringEquals(ReadOnlySpan<char> span1, ReadOnlySpan<char> span2, StringComparison comparisonType)
```
JimBobSquarePants, line 1068: There's not really much point passing the comparer here since we use this method once with a single comparer type. 

It's possible to [reimplement](https://source.dot.net/#System.Private.CoreLib/MemoryExtensions.Globalization.cs,72285d8c26ebb991) the comparison for legacy platforms but for such an obscure feature probably not worth the effort.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: You're right, my mistake. I forgot the reference source was a thing. What's the license for that, and should that implementation just be re-implemented here? Basically it's just using a lookup table instead.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: It seemed to be the intended way to handle things here, but I could be wrong. The non-internal methods are generic, and we don't know the types of the values. The `Values` property is an `IReadOnlyList`.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: Done, the unnecessary parameters have been removed from `StringEquals` (now `StringEqualsInsensitive`) and `ParseInt32`.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: Lemme have a look...

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: MIT so you can reimplement here 😁. I would create a static `HexEncoder` type so we can easily test and benchmark it.

jubilant-enigma: @JimBobSquarePants Finally had some more time to look at this. Where should the `HexEncoder` class be created?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: Hey there @JimBobSquarePants, did you find a different function I should be using instead?

JimBobSquarePants: I’ll have a look later tonight. Been taking some much needed time off

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: Stick it under Common/Helpers as an internal class.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: Stick with the internal methods for now. That looks like the best way. We should look to add a `Merge` method to `ExifProfile` for the future.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1030,7 +1031,251 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
 
             if (this.TryUncompressTextData(compressedData, PngConstants.Encoding, out string uncompressed))
             {
-                metadata.TextData.Add(new PngTextData(name, uncompressed, string.Empty, string.Empty));
+                if (name.Equals("Raw profile type exif", StringComparison.OrdinalIgnoreCase) &&
+                    this.TryReadLegacyExifTextChunk(baseMetadata, uncompressed))
+                {
+                    // Successfully parsed exif data stored as text in this chunk
```
JimBobSquarePants, line 1037: Reverse the condition here so you don't have the odd empty scope.

jubilant-enigma: I ended up moving this into its own function, `TryReadTextChunkMetadata`, since I was missing support for loading legacy exif data from uncompressed text chunks, and I also wanted to have an easy spot to add support for other metadata chunks in the future. I suppose I could also just add IPTC support in this PR too if you want, but I figure it's getting somewhat large and it might be better to just make another PR with that feature after this one gets merged.

JimBobSquarePants (commented):
@jubilant-enigma Can you double check that your local fork is using the updated submodules. There shouldn't be a change to them compared to our master.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: Yeah, let's keep IPTC  separate for now.

JimBobSquarePants (approved):
LGTM. Great job 👍

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1028,12 +1034,185 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
...
+                // Check for the presence of the exif header in the hex-encoded binary data
+                byte[] tempExifBuf = exifBlob;
+                if (exifBlob.Length < exifHeader.Length)
+                {
+                    // Need to allocate a temporary array, this should be an extremely uncommon (TODO: impossible?) case
+                    tempExifBuf = new byte[exifHeader.Length];
+                }
+
+                HexConverter.HexStringToBytes(dataSpan.Slice(0, exifHeader.Length * 2), tempExifBuf);
+                if (!tempExifBuf.AsSpan().Slice(0, exifHeader.Length).SequenceEqual(exifHeader))
```
turbedi, line 1118: `if (!tempExifBuf.AsSpan(0, exifHeader.Length).SequenceEqual(exifHeader))`

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1028,12 +1034,185 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
...
+                {
+                    ReadOnlySpan<char> lineSpan = dataSpan;
+
+                    int newlineIndex = dataSpan.IndexOf('\n');
+                    if (newlineIndex != -1)
+                    {
+                        lineSpan = dataSpan.Slice(0, newlineIndex);
+                    }
+
+                    i += HexConverter.HexStringToBytes(lineSpan, exifBlob.AsSpan().Slice(i));
```
turbedi, line 1139: `i += HexConverter.HexStringToBytes(lineSpan, exifBlob.AsSpan(i));`

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1028,12 +1034,185 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
...
+            if (!StringEqualsInsensitive(dataSpan.Slice(0, 4), "exif".AsSpan()))
+            {
+                // "exif" identifier is missing from the beginning of the text chunk
+                return false;
+            }
+
+            // Skip to the data length
+            dataSpan = dataSpan.Slice(4).TrimStart();
+            int dataLengthEnd = dataSpan.IndexOf('\n');
+            int dataLength = ParseInt32(dataSpan.Slice(0, dataSpan.IndexOf('\n')));
```
turbedi, line 1089: `int dataLength = ParseInt32(dataSpan.Slice(0, dataLengthEnd));`

jubilant-enigma:
> @jubilant-enigma Can you double check that your local fork is using the updated submodules. There shouldn't be a change to them compared to our master.
> 
> EDIT. I pulled your fork and did the update myself to fix.

Sorry about that, thanks for taking care of it! Unfortunately my available free time has been more intermittent than I hoped.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
jubilant-enigma: Thanks for pointing these issues out @turbedi. Ideally I'll fix them sometime next week when I create a PR for IPTC metadata loading from text chunks.

Issue #1345, opened 2020-09-07
Add Abgr32 pixel type


antonfirsov:
Follow up on https://github.com/SixLabors/ImageSharp/discussions/1344#discussioncomment-63349

We have: `Rgba32`, `Argb32`, `Bgra32`
`Abgr32` looks like a missing piece. Can be optimized using [existing patterns](https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/PixelFormats/PixelImplementations/Generated/_Common.ttinclude#L20).

ynse01:
Now that #1886 is merged into master, we can close this issue, right?

JimBobSquarePants:
Yep! Thanks!

Issue #1884, opened 2021-12-09
Substantial Jpeg performance degradation on master branch as of 09.12.2021


br3aker:
### Description
Looks like it was caused by Memory Allocator PR but I'm not sure, just spotted it @antonfirsov @JimBobSquarePants 

Master before allocator PR
|                              Method - Decode |      Mean |     Error |    StdDev |
|------------------------------------ |----------:|----------:|----------:|
|        'Baseline 4:4:4 Interleaved' | 11.504 ms | 0.0719 ms | 0.0673 ms |
|        'Baseline 4:2:0 Interleaved' |  8.605 ms | 0.0372 ms | 0.0348 ms |
|        'Baseline 4:0:0 (grayscale)' |  1.575 ms | 0.0054 ms | 0.0042 ms |
| 'Progressive 4:2:0 Non-Interleaved' | 13.207 ms | 0.0472 ms | 0.0369 ms |

Current master
|                              Method - Decode |      Mean |     Error |    StdDev |
|------------------------------------ |----------:|----------:|----------:|
|        'Baseline 4:4:4 Interleaved' | 13.532 ms | 0.1669 ms | 0.1561 ms |
|        'Baseline 4:2:0 Interleaved' |  9.081 ms | 0.0397 ms | 0.0332 ms |
|        'Baseline 4:0:0 (grayscale)' |  1.850 ms | 0.0109 ms | 0.0085 ms |
| 'Progressive 4:2:0 Non-Interleaved' | 13.581 ms | 0.1397 ms | 0.1238 ms |

What is strange is that progressive code wasn't really affected as baseline decoder.

---

Master before allocator PR
|                      Method - Encode | Quality |     Mean |    Error |   StdDev | Ratio |
|---------------------------- |-------- |---------:|---------:|---------:|------:|
| 'System.Drawing Jpeg 4:2:0' |      75 | 30.04 ms | 0.540 ms | 0.479 ms |  1.00 |
|     'ImageSharp Jpeg 4:2:0' |      75 | 19.32 ms | 0.290 ms | 0.257 ms |  0.64 |
|     'ImageSharp Jpeg 4:4:4' |      75 | 26.76 ms | 0.332 ms | 0.294 ms |  0.89 |
|                             |         |          |          |          |       |
| 'System.Drawing Jpeg 4:2:0' |      90 | 32.82 ms | 0.184 ms | 0.163 ms |  1.00 |
[...]

PR #1886, opened 2021-12-10
Add Abgr32 pixel type

ynse01:
### Prerequisites

### Description
This adds the Abgr32 Pixel Format, as described in issue #1345.

For most of its implementation, I took inspiration from the Bgra32 pixel format code. It includes the SIMD conversions to other RGB based formats, so conversion performance should be pretty good.

I also fixed, what I think is a bug, in IComponentShuffle. Please check if my understanding of byte shuffling is correct.

I've not added Benchmarks, as there weren't any for Bgra32 also. I expect the performance will be similar to the other RGBA Pixel Formats. Let me know if you want me to add benchmarks.


brianpopow:
@antonfirsov there seem to be failing tests on macos, which may be related to #1730, could you take a look at this: [Failing Tests](https://github.com/SixLabors/ImageSharp/runs/4485782267?check_suite_focus=true)

src/ImageSharp/PixelFormats/PixelImplementations/Bgr24.cs
```diff
@@ -185,6 +185,15 @@ public void FromRgb24(Rgb24 source)
             this.B = source.B;
         }
 
+        /// <inheritdoc/>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public void FromAbgr32(Abgr32 source)
+        {
+            this.R = source.R;
+            this.G = source.G;
+            this.B = source.B;
```
gfoidl, line 194: Not as readable, but [less machinecode](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgBEBLbAcwDsJcGDmFwsAwhAAmMAIK9sAGwCeuDrgDcNek1YA5AK74YUYRq2NmLAEr7eQo+Ij4ADhwXGAysYBuwmGdoLVhs7DgcASTtjCGcvKF8wf01qGgBtDwwofTAMABlsJQh9DAAKfMLigGkOXkkWLwBHfRhQxQBKAF0tAGYGQSychhlgbihu0hoAbxoGWaZe4CUMGCHkufmGReWGACE1ueIFpZWAcX3Zw83jhitkgF8aNIyBvIKi0vL36tr6mCaWoTtLq0Xr9bIYXajUgoKYzA5HbZ7OEXBGnc4bLYrW7Ihg4y7EFAMABiUCcw1G4xK5LGpD6RSgiTaOOm1HW6wwAAs1NYGABeOn6Bkwazo9lc0QnPkCoUsM44sXcnZS3D0xIsJGsuYPTWzPG9AnE0n4anjAD6MipIxp0sZzPlbNmsAAZldtirBYkrDAXfznQwAKq8XDYJ3CmS4AA8JtIaFdMAAfCU/e6hW1RQ6GH7MTaYF6nSSnErfd6A0GQ2HJJIdscAPJOp24GClZOq3Pe2PkNP2h2ctRSwPB0MscMRzGxnZQlCJlsetv5o07Ls6hjau5AA)
```suggestion
            ref byte sourceRef = ref Unsafe.As<Abgr32, byte>(ref source);
            ref byte sourceRefFromB = ref Unsafe.AddByteOffset(ref sourceRef, 1);
            this = Unsafe.As<byte, Bgr24>(ref sourceRefFromB);
```
I'm not sure if I would take this suggestion (readability of code wise).

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Generated/Abgr32.PixelOperations.Generated.cs
```diff
@@ -0,0 +1,346 @@
...
+{
+    /// <content>
+    /// Provides optimized overrides for bulk operations.
+    /// </content>
+    public partial struct Abgr32
+    {
+        /// <summary>
+        /// Provides optimized overrides for bulk operations.
+        /// </summary>
+        internal partial class PixelOperations : PixelOperations<Abgr32>
```
gfoidl, line 22: Can this be `sealed`?
Sealing would open the door in the JIT to allow de-virtualization, especially with dynamic profile guided optimizaiton (D-PGO).

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Generated/Abgr32.PixelOperations.Generated.cs
```diff
@@ -0,0 +1,346 @@
...
+                ReadOnlySpan<Abgr32> sourcePixels,
+                Span<L8> destinationPixels)
+            {
+                Guard.NotNull(configuration, nameof(configuration));
+                Guard.DestinationShouldNotBeTooShort(sourcePixels, destinationPixels, nameof(destinationPixels));
+
+                ref Abgr32 sourceRef = ref MemoryMarshal.GetReference(sourcePixels);
+                ref L8 destRef = ref MemoryMarshal.GetReference(destinationPixels);
+
+                for (int i = 0; i < sourcePixels.Length; i++)
```
gfoidl, line 207:
```suggestion
                for (nint i = 0; i < sourcePixels.Length; i++)
```
The same on other places.

I comment here deputy for other generated code too.

src/ImageSharp/PixelFormats/PixelImplementations/Abgr32.cs
```diff
@@ -0,0 +1,394 @@
...
+        public void FromBgra5551(Bgra5551 source) => this.FromScaledVector4(source.ToScaledVector4());
+
+        /// <inheritdoc/>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public void FromBgr24(Bgr24 source)
+        {
+            this.R = source.R;
+            this.G = source.G;
+            this.B = source.B;
+            this.A = byte.MaxValue;
```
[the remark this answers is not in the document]
ynse01: I will take this suggestion, but with a small code comments to explain the _magic_ happening here.

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
```diff
@@ -194,4 +194,37 @@ public void RunFallbackShuffle(ReadOnlySpan<byte> source, Span<byte> dest)
...
+        }
+
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public void RunFallbackShuffle(ReadOnlySpan<byte> source, Span<byte> dest)
+        {
+            ref uint sBase = ref Unsafe.As<byte, uint>(ref MemoryMarshal.GetReference(source));
+            ref uint dBase = ref Unsafe.As<byte, uint>(ref MemoryMarshal.GetReference(dest));
+            int n = source.Length / 4;
+
+            for (int i = 0; i < n; i++)
```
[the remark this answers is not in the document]
ynse01: Indeed, I see. Applied it to all methods in this file

src/ImageSharp/PixelFormats/PixelImplementations/Bgr24.cs
ynse01: I do take this suggestion, but adding a small code comment to explain the _magic_ happening here.

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Generated/Abgr32.PixelOperations.Generated.cs
ynse01: Updated the common.tt template and regenerated all the PixelOperations files that used it.

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
```diff
@@ -194,4 +194,37 @@ public void RunFallbackShuffle(ReadOnlySpan<byte> source, Span<byte> dest)
...
+                uint packed = Unsafe.Add(ref sBase, i);
+
+                // packed              = [W Z Y X]
+                // tmp1                = [0 Z 0 X]
+                // tmp2                = [W 0 Y 0]
+                // tmp3=ROTL(16, tmp2) = [Y 0 W 0]
+                // tmp1 + tmp3         = [Y Z W X]
+                uint tmp1 = packed & 0x00FF00FF;
+                uint tmp2 = packed & 0xFF00FF00;
+                uint tmp3 = (tmp2 << 16) | (tmp2 >> 16);
```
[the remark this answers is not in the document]
ynse01: I added a PolyFill class for BitOperations on platforms other then 3.1. Added RotateLeft there, as the implementation is only a single line.

src/ImageSharp/Common/Helpers/BitOperations.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System.Runtime.CompilerServices;
+
+#if !NETCOREAPP3_1
```
gfoidl, line 6:
```suggestion
#if !NETCOREAPP3_1_OR_GREATER
```
is this macro available here?

Otherwise it's easy to miss if new targets are enabled. 
E.g. for .NET 6.0 this code would be hit, instead the one from the built-in BitOperations.

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
```diff
@@ -97,15 +99,15 @@ public void RunFallbackShuffle(ReadOnlySpan<byte> source, Span<byte> dest)
...
             ref uint sBase = ref Unsafe.As<byte, uint>(ref MemoryMarshal.GetReference(source));
             ref uint dBase = ref Unsafe.As<byte, uint>(ref MemoryMarshal.GetReference(dest));
-            int n = source.Length / 4;
+            nuint n = (nuint)(uint)source.Length / 4;
 
-            for (int i = 0; i < n; i++)
+            for (nuint i = 0; i < n; i++)
             {
-                uint packed = Unsafe.Add(ref sBase, i);
+                uint packed = Unsafe.Add(ref sBase, (int)i);
```
gfoidl, line 106:
```suggestion
                uint packed = Unsafe.Add(ref sBase, i);
```
The whole point of `nint` here is to avoid a sign extending move (`movsxd`)* in generated machine code. `nint` has the native size of the interge, i.e. it's like a `int` on 32-bit, and a `long` on 64-bit.
The cast back to `(int)` is actually a de-optimization. Just keep the `nint`.

\* this is a rather slow one, especially compared to normal `mov`-operations.
See [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYmpBkIYBvGgzNNdwAJ4YYDALIAKWADMG12w1wAhPDDQMvDyBAJQMALwAfAwAqly42C6CAIIAJqnOMG4+fgG8IRrU5hbuNnZOrqWeObj+DFxBGKER0XEJSSxpGZU1dfmFxTpV5ZluHna9AVwcjc1RsfGJKemjXr61eQU0AL5AA==).

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
```diff
@@ -97,15 +99,15 @@ public void RunFallbackShuffle(ReadOnlySpan<byte> source, Span<byte> dest)
         {
             ref uint sBase = ref Unsafe.As<byte, uint>(ref MemoryMarshal.GetReference(source));
             ref uint dBase = ref Unsafe.As<byte, uint>(ref MemoryMarshal.GetReference(dest));
-            int n = source.Length / 4;
+            nuint n = (nuint)(uint)source.Length / 4;
```
gfoidl, line 102: If you wonder why these casts were suggested: it's to tell the JIT that it's unsigned, so it can emit a bit-shift for the division by 4, instead of the expensive integer division.

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Generated/_Common.ttinclude
```diff
@@ -104,10 +104,10 @@ using SixLabors.ImageSharp.PixelFormats.Utils;
                 ref <#=fromPixelType#> sourceRef = ref MemoryMarshal.GetReference(sourcePixels);
                 ref <#=toPixelType#> destRef = ref MemoryMarshal.GetReference(destinationPixels);
 
-                for (int i = 0; i < sourcePixels.Length; i++)
+                for (nint i = 0; i < sourcePixels.Length; i++)
                 {
-                    ref <#=fromPixelType#> sp = ref Unsafe.Add(ref sourceRef, i);
-                    ref <#=toPixelType#> dp = ref Unsafe.Add(ref destRef, i);
+                    ref <#=fromPixelType#> sp = ref Unsafe.Add(ref sourceRef, (int)i);
+                    ref <#=toPixelType#> dp = ref Unsafe.Add(ref destRef, (int)i);
```
gfoidl, line 110:
```suggestion
                    ref <#=fromPixelType#> sp = ref Unsafe.Add(ref sourceRef, i);
                    ref <#=toPixelType#> dp = ref Unsafe.Add(ref destRef, i);
```

src/ImageSharp/Common/Helpers/BitOperations.cs
ynse01: Yes, it is defined. So, used it instead.

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
ynse01: With the nint usage now, the intermediate cast to uint is removed.

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Generated/_Common.ttinclude
ynse01: Removed this cast

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
ynse01: Ok, wasn't aware of this difference between the different machine code calls. Usin `nint`'s now

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
gfoidl: Hm, now I can't follow.

Here it should be `nuint` (note the **u** for unsigned), as [codegen](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYmpBkIYBvGgzNNdvHgwBiECAAo2MbABMA8l0kBPWWOxcADzA3hgwAHwMuP5cAJQMALyR0QEsADIwXADmGAAWDAD0DCga1OYWDBxWGLb25E4uHl6+McGhEVEx8UkMDlU8sSlc6Zk5+UUlpuY6DFzVDABC2FANbp4+fgFtYcldiZEOcwNDI9l5hcWl5TNHNUtQ9c5rzZtBITudAd0Ht7F91YMYqcxhdJmVprouP07tgAF6rJobVrvDpDb69KEAk4ZM7jS5TMw3aGLOGPRrrFpbFG7L77DHQv4M7Gjc4TUoAXyAA==) is best. The intermediate cast to `uint` removes the `movsxd` (which I don't like 😉).

For `Unsafe.Add` both `nint` and `nuint` produce equal code. Hence go here with `nuint`.

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
```diff
@@ -99,15 +99,15 @@ public void RunFallbackShuffle(ReadOnlySpan<byte> source, Span<byte> dest)
         {
             ref uint sBase = ref Unsafe.As<byte, uint>(ref MemoryMarshal.GetReference(source));
             ref uint dBase = ref Unsafe.As<byte, uint>(ref MemoryMarshal.GetReference(dest));
-            nuint n = (nuint)(uint)source.Length / 4;
+            nint n = (nint)source.Length / 4;
```
gfoidl, line 102:
```suggestion
            nuint n = (nuint)(uint)source.Length / 4;
```
See other comments.
This way, the JIT emits a bit-hack (a shit) instead of the costly integer division.

It's a bit tricky, but the JIT has to conform correctness, so we need to help the JIT a bit.

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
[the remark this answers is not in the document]
JimBobSquarePants: Can we double check to ensure we’re not introducing new guard checks here and that the codegen is not negatively affected? We’ll kill performance if so.

.gitattributes
```diff
@@ -87,13 +87,15 @@
 *.eot               binary
 *.exe               binary
 *.otf               binary
+*.pbm               binary
```
JimBobSquarePants, line 90: Sub module changes need to be reverted. These are undoing latest changes

src/ImageSharp/Common/Helpers/BitOperations.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        /// Rotates the specified value left by the specified number of bits.
+        /// </summary>
+        /// <param name="value">The value to rotate.</param>
+        /// <param name="offset">The number of bits to roate with.</param>
+        /// <returns>The rotated value.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static uint RotateLeft(uint value, int offset)
+            => (value << offset) | (value >> (32 - offset));
```
JimBobSquarePants, line 23: We’ve been using Numerics class for stuff like this

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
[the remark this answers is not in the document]
JimBobSquarePants: Ok. Checked. We’re good

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Generated/Abgr32.PixelOperations.Generated.cs
JimBobSquarePants: That’s ok

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
ynse01: The issue is with the combination of these two lines:

`nuint n = (nuint)(uint)span.Length / 4;`
`uint packed = Unsafe.Add(ref sBase, n);`

This all looks good on SharpIo, but when I compile it, it complains about the type of `n` in the second line. `Unsafe.Add` requires its second argument to be of type `int`. Hence the cast I put in earlier, which puts in the `movsxd` instruction again.

Please advise on the entire method.

gfoidl: > `Unsafe.Add` requires its second argument to be of type `int`

There are [overloads](https://apisof.net/catalog/c9442b39-7483-70da-b3b4-817e3281a1ed) for `nint` and `nuint` too. Note: `IntPtr` <-> `nint` and `UIntPtr` <-> `nuint` for the overloads, as these map transparently.

antonfirsov: Builds still need to support .NET Standard 1.3 which [doesn't seem to have them](https://apisof.net/catalog/c9442b39-7483-70da-b3b4-817e3281a1ed).

Maybe we can remove 1.3 in ImageSharp 2.0, I'm not aware of any significant platforms where upgrading to netstandard 2.0 is a significant blocker.  @JimBobSquarePants  thoughts?

JimBobSquarePants: Yeah kill it. Compatibility chart looks good to me

gfoidl: > still need to support .NET Standard 1.3

Ah that's the point I missed. Sorry for not remembering this.
If support for NS 1.3 is needed for ImageSharp 2.0 too, would it make sense to have a shim for missing methods?

antonfirsov: > Yeah kill it.

Though it's too big scope in scope for this PR I guess. There should be a separate PR, this one will be blocked on if that's acceptable. I wonder who can take it, my next item in the queue is #1884.

JimBobSquarePants: I can do it easily enough.

src/ImageSharp/Common/Helpers/BitOperations.cs
ynse01: Ah, using this existing class now.

.gitattributes
ynse01: Indeed, this was unintentional. Reverted.

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
ynse01: I agree that a too big a change for here. 

What do we do with IComponentShuffle in **_this_** PR ?

JimBobSquarePants: If you can wait 24 hours I can get the target framework removal merged in and you can avoid any hackery

JimBobSquarePants: OK... 3 hours. 1.3 is now gone!

ynse01: OK, great, than we can go with the `nuint` version of IComponentShuffle.
Checked it on [sharplab.io](https://sharplab.io/#gist:bae34a10615c1574892da98bf3e476af) and the generated machine code looks good to me!

JimBobSquarePants: That link still uses the cast no? `nuint` directly is better.
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYmpBkIYBvGgzNNdHXjwYBZABSwAZgys3cAITww0DLm4wGXgBKY1NzCICGMWwwAGsYABMGAF4GAFUuXGwnQQBBRMTHGBdPb18QjWoIiOIAdmjYhMSqiIBfGjagA=

ynse01: Nope, the compiler assumes the `int` overload than and throws an error.

JimBobSquarePants: Removing netstandard 1.3 was to facilitate the use of the overloads. Are we now saying they are not available?

JimBobSquarePants: So.... It looks like you need version 6 for this to work. Which is unsupported in .NET Core 2.1

ynse01: The [docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafe.add?view=net-6.0) are also not very clear. They speak about _.NET Platform Extensions_, but even those versions do not support the UIntPtr overload (which we would like to use to avoid the `movsxd` machine instruction).

The presence of a NuGet version, doesn't mean it has all the overloads, right? NS 2.0 might still have a reduced API and still miss the overload we're looking for.

If this is the case, I feel we're better off using `int` as loop variables. Let me know what you think is the best way forward here.

JimBobSquarePants: I agree. Let’s leave it as int for now

ynse01: OK, done that.

src/ImageSharp/Common/Helpers/Shuffle/IComponentShuffle.cs
ynse01: We're back to using `int`'s now, so this comment is no longer applicable. Thanks for the suggestions!

antonfirsov (dismissed):
LGTM, thanks!

This is a breaking change due to `IPixel<T>` interface extension.

antonfirsov (approved):
Nevermind, all good.

PR #1898, opened 2021-12-16
PNG: Verify CRC of IDAT chunk is correct

brianpopow:
### Prerequisites

### Description

This PR changes the PNG decoder to verify that the CRC of IDAT chunk is correct.

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -1206,22 +1206,29 @@ namespace SixLabors.ImageSharp.Formats.Png
 
             PngChunkType type = this.ReadChunkType();
 
-            // NOTE: Reading the Data chunk is the responsible of the caller
             // If we're reading color metadata only we're only interested in the IHDR and tRNS chunks.
             // We can skip all other chunk data in the stream for better performance.
-            if (type == PngChunkType.Data || (this.colorMetadataOnly && type != PngChunkType.Header && type != PngChunkType.Transparency))
+            if (this.colorMetadataOnly && type != PngChunkType.Header && type != PngChunkType.Transparency)
             {
                 chunk = new PngChunk(length, type);
 
                 return true;
             }
 
+            long pos = this.currentStream.Position;
             chunk = new PngChunk(
                 length: length,
                 type: type,
                 data: this.ReadChunkData(length));
 
             this.ValidateChunk(chunk);
+
+            // Restore the stream position for IDAT chunks, because it will be decoded later and
+            // was only read to verifying the CRC is correct.
+            if (type == PngChunkType.Data)
+            {
+                this.currentStream.Position = pos;
+            }
 
             return true;
         }
--- a/tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
@@ -255,7 +255,7 @@ namespace SixLabors.ImageSharp.Tests.Formats.Png
         public void Decode_MissingDataChunk_ThrowsException<TPixel>(TestImageProvider<TPixel> provider)
             where TPixel : unmanaged, IPixel<TPixel>
         {
-            System.Exception ex = Record.Exception(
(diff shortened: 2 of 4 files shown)


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1206,23 +1206,30 @@ private bool TryReadChunk(out PngChunk chunk)
...
                 data: this.ReadChunkData(length));
 
             this.ValidateChunk(chunk);
 
+            // Restore the stream position for IDAT chunks, because it will be decoded later and
+            // was only read to verifying the CRC is correct.
+            if (type == PngChunkType.Data)
+            {
+                this.currentStream.Position = pos;
+            }
```
antonfirsov, line 1231: Why do we need to read the IDAT chunk twice? Upon rewinding the stream, this means creating & filling `PngChunk.Data` twice. Can't we avoid this?

tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
```diff
@@ -297,13 +297,28 @@ public void Decode_InvalidColorType_ThrowsException<TPixel>(TestImageProvider<TP
...
+        [WithFile(TestImages.Png.Bad.WrongCrcDataChunk, PixelTypes.Rgba32)]
+        public void Decode_InvalidDataChunkCrc_ThrowsException<TPixel>(TestImageProvider<TPixel> provider)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            Exception ex = Record.Exception(
+                () =>
+                {
+                    using Image<TPixel> image = provider.GetImage(PngDecoder);
+                    image.DebugSave(provider);
+                });
```
antonfirsov, line 310:
```suggestion
            InvalidImageContentException ex = Assert.Throws<InvalidImageContentException>(
                () =>
                {
                    using Image<TPixel> image = provider.GetImage(PngDecoder);
                });
```

src/ImageSharp/Formats/Png/PngDecoderCore.cs
brianpopow: I dont see how we can avoid reading the stream twice. We need to read the stream completely once to be able to calculate the CRC. Then we are sure the byte stream is not corrupted and can start decoding it in `ZlibInflateStream`. Otherwise we would need to introduce PNG specific logic into `ZlibInflateStream`, which i dont think would be good. 
Also the compressed stream can span over multiple IDAT chunks, which would make it even more complicated.

Note also the `PngChunk.Data` is not filled twice, just once. Before this PR the `PngChunk.Data` was left empty.
So we have now with this PR an additional allocation of the size of `IDAT` chunk.

JimBobSquarePants: Does libpng check it? I know it’s optional but haven’t had the chance to dig through the code

brianpopow: yes, libpng does check it

brianpopow: One way to possible avoid a large allocation here for the `IDAT` chunk just to calculate the CRC could be to read the `IDAT` data in smaller portions from the stream (lets say 8192 bytes) and calculate the CRC on those smaller chunk's of data.

Is it worth introducing such special handling for calculating the CRC of `IDAT`?

antonfirsov: ZlibInflateStream operates over `BufferedReadStream`, so we can consider making CRC32 optional part of `BufferedReadStream` buffer fills, although it's definitely not a 2.0 optimization, I would open a separate issue to investigate this.

> read the `IDAT` data in smaller portions from the stream (lets say 8192 bytes) and calculate the CRC on those smaller chunk's of data.

If my understanding is correct, PNG-s with single very large IDAT chunks are rare, so it's probably not worth it, I would better investigate the `BufferedReadStream` approach.

antonfirsov (approved):
Looks good for 2.0, but might worth to investigate on-the-fly CRC32 calculation in the future.

PR #1899, opened 2021-12-17
Jpeg sampling factor validation #1894

### LINKED ISSUE #1894: Proposal to remove non-complient images from Jpeg test suite and introduce new validation check
Jpeg test suite has [invalid image](https://github.com/SixLabors/ImageSharp/blob/master/tests/Images/Input/Jpg/issues/fuzz/Issue824-IndexOutOfRangeException-C.jpg). It was introduced due to the index out of range exception but right now it's used to actually test jpeg decoder with reference image.

This image is invalid according to the itu spec: 
image

As it has `Hi` parameter equal to 10:
image

Current implementation does not validate this value in any way and tries to decode spectral data with input parameters. Test suite assumes this image to be parsable just because libjpeg can parse it. Problem is that this image is small enough not to be affected by artifacts, any real world size image would produce pixel mess with invalid sampling rate. Thus this image should not be used to test decoder.

As for the solution, we can force invalid values to `1` or do `% 4` but I don't think it's a good way to handle this. It's more convenient to get an exception of what is wrong (ideally with byte(s) start index) rather than random collection of pixels.

P.S.
This is not really relevant to the 'no itu spec compliance' reasoning but it also blocks my decoder optimization with ugly redundant checks.

br3aker:
### Prerequisites

### Description
Closes #1894. Moved mentioned image from decoder tests to 'throws exception' tests.

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -1033,13 +1033,13 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
                 // Validate: 1-4 range
                 if (h is < 1 or > 4)
                 {
-                    JpegThrowHelper.ThrowInvalidImageContentException($"Bad horizontal sampling factor: {h}");
+                    JpegThrowHelper.ThrowBadSampling(h);
                 }
 
                 // Validate: 1-4 range
                 if (v is < 1 or > 4)
                 {
-                    JpegThrowHelper.ThrowInvalidImageContentException($"Bad vertical sampling factor: {v}");
+                    JpegThrowHelper.ThrowBadSampling(v);
                 }
 
                 if (maxH < h)
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -1022,9 +1022,25 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
             int index = 0;
             for (int i = 0; i < componentCount; i++)
             {
+                // 1 byte: component identifier
+                byte componentId = this.temp[index];
+
+                // 1 byte: component sampling factors
                 byte hv = this.temp[index + 1];
                 int h = (hv >> 4) & 15;
                 int v = hv & 15;
+
+                // Validate: 1-4 range
+                if (h is < 1 or > 4)
+                {
+                    JpegThrowHelper.ThrowInvalidImageContentException($"Bad horizontal sampling factor: {h}");
+                }
+
+                // Validate: 1-4 range
+                if (v is < 1 or > 4)
+                {
+                    JpegThrowHelper.ThrowInvalidImageContentException($"Bad vertical sampling factor: {v}");
+                }
 
                 if (maxH < h)
                 {
@@ -1036,10 +1052,19 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
                     maxV = v;
                 }
 
-                var component = new JpegComponent(this.Configuration.MemoryAllocator, this.Frame, this.temp[index], h, v, this.temp[index + 2], i);
+                // 1 byte: quantization table destination selector
+                byte quantTableIndex = this.temp[index + 2];
+
+                // Validate: 0-3 range
+                if (quantTableIndex > 3)
+                {
+                    JpegThrowHelper.ThrowBadQuantizationTableIndex(quantTableIndex);
+                }
+
+                var component = new JpegComponent(this.Configuration.MemoryAllocator, this.Frame, componentId, h, v, quantTableIndex, i);
 
                 this.Frame.Components[i] = component;
-                this.Frame.ComponentIds[i] = component.Id;
+                this.Frame.ComponentIds[i] = componentId;
 
                 index += componentBytes;
             }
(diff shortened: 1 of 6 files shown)


src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -1022,10 +1022,26 @@ private void ProcessStartOfFrameMarker(BufferedReadStream stream, int remaining,
...
+                if (h is < 1 or > 4)
+                {
+                    JpegThrowHelper.ThrowInvalidImageContentException($"Bad horizontal sampling factor: {h}");
+                }
+
+                // Validate: 1-4 range
+                if (v is < 1 or > 4)
+                {
+                    JpegThrowHelper.ThrowInvalidImageContentException($"Bad vertical sampling factor: {v}");
+                }
```
antonfirsov, line 1043: Wish we had an easy way to construct tests that can exercise all the error cases.

antonfirsov (approved):
Looks good, thanks!

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
br3aker: We can expose this marker parsing methods as static ones and them separately which is good enough IMO. Somewhat automatic fuzzer would also be good but too much work for now, working on decoder optimizations for now which also need a lot of tests.

antonfirsov: Testing static methods against constructed binary data is also a non trivial effort, but might worth investigating eventually.

One more thing for this PR:
Some of the coverage change may indicate that unused methods, if so, we should delete them:
https://app.codecov.io/gh/SixLabors/ImageSharp/compare/1899/changes

br3aker: There actually were 2 unused methods, removed them. I've also introduced a new helper throw method to remove those doubled hand-written sampling exceptions.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -1022,10 +1022,26 @@ private void ProcessStartOfFrameMarker(BufferedReadStream stream, int remaining,
...
                 int v = hv & 15;
 
+                // Validate: 1-4 range
+                if (h is < 1 or > 4)
+                {
+                    JpegThrowHelper.ThrowBadSampling(h);
+                }
+
+                // Validate: 1-4 range
+                if (v is < 1 or > 4)
```
gfoidl, line 1040: This reads really nice, but codegen has to two conditional jumps. With uint and subtraction, there could be only I conditional jump.

Therefore I propose a helper like
```c#
static bool IsOutOfRange(int value, int min, int max)
    => (uint)(value - min) > (uint)(max - min);
```
then it can be written as
```suggestion
                if (IsOutOfRange(v, 1, 4))
```
The JIT will use the constants, and produce special inlined code for that.

For codegen see [example on sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKBuIGYACMxgYUYG8bGfmnjySRsAgQANowCy5ABQBLAHYZGANwCUjALwA+RjJWM5uRgB5G5RtEa6UagNzdeDZoOGiJk0vKWqNOxgCSuADyAK4YwQBmAErYCgDmMPpo5im2DtS8jI48zgJCisoAYqKyhb45nJVZcpF6Bkam5pZQ1oy21bxcmVm9PAAqABZQEADuMvadPAC+U9k9fcQA7KoZWbML81l5ruUlEF7l6pXdfTy1ekFhETFxicmp7Wpqc6dnWUMj45Ob63Nzy1WlQ2vEqlQADlA5CpsBgYC4hCJxIEQuEorEEkkjtgxKEYClyvhFASfPhsAgXr9eP4ZDJQoU1PocXjGHBGESFM82rT6UpGWSEKz2YpnmtQZtIdDYfD8swUIxPmMJlpdBhhmNGAoYKNGABRBBgGDgjByCAKCYZaZAA==)

antonfirsov: Note that this is not a hot path, but if we create a reusable `Numerics.IsOutOfRange` method, I'm not against. Up to @br3aker to decide.

br3aker: I'm up for a helper method. I personally do and do not like pattern matching at the same time. And if it has worse performance - let's introduce `IsOutOfRange` method.

antonfirsov: >  I personally do and do not like pattern matching at the same time.

Same here. It fragments the language, pushing it towards the C++ direction where there are literally 5-10 different ways to achieve the same thing sometimes.

br3aker: Done, I've also added tests for new method.

tests/ImageSharp.Tests/Common/NumericsTests.cs
```diff
@@ -97,5 +97,25 @@ public void DivideCeil_RandomValues(int seed, int count)
...
             }
         }
+
+        private static bool IsOutOfRange_ReferenceImplementation(int value, int min, int max) => value < min || value > max;
+
+        [Theory]
+        [InlineData(1, 100)]
+        public void IsOutOfRange(int seed, int count)
+        {
+            var rng = new Random(seed);
```
gfoidl, line 107: Instead of random numbers to test, I'd rather test edge-cases.

E.g.
```
min := 7
max := 92

value = { int.MinValue, -1, 0, 1, 6, 7, 8, 91, 92, 93, int.MaxValue }
```
then the whole range is covered.

br3aker: Agreed, fixed.

PR #1906, opened 2021-12-21
Expose ImageFrame<TPixel>.PixelBuffer

JimBobSquarePants:
### Prerequisites

### Description

We need to expose this property to allow migrating the ImageSharp.Drawing brush API to the latest ImageSharp v2 versions without runtime error.

### DIFF
--- a/src/ImageSharp/ImageFrame{TPixel}.cs
+++ b/src/ImageSharp/ImageFrame{TPixel}.cs
@@ -145,13 +145,8 @@ namespace SixLabors.ImageSharp
             source.PixelBuffer.FastMemoryGroup.CopyTo(this.PixelBuffer.FastMemoryGroup);
         }
 
-        /// <summary>
-        /// Gets the image pixels. Not private as Buffer2D requires an array in its constructor.
-        /// </summary>
-        internal Buffer2D<TPixel> PixelBuffer { get; private set; }
-
         /// <inheritdoc/>
-        Buffer2D<TPixel> IPixelSource<TPixel>.PixelBuffer => this.PixelBuffer;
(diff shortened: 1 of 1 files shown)


src/ImageSharp/ImageFrame{TPixel}.cs
```diff
@@ -145,13 +145,8 @@ internal ImageFrame(Configuration configuration, ImageFrame<TPixel> source)
...
         }
 
-        /// <summary>
-        /// Gets the image pixels. Not private as Buffer2D requires an array in its constructor.
-        /// </summary>
-        internal Buffer2D<TPixel> PixelBuffer { get; private set; }
-
         /// <inheritdoc/>
-        Buffer2D<TPixel> IPixelSource<TPixel>.PixelBuffer => this.PixelBuffer;
+        public Buffer2D<TPixel> PixelBuffer { get; private set; }
```
antonfirsov, line 149: Can `<inheritdoc/>` publicly inherit docs from an internal interface?

JimBobSquarePants: I honestly have no idea. I think it triggers a copy of the docs though during build rather than reference the original

antonfirsov: Ok, we check it during doc generation, and fix if it doesn't work.

antonfirsov (approved):
LGTM.

src/ImageSharp/ImageFrame{TPixel}.cs
JimBobSquarePants: Intellisense seems to be working in consuming project.

PR #1915, opened 2021-12-29
Remove return value documentation for void return types.

wischi-chr:
### Prerequisites

### Description

Very minor code documentation fixes.


dlemstra (approved):
Tnx for fixing this for us.

JimBobSquarePants:
Needs to be fixed at the template level.

https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Formats/ImageExtensions.Save.tt

wischi-chr:
Should I change it only in the template or in both files?

dlemstra:
Both files should be changed.

wischi-chr:
The patch now includes the fix in the template as well.

PR #1907, opened 2021-12-21
Don't throw when Position is out of range. 

JimBobSquarePants:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

According to the docs for [Stream.Position](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.position?view=net-6.0#remarks)

>Seeking to any location beyond the length of the stream is supported.

Removing our explicit throw make the implementation compliant. Fix #1903

PR #1918, opened 2022-01-02
Support for XMP metadata

### LINKED ISSUE #858: Support for XMP metadata
As discussed on Gitter, ImageSharp doesn't support XMP at the moment, it would be interesting to have support for it

ynse01:
### Prerequisites

### Description
This PR fixes #858.

Adding the ability to read and write XMP metadata from the following image formats:

- GIF
- JPEG
- PNG
- TIFF
- WEBP

XMP metadata is placed in a dedicated XmpProfile class, which exposes the raw XML data.

Some highlights on the changes:
- Relies on a solution for bug #1907 
- WebP decoder is no longer dependant on the order in which the headers appear in the file. This also exposes EXIF on some images where they where hidden previously.
- Extended JPEG is supported, where the XMP metadata is larger then 64 kB. Added a test images for this specific case.
- Support the image that @rickbrew shared
- GIF does not provide a size of the XMP metadata, so I opted to read in blocks of 256 bytes and combine them later.
- IGifExtension got a new property, to predict the buffer size required during encoding.
- Added some test images, to ensure all formats have at least 1 image with XMP metadata. Re-used existing test images where applicable.
- Unit test for all image formats.

Open points, not part of this review:
- XmpProfile does not provide any strongly typed classes for the XMP properties.
- Metadata profiles should use a MemoryAllocator, as these profiles get quite large for realistic images. This is most notable in the GIF decoding. Doing so, causes a breaking change though, as it makes Metadata classes `IDisposable` and require the constructors to have a `Configuration` argument.


src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -250,33 +251,50 @@ private void ReadLogicalScreenDescriptor()
...
             {
-                this.stream.Skip(appLength);
-                int subBlockSize = this.stream.ReadByte();
+                this.stream.Read(this.buffer, 0, GifConstants.ApplicationBlockSize);
+                bool isXmp = true;
+                ReadOnlySpan<byte> idBytes = GifConstants.XmpApplicationIdentificationBytes;
+                for (int i = 0; i < idBytes.Length; i++)
+                {
+                    isXmp &= this.buffer[i] == idBytes[i];
+                }
```
gfoidl, line 271:
```suggestion
                bool isXmp = buffer.AsSpan().StartsWith(idBytes);
```
?

`StartsWith` is optimized (and vectorized).

src/ImageSharp/Formats/Gif/Sections/GifXmpApplicationExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// </summary>
+        /// <param name="stream">The stream to read from.</param>
+        /// <returns>The XMP metadata</returns>
+        /// <exception cref="ImageFormatException">Thrown if the XMP block is not properly terminated.</exception>
+        public static GifXmpApplicationExtension Read(Stream stream)
+        {
+            // Read data in blocks, until an \0 character is encountered.
+            // We overshoot, indicated by the terminatorIndex variable.
+            const int bufferSize = 256;
+            var list = new List<byte[]>();
```
gfoidl, line 34: This is quite allocation heavy. Would it make sense to build this around [ReadOnlySequence<T>](https://docs.microsoft.com/en-us/dotnet/standard/io/buffers#readonlysequencet) and / or [Pipelines](https://docs.microsoft.com/en-us/dotnet/standard/io/pipelines)?

Of course, functionality can be brought in and optimized later on.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1364,6 +1371,21 @@ private bool TryReadTextKeyword(ReadOnlySpan<byte> keywordBytes, out string name
...
+            bool result = keywordBytes.Length == expected.Length;
+            if (result)
+            {
+                for (int i = 0; i < keywordBytes.Length; i++)
+                {
+                    result |= keywordBytes[i] == expected[i];
+                }
+            }
+
+            return result;
```
gfoidl, line 1386:
```suggestion
            return keywordBytes.SequenceEqual(expected);
```

src/ImageSharp/Formats/Webp/WebpDecoderCore.cs
```diff
@@ -440,6 +447,38 @@ private void ParseOptionalExtendedChunks(WebpChunkType chunkType, WebpFeatures f
...
+                case WebpChunkType.Exif:
+                    uint exifChunkSize = this.ReadChunkSize();
+                    if (this.IgnoreMetadata)
+                    {
+                        this.currentStream.Skip((int)exifChunkSize);
+                    }
+                    else
+                    {
+                        byte[] exifData = new byte[exifChunkSize];
+                        this.currentStream.Read(exifData, 0, (int)exifChunkSize);
```
gfoidl, line 459: Stream.Read doesn't guarantee that the requested length is actually read. It may be less.
Is that case covered? (on other places as well)

src/ImageSharp/Formats/Png/PngDecoderCore.cs
ynse01: Good suggestion, made it into a one-liner now.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
ynse01: Good suggestion

src/ImageSharp/Formats/Webp/WebpDecoderCore.cs
br3aker: It is handled by `BufferedReadStream` stream wrapper:

https://github.com/SixLabors/ImageSharp/blob/fe044d131c45d89aa52281e2eb442c82b4424ce1/src/ImageSharp/Formats/Webp/WebpDecoderCore.cs#L81

In actual `Read` methods:

https://github.com/SixLabors/ImageSharp/blob/fe044d131c45d89aa52281e2eb442c82b4424ce1/src/ImageSharp/IO/BufferedReadStream.cs#L342-L352

Shouldn't we also save stream reference as explicit `BufferedReadStream` instead of abstract `Stream` for de-virtualization?

JimBobSquarePants:
Will do a proper review later on but wasn’t specifically planning on supporting this for V2 so depending on the design we might have to make this internal to ensure we have the time to get it right.

Regarding Metadata, we cannot use the allocator since we require the ability to explicitly set profile instances to null.

src/ImageSharp/Formats/Gif/Sections/GifXmpApplicationExtension.cs
JimBobSquarePants: I’d like to revisit a lot of stuff for V3 where we can trim framework targets.

src/ImageSharp/Formats/Webp/WebpDecoderCore.cs
gfoidl: > Shouldn't we also save stream reference as explicit `BufferedReadStream` instead of abstract `Stream` for de-virtualization?

Yep that would be better.
JIT's dynamic PGO (profile guided optimization) could handle this, but right now this is opt-in by setting the environment varialbe `export DOTNET_TieredPGO=1` in the host-process.

Besides that it would be clear that's a `BufferedReadStream` and not the base `Stream`, so concerns about reading less than the requested lenght won't arise 😉.

src/ImageSharp/Metadata/Profiles/XMP/XmpProfile.cs
```diff
@@ -0,0 +1,199 @@
...
+            {
+                this.Data = new byte[other.Data.Length];
+                other.Data.AsSpan().CopyTo(this.Data);
+            }
+        }
+
+        /// <summary>
+        /// Gets the rax XML document containing the XMP profile.
+        /// </summary>
+        public XDocument Document
```
antonfirsov, line 63: `XDocument` is not read-only. A user adding nodes to the doc may expect their changes to be written back by re-encoding. We definitely don't want to do that, since we are not validating the contents.

My recommendation is to turn this into a `.GetDocument()` method that will create a new document on each call.

@JimBobSquarePants thoughts?

src/ImageSharp/Metadata/Profiles/XMP/XmpProfile.cs
```diff
@@ -0,0 +1,199 @@
...
+            using var writer = new StreamWriter(stream, Encoding.UTF8);
+            this.document.Save(writer);
+            this.Data = stream.ToArray();
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode() => base.GetHashCode();
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
```
antonfirsov, line 141: I don't see equality being implemented for other profiles. Also see previous comment. We should drop this feature IMO.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -654,6 +655,43 @@ private void WriteExifChunk(Stream stream, ImageMetadata meta)
...
+
+            if (xmpData.Length == 0)
+            {
+                return;
+            }
+
+            byte[] payload = new byte[xmpData.Length + PngConstants.XmpKeyword.Length + 5];
+            PngConstants.XmpKeyword.CopyTo(payload);
+            int bytesWritten = PngConstants.XmpKeyword.Length;
+            payload[bytesWritten++] = 0; // Keyword string terminator
```
JimBobSquarePants, line 686: Why zero out here if the array is already zeroed? Can we do this without allocating and use our pool?

src/ImageSharp/Metadata/Profiles/XMP/XmpProfile.cs
JimBobSquarePants: The profile is supposed to be writeable no? If not I don't see a clear purpose for this. 

I'm not altogether happy with the lack of types and read/write functionality tbh. Our metadata profiles EXIF/XMP should follow a similar structure.

src/ImageSharp/Metadata/Profiles/XMP/XmpProfile.cs
JimBobSquarePants: Yeah, definitely no need for this.

JimBobSquarePants (commented):
Lovely bit of work so far but I'm not sure about where we're going with this. Needs more discussion.

src/ImageSharp/Metadata/Profiles/XMP/XmpProfile.cs
antonfirsov: From https://github.com/SixLabors/ImageSharp/issues/858#issuecomment-998446230:

> Creating an object model against XMP is IMO not worth the effort unless you find strong need for it.

Considering the format seems to be quite complicated, I can agree with this. I think preserving the data on load+resave has a value on it's own, we just need to make sure the types are extensible enabling to add actual contents.

JimBobSquarePants: We already had preservation in Tiff using the raw byte array. I can go with preservation but it feels half baked to expose the thing in a readonly format while also expecting the user to actually parse the document.

antonfirsov: I think half-baked is fine, as long as: (1) users find it valuable, (2) the solution is extensible to make it fully baked.

I don't think we have the capacity to validate and review a PR adding full XMP support now.

antonfirsov:
> Metadata profiles should use a MemoryAllocator, as these profiles get quite large for realistic images.

> Regarding Metadata, we cannot use the allocator since we require the ability to explicitly set profile instances to null.

The main challenge here is that metadata profiles shall be `IDisposable` in order to hold allocated buffers. (This way profile setters can dispose them too.) In case we are concerned that profiles (of any type) can get allocation-heavy, we should revisit the API.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1045,13 +1046,14 @@ private void ReadCompressedTextChunk(PngMetadata metadata, ReadOnlySpan<byte> da
         /// </summary>
         /// <param name="metadata">The metadata to decode to.</param>
         /// <param name="data">The <see cref="T:Span"/> containing the data.</param>
-        private void ReadInternationalTextChunk(PngMetadata metadata, ReadOnlySpan<byte> data)
+        private void ReadInternationalTextChunk(ImageMetadata metadata, ReadOnlySpan<byte> data)
```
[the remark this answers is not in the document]
ynse01: Because the `XmpProfile` property is on the `ImageMetadata` class, not on `PngMetadata`

src/ImageSharp/Metadata/Profiles/XMP/XmpProfile.cs
ynse01: Some background on why there are equality checks in `XmpProfile`

Other profiles can rely on their byte array equality check.
In XMP however, there can be non-relevant changes, like BOM marker, whitespace, ordering of elements and attributes.

All these **_do_** change the byte array, but **_do not_** change the XML. Hence these should not be causing the equality check of `XmpProfile` to return `false`.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -121,11 +122,8 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream, CancellationToken
             // Write the comments.
             this.WriteComments(gifMetadata, stream);
 
-            // Write application extension to allow additional frames.
-            if (image.Frames.Count > 1)
```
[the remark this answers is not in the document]
ynse01: There is a (previously redundant) check for the same in `WriteApplicationExtensions` on line 332

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -420,14 +427,25 @@ private void WriteGraphicalControlExtension(GifFrameMetadata metadata, int trans
         private void WriteExtension<TGifExtension>(TGifExtension extension, Stream stream)
             where TGifExtension : struct, IGifExtension
         {
-            this.buffer[0] = GifConstants.ExtensionIntroducer;
-            this.buffer[1] = extension.Label;
+            byte[] buffer;
+            int extensionSize = extension.ContentLength;
+            if (extensionSize > this.buffer.Length - 3)
+            {
+                buffer = new byte[extensionSize + 3];
```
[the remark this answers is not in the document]
ynse01: Nope, this was an oversight. I'll change it to use Memory Allocator.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
ynse01: Indeed, we can just skip over these bytes

src/ImageSharp/Metadata/Profiles/XMP/XmpProfile.cs
antonfirsov: My point is that other profiles do not implement equality in any form (neither byte-array, nor semantic). I don't see how / why should a complex type like this implement equality out of the box. If a user needs it, they can implement XML-based comparison in their own code.
Out of curiosity: Do you have an important personal use case here?

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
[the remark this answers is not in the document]
JimBobSquarePants: Line 332 checks the repeat count (metadata) not the frame count. We need both.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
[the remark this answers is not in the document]
JimBobSquarePants: Ah yes, didn't spot the stuff at 1107 👍

src/ImageSharp/Metadata/Profiles/XMP/XmpProfile.cs
jubilant-enigma: I was actually making some headway on trying to create an object model against XMP. It's far from complete, but I don't think it would be *too* complicated to have a user-friendly way to read and modify XMP. Would probably still be best if it is (potentially) added in a later PR though, and this PR just exposes the `XDocument`.

JimBobSquarePants: Point 1 is the important one for V2. I would expect us to deliver a full mutable API for V3. 

`.GetDocument()` would definitely be better API-wise. I also don't believe we should expose the raw `Data` property and use `ToByteArray()` like our `ExifProfile`.

ynse01: Ok, sounds like a plan.

The `ToByteArray` method return a clone, as opposed to the `Data` property previously. The `ExifProfile` does this implicitly, as it parses its internal strong types.

The `GetDocument` method does no longer influence the original byte array.

And last but not least, the equality methods and operators has been removed.

@JimBobSquarePants This would then be the V2 API, as you describe.
@jubilant-enigma Feel free to implement your object model on top of this PR.

src/ImageSharp/Metadata/Profiles/XMP/XmpProfile.cs
ynse01: Equality methods and operators are removed now.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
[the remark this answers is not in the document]
ynse01: Clear, implemented both checks inside `WriteApplicationExtensions` now.

ynse01:
I had to apply the fix of #1907 to `BufferedReadStream` also and updated its tests accordingly. Hope this is OK with you guys.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -73,6 +74,16 @@ internal sealed class JpegDecoderCore : IRawJpegData, IImageDecoderInternals
         /// </summary>
         private byte[] iptcData;
 
+        /// <summary>
+        /// Whether the image has a XMP data.
+        /// </summary>
+        private bool isXmp;
```
antonfirsov, line 80: Naming: `hasXmp`

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -686,6 +712,21 @@ private void ProcessApp1Marker(BufferedReadStream stream, int remaining)
...
                 }
             }
+
+            if (ProfileResolver.IsProfile(profile, ProfileResolver.XmpMarker))
+            {
+                this.isXmp = true;
+                if (this.xmpData is null)
+                {
+                    // The first 29 bytes will be skipped, because this is Jpeg specific
+                    this.xmpData = profile.AsSpan(XmpNsLength).ToArray();
```
antonfirsov, line 722: If this can grow large, we should request `profile`  from `MemoryAllocator` so it's pooled, to at least avoid double GC allocation. (Once with `new byte[remaining]` once with `.ToArray()`)

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -466,6 +467,54 @@ private void WriteIptcProfile(IptcProfile iptcProfile)
...
+            if (xmpProfile is null)
+            {
+                return;
+            }
+
+            const int XmpOverheadLength = 29;
+            const int Max = 65533;
+            const int MaxData = Max - XmpOverheadLength;
+
+            byte[] data = xmpProfile.ToByteArray();
```
antonfirsov, line 488: We can expose an *internal* `byte[] XmpProfile.Data` property to avoid an allocation and copying of the byte array.

antonfirsov (commented):
Some memory optimization suggestions. Commented on Jpeg codec, but applicable to all.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
ynse01: There are already 3 other properties with this naming convention, `isExif`, `isIptc` and `isIcc`. 
Shall I update all 4 then ?

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
ynse01: Makes sense. `XmpProfile` is now effectively read-only from outside, which means we can assume the `data` field can be written directly in all the Encoders.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
ynse01: I have reduced the allocations, but MemoryAllocator is not an option for profiles, see previous comment from @JimBobSquarePants .

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
JimBobSquarePants: Yes please!

ynse01: With pleasure !

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -326,15 +324,24 @@ private int GetTransparentIndex<TPixel>(IndexedImageFrame<TPixel> quantized)
...
         /// <param name="stream">The stream to write to.</param>
+        /// <param name="frameCount">The frame count fo this image.</param>
         /// <param name="repeatCount">The animated image repeat count.</param>
-        private void WriteApplicationExtension(Stream stream, ushort repeatCount)
+        /// <param name="xmpProfile">The XMP metadata profile. Null if profile is not to be written.</param>
+        private void WriteApplicationExtensions(Stream stream, int frameCount, ushort repeatCount, XmpProfile xmpProfile)
         {
-            // Application Extension Header
+            // Application Extension: Loop repeat count.
             if (repeatCount != 1)
```
JimBobSquarePants, line 333: Looks like you're not actually using the frameCount. 

```c#
 if (frameCount > 1 && repeatCount != 1)
```

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -654,6 +655,38 @@ private void WriteExifChunk(Stream stream, ImageMetadata meta)
...
+            }
+
+            var xmpData = meta.XmpProfile.Data;
+
+            if (xmpData.Length == 0)
+            {
+                return;
+            }
+
+            byte[] payload = new byte[xmpData.Length + PngConstants.XmpKeyword.Length + 5];
```
JimBobSquarePants, line 683: This should be allocated using our memory allocator

JimBobSquarePants (changes requested):
Just a single missed allocation optimization opportunity remaining and this looks good to go in. I've gotta say, I'm super impressed with how quickly you've added this across multiple formats. Really nice 👍

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
ynse01: Oops. Checking it now.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
ynse01: Indeed

ynse01: @JimBobSquarePants Test results on Github vary now, for the same Png round trip test case. I do NOT get this same error locally.

Can it be that this last memory optimization is causing instability? If so, I do not understand what might be the cause.

ynse01: Found it, the Memory Allocator does not zero out the buffers it allocates.
Doing this explitly now.

Btw: I found some unrelated methods in `PngEncoderCore` that suffer from the same allocation issue. I'll fix these also.

JimBobSquarePants:
Let's get this merged 😄  Thanks @ynse01 for the code and everyone else for the excellent reviews. Much appreciated!

Issue #1921, opened 2022-01-03
Possible invalid/missing EXIF IFD end directory link


danpetitt:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
The IFD does not contain the final 4 byte 0 after the last directory which makes it hard to determine if there are any other linked directories or not.

From the TIFF specification (https://www.itu.int/itudoc/itu-t/com16/tiff-fx/docs/tiff6.pdf):
> An Image File Directory (IFD) consists of a 2-byte count of the number of directory entries (i.e., the number of fields), followed by a sequence of 12-byte field entries, followed by a 4-byte offset of the next IFD (or 0 if none). **(Do not forget to write the 4 bytes of 0 after the last IFD.)**

(the emphasis on the final sentence in brackets is mine).

### Steps to Reproduce
```c#
using( var image = Image.Load( "metadata.jpg" ) )
{
  var exif = new ExifProfile();
  exif.SetValue<byte[]>( ExifTag.XPAuthor, Encoding.GetEncoding( "UCS-2" ).GetBytes( "Dan Petitt" ) );
  image.Metadata.ExifProfile = exif;

  image.Save( "metadata-with-exif.jpg" );
}
```

Binary dump of above shows the beginning of the EXIF/TIFF chunk (showing the byte order marker bytes) which is where the data offsets will start from (`30`:
![image](https://user-images.githubusercontent.com/1733292/147985702-9845739f-2e89-4189-ba49-0a7a08a7b0c7.png)

Below shows the first offset (which is always 8 from the start offset):
![image](https://user-images.githubusercontent.com/1733292/147985869-f69b7790-0dd0-4516-bb06-b98ba0201922.png)

Then we start the 12 byte data chunks for each bit of metadata from offset `38`; the first 2 bytes shows the directory count:
![image](https://user-images.githubusercontent.com/1733292/147985949-07b98aac-7a54-4138-9166-618c79541377.png)

Next 10 bytes is the component size, count and the last 4 bytes will be the offset where the actual data lives which is `1A` or 26 + 30 start offset = 56:
[...]

PR #1923, opened 2022-01-04
EXIF IDF pointer

IldarKhayrutdinov:
### Prerequisites

### Description

Fixes for #1921


antonfirsov:
Can we add a test case based on the #1921 repro steps?

IldarKhayrutdinov:
I needs to additional reviewing of EXIF writer methods, it seems there are problems of IFD writing (regarding next IFD pointers and gps sub IFD). Next IFD pointer should be written after each IFD, but at the moment it's written only after the last IFD.

src/ImageSharp/Metadata/Profiles/Exif/ExifWriter.cs
```diff
@@ -56,39 +55,38 @@ public byte[] GetData()
...
+            byte[] result = new byte[length];
 
             int i = 0;
 
             // The byte order marker for little-endian, followed by the number 42 and a 0
             ExifConstants.LittleEndianByteOrderMarker.CopyTo(result.AsSpan(start: i));
             i += ExifConstants.LittleEndianByteOrderMarker.Length;
 
-            uint ifdOffset = ((uint)i - startIndex) + 4U;
-            uint thumbnailOffset = ifdOffset + ifdLength + exifLength + gpsLength;
```
IldarKhayrutdinov, line 84: in fact, the thumbnail isn't written, this is an unnecessary line

JimBobSquarePants (approved):
This looks great! Thanks for stepping up 👍

PR #1925, opened 2022-01-06
HuffmanTable related code improvements

br3aker:
### Prerequisites

### Description
Main purpose of this PR is to reduce huffman-table-related code: previously huffman tables were built only on request from each component which led some redundant checks and extra code in the binary decoder. Now tables are build in the ctor.

I've also marked some `HuffmanTable` field private, it doesn't really add anything except code readability (imo) but it felt like a right thing to do considering those fields are unsafe raw pointers.

Updated code does contain a little bit of optimizations. There's less empty calls on each scan parsing routine but you know, there's not so many scans per image :P

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
@@ -14,9 +14,20 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
     internal unsafe struct HuffmanTable
     {
         /// <summary>
+        /// Derived from the DHT marker. Sizes[k] = # of symbols with codes of length k bits; Sizes[0] is unused.
+        /// </summary>
+        public fixed byte Sizes[17];
+
+        /// <summary>
         /// Derived from the DHT marker. Contains the symbols, in order of incremental code length.
         /// </summary>
         private fixed byte values[256];
+
+        /// <summary>
+        /// Contains the largest code of length k (0 if none). MaxCode[17] is a sentinel to
+        /// ensure <see cref="HuffmanScanBuffer.DecodeHuffman"/> terminates.
+        /// </summary>
+        public fixed ulong MaxCode[18];
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
```diff
@@ -62,20 +55,7 @@ internal unsafe struct HuffmanTable
...
-        public void Configure()
-        {
-            if (this.isConfigured)
-            {
-                return;
-            }
+            Unsafe.CopyBlockUnaligned(ref this.values[0], ref MemoryMarshal.GetReference(values), (uint)values.Length);
 
             Span<char> huffSize = stackalloc char[257];
             Span<uint> huffCode = stackalloc uint[257];
```
br3aker, line 61: @antonfirsov this concerns me a bit, are we okay allocating 1.5kb on stack here? This part was taken from current master implementation.

JimBobSquarePants: Hmmm yeah. Never did the maths on that previously. Anything larger than a kilobyte is generally considered a concern.

br3aker:
Strange crash, all tests passed locally. My guess that I've screwed up with table structure, will update in a bit.

src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
br3aker: It would look ugly but we are already pooling some memory for table parsing:

https://github.com/SixLabors/ImageSharp/blob/a54a19fa994ebed83f3c76a30482c96618a8835d/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs#L1142-L1144

We can pool those 256 bytes + 1.5kb in one package and split it into 3 different spans for table building workspace.

br3aker: Marked this as a draft until I'll implement this.

br3aker:
Actually I've decided to redo this from scratch with `stackalloc` fix, that test crash concerns me.

Issue #839, opened 2019-02-20
Image.Load throws ExecutionEngineException


Metalnem:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
**Image.Load** throws **ExecutionEngineException** when loading malformed JPG file. The stack trace looks almost the same as in the #827, which probably means the previous issue hasn't been fixed completely.

### Steps to Reproduce
Run the following code snippet with the ```path``` variable containing the path to the file from the [attached archive](https://github.com/SixLabors/ImageSharp/files/2884291/ExecutionEngineException.zip):

```csharp
Image.Load(path);
```

SOS exception details:

```
Exception object: 000000018eb911f8
Exception type:   System.ExecutionEngineException
Message:          <none>
InnerException:   <none>
StackTrace (generated):
<none>
StackTraceString: <none>
HResult: 80131506
```

Stack trace:

```
00007FFEE12723E0 00007fff785f923e [FaultingExceptionFrame: 00007ffee12723e0] 
00007FFEE1273850 000000011DC57772 SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder.HuffmanTable..ctor(SixLabors.Memory.MemoryAllocator, System.ReadOnlySpan`1<Byte>, System.ReadOnlySpan`1<Byte>) [/Users/Metalnem/Temp/ImageSharp/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs @ 114]
00007FFEE12738E0 000000011DC56D55 SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.ProcessDefineHuffmanTablesMarker(Int32)
00007FFEE1274000 000000011DC51FF7 SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.ParseStream(System.IO.Stream, Boolean) [/Users/Metalnem/Temp/ImageSharp/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs @ 313]
[...]

PR #1926, opened 2022-01-06
Jpeg HuffmanTable improvements

br3aker:
### Prerequisites

### Description
This PR removes redundant `Configure()` calls for `HuffmanTable` and removes stack pressure from `stackalloc` calls.

### TODO
- [x] Remove `Configure()` method 
- [x] Remove `stackalloc` calls
- [x] Create out-of-bounds huffman table test image

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
@@ -57,27 +57,16 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
         {
             Unsafe.CopyBlockUnaligned(ref this.Values[0], ref MemoryMarshal.GetReference(values), (uint)values.Length);
 
-            Span<byte> huffSize = stackalloc byte[257];
             Span<uint> huffCode = stackalloc uint[257];
 
-            // Figure C.1: make table of Huffman code length for each symbol
-            int p = 0;
-            for (int j = 1; j <= 16; j++)
-            {
-                int count = codeLengths[j];
-                huffSize.Slice(p, count).Fill((byte)j);
-                p += count;
-            }
-
-            huffSize[p] = 0;
-
-            // Figure C.2: generate the codes themselves
+            // Generate codes
             uint code = 0;
-            int si = huffSize[0];
-            p = 0;
-            while (huffSize[p] != 0)
+            int si = 1;
+            int p = 0;
+            for (int i = 1; i <= 16; i++)
             {
-                while (huffSize[p] == si)
+                int count = codeLengths[i];
+                for (int j = 0; j < count; j++)
                 {
                     huffCode[p++] = code;
                     code++;
--- a/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
@@ -94,6 +94,17 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
                         code++;
                     }
 
+                    // 'code' is now 1 more than the last code used for codelength 'si'
+                    // in the valid worst possible case 'code' would have the least
+                    // significant bit set to 1, e.g. 1111(0) +1 => 1111(1)
+                    // but it must still fit in 'si' bits since no huffman code can be equal to all 1s
+                    // if last code is all ones, e.g. 1111(1), then incrementing it by 1 would yield
+                    // a new code which occupies one extra bit, e.g. 1111(1) +1 => (1)1111(0)
+                    if (code >= (1 << si))
+                    {
+                        JpegThrowHelper.ThrowInvalidImageContentException("Bad huffman table.");
+                    }
+
                     code <<= 1;
                     si++;
                 }
--- a/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
@@ -13,13 +13,6 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
     [StructLayout(LayoutKind.Sequential)]
     internal unsafe struct HuffmanTable
     {
-        private bool isConfigured;
-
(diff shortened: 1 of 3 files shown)


src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
```diff
@@ -62,20 +55,7 @@ internal unsafe struct HuffmanTable
         /// <param name="values">The huffman values</param>
         public HuffmanTable(ReadOnlySpan<byte> codeLengths, ReadOnlySpan<byte> values)
         {
-            this.isConfigured = false;
-            Unsafe.CopyBlockUnaligned(ref this.Sizes[0], ref MemoryMarshal.GetReference(codeLengths), (uint)codeLengths.Length);
```
JimBobSquarePants, line 66: As I recall this copy exists because the input span is allocated once and reused for each Huffman table.

br3aker: Which one? `codeLengths`? It may be unique for each table. Right now ctor copies input spans and then builds actual table representation during `Configure()` call - that's the only reason it copies first span param to internal array. While calling this method a couple of times with short-circuit `return` is not bad, it's impossible to pass workspace spans from allocator this way. That's why I want to build table in the ctor but that remote executor crash...

I hope it's okay if I can use github actions for some testing. That random crash is very strange and I'm not sure I can debug it locally.

JimBobSquarePants: Ah yes, The reason for the lazy invocation was to avoid building the tables to allow the validation of successive scan parameters for progressive images. See `HuffmanScanDecoder.CheckProgressiveData`. This is the same thing that libjpeg and libjpeg-turbo does. https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/jdhuff.c#L77

JimBobSquarePants: Also yes, experiment away. Maybe disable the codecov run first though to save time.

br3aker:
Gotcha

```
Message: 
    Assert.Throws() Failure
    Expected: typeof(SixLabors.ImageSharp.InvalidImageContentException)
    Actual:   typeof(System.IndexOutOfRangeException): Index was outside the bounds of the array.
    ---- System.IndexOutOfRangeException : Index was outside the bounds of the array.

  Stack Trace: 
    HuffmanTable.ctor(ReadOnlySpan`1 codeLengths, ReadOnlySpan`1 values) line 58
    JpegDecoderCore.ProcessDefineHuffmanTablesMarker(BufferedReadStream stream, Int32 remaining) line 1098
    JpegDecoderCore.ParseStream(BufferedReadStream stream, HuffmanScanDecoder scanDecoder, CancellationToken cancellationToken) line 358
    ...
```

One of the invalid test images was going out of bounds during new `HuffmanTable` ctor. The only reason it wasn't spotted before is that exposed code in `Configure()` was called after checks for invalid image which led to this crashes.

src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
br3aker: Hm, that check seems valid but table building is not that expensive, especially compared to exception path imo. Should I still remove `Configure` and move everything to the ctor for simplicity and reduced `stackalloc`s?

JimBobSquarePants: Let’s give it a go yeah!

JimBobSquarePants:
Ah nice, well done finding that. So now I guess we have to find a way to sanitise cheaply and then try to move the table building to the constructor.

br3aker:
It's actually a bug, there's a possibility for malformed image go out of bounds without any exceptions in current master. I found a way to check that without a noticable slowdown. I'd also create an example image for a test suite.

src/ImageSharp/Formats/Jpeg/Components/Decoder/HuffmanTable.cs
```diff
@@ -94,6 +94,17 @@ public HuffmanTable(ReadOnlySpan<byte> codeLengths, ReadOnlySpan<byte> values)
...
 
+                    // 'code' is now 1 more than the last code used for codelength 'si'
+                    // in the valid worst possible case 'code' would have the least
+                    // significant bit set to 1, e.g. 1111(0) +1 => 1111(1)
+                    // but it must still fit in 'si' bits since no huffman code can be equal to all 1s
+                    // if last code is all ones, e.g. 1111(1), then incrementing it by 1 would yield
+                    // a new code which occupies one extra bit, e.g. 1111(1) +1 => (1)1111(0)
+                    if (code >= (1 << si))
+                    {
+                        JpegThrowHelper.ThrowInvalidImageContentException("Bad huffman table.");
```
br3aker: @JimBobSquarePants do need a more specific exception message for this?

JimBobSquarePants: No, that should be fine

br3aker:
High hopes it would pass all the tests, it is ready to be reviewed:

- Rewritten part of the huffman table building routine, it now requires 256 bytes less memory workspace
- No `stackalloc` calls
- Rewritten DHT marker parsing code, now it fetches a big single buffer for binary parsing and table building routine
- Added sanity checks for huffman table input, now it should be impossible to break decoder with malformed DHT marker

There's no need for a new test image, image from [#839](https://github.com/SixLabors/ImageSharp/issues/839) has malformed DHT marker which goes out-of-bounds during huffman table for unsafe pointers (which is unspottable without wrapping them with `Span`). I've left a note just in case but this PR should nullify this error.

*this is my favourite*
On top of a critical security breach fix and extra checks we are not  actually slower:

```
MASTER
// Decoding
// Elapsed: 4643ms across 400 iterations
// Average: 11,6075ms

PR
// Decoding
// Elapsed: 4634ms across 400 iterations
// Average: 11,585ms
```

JimBobSquarePants (commented):
Faster... Fantastic 🤣

This is very, very good as always. Thanks! 👍

JimBobSquarePants (approved):

PR #1929, opened 2022-01-08
Change the constructor of PixelTypeInfo and ImageMetadata from internal to public so that a user can implement IImageInfo

jz5:
### Prerequisites

### Description

Case a user implements `IImageInfo` and supports user-specific image formats:
`IImageInfo` has `PixelTypeInfo` and `ImageMetadata` properties. But
* `ImageMetadata` cannot be instanced because it has an internal constructor.
* `PixelTypeInfo` cannot be instanced because it has an internal constructor.

So change the constructor of `PixelTypeInfo` and `ImageMetadata` from internal to public

Related discussion: https://github.com/SixLabors/ImageSharp/discussions/1914

### DIFF
--- a/src/ImageSharp/Formats/PixelTypeInfo.cs
+++ b/src/ImageSharp/Formats/PixelTypeInfo.cs
@@ -16,7 +16,7 @@ namespace SixLabors.ImageSharp.Formats
         /// </summary>
         /// <param name="bitsPerPixel">Color depth, in number of bits per pixel.</param>
         /// <param name="alpha">Tthe pixel alpha transparency behavior.</param>
-        internal PixelTypeInfo(int bitsPerPixel, PixelAlphaRepresentation? alpha = null)
+        public PixelTypeInfo(int bitsPerPixel, PixelAlphaRepresentation? alpha = null)
         {
             this.BitsPerPixel = bitsPerPixel;
             this.AlphaRepresentation = alpha;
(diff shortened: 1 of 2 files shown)


src/ImageSharp/Formats/PixelTypeInfo.cs
```diff
@@ -16,7 +16,7 @@ public class PixelTypeInfo
         /// </summary>
         /// <param name="bitsPerPixel">Color depth, in number of bits per pixel.</param>
         /// <param name="alpha">Tthe pixel alpha transparency behavior.</param>
-        internal PixelTypeInfo(int bitsPerPixel, PixelAlphaRepresentation? alpha = null)
+        public PixelTypeInfo(int bitsPerPixel, PixelAlphaRepresentation? alpha = null)
```
JimBobSquarePants, line 19: @antonfirsov I'm wondering whether we should avoid the nullable parameter here and either provide an overload or break things for V2.

I'm not overly concerned about the implications of making this public because I'd like to do a thorough review for V3 to make sure we capture everything we should.

antonfirsov: I think this should be refactored into 2 constructors for now. In long term, I would explore [init-only setters](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/init) for such API-s.

antonfirsov: As for the V3 API concern: I think minor breaking changes are fine in advanced API-s.

JimBobSquarePants: Yeah, that keeps with our pattern for avoiding nullable parameters when possible. @jz5 Could you please update the `PixelTypeInfo` constructor to use separate overloads with and without the `PixelAlphaRepresentation` (not nullable) parameter. Thanks!

jz5: I've updated it. How about this?

antonfirsov (approved):
LGTM, thanks!

Issue #1919, opened 2022-01-02
Exif metadata for "User Comment" maybe encoded incorrectly?


danpetitt:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I believe, when writing out a "User Comment" Exif tag, the tag is an 'undefined' type which means, according to the spec, that the encoding type should be written out in the first 8 bytes of the value data.

This is seen if you set the metadata and use https://www.metadata2go.com/ to verify the data, it will produce a warning for the "User Comment" tag saying "Invalid EXIF text encoding for UserComment".

If you then edit the metadata on the site and download you will get the additional 8 bytes written before the tag text area as defined in the spec and it then validates okay.

Refer to the specification at https://www.cipa.jp/std/documents/e/DC-008-2012_E.pdf on page 46 for user comment.

I couldnt see anywhere where it states this is optional; but I may be misreading things here so apologies.

On another site (https://www.exiv2.org/tags.html) it states that this may be applicable to other tags which are also of the 'undefined' or 'comment' type:
> The string encoding is defined using the same scheme as UserComment
* GPSProcessingMethod
* GPSAreaInformation

### Steps to Reproduce
```c#
using( var image = Image.Load( "metadata.jpg" ) )
{
  var exif = new ExifProfile();
  exif.SetValue<byte[]>( ExifTag.XPAuthor, Encoding.GetEncoding( "UCS-2" ).GetBytes( "Dan Petitt" ) );
  exif.SetValue<byte[]>( ExifTag.XPTitle, Encoding.GetEncoding( "UCS-2" ).GetBytes( "A bit of test metadata for image title" ) );
  exif.SetValue<byte[]>( ExifTag.UserComment, Encoding.ASCII.GetBytes( "A bit of normal comment text" ) ); // Possibly needs to a parameter to specify what encoding is being used to write the bytes out perhaps?
  image.Metadata.ExifProfile = exif;

  image.Save( "metadata-exif.jpg" );
}
```

`metadata.jpg` shows the file written out by ImageSharp:
[...]

PR #1934, opened 2022-01-10
Use a scheduled job for codecov

JimBobSquarePants:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

Speed up our CI times by moving code coverage checks to a scheduled workflow. Tue,Thur at 2AM

Note: I've done the bare minimum required to enable this. I didn't want to edit `*.ps1` files.

https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows

PR #1935, opened 2022-01-10
[EXIF] Support UCS2 and 8-byte encoded string tags

IldarKhayrutdinov:
### Prerequisites

### Description

Fixes for https://github.com/SixLabors/ImageSharp/issues/1919

* Support reading and writing `encoded string` tags:
  * [UserComment](https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/exif/usercomment.html)
  * [GPSProcessingMethod](https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/gps/gpsprocessingmethod.html)  
  * [GPSAreaInformation](https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/gps/gpsareainformation.html)
* Support reading and writing `UCS2` encoded string tags:

<html>
<body>

0x9c9b | 40091 | Image | Exif.Image.XPTitle | Byte | Title tag used by Windows, encoded in UCS2
-- | -- | -- | -- | -- | --
0x9c9c | 40092 | Image | Exif.Image.XPComment | Byte | Comment tag used by Windows, encoded in UCS2
0x9c9d | 40093 | Image | Exif.Image.XPAuthor | Byte | Author tag used by Windows, encoded in UCS2
0x9c9e | 40094 | Image | Exif.Image.XPKeywords | Byte | Keywords tag used by Windows, encoded in UCS2
0x9c9f | 40095 | Image | Exif.Image.XPSubject | Byte | Subject tag used by Windows, encoded in UCS2

</body>
</html>

See: https://www.exiv2.org/tags.html
 

#### Todo
- [x] Review API
- [x] Add tests
- [x] Review UCS-2 encoded tags


src/ImageSharp/Metadata/Profiles/Exif/ExifConstants.cs
```diff
@@ -2,11 +2,25 @@
...
 namespace SixLabors.ImageSharp.Metadata.Profiles.Exif
 {
     internal static class ExifConstants
     {
+        private const ulong AsciiCode = 0x_41_53_43_49_49_00_00_00;
+        private const ulong JISCode = 0x_4A_49_53_00_00_00_00_00;
+        private const ulong UnicodeCode = 0x_55_4E_49_43_4F_44_45_00;
+        private const ulong UndefinedCode = 0x_00_00_00_00_00_00_00_00;
+
+        private static readonly byte[] AsciiCodeBytes = { 0x41, 0x53, 0x43, 0x49, 0x49, 0, 0, 0 };
```
gfoidl, line 17:
```suggestion
        // This uses C# compiler's optimization to refer to the data segment directly. No allocation occurs.
        private static ReadOnlySpan<byte> AsciiCodeBytes => new byte[] { 0x41, 0x53, 0x43, 0x49, 0x49, 0, 0, 0 };
```
With the `byte[]` the runtime needs to allocate an array on the heap. With the ROS<byte> the C# compiler will do a trick, and refer to the assembly's static data segment without any further allocation.

This is already used in other places in this project (e.g. L24 below), so should be used here too.

src/ImageSharp/Metadata/Profiles/Exif/ExifConstants.cs
```diff
@@ -22,5 +36,55 @@ internal static class ExifConstants
             0x00,
             0x2A
         };
+
+        public static Encoding DefaultAsciiEncoding => Encoding.UTF8;
```
gfoidl, line 40:
```suggestion
        public static Encoding DefaultAsciiEncoding => Encoding.Ascii;
```
? Or adjust the property name.

tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.Metadata.cs
```diff
@@ -288,5 +289,35 @@ public void IgnoreMetadata_ControlsWhetherMetadataIsParsed(bool ignoreMetadata)
...
+        {
+            byte[] exifBytes;
+            using var memoryStream = new MemoryStream();
+            using (var image = Image.Load(TestFile.GetInputFileFullPath(TestImages.Jpeg.Baseline.Calliphora)))
+            {
+                var exif = new ExifProfile();
+                exif.SetValue(ExifTag.XPAuthor, Encoding.GetEncoding("UCS-2").GetBytes("Dan Petitt"));
+
+                exif.SetValue(ExifTag.XPTitle, Encoding.GetEncoding("UCS-2").GetBytes("A bit of test metadata for image title"));
+                exif.SetValue(ExifTag.UserComment, new EncodedString("A bit of normal comment text", EncodedStringCode.ASCII));
```
IldarKhayrutdinov, line 304: @JimBobSquarePants This is API example for `UserComment` tag. 
Doc: https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/exif/usercomment.html

src/ImageSharp/Metadata/Profiles/Exif/ExifConstants.cs
JimBobSquarePants: >but in fact the code use UTF

What code, ours or a reference implementation?

IldarKhayrutdinov: I think, UTF8 is better than ASCII:  
https://stackoverflow.com/questions/19284205/safe-to-use-utf8-decoding-for-exif-property-marked-as-ascii/19284524#19284524

JimBobSquarePants: Thanks for the info, yeah, the spec let's everyone down there. Shall we rename the constant then. `DefaultEncoding` and add a comment stating why we use UTF8.

tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.Metadata.cs
JimBobSquarePants: Yeah I like it. 👍 I'd swap the arguments in the constructor though.

src/ImageSharp/Metadata/Profiles/Exif/Values/EncodedString.cs
```diff
@@ -13,45 +13,67 @@ namespace SixLabors.ImageSharp.Metadata.Profiles.Exif
...
             this.Text = text;
             this.Code = code;
         }
 
         /// <summary>
-        /// Gets the text.
+        /// The 8-byte character code enum.
         /// </summary>
-        public string Text { get; }
+        public enum CharacterCode
```
JimBobSquarePants, line 36: I like this name better 👍

IldarKhayrutdinov:
@JimBobSquarePants  Codecov is not working

JimBobSquarePants:
>@JimBobSquarePants Codecov is not working

Yeah I turned it off for PRs in https://github.com/SixLabors/ImageSharp/pull/1934 because the build/test times were prohibitively expensive. It now runs on a schedule

tests/ImageSharp.Tests/Metadata/Profiles/Exif/Values/ExifValuesTests.cs
```diff
@@ -593,5 +597,19 @@ public void ExifUndefinedArrayTests(ExifTag tag)
             var typed = (ExifByteArray)value;
             Assert.Equal(expected, typed.Value);
         }
+
+        [Theory]
+        [MemberData(nameof(EncodedStringTags))]
+        public void ExifEncodedStringTests(ExifTag tag)
+        {
+            var expected = new EncodedString(EncodedString.CharacterCode.JIS, "test string");
```
JimBobSquarePants, line 605: You could enumerate through the different `CharacterCode` values here and assert each one.

JimBobSquarePants:
@IldarKhayrutdinov Is there anything else you want to add to this? It's looking great so far!

IldarKhayrutdinov:
@JimBobSquarePants Yes, I locally added tests and found some bugs. Nothing major, minor tweaks needed.

JimBobSquarePants:
@IldarKhayrutdinov Hopefully my recent push hasn't caused any conflicts for you. Was just trying to help out.

IldarKhayrutdinov:
@JimBobSquarePants It's ok, there should be no problems. I'll hurry to complete this PR.

src/ImageSharp/ImageSharp.csproj
```diff
@@ -49,6 +49,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="5.0.0" />
+    <PackageReference Include="System.Text.Encoding.CodePages" Version="5.0.0" />
```
IldarKhayrutdinov, line 52: had to add this reference, for `JIS` encoding

src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
```diff
@@ -25,7 +25,13 @@ internal static class ExifEncodedStringHelpers
 
         private static ReadOnlySpan<byte> UndefinedCodeBytes => new byte[] { 0, 0, 0, 0, 0, 0, 0, 0 };
 
-        private static Encoding JIS0208Encoding => Encoding.GetEncoding(932);
+        private static Encoding JIS0208Encoding => CodePagesEncodingProvider.Instance.GetEncoding(932);
```
JimBobSquarePants, line 28: [Docs](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding?view=netframework-4.7.2) seem to say the original should work. What was wrong?

IldarKhayrutdinov: It doesn't work for me locally without `System.Text.Encoding.CodePages`. I don't know maybe it's my machine problem.

JimBobSquarePants: Ah, I completely misread the page there. Yeah, looks like you definitely need the additional package.

IldarKhayrutdinov:
Add new test. The JIS write and read tests pass successfully, but for some reason ExifTool cannot correctly display the JIS encoded text.

`ExifTool` output:

```
ExifTool Version Number         : 12.39
File Name                       : Calliphora_encoded_strings.jpg
Directory                       : .
File Size                       : 252 KiB
File Modification Date/Time     : 2022:01:30 17:50:40+03:00
File Access Date/Time           : 2022:01:30 17:58:59+03:00
File Creation Date/Time         : 2022:01:06 11:42:08+03:00
File Permissions                : -rw-rw-rw-
File Type                       : JPEG
File Type Extension             : jpg
MIME Type                       : image/jpeg
JFIF Version                    : 1.01
Resolution Unit                 : inches
X Resolution                    : 96
Y Resolution                    : 96
Exif Byte Order                 : Little-endian (Intel, II)
XP Title                        : A bit of test metadata for image title
XP Comment                      : A bit of test metadata for image comment
XP Author                       : Dan Petitt
XP Keywords                     : Keyword1;Keyword2
XP Subject                      : This is a subject
User Comment                    : 箖望춂낐嚓좂沸喇
GPS Date Stamp                  : 2022:01:06
GPS Processing Method           : GPS processing method (ASCII)
GPS Area Information            : GPS area info (Unicode)
Image Width                     : 804
Image Height                    : 1198
Encoding Process                : Baseline DCT, Huffman coding
Bits Per Sample                 : 8
Color Components                : 3
Y Cb Cr Sub Sampling            : YCbCr4:4:4 (1 1)
Image Size                      : 804x1198
Megapixels                      : 0.963

```

IldarKhayrutdinov:
I change JIS encoding, though `Shift JIS` [standardized as JIS X 0208 Appendix 1](https://en.wikipedia.org/wiki/Shift_JIS) and popular encoding.  By [docs](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding?view=net-6.0) for `JIS 0208-1990`  needs to use 20932 code.

src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
```diff
@@ -61,17 +67,17 @@ public static bool TryCreate(ReadOnlySpan<byte> buffer, out EncodedString encode
         public static uint GetDataLength(EncodedString encodedString) =>
             (uint)GetEncoding(encodedString.Code).GetByteCount(encodedString.Text) + CharacterCodeBytesLength;
 
-        public static bool IsEncodedString(ExifTagValue tag)
+        public static byte[] GetData(EncodedString encodedString)
```
JimBobSquarePants, line 72: This is used only by the writer yeah? If so I would return an `IDisposable` containing a pooled buffer which you could use to avoid the buffer allocation.

IldarKhayrutdinov: directly writing to output buffer 5d50298

JimBobSquarePants: Ace, thanks! I'll review this today.

JimBobSquarePants (approved):
This all looks excellent. Thanks so much for helping out with the fix! 👍 

I had a good read of the reference material you provided regard the JIS encoding and the one you have chosen does match the specification. I've unlocked the spec PDF to make this more easily searchable.

PR #1936, opened 2022-01-11
ImageFormatDetector Configuration Cleanup

### LINKED ISSUE #1850: Do we need public setter for Configuration.ImageFormatManager?
Currently it's not guarded against setting it to `null`:

https://github.com/SixLabors/ImageSharp/blob/e11f318144d91bac6ed84c6b08e81c6d8850f292/src/ImageSharp/Configuration.cs#L114-L117

I'm wondering if we need the public setter at all? No test cases exercise it, every call site seems to modify the existing collection.

JimBobSquarePants:
### Prerequisites

### Description

Updates around `ImageFormatDetector`. Fixes #1850 

Minimum changes. 
- Removes the public setter from the `Configuration` type
- Minor style cleanup.

This still leaves it possible to clear the collection of detectors, therefore limiting the collection of available image formats.

### DIFF
--- a/src/ImageSharp/Configuration.cs
+++ b/src/ImageSharp/Configuration.cs
@@ -122,25 +122,25 @@ namespace SixLabors.ImageSharp
         public ReadOrigin ReadOrigin { get; set; } = ReadOrigin.Current;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


antonfirsov (approved):
LGTM

Issue #1942, opened 2022-01-13
Overflow exception related to IptcProfile


eelke-at-bolt:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Loading the attached image causes en OverflowException
```
System.OverflowException: Arithmetic operation resulted in an overflow.
   at SixLabors.ImageSharp.Metadata.Profiles.Iptc.IptcProfile.Initialize()
   at SixLabors.ImageSharp.Metadata.Profiles.Iptc.IptcProfile..ctor(Byte[] data)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.InitIptcProfile()
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.Identify(BufferedReadStream stream, CancellationToken cancellationToken)
   at SixLabors.ImageSharp.Formats.ImageDecoderUtilities.Identify(IImageDecoderInternals decoder, Configuration configuration, Stream stream)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoder.Identify(Configuration configuration, Stream stream)
   at SixLabors.ImageSharp.Image.InternalIdentity(Stream stream, Configuration config)
   at SixLabors.ImageSharp.Image.<>c__DisplayClass111_0.<Identify>b__0(Stream s)
   at SixLabors.ImageSharp.Image.WithSeekableStream[T](Configuration configuration, Stream stream, Func`2 action)
   at SixLabors.ImageSharp.Image.Identify(Configuration configuration, Stream stream, IImageFormat& format)
   at SixLabors.ImageSharp.Image.Identify(Configuration configuration, String filePath, IImageFormat& format)
   at SixLabors.ImageSharp.Image.Identify(String filePath, IImageFormat& format)
   at SixLabors.ImageSharp.Image.Identify(String filePath)
   at imageTest.Program.Main(String[] args) in C:\Users\nobody\source\repos\imageTest\Program.cs:line 14
```

### Steps to Reproduce

Load the attached file with `SixLabors.ImageSharp.Image.Load(filename);`

[...]

PR #1944, opened 2022-01-17
Fix for issue #1942

brianpopow:
### Prerequisites

### Description

The real cause of issue #1942 is, the provided testimage seems to have a invalid IPTC profile. The second IPTC value does not start with the expected 0x1c tag marker, its 0 instead.

This PR will skips this tag and ignores IPCT's Extended DataSet feature tags.


src/ImageSharp/Metadata/Profiles/IPTC/IptcProfile.cs
```diff
@@ -264,34 +268,34 @@ private void Initialize()
...
             }
 
-            int i = 0;
-            while (i + 4 < this.Data.Length)
+            int offset = 0;
+            while (offset + 4 < this.Data.Length)
             {
-                if (this.Data[i++] != 28)
+                bool isValidTagMarker = this.Data[offset++] == IptcTagMarkerByte;
+                byte recordType = this.Data[offset++];
```
JimBobSquarePants, line 280: I see `recordType` is unused but I'm struggling to correlate this with the spec. Got any pointers.

src/ImageSharp/Metadata/Profiles/IPTC/IptcProfile.cs
```diff
@@ -242,7 +246,7 @@ public void UpdateData()
             int i = 0;
             foreach (IptcValue value in this.Values)
             {
-                this.Data[i++] = 28;
```
JimBobSquarePants, line 245: I wonder whether we should add a little ascii table just indicating what the dataset tags look like?

src/ImageSharp/Metadata/Profiles/IPTC/IptcProfile.cs
brianpopow: section 1.5 iii: 
```
Octet 2 is the binary representation of the record number. Note that the
envelope record number is always 1, and that the application records are
numbered 2 through 6, the pre-object descriptor record is 7, the object record
is 8, and the post-object descriptor record is 9.
```

I was not sure what to do with this info. Maybe it would be wise to consider the value invalid if its not in the range of 1 to 9.

JimBobSquarePants: > Maybe it would be wise to consider the value invalid if its not in the range of 1 to 9.

Probably wise.

src/ImageSharp/Metadata/Profiles/IPTC/IptcProfile.cs
```diff
@@ -264,34 +268,34 @@ private void Initialize()
...
-                i++;
-
-                var tag = (IptcTag)this.Data[i++];
-
-                int count = BinaryPrimitives.ReadInt16BigEndian(this.Data.AsSpan(i, 2));
-                i += 2;
-
-                var iptcData = new byte[count];
-                if ((count > 0) && (i + count <= this.Data.Length))
+                if (isValidTagMarker && byteCount > 0 && (offset + byteCount <= this.Data.Length))
```
gfoidl, line 291:
```suggestion
                if (isValidTagMarker && byteCount > 0 && (offset <= this.Data.Length - byteCount))
```
To prevent potential overflow. I see that this won't happen in practice, but I got conditioned to be paranoid about these cases (to be safe against malicious code, etc.).

src/ImageSharp/Metadata/Profiles/IPTC/IptcProfile.cs
```diff
@@ -264,34 +268,34 @@ private void Initialize()
...
-            if (this.Data == null || this.Data[0] != 0x1c)
+            if (this.Data == null || this.Data[0] != IptcTagMarkerByte)
             {
                 return;
             }
 
-            int i = 0;
-            while (i + 4 < this.Data.Length)
+            int offset = 0;
+            while (offset + 4 < this.Data.Length)
```
gfoidl, line 277:
```suggestion
            while (offset < this.Data.Length - 4)
```
So the "loop variable" `offset` can be kept in a register, avoiding the repeated addition. Or in other words `this.Data.Length - 4` is constant for the loop, thus will be held in a register.

Also see comment about overflow.

JimBobSquarePants (approved):
Lovely stuff @brianpopow thanks for working on this. 

Also thanks @gfoidl for the reviews.

PR #1958, opened 2022-01-25
Jpeg encoder optimization via removing second transpose call from FDCT

### LINKED ISSUE #1778: Jpeg encoder can be further optimized by fusing transpose and zig-zag steps
Current **SIMD** pipeline (which is stable and most likely won't ever be changed) of block compression is this:

```csharp
// FDCT step
Transpose
FDCT8x8_1D
Transpose
FDCT8x8_1D

// quantization & zigzag step
Divide by quantization table and reduce to int16
Apply zig-zag ordering

// huffman compression
Encode_Block
```

Due to the fact that we are doing 2D DCT we can change matrix transpose ordering to this:
```csharp
// FDCT step
FDCT8x8_1D
Transpose
FDCT8x8_1D
Transpose

// quantization & zigzag step
Apply zig-zag ordering
```

Transpose is followed by zig-zag ordering (quantization is omited because it doesn't change anything - we can modify/transpose/zig-zag quantization table how we want). These operations can be fused into single operation reducing time spent on Transpose calls by 50%.

P.S.
I didn't measure exact time spent on blocks transposing so if anyone wants to implement this - benchmark before potentially wasting your time!

~~P.P.S
This can also be applied to the decoder pipeline.~~

br3aker:
### Prerequisites

### Description
Closes #1778 for the encoder part, decoder part was implemented long time ago so it can be closed after this PR.

Basically, this removes second matrix transpose call from FDCT-Quantize-Huffman pipeline in the jpeg encoder. Performance gain is not **that** big but it's still nice:

Current master
```
// Encoding q=98 | color=YCbCrRatio444
// Elapsed: 11934ms across 1000 iterations
// Average: 11,934ms
```

PR, roughly **4% faster**
```
// Encoding q=98 | color=YCbCrRatio444
// Elapsed: 11457ms across 1000 iterations
// Average: 11,457ms
```

Tests were done using [Calliphora.jpg](https://github.com/SixLabors/ImageSharp/blob/master/tests/Images/Input/Jpg/baseline/Calliphora.jpg) from test suite. Quality 98, no chroma subsampling.

This optimization has 'constant' delta independent of output quality so lower qualities gain better performance boost:

Current master
```
// Encoding q=75 | color=YCbCrRatio444
// Elapsed: 9386ms across 1000 iterations
// Average: 9,386ms
```

PR, roughly **5% faster**
```
// Encoding q=75 | color=YCbCrRatio444
// Elapsed: 8900ms across 1000 iterations
// Average: 8,9ms
```

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
@@ -149,7 +149,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
         {
             DebugGuard.IsTrue(Ssse3.IsSupported, "Ssse3 support is required to run this operation!");
 
-            fixed (byte* shuffleVectorsPtr = SseShuffleMasks)
+            fixed (byte* shuffleVectorsPtr = &MemoryMarshal.GetReference(SseShuffleMasks)
             {
                 Vector128<byte> rowA = block.V0.AsByte();
                 Vector128<byte> rowB = block.V1.AsByte();
--- a/src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
@@ -3,6 +3,7 @@
 
 #if SUPPORTS_RUNTIME_INTRINSICS
 using System;
+using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics;
 using System.Runtime.Intrinsics.X86;
 
@@ -149,7 +150,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
         {
             DebugGuard.IsTrue(Ssse3.IsSupported, "Ssse3 support is required to run this operation!");
 
-            fixed (byte* shuffleVectorsPtr = &MemoryMarshal.GetReference(SseShuffleMasks)
+            fixed (byte* shuffleVectorsPtr = &MemoryMarshal.GetReference(SseShuffleMasks))
             {
                 Vector128<byte> rowA = block.V0.AsByte();
                 Vector128<byte> rowB = block.V1.AsByte();
@@ -243,7 +244,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
         {
             DebugGuard.IsTrue(Avx2.IsSupported, "Avx2 support is required to run this operation!");
 
-            fixed (byte* shuffleVectorsPtr = AvxShuffleMasks)
+            fixed (byte* shuffleVectorsPtr = &MemoryMarshal.GetReference(AvxShuffleMasks))
             {
                 Vector256<byte> rowAB = block.V01.AsByte();
                 Vector256<byte> rowCD = block.V23.AsByte();
--- a/src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
@@ -18,120 +18,138 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components
 #pragma warning restore SA1309
 
         /// <summary>
-        /// Gets shuffle vectors for <see cref="ApplyZigZagOrderingSsse3"/>
+        /// Gets shuffle vectors for <see cref="ApplyTransposingZigZagOrderingSsse3"/>
         /// zig zag implementation.
         /// </summary>
         private static ReadOnlySpan<byte> SseShuffleMasks => new byte[]
         {
-            // row0
-            0, 1, 2, 3, _, _, _, _, _, _, 4, 5, 6, 7, _, _,
-            _, _, _, _, 0, 1, _, _, 2, 3, _, _, _, _, 4, 5,
-            _, _, _, _, _, _, 0, 1, _, _, _, _, _, _, _, _,
+#pragma warning disable SA1515
+            /* row0 - A0 B0 A1 A2 B1 C0 D0 C1 */
+            // A
+            0, 1, _, _, 2, 3, 4, 5, _, _, _, _, _, _, _, _,
+            // B
+            _, _, 0, 1, _, _, _, _, 2, 3, _, _, _, _, _, _,
+            // C
+            _, _, _, _, _, _, _, _, _, _, 0, 1, _, _, 2, 3,
 
-            // row1
-            _, _, _, _, _, _, _, _, _, _, _, _, 8, 9, 10, 11,
-            2, 3, _, _, _, _, _, _, 4, 5, _, _, _, _, _, _,
-            _, _, 0, 1, _, _, 2, 3, _, _, _, _, _, _, _, _,
+            /* row1 - B2 A3 A4 B3 C2 D1 E0 F0 */
+            // A
+            _, _, 6, 7, 8, 9, _, _, _, _, _, _, _, _, _, _,
+            // B
+            4, 5, _, _, _, _, 6, 7, _, _, _, _, _, _, _, _,
 
-            // row2
-            _, _, _, _, _, _, 2, 3, _, _, _, _, _, _, 4, 5,
-            _, _, _, _, _, _, _, _, 0, 1, _, _, 2, 3, _, _,
+            /* row2 - E1 D2 C3 B4 A5 A6 B5 C4 */
+            // A
+            _, _, _, _, _, _, _, _, 10, 11, 12, 13,  _,  _, _, _,
+            // B
+            _, _, _, _, _, _, 8, 9,  _,  _,  _,  _, 10, 11, _, _,
+            // C
+            _, _, _, _, 6, 7, _, _,  _,  _,  _,  _,  _,  _, 8, 9,
 
-            // row3
-            _, _, _, _, _, _, 12, 13, 14, 15, _, _, _, _, _, _,
-            _, _, _, _, 10, 11, _, _, _, _, 12, 13, _, _, _, _,
-            _, _, 8, 9, _, _, _, _, _, _, _, _, 10, 11, _, _,
-            6, 7, _, _, _, _, _, _, _, _, _, _, _, _, 8, 9,
+            /* row3 - D3 E2 F1 G0 H0 G1 F2 E3 */
+            // E
+            _, _, 4, 5, _, _, _, _, _, _, _, _, _, _, 6, 7,
+            // F
+            _, _, _, _, 2, 3, _, _, _, _, _, _, 4, 5, _, _,
+            // G
+            _, _, _, _, _, _, 0, 1, _, _, 2, 3, _, _, _, _,
 
-            // row4
-            _, _, 4, 5, _, _, _, _, _, _, _, _, 6, 7, _, _,
-            _, _, _, _, 2, 3, _, _, _, _, 4, 5, _, _, _, _,
-            _, _, _, _, _, _, 0, 1, 2, 3, _, _, _, _, _, _,
+            /* row4 - D4 C5 B6 A7 B7 C6 D5 E4 */
+            // B
+            _, _,  _,  _, 12, 13, _, _, 14, 15,  _,  _,  _,  _, _, _,
+            // C
+            _, _, 10, 11,  _,  _, _, _,  _,  _, 12, 13,  _,  _, _, _,
+            // D
+            8, 9,  _,  _,  _,  _, _, _,  _,  _,  _,  _, 10, 11, _, _,
 
-            // row5
-            _, _, 12, 13, _, _, 14, 15, _, _, _, _, _, _, _, _,
-            10, 11, _, _, _, _, _, _, 12, 13, _, _, _, _, _, _,
+            /* row5 - F3 G2 H1 H2 G3 F4 E5 D6 */
+            // F
+            6, 7, _, _, _, _, _, _, _, _, 8, 9, _, _, _, _,
+            // G
+            _, _, 4, 5, _, _, _, _, 6, 7, _, _, _, _, _, _,
+            // H
+            _, _, _, _, 2, 3, 4, 5, _, _, _, _, _, _, _, _,
 
-            // row6
-            _, _, _, _, _, _, _, _, 12, 13, _, _, 14, 15, _, _,
-            _, _, _, _, _, _, 10, 11, _, _, _, _, _, _, 12, 13,
-            4, 5, 6, 7, _, _, _, _, _, _, _, _, _, _, _, _,
+            /* row6 - C7 D7 E6 F5 G4 H3 H4 G5 */
+            // G
+            _, _, _, _, _, _, _, _, 8, 9, _, _, _, _, 10, 11,
+            // H
+            _, _, _, _, _, _, _, _, _, _, 6, 7, 8, 9,  _,  _,
 
-            // row7
-            10, 11, _, _, _, _, 12, 13, _, _, 14, 15, _, _, _, _,
-            _, _, 8, 9, 10, 11, _, _, _, _, _, _, 12, 13, 14, 15
+            /* row7 - F6 E7 F7 G6 H5 H6 G7 H7 */
+            // F
+            12, 13, _, _, 14, 15,  _,  _,  _,  _,  _,  _,  _,  _, _, _,
+            // G
+            _,  _, _, _,  _,  _, 12, 13,  _,  _,  _,  _, 14, 15, _, _,
+            // H
+            _,  _, _, _,  _,  _,  _,  _, 10, 11, 12, 13,  _,  _, 14, 15,
+#pragma warning restore SA1515
         };
 
         /// <summary>
-        /// Gets shuffle vectors for <see cref="ApplyZigZagOrderingAvx2"/>
+        /// Gets shuffle vectors for <see cref="ApplyTransposingZigZagOrderingAvx2"/>
         /// zig zag implementation.
         /// </summary>
         private static ReadOnlySpan<byte> AvxShuffleMasks => new byte[]
         {
-                // 01_AB/01_EF/23_CD - cross-lane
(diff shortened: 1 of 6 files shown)


src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
```diff
@@ -18,120 +18,138 @@ internal static partial class ZigZag
...
         /// Applies zig zag ordering for given 8x8 matrix using SSE cpu intrinsics.
         /// </summary>
         /// <param name="block">Input matrix.</param>
-        public static unsafe void ApplyZigZagOrderingSsse3(ref Block8x8 block)
+        public static unsafe void ApplyTransposingZigZagOrderingSsse3(ref Block8x8 block)
         {
             DebugGuard.IsTrue(Ssse3.IsSupported, "Ssse3 support is required to run this operation!");
 
-            fixed (byte* maskPtr = SseShuffleMasks)
+            fixed (byte* shuffleVectorsPtr = SseShuffleMasks)
```
gfoidl, line 152:
```suggestion
            fixed (byte* shuffleVectorsPtr = &MemoryMarshal.GetReference(SseShuffleMasks)
```
Here we know that `SseShuffleMasks` isn't empty, to this avoid the `Length == 0`-check done by `GetPinnableReference` (which is generated by Roslyn for the `fixed`-statement as written above).

Note: for .NET 6 codegen is equal, as the JIT got optimizied, but older runtimes can benefit from that change.

gfoidl: Sorry, I missed a `)`.

br3aker: Np, there's 2 such places so I'll manually create a commit, thanks for the review!

src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
```diff
@@ -225,69 +239,61 @@ public static unsafe void ApplyZigZagOrderingSsse3(ref Block8x8 block)
...
-                Vector256<int> rows_AB01_EF01_CD23_shuffleMask = Avx.LoadVector256(shuffleVectorsPtr + (0 * 32)).AsInt32();
-                Vector256<byte> row01_AB = Avx2.PermuteVar8x32(rowsAB.AsInt32(), rows_AB01_EF01_CD23_shuffleMask).AsByte();
+                Vector256<byte> rowAB = block.V01.AsByte();
+                Vector256<byte> rowCD = block.V23.AsByte();
+                Vector256<byte> rowEF = block.V45.AsByte();
+                Vector256<byte> rowGH = block.V67.AsByte();
+
+                /* row01 - A0 B0 A1 A2 B1 C0 D0 C1 | B2 A3 A4 B3 C2 D1 E0 F0 */
+                Vector256<int> crln_01_AB_CD = Avx.LoadVector256(shuffleVectorsPtr + (0 * 32)).AsInt32();
+                Vector256<byte> row01_AB = Avx2.PermuteVar8x32(rowAB.AsInt32(), crln_01_AB_CD).AsByte();
```
antonfirsov, line 255: Any chance we can group Loads and PermuteVars along the implementation (without running out of registers) so they are pipelined? 

See https://github.com/SixLabors/ImageSharp/pull/1242#issuecomment-735346693

br3aker: Cross-lane permutations are not pipelined:
image

Loads are the main bottlenecks here with latency of **7** which can be pipelined:
image

Even though shuffles and permutes don't pipeline well, shuffles can be pipelined on the newer architecture:
image

So I'll try to group them without register spilling as well but I don't have such CPU to test performance.

br3aker: Nah, grouped loads code is a bit slower:

``` ini

.NET Core SDK=6.0.100-preview.3.21202.5
  [Host]     : .NET Core 3.1.21 (CoreCLR 4.700.21.51404, CoreFX 4.700.21.51508), X64 RyuJIT
  DefaultJob : .NET Core 3.1.21 (CoreCLR 4.700.21.51404, CoreFX 4.700.21.51508), X64 RyuJIT


```
|          Method |     Mean |     Error |    StdDev | Code Size |
|---------------- |---------:|----------:|----------:|----------:|
| &#39;Avx optimized&#39; | 5.666 ns | 0.1440 ns | 0.2018 ns |     478 B |
| &#39;Avx pipelined&#39; | 6.076 ns | 0.1516 ns | 0.2175 ns |     501 B |

While loads are grouped at the assembler level, 'pipelined' version saves 6 `xmm` (yes, `xmm`, not `ymm`) registers to the stack while current PR versions saves 4 registers to the stack for whatever reason, thus the slowdown.

br3aker:
@gfoidl sorry for pinging but this may interest you.

AVX implementation has a bit different asm output and performance on `core 3.1` and `.net 6.0`:

**.NET Core 3.1.21** (CoreCLR 4.700.21.51404, CoreFX 4.700.21.51508), X64 RyuJIT
|          Method |     Mean |     Error |    StdDev | Code Size |
|---------------- |---------:|----------:|----------:|----------:|
| &#39;Avx optimized&#39; | 5.666 ns | 0.1440 ns | 0.2018 ns |     478 B |

**.NET Core 6.0.0** (CoreCLR 6.0.21.20104, CoreFX 6.0.21.20104), X64 RyuJIT
|          Method |     Mean |     Error |    StdDev | Code Size |
|---------------- |---------:|----------:|----------:|----------:|
| &#39;Avx optimized&#39; | 7.009 ns | 0.1292 ns | 0.1209 ns |     588 B |

I'm not even close to understand assembly nuances but `.net 6.0` does mask loads a bit different than `core 3.1`. Take a look at first 2 rows calculation:

`core 3.1`
```asm
sub       rsp,58
vzeroupper
vmovaps   [rsp+40],xmm6
vmovaps   [rsp+30],xmm7
vmovaps   [rsp+20],xmm8
vmovaps   [rsp+10],xmm9
xor       eax,eax
mov       [rsp+8],rax
mov       rax,26ECE25CA20
mov       [rsp+8],rax
vmovupd   ymm0,[rcx]
vmovupd   ymm1,[rcx+20]
vmovupd   ymm2,[rcx+40]
vmovupd   ymm3,[rcx+60]
vmovdqu   ymm4,ymmword ptr [rax]
vpermd    ymm5,ymm4,ymm1
vpermd    ymm4,ymm4,ymm0
vmovdqu   ymm6,ymmword ptr [rax+20]
vpshufb   ymm4,ymm4,ymm6
vmovdqu   ymm6,ymmword ptr [rax+40]
vpshufb   ymm5,ymm5,ymm6
vmovdqu   ymm6,ymmword ptr [rax+60]
vpermd    ymm7,ymm6,ymm2
vmovdqu   ymm8,ymmword ptr [rax+80]
vpshufb   ymm8,ymm7,ymm8
vpor      ymm5,ymm5,ymm8
vpor      ymm4,ymm4,ymm5
...
```

`.net 6.0`
```asm
sub       rsp,58
vzeroupper
vmovaps   [rsp+40],xmm6
vmovaps   [rsp+30],xmm7
vmovaps   [rsp+20],xmm8
vmovaps   [rsp+10],xmm9
mov       rax,1CDF57ECA60
mov       [rsp+8],rax
vmovupd   ymm0,[rcx]
vmovupd   ymm1,[rcx+20]
vmovupd   ymm2,[rcx+40]
vmovupd   ymm3,[rcx+60]
vmovdqu   ymm4,ymmword ptr [rax]
vpermd    ymm5,ymm4,ymm1
vpermd    ymm4,ymm4,ymm0
mov       rax,1CDF57ECA80
vmovdqu   ymm6,ymmword ptr [rax]
vpshufb   ymm4,ymm4,ymm6
mov       rax,1CDF57ECAA0
vmovdqu   ymm6,ymmword ptr [rax]
vpshufb   ymm5,ymm5,ymm6
mov       rax,1CDF57ECAC0
vmovdqu   ymm6,ymmword ptr [rax]
vpermd    ymm7,ymm6,ymm2
mov       rax,1CDF57ECAE0
vmovdqu   ymm8,ymmword ptr [rax]
vpshufb   ymm8,ymm7,ymm8
vpor      ymm5,ymm5,ymm8
vpor      ymm4,ymm4,ymm5
...
```

Aren't those multiple `mov rax addr` slower than `[rax + const]`?

antonfirsov:
The test failure is result of eager merge of #1960. I will revert bbbf687 later.

src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
antonfirsov: > Cross-lane permutations are not pipelined:

To me Latency = 3, Througput = 1 means that you can pipeline 3 instructions so the overall latency will be 3 cycles instead of 3 * 3 cycles, which is already a win.

> While loads are grouped at the assembler level, 'pipelined' version saves 6 `xmm` (yes, `xmm`, not `ymm`) registers to the stack while current PR version saves 4 registers to the stack for whatever reason, thus the slowdown.

That's super weird, especially the xmm part.

Anyways, this was just an idea, the improvement in the PR is already good enough.

gfoidl:
> Aren't those multiple `mov rax addr` slower than using `[rax + const]` directly?

IMO `[rax + const]` should be prefered (as you expect too). Code size is also quite smaller in .NET Core 3.1. (by 110 bytes). So it looks like a regression in .NET 6.
When I look with llvm-mca at the posted assembly, the .NET Core 3.1 version has about 400 µOps less to execute -- this reflects in the shown perf-numbers too.

Can we get a minimal repro so that an issue in dotnet/runtime can be filed?

(Don't hesitate to ping me, but in the next days I'm a bit slow to respond (out of office)).

br3aker:
> Can we get a minimal repro so that an issue in dotnet/runtime can be filed?

I'll play around with it and ping with results (tomorrow, probably), thanks!

src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
br3aker: It's the opposite: latency is how many 'cycles' single independent op takes, throughput is how many instructions can be started in one 'cycle', it's a reciprocal value though.

Let's take unaligned loads for example, latency of 7 and thoughput 0.5 on newer architecture, let's say we have this code:
```asm
...
vmovupd   ymm0,[rcx]
vmovupd   ymm1,[rcx+20]
vmovupd   ymm2,[rcx+40]
vmovupd   ymm3,[rcx+60]
...
```

Due to the fancy cpu out-of-order first two `vmovupd` loads can be issued simultaneously because they are independent. In the best case possible code above would be executed in 8 cycles:

```
|1 cycle|2 cycle|3 cycle|4 cycle|5 cycle|6 cycle|7 cycle|8 cycle    |
|vmovupd                                                |use results|
|vmovupd                                                |use results|
        |vmovupd                                                    |use results|
        |vmovupd                                                    |use results|
```

This is a very rough 'view' of what could happen because all these `vmovupd` must have been decoded and present in instruction queue, there should be no other port competing instructions there and other complicated stuff. But general idea is this.

saucecontrol: I think you have pipelining confused with parallel dispatch.

A pipelined instruction is one that takes more than one cycle to complete but does not tie up the execution unit completely during that time, so instructions can overlap execution within that latency window.  You've correctly illustrated that for `vmovupd` above, where both pipelining and parallel dispatch apply, because 2 ports can accept that instruction simultaneously and up to 7 instructions can be in various stages of completion within the EU's pipeline.

In the case of `vpermd`, the latency is 3, meaning it takes 3 cycles from dispatch to completion, and its reciprocal throughput is 1, meaning an instruction can be dispatched and retired every cycle (this one can only be dispatched on a single port).  The total execution time for 2 `vpermd`s would be 4 cycles, with both results available on the 5th.

```
|1 cycle|2 cycle|3 cycle|4 cycle    |5 cycle    |
|vpermd                 |use results|
        |vpermd                     |use results|
```

So there is potential benefit in unrolling those.

saucecontrol:
Here's a simple repro for that codegen issue, along with "one weird trick" to work around it 😂

[Sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4a9Jq049+ggJI8ZEMfKhKV6zY2btufASyMYovLrmXXatnQ/3OPG4eXiwAGgAcSBrUmgDM2kgM3LjYAGYwTKQMQgwA3jQMRQxibgrYGJnMSWww2AAmAPJckgCesmLYXAA8wK2VAHwMfZW4DAC8Q1wwAO7D/TAA2gC6+YXFG3RoDOTbpNtx2+gMAKzbSNsA7GjrG0VbO3sHR9tnDBcMl7cMAL4xG99iAlqgx3BgGABZAAUAEo1tQ7sURjAAFQlVwTBhQ5EomEAVQ86UEAEFcAAFCBgmRQ2BpSEwfDQVoQ7BQXAAC2wkhYAHEYBhahlYFwVNiFrgYTD/oiGN8NsRLgxZLgYKQWMT6vUoXKZcrVSwADIQBoANRgYAw0HIpAiULErhh6twLjipFhNwRMo2erVRtN5stUGttvtUAYAGoGGaLVabb0FgNhBAHI7SS63TCdUVHQAVCCyMBc1mw6VFH40QHA8hJMEMABysPhXtp80qwzD4wYLYJqQy6s1NJgdIhDKZLLZnO5fIFQ5kMBFMDFoxh2zoUqzss9iIVSpVao1Wo3Gx7RKd8cq22jgeD3TBAwGg7pwCgK6PxRPfdJ55gl4DsYit48Pej4MB+JIDi2z6/jGQZxsiiYiCmmZbncub5oWkjFuuKHltQPxAA===)

br3aker:
Got a bit ill so couldn't work these 2 days.

> "one weird trick" to work around it

I've actually tried to do entire thing with 'ref pointers' and got worse performance, didn't check assembly though.

saucecontrol:
The codegen is extremely finicky.  Note the `Unsafe.Add` with `0` offset in that example.  Take that out, and it regresses again.  We're discussing that over in the #lowlevel channel on csharp discord.  You may want to get in on that https://discord.gg/csharp

Hope you feel better!

src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
br3aker: I actually did, thanks for clarifying!
But as I said, grouped loads implementation is slower. Maybe grouped permutations and shuffles would yield better performance with enough manual tweaking and looking bad at the binary but with 'naive' grouping them  in code yields worse/same results:

|                                           Method |     Mean |     Error |    StdDev | Code Size |
|------------------------------------------------- |---------:|----------:|----------:|----------:|
|                                  'Avx detailed masks' | 9.309 ns | 0.0382 ns | 0.0319 ns |     569 B |
|                                  'Avx optimized' | 5.965 ns | 0.0696 ns | 0.0617 ns |     478 B |
|                            'Avx pipelined loads' | 6.095 ns | 0.0346 ns | 0.0307 ns |     501 B |
|                    'Avx pipelined the right way' | 6.424 ns | 0.0314 ns | 0.0278 ns |     488 B |

Legend:
Detailed masks - first implementation with detailed permutation/shuffle masks
Optimized - reduced permutation/shuffle masks table
Pipelined loads - grouped mask loads
The right way - grouped everything manually to get the best codegen from the 'pipeline loads' implementation

For example, here's the codegen for the first 2 rows calculation:

```asm
...
vmovupd   ymm0,[rcx]
vmovupd   ymm1,[rcx+20]
vmovupd   ymm2,[rcx+40]
vmovdqu   ymm3,ymmword ptr [rax]
vmovdqu   ymm4,ymmword ptr [rax+20]
vmovdqu   ymm5,ymmword ptr [rax+40]
vmovdqu   ymm6,ymmword ptr [rax+60]
vmovdqu   ymm7,ymmword ptr [rax+80]
vpermd    ymm8,ymm3,ymm0
vpermd    ymm3,ymm3,ymm1
vpermd    ymm9,ymm6,ymm2
vpshufb   ymm8,ymm8,ymm4
vpshufb   ymm3,ymm3,ymm5
vpshufb   ymm4,ymm9,ymm7
vpor      ymm3,ymm8,ymm3
vpor      ymm3,ymm3,ymm4
vmovupd   [rdx],ymm3
...
```

But yet this produces same performance (if not worse, benchmark is a bit unstable for whatever reason). Maybe I'm doing something wrong, I can provide full code for all of those benchmark versions if you want to play with this code but I'm honestly okay with current performance :)

saucecontrol: Yep, that codegen looks good.  I'd be happy to look at your variants if you want to share them, but if you're already happy with the perf, I'm in no position to argue ;)

This is something that may be worth revisiting once netcoreapp3.1 is EOL.  The constant vector and vector CSE support in net6.0 might help.  It's difficult to get optimal code on both runtimes at present.

JimBobSquarePants: >I'm honestly okay with current performance

As am I 😄 

This has been a fantastic discussion btw, I've learned a lot!

JimBobSquarePants (approved):

src/ImageSharp/Formats/Jpeg/Components/ZigZag.Intrinsic.cs
br3aker: I'll leave it here just in case you'd want to take a look:
https://gist.github.com/br3aker/4da983868b8915180747ba1f8274862e

PR #1969, opened 2022-02-01
Add basic memory diagnostic API

### LINKED ISSUE #1883: Basic memory diagnostics
Spinning off https://github.com/SixLabors/ImageSharp/issues/1739#issuecomment-984667382 to an issue.

We need expose some basic memory diagnostic utilities from library code:

```C#
namespace SixLabors.ImageSharp.Diagnostics
{
    public static class MemoryTracker
    {
        // Count outstanding MemoryAllocator allocations. Dispose reduces the counter, finalizer doesn't.
        public static long TotalUndisposedLogicalAllocationCount { get; }

        // Make the library throw in finalizers if stuff was left undisposed
        public static bool EnableStrictDisposeWatcher { get; set; }
    }
}
````

We should also introduce standard [event counters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/event-counters), but that's a different story for a separate issue later.

### PR DESCRIPTION
**Pull request by antonfirsov**
### Prerequisites

### Description

Resolves #1883. I went with the following API in the end, feels nicer and simpler:

```C#
public delegate void UndisposedMemoryResourceDelegate(string allocationStackTrace);

public static class MemoryDiagnostics
{
    // Reports the stack trace of leaked allocations. Subscribe for troubleshooting.
    public static event UndisposedMemoryResourceDelegate UndisposedAllocation;

    public static int TotalUndisposedAllocationCount { get; }
}
```


JimBobSquarePants (approved):
Well this is clever!!! 🤲

src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// Fires when an ImageSharp object's undisposed memory resource leaks to the finalizer.
+        /// The event brings significant overhead, and is intended to be used for troubleshooting only.
+        /// For production diagnostics, use <see cref="TotalUndisposedAllocationCount"/>.
+        /// </summary>
+        public static event UndisposedMemoryResourceDelegate UndisposedAllocation
+        {
+            add
+            {
+                Interlocked.Increment(ref undisposedMemoryResourceSubscriptionCounter);
+                undisposedMemoryResource += value;
```
gfoidl, line 34: Instead of `Interlocked.Increment/Decrement` I'd put both operations under one `lock` that's also used in `remove` (the lock object).

`add`/`remove` isn't high perf, and so both operations are protected. Current implementation could suffer from a race.

src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
```diff
@@ -0,0 +1,63 @@
...
+                Interlocked.Decrement(ref undisposedMemoryResourceSubscriptionCounter);
+            }
+        }
+
+        /// <summary>
+        /// Gets a value indicating the total number of memory resource objects leaked to the finalizer.
+        /// </summary>
+        public static int TotalUndisposedAllocationCount => totalUndisposedAllocationCount;
+
+        internal static bool MemoryResourceLeakedSubscribed => undisposedMemoryResourceSubscriptionCounter > 0;
```
gfoidl, line 49:
```suggestion
        internal static bool MemoryResourceLeakedSubscribed => Volatile.Read(ref undisposedMemoryResourceSubscriptionCounter) > 0;
```
to be on the safe side?

x86/x64 should be safe anyway, but other platforms may behave different.

src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
```diff
@@ -0,0 +1,63 @@
...
+        internal static void IncrementTotalUndisposedAllocationCount() =>
+            Interlocked.Increment(ref totalUndisposedAllocationCount);
+
+        internal static void DecrementTotalUndisposedAllocationCount() =>
+            Interlocked.Decrement(ref totalUndisposedAllocationCount);
+
+        internal static void RaiseUndisposedMemoryResource(string allocationStackTrace)
+        {
+            // Schedule on the ThreadPool, to avoid user callback messing up the finalizer thread.
+            ThreadPool.QueueUserWorkItem(_ => undisposedMemoryResource?.Invoke(allocationStackTrace));
```
gfoidl, line 60: Perform the null-check before queuing to thread-pool? So if nothing is registered, the thread-pool has less work to do. Ala
```suggestion
            UndisposedMemoryResourceDelegate tmp = undisposedMemoryResource;
            if (tmp is not null)
            {
                ThreadPool.QueueUserWorkItem(static state =>
                {
                    var (@event, stackTrace) = state;
                    @event(stackTrace);
                }, (tmp, allocationStackTrace), preferLocal: false);
            }
```
This also avoids the closures for the stacktrace and the delegate, thus avoiding allocations*.
Though as noted in the PR's description this API is for diagnostics and not intended for production, so perf doesn't matter as much one could go a simpler route:
```suggestion
            UndisposedMemoryResourceDelegate tmp = undisposedMemoryResource;
            if (tmp is not null)
            {
                ThreadPool.QueueUserWorkItem(stackTrace => tmp(stackTrace), allocationStackTrace, preferLocal: false);
            }
```

Anyway my main point is: cache to local, perform the null-check, and iif not null queue to thread-pool to have minimal impact for production.

---

\* in thread-pool internals there's still an allocation happening, which could be avoided by a type that implements `IThreadPoolWorkItem` which can be re-used. But that's overkill here.

antonfirsov: >  in thread-pool internals there's still an allocation happening, which could be avoided by a type that implements IThreadPoolWorkItem which can be re-used. But that's overkill here.

Yeah I considered that, but definitely not worth it. Users should not subscribe to this delegate in production anyways, it takes a snapshot of `Environment.StackTrace` on *every* buffer allocation.

antonfirsov:
@gfoidl as always, thanks for the remarks!

src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
```diff
@@ -56,8 +63,16 @@ public static class MemoryDiagnostics
...
+            {
+                return;
+            }
+
             // Schedule on the ThreadPool, to avoid user callback messing up the finalizer thread.
-            ThreadPool.QueueUserWorkItem(_ => undisposedMemoryResource?.Invoke(allocationStackTrace));
+            ThreadPool.QueueUserWorkItem(
+                stackTrace => undisposedAllocation?.Invoke(stackTrace),
+                allocationStackTrace,
+                preferLocal: true);
```
gfoidl, line 75:
```suggestion
                preferLocal: false);
```
I believe it should go to the global thread-pool queue not the local one.

src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
```diff
@@ -69,10 +68,16 @@ internal static void RaiseUndisposedMemoryResource(string allocationStackTrace)
             }
 
             // Schedule on the ThreadPool, to avoid user callback messing up the finalizer thread.
+#if NETSTANDARD2_1 || NETCOREAPP2_1_OR_GREATER
```
JimBobSquarePants, line 71: https://apisof.net/catalog/737fc3e3-23d4-dfab-294f-8c59c58c5852

antonfirsov: Damn, thanks!

JimBobSquarePants: No worries. Thought I'd better wait until you'd seen this before merging. I'll let you do the honors when you're happy.

Issue #1802, opened 2021-11-01
WebP: Missing features


brianpopow:
### Description

This is a tracking issue for missing webp features.
The following features for webp are still missing.

- ~Decode animated webp images.~ done with #1985
- ~Encode animated webp images.
- ~Decode and store XMP profile.~ done with #1918
- ~Encode Alpha channel of Lossy images with the lossless encoder and store it in `ALPH` chunk.~ done in #1971

Any help from the community would be highly appreciated.

JimBobSquarePants:
@bytehydrant Do you still have your fork with the in-progress animation support?

We're in a good place now to continue working against. https://github.com/SixLabors/ImageSharp/issues/121#issuecomment-890502421

bytehydrant:
@JimBobSquarePants, sorry for the (yikes) 3-month delay in response. Apparently, I had notifications turned off and didn't see this 'til just now. Yeah, the fork linked from my original post is still there. I've not touched it since then.

JimBobSquarePants:
No worries  @bytehydrant , Thanks for getting back to me. 

@brianpopow has recently opened a PR for supporting decoding animated Webp in #1985 It's be really useful to us if you could cast an eye over that code and do a review. Thanks!

JimBobSquarePants:
Fixed with #2569

PR #1971, opened 2022-02-01
Add support for encoding lossy webp images with alpha channels

brianpopow:
### Prerequisites

### Description

This PR adds support for encoding lossy webp images with alpha channels.
The alpha data can be either compressed (with webp lossless compression) or uncompressed.

Related to #1802

### DIFF
--- a/src/ImageSharp/Formats/Webp/AlphaEncoder.cs
+++ b/src/ImageSharp/Formats/Webp/AlphaEncoder.cs
@@ -13,8 +13,10 @@ namespace SixLabors.ImageSharp.Formats.Webp
     /// <summary>
     /// Methods for encoding the alpha data of a VP8 image.
     /// </summary>
-    internal static class AlphaEncoder
+    internal class AlphaEncoder : IDisposable
     {
+        private IMemoryOwner<byte> alphaData;
+
         /// <summary>
         /// Encodes the alpha channel data.
         /// Data is either compressed as lossless webp image or uncompressed.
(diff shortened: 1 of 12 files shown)


src/ImageSharp/Formats/Webp/AlphaEncoder.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// <param name="memoryAllocator">The memory manager.</param>
+        /// <param name="compress">Indicates, if the data should be compressed with the lossless webp compression.</param>
+        /// <param name="size">The size in bytes of the alpha data.</param>
+        /// <returns>The encoded alpha data.</returns>
+        public static IMemoryOwner<byte> EncodeAlpha<TPixel>(Image<TPixel> image, Configuration configuration, MemoryAllocator memoryAllocator, bool compress, out int size)
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            int width = image.Width;
+            int height = image.Height;
+            IMemoryOwner<byte> alphaData = ExtractAlphaChannel(image, configuration, memoryAllocator);
```
JimBobSquarePants, line 34: I can't see where this is getting disposed?

brianpopow: Its done in the `Dispose` of the `Vp8Encoder`

brianpopow: Maybe it would be better to keep the responsibility of disposing to the `AlphaEncoder` instead?

JimBobSquarePants: Yeah, I think that's wise. Difficult to keep track of then.

src/ImageSharp/Formats/Webp/Lossy/Vp8Encoder.cs
```diff
@@ -71,10 +71,7 @@ internal class Vp8Encoder : IDisposable
         /// </summary>
         private int uvAlpha;
 
-        /// <summary>
-        /// Scratch buffer to reduce allocations.
-        /// </summary>
-        private readonly int[] scratch = new int[16];
+        private readonly bool alphaCompression;
 
         private readonly byte[] averageBytesPerMb = { 50, 24, 16, 9, 7, 5, 3, 2 };
```
gfoidl, line 76:
```suggestion
        // This uses C#'s optimization to refer to the static data segment of the assembly, no allocation occurs.
        private static ReadOnlySpan<byte> AverageBytesPerMb => new byte[] { 50, 24, 16, 9, 7, 5, 3, 2 };
```
While you're on it.

Need to update https://github.com/SixLabors/ImageSharp/blob/484bd77d8d2e0cca5bdbad6e528337fc21b4af94/src/ImageSharp/Formats/Webp/Lossy/Vp8Encoder.cs#L327 to new name too.

JimBobSquarePants (approved):
LGTM. Thanks for pushing this through so quickly!

PR #1976, opened 2022-02-03
Convert OrientationMode into something usable.

JimBobSquarePants:
### Prerequisites

### Description

I noticed while working on https://github.com/SixLabors/ImageSharp.Web/pull/215 that our `OrientationMode` enum was internal. I'm assuming because explicitly casting to a `ushort` each time was not a great developer experience. By converting it to a static class with constants we're able to expose the type in a much more useful form.


src/ImageSharp/Processing/Processors/Transforms/Linear/AutoOrientProcessor{TPixel}.cs
```diff
@@ -81,32 +81,32 @@ protected override void OnFrameApply(ImageFrame<TPixel> sourceBase)
...
             if (value.DataType == ExifDataType.Short)
             {
-                orientation = (OrientationMode)value.Value;
+                orientation = value.Value;
             }
             else
             {
-                orientation = (OrientationMode)Convert.ToUInt16(value.Value);
+                orientation = Convert.ToUInt16(value.Value);
                 source.Metadata.ExifProfile.RemoveValue(ExifTag.Orientation);
```
ronaldbarendse, line 106: This line should be removed, because the value is always set below:
```suggestion
```

src/ImageSharp/Processing/Processors/Transforms/Linear/AutoOrientProcessor{TPixel}.cs
```diff
@@ -81,32 +81,32 @@ protected override void OnFrameApply(ImageFrame<TPixel> sourceBase)
...
             }
 
-            source.Metadata.ExifProfile.SetValue(ExifTag.Orientation, (ushort)OrientationMode.TopLeft);
+            source.Metadata.ExifProfile.SetValue(ExifTag.Orientation, ExifOrientationMode.TopLeft);
```
ronaldbarendse, line 109: Would you be open to make it possible to opt-in to removing the value, instead of resetting it to `ExifOrientationMode.TopLeft`? This would require a single `bool RemoveExifValue { get; }` property and the following code:

```c#
if (this.RemoveExifValue)
{
    source.Metadata.ExifProfile.RemoveValue(ExifTag.Orientation);
}
else
{
    source.Metadata.ExifProfile.SetValue(ExifTag.Orientation, ExifOrientationMode.TopLeft);
}

antonfirsov (approved):
I wish there was a good way to consume enums in `ExifTag.SetValue()`, but I don't see how would it be possible, so LGTM.

src/ImageSharp/Processing/Processors/Transforms/Linear/AutoOrientProcessor{TPixel}.cs
JimBobSquarePants: No, we'll keep it since the value was explicitly set by the previous encoder.

src/ImageSharp/Processing/Processors/Transforms/Linear/AutoOrientProcessor{TPixel}.cs
JimBobSquarePants: We remove to ensure that any invalid values are gone. We've seen in the wild some odd implementations of this property. `SetValue` actually calls `TrySetValue` on the value which can return false.

Issue #1978, opened 2022-02-05
Some old jpg images are saved with glitches (regression in alpha)


botinko:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Some pictures of our clients get corrupted when resizing and even when simply resaving by ImageSharp. 
Issue exists in `2.0.0-alpha.0.168` but doesn't exists in `v1.0.4`. 

### Steps to Reproduce
```
            var img = Image.Load(@"issue.jpg");
            img.Save("copy.jpg");
```
before
![issue](https://user-images.githubusercontent.com/2981998/152630871-cb8c08c5-17b9-43e9-852c-44a5ed11d56e.jpg)
after
![copy](https://user-images.githubusercontent.com/2981998/152630935-2706ff5f-dfd5-4a33-92d6-641a167857c7.jpg)

Link to original image and image after resaving https://drive.google.com/file/d/1_k54RFuPciKeKBtFDLftIKetOPsWkOFO/view?usp=sharing

If you resave the image in any editor, the problem stops reproducing. So you need an original image from archive to reproduce. 

### System Configuration
- Other ImageSharp packages and versions: 2.0.0-alpha.0.168
- Environment (Operating system, version and so on): Windows 10, docker
- .NET Framework version: dotnet 5, 6
- Additional information: issue doesn't exists in v1.0.4.

br3aker:
Can confirm on my machine. Will work on it asap.

br3aker:
I've found error leading to this. It was broken in https://github.com/SixLabors/ImageSharp/pull/1694 which did spectral -> color conversion withing spectral strides instead of full image.

Everything does work in current tests because we don't have a single component jpeg with non 1x1 sampling factors in current test suite. ITU spec says that any single component jpeg must be decoded as 'non-interleaved' jpeg:

> When Ns = 1 (where Ns is the number of components in a scan), the order of data units within a scan shall be left-to-right
and top-to-bottom, as shown in Figure A.2. This ordering applies whenever Ns = 1, regardless of the values of
H1 and V1.
[...]

PR #1979, opened 2022-02-06
Fixed decoding grayscale jpegs with exotic sampling factors

br3aker:
### Prerequisites

### Description
Fixes https://github.com/SixLabors/ImageSharp/issues/1978. ~~Issue author didn't respond (yet) whether we can use attached image in the test suite so~~ I've created a test image based on one of grayscale images from the test suit:

before

This image after decode-encode on current master:

after


br3aker:
@antonfirsov  we've got a permission to use provided image in test suite, do we need extra image or one is enough?

JimBobSquarePants:
Great work @br3aker ! I don't think we'll need two references if the sampling factors are the same for both.

JimBobSquarePants (commented):
Confirmed the sampling factors so I believe this to be good to go.

JimBobSquarePants (approved):

PR #1981, opened 2022-02-07
Fix documentation of HorizontalResolution and VerticalResolution properties

0xced:
Also improve the formatting of the ResolutionUnits documentation.

### DIFF
--- a/src/ImageSharp/Metadata/ImageMetadata.cs
+++ b/src/ImageSharp/Metadata/ImageMetadata.cs
@@ -72,7 +72,7 @@ namespace SixLabors.ImageSharp.Metadata
 
         /// <summary>
         /// Gets or sets the resolution of the image in x- direction.
-        /// It is defined as the number of dots per inch and should be an positive value.
+        /// It is defined as the number of dots per <see cref="ResolutionUnits"/> and should be an positive value.
         /// </summary>
         /// <value>The density of the image in x- direction.</value>
         public double HorizontalResolution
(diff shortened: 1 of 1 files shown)


JimBobSquarePants (approved):
Ah JUST missed the V2 release.... 

This looks great, much appreciated 👍

0xced:
Too bad I just missed the v2 release!

I noticed this documentation glitch while porting the ClosedXML library from `System.Drawing` to ImageSharp (https://github.com/ClosedXML/ClosedXML/issues/1694#issuecomment-1031590337). Except for this small glitch the migration is going pretty smoothly. It's really a pleasure to work with such a well engineered and well documented library. 😍

JimBobSquarePants:
Thanks! That's really great to hear!

Just had a very quick look at the port. You've got a mix of the old and new major versions of ImageSharp, Drawing, and, Fonts going on there which won't work well. I'm gonna be pushing new betas for both Fonts and Drawing today which will bring some changes (massive improvements) around text rendering. I'd strongly recommend bumping to a minimum of net472 also due the the [redirect issues with older targets](https://nickcraver.com/blog/2020/02/11/binding-redirects/)

0xced:
Actually, Drawing is not used. Nevertheless, I updated Fonts to the latest beta (16) and it was a matter of updating `IFontCollection.Install` to `IFontCollection.Add`, `IReadOnlyFontCollection.Find` to `IReadOnlyFontCollection.Get` and `RenderOptions` to the new `TextOptions`.

> I'd strongly recommend bumping to a minimum of net472 also due the the [redirect issues with older targets](https://nickcraver.com/blog/2020/02/11/binding-redirects/)

I'm fully aware of the issues with .NET Standard 2.0 *theoretically* supporting .NET Framework 4.6.1 but *actually* only supporting .NET Framework 4.7.2. But in such a large pull request, I prefer te be conservative with the changes to the target framework in order to focus on the rest of the pull request.

Issue #1997, opened 2022-02-14
Image.LoadAsync(stream) uses synchronous IO if the stream is seekable 


kroymann:
### ImageSharp version

1.0.4 (verified issue is still present in 2.0.0)

### Other ImageSharp packages and versions

ImageSharp.Web 1.0.5 (+ fixes from PR 222)

### Environment (Operating system, version and so on)

Azure App Service (Windows-based)

### .NET Framework version

6.0

### Description

`Image.LoadAsync(stream)` ends up synchronously reading from the source stream in both `Image.InternalDetectFormatAsync()` and then again in `ImageDecoderUtilities.DecodeAsync()`.  When running in a webservice under high-load scenarios, this can quickly lead to thread pool exhaustion as all threads are blocked waiting on network IO.

```cs
private static Task<IImageFormat> InternalDetectFormatAsync(Stream stream, Configuration config)
{
    return Task.FromResult(InternalDetectFormat(stream, config));
}
```

```cs
public static Task<Image<TPixel>> DecodeAsync<TPixel>(...)
{
    using var bufferedReadStream = new BufferedReadStream(configuration, stream);
    Image<TPixel> image = decoder.Decode<TPixel>(bufferedReadStream, cancellationToken);
    return Task.FromResult(image);
}
```

### Steps to Reproduce

Try to load an image asynchronously and observe that threads end up blocked on synchronous reads

kroymann:
I initially started to implement a PR that would fix this by making both of those methods _actually_ async, but that has long tendrils that extend deep into the image decoding code.  And even if you ignore the amount of duplicate sync/async code paths such a PR would create, I'm now questioning whether that's actually even the right way to fix this.

As a bit of background, this issue only occurs if the source stream passed in to `Image.LoadAsync()` is seekable, because in that scenario the image decoding reads directly from the source stream.  If the source stream is _not_ seekable, then the code first creates a new `ChunkedMemoryStream`, then asynchronously copies the source data into that in-memory stream, and finally performs the rest of the actions on that in-memory stream.  In that latter scenario, synchronous reads are not only fine, but are actually preferrable as they avoid unnecessary overhead.

[...]

PR #2006, opened 2022-02-16
Fix async image load

kroymann:
### Prerequisites

### Description
This addresses the issue outlined in #1997 where `Image.LoadAsync(stream)` would incorrectly use synchronous IO if the stream is seekable.  The fix is simply to remove the optimization that avoids prebuffering the stream into an in-memory buffer if the stream is seekable, and instead _always_ prebuffer the stream.  This ensures that async IO is always used to read the stream.


src/ImageSharp/Formats/IImageDecoder.cs
```diff
@@ -19,19 +19,21 @@ public interface IImageDecoder
         /// <typeparam name="TPixel">The pixel format.</typeparam>
         /// <param name="configuration">The configuration for the image.</param>
         /// <param name="stream">The <see cref="Stream"/> containing image data.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
         /// <returns>The <see cref="Image{TPixel}"/>.</returns>
         // TODO: Document ImageFormatExceptions (https://github.com/SixLabors/ImageSharp/issues/1110)
-        Image<TPixel> Decode<TPixel>(Configuration configuration, Stream stream)
+        Image<TPixel> Decode<TPixel>(Configuration configuration, Stream stream, CancellationToken cancellationToken = default)
```
antonfirsov, line 25: Since no users are expected to call this method directly, we should avoid making the `cancellationToken` argument optional.

kroymann: Really?  Is that a stylistic choice within the ImageSharp codebase not to have optional arguments?  Given that these are methods on a public interface, I would have assumed that there might be people using these APIs directly, so providing a default for the `cancellationToken` seemed like a good way to minimize the breaking nature of the change.

antonfirsov: Yeah we tend to avoid optional arguments, because there is higher chance for making binary breaking changes. (It can be a source of endless arguments whether this is right or not, but it's not the point here and `CancellationToken` is actually an exception for this rule.)

What's important in this context is that `IImageDecoder` is not meant to be consumed, but to *be implemented*. With this in mind, avoiding `cancellationToken = default` makes things simpler.

src/ImageSharp/Image.FromStream.cs
```diff
@@ -770,25 +770,11 @@ public static Image Load(Configuration configuration, Stream stream, out IImageF
                 throw new NotSupportedException("Cannot read from the stream.");
             }
 
-            // To make sure we don't trigger anything with aspnetcore then we just need to make sure we are
-            // seekable and we make the copy using CopyToAsync if the stream is seekable then we aren't using
-            // one of the aspnetcore wrapped streams that error on sync api calls and we can use it without
-            // having to further wrap
-            if (stream.CanSeek)
-            {
-                if (configuration.ReadOrigin == ReadOrigin.Begin)
```
JimBobSquarePants, line 779: So here we're resetting the input stream position before we attempt to decode because [`Configuration.ReadOrigin`](https://docs.sixlabors.com/api/ImageSharp/SixLabors.ImageSharp.Configuration.html#SixLabors_ImageSharp_Configuration_ReadOrigin) has been set. We need to ensure that we still respect that configuration option.

kroymann: Ahhh now I get what you were saying.  Somehow I completely missed that bit.  I'll add that back in.

JimBobSquarePants:
Everything looks good so far from an implementation perspective. We'll need to figure out a strategy now to compare the new Async performance with Main.

JimBobSquarePants:
@antonfirsov I just pushed a change to `ChunkedMemoryStream` that allows the size of the pooled buffer chunks to scale with the total allocation.

Allocation blocks match this scale.

```
LargeChunks = Min(4Mb, MemoryAllocator.BufferCapacityInBytes())
LargeChunkThreshold = Min(1Mb, LargeChunk / 4)
SmallChunks = Min(128Kb, LargeChunks / 32)
```

Once the total allocation within the stream hits our threshold we immediately switch to allocating larger chunk sizes.

I chose the the the values for the following reasons

- Small @ 128Kb because that's the default chunk size `RecycledMemoryStream` uses. 
- Threshold @ 1Mb because that's the threshold of the ArrayPool. I was tempted to go for 512Kb here.
- Large @ 4Mb because that matches the best performing pool size for our current allocator.

I'm not precious about any of these values to if you feel like they should change then please suggest alternatives.

JimBobSquarePants (approved):
I'm approving this but want a second opinion before merging.

antonfirsov:
I want to take a look next week. Quite distracted from everything programming-related thanks to the ongoimg Fourth Reich experiment on my continent.

JimBobSquarePants:
Of course mate. Absolute horror show over there!

antonfirsov:
@kroymann can you please give me permission to push to your branch? (Thought it is always allowed for contributors for PR branches.)
I want to push my load tests:
https://github.com/antonfirsov/ImageSharp/commit/63e897e973a6f6fbe58e5b0edcc070901281c60c

antonfirsov:
I did before / after runs of asnchronous load testing:
- It looks like we can nicely see the result of thread pool starvation on `main` branch: The synchronous parallel run finishes in about 6.3 seconds on my 10 core machine, while the async version needs ~9.2 seconds.
- The PR seems to fix this nicely, bringing down async execution time close to the synchronous time
- Unfortunately, there is a significant *~25% memory regression*

#### Before
image

#### After
image

We should try to address it. Will take quick look at the code changes now.

antonfirsov (approved):
Ok, I have instant idea to improve this.

If you guys both agree that 25% extra memory is an acceptable price for correct async, I'm fine merging this. We can open an issue to track the sync/async memory gap.

Just please remember to merge my load test addition!

antonfirsov:
Btw @kroymann thanks for discovering the problem, and pushing through the fix, really great job!

JimBobSquarePants:
@antonfirsov Your permissions issues were likely due to Git LFS issues with GitHub. 

There's a workaround that should allow you to push to the fork.
```
rm .git/hooks/pre-push
```

kroymann:
@antonfirsov I think @JimBobSquarePants is right about the github permissions, but I went ahead and cherry-picked your commit just to expedite things.  Thank you for adding those tests!

antonfirsov:
> I'm assuming the memory overheads come from my buffer allocation? Would a drop to 2MB segments be useful?

2MB logical allocations lead to 4MB actual allocations since buffers are uniform.

Actually, there might be a way to create a smarter buffer growing strategy.

This is how subsequent chunks sizes are allocated currently if I get it right.
```
128K -> 128K -> 128K -> 128K -> 128K -> 128K -> 128K -> 128K ->
4MB -> 4MB -> 4MB  -> ...
```
This stressing the 128K ArrayPool unnecessarily, and often leads to a wasteful allocation in the end.

Instead, I would try to do something like:

```
128K -> 128K -> 128K -> 128K ->
256K -> 256K ->  256K -> 256K ->
512K -> 512K ->  512K -> 512K ->
1MB -> 1MB -> 1MB -> 1MB -> 
4MB -> 4MB -> 4MB  -> ...
``` 

This would keep most of the buffers rented from `ArrayPool.Shared` without stressing individual buckets too much and avoid wasting space. The unmanaged pool wouldn't be touched for images smaller than 7 MB.

JimBobSquarePants:
Yeah, that's how it's allocated currently. Shouldn't be too hard to update it to follow your pattern.

antonfirsov:
Here is a closed formula to get the i-th chunk size according to my logic, assuming you increment `i` each time you allocate a chunk:

```C#
private static int GetChunkSize(int i)
{
    const int _128K = 1 << 17;
    const int _4M = 1 << 22;
    return i < 16 ? _128K * (1 << (i / 4)) : _4M;
}
```

Tomorrow is the last day I have access to my benchmark machine to redo the benchmarks, I will leave for two weeks afterwards.

JimBobSquarePants:
It's always tomorrow in Aus!

That's the changes pushed. I added a check to ensure we always use the min of the allocator capacity or the calculated size, in case we change things in the future or someone implements their own allocator.

antonfirsov:
Looks like there is a slight improvement. The total memory gap seems to be of 20%.

image

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -547,6 +543,16 @@ private void ReleaseMemoryChunks(MemoryChunk chunk)
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static int GetChunkSize(int i)
+        {
+#pragma warning disable IDE1006 // Naming Styles
+            const int _128K = 1 << 17;
```
antonfirsov, line 550: A message to our future selves.
```suggestion
            // Increment chunks sizes with moderate speed, but without using too many buffers from the same ArrayPool bucket of the default MemoryAllocator.
            //https://github.com/SixLabors/ImageSharp/pull/2006#issuecomment-1066244720
#pragma warning disable IDE1006 // Naming Styles
            const int _128K = 1 << 17;
```

antonfirsov (approved):
LGTM, thanks for the patience!

PR #2013, opened 2022-02-18
Tiff Decoder option to decode only first frame

brianpopow:
### Prerequisites

### Description

As discussed in #1982, this adds the TiffDecoder option `FrameDecodingMode`, which allows to decode only the first frame.


JimBobSquarePants:
Looks good to me but I think we should hold for 2.1 release before merging this. I want the 2.0.1 release to contain critical bug fixes only.

brianpopow:
> Looks good to me but I think we should hold for 2.1 release before merging this. I want the 2.0.1 release to contain critical bug fixes only.

Yes sure, makes sense.

JimBobSquarePants:
@brianpopow I've decoded to skip 2.0.1 and do a 2.1 release instead. I think we've got a few good things to ship plus fixes.

JimBobSquarePants (approved):

PR #2015, opened 2022-02-18
WIP: Add ARM version of adler32

brianpopow:
### Prerequisites

### Description

I was experimenting a bit with ARM hardware intrinsics and decided to port the adler32 version of chromium to ImageSharp.

BenchmarkDotNet=v0.13.1, OS=ubuntu 20.04
Unknown processor
.NET SDK=5.0.402
  [Host]     : .NET 5.0.11 (5.0.1121.47308), Arm64 RyuJIT
  DefaultJob : .NET 5.0.11 (5.0.1121.47308), Arm64 RyuJIT

|        Method |      Mean |     Error |    StdDev |
|-------------- |----------:|----------:|----------:|
|   AdlerARM    |  1.082 us | 0.0123 us | 0.0109 us |
|   AdlerScalar |  3.823 us | 0.0165 us | 0.0146 us |

CPU info:
```
Architecture:        aarch64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              6
On-line CPU(s) list: 0-5
Thread(s) per core:  1
Core(s) per socket:  3
Socket(s):           2
Vendor ID:           ARM
Model:               4
Model name:          Cortex-A53
Stepping:            r0p4
CPU max MHz:         1896.0000
CPU min MHz:         100.0000
BogoMIPS:            48.00
Flags:               fp asimd evtstrm aes pmull sha1 sha2 crc32
```

Benchmark was run with the following data:

```
int N = 4096;
byte[] data = new byte[N];
new Random(1).NextBytes(data);
```

**There is still a bug, adler results for N 4096:**

Arm: 1010560139
Scalar: 1010560140

### DIFF
--- a/src/ImageSharp/Compression/Zlib/Adler32.cs
+++ b/src/ImageSharp/Compression/Zlib/Adler32.cs
@@ -6,6 +6,9 @@ using System.Runtime.CompilerServices;
 #if SUPPORTS_RUNTIME_INTRINSICS
 using System.Runtime.Intrinsics;
 using System.Runtime.Intrinsics.X86;
+#if NET5_0_OR_GREATER
+using System.Runtime.Intrinsics.Arm;
+#endif
 #endif
 
 #pragma warning disable IDE0007 // Use implicit type
@@ -23,13 +26,15 @@ namespace SixLabors.ImageSharp.Compression.Zlib
         public const uint SeedValue = 1U;
 
         // Largest prime smaller than 65536
-        private const uint BASE = 65521;
+        private const uint Base = 65521;
 
         // NMAX is the largest n such that 255n(n+1)/2 + (n+1)(BASE-1) <= 2^32-1
-        private const uint NMAX = 5552;
+        private const uint Nmax = 5552;
 
 #if SUPPORTS_RUNTIME_INTRINSICS
         private const int MinBufferSize = 64;
+
+        private const uint ArmBlockSize = 1 << 5;
 
         // The C# compiler emits this as a compile-time constant embedded in the PE file.
         private static ReadOnlySpan<byte> Tap1Tap2 => new byte[]
@@ -67,11 +72,14 @@ namespace SixLabors.ImageSharp.Compression.Zlib
             {
                 return CalculateSse(adler, buffer);
             }
-
-            return CalculateScalar(adler, buffer);
-#else
-            return CalculateScalar(adler, buffer);
+#if NET5_0_OR_GREATER
+            if (AdvSimd.IsSupported)
+            {
+                return CalculateArm(adler, buffer);
+            }
 #endif
+#endif
+            return CalculateScalar(adler, buffer);
         }
 
         // Based on https://github.com/chromium/chromium/blob/master/third_party/zlib/adler32_simd.c
@@ -95,7 +103,7 @@ namespace SixLabors.ImageSharp.Compression.Zlib
                 fixed (byte* tapPtr = Tap1Tap2)
                 {
                     index += (int)blocks * BLOCK_SIZE;
-                    var localBufferPtr = bufferPtr;
+                    byte* localBufferPtr = bufferPtr;
 
                     // _mm_setr_epi8 on x86
                     Vector128<sbyte> tap1 = Sse2.LoadVector128((sbyte*)tapPtr);
@@ -105,7 +113,7 @@ namespace SixLabors.ImageSharp.Compression.Zlib
 
                     while (blocks > 0)
                     {
-                        uint n = NMAX / BLOCK_SIZE;  /* The NMAX constraint. */
+                        uint n = Nmax / BLOCK_SIZE;  /* The NMAX constraint. */
                         if (n > blocks)
                         {
                             n = blocks;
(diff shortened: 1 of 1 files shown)


src/ImageSharp/Compression/Zlib/Adler32.cs
```diff
@@ -192,35 +199,167 @@ private static unsafe uint CalculateSse(uint adler, ReadOnlySpan<byte> buffer)
...
+                    {
+                        n = (uint)blocks;
+                    }
+
+                    blocks -= n;
+
+                    // Process n blocks of data. At most nMax data bytes can be
+                    // processed before s2 must be reduced modulo Base.
+                    var vs2 = Vector128.Create(0, 0, 0, s1 * n);
+                    Vector128<uint> vs1 = Vector128<uint>.Zero;
```
gfoidl, line 245:
```suggestion
                    Vector128<uint> vs1 = Vector128<uint>.Zero;
                    Vector128<uint> vs2 = vs1.WithElement(3, s1 * n);
```
I don't know if codegen for ARM will be better this way...don't have a test-setup for this now.
But for [amd64](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8YUCAAcAymIBuvMDFwBuGvSatOPfoJEYovLrgUqaagMwNuubADMYTUgwDCDAN40GXhhIOzsGA7MSAwAFABqMGAY0OSkABwAPByGGAB8aAyR0bEJyalpAJQMAGIQEKEpPAy45AByhZ7eHtTebVlRMVBxSVXpDLK4TgC8HTndCSzOsAEwoXSZCwxLtQ2qre1e2V09+TxpA7UMo9u5vQUsAFpiEOubXk2bxADsYYPkmYOkhXfeAL6PBiA3y8fyBDQhCKdM57dKZU4Tc77YoAIWwUEqqRq9UaGy8LXuYx2eT6B3exyJMNJVxuv02CN2pMOI0OrAA6rwMAALACiABsYAIeKFzJlVj9AW1Jd4Xm9ap8hhK8QwAdQ/kA===) codegen is better, and I assume the JIT will emit similar code for ARM.

src/ImageSharp/Compression/Zlib/Adler32.cs
```diff
@@ -192,35 +199,167 @@ private static unsafe uint CalculateSse(uint adler, ReadOnlySpan<byte> buffer)
...
 #endif
 
         [MethodImpl(InliningOptions.HotPath | InliningOptions.ShortMethod)]
         private static unsafe uint CalculateScalar(uint adler, ReadOnlySpan<byte> buffer)
         {
             uint s1 = adler & 0xFFFF;
             uint s2 = (adler >> 16) & 0xFFFF;
-            uint k;
 
             fixed (byte* bufferPtr = buffer)
```
gfoidl, line 355:
```suggestion
            fixed (byte* bufferPtr = &MemoryMarshal.GetReference(buffer))
```

gfoidl (commented):
For .NET 7 cross-platform intrinsics will be available, so that code can be written on behalf of `Vector128<T>` and not via `Sse2` / `AdvSimd`.

I know that it's still a long time until is .NET 7 is available and adopted, but is this something that should be focused on now already?

brianpopow:
> For .NET 7 cross-platform intrinsics will be available, so that code can be written on behalf of `Vector128<T>` and not via `Sse2` / `AdvSimd`.
> 
> I know that it's still a long time until is .NET 7 is available and adopted, but is this something that should be focused on now already?

That's interesting, I was not aware of this. How would that work? SSE has intrinsics which ARM does not have and vice versa.

I think its to early to early to consider this now. Anything ARM related is now only pretty much experimental and will not land in the main branch anytime soon.

gfoidl:
> How would that work? SSE has intrinsics which ARM does not have and vice versa

Only the "common subset" is exposes as xplat-intrinsics. The journey started in https://github.com/dotnet/runtime/issues/49397 but more operations got added so far -- I can't find a up-to-date summary of all operations 😢.
For the special cases one needs to fall back to the current intrinsics style (Sse.IsSupported / AdvSimd.IsSupported).

> I think its to early to early to consider this now

This is also my expectation. Just wanted to point to it (as a reminder that something is on the way).

brianpopow:
> This is also my expectation. Just wanted to point to it (as a reminder that something is on the way).

Thanks for sharing, that's very interesting to know.

antonfirsov:
> I know that it's still a long time until is .NET 7 is available and adopted, but is this something that should be focused on now already?

No, since - unlike the BCL - we have to maintain multiple .NET versions backwards, not just the shiniest-latest one, which means we would end up complicating our code with lots of `#if`-s instead of simplifying it. These are cool features for dotnet/runtime itself, but radically less useful for 3rd libraries until .NET 6 goes out of support.

gfoidl:
@antonfirsov I partly agree.

Total agreement for backwards-compat and what you said about the BCL.

For the case of ARM I neither agree nor disagree. ARM isn't supported right now, so no problems with backwards-compat, hence this opens the opportunity to use one code path with the xplat-intrinsics for ARM and x86/x64. '#if'def would then only be for `NET7_0_OR_GREATER`.

Right now it looks like non-BCL libraries have to go special routes / code-pathes for each type of HW-platform, then with .NET 7 (and when .NET 6 isn't supported anymore) these code-pathes can be merged by using the xplat-intrinsics.

So if ARM should be supported by .NET 6 the different code-pathes is IMO the right way to go.
If ARM support is sufficient to have for .NET 7 onwards, the one should go with the xplat-intrinsics.

But it's relatively new concept, so it has to evolve and so it's good to have some discussion here to see what's doable and what's not.

Issue #2018, opened 2022-02-19
Get "Invalid PNG data." exception during Identify phase of loading PNG image when image contains sBIT chunk


flew2bits:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

v2.0.0

### Other ImageSharp packages and versions

ImageSharp.Drawing v1.0.0beta14

### Environment (Operating system, version and so on)

Windows 10, Rider

### .NET Framework version

6.0.1

### Description

When loading an image that contains an sBIT chunk, the Identify phase throws an exception stating that the data is invalid. The switch statement in the identify function (around line 250 of PngDecoderCore.cs) does not include the PngChunkType.SignificantBits as a case, nor does it have a default clause. Adding a default clause as follows fixes the issue:

`default:
    this.SkipChunkDataAndCrc(chunk);
    break;`

I may try to submit a PR if I can figure out how to do so.

Unfortunately, I don't have permission to upload the image that is causing this problem.

### Steps to Reproduce

Image.Load(stream) from a stream that has a PNG with a sBIT chunk.

brianpopow:
> Unfortunately, I don't have permission to upload the image that is causing this problem.

You cannot attach the image here? Just drag and drop it in the text box should work.

flew2bits:
It's not that I'm unable to attach the image -- it's a copyright image that I don't have permission to upload here.

brianpopow:
> It's not that I'm unable to attach the image -- it's a copyright image that I don't have permission to upload here.

Ok, I misunderstood that. Then we need to find another image with the same chunk for a unit test.

Thanks for opening a PR for this.

flew2bits:
It looks like I'll need to figure out why the three resolution tests are failing.

flew2bits:
It looks like there's some other reason my file isn't working right. Several of the test files also have sBIT chunks and they work fine. However, Windows and Chrome can both open my file just fine, and a file analyzer I used on it didn't find any problems, so I'll see if I can find an alternate solution.

brianpopow:
[...]

PR #2019, opened 2022-02-19
Fix issue in PNG identify method to skip "uninteresting" chunks, incl…

flew2bits:
…uding sBIT.

### Prerequisites

### Description
Add a default clause in the PNG Identify method to skip over "uninteresting" chunks so that the data is properly read. 
This fixes issue #2018

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -336,6 +336,14 @@ namespace SixLabors.ImageSharp.Formats.Png
                                 break;
                             case PngChunkType.End:
                                 goto EOF;
+
+                            default:
+                                if (this.colorMetadataOnly)
+                                {
+                                    this.SkipChunkDataAndCrc(chunk);
+                                }
+
+                                break;
                         }
                     }
                     finally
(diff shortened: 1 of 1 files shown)


flew2bits:
The call to SkipChunkDataAndCrc was in the wrong location to fix the problem. I've moved it to a location that should fix the problem.

flew2bits:
I moved the call to SkipChunkDataAndCrc back to the default switch clause based on how other chunks are handled in the switch statement. It's not wrapped in a check to see if the identification is for color metadata only, also based on other clauses.

brianpopow:
I have found an png with a sBit chunk:
length_sbit

Would that reproduce the described issue in #2018? If so, it would be good to have  a unit test covering this issue. See [PngDecoderTests](https://github.com/SixLabors/ImageSharp/blob/92da254bb8203ea5eb2fc8f58f849f487921466e/tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs#L246) for an example howto do that.

flew2bits:
Unfortunately that image doesn't exhibit the problem. I'll see if I can put one together that can be used for testing.

JimBobSquarePants:
I'm actually happy to push this in without the offending image. As a performance optimization alone it's worth the addition. 

@flew2bits if you could supply an image at some point that would be really great.

JimBobSquarePants:
@brianpopow Yeah... Let's get it merged now.

PR #2014, opened 2022-02-18
Gif decoder. Fix memory access violations.

JimBobSquarePants:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

Fix #2012

- Fixes `GifXmpApplicationExtension` by using a more robust (_handles empty profiles_) and less allocation heavy reader.
- Provides protection against invalid memory operations in `LzwDecoder`
- Fixes cloning of `XmpProfile` when cloning `ImageMetadata`

Integrity of reencoded metadata has been tested externally using ExifTools

PR #2020, opened 2022-02-19
PNG: Throw exception, if palette chunk is missing

brianpopow:
### Prerequisites

### Description

This PR changes the PNG decoder to throw an appropriate exception, if the palette chunk is missing.

### DIFF
--- a/src/ImageSharp/Formats/Png/PngThrowHelper.cs
+++ b/src/ImageSharp/Formats/Png/PngThrowHelper.cs
@@ -22,7 +22,7 @@ namespace SixLabors.ImageSharp.Formats.Png
         public static void ThrowNoData() => throw new InvalidImageContentException("PNG Image does not contain a data chunk");
 
         [MethodImpl(InliningOptions.ColdPath)]
-        public static void ThrowMissingPalette() => throw new InvalidImageContentException("PNG Image does not contain palette chunk");
+        public static void ThrowMissingPalette() => throw new InvalidImageContentException("PNG Image does not contain a palette chunk");
 
         [MethodImpl(InliningOptions.ColdPath)]
         public static void ThrowInvalidChunkType() => throw new InvalidImageContentException("Invalid PNG data.");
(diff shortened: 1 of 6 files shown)


src/ImageSharp/Formats/Png/PngThrowHelper.cs
```diff
@@ -21,6 +21,9 @@ public static void ThrowInvalidImageContentException(string errorMessage, Except
         [MethodImpl(InliningOptions.ColdPath)]
         public static void ThrowNoData() => throw new InvalidImageContentException("PNG Image does not contain a data chunk");
 
+        [MethodImpl(InliningOptions.ColdPath)]
+        public static void ThrowMissingPalette() => throw new InvalidImageContentException("PNG Image does not contain palette chunk");
```
dlemstra, line 25: Maybe `a palette chunk`?

JimBobSquarePants:
What's the current behavior? Out of range exception?

https://media.githubusercontent.com/media/SixLabors/ImageSharp/29ddc6053e1a27ab095e87d8baa015f7428c53ef/tests/Images/Input/Png/missing_plte.png

The first referenced example is unopenable using conventional editors Edge and Chrome can open it displaying it as a 16x16 transparent png. 

I haven't tried the second but I wonder whether we should do the same? #2014 updates our Gif decoder to be more robust in this manner when decoding images whose color palette cannot be assigned. If we did the same for Png then the behavior for both  decoders would then match browser behavior.

brianpopow:
> What's the current behavior? Out of range exception?

System.NullReferenceException: 'Object reference not set to an instance of an object.' when accessing the empty palette span.

> https://media.githubusercontent.com/media/SixLabors/ImageSharp/29ddc6053e1a27ab095e87d8baa015f7428c53ef/tests/Images/Input/Png/missing_plte.png
> 
> The first referenced example is unopenable using conventional editors Edge and Chrome can open it displaying it as a 16x16 transparent png.
> 
> I haven't tried the second but I wonder whether we should do the same? #2014 updates our Gif decoder to be more robust in this manner when decoding images whose color palette cannot be assigned. If we did the same for Png then the behavior for both decoders would then match browser behavior.

The png is corrupt and in my opinion throwing an exception and stating what's wrong with image is the right thing to do in such case. I dont think silently ignoring this would be good.

JimBobSquarePants:
>The png is corrupt and in my opinion throwing an exception and stating what's wrong with image is the right thing to do in such case. I dont think silently ignoring this would be good.

Yeah, I've been flip-flopping around that. I'm gonna update my Gif PR to do what you do. 👍

JimBobSquarePants (approved):
LGTM Thanks! 👍

PR #2021, opened 2022-02-21
PNG: Ignore invalid gamma chunks

brianpopow:
### Prerequisites

### Description

This PR handles another edge case when the gamma chunk does not contain enough data.


JimBobSquarePants:
What does libpng do in this situation? Gamma chunk is not mandatory and can use default values in certain situations.

http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html

JimBobSquarePants (approved):
LGTM 👍

PR #2023, opened 2022-02-21
Minor PNG Encoding Optimization for DeflaterEngine and WuQuantizer

ratok-mk:
### Prerequisites

### Description
I investigated some possible code optimizations and found minor issues with unnecessary operations in:
- DeflaterEngine
- WuQuantizer

It is connected to the following [discussion](https://github.com/SixLabors/ImageSharp/discussions/2010)

A comparison with BenchmarkDotNet showed ~3% improvement for a 2560 x 1920 image 16 Bit Grayscale Image.
Also used existing test 'ImageSharp Png' for .NET 5.0:
Before (mean): 36.07ms 
Optimized (mean): 35.62 ms

### DIFF
--- a/src/ImageSharp/Compression/Zlib/DeflaterEngine.cs
+++ b/src/ImageSharp/Compression/Zlib/DeflaterEngine.cs
@@ -486,15 +486,16 @@ namespace SixLabors.ImageSharp.Compression.Zlib
             this.matchLen = Math.Max(this.matchLen, DeflaterConstants.MIN_MATCH - 1);
             int matchLength = this.matchLen;
 
-            if (scan + matchLength > scanMax)
+            int scanEndPosition = scan + matchLength;
+            if (scanEndPosition > scanMax)
             {
                 return false;
             }
 
             byte* pinnedWindow = this.pinnedWindowPointer;
             int scanStart = this.strstart;
-            byte scanEnd1 = pinnedWindow[scan + matchLength - 1];
-            byte scanEnd = pinnedWindow[scan + matchLength];
+            byte scanEnd1 = pinnedWindow[scanEndPosition - 1];
+            byte scanEnd = pinnedWindow[scanEndPosition];
 
             // Do not waste too much time if we already have a good match:
             if (matchLength >= this.goodLength)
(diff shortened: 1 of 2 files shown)


src/ImageSharp/Compression/Zlib/DeflaterEngine.cs
```diff
@@ -486,15 +486,16 @@ private bool FindLongestMatch(int curMatch)
             this.matchLen = Math.Max(this.matchLen, DeflaterConstants.MIN_MATCH - 1);
             int matchLength = this.matchLen;
```
gfoidl, line 487:
```suggestion
            int matchLength = this.matchLen;
            matchLength  = Math.Max(matchLength, DeflaterConstants.MIN_MATCH - 1);
            this.matchLen = matchLength;
```
Should give better codegen -- especially for older runtimes (JIT).

src/ImageSharp/Compression/Zlib/DeflaterEngine.cs
```diff
@@ -486,15 +486,16 @@ private bool FindLongestMatch(int curMatch)
...
 
-            if (scan + matchLength > scanMax)
+            int scanEndPosition = scan + matchLength;
+            if (scanEndPosition > scanMax)
             {
                 return false;
             }
```
gfoidl, line 493: Is there a chance that `scan + matchLength` can overflow? If not, please ignore this suggestion.

To avoid potential overflow:
```suggestion
            if (scan > scanMax - matchLength)
            {
                return false;
            }
            int scanEndPosition = scan + matchLength;
```

src/ImageSharp/Compression/Zlib/DeflaterEngine.cs
ratok-mk: It looks strange and somehow duplicate in code. So if it helps the JIT codegen, I agree to use it.

src/ImageSharp/Compression/Zlib/DeflaterEngine.cs
ratok-mk: It should be very rare for really large files. Then the suggestion would help.

JimBobSquarePants (approved):
LGTM. Thanks! 👍

PR #2039, opened 2022-03-06
AVX conversions for Luminance and Rgb

ynse01:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided _benchmark_ coverage for my change (where applicable)

### Description
This PR fixed #1763

Implemented AVX versions of Luminance and RGB conversions in JPEG Encoding, as described in the issue.
Added Benchmarks for these cases.

Results with and without AVX:
|                                  Method | Quality |     Mean |    Error |   StdDev |
|---------------------------------------- |-------- |---------:|---------:|---------:|
|     ImageSharp (greyscale) Jpeg 4:0:0 |      75 | 12.90 ms | 0.074 ms | 0.069 ms |
| ImageSharp (greyscale) Jpeg 4:0:0 AVX |      75 | 11.96 ms | 0.076 ms | 0.072 ms |
|                   ImageSharp Jpeg rgb |      75 | 28.45 ms | 0.130 ms | 0.115 ms |
|               ImageSharp Jpeg rgb AVX |      75 | 26.22 ms | 0.051 ms | 0.046 ms |
|     ImageSharp (greyscale) Jpeg 4:0:0 |      90 | 16.23 ms | 0.273 ms | 0.255 ms |
| ImageSharp (greyscale) Jpeg 4:0:0 AVX |      90 | 15.15 ms | 0.234 ms | 0.219 ms |
|                   ImageSharp Jpeg rgb |      90 | 36.84 ms | 0.258 ms | 0.241 ms |
|               ImageSharp Jpeg rgb AVX |      90 | 34.56 ms | 0.111 ms | 0.104 ms |
|     ImageSharp (greyscale) Jpeg 4:0:0 |     100 | 20.57 ms | 0.080 ms | 0.075 ms |
| ImageSharp (greyscale) Jpeg 4:0:0 AVX |     100 | 19.55 ms | 0.070 ms | 0.086 ms |
|                   ImageSharp Jpeg rgb |     100 | 54.64 ms | 0.102 ms | 0.090 ms |
|               ImageSharp Jpeg rgb AVX |     100 | 51.67 ms | 0.112 ms | 0.105 ms |

PR #2051, opened 2022-03-09
Improved jpeg encoding benchmarks

br3aker:
### Prerequisites

### Description
Based on discussion from https://github.com/SixLabors/ImageSharp/pull/2039. I've split jpeg encoding benchmark into ImageSharp-only benchmark with all supported color types and 'cross-codec' which compares ImageSharp with other codec (only SkiaSharp currently as `System.Drawing` is unreliable on Linux).

I've also deleted multi-image benchmark as new benchmarks can be used via replacing filename from const string to string with `Params` attribute.

Ideally, we would want to create a encode-decode benchmark as it's would be a lot easier to trace and compare to something like MagicScaler as its decoding is tighly coupled with encoding.

Decode-Resize-Encode benchmark would also be helpful and I'd implement it during jpeg dct scaling PR somewhere in the future - no point implementing it now.


tests/ImageSharp.Benchmarks/Codecs/Jpeg/EncodeJpegFeatures.cs
```diff
@@ -0,0 +1,89 @@
...
+| Benchmark |    YCbCrRatio420 |      75 |  6.463 ms | 0.0282 ms | 0.0235 ms |
+| Benchmark |    YCbCrRatio444 |      75 |  8.616 ms | 0.0422 ms | 0.0374 ms |
+| Benchmark |        Luminance |      90 |  7.011 ms | 0.0361 ms | 0.0301 ms |
+| Benchmark |              Rgb |      90 | 13.119 ms | 0.0947 ms | 0.0886 ms |
+| Benchmark |    YCbCrRatio420 |      90 |  6.786 ms | 0.0328 ms | 0.0274 ms |
+| Benchmark |    YCbCrRatio444 |      90 |  8.672 ms | 0.0772 ms | 0.0722 ms |
+| Benchmark |        Luminance |     100 |  9.554 ms | 0.1211 ms | 0.1012 ms |
+| Benchmark |              Rgb |     100 | 19.475 ms | 0.1080 ms | 0.0958 ms |
+| Benchmark |    YCbCrRatio420 |     100 | 10.146 ms | 0.0585 ms | 0.0519 ms |
+| Benchmark |    YCbCrRatio444 |     100 | 15.317 ms | 0.0709 ms | 0.0592 ms |
```
br3aker, line 88: Luminance encoding has same (or even worse) performance as YCbCr 4:2:0 because Rgb24 -> L8 conversion uses scalar code.

JimBobSquarePants: Yeah theres definitely optimisation opportunity there.

JimBobSquarePants (approved):
Thanks for this, looks great.

PR #2052, opened 2022-03-09
Fix resize pad color application for non Nearest Neighbor samplers

JimBobSquarePants:
### Prerequisites

### Description

The new `ResizeOptions.PadColor` property fill operation was being overwritten by the `ResizeWorker`. This changes fixes the worker to only allocate buffers matching the target rectangle of interest writing them to the correct target span.


src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs
```diff
@@ -101,7 +101,7 @@ protected virtual void Dispose(bool disposing)
         /// Returns a <see cref="ResizeKernel"/> for an index value between 0 and DestinationSize - 1.
         /// </summary>
         [MethodImpl(InliningOptions.ShortMethod)]
-        internal ref ResizeKernel GetKernel(int destIdx) => ref this.kernels[destIdx];
+        internal ref ResizeKernel GetKernel(nint destIdx) => ref this.kernels[destIdx];
```
gfoidl, line 104: For the plain array access `nint` doesn't bring any advantage.
Here the JIT misses a optimization (a bug is tracked, but can't find it at the moment).

If you want best machine code here, look at [Sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgBEBLbAcwDsJcGDmFwBuGvSasASgFdeQ/DBYBhCPgAOHADYwoAZT0A3YTDETGzFnIUclLAJIK9EDYagmwZ8dRodnULzY2kykDCoMAN40DLEMGlAcRtgYMAyw2AAmELzaAJ4M0mYcAF4wANJ6vDDaANoAugwA+gDWVTXm1HEMMXHEAMzpMABmhcVllYE1DADiMBiT1doAggAU/hgMmWYYDpkIAJQMALwAfEOjre3auLXbgnsI9T7dAxdjuKUV17PzizUAIVWvA2Wx2jyOZ3eVymNzu4P2z16sTesFGRU+Ex+cwW1xUwNB912+wOyKiZO6bBgwFk3BYy1wuD0GFWRMeDFOxwYjAAZDywQ99gwADzNNqw3AsAAyMF43AwAAsDi9unE0R8vv8QuKlkVRlz1QBZGD4aB5Q3YKC4BXBFg45ZQKDYPJsFLYPV6WVeVYwpa4ZUUvoAdneAFVeLhsMNlMtMplVuqdTU9WgBcTDirYgBfGg53zUDZVYJDLI5fIMeSR6MMQRQWRgTYYzXXGjRLpxBJJFJpDLZXIFYbaCApABUDBpw2jUAAChgoD4yQB6Zci3CyfD4S15U5LldODhCYJfXAMbAMaoAdwY/kE2F4XgYEFGirSwqZaTAaOOACIm1jYd+i7nLW9YYCwu6LiKi5rhuW47u2sSFoExZ/t8sLrAoNYYJauy8NsCCpoOw4YGOE5TrOUCpqCuhyoqpIIeSDHdIqHCSvo2FQLh+EnFhOFOPhmaqixkpkXoFE8aJM5zoJzEKqx0qyvKCo8TRSmCXmEGruum5QNumk4ieL68Zx154TACAMMM0AMEZRL+CkHA5OkEAXuBDHLpBwrQdpcFkm8oLsXxZkIGSbaqnE3DzOpNCaW+Pm6fB3Qeb8GCGQqaSqYqj7PulDBJtoblJSuXkwTpekMf5mEyrRCqhYGsSRRg0XUFmQA=) method `GetKernelC`.
It's more C# code and I doubt it's needed here. I'd just go with `int` (or `nint` as you have it now) and wait for the JIT fix.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeWorker.cs
```diff
@@ -131,17 +130,18 @@ public void FillDestinationPixels(RowInterval rowInterval, Buffer2D<TPixel> dest
                 ref Vector4 tempRowBase = ref MemoryMarshal.GetReference(tempColSpan);
 
                 int top = kernel.StartIndex - this.currentWindow.Min;
+
                 ref Vector4 fpBase = ref transposedFirstPassBufferSpan[top];
 
-                for (int x = 0; x < this.destWidth; x++)
+                for (nint x = left; x < right; x++)
```
gfoidl, line 136: Here it's good to have `nint`, as in the `Unsafe.Add`s below this removes the sign extending move `movsxd` which is beneficial.

gfoidl: Would it be better to have the loop as
```suggestion
                for (nint x = 0; x < (right - left); x++)
```
so that the subtractions below `x - left` can be eliminated (shifted to the loop variable), so that it's just `x`.
I assume current JIT won't (completely) CSE that expression.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeWorker.cs
```diff
@@ -184,13 +186,13 @@ private void CalculateFirstPassValues(RowInterval calculationInterval)
                 // Span<Vector4> firstPassSpan = transposedFirstPassBufferSpan.Slice(y - this.currentWindow.Min);
                 ref Vector4 firstPassBaseRef = ref transposedFirstPassBufferSpan[y - this.currentWindow.Min];
 
-                for (int x = this.targetWorkingRect.Left; x < this.targetWorkingRect.Right; x++)
+                for (nint x = left; x < right; x++)
```
gfoidl, line 189: Same here:
```suggestion
                for (nint x = 0; x < (right - left); x++)
```
? 
And adjust below to just `x` instead of `x - left`.

JimBobSquarePants: Ah, I have to add an extra incremental here. I need both `x` & `x - left`.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeWorker.cs
JimBobSquarePants: Good spot. Not sure how I missed that.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernelMap.cs
JimBobSquarePants: No `MemoryMarshal.GetArrayDataReference` for us yet unfortunately..... One day!

Issue #427, opened 2018-01-05
Feature proposal: handling tileable images


vpenades:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

I was wondering if there's any provision for handling tileable images... I know that brushes are tileable by default, but in most processors, image tiling/wrapping is simply not supported.

For example, let's say I have an image (texture) I know it's tileable, and I want to apply a gaussian blur with radius 5. The current behavior is that processed pixels close to a boundary would not take in account the pixels of the opposite boundary. The resulting image, when tiled, would show a noticeable sharp seam between tile limits.

Tipically when texture/tiling has an enumeration like this:

WrapModes:

- None
- WrapX
- WrapY
- MirrorX
- MirrorY

antonfirsov:
@vpenades interesting proposal, but really not easy to implement because of the cross-library architectural impact it brings. Might be a candidate feature for 2.0, but with our current focus, I'm afraid it's a bit off the radar, unless someone from the community takes care of it.

But we really want to hear the voice of the users, so anyone else having interest in this, please vote/leave your comments! :)

JimBobSquarePants:
I'd say it only affects only the convolution based processors so theoretically shouldn't be too big a deal. That said a PR would be most welcome.

antonfirsov:
It probably also affects drawing.

JimBobSquarePants:
You reckon? I don't think any of the drawing stuff really cares about edge pixels. I could be wrong though, it's not an area I've looked at in a while.

tocsoft:
Only thing that could be effected in drawing is the `ImageBrush` and that already tiles the fill image.

antonfirsov:
Supporting tiling would mean to me that we need to support "cross boundary" drawing in a way like this:
![image](https://user-images.githubusercontent.com/6835152/34841984-60d8a46e-f70a-11e7-92d7-87005ca0816c.png)

Black lines represent image boundary, the green hearth represents the shape drawn by `ImageSharp.Drawing`.

tocsoft:
[...]

PR #2060, opened 2022-03-13
Convolution: Border wrapping modes

ynse01:
### Prerequisites

### Description
I went through the somewhat older issues and found #427 to be of my interest to implement.

This PR creates three modes in which to treat the border pixels in a Convolution operation. These modes are:

1. Clamp (the current implementation)
2. Wrap
3. Mirror

I don't like the `None` name proposal in the original issue, as I feel you **have** to make a choice about your borders. And with the name `Clamp` we make clear what operation we perform.

Please advise on how to expose this in the public Convolution API's. 
For example, please specify your preference on questions like:

- On which class to expose the `BorderWrappingMode` property ?
- Specify the modes of X and Y directions separately? Or just a single property for both directions?
- Switch the default from `Clamp` to `Mirror` ?
- Add this functionality to Resize also ?

Other feedback or suggestion are also welcome !


src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
```diff
@@ -105,5 +81,79 @@ public void Dispose()
...
+                        Numerics.Clamp(span.Slice(0, affectedSize), min, max);
+                        Numerics.Clamp(span.Slice(span.Length - affectedSize), min, max);
+                        break;
+                    case BorderWrappingMode.Mirror:
+                        for (int i = 0; i < affectedSize; i++)
+                        {
+                            var value = span[i];
+                            if (value < min)
+                            {
+                                span[i] = min - value + min;
```
br3aker, line 121: Not sure whether this would really affect performance as it depends on the span size but you can cache `2 * min` before the loop:
```cs
// if min is always positive (uint)min * 2 for maximum performance 
// due to binary shift multiplication
int min2 = min + min;
for (int i = 0; i < affectedSize; i++)
{
    var value = span[i];
    if (value < min)
    {
        span[i] = min2 - value;
    }
}
```

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
```diff
@@ -105,5 +81,79 @@ public void Dispose()
...
+                            }
+                        }
+
+                        for (int i = span.Length - affectedSize; i < span.Length; i++)
+                        {
+                            var value = span[i];
+                            if (value > max)
+                            {
+                                span[i] = max - value + max;
```
br3aker, line 130: Same caching of `2 * max` as above.

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
```diff
@@ -105,5 +81,79 @@ public void Dispose()
...
+                        }
+
+                        break;
+                    case BorderWrappingMode.Wrap:
+                        for (int i = 0; i < affectedSize; i++)
+                        {
+                            var value = span[i];
+                            if (value < min)
+                            {
+                                span[i] = max - min + value + 1;
```
br3aker, line 141: Same caching of `max - min + 1` as above.

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
```diff
@@ -105,5 +81,79 @@ public void Dispose()
...
+                            }
+                        }
+
+                        for (int i = span.Length - affectedSize; i < span.Length; i++)
+                        {
+                            var value = span[i];
+                            if (value > max)
+                            {
+                                span[i] = min + value - max - 1;
```
br3aker, line 150: Same caching of `min - max - 1` as above.

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
br3aker: Using `ref`'s instead of span indexing should also yield better performance.

src/ImageSharp/Processing/Processors/Convolution/BorderWrappingMode.cs
```diff
@@ -0,0 +1,21 @@
...
+    public enum BorderWrappingMode : byte
+    {
+        /// <summary>Repeat the border pixel value: aaaaaa|abcdefgh|hhhhhhh</summary>
+        Repeat = 0,
+
+        /// <summary>Take values from the opposite edge: cdefgh|abcdefgh|abcdefg</summary>
+        Wrap = 1,
+
+        /// <summary>Mirror the last few border values: fedcb|abcdefgh|gfedcb</summary>
+        /// <remarks>Please note this mode doe not repeat the very border pixel, as this gives better image quality.</remarks>
```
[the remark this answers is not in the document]
ynse01: Corrected

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
ynse01: Added `min2` variable, as suggested.

Typical `kernelSize` values are 3 and 5. Using references does however reduce readability. 
Do you think it is worth it?

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
ynse01: Done

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
ynse01: Done

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
ynse01: Done, with same variable as above.

src/ImageSharp/Processing/Processors/Convolution/KernelSamplingMap.cs
br3aker: Agreed on readability, didn't know about typical sizes so yeah, it shouldn't matter.

JimBobSquarePants:
Oh wow! This is a pleasant surprise! I never looked at this after refactoring convolution so thought the changes would still be really viral.

src/ImageSharp/Processing/Processors/Convolution/BorderWrappingMode.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>Repeat the border pixel value: aaaaaa|abcdefgh|hhhhhhh</summary>
+        Repeat = 0,
+
+        /// <summary>Take values from the opposite edge: cdefgh|abcdefgh|abcdefg</summary>
+        Wrap = 1,
+
+        /// <summary>Mirror the last few border values: fedcb|abcdefgh|gfedcb</summary>
+        /// <remarks>Please note this mode does not repeat the very border pixel, as this gives better image quality.</remarks>
+        Mirror = 2
```
JimBobSquarePants, line 19: Does this negatively affect the tiling of convoluted images?

ynse01: Not sure if I understand you correctly, but for image tiles, you would use Wrap mode.

To my knowledge Mirror is the most natural mode for photos. If you would repeat the border pixel (a and h in the example) the image might get a small area of equal pixel value, which looks unnatural.

However, for Graphics, especially with solid background, Clamp is the best choice.

Having said that, applying the wrong mode for the subject can have awkward effects.

JimBobSquarePants: Yep, sorry, that makes sense. 

Let's make sure we use Clamp over Repeat for naming to match CUDA definitions. 

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g03e1bbd2c491d809279c7a47e2cd0351

It appears to me that they repeat the border pixel though (I may be reading this wrong)

https://stackoverflow.com/a/27411685/427899

ynse01: The names now match with CUDA's, with the deviation you mention indeed. 

Alternatively, [OpenCV](https://vovkos.github.io/doxyrest-showcase/opencv/sphinx_rtd_theme/enum_cv_BorderTypes.html) has both Mirror modes (which it calls `Reflect` and `Reflect_101`). OpenCV does take the Mirror mode without duplicating the border pixel as the default.

I bet you, if you find a third library, it has its border enum implemented even differently.

I can easily make an extra Mirror mode with the border pixel repeated. Main difficulty there is coming up with a good name for the 2 variants of Mirror.

JimBobSquarePants: Ooh YES! I love it!

ynse01: Bounce mode is in

JimBobSquarePants: Ace. Will review ASAP. Thanks for your patience.

ynse01: Also, please advise how to expose the `BorderWrappingMode` and in which ImageProcessor (or any other public API).

JimBobSquarePants: OK. 

- I don't think the functionality should be added to Resize. 
- We should add both X and Y properties as overloads to our Gaussian Blur/Sharpen methods and Box Blur extensions. I don't think the Bokeh blur functionality can be augmented in that fashion.

ynse01: Ok, done as stated above

ynse01:
Please let me know if there is anything missing in this PR.

JimBobSquarePants:
Will do. We’re currently investigating some access violation issues  #2075 that we need to resolve before we merge any more code.

JimBobSquarePants (approved):
I think we can get this merged now. Thanks for your patience @ynse01 this is an awesome addition to the library!

PR #2062, opened 2022-03-13
Add support for decoding Tiff images with associated alpha data

brianpopow:
### Prerequisites

### Description

This PR adds support for decoding Tiff images with associated alpha data: The alpha data is pre-multiplied.


src/ImageSharp/Formats/Tiff/PhotometricInterpretation/Rgba16161616TiffColor{TPixel}.cs
```diff
@@ -69,6 +83,12 @@ public override void Decode(ReadOnlySpan<byte> data, Buffer2D<TPixel> pixels, in
                         pixelRow,
                         pixelRow.Length);
 
+                    if (hasAssociatedAlpha)
+                    {
+                        PixelOperations<TPixel>.Instance.ToVector4(this.configuration, pixelRow, vectorsSpan);
```
JimBobSquarePants, line 88: Can't we use `PixelOperations<TPixel>.Instance.FromVector4Destructive` passing `PixelConversionModifiers.Premultiply` here? This allows us to use Avx2 unpremultiplication.

JimBobSquarePants, line 88: Can't we use `PixelOperations<TPixel>.Instance.FromVector4Destructive` passing `PixelConversionModifiers.Premultiply` here? This allows us to use Avx2 unpremultiplication.

src/ImageSharp/Formats/Tiff/Utils/TiffUtils.cs
```diff
@@ -126,6 +151,27 @@ public static TPixel ColorScaleTo32Bit<TPixel>(ulong intensity, TPixel color)
             return color;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void UnPremultiplyRow<TPixel>(Span<Vector4> vectors, Span<TPixel> pixelRow, TPixel color)
```
JimBobSquarePants, line 155: This looks like we're replicating existing functionality. We have `PixelOperations` for this.

JimBobSquarePants, line 155: This looks like we're replicating existing functionality. We have `PixelOperations` for this.

src/ImageSharp/Formats/Tiff/Utils/TiffUtils.cs
```diff
@@ -126,6 +151,27 @@ public static TPixel ColorScaleTo32Bit<TPixel>(ulong intensity, TPixel color)
...
+        {
+            for (int x = 0; x < vectors.Length; x++)
+            {
+                Vector4 vec = vectors[x];
+                pixelRow[x] = UnPremultiply(vec, color);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static TPixel UnPremultiply<TPixel>(Vector4 vec, TPixel color)
```
JimBobSquarePants, line 166: Do you call this method when not using  collection? We have `Numerics.UnPremultiply(ref Vector4 vector)`

JimBobSquarePants, line 166: Do you call this method when not using  collection? We have `Numerics.UnPremultiply(ref Vector4 vector)`

JimBobSquarePants:
@brianpopow Do you want this in 2.1? I can hold a release 'til we get it done.

JimBobSquarePants:
@brianpopow I updated your PR to use the built in operations. I had to tweak the tolerance slightly but the output is mostly the same (better alpha blending of edge pixels with `Numerics`).

image

I'm curious as to why there is such a significant difference from the reference as I cannot see an obvious issue.

JimBobSquarePants:
@brianpopow I might have found quite a serious issue with the Tiff decoder I hadn't spotted before. We use `FromVector4` all through the decoder codebase when we should be using `FromScaledVector4`.

JimBobSquarePants (approved):
LGTM 👍

brianpopow:
> @brianpopow Do you want this in 2.1? I can hold a release 'til we get it done.

I dont think this should block 2.1 for this and it can be added for future.

> @brianpopow I might have found quite a serious issue with the Tiff decoder I hadn't spotted before. We use FromVector4 all > through the decoder codebase when we should be using FromScaledVector4. There's some cases in the Bmp decoder also.
> I'm going to update and fix in this PR.

Good spot 👍 

> Can't we use PixelOperations<TPixel>.Instance.FromVector4Destructive passing PixelConversionModifiers.Premultiply here? This allows us to use Avx2 unpremultiplication.

That was also my first thought, but I was getting different results then the reference decoder, so I was leaving it to the manual calculation.

> Can't we use PixelOperations<TPixel>.Instance.FromVector4Destructive passing PixelConversionModifiers.Premultiply here? This allows us to use Avx2 unpremultiplication.

Let me double check that, but last time I checked any difference to the reference decoder was always off by exact 1, which I blamed on floating point inaccuracies.

JimBobSquarePants:
@brianpopow I just did a comparison using the System.Drawing reference decoder and the tests pass with the original comparison tolerance of .0.001F. This is definitely good to go.

PR #2073, opened 2022-03-24
Add support for decoding jpeg's with arithmetic coding

brianpopow:
### Prerequisites

### Description

This PR adds support for decoding jpeg's with arithmetic coding.

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticScanDecoder.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticScanDecoder.cs
@@ -57,9 +57,6 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
 
         private readonly CancellationToken cancellationToken;
 
-        // Uses C# compiler's optimization to refer to static data segement directly, without any further allocation.
-        private static ReadOnlySpan<byte> FixedBin => new byte[] { 113, 0, 0, 0 };
-
         private static readonly int[] ArithmeticTable =
         {
             Pack(0x5a1d,   1,   1, 1),
@@ -199,6 +196,9 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
             this.a = 0;
             this.ct = -16; // Force reading 2 initial bytes to fill C.
         }
+
+        // Uses C# compiler's optimization to refer to static data segment directly, without any further allocation.
+        private static ReadOnlySpan<byte> FixedBin => new byte[] { 113, 0, 0, 0 };
 
         /// <inheritdoc/>
         public int ResetInterval
@@ -916,7 +916,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
 
                     if (this.DecodeBinaryDecision(ref reader, ref Unsafe.Add(ref st, 1)) != 0)
                     {
-                        bool flag = this.DecodeBinaryDecision(ref reader, ref this.GetFixedBinReference()) != 0)
+                        bool flag = this.DecodeBinaryDecision(ref reader, ref this.GetFixedBinReference()) != 0;
                         coef = (short)(coef + (flag ? m1 : p1));
 
                         break;
--- a/src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticScanDecoder.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticScanDecoder.cs
@@ -57,7 +57,8 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
 
         private readonly CancellationToken cancellationToken;
 
-        private readonly byte[] fixedBin = { 113, 0, 0, 0 };
+        // Uses C# compiler's optimization to refer to static data segement directly, without any further allocation.
+        private static ReadOnlySpan<byte> FixedBin => new byte[] { 113, 0, 0, 0 };
 
         private static readonly int[] ArithmeticTable =
         {
@@ -233,7 +234,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
             }
         }
 
-        private ref byte GetFixedBinReference() => ref this.fixedBin[0];
+        private ref byte GetFixedBinReference() => ref MemoryMarshal.GetReference(FixedBin);
 
         /// <summary>
         /// Decodes the entropy coded data.
@@ -580,8 +581,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
                 for (int i = 0; i < mcusPerLine; i++)
                 {
                     // Scan an interleaved mcu... process components in order.
-                    int mcuRow = mcu / mcusPerLine;
-                    int mcuCol = mcu % mcusPerLine;
+                    int mcuRow = Math.DivRem(mcu, mcusPerLine, out int mcuCol);
                     for (int k = 0; k < this.scanComponentCount; k++)
                     {
                         int order = this.frame.ComponentOrder[k];
@@ -908,14 +908,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
                     {
                         if (this.DecodeBinaryDecision(ref reader, ref Unsafe.Add(ref st, 2)) != 0)
                         {
-                            if (coef < 0)
-                            {
-                                coef = (short)(coef + m1);
-                            }
-                            else
-                            {
-                                coef = (short)(coef + p1);
-                            }
+                            coef = (short)(coef + (coef < 0 ? m1 : p1));
                         }
 
                         break;
@@ -923,14 +916,8 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
 
                     if (this.DecodeBinaryDecision(ref reader, ref Unsafe.Add(ref st, 1)) != 0)
                     {
-                        if (this.DecodeBinaryDecision(ref reader, ref this.GetFixedBinReference()) != 0)
-                        {
-                            coef = (short)(coef + m1);
-                        }
-                        else
-                        {
-                            coef = (short)(coef + p1);
-                        }
+                        bool flag = this.DecodeBinaryDecision(ref reader, ref this.GetFixedBinReference()) != 0)
+                        coef = (short)(coef + (flag ? m1 : p1));
 
                         break;
                     }
--- a/src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticScanDecoder.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticScanDecoder.cs
@@ -451,6 +451,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
 
                         // Scan out an mcu's worth of this component; that's just determined
                         // by the basic H and V specified for the component.
+                        int mcuColMulh = mcuCol * h;
                         for (int y = 0; y < v; y++)
                         {
                             Span<Block8x8> blockSpan = component.SpectralBlocks.DangerousGetRowSpan(y);
@@ -466,11 +467,11 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
                                     return;
                                 }
 
-                                int blockCol = (mcuCol * h) + x;
+                                int blockCol = mcuColMulh + x;
 
                                 this.DecodeBlockBaseline(
                                     component,
-                                    ref Unsafe.Add(ref blockRef, blockCol),
+                                    ref Unsafe.Add(ref blockRef, (nint)(uint)blockCol),
                                     ref acDecodingTable,
                                     ref dcDecodingTable);
                             }
@@ -521,7 +522,7 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
 
                         this.DecodeBlockBaseline(
                             component,
-                            ref Unsafe.Add(ref blockRef, k),
+                            ref Unsafe.Add(ref blockRef, (nint)(uint)k),
                             ref acDecodingTable,
                             ref dcDecodingTable);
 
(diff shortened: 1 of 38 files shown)


src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticStatistics.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System;
+
+namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
+{
+    internal class ArithmeticStatistics
```
gfoidl, line 8: Can this be written as
```c#
internal unsafe struct ArithmeticStatistics
{
    private fixed byte statistics[256];

    public ArithmeticStatistics(bool dc, int identifier)
    {
        this.IsDcStatistics = dc;
        this.Identifier = identifier;
    }

    public bool IsDcStatistics { get; private set; }

    public int Identifier { get; private set; }

    public ref byte GetReference() => ref this.statistics[0];

    public ref byte GetReference(int offset) => ref this.statistics[(uint)offset];

    public void Reset()
    {
        Unsafe.InitBlockUnaligned(ref this.GetReference(), 0x00, (uint)(this.IsDcStatistics ? 64 : 256));
    }
}
```
?
* avoids the additional allocation for each stats-object
* downside: size is _fixed_ to 256, so potentially the list will be larger
* [codegen](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNXjxldskht1zYAZjAa4MUDmAwMAglF4YAFgL5hZGbH0vLcNADeNAyhDGJOCj7mJrwIMAAmDMAAnhjmlj68fmC4ANqkAKxIALoa1CFhxADMDk6u7speWTm4ABTAEBAGCWBoDNp2vAkwPLyxMgCUlaHB1GELDK7ZLACSuAAint6+HrgMALwMveWLYcu4ayNjE1CHA9d8t6dhAL6a81W1nd0M61vNXb+BiBBgAcxgGDU4Ui0QskOh7wqn1CNQGPD+j3GvBkIPBCJhvCi6XhUIYSJmTFqsBMyTS5gA4pC2DAzLAuCo2pNDgA+Bg0pYuFaZIG5PJ0MofBZogWpElMjAstmjTmDBgQEwmXCQ7kHPkCi4sEXZPZ5NocQaTDVayGS5HS2rEFAMFnajBcylzM6hACqXGMZjWXGcACFJBAwABrP36XhgriJNoGoWXBVKmQqmBc/p0BB0Oj9c2WtqG/7bFp7BgAfgYqAYIAYRSQk0mL1CSNeQA) is better than [with the class](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKAEsA7DGKe7AGwAIyOBBKWjAAt8MDLTABlDNjG4xYXDQ7KOAbyUrNAB34A3GTA6xsAEwj02ATw7BLTANoBdDnJm0543AG4aGzcuIAZl5+IRF5KTcPBQAKYAgIThMwNA4GDDSTGEZaADNaZgBKP391an8KjiF3ADoASVwAEUlpWXlcDgBeDmSfcsrNatx6rJz85i7M7LFxqD6BwcFa1zbPSeSOAH4OehgAdxs7GHtUZxAd/cOHUgBWJEd5yoBfX36KoJsEzgbmyNWFNQcADmoi8HB0tH0TBcoI4L2oJU0H3SHDqoxmBSggJBGDBEKhhlwsPhiJUH1guSuhgA4qIAEowXLMbJgGAxQpdAB8RkZVSWwxW7na9gADA9XgNybzbNDaRgGUzYPRWTEURBcrkiRgOZ1uRS+ctWkLPPZ1ZrROKEW9/B9iCgOAyteyuQaBUbosMeLgJFpsPR2TUAMJsGDYKDsx7KJ5AA=) especially for `GetReference` and `Reset` (which are the hottest methods here?)

src/ImageSharp/Formats/Jpeg/Components/Decoder/IJpegComponent.cs
```diff
@@ -10,6 +10,11 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder
     /// </summary>
     internal interface IJpegComponent
     {
+        /// <summary>
+        /// Gets the component id.
+        /// </summary>
+        public byte Id { get; }
```
gfoidl, line 16: The existing interface members don't have the access modifier specified.
It's not strictly needed, but C# allows this (since C# 8 I think).

At least it should be consistent withing this file.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -1036,17 +1104,21 @@ private void ProcessStartOfFrameMarker(BufferedReadStream stream, int remaining,
                 JpegThrowHelper.ThrowInvalidImageContentException("Multiple SOF markers. Only single frame jpegs supported.");
             }
 
-            // Read initial marker definitions
+            // Read initial marker definitions.
             const int length = 6;
-            stream.Read(this.temp, 0, length);
+            int bytesRead = stream.Read(this.temp, 0, length);
```
gfoidl, line 1109: To double check: stream (in general) isn't guaranteed to read exactly the requested length, it may read up to length bytes. So this can be a latent bug, and usually there's a `while (bytesRead < length)`-loop.

I see the check below, and assume that BufferedReadStream's implemented takes this into mind (via CopyTo over there?).

So this is more or less a question to avoid a latent, nasty to find bug.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -1055,18 +1127,18 @@ private void ProcessStartOfFrameMarker(BufferedReadStream stream, int remaining,
             // 2 byte: Width
             int frameWidth = (this.temp[3] << 8) | this.temp[4];
 
-            // Validate: width/height > 0 (they are upper-bounded by 2 byte max value so no need to check that)
+            // Validate: width/height > 0 (they are upper-bounded by 2 byte max value so no need to check that).
             if (frameHeight == 0 || frameWidth == 0)
```
gfoidl, line 1131:
```suggestion
            if ((frameHeight | frameWidth) == 0)
```
produces [less code](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYmpBgDEIEAELYoNAN40GtpruDHJh4+QAUvHgwBmUbAIAJGF4AcwALDDQGTwwfPwEAdV4AEwwwgEoGAF4APjj/GCDQiOyshkZCQnzElLTS8o1qO3sGRwhnIwhSDy9fAqLwyOje+Jgk1IzsvLc+wODBhirZsdrJrLK6RoBfIA) and avoids a branch.

Note: current JIT will do this optimizaiton for use, but older runtimes miss that optimization.

gfoidl (commented):
I always enjoy reading such fine code.
At the same time I'm sorry for so much comments.

br3aker:
I know that I'm asking for too much but can we postpone this PR from merging for a bit?
There's so much virtual types existing in current jpeg decoder and new ones from this PR which already led to explicit casting. I'm working on the downscaling decoding for jpeg which would also need some sort of a virtuality for both the huffman decoder and component post processing which combined can cause a mess IMO.

I'll push some fixes on your branch if you do't mind.

brianpopow:
@gfoidl thanks for your review, its always appreciated!

@br3aker:

> I know that I'm asking for too much but can we postpone this PR from merging for a bit?

Its not a problem, there is no need to rush this PR.

> I'll push some fixes on your branch if you don't mind.

I would not mind, if you push directly to this branch, your help is very welcome, but I dont think you have permission to do so.
I am pretty sure its somehow possible to give you those write access to this branch, but I did not see how at the moment.
You will have to create a regular PR with this branch as the target branch instead of `main`.

br3aker:
Code itself is very good, it's just those tiny little spots I'd want to fix but they do not block merging so it's okay. I will create a PR with general refactoring of the decoder part.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
JimBobSquarePants: Yeah `BufferedReadStream` uses a while loop to read internally.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticScanDecoder.cs
```diff
@@ -0,0 +1,1250 @@
...
+                                    // so we need to decode what's left and return (or maybe throw?)
+                                    this.spectralConverter.ConvertStrideBaseline();
+                                    return;
+                                }
+
+                                int blockCol = (mcuCol * h) + x;
+
+                                this.DecodeBlockBaseline(
+                                    component,
+                                    ref Unsafe.Add(ref blockRef, blockCol),
```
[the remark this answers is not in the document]
brianpopow: @gfoidl I dont understand the suggestion: How can I use `uint` here, when `Unsafe.Add` expects `int`?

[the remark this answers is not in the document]
gfoidl: Here it's for the overload `Unsafe.Add<T>(ref T source, nuint elementOffset)` and uses the implicit conversion from `uint` -> `nuint`.
So the long form of the suggestion is
 ```suggestion
                                    ref Unsafe.Add(ref blockRef, (nuint)(uint)blockCol),
```
but as the compiler does the second conversion for us, I've shortened it to
```suggestion
                                    ref Unsafe.Add(ref blockRef, (uint)blockCol),
```

---

In essence, and to avoid the `movsxd` (the sign extending move), we need to convert the `int` to any of `IntPtr` / `nint` or `UIntPtr` / `nuint`. As first step here is to prove to the JIT that there's only positive numbers by the `uint`-cast, which elides the `movsxd` emitting. 
The actual target type (`nint` or `nuint` respectively there IntPtr base variants) doesn't really matter then.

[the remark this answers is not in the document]
brianpopow: Thank you for the explanation. Now I see why it does not work: The uint overload is only available on .net6.0, so cant use that here.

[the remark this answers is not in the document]
gfoidl: Ah, sorry, didn't check that. The use `(nint)(uint)` for the cast.

src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticScanDecoder.cs
```diff
@@ -0,0 +1,1250 @@
...
+            {
+                var component = this.components[i] as ArithmeticDecodingComponent;
+                this.dcDecodingTables[i] = this.GetArithmeticTable(arithmeticDecodingTables, true, component.DcTableId);
+                component.DcStatistics = this.CreateOrGetStatisticsBin(true, component.DcTableId);
+                this.acDecodingTables[i] = this.GetArithmeticTable(arithmeticDecodingTables, false, component.AcTableId);
+                component.AcStatistics = this.CreateOrGetStatisticsBin(false, component.AcTableId);
+            }
+        }
+
+        private ref byte GetFixedBinReference() => ref this.fixedBin[0];
```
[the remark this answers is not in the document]
brianpopow: For some reason this change introduced a access violation issue. I cant spot why that is:

```
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```

See CI output for full stack trace.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -97,7 +98,17 @@ internal sealed class JpegDecoderCore : IRawJpegData, IImageDecoderInternals
...
+
+        /// <summary>
+        /// The arithmetic decoding tables.
+        /// </summary>
+        private List<ArithmeticDecodingTable> arithmeticDecodingTables;
+
+        /// <summary>
+        /// The restart interval.
+        /// </summary>
+        private int? resetInterval;
```
br3aker, line 111: I don't like this arithmetic coding specific property is leaked in the 'master' class, same goes for reset interval which should only be used inside `scanDecoder` but it's a broad question for some further refactoring - I'll work on it right after scaled decoding PR.

src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegFrame.cs
```diff
@@ -86,7 +86,7 @@ public JpegFrame(JpegFileMarker sofMarker, byte precision, int width, int height
         /// <summary>
         /// Gets or sets the frame component collection.
         /// </summary>
-        public JpegComponent[] Components { get; set; }
+        public IJpegComponent[] Components { get; set; }
```
br3aker, line 89: As far as I remember, `IJpegComponent` interface is used in tests for libjpeg output comparison but I think we shouldn't use this interface everywhere - [this leads to virtual calls for every component field getter ](https://sharplab.io/#v2:EYLgHgbALANALiAhgZwLYB8ACAmAjAWAChMBmAAk1wguzIGEyBvIs1i8gSwDs4BTAJwBmiAMa8yASToB7VAAdpXXjzIs2zQmy1lg06QBsyAMX2IA5kzO84AbgC+a1g8KP2NerIVKVISTPmKynCuGtpspDp6hibmltb2rs5arhGU1LoGZADi1jFmABRSnoEqIp4AlGQAvAB8ZGXyAHR5NinkaZGZOXB5+f5eQfUV1XUNcs2mZq2EdkA==).

br3aker (changes requested):
Looks good overall but please revert component type to `JpegComponent` in `JpegDecoderCore`.

br3aker (approved):

br3aker:
@brianpopow hi! Anything blocking this PR from being merged?

brianpopow:
> @brianpopow hi! Anything blocking this PR from being merged?

@br3aker: I think this can be merged now. 
@JimBobSquarePants do you have any objection on merging this?

JimBobSquarePants:
Nope. Rock on

PR #2077, opened 2022-03-27
Add missing decoder sanitation checks

### LINKED ISSUE #2075: Access Violation in 2.1, works fine in 1.04
### Prerequisites

### ImageSharp version

2.1.0

none

Windows 2016

### .NET Framework version

6.0

### Description

We have an ASP.NET 6 app that uses ImageSharp 2.1 to resize images and create thumbnails (and that's it) for UGC. Several times a day our app crashes with this error:

```
Application: w3wp.exe
CoreCLR Version: 6.0.322.12309
.NET Version: 6.0.3
Description: The process was terminated due to an internal error in the .NET Runtime at IP 00007FF932A2A5A9 (00007FF932940000) with exit code 80131506.
```

It's very simple code. Some images throw this uncatchable error, which crashes our entire process.

Reverting back to 1.0.4 seems to help.

### Steps to Reproduce

Unfortunately I'm unable to reliably tell which images caused this, since it's a busy app with millions of HTTP requests. Also I had no time to create process dumps or anything, But I did narrow this down to ImageSharp - reverting nuget from 2.1 to 1.0.4 fixed the issue immediately. So I thought you should know.

PS. Also, posting this is to help someone who's going to google for a solution desperately

### Images

_No response_

JimBobSquarePants:
### Prerequisites

### Description

A PR to add additional sanitation checks with a view to fix #2075 

Any additional sanitation fixes added to other open branches should be added here.

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -722,7 +722,14 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
 
             if (ProfileResolver.IsProfile(this.temp, ProfileResolver.XmpMarker.Slice(0, ExifMarkerLength)))
             {
-                int remainingXmpMarkerBytes = XmpMarkerLength - ExifMarkerLength;
+                const int remainingXmpMarkerBytes = XmpMarkerLength - ExifMarkerLength;
+                if (remaining < remainingXmpMarkerBytes || this.IgnoreMetadata)
+                {
+                    // Skip the application header length.
+                    stream.Skip(remaining);
+                    return;
+                }
+
                 stream.Read(this.temp, ExifMarkerLength, remainingXmpMarkerBytes);
                 remaining -= remainingXmpMarkerBytes;
                 if (ProfileResolver.IsProfile(this.temp, ProfileResolver.XmpMarker))
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -1334,8 +1334,12 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
                 component.ACHuffmanTableId = acTableIndex;
             }
 
-            // 3 bytes: Progressive scan decoding data
-            stream.Read(this.temp, 0, 3);
+            // 3 bytes: Progressive scan decoding data.
+            int bytesRead = stream.Read(this.temp, 0, 3);
+            if (bytesRead != 3)
+            {
+                JpegThrowHelper.ThrowInvalidImageContentException("Not enough data to read progressive scan decoding data");
+            }
 
             int spectralStart = this.temp[0];
             this.scanDecoder.SpectralStart = spectralStart;
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -1362,7 +1362,12 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
         [MethodImpl(InliningOptions.ShortMethod)]
         private ushort ReadUint16(BufferedReadStream stream)
         {
-            stream.Read(this.markerBuffer, 0, 2);
+            int bytesRead = stream.Read(this.markerBuffer, 0, 2);
+            if (bytesRead != 2)
+            {
+                JpegThrowHelper.ThrowInvalidImageContentException("jpeg stream does not contain enough data, could not read ushort.");
+            }
+
             return BinaryPrimitives.ReadUInt16BigEndian(this.markerBuffer);
         }
     }
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -232,7 +232,12 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
             using var stream = new BufferedReadStream(this.Configuration, ms);
 
             // Check for the Start Of Image marker.
-            stream.Read(this.markerBuffer, 0, 2);
+            int bytesRead = stream.Read(this.markerBuffer, 0, 2);
+            if (bytesRead != 2)
+            {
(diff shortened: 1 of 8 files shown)


src/ImageSharp/Formats/Webp/WebpDecoderCore.cs
```diff
@@ -456,7 +462,12 @@ private void ParseOptionalExtendedChunks(WebpChunkType chunkType, WebpFeatures f
                     else
                     {
                         byte[] exifData = new byte[exifChunkSize];
-                        this.currentStream.Read(exifData, 0, (int)exifChunkSize);
+                        bytesRead = this.currentStream.Read(exifData, 0, (int)exifChunkSize);
+                        if (bytesRead != exifChunkSize)
+                        {
+                            WebpThrowHelper.ThrowInvalidImageContentException("Not enough data to read the exif chunk");
```
JimBobSquarePants: Are there rules in WebP about where a these chunks should be. I.E are they always before the image data?

brianpopow: ICCP and ANIM chunks should always come before the image data, EXIF and XMP chunks should come after the image data.

https://developers.google.com/speed/webp/docs/riff_container#extended_file_format

JimBobSquarePants: If after, I don't think we should throw then and skip instead. I would consider those non-critical.

br3aker:
I would suggest another sanitation check for jpeg decoder in general. Each marker has specific predefined size which we already have in each marker parser as a parameter - `remaining`. Checking whether stream has enough bytes via `stream.Length - stream.Position >= remaining` would eliminate any stream read checks inside marker parsers and would make code cleaner imo.

I can't contribute to this branch directly so I can make this as a separate PR.

JimBobSquarePants:
> Checking whether stream has enough bytes via `stream.Length - stream.Position >= remaining` would eliminate any stream read checks inside marker parsers and would make code cleaner imo.

I would agree. If you have the time to do that'd very much appreciate it.

br3aker:
Sorry for late response, been studying scaled IDCT for other PR. Will try to do a PR for jpeg markers on this week.

JimBobSquarePants:
I looks increasingly unlikely that we'll get any test images for this but I think we should still push on with the sanitation code. 
I'm going to create a release/2.1.1 branch and open a second PR from this branch so we can push on with V3.

br3aker:
Actually there's kind of a problem in current jpeg decoder. It shouldn't be a cause of mentioned issue but it's still a bug

Current decoding pipeline:

```
1. Gather jpeg info (components, tables, frame size, etc.) <- pixel buffer is allocated here
2. For each scan:
       For each mcu row:
           Decode: huffman -> spectral -> rgb -> TPixel
3. Create Image object using allocated pixel buffer
```

Any exceptions between step 1 and 3 would cause allocated pixel buffer to be gc'd leading to a false positive memory leak reported by memory diagnostics. Any invalid marker between scans can cause this, it shouldn't crash the entire process or even lead to an actual memory leak but it's still a thing. 

I'll create a separate issue this week, posting this here just to let you know.

br3aker:
> I would suggest another sanitation check for jpeg decoder in general. Each marker has specific predefined size which we already have in each marker parser as a parameter - remaining. Checking whether stream has enough bytes via stream.Length - stream.Position >= remaining would eliminate any stream read checks inside marker parsers and would make code cleaner imo.

Extra jpeg sanity checks implemented at https://github.com/SixLabors/ImageSharp/pull/2084.

JimBobSquarePants:
@SixLabors/core I'm marking this as ready to review. I think with these checks and #2084 we've probably handled any potential issues during decode that trigger #2075

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -232,15 +232,25 @@ public void LoadTables(byte[] tableBytes, HuffmanScanDecoder huffmanScanDecoder)
             using var stream = new BufferedReadStream(this.Configuration, ms);
 
             // Check for the Start Of Image marker.
-            stream.Read(this.markerBuffer, 0, 2);
+            int bytesRead = stream.Read(this.markerBuffer, 0, 2);
+            if (bytesRead != 2)
+            {
+                JpegThrowHelper.ThrowInvalidImageContentException("Not enough data to read the SOI marker");
+            }
```
br3aker, line 239: I think we can remove this and similar checks from other marker parsing methods as it's now guaranteed that stream has at least `remaining` bytes available.

JimBobSquarePants: Ah... That method is actually called from TIFF only.

br3aker: But we still can transform 2 read checks to 1 `AvailableBytes` check :)
I can't comment on non diff lines but that method creates stream:

```cs
using var ms = new MemoryStream(tableBytes);
```

and then do 2 reads of 2 bytes with sequential if checks.

We can do this check:

```cs
if(tableBytes.Length < 4)  throw ...
```

before even creating a stream and remove 2 checks afters reads as those would always be true.

br3aker:
Sorry, I forgot to alter TIFF jpeg loading method with here:

https://github.com/SixLabors/ImageSharp/blob/c5f14f751d78ae4a0942438a0b081c504aac0284/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs#L257-L264

It should be this:

```cs
int markerContentByteSize = this.ReadUint16(stream) - 2;

// Check whether stream actually has enought bytes to read
// markerContentByteSize is always positive so we cast
// to uint to avoid sign extension
if (stream.RemainingBytes < (uint)markerContentByteSize)
{
    JpegThrowHelper.ThrowNotEnoughBytesForMarker(fileMarker.Marker);
}

swtich(fileMarker.Marker)
{
    // ...
}
```

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
JimBobSquarePants: Cool. Will have a look.

JimBobSquarePants:
Peeps I'm gonna merge this. I want a build that can be tested in #2075 and it's blocking all other progress.

Issue #2080, opened 2022-04-09
Memory leak in PNG decoder


miere43:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

2.1.0

### Other ImageSharp packages and versions

None

### Environment (Operating system, version and so on)

Windows 10 Pro, version 21H1, OS build 19043.1586

### .NET Framework version

6.0.3

### Description

Using `Image.Load` with PNG image leaks memory after calling `Dispose` on that image.

### Steps to Reproduce

Code for reproduction:
```csharp
using System.Diagnostics;
using SixLabors.ImageSharp;
using SixLabors.ImageSharp.Diagnostics;

MemoryDiagnostics.UndisposedAllocation += stackTrace =>
{
    // This callback is never called for some reason.
    Console.WriteLine(stackTrace);
    Debugger.Break();
};

static void Test()
{
    using var stream = File.OpenRead("C:\image.png");
    using var image = Image.Load(stream);
}

Test();

Console.WriteLine($"Leaks: {MemoryDiagnostics.TotalUndisposedAllocationCount}");
Console.ReadLine();
```

Program output:
```
Leaks: 1
```

UndisposedAllocation callback doesn't work for some reason, this is one of stack traces from my project:
```
   at System.Environment.get_StackTrace()
   at SixLabors.ImageSharp.Memory.Internals.SharedArrayPoolBuffer`1..ctor(Int32 lengthInElements)
   at SixLabors.ImageSharp.Memory.UniformUnmanagedMemoryPoolMemoryAllocator.Allocate[T](Int32 length, AllocationOptions options)
   at SixLabors.ImageSharp.Formats.Png.PngDecoderCore.TryReadChunk(PngChunk& chunk)
   at SixLabors.ImageSharp.Formats.Png.PngDecoderCore.ReadNextDataChunk()
   at SixLabors.ImageSharp.Compression.Zlib.ZlibInflateStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.Compression.DeflateStream.ReadCore(Span`1 buffer)
[...]

PR #2081, opened 2022-04-09
Fix chunk data memory leak when decoding PNG

miere43:
### Prerequisites

### Description

Added missing `PngChunk.Dispose` call to avoid memory leak in `PngDecoderCore`

All relevant information in issue #2080

### DIFF
--- a/tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Threading.Tasks;
 using Microsoft.DotNet.RemoteExecutor;
 
+using SixLabors.ImageSharp.Diagnostics;
 using SixLabors.ImageSharp.Formats.Png;
 using SixLabors.ImageSharp.Memory;
 using SixLabors.ImageSharp.PixelFormats;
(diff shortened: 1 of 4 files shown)


tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs
```diff
@@ -491,6 +492,34 @@ public void Issue410_MalformedApplePng<TPixel>(TestImageProvider<TPixel> provide
...
+                TestImageProvider<Rgba32> provider
+                    = BasicSerializer.Deserialize<TestImageProvider<Rgba32>>(providerDump);
+
+                provider.Configuration = Configuration.CreateDefaultInstance();
+
+                // Replace memory allocator to bypass image cache.
+                // Allocator must be unmanaged to enable allocation tracing.
+                provider.Configuration.MemoryAllocator = new UniformUnmanagedMemoryPoolMemoryAllocator(null);
+
+                int undisposedAllocationCount = MemoryDiagnostics.TotalUndisposedAllocationCount;
```
tocsoft, line 511: I'm fairly confident trying to write consuming `MemoryDiagnostics.TotalUndisposedAllocationCount` will produce flaky tests, I believe the tests run in parallel and as such other tests could be running at the same time as this one and thus can also be influencing this counter mid test.

I feel we might need to add something to the MemoryAllocator itself to track stats per Manager to make this workable or we might need to alter the MemoryDiagnostics to allow resetting/tracking via an AsyncLocal to allow isolating stats per test.

miere43: Test code runs using `RemoteExecutor`, which I believe creates new process to execute code. So it won't be flaky.

tocsoft: ah yes your right... I missed the `RemoteExecutor` usage... ignore my concerns I can't see how other tests will influence them. 👍

tocsoft:
@miere43 Thanks a lot for this fix.

Turns out most of our test images suffer from this same issue, I have merged your fix (but not your tests) into #2082 so your fix has been merged via that PR instead so I'm going to close this one.

PR #2082, opened 2022-04-09
Memory allocation allocation tracker test helper

tocsoft:
### Prerequisites

### Description

This adds some new test helpers APIs + updates the `MemoryDiagnostics` class to consistently track the memory buffers per test.

Also adds 2 helpers in to the tests 
1. `ValidateDisposedMemoryAllocationsAttribute` - This attributes a can be applied to test classes and `Fact`/`Theory`s directly to automatically apply the memory application tracking and validation.
2. `MemoryAllocatorValidator` static class can be used in places the attribute is not appropriate, i.e. inside Remote Executors.

I've also applied some fixes targeting the Decoder tests

this includes the fix from #2081 (which I will merge into this PR so we don't loose attribute of the fix) if we decide this approach is suitable...otherwise we should just merge that one.

### DIFF
--- a/src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
+++ b/src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
@@ -15,10 +15,8 @@ namespace SixLabors.ImageSharp.Diagnostics
     /// </summary>
     public static class MemoryDiagnostics
     {
-        private static int totalUndisposedAllocationCount;
-
-        private static UndisposedAllocationDelegate undisposedAllocation;
-        private static int undisposedAllocationSubscriptionCounter;
+        internal static readonly InteralMemoryDiagnostics Default = new();
+        private static AsyncLocal<InteralMemoryDiagnostics> localInstance = null;
         private static readonly object SyncRoot = new();
 
         /// <summary>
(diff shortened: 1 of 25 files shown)


tocsoft:
@antonfirsov It would be useful if you could confirm I'm not doing anything too critical/stupid fiddling with the `MemoryDiagnostics` class.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -187,20 +187,27 @@ public Image<TPixel> Decode<TPixel>(BufferedReadStream stream, CancellationToken
...
+                return new Image<TPixel>(
+                    this.Configuration,
+                    spectralConverter.GetPixelBuffer(cancellationToken),
+                    this.Metadata);
+            }
+            catch
+            {
+                this.Frame?.Dispose();
+                throw;
+            }
```
br3aker, line 210: There's no need to do a try-catch here I believe as it is handled by the caller here:

https://github.com/SixLabors/ImageSharp/blob/136cc143bba2fca255328751dd4580e4f8d413c3/src/ImageSharp/Formats/Jpeg/JpegDecoder.cs#L19-L26

tocsoft: yeah looks like you are right... I think I must have added it on first pass trying to find the leaky code.

br3aker: It's very tricky because jpeg is the only decoder which has a disposable internal `DecoderCore` implementation :)

JimBobSquarePants: Should we make it so it’s doesn’t implement IDisposable?

br3aker: Hmm... Considering it's only used for `Frame` field which is a class field for the sake of not passing it to a bunch of methods we can make it non-disposable for consistency.

If you do so, try-catch must be replaced by try-finally or `Frame` would leak in a no-exception scenario. Also don't forget about `Identify` method - it also creates a `Frame` field so it should have a proper try-finally too.

JimBobSquarePants (approved):
Ah this is fantastic! Exactly what we need. 😍

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
JimBobSquarePants: Thanks. Ace, that’s what I thought. That can be a separate PR though.

br3aker (approved):
`ValidateDisposedMemoryAllocationsAttribute` is a blessing, also cool that current coverage reports no leaks :sunglasses:

tocsoft:
> `ValidateDisposedMemoryAllocationsAttribute` is a blessing, also cool that current coverage reports no leaks 😎

no leaks (anymore) in the few tests I added it to anyway :D 

There will be a need to follow up adding this to a lot more of the test... I didn't want to blow out this PR fixing leaks hiding the framework changes.

Despite this being approved I think its worth leave this one open a bit longer before merging to give anyone with more knowledge on how my changed to the `MemoryDiagnostics` class might impact perf.

src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
```diff
@@ -47,18 +49,42 @@ public static class MemoryDiagnostics
...
+
+                localInstance.Value = value;
+            }
+        }
+
         /// <summary>
         /// Gets a value indicating the total number of memory resource objects leaked to the finalizer.
         /// </summary>
-        public static int TotalUndisposedAllocationCount => totalUndisposedAllocationCount;
+        public static int TotalUndisposedAllocationCount => Current.TotalUndisposedAllocationCount;
```
antonfirsov, line 81: `TotalUndisposedAllocationCount` was intended to report global allocation count. If I get the docs of [`AsyncLocal<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=net-6.0) right, the PR changes the semantics, so it becomes specific to an asynchronous execution flow, and will reset outside of the flow. I don't think this public behavior change is beneficial for the users of `MemoryDiagnostics.TotalUndisposedAllocationCount `. We need a separate API for `MemoryAllocatorValidator`.

src/ImageSharp/Formats/Tiff/Compression/Decompressors/JpegTiffCompression.cs
```diff
@@ -65,7 +65,8 @@ protected override void Decompress(BufferedReadStream stream, int byteCount, int
                         jpegDecoder.ParseStream(stream, scanDecoderGray, CancellationToken.None);
 
                         // TODO: Should we pass through the CancellationToken from the tiff decoder?
-                        CopyImageBytesToBuffer(buffer, spectralConverterGray.GetPixelBuffer(CancellationToken.None));
+                        using var decompressedBuffer = spectralConverterGray.GetPixelBuffer(CancellationToken.None);
+                        CopyImageBytesToBuffer(buffer, decompressedBuffer);
```
antonfirsov, line 69: Do we also have separate PR-s for the Jpeg+Tiff+WebP+Bmp fixes or just for PNG?

tests/ImageSharp.Tests/MemoryAllocatorValidator.cs
```diff
@@ -0,0 +1,43 @@
...
+using System;
+using System.Diagnostics;
+using SixLabors.ImageSharp.Diagnostics;
+using Xunit;
+
+namespace SixLabors.ImageSharp.Tests
+{
+    public static class MemoryAllocatorValidator
+    {
+        public static IDisposable MonitorAllocations(int max = 0)
```
antonfirsov, line 13:
```suggestion
        public static TestMemoryAllocatorDisposable MonitorAllocations(int max = 0)
```
No point for `TestMemoryAllocatorDisposable` being a struct, if we box it into `IDisposable` in the end.

tests/ImageSharp.Tests/MemoryAllocatorValidator.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            MemoryDiagnostics.Current = new();
+            return new TestMemoryAllocatorDisposable(max);
+        }
+
+        public static void ValidateAllocation(int max = 0)
```
antonfirsov, line 19: Is there a use case for `max > 0`? Do we want to allow any test with monitoring enabled to leak *some* memory?

antonfirsov (changes requested):
We need to separate the `AsyncLocal` test-scoped validation from the global `MemoryDiagnostics.TotalUndisposedAllocationCount` counter.

antonfirsov:
What I would do is to expose (for now internal) `BufferAllocated` / `BufferDestroyed` events and do something like:

```C#
internal static void IncrementTotalUndisposedAllocationCount()
{
    Interlocked.Increment(ref totalUndisposedAllocationCount);
    bufferAllocatedBackingDelegate?.Invoke();
}
```

Then move the `AsyncLocal` utility to test code and control it by subscribing to the events above.

src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
tocsoft: It way its setup it will only report the local context when its been overwritten in tests (internal api)   by default the backing fields are backs by the `MemoryDiagnostics.Default` static instance which will be used in production usage outside of testing.

tocsoft:
> What I would do is to expose (for now internal) `BufferAllocated` / `BufferDestroyed` events and do something like:
> 
> ```cs
> internal static void IncrementTotalUndisposedAllocationCount()
> {
>     Interlocked.Increment(ref totalUndisposedAllocationCount);
>     bufferAllocatedBackingDelegate?.Invoke();
> }
> ```
> 
> Then move the `AsyncLocal` utility to test code and control it by subscribing to the events above.

I think I like that better its less impactful on the prod code.... i'll have a play and see how feels.

src/ImageSharp/Diagnostics/MemoryDiagnostics.cs
antonfirsov: I see now, I missed the purpose of the branching. Though probably still better to keep the internal API as minimalistic and non-intrusive as possible.

tests/ImageSharp.Tests/MemoryAllocatorValidator.cs
```diff
@@ -0,0 +1,77 @@
...
+        public static void ValidateAllocations(int expectedAllocationCount = 0)
+            => LocalInstance.Value?.Validate(expectedAllocationCount);
+
+        public class TestMemoryDiagnostics : IDisposable
+        {
+            public int TotalAllocated { get; set; }
+
+            public int TotalRemainingAllocated { get; set; }
+
+            public void Validate(int expectedAllocationCount)
```
antonfirsov, line 60: `expectedAllocationCount` seems to be always 0.

tests/ImageSharp.Tests/MemoryAllocatorValidator.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public static void StopMonitoringAllocations() => LocalInstance.Value = null;
+
+        public static void ValidateAllocations(int expectedAllocationCount = 0)
+            => LocalInstance.Value?.Validate(expectedAllocationCount);
+
+        public class TestMemoryDiagnostics : IDisposable
+        {
+            public int TotalAllocated { get; set; }
```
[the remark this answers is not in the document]
tocsoft: I was pre-empting being useful for tests that want to track/verify total number of allocated buffers not just retained buffers... but your right currently not being used.

tests/ImageSharp.Tests/MemoryAllocatorValidator.cs
tocsoft: In our current tests your are right, but I was envisioning use cases where we might want to want to validate before all buffers are returned and this is to allow for testing a specific/expected number of buffers are currently allocated still.

tests/ImageSharp.Tests/MemoryAllocatorValidator.cs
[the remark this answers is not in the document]
antonfirsov: I'm generally reluctant to add unused/untested features (= stuff to maintain) even to test utility code. When actual need appears, we can easily add them.

But it's really a nit, so I'm not going to block the PR on this.

antonfirsov (approved):

Issue #2087, opened 2022-04-13
ArgumentNullException thrown when resizing jpg image


saintarian:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

2.1.0

### Other ImageSharp packages and versions

None

### Environment (Operating system, version and so on)

MacOS 12.1 on M1 chip

### .NET Framework version

6.0

### Description

Running the xunit test below produces the following exception:

```
[xUnit.net 00:00:00.93]     ImageSharpResizeTest.TestExifIssue [FAIL]
  Failed ImageSharpResizeTest.TestExifIssue [454 ms]
  Error Message:
   System.ArgumentNullException : Array cannot be null. (Parameter 'bytes')
  Stack Trace:
     at System.Text.UTF8Encoding.GetBytes(Char* chars, Int32 charCount, Byte* bytes, Int32 byteCount)
   at SixLabors.ImageSharp.Metadata.Profiles.Exif.ExifEncodedStringHelpers.Write(Encoding encoding, String value, Span`1 destination)
   at SixLabors.ImageSharp.Metadata.Profiles.Exif.ExifEncodedStringHelpers.Write(EncodedString encodedString, Span`1 destination)
   at SixLabors.ImageSharp.Metadata.Profiles.Exif.ExifWriter.WriteValue(IExifValue exifValue, Span`1 destination, Int32 offset)
   at SixLabors.ImageSharp.Metadata.Profiles.Exif.ExifWriter.WriteData(UInt32 startIndex, List`1 values, Span`1 destination, Int32 offset)
   at SixLabors.ImageSharp.Metadata.Profiles.Exif.ExifWriter.GetData()
   at SixLabors.ImageSharp.Metadata.Profiles.Exif.ExifProfile.ToByteArray()
   at SixLabors.ImageSharp.Formats.Jpeg.JpegEncoderCore.WriteExifProfile(ExifProfile exifProfile)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegEncoderCore.WriteProfiles(ImageMetadata metadata)
[...]

PR #2059, opened 2022-03-11
Fix null reference exception on Exif clone of null ExifArrayValue

ynse01:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description
This PR fixes #2056 

When an `ExifArrayValue` has a `null` value, don't bother to clone that value. Just write the `null` value in the clone also.

PR #2088, opened 2022-04-14
Fixed string exif value corner case exception

br3aker:
### Prerequisites

### Description
Closes https://github.com/SixLabors/ImageSharp/issues/2087.

Given image from linked issue contains empty user comment tag which is written last. Allocated buffer is passed as `Span` which is sliced during different exif values rendering to the given buffer. This exact situation led to copying zero bytes to zero-length span aaaaand looks like .net treats empty spans with valid memory pointer as null pointer thus mentioned `ArgumentNullException`.

Fix is fairly simple: use spans instead of working with raw pointers without any fixed statements (love this performance side effect) and unsafe stuff, `Encoding.GetBytes(ReadOnlySpan, Span)` respects empty spans.

~~P.S.
Waiting for issue image usage approval from the issue author.~~

### DIFF
--- a/src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
+++ b/src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
@@ -79,16 +79,8 @@ namespace SixLabors.ImageSharp.Metadata.Profiles.Exif
             return CharacterCodeBytesLength + count;
         }
 
-        public static unsafe int Write(Encoding encoding, string value, Span<byte> destination)
-        {
-            fixed (char* c = value)
-            {
-                fixed (byte* b = destination)
-                {
-                    return encoding.GetBytes(c, value.Length, b, destination.Length);
-                }
-            }
-        }
+        public static unsafe int Write(Encoding encoding, string value, Span<byte> destination) =>
+            encoding.GetBytes(value.AsSpan(), destination);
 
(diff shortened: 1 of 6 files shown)


br3aker:
@JimBobSquarePants can you please advice where should I place the image from this issue?
I don't see any similar tests.

antonfirsov:
@br3aker I think adding yet another entry to [BaselineTestJpegs](https://github.com/SixLabors/ImageSharp/blob/e54219acec10d2d2dc247ed5fd4230204a89c07b/tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.Images.cs#L12) or [ProgressiveTestJpegs](https://github.com/SixLabors/ImageSharp/blob/e54219acec10d2d2dc247ed5fd4230204a89c07b/tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.Images.cs#L50) is good enough, we have a bunch of Exif regression tests there already.

JimBobSquarePants:
Yep, just add it to the collection there

br3aker:
Problem is that this image decodes correctly, it throws on encoding during writing exif profile.

JimBobSquarePants:
It’s cool, we did the same here. 

https://github.com/SixLabors/ImageSharp/pull/2059

br3aker:
Done, hope I got your idea right.

JimBobSquarePants (approved):
Looks perfect. Thanks 👍

PR #2092, opened 2022-04-16
Obsolete ImageFormat out params and return on Metadata

tocsoft:
### Proposal
This goes along with the https://github.com/SixLabors/ImageSharp/discussions/2090 proposal so show the impact to the API.

### Description
Obsoletes all the overloads of `Image.Load` that return on `IImageFormat`, removes all internal dependencies on those APIs, returns `OrginalImageFormat` on all images/identify call sites.

`out IImageFormat` overloads no depend on non `out` versions rather than the other way around.

### DIFF
--- a/src/ImageSharp/Image.Decode.cs
+++ b/src/ImageSharp/Image.Decode.cs
@@ -166,8 +166,13 @@ namespace SixLabors.ImageSharp
                 return null;
             }
 
-            IImageInfo info = detector?.Identify(config, stream, cancellationToken);
-            info.Metadata.OrigionalImageFormat = format;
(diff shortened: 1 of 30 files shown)


JimBobSquarePants:
@tocsoft I'm currently doing a rework of this based on your example. It's easier to start from scratch due to all the conflicts following the decoder API churn.

tocsoft:
> @tocsoft I'm currently doing a rework of this based on your example. It's easier to start from scratch due to all the conflicts following the decoder API churn.

sounds good to me..this was mostly ment as an example anyway.

PR #2098, opened 2022-04-22
Use .NET 6 only

JimBobSquarePants:
### Prerequisites

### Description

Reduces our target framework collection to .NET 6 only (.NET 7 when CI testing preview). See https://github.com/SixLabors/ImageSharp/discussions/2072

I'll do a follow up PR to clear out any compiler conditionals for items with guaranteed support. 
e.g `#if SUPPORTS_RUNTIME_INTRINSICS`

### DIFF
--- a/tests/ImageSharp.Tests.ProfilingSandbox/ImageSharp.Tests.ProfilingSandbox.csproj
+++ b/tests/ImageSharp.Tests.ProfilingSandbox/ImageSharp.Tests.ProfilingSandbox.csproj
@@ -12,24 +12,19 @@
     <!--Used to hide test project from dotnet test-->
     <IsTestProject>false</IsTestProject>
     <EnsureNETCoreAppRuntime>false</EnsureNETCoreAppRuntime>
-    <Configurations>Debug;Release;Debug-InnerLoop;Release-InnerLoop</Configurations>
+    <Configurations>Debug;Release</Configurations>
     <ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>
   </PropertyGroup>
 
   <Choose>
     <When Condition="$(SIXLABORS_TESTING_PREVIEW) == true">
       <PropertyGroup>
-        <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;netcoreapp2.1;net472</TargetFrameworks>
-      </PropertyGroup>
-    </When>
-    <When Condition="$(Configuration.EndsWith('InnerLoop')) == true">
-      <PropertyGroup>
-        <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+        <TargetFrameworks>net7.0;net6.0</TargetFrameworks>
       </PropertyGroup>
     </When>
     <Otherwise>
       <PropertyGroup>
-        <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp2.1;net472</TargetFrameworks>
+        <TargetFrameworks>net6.0</TargetFrameworks>
       </PropertyGroup>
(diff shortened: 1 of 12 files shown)


tests/ImageSharp.Tests.ProfilingSandbox/ImageSharp.Tests.ProfilingSandbox.csproj
```diff
@@ -12,24 +12,19 @@
     <!--Used to hide test project from dotnet test-->
```
JimBobSquarePants, line 12: Do we still need this project?

tests/ImageSharp.Tests/TestUtilities/Tests/FeatureTestRunnerTests.cs
```diff
@@ -55,13 +55,17 @@ public void AllowsAllHwIntrinsicFeatures()
                 HwIntrinsics.AllowAll);
         }
 
+#if !NET7_0_OR_GREATER
+        // COMPlus_EnableSIMD isn't thing anymore.
+        // https://github.com/dotnet/runtime/issues/66206
```
tannergooding, line 60: Just as an aside, it also already wasn't a thing for `Arm64` since SIMD support was required by the underlying platform calling convention.

So if a dev were to run this test on an Arm64 device, it would likely fail.

JimBobSquarePants: That’s really useful to know, thanks!

With that knowledge I’ll likely just remove the value and update any relevant tests. With .NET 6 we don’t really need it.

JimBobSquarePants: Or do I still need to think about Arm32 (is that a thing?)

tannergooding: Arm32 doesn't support SIMD acceleration today, the way it has to allocate SIMD registers is significantly different and it's not been a priority since most people are focusing on 64-bit anyways.

Basically `COMPlus_EnableSIMD` is fairly integral to performance, interop, and just general JIT functionality and so you can realistically expect no one is going to disable it or have it disabled, so it's probably safe to not test it (and that's why we've removed it in .NET 7).

The real thing you want to test for, IMO, is a few key instruction set levels. As these represent hardware differences consumers may have.

`COMPlus_EnableHWIntrinsic=0` will basically disable all acceleration. This is mainly useful for testing the "software fallback" path that new platforms may hit and is the most similar to what you probably want instead of `COMPlus_EnableSIMD`.

For other `COMPlus_Enable*=0` settings, that'd be roughly along the lines of that other comment I left a month or two back (the one describing the runtime requirements and the key "hardware versions" you likely want to consider to get broadest reach with the "least" overall work).

JimBobSquarePants: Perfect. I'll just get rid of it all together again. I need to dig out your old comment. Full of information we're gonna need over the next few months. Thanks again!

tannergooding: Sure thing. And feel free to ping me on any of these. Always happy to drop in and help answer questions or give insight.

ImageSharp is a good library and this is one of the ways I can reasonably "contribute" with my otherwise busy schedule.

JimBobSquarePants: Thanks! I really appreciate any assistance you provide.

Issue #2107, opened 2022-05-06
webp (from jpeg) gives inaccurate colors


marklonquist:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp.Web
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp.Web/issues) and [closed](https://github.com/SixLabors/ImageSharp.Web/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp.Web version

2.0.0

### Other Six Labors packages and versions

ImageSharp 2.1.1

### Environment (Operating system, version and so on)

Linux container in GCP running on Kubernetes

### .NET Framework version

NET6

### Description

Colors are inaccurate when setting format to webp from a jpeg original.

### Steps to Reproduce

Set `format=webp` on a jpeg original in a side-by-side comparison and observe the inaccurate colors

### Images

Left side: WebP - Right side: Original JPEG
![Screenshot 2022-05-06 at 10 00 51](https://user-images.githubusercontent.com/1456819/167091764-df4e6791-5e35-4332-b40c-8904b207839f.png)

br3aker:
Can you provide an image? It's almost impossible to determine errors just from before/after comparison.

brianpopow:
Converting to bmp and tga also has this incorrect color, while tiff seems to be ok.

br3aker:
Yep, checked with JPEGsnoop - this image has color profile.

@marklonquist unfortunately, this is yet to be implemented in https://github.com/SixLabors/ImageSharp/issues/74.

marklonquist:
> Yep, checked with JPEGsnoop - this image has color profile.
> 
> @marklonquist unfortunately, this is yet to be done in #1567.

Thanks for the info.

brianpopow:
There is something we can do actually: Currently we are not preserving the ICCP when encoding webp and bmp. We can at least change that.

brianpopow:
With #2108, #2109 and #2110 merged to `main` branch, the color profile will be preserved when encoding webp, bmp and png images. With that the image color should look correct when the viewer of the image supports color profiles (most browsers should do that).

marklonquist:
[...]

PR #2110, opened 2022-05-07
Preserve color profile when encoding PNG images

brianpopow:
### Prerequisites

### Description

This PR changes the PNG encoder to preserve the color profile when encoding an image.

Related issue: #2107

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -1223,26 +1223,26 @@ namespace SixLabors.ImageSharp.Formats.Png
             fixed (byte* compressedDataBase = compressedData)
             {
                 using (IMemoryOwner<byte> destBuffer = this.memoryAllocator.Allocate<byte>(this.Configuration.StreamProcessingBufferSize))
-                using (var memoryStream = new UnmanagedMemoryStream(compressedDataBase, compressedData.Length))
-                using (var bufferedStream = new BufferedReadStream(this.Configuration, memoryStream))
+                using (var memoryStreamOutput = new MemoryStream(compressedData.Length))
+                using (var memoryStreamInput = new UnmanagedMemoryStream(compressedDataBase, compressedData.Length))
+                using (var bufferedStream = new BufferedReadStream(this.Configuration, memoryStreamInput))
                 using (var inflateStream = new ZlibInflateStream(bufferedStream))
                 {
-                    Span<byte> dest = destBuffer.GetSpan();
+                    Span<byte> destUncompressedData = destBuffer.GetSpan();
                     if (!inflateStream.AllocateNewBytes(compressedData.Length, false))
                     {
                         uncompressedBytesArray = Array.Empty<byte>();
                         return false;
                     }
 
-                    var uncompressedBytes = new List<byte>(compressedData.Length);
-                    int bytesRead = inflateStream.CompressedStream.Read(dest, 0, dest.Length);
+                    int bytesRead = inflateStream.CompressedStream.Read(destUncompressedData, 0, destUncompressedData.Length);
                     while (bytesRead != 0)
                     {
-                        uncompressedBytes.AddRange(dest.Slice(0, bytesRead).ToArray());
-                        bytesRead = inflateStream.CompressedStream.Read(dest, 0, dest.Length);
+                        memoryStreamOutput.Write(destUncompressedData.Slice(0, bytesRead));
+                        bytesRead = inflateStream.CompressedStream.Read(destUncompressedData, 0, destUncompressedData.Length);
                     }
 
-                    uncompressedBytesArray = uncompressedBytes.ToArray();
+                    uncompressedBytesArray = memoryStreamOutput.ToArray();
                     return true;
                 }
             }
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -1206,27 +1206,45 @@ namespace SixLabors.ImageSharp.Formats.Png
 
             ReadOnlySpan<byte> compressedData = data.Slice(zeroIndex + 2);
 
-            using (var memoryStream = new MemoryStream(compressedData.ToArray()))
-            using (var bufferedStream = new BufferedReadStream(this.Configuration, memoryStream))
-            using (var inflateStream = new ZlibInflateStream(bufferedStream))
+            if (this.TryUncompressZlibData(compressedData, out byte[] iccpProfileBytes))
             {
-                if (!inflateStream.AllocateNewBytes(compressedData.Length, false))
-                {
-                    return;
-                }
-
-                var uncompressedBytes = new List<byte>();
-
-                // Note: this uses a buffer which is only 4 bytes long to read the stream, maybe allocating a larger buffer makes sense here.
-                int bytesRead = inflateStream.CompressedStream.Read(this.buffer, 0, this.buffer.Length);
-                while (bytesRead != 0)
-                {
-                    uncompressedBytes.AddRange(this.buffer.AsSpan(0, bytesRead).ToArray());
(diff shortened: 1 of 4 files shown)


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1174,6 +1178,58 @@ private bool TryReadLegacyExifTextChunk(ImageMetadata metadata, string data)
...
+            if (!this.TryReadTextKeyword(keywordBytes, out string name))
+            {
+                return;
+            }
+
+            ReadOnlySpan<byte> compressedData = data.Slice(zeroIndex + 2);
+
+            using (var memoryStream = new MemoryStream(compressedData.ToArray()))
+            using (var bufferedStream = new BufferedReadStream(this.Configuration, memoryStream))
+            using (var inflateStream = new ZlibInflateStream(bufferedStream))
```
JimBobSquarePants, line 1211: I think we can avoid the allocation here. 

https://github.com/dotnet/runtime/issues/24622#issuecomment-356465832

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -687,19 +693,47 @@ private void WriteXmpChunk(Stream stream, ImageMetadata meta)
                 PngConstants.XmpKeyword.CopyTo(payload);
                 int bytesWritten = PngConstants.XmpKeyword.Length;
 
-                // Write the iTxt header (all zeros in this case)
+                // Write the iTxt header (all zeros in this case).
                 payload[bytesWritten++] = 0;
                 payload[bytesWritten++] = 0;
                 payload[bytesWritten++] = 0;
                 payload[bytesWritten++] = 0;
                 payload[bytesWritten++] = 0;
```
gfoidl, line 701: Not directly related to this PR, but
```suggestion
                Span<byte> iTxtHeader = payload.Slice(bytesWritten);
                iTxtHeader[4] = 0;
                iTxtHeader[3] = 0;
                iTxtHeader[2] = 0;
                iTxtHeader[1] = 0;
                iTxtHeader[0] = 0;
                bytesWritten += 5;
```
will avoid extra bound checks for each element access, except for the first one (index 4). Cf. [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKBuIGYACMxgYUYG8bGfmniUjALLkAFAGUADtgB2AHmABPDDAB8jaYoA2EbABM0jAJYyMjJStwB1KEYwqZASkbdeXar08bs23XoDaFjDWtvYwMgDUEQC6jAC8jAAMANyuXt6++oHKwTZ2DlGxCSlpXpo6WUEh+eGF8UmpHukZFQFVeWGRMfUlTenlftmWHQXdxY2eAL6lpQzMgkKkEtLyQeoD+oYmZu2hDs6l7s2MUrIKOepGACoIGAASMPowUPUbegB04lpGYDCiuzUnBNmtdbg8nlB/Cgig1Sp5QfdHnpnv56DDescEeDkZDSOjgeksUiUeR8XDeESIf5EmS+oxpnTZvwFvRlmc1i0/FtTOYctVOgc6UdmqdVhdjDdERDXj5Wp9vr9/nyRuFHASvJScdTaZjJdiSTqQXribjDYTjVS0T11fCLVrodbSgyGXR+OQkMxSIwAAoyADmrAgMlwGFkGFwNGFPDmkGDZm2whMVxgtwA0jBFAB3aB6AAy4T9GAAFvV6OqY0GQ8YeUJsAhk2mM9moHmC8X6gB2RqTIA)

Note: for .NET 7 the indices can be written in ascending order too.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -687,19 +693,47 @@ private void WriteXmpChunk(Stream stream, ImageMetadata meta)
...
+
+            byte[] compressedData = this.GetZlibCompressedBytes(iccProfileBytes);
+            int payloadLength = ColorProfileName.Length + compressedData.Length + 2;
+            using (IMemoryOwner<byte> owner = this.memoryAllocator.Allocate<byte>(payloadLength))
+            {
+                Span<byte> outputBytes = owner.GetSpan();
+                PngConstants.Encoding.GetBytes(ColorProfileName).CopyTo(outputBytes);
+                int bytesWritten = ColorProfileName.Length;
+                outputBytes[bytesWritten++] = 0; // Null separator.
+                outputBytes[bytesWritten++] = 0; // Compression.
```
gfoidl, line 731: Here the above "trick" doesn't make sense, as for slicing there's an argument validation, so we just would exchange one comparison for the bound-check against the argument validation. So no net-win.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1174,6 +1178,76 @@ private bool TryReadLegacyExifTextChunk(ImageMetadata metadata, string data)
...
+                {
+                    if (!inflateStream.AllocateNewBytes(compressedData.Length, false))
+                    {
+                        uncompressedBytesArray = Array.Empty<byte>();
+                        return false;
+                    }
+
+                    var uncompressedBytes = new List<byte>(compressedData.Length);
+
+                    // Note: this uses a buffer which is only 4 bytes long to read the stream, maybe allocating a larger buffer makes sense here.
```
brianpopow, line 1237: Should we use the memory allocator to to create a bigger buffer to read the uncompressed data into here? using the scratch buffer feels like it could be very inefficient.

JimBobSquarePants: Yeah, lot's of small allocations. I would use the allocator to allocate a buffer of length `Configuration.StreamProcessingBufferSize`.

brianpopow: I was thinking about this for a while now, but if we always use the allocator this could be also inefficient, if we have alot of small compressed text string's.

Maybe something like this:
```
Span<byte> destBuffer = this.buffer.AsSpan();
using IMemoryOwner<byte> foo = compressedData.Length > 1024 ? this.memoryAllocator.Allocate<byte>(1024) : null;
if (compressedData.Length > 1024)
{
   destBuffer = foo.GetSpan();
}
```

For the the buffer to uncomress into?

JimBobSquarePants:
```c#
uncompressedBytes.AddRange(this.buffer.AsSpan(0, bytesRead).ToArray());
```

It's actually this line that concerns me the most. Zlib compression is pretty good and you could end up with a large number of `ToArray()` calls. That's why I favour a much larger array and simply slicing it. Pooling is pretty cheap.

brianpopow: I dont see how this `ToArray` call can be avoided, though. We cannot pass a Span to `AddRange`, see [dotnet1530](https://github.com/dotnet/runtime/issues/1530)

br3aker: Maybe `MemoryStream`?

Can be used with spans and can expand like a list.
No virt calls as we would work with direct type.
Should be relatively fast as it uses `Buffer.InternalBlockCopy` for big buffers.

brianpopow: @br3aker good idea, that will work: a0e38c87

JimBobSquarePants (approved):
Very cool 👍

brianpopow:
I wanted to merge this now, but two tests for `net7.0` are failing now with the CI. Tried to restart two times the test run's, but its still failing. 

I kind of doubt this is related to the PR changes.

JimBobSquarePants:
Definitely not us. There’s been a new .NET7 preview release. We should check the output and report any issues.

brianpopow:
> Definitely not us. There’s been a new .NET7 preview release. We should check the output and report any issues.

I will merge this PR and create a tracking issue for the .NET7 issue

Issue #1762, opened 2021-09-14
Revise how constant SIMD vectors are defined


gfoidl:
Starting with .NET 5 the JIT emits code to read constant SIMD vectors from the data section which gives nice machine code and some related advantages like the ones listed in https://github.com/SixLabors/ImageSharp/pull/1761#discussion_r708063156.
See also https://github.com/dotnet/runtime/issues/44115 for more context on that issue.

But changing the usage of constant vectors to the "inline variant" will regress pre .NET 5 targets, thus it seems best to keep current code with `static readonly`s, and update to the inline variant once .NET Core 3.1 will go out of support on [3rd Dec 2022](https://dotnet.microsoft.com/platform/support/policy) (the purpose of this issue is to have a remainder for that task).

antonfirsov:
3rd Dec 2022 is pretty far, ideally we would release ImageSharp 3.0 before that, and personally I wouldn't remove .NET Core 3.1 without bumping the major version, so I'm afraid it's not anything for the near future.

Would an inclined property be equivalent of "inline vector create"? Can we do a trick like the following?

```C#
#if NET5_OR_NEWER
private static Vector256<float> Constant =>  Vector256.Create(0.707106781f);
#else
private static readonly  Vector256<float> Constant = Vector256.Create(0.707106781f);
#endif
```

I would only do it if there is measurable benefit justifying the increased complexity.

gfoidl:
The trick [seems to work](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8ovLrl5hcAbhr0mrTj36CRGMRKm4WADQAcSOdXkBmJqQYBhBjQDeNBk4YAHMQDdsGGIqQMAajBgGNCkAKxIADwAZgA2EF4AfNYQEhjYPAwAvMkBQSFQ4UgsVrBeMAAUdCwA7HQ15HRINfrk0QCUxs621N3E5oHBoREx8UkMALIVgwVFowkYydjtPd0Ovd3OxDUMAILuCCy7ACbHFdhoKWkZGJ2OzgC+907PTAP5w1FxC8kT5NMfQojb7jZavdabLY7faHE5nC55IZA4qlGDlKq1eqNZqtDp3DZOJ7UB5AA==) (at least for simple cases, where it's easy for the JIT to inline the property).

[...]

Issue #2121, opened 2022-05-18
Vector{128,256} operations that use MmShuffle fall back to method call


gfoidl:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

Current main branch

### Other ImageSharp packages and versions

none

### Environment (Operating system, version and so on)

all .NET supported

### .NET Framework version

all

### Description

While working on https://github.com/SixLabors/ImageSharp/issues/1762 I recognized that methods that use https://github.com/SixLabors/ImageSharp/blob/c661ab1478419004271a1bd2281f5d4e478840b9/src/ImageSharp/Common/Helpers/SimdUtils.Shuffle.cs#L236-L238 won't emit platform-intrinsics, rather fallback to a method call as the value isn't a constant.

E.g. `Vp8Encoding.FTransformPass1SSE2` looks after inlining the vector constants like
```asm
       push      rdi
       push      rsi
       push      rbx
       sub       rsp,60
       vzeroupper
       xor       eax,eax
       mov       [rsp+50],rax
       mov       [rsp+58],rax
       mov       rsi,rdx
       mov       rdi,r8
       mov       rbx,r9
       vmovupd   xmm0,[rcx]
       lea       rcx,[rsp+40]
       vmovapd   [rsp+30],xmm0
       lea       rdx,[rsp+30]
       mov       r8d,0B1
       call      System.Runtime.Intrinsics.X86.Sse2.ShuffleHigh(System.Runtime.Intrinsics.Vector128`1<Int16>, Byte)
       vmovupd   xmm0,[rsi]
       lea       rcx,[rsp+50]
       vmovapd   [rsp+30],xmm0
       lea       rdx,[rsp+30]
       mov       r8d,0B1
       call      System.Runtime.Intrinsics.X86.Sse2.ShuffleHigh(System.Runtime.Intrinsics.Vector128`1<Int16>, Byte)
       vmovapd   xmm0,[rsp+40]
       vpunpcklqdq xmm0,xmm0,[rsp+50]
       vmovapd   xmm1,[rsp+40]
[...]

PR #2122, opened 2022-05-18
Used inline SIMD vectors if they are constants

gfoidl:
### Prerequisites

### Description

Starting with .NET 5 constant SIMD vectors can be used / created "inline", so that the JIT can read the values from the data section. Thus the need of static readonly fields for these kind of vectors is gone.

Codegen (x64) was spot-checked. Basically it's
```diff
-      mov       rax,19D98002F38
-      mov       rax,[rax]
-      vmovupd   xmm8,[rax+8]
+      vmovupd   xmm0,[7FFF734BC5F0]
```
For loops or other code that isn't tiered compiled by the JIT, the static init check needed by the runtime is also elided with this change.

In micro-benchmarks that change allone I expect to be just below noise, as all data is in the cache and the branch-predictor (static init check) will be trainined. The difference could (hopefully be should) be visible in real-world apps.

Note: for micro-benchmarks https://github.com/SixLabors/ImageSharp/issues/2121 will have an effect, see machine code comparison over there.

Fixes https://github.com/SixLabors/ImageSharp/issues/1762
Contributes to https://github.com/SixLabors/ImageSharp/issues/2121

#### Patterns applied

* If a constant vector is used only once outside a loop, then `Vector{128,256}.Create` is called in-place of the argument. That way the JIT is able to produce code like
  ```asm
  vmulps reg, reg, [mem]
  ```
  and no additional (logical) register is needed.

* If a constant vector is used once inside a loop, then the `Vector{128,256}.Create` is hoisted above the loop(s), as the JIT (currently) won't do this for us.

* If a constant vector is used several times outside a loop, then `Vector{128,256}.Create` is assigned to a vector variable as early as possible (to hide the latency of loading from mem), but not too far away to keep the logical context of the code.


src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
```diff
@@ -65,48 +65,16 @@ internal static unsafe class Vp8Encoding
...
-        private static readonly Vector128<int> K51000 = Vector128.Create(51000);
-
-        private static readonly byte MmShuffle2301 = SimdUtils.Shuffle.MmShuffle(2, 3, 0, 1);
-
-        private static readonly byte MmShuffle1032 = SimdUtils.Shuffle.MmShuffle(1, 0, 3, 2);
-#pragma warning restore SA1310 // Field names should not contain underscore
-#endif
-
-        static Vp8Encoding()
+        private static byte[] GetClip1()
```
gfoidl, line 68: Instead of the static constructor (cctor) this method is now called from the `static readonly` field.
Avoiding a cctor and using `beforefieldinit` (IL) is the prefered way in dotnet/runtime too, as there are several pro reasons for it, liking timing, static init check, etc.

That's the only other code change not mentioned in PR's description.

gfoidl:
The CI leg for _windows-latest_ is failling with
```

**Comment by JimBobSquarePants** [comment]
@gfoidl 

Yep, see https://github.com/SixLabors/ImageSharp/issues/2117

Not sure what the cause is yet. It's only happening with the latest .NET preview 4.

If you update your branch from main those tests will be skipped for now on that platform.

**Comment by gfoidl** [comment]
The CI leg for windows-latest is failing with
```
...ImageDifferenceIsOverThresholdException : Image difference is over threshold!
Test Environment OS : Windows
Test Environment is CI : True
Test Environment is .NET Core : True
Report ImageFrame 0:
Total difference: 29.9761%
...
```
Local tests are passing (VS, standard checkout as given in ReadMe).
Please give me some advice on how to investigate this.

gfoidl:
@JimBobSquarePants thanks, now we had a race condition.
The short comment (now deleted) GH created while copying the error message (don't know why).

Thanks!

Issue #2123, opened 2022-05-20
AccessViolationException when loading TIFF image


botinko:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

2.1.1

### Other ImageSharp packages and versions

2.1.1

### Environment (Operating system, version and so on)

Windows 10 21H2, Linux

### .NET Framework version

.NET 5

### Description

```
    using var fileStream = File.Open("error.tiff", FileMode.Open);
    Image.Load(fileStream);
```

```
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at SixLabors.ImageSharp.Formats.Tiff.PhotometricInterpretation.YCbCrTiffColor`1[[SixLabors.ImageSharp.PixelFormats.Rgba32, SixLabors.ImageSharp, Version=2.0.0.0, Culture=neutral,
 PublicKeyToken=d998eea7b14cab13]].Decode(System.ReadOnlySpan`1<Byte>, SixLabors.ImageSharp.Memory.Buffer2D`1<SixLabors.ImageSharp.PixelFormats.Rgba32>, Int32, Int32, Int32, Int32)
   at SixLabors.ImageSharp.Formats.Tiff.TiffDecoderCore.DecodeStripsChunky[[SixLabors.ImageSharp.PixelFormats.Rgba32, SixLabors.ImageSharp, Version=2.0.0.0, Culture=neutral, PublicK
eyToken=d998eea7b14cab13]](SixLabors.ImageSharp.ImageFrame`1<SixLabors.ImageSharp.PixelFormats.Rgba32>, Int32, System.Span`1<UInt64>, System.Span`1<UInt64>, System.Threading.Cancell
ationToken)
   at SixLabors.ImageSharp.Formats.Tiff.TiffDecoderCore.DecodeFrame[[SixLabors.ImageSharp.PixelFormats.Rgba32, SixLabors.ImageSharp, Version=2.0.0.0, Culture=neutral, PublicKeyToken
[...]

PR #2124, opened 2022-05-20
Jpeg compressed tiff: jpeg decoder should handle the conversion from YCbCr to RGB

brianpopow:
### Prerequisites

### Description

This PR provides a fix for #2123: If the image uses Jpeg Compression, the jpeg decoder should handle the conversion from YCbCr to RGB and not the Tiff Decoder, See discussion in #2123


JimBobSquarePants:
@brianpopow You're not gonna like this but since it's such a critical bug we'll have to backport this PR to the [release/2.1.x](https://github.com/SixLabors/ImageSharp/tree/release/2.1.x) branch.

JimBobSquarePants:
@brianpopow This looks to be how libjpeg-turbo figured it out.

https://github.com/libjpeg-turbo/libjpeg-turbo/blob/1f55ae7b0fa3acc348a630171617d0e56d922b68/jdapimin.c#L127

brianpopow:
I think it is working now as it should. I had to change how the jpeg decoder deduces the colorspace. That's why this should be reviewed carefully. 

It is close to how libjpeg does it, but not exactly the same. libjpeg checks for a JFIF marker first and then concludes it should be YCbCr. We first check for adobe with a color transform and if the component Id's are ASCII R, G, B. The reason for that is, that there is a test image `Tiff\rgb_jpegcompressed_nojpegtable.tiff`, which has a JFIF marker, but the component Id's indicate it's in fact RGB.

@br3aker if you have time to review this part, your feedback would be very much appreciated.

br3aker:
Ups, completely missed this, will review tomorrow!

br3aker (commented):
Strictly speaking, [oracle docs](https://docs.oracle.com/javase/7/docs/api/javax/imageio/metadata/doc-files/jpeg_metadata.html#color) are not accurate according to jpeg specs. We can have 2 different 'versions' for the jpeg image - jfif and adobe jfif (exif of whatever it's called). So we can have 2 different paths:

1. Have jfif marker
2. Have app14 adobe marker

But we can't have both at the same time according to the specs. I'd suggest to write it like this:

```cs
if (componentCount == 1)
{
    return grayscale;
}

if (componentCount == 3)
{
    // we prioritize adobe marker over jfif marker
    // if somebody really encoded this image with redundant adobe marker
    // then it's most likely an adobe jfif image IMO
    if (this.adobeMarker is not null)
    {
        if(this.adobeMarker.Transform == ycbcr) return ycbcr;
        else if(this.adobeMarker.Transform == unknown) return rgb;

        // throw if color transform flag has invalid value?
    }

    if (this.jfifMarker is not null)
    {
        return ycbcr;
    }

    // fallback to the id color deduction
    return this.Components[2].Id == 66 && this.Components[1].Id == 71 && this.Components[0].Id == 82 ? rgb : ycbcr;
}

if (componentCount == 4)
{
    // jfif images don't not support 4 component images
    // so we only check adobe
    if (this.adobeMarker is not null)
    {
        if(this.adobeMarker.Transform == ycck) return ycck;
        else if(this.adobeMarker.Transform == unknown) return cmyk;

        // throw if color transform flag has invalid value?
    }

    // fallback to cmyk as neither of cmyk nor ycck have 'special' component ids
    return cmyk;
}

JpegThrowHelper.ThrowNotSupportedComponentCount(componentCount);
return default;
```

P.S.
I personally don't like component id color deduction as ids are used to determine components in scans, they have nothing to do with encoding color. AFAIR 'RGB' ASCII ids were proposed by Adobe before jpeg extension via APP14 marker was published.

brianpopow:
@br3aker thanks for your feedback, I have changed the color deduction accordingly.

For `Tiff\rgb_jpegcompressed_nojpegtable.tiff`: I am pretty sure this was wrongly encoded (by myself probably). The jpeg data was not following the spec. This is changed now.

JimBobSquarePants:
If @br3aker is happy with this then I am 😄 

@brianpopow I'm gonna create a PR using these files against the [release/2.1.x](https://github.com/SixLabors/ImageSharp/tree/release/2.1.x) branch and once it's built will merge both.

antonfirsov:
>  I was a bit unsure how to do it, never made a release.

After merging this PR, we should rebase all the commits in this PR on `release/2.1.x` and do a separate "backport" PR. We may also consider squashing them into a single commit, if that makes things simpler.

@JimBobSquarePants I'm not sure what happens after merging the backport PR, is it enough to just tag the merge commit with `2.1.2` or are there any manual steps you have to perform? How does MinVer figure out that our next `patch=2` in the package version?

antonfirsov:
@JimBobSquarePants is faster doing than me talking 😆

JimBobSquarePants:
Haha  😄  Yeah, See #2126 

It should be enough to PR against the release branch then create a tagged release via the UI here ensuring we are pointing at the correct branch to create the release from.

Issue #2133, opened 2022-05-29
2.1.2: Jpeg YCbcr encode cause pink color blur


lifeglam:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

2.1.2

### Other ImageSharp packages and versions

2.1.1

### Environment (Operating system, version and so on)

Windows 11, x86

### .NET Framework version

.Net Core 3.0

### Description

Jpeg image with YCbCR color space encode with pink blur (sample image added below). This only repros with latest 2.1.2. (2.1.1 works without issue)

### Steps to Reproduce

`using (Image<Rgba32> image = Image.Load<Rgba32>(File.OpenRead(sourceFilePath)))
            {
                using (var destinationImageStream = new MemoryStream())
                {
                    destinationImageStream.Position = 0;
                    image.SaveAsJpeg(destinationImageStream);
                }
            }`

### Images

![original](https://user-images.githubusercontent.com/13856985/170858158-9e55ea35-53ed-40e1-84e6-eff74b7c76f7.jpg)
![converted_pink](https://user-images.githubusercontent.com/13856985/170858177-58e820fa-5908-430c-a933-712d406ab494.jpeg)

brianpopow:
closed via #2135

PR #2135, opened 2022-05-29
Deduce color space as YCbCr if component id's are 1, 2, 3

brianpopow:
### Prerequisites

### Description

Another case of deducing the color space of a jpeg image based on the component id's.

Related issue #2133

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -556,6 +556,12 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
                     return JpegColorSpace.RGB;
                 }
 
+                // If these values are 1-3 for a 3-channel image, then the image is assumed to be YCbCr.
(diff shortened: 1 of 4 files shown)


br3aker (approved):
Looks good. Image from the issue is neither jfif nor adobe jfif, it was encoded with app1 marker which we indeed ignored during color deduction. I didn't find any information regarding jpeg encoding color type from exif specs so component id color deduction seems to be a solution.

brianpopow:
> Looks good. Image from the issue is neither jfif nor adobe jfif, it was encoded with app1 marker which we indeed ignored during color deduction. I didn't find any information regarding jpeg encoding color type from exif specs so component id color deduction seems to be a solution.

@br3aker thanks for your feedback!

JimBobSquarePants:
Do you think we should backport this since we introduced the issue in the last patch?

brianpopow:
> Do you think we should backport this since we introduced the issue in the last patch?

I would not consider this a critical issue, it's more an edge case, so I don't think we need to backport this.

Issue #2136, opened 2022-05-30
InvalidImageContentException "Input stream does not have enough bytes  to parse declared contents of the 00 marker" when loading JPG created by Brother scanner


w-michal:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

2.1.2

### Other ImageSharp packages and versions

N/A

### Environment (Operating system, version and so on)

Windows 10 x64, 21H2

### .NET Framework version

net6.0

### Description

Looks like my scanner (Brother 920dw) creates broken JPEGs by default... 

Call stack
```
An unhandled exception of type 'SixLabors.ImageSharp.InvalidImageContentException' occurred in SixLabors.ImageSharp.dll: 'Input stream does not have enough bytes to parse declared contents of the 00 marker.'
   at SixLabors.ImageSharp.Formats.Jpeg.JpegThrowHelper.ThrowNotEnoughBytesForMarker(Byte marker)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.ParseStream(BufferedReadStream stream, HuffmanScanDecoder scanDecoder, CancellationToken cancellationToken)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoderCore.Decode[TPixel](BufferedReadStream stream, CancellationToken cancellationToken)
   at SixLabors.ImageSharp.Formats.ImageDecoderUtilities.Decode[TPixel](IImageDecoderInternals decoder, Configuration configuration, Stream stream, Func`3 largeImageExceptionFactory, CancellationToken cancellationToken)
   at SixLabors.ImageSharp.Formats.ImageDecoderUtilities.Decode[TPixel](IImageDecoderInternals decoder, Configuration configuration, Stream stream, CancellationToken cancellationToken)
   at SixLabors.ImageSharp.Formats.Jpeg.JpegDecoder.Decode[TPixel](Configuration configuration, Stream stream, CancellationToken cancellationToken)
[...]

PR #2147, opened 2022-06-05
Skip invalid markers during jpeg decoding

br3aker:
### Prerequisites

### Description
Fixes https://github.com/SixLabors/ImageSharp/issues/2136

Given image contained extra unused bytes right at the end of the SOS segment containing RST markers and `0xFF00` sequences. This lead to non-SOS parser to count them as valid markers and failed with exception. With this PR jpeg decoder now skips any invalid markers during non-SOS parsing so we can at least give the user something instead of exception.

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -489,7 +489,6 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
 
                 // Read on.
                 fileMarker = FindNextFileMarker(stream);
-                Console.WriteLine($"Found marker: {fileMarker.Marker} at {fileMarker.Position}");
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -491,7 +488,8 @@ internal void ParseStream(BufferedReadStream stream, SpectralConverter spectralC
                 }
 
                 // Read on.
-                fileMarker = FindNextFileMarker(this.markerBuffer, stream);
+                fileMarker = FindNextFileMarker(stream);
+                Console.WriteLine($"Found marker: {fileMarker.Marker} at {fileMarker.Position}");
```
brianpopow, line 492: This Console.WriteLine is probably a leftover from debugging

br3aker: Ups!

brianpopow (approved):
Looks good, thank you!

PR #2172, opened 2022-07-13
Fix run-length encode compression implementation mistake in tga encoder

l1nxy:
### Prerequisites

### Description

According to the [Tga format spec](https://www.dca.fee.unicamp.br/~martino/disciplinas/ea978/tgaffs.pdf) page 24:

> Run-length Packets should never encode pixels from more than one scan line. Even if the end of one
scan line and the beginning of the next contain pixels of the same value, the two should be encoded
as separate packets. In other words, Run-length Packets should not wrap from one line to another.
This scheme allows software to create and use a scan line table for rapid, random access of
individual lines. Scan line tables are discussed in further detail in the Extension Area section of this
document.

It should not cross multiple lines when using Run-length encode compression. Some libraries can not fix the rle mistake when importing images saved by ImageSharp, and will throw an exception.


brianpopow:
@l1nxy thanks for providing a fix for this.

Some questions I have:
- Can you provide a test image where this error happens?
- Does ImageMagick detect this error? If so, we should add a test for this case (we do use ImageMagick as a ReferenceDecoder for unit tests).

l1nxy:
I will close this pr and make a new one, which should add a new option to control which RLE compression is selected.

brianpopow:
> I will close this pr and make a new one, which should add a new option to control which RLE compression is selected.

Mhm, I dont think this is a good idea. Why would someone want to have an option to write a invalid tga? If its against the spec, we should not do it.

l1nxy:
> > I will close this pr and make a new one, which should add a new option to control which RLE compression is selected.
> 
> Mhm, I dont think this is a good idea. Why would someone want to have an option to write a invalid tga? If its against the spec, we should not do it.

But the legacy format is used widely, ImageMagick and Many picture viewers can detect the legacy format. 
What do you think about it?

brianpopow:
> > > I will close this pr and make a new one, which should add a new option to control which RLE compression is selected.
> > 
> > 
> > Mhm, I dont think this is a good idea. Why would someone want to have an option to write a invalid tga? If its against the spec, we should not do it.
> 
> But the legacy format is used widely, ImageMagick and Many picture viewers can detect the legacy format. What do you think about it?

I think its ok, if we can decode such image, but I dont think we should encode such tga which do not follow the spec. Can we actually decode such images? You have not provided a test image yet.

l1nxy:
> > > > I will close this pr and make a new one, which should add a new option to control which RLE compression is selected.
> > > 
> > > 
> > > Mhm, I dont think this is a good idea. Why would someone want to have an option to write a invalid tga? If its against the spec, we should not do it.
> > 
> > 
> > But the legacy format is used widely, ImageMagick and Many picture viewers can detect the legacy format. What do you think about it?
> 
> I think its ok, if we can decode such image, but I dont think we should encode such tga which do not follow the spec. Can we actually decode such images? You have not provided a test image yet.
> 
> edit: Also note you can add commits to this PR without the need to re-open a new one.

OK, thank you very much. i will provide some test pic later. and test the encoder.

src/ImageSharp/Formats/Tga/TgaEncoderCore.cs
```diff
@@ -225,35 +228,29 @@ private void WriteRunLengthEncodedImage<TPixel>(Stream stream, ImageFrame<TPixel
...
         /// <returns>The number of equal pixels.</returns>
-        private byte FindEqualPixels<TPixel>(Buffer2D<TPixel> pixels, int xStart, int yStart)
+        private byte FindEqualPixels<TPixel>(Buffer2D<TPixel> pixels, int xStart, int yPos)
             where TPixel : unmanaged, IPixel<TPixel>
         {
             byte equalPixelCount = 0;
-            bool firstRow = true;
-            TPixel startPixel = pixels[xStart, yStart];
-            for (int y = yStart; y < pixels.Height; y++)
+            TPixel startPixel = pixels[xStart, yPos];
```
brianpopow, line 237: Instead of using the indexer, you could use `DangerousGetRowSpan`, which should be faster. Like this:

```
Span<TPixel> pixelRow = pixels.DangerousGetRowSpan(yPos).Slice(xStart);
```

brianpopow (changes requested):
I would like two unit tests to be added to make sure it works now as it should. (I would do it by myself, but for some weird reason gitlfs does not let me push to this PR)

1. One test for decoding the legacy format (even though we can decode it already, a test to make sure would be nice).
Add the following to `TgaDecoderTests.cs`
```
[Theory]
[WithFile(Github_RLE_legacy, PixelTypes.Rgba32)]
public void TgaDecoder_CanDecode_LegacyFormat<TPixel>(TestImageProvider<TPixel> provider)
{
    using (Image<TPixel> image = provider.GetImage(TgaDecoder))
    {
        image.DebugSave(provider);
        ImageComparingUtils.CompareWithReferenceDecoder(provider, image);
    }
}
```

and the `Github_RLE_legacy` to `TestImages.cs`.

2. Make sure the encoded bytes now matches the expected size. Add to `TgaEncoderTests.cs`

```
// Run length encoded pixels should not exceed row boundaries.
// https://github.com/SixLabors/ImageSharp/pull/2172
[Theory]
[MemberData(nameof(TgaBitsPerPixelFiles))]
public void TgaEncoder_RunLengthDoesNotCrossRowBoundaries(string imagePath, TgaBitsPerPixel bmpBitsPerPixel)
{
    var options = new TgaEncoder();

    var testFile = TestFile.Create(imagePath);
    using (Image<Rgba32> input = testFile.CreateRgba32Image())
    {
        using (var memStream = new MemoryStream())
        {
            input.Save(memStream, options);
            // TODO assert the current encoded bytes match.
        }
    }
}
```

JimBobSquarePants:
>I would do it by myself, but for some weird reason gitlfs does not let me push to this PR

There's a whole heap of issues related to GitHub, Git LFS and forks. Looks like they're finally on to figuring out a cause though which is ace.

https://github.com/git-lfs/git-lfs/issues/5001

I use a workaround for pushing to forks when there is no LFS files to be added but that won't help in this case.

src/ImageSharp/Formats/Tga/TgaEncoderCore.cs
```diff
@@ -225,35 +228,29 @@ private void WriteRunLengthEncodedImage<TPixel>(Stream stream, ImageFrame<TPixel
...
-
-                    if (equalPixelCount >= 127)
-                    {
-                        return equalPixelCount;
-                    }
+                    return equalPixelCount;
                 }
 
-                firstRow = false;
+                if (equalPixelCount >= 127)
```
brianpopow, line 250: Reading the spec again, I believe this should be >=128 instead.

From the spec page 25: ```1. Pixel Count: The pixel count can range from 0 to 127. Since a run-length packet never encodes
zero pixels, and in order to make use of all of the values available with seven bits, a zero (0) in Pixel Count
is defined to mean that 1 pixel is encoded by the packet. A 1 in Pixel Count means that 2 pixels are
actually contained in the packet. In other words, the value in Pixel Count is always 1 less than the actual
number of pixels encoded in the packet. Thus, you may encode between 1 and 128 pixels with a single
packet. If you have a run which is longer than 128 bytes, you must encode it using multiple packets.```

This highlights two more problems: 
1. **run-length packet never encodes zero pixels**, which we dont respect. We should add a case for this.
2. **1 in Pixel Count means that 2 pixels are actually contained in the packet.** I believe we are not doing this correct either.

brianpopow: Mhm I think, I miss-judged this yesterday. After reading it once again, I think finding equal pixel count is fine with >= 127.
The loop starts with `x = xStart + 1`, which means if we find 127 equal pixels, with the first pixel we encode 128 in one run at maximum.

Also we do follow **run-length packet never encodes zero pixels**: If `equalPixelCount` is zero, the run contains 1 pixel.

I was irritated about the difference about our encoded data and the one from imagemagick and thought this was the cause. Sorry for the confusion.

brianpopow:
> But there is a wired thing is the image size generated by ImageSharp bigger than the file generated by ImageMagick.

I had some more time looking into this. The size difference currently fort the given testimage is:
- ImageSharp (main): 50.2 kB
- ImageMagick: 35.6 kB

The reason for that is, that we always write Run-Length Packets. If consecutive pixels are not the same, this will add an additional byte overhead to each pixel data. This will add up alot, if there are not many pixels equal. (note: the test image looks good for RLE, but it has transparency).

So to fix this, we need to change this to not always write RLE packets.
This does not have to be part of this PR and can be a follow up PR.

brianpopow:
I will go ahead and merge this. Code itself looks good, just tests are missing. I will do a follow up PR to add tests and fix the image size issue, since I do not know how to work around the git lfs issues

JimBobSquarePants:
>Code itself looks good

It's still using the [indexer](https://github.com/SixLabors/ImageSharp/pull/2172/files#diff-be4e2e4cce711dbb3cc8e37a23bb9993dd82ecc81364e29b64bf57336652c9b1R237) to access pixels which will be slow. Best make sure your follow up fixes that.

Issue #2171, opened 2022-07-13
Webp with lossy conversion weighs a lot more than lossless.


MrZnake:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

2.1.3

### Other ImageSharp packages and versions

-

### Environment (Operating system, version and so on)

x86-64 Win11

### .NET Framework version

.NET 6.0

### Description

Webp with lossy conversion weighs more than lossless regardless of quality used.

![image](https://user-images.githubusercontent.com/20782978/178638409-fd70af06-4624-4c02-8669-aa54a9d53740.png)

### Steps to Reproduce

```csharp
using SixLabors.ImageSharp;

var image = SixLabors.ImageSharp.Image.Load("startFile.png");

image.Save("outLossless.webp", new SixLabors.ImageSharp.Formats.Webp.WebpEncoder
{
    FileFormat = SixLabors.ImageSharp.Formats.Webp.WebpFileFormatType.Lossless
});

image.Save("outLosssy90.webp", new SixLabors.ImageSharp.Formats.Webp.WebpEncoder
{
    FileFormat = SixLabors.ImageSharp.Formats.Webp.WebpFileFormatType.Lossy,
    Quality = 90,
});

image.Save("outLosssy80.webp", new SixLabors.ImageSharp.Formats.Webp.WebpEncoder
{
    FileFormat = SixLabors.ImageSharp.Formats.Webp.WebpFileFormatType.Lossy,
    Quality = 80,
});

image.Save("outLosssy50.webp", new SixLabors.ImageSharp.Formats.Webp.WebpEncoder
{
    FileFormat = SixLabors.ImageSharp.Formats.Webp.WebpFileFormatType.Lossy,
    Quality = 50,
});

image.Save("outLosssy10+Alpha.webp", new SixLabors.ImageSharp.Formats.Webp.WebpEncoder
{
    FileFormat = SixLabors.ImageSharp.Formats.Webp.WebpFileFormatType.Lossy,
    UseAlphaCompression = true,
    Quality = 10,
});

image.Dispose();
```

### Images

[...]

Issue #2117, opened 2022-05-12
Failing tests on .NET7


brianpopow:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

Current main branch

### Other ImageSharp packages and versions

none

### Environment (Operating system, version and so on)

Windows 10

### .NET Framework version

- 7.0.100-preview.4.22252.9
- 7.0.100-preview.6.22352.1
- 7.0.100-preview.7.22377.5
- 7.0.100-rc.1.22431.12

### Description

The following tests fail on .NET 7.0 Windows only. Introduced in 7.0.100-preview.4.22252.9:

- `Skew_IsNotBoundToSinglePixelType<Bgra32>(provider: TestPattern100x50[Bgra32], x: 20, y: 10)`
- `Skew_IsNotBoundToSinglePixelType<Bgra32>(provider: TestPattern100x50[Bgra32], x: -20, y: -10)`

The issue seems to be only occurring in Release mode. Also it seems somehow related to the PixelFormat `Bgra32`

Environment Info

```
OS=Windows 10.0.20348
Intel Xeon Platinum 8272CL CPU 2.60GHz, 1 CPU, 2 logical and 2 physical cores
.NET SDK=7.0.100-preview.6.22352.1
  [Host] : .NET 7.0.0 (7.0.22.32404), X64 RyuJIT
```
<strike>
In addition on Mac .NET 7.0  we are seeing the following triggered by a a call to `System.Runtime.Intrinsics.Vector256.Create(Single value)`

```
Error Message:
   System.PlatformNotSupportedException : Operation is not supported on this platform.
  Stack Trace:
     at System.Runtime.Intrinsics.Vector256.Create(Single value)
   at SixLabors.ImageSharp.Formats.Jpeg.Components.FastFloatingPointDCT.<IDCT8x8_Avx>g__IDCT8x8_1D_Avx|16_0(Block8x8F& block) in /_/src/ImageSharp/Formats/Jpeg/Components/FastFloatingPointDCT.Intrinsic.cs:line 141
[...]

PR #2173, opened 2022-07-13
Fix issue when writing lossy webp with alpha data

brianpopow:
### Prerequisites

### Description

Fix for issue #2171: The alpha chunk data was written completely (the size of the image in pixels) and not only the compressed data when writing a lossy webp image.


brianpopow:
There are new failing tests with .Net7.0, this time on macos only. 222 are failing:

```
Error Message:
   System.PlatformNotSupportedException : Operation is not supported on this platform.
  Stack Trace:
     at System.Runtime.Intrinsics.Vector256.Create(Single value)
   at SixLabors.ImageSharp.Formats.Jpeg.Components.FastFloatingPointDCT.<IDCT8x8_Avx>g__IDCT8x8_1D_Avx|16_0(Block8x8F& block) in /_/src/ImageSharp/Formats/Jpeg/Components/FastFloatingPointDCT.Intrinsic.cs:line 141
   at SixLabors.ImageSharp.Formats.Jpeg.Components.Decoder.JpegComponentPostProcessor.CopyBlocksToColorBuffer(Int32 spectralStep) in /_/src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegComponentPostProcessor.cs:line 70
```

JimBobSquarePants:
That’s concerning. @tannergooding sorry to page but are there changes coming to .NET 7 we should know about?

tannergooding:
`System.Runtime.Intrinsics.Vector256.Create(Single value)` should work on all platforms, it might just not be accelerated on certain ones (x64 machines without AVX and Arm64 machines in general).

Do you have a simple repro? I don't immediately see any changes to the .NET 7 version or to any code newly using `Vector256` in this PR.

brianpopow:
> `System.Runtime.Intrinsics.Vector256.Create(Single value)` should work on all platforms, it might just not be accelerated on certain ones (x64 machines without AVX and Arm64 machines in general).
> 
> Do you have a simple repro? I don't immediately see any changes to the .NET 7 version or to any code newly using `Vector256` in this PR.

Hi @tannergooding. I am not 100% sure if this will be enough, but from the error messages in this [CI run](https://github.com/SixLabors/ImageSharp/runs/7322478738?check_suite_focus=true) I would assume just calling:

```
var scale = Vector256.Create(200.0f);
```

on a Mac should be enough to trigger this. Since I do not have a Mac, I cannot verify that.
@JimBobSquarePants Maybe you have a chance to test, if that is indeed enough?

tannergooding:
This might be an issue specifically around a machine with `AVX` but no `AVX2` support. I'll see if I can get a repro

JimBobSquarePants:
Thanks @tannergooding I've been unable to reproduce this on my old 2.7 GHz Dual-Core Intel Core i5

JimBobSquarePants:
I got some environmental info via #2175 

I'm at a loss to explain the cause.

```
OS=macOS Big Sur 11.6.7 (20G630) [Darwin 20.6.0]
Intel Xeon CPU E5-1650 v2 3.50GHz (Max: 3.34GHz), 1 CPU, 3 logical and 3 physical cores
.NET SDK=7.0.100-preview.6.22352.1
[Host] : .NET 7.0.0 (7.0.22.32404), X64 RyuJIT
```

brianpopow:
> I got some environmental info via #2175
> 
> I'm at a loss to explain the cause.
> 
> ```
> OS=macOS Big Sur 11.6.7 (20G630) [Darwin 20.6.0]
> Intel Xeon CPU E5-1650 v2 3.50GHz (Max: 3.34GHz), 1 CPU, 3 logical and 3 physical cores
> .NET SDK=7.0.100-preview.6.22352.1
> [Host] : .NET 7.0.0 (7.0.22.32404), X64 RyuJIT
> ```

I think Tanner is on the right track with his guess that it's maybe related to a machine with AVX but not AVX2. Intel Xeon CPU [E5-1650](https://ark.intel.com/content/www/de/de/ark/products/64601/intel-xeon-processor-e51650-12m-cache-3-20-ghz-0-0-gts-intel-qpi.html) only has AVX from what I can tell.

JimBobSquarePants:
>Maybe we can workaround these issues to get back clean CI until there is an upstream fix? We can disable Avx path on .NET7 when Avx2.IsSupported == false.

You mean for this instance? I think we're just hitting this one cos it comes first. We use `Vector256.Create(single)` in a few places. Theoretically we could turn off the support by setting the correct environmental variable during CI but I think I'd rather just disable .NET 7 Mac tests for now.

tannergooding:
Ok, I can reproduce the issue and will work on getting a fix up.

JimBobSquarePants:
AWESOME!

antonfirsov:
> I'd rather just disable .NET 7 Mac tests for now.

Yeah, makes sense.

tannergooding:
@JimBobSquarePants, the issue discussed above should be fixed as of https://github.com/dotnet/runtime/pull/72522 and will land in .NET 7 RC1.

-- Also, sorry for the ping after this has been closed for two weeks; wasn't sure of a better place to relay this small note.

JimBobSquarePants:
@tannergooding That's fantastic news, thanks for letting me know. 

That's a very interesting conversation regarding AVX/AVX2 issues. I wonder if any of that is causing our NaN handling issues in #2117 . I've struggled to find the specs for Intel Xeon Platinum 8272CL

Issue #2174, opened 2022-07-14
Jpeg loaded with pink color blur


teamnimbus:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

2.1.2

### Other ImageSharp packages and versions

2.1.3, 3.0.0-alpha.0.11, 3.0.0-alpha.0.21

### Environment (Operating system, version and so on)

Windows 10 21H1 x64

### .NET Framework version

.Net 6.0

### Description

We stumbled on a Jpeg image that gets loaded with wrong colors on all versions after 2.1.1. Tested with 2.1.2, 2.1.3, 3.0.0-alpha.0.11 and 3.0.0-alpha.0.21.
This is very similar to issues #2133, #2141 and #2152. However we don't think it's a duplicate since it's not fixed in either 2.1.3 (like #2133) or 3.0.0-alpha.0.11 (like #2141 / #2152).

### Steps to Reproduce

This NUnit test passes on 2.1.1 and fails on all later versions with "Expected: Rgba32(255, 255, 255, 255) But was:  Rgba32(255, 128, 128, 255)".

```
[Test]
public async Task ImageShouldNotBePink()
{
	var response = await new HttpClient().GetAsync("https://www.mottafashionplace.com/media/catalog/product/1/0/10085001_11.jpg");
	await using var imageStream = await response.Content.ReadAsStreamAsync();

	var image = Image.Load<Rgba32>(imageStream);

	Assert.That(image[0, 0], Is.EqualTo(new Rgba32(255, 255, 255, 255)));
}
```

### Images

We don't own the problematic image so we can't distribute it. You can [download it](https://www.mottafashionplace.com/media/catalog/product/1/0/10085001_11.jpg) and use it as a starting point to create another image with the same properties.

brianpopow:
The image has 3 components with no subsampling, that's why ImageSharp deduces the colorspace to RGB. It seems it should be YCbCr instead, but I am not sure how we can come to this conclusion from the components info's. The component Id's are 0, 1 and 2.

[...]

PR #2177, opened 2022-07-16
Jpeg color space deduction fix

br3aker:
### Prerequisites

### Description
Fixes https://github.com/SixLabors/ImageSharp/issues/2174


brianpopow:
There were 3 failing tiff tests, but I believe it was an issue with the Tiff JpegDecompressor. Hopefully this 1959bf65 fixes it now.

br3aker:
> There were 3 failing tiff tests, but I believe it was an issue with the Tiff JpegDecompressor. 

I believe it's not not that simple :D

TIFF with PhotometricInterpretation=RGB will have YCbCr signature at jpeg decoding level which would lead to invalid results.
Always using Rgb24 converter would lead to YCbCr encoded jpeg having RGB color conversion. I guess I have an idea: we can just pass PhotometricInterpretation to new `TiffSpectralConverter` and resolve color converter using that value.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -506,90 +506,48 @@ public void Dispose()
         /// Returns the correct colorspace based on the image component count and the jpeg frame component id's.
```
brianpopow, line 506: The comment is no longer valid, maybe change it like this:
```suggestion
        /// Returns the correct colorspace based on the image component count and adobe APP14 marker's color transform flag.
```

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -506,90 +506,48 @@ public void Dispose()
...
         /// </summary>
         /// <param name="componentCount">The number of components.</param>
+        /// <param name="adobeMarker">Parsed adobe APP14 marker.</param>
         /// <returns>The <see cref="JpegColorSpace"/></returns>
-        private JpegColorSpace DeduceJpegColorSpace(byte componentCount)
+        internal static JpegColorSpace DeduceJpegColorSpace(byte componentCount, ref AdobeMarker adobeMarker)
         {
-            if (componentCount == 1)
```
brianpopow, line 512: I have found a jpeg image with an app14 marker which is gray, should that not be possible according to the spec?
gray-sample

br3aker: Adobe spec doesn't say anything about grayscale images but I think we can include it as decoding it doesn't violate the spec :)

br3aker:
I've implemented tests for the specific method which resolves encoded jpeg color space. Not only it's easier to maintain, it is also much faster and there's no need to find a ton of images with different setups.

tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.Internal.cs
```diff
@@ -0,0 +1,71 @@
...
+    [Trait("Format", "Jpg")]
+    public partial class JpegDecoderTests
+    {
+        [Theory]
+        [InlineData(1, 0, JpegColorSpace.Grayscale)]
+        [InlineData(3, JpegConstants.Adobe.ColorTransformUnknown, JpegColorSpace.RGB)]
+        [InlineData(3, JpegConstants.Adobe.ColorTransformYCbCr, JpegColorSpace.YCbCr)]
+        [InlineData(4, JpegConstants.Adobe.ColorTransformUnknown, JpegColorSpace.Cmyk)]
+        [InlineData(4, JpegConstants.Adobe.ColorTransformYcck, JpegColorSpace.Ycck)]
+        internal void DeduceJpegColorSpaceAdobeMarker_ShouldReturnValidColorSpace(byte componentCount, byte adobeFlag, JpegColorSpace expectedColorSpace)
```
JimBobSquarePants, line 33: This is all very nice 👍

JimBobSquarePants (approved):
This all looks great! Thanks @br3aker

Issue #272, opened 2017-07-07
Async Support is missing


TsengSR:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageSharp/issuess) and [closed](https://github.com/JimBobSquarePants/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I've skipped through the source and can't seem to find a single real async method, not even for the most basic methods which read images from file system like `Image.Load(...)`. None of them is awaitable. 

It makes the library kind of useless for anything related to ASP.NET Core since I/O operations (reading from network connection, file system or remote azure storage) will block the request thread, killing any scale-ability of the web application.

### System Configuration

- ImageSharp version: 1.0.0-alpha9-00170
- Environment (Operating system, version and so on): Windows for development, Linux for deployment
- .NET Framework version: .NET Core 1.1
- Additional information:

JimBobSquarePants:
Use Gitter please

antonfirsov:
Yeah gitter is our forum for initial feature discussions.

A short answer though:
You can workaround this by reading your network/file stream into `MemoryStream` asynchronously, then pass it to `Image.Load()`. Our initial implementation would be the same, implementing fine-grain async stream processing in codecs is not a key priority for at the moment. It's extremely hard for a format like jpeg, we need to make our  jpeg decoder faster & more memory friendly first.

PR #2180, opened 2022-07-17
Introduce Shared General Decoder Options plus Specialization

JimBobSquarePants:
### Prerequisites

### Description

Based on the discussion #2091 I've refactored our decoding API to do a few things.

- Introduced shared general decoder options
- Introduced specialized decoder specific options
- Introduced a general API for resizing images on decode.
- Introduced a new base `ImageDecoder<T>` class where `T` is `ISpecializedDecoderOptions` which allows external implementations to easily implement decoders using our new resize-on-decode strategy.
- Removed the ability to pass a decoder to `Image.Load(...)` as it was leading to too many user errors based on incorrect encoding assumptions.
- Removed all array overloads to our load APIs now that we have implicit `ReadonlySpan<T>` conversion

The general decoder options look like this. 

```csharp
/// <summary>
/// Provides general configuration options for decoding image formats.
/// </summary>
public sealed class DecoderOptions
{
    private static readonly Lazy<DecoderOptions> LazyOptions = new(() => new());

    private uint maxFrames = int.MaxValue;

    /// <summary>
    /// Gets the shared default general decoder options instance.
    /// </summary>
    public static DecoderOptions Default { get; } = LazyOptions.Value;

    /// <summary>
    /// Gets or sets a custom Configuration instance to be used by the image processing pipeline.
    /// </summary>
    public Configuration Configuration { get; set; } = Configuration.Default;

    /// <summary>
    /// Gets or sets the target size to decode the image into.
    /// </summary>
    public Size? TargetSize { get; set; } = null;

    /// <summary>
    /// Gets or sets a value indicating whether to ignore encoded metadata when decoding.
    /// </summary>
    public bool SkipMetadata { get; set; } = false;

    /// <summary>
    /// Gets or sets the maximum number of image frames to decode, inclusive.
    /// </summary>
    public uint MaxFrames { get => this.maxFrames; set => this.maxFrames = Math.Min(Math.Max(value, 1), int.MaxValue); }
}
```

The options replace Configuration as the primary parameter in all our decoding APIs

``` c#
/// <summary>
/// Decode a new instance of the <see cref="Image"/> class from the given stream.
/// </summary>
/// <param name="options">The general decoder options.</param>
/// <param name="stream">The stream containing image information.</param>
/// <exception cref="ArgumentNullException">The options are null.</exception>
/// <exception cref="ArgumentNullException">The stream is null.</exception>
/// <exception cref="NotSupportedException">The stream is not readable or the image format is not supported.</exception>
/// <exception cref="UnknownImageFormatException">Image format not recognised.</exception>
/// <exception cref="InvalidImageContentException">Image contains invalid content.</exception>
/// <returns>A new <see cref="Image"/>.</returns>
public static Image Load(DecoderOptions options, Stream stream)
```

I've deliberately chosen not to allow passing `ResizeOptions` to the decoder as I want to leave a clear path between general and specialist pipelines. Default resize-on-decode uses `ResizeMode.Max` with the `KnownResamplers.Box`.

An example `ISpecializedDecoderOptions` implementation looks like the following.

```csharp
/// <summary>
/// Configuration options for decoding Windows Bitmap images.
/// </summary>
public sealed class BmpDecoderOptions : ISpecializedDecoderOptions
{
    /// <inheritdoc/>
    public DecoderOptions GeneralOptions { get; set; } = new();

    /// <summary>
    /// Gets or sets the value indicating how to deal with skipped pixels,
    /// which can occur during decoding run length encoded bitmaps.
    /// </summary>
    public RleSkippedPixelHandling RleSkippedPixelHandling { get; set; }
}
```

Usage is allowed only on decoder instances via individual `IImageDecoderSpecialized<T>` implementations

```csharp
/// <summary>
/// Decodes the image from the specified stream to an <see cref="Image{TPixel}"/> of a specific pixel type.
/// </summary>
/// <param name="options">The specialized decoder options.</param>
/// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
/// <returns>The <see cref="Image{TPixel}"/>.</returns>
/// <exception cref="ImageFormatException">Thrown if the encoded image contains errors.</exception>
public abstract Image<TPixel> DecodeSpecialized<TPixel>(T options, Stream stream, CancellationToken cancellationToken)

/// <summary>
/// Decodes the image from the specified stream to an <see cref="Image"/> of a specific pixel type.
/// </summary>
/// <param name="options">The specialized decoder options.</param>
/// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
/// <returns>The <see cref="Image{TPixel}"/>.</returns>
/// <exception cref="ImageFormatException">Thrown if the encoded image contains errors.</exception>
public abstract Image DecodeSpecialized(T options, Stream stream, CancellationToken cancellationToken);
```

All but the `JpegDecoder` utilize the same resizing strategy for specialized methods as the default loading behavior. `JpegDecoder` will only downscale to the nearest size that is supported by the decoder allowing users to apply their own resize behavior in advanced scenarios. CC @br3aker


JimBobSquarePants:
I must have broken something with the last commit. I promise I had all tests passing! 😝

br3aker:
Phenomenal, will try to review this week!

gfoidl (commented):
Just scrolled through the changes, some nits.
Didn't read the xml-comments and tests though.

src/ImageSharp/Formats/DecoderOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to ignore encoded metadata when decoding.
+        /// </summary>
+        public bool SkipMetadata { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets the maximum number of image frames to decode, inclusive.
+        /// </summary>
+        public uint MaxFrames { get => this.maxFrames; set => this.maxFrames = Math.Min(Math.Max(value, 1), int.MaxValue); }
```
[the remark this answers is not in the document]
JimBobSquarePants: I'm stuck in the past! 😁

src/ImageSharp/Image.FromStream.cs
```diff
@@ -558,7 +401,7 @@ public static Image<TPixel> Load<TPixel>(Configuration configuration, Stream str
             var sb = new StringBuilder();
```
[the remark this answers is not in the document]
JimBobSquarePants: Good idea. I'll do it now.

br3aker (commented):
Looks good!
Glad that new API allows users to decode to target size in format-agnostic way.

src/ImageSharp/Formats/ImageDecoder{T}.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <summary>
+        /// Decodes the image from the specified stream to an <see cref="Image{TPixel}"/> of a specific pixel type.
+        /// </summary>
+        /// <typeparam name="TPixel">The pixel format.</typeparam>
+        /// <param name="options">The specialized decoder options.</param>
+        /// <param name="stream">The <see cref="Stream"/> containing image data.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <returns>The <see cref="Image{TPixel}"/>.</returns>
+        /// <exception cref="ImageFormatException">Thrown if the encoded image contains errors.</exception>
+        public abstract Image<TPixel> DecodeSpecialized<TPixel>(T options, Stream stream, CancellationToken cancellationToken)
```
antonfirsov, line 67: `DecodeSpecialized<T>` doesn't look like a real abstraction, it's rather a pattern. We never call it though the abstract base method. We could just define the methods in decoder subclasses without overriding anything.

I think that using inheritance and generics to enforce a pattern is a wrong design decision. Can't refer to any design guideline paragraphs on this, but I'm not aware of any precedents in the BCL aspnet, or other libraries.

src/ImageSharp/Formats/Bmp/BmpDecoderOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace SixLabors.ImageSharp.Formats.Bmp
+{
+    /// <summary>
+    /// Configuration options for decoding Windows Bitmap images.
+    /// </summary>
+    public sealed class BmpDecoderOptions : ISpecializedDecoderOptions
+    {
+        /// <inheritdoc/>
+        public DecoderOptions GeneralOptions { get; set; } = new();
```
antonfirsov, line 12: This is the only specialized decoder option class that actually contains a feature. All the others are dummies, their only purpose to satisfy the pattern introduced. I believe this is alarming.

src/ImageSharp/Formats/DecoderOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Gets or sets a custom Configuration instance to be used by the image processing pipeline.
+        /// </summary>
+        public Configuration Configuration { get; set; } = Configuration.Default;
+
+        /// <summary>
+        /// Gets or sets the target size to decode the image into.
+        /// </summary>
+        public Size? TargetSize { get; set; } = null;
```
antonfirsov, line 30: It looks like we are missing tests validating various usages of `TargetSize` (decode to specific size & keep aspect ratio). These tests can be also used as codes sample to ask some community members to do a final sanity-check on the target size decoding API.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -201,48 +206,33 @@ public static JpegFileMarker FindNextFileMarker(BufferedReadStream stream)
         /// <inheritdoc/>
         public Image<TPixel> Decode<TPixel>(BufferedReadStream stream, CancellationToken cancellationToken)
             where TPixel : unmanaged, IPixel<TPixel>
-            => this.Decode<TPixel>(stream, targetSize: null, cancellationToken);
-
-        /// <inheritdoc/>
-        public IImageInfo Identify(BufferedReadStream stream, CancellationToken cancellationToken)
         {
-            this.ParseStream(stream, spectralConverter: null, cancellationToken);
+            using var spectralConverter = new SpectralConverter<TPixel>(this.configuration, this.Options.GeneralOptions.TargetSize);
```
antonfirsov, line 210: It would be very important to have some tests validating how resize works together with #2076, stressing different dimensions including ones that are not divisible by 8. /cc @br3aker

antonfirsov (changes requested):
IMHO we should iterate on the design there are just too many alarming points that do not look good. One format doesn't justify introducing and maintaining a complicated and questionable pattern.

If we cannot come up with a good idea, I would just remove all the "specialized decoder options", and add a single special method to `BmpDecoder` for now. Maybe we will be able to come up with something better in the future.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
br3aker: Shouldn't be too hard as reference results are just scaled input images with box filter applied. I can create a PR when actual API is set in stone - you gave a lot to think of (in a positive way)!

src/ImageSharp/Formats/DecoderOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+    public sealed class DecoderOptions
+    {
+        private static readonly Lazy<DecoderOptions> LazyOptions = new(() => new());
+
+        private uint maxFrames = int.MaxValue;
+
+        /// <summary>
+        /// Gets the shared default general decoder options instance.
+        /// </summary>
+        public static DecoderOptions Default { get; } = LazyOptions.Value;
```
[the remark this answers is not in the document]
JimBobSquarePants: Yep agreed.

src/ImageSharp/Formats/DecoderOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to ignore encoded metadata when decoding.
+        /// </summary>
+        public bool SkipMetadata { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets the maximum number of image frames to decode, inclusive.
+        /// </summary>
+        public uint MaxFrames { get => this.maxFrames; set => this.maxFrames = Math.Clamp(value, 1, int.MaxValue); }
```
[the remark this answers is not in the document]
JimBobSquarePants: This is from the initial discussion at #2091 I wasn't 100 sure on it but nobody had argued against it and it was easy enough to implement. A `Range` might be more useful but I think anything too complicated would be best left to frame collection.

src/ImageSharp/Formats/ImageDecoder{T}.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// The base class for all image decoders.
+    /// </summary>
+    /// <typeparam name="T">The type of specialized decoder options.</typeparam>
+    public abstract class ImageDecoder<T> : IImageInfoDetector, IImageDecoder
+        where T : class, ISpecializedDecoderOptions, new()
+    {
+        /// <inheritdoc/>
+        public IImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
+        {
+            T specializedOptions = new() { GeneralOptions = options };
```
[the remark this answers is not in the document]
JimBobSquarePants: `Activator.CreateInstance<T>()` was optimized for .NET Core 3.0

Relevant issue/fix/benchmarks here. In the bigger scheme of things I imagine the numbers would be negligible. 

https://github.com/dotnet/runtime/issues/6801
https://github.com/dotnet/coreclr/pull/25145
https://github.com/dotnet/performance/pull/567

I chose the constraint specifically be cause I wanted to allow simple creation from our known type.

src/ImageSharp/Formats/ImageDecoder{T}.cs
JimBobSquarePants: I'm not following... The abstract base class calls it from `Decode`.

https://github.com/SixLabors/ImageSharp/pull/2180/files/085ff1ae465e67e27a80dae0601a56ae3149c66d#diff-7618afaee324d1c3d35802c1d55049a6226f471b8f73295c454e80d8f01e5290R29-R30

src/ImageSharp/Formats/Bmp/BmpDecoderOptions.cs
JimBobSquarePants: I wanted something that was guaranteed to handle some of the more exotic codes that we'll end up implementing so that we don't have to further breaks. I can imagine both AVIF and Jpeg-XL requiring customization. 

I'm actually very surprised we don't already have something for Tiff and I believe that is most likely an oversight thus far.

src/ImageSharp/Formats/DecoderOptions.cs
JimBobSquarePants: You mean validating the result of given dimensions? I would have thought that would already be covered by `ResizeHelperTests`? We simply pass the size to that and `JpegDecoderCore` already has tests for scaled decoding.

JimBobSquarePants:
Thanks @antonfirsov for the thorough review. I'll make the default options internal for now but I want to discuss further before making other changes to be sure we're on the same page.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
JimBobSquarePants: Thanks @br3aker 👍

I didn't add any specific ones because we had discussed the output quality vs performance tradeoffs in #2076 . Output seemed adequate and perf was far too negatively affected when using other resamplers.

src/ImageSharp/Formats/ImageDecoder{T}.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        /// <summary>
+        /// Reads the raw image information from the specified stream.
+        /// </summary>
+        /// <param name="options">The specialized decoder options.</param>
+        /// <param name="stream">The <see cref="Stream"/> containing image data.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <returns>The <see cref="IImageInfo"/> object.</returns>
+        /// <exception cref="ImageFormatException">Thrown if the encoded image contains errors.</exception>
+        public abstract IImageInfo IdentifySpecialized(T options, Stream stream, CancellationToken cancellationToken);
```
JimBobSquarePants, line 56: I'm going to see if I can refactor to remove this. There should be no need for specialized options during identification.

src/ImageSharp/Formats/DecoderOptions.cs
antonfirsov: I mean end-to-end testing resize-decode the way users will work with the feature, including aspect-ratio preservation. The only setter call I see for `DecoderOptions.TargetSize` is in `LoadResizeSaveStressRunner`.

Even though currently it's implemented with a simple `Resize`, it's a separate feature that needs testing.  Later it could be optimized by utilizing resizer internals (feeding the image to `ResizeWorker` row-by-row).

src/ImageSharp/Formats/ImageDecoder{T}.cs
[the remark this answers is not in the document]
antonfirsov: Even with those optimizations, `new T()` is still a sugar for `Activator.CreateInstance<T>()` which is practically reflection. I consider it a risk with all the AOT platforms out there.

src/ImageSharp/Formats/Bmp/BmpDecoderOptions.cs
antonfirsov: Even with Bmp + Tiff + AVIF we would have too much of boilerplate of generics and unnecessary empty classes IMO. And Tiff + AVIF are purely hypothetical at the moment, we don't have a proof this feature really needed & will be in fact implemented in the near future. I would prefer a design, where we could YAGNI this, and postpone creating specialized types until the moment they are proven to be actually needed for a format.

On top of that, I realized now, that we are creating an API disparity between decoders and encoders:
Decoders are becoming stateless objects with this PR with separate types for decoding options, while encoder objects carry the encoding options.


I know that I was an advocate of stateless decoders/encoders, but now I disagree with my past self and it feels like the simplest thing would be to move the format specific options back to the decoder objects, so users could do specialized decoding just:

```C#
var decoder = new BmpDecoder { RleSkippedPixelHandling = RleSkippedPixelHandling.Transparent };
using Image img = decoder.Decode(...);
```


Alternatively, we can also consider turning the current abstaction into a pattern, but this would not solve the decoder-encoder API disparity:

```C#
public class ImageDecoder : IImageDecoder {
    public Image<TPixel> Decode<TPixel>(DecoderOptions options, Stream stream, CancellationToken cancellationToken);
}

public class BmpDecoderOptions{
    public DecoderOptions GeneralOptions { get; set; }
}

public BmpDecoder : ImageDecoder {
    public Image<TPixel> DecodeSpecialized(BmpDecoderOptions options, Stream stream, CancellationToken cancellationToken);
}

```

src/ImageSharp/Formats/ImageDecoder{T}.cs
antonfirsov: I missed this. Still not big fan of the design, since fact the specialized method (and as a consequence the abstraction) is unnecessary for many formats, see next comment.

src/ImageSharp/Formats/DecoderOptions.cs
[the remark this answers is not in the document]
antonfirsov: This feels like a tiny detail now, would come back to it when we have consensus about the big picture.

src/ImageSharp/Formats/DecoderOptions.cs
br3aker: > I only see unit tests for SpectralConverter.CalculateResultingImageSize. Or am I missing something?

Yep, I've left actual image tests to the moment when API is ready as jpeg resize-decode entry point was somewhat of a stub. As that was internal we didn't risk exposing invalid code to users but this PR is a game changer. I know this solution isn't 'synchronized' but I'll write proper test suite for jpeg after API is set in stone.

antonfirsov: I'm fine delaying it for jpeg, the implementation is somewhat special anyways, but we could add test for another decoder (eg. PNG) in this PR. Then later we can extend as we optimize decoders.

src/ImageSharp/Formats/Bmp/BmpDecoderOptions.cs
JimBobSquarePants: > On top of that, I realized now, that we are creating an API disparity between decoders and encoders:
Decoders are becoming stateless objects with this PR with separate types for decoding options, while encoder objects carry the encoding options.

I think we'll end up doing something similar to encoders anyway. 

How would you enforce the options pattern for specialized decoders? I really want to ensure that for specialization we are forced into following something. 

The problem I see with the pattern above is that now you actually have more boilerplate for non-specialized decoding since you now have to implement the resize functionality for each concrete `Decode` implementation.

src/ImageSharp/Formats/ImageDecoder{T}.cs
JimBobSquarePants: I'll see what I can do but it's tricky. There's no concrete implementation for the core decoders so I can't implement a single abstract way to ensure the decoder honors the resize operation without reimplementing each time.

src/ImageSharp/Formats/ImageDecoder{T}.cs
JimBobSquarePants: I think I'd just make this abstract and pass `DecoderOptions`.

src/ImageSharp/Formats/Bmp/BmpDecoderOptions.cs
JimBobSquarePants: Maybe we can revert the specialized options and make the decoder stateful again?

`DecodeSpecialized` is simply the concrete implementation and each decoder then becomes responsible for passing addition options to the core.

antonfirsov: > you now have to implement the resize functionality for each concrete `Decode` implementation. 

We have to do it anyways, having a streaming + resizing, low-memory decoding path would be preferable introduce for each codec in 3.0. 

> I can't implement a single abstract way to ensure the decoder honors the resize operation without reimplementing each time.

We can move the current unoptimized implementation to some internal helper method to avoid copypasta.

> Maybe we can revert the specialized options and make the decoder stateful again?

That's basically the first suggestion in my comment. I wonder if we have the same understanding though?

src/ImageSharp/Formats/ImageDecoder{T}.cs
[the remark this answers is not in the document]
JimBobSquarePants: I think it's AOT safe now but I'll see if I can avoid it.

src/ImageSharp/Formats/Bmp/BmpDecoderOptions.cs
JimBobSquarePants: I think I understand where you are going. 

This is where I'm currently sitting with an implementation that has custom values and resizes during specialization. 

My only concern here is the separation of options into two separate types. It feels a little clunky 

```c#
/// <summary>
/// Image decoder for generating an image out of a Windows bitmap stream.
/// </summary>
public class BmpDecoder : ImageDecoder
{
    /// <summary>
    /// Gets or sets the value indicating how to deal with skipped pixels,
    /// which can occur during decoding run length encoded bitmaps.
    /// </summary>
    public RleSkippedPixelHandling RleSkippedPixelHandling { get; set; }

    /// <inheritdoc/>
    public override IImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
    {
        Guard.NotNull(stream, nameof(stream));

        return new BmpDecoderCore(this, options).Identify(options.Configuration, stream, cancellationToken);
    }

    /// <inheritdoc/>
    public override Image<TPixel> DecodeSpecialized<TPixel>(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
    {
        Guard.NotNull(stream, nameof(stream));

        BmpDecoderCore decoder = new(this, options);
        Image<TPixel> image = decoder.Decode<TPixel>(options.Configuration, stream, cancellationToken);

        Resize(options, image);

        return image;
    }

    /// <inheritdoc/>
    public override Image DecodeSpecialized(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
        => this.DecodeSpecialized<Rgba32>(options, stream, cancellationToken);
}
```

JimBobSquarePants: I'm gonna try your pattern above.

JimBobSquarePants: OK.... So here's a play on your pattern...

I think this actually solves all our problems. 

- Uses abstraction.
- Add specialized implementation only where it is needed.
- Avoids `Activator`
- Groups options types.

Note. I'm not enforcing a generic type for `ImageDecoderSpecialized<T>` so that we can pass `DecoderOptions` as the generic parameter in the `JpegDecoder` implementation since it requires no additional properties yet requires specialist behavior (_no additional resize_).

```c#
/// <summary>
/// Image decoder for generating an image out of a Windows bitmap stream.
/// </summary>
public class BmpDecoder : ImageDecoderSpecialized<BmpDecoderOptions>
{
    /// <inheritdoc/>
    public override IImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
    {
        Guard.NotNull(options, nameof(options));
        Guard.NotNull(stream, nameof(stream));

        return new BmpDecoderCore(new() { GeneralOptions = options }).Identify(options.Configuration, stream, cancellationToken);
    }

    /// <inheritdoc/>
    public override Image<TPixel> Decode<TPixel>(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
        => this.DecodeSpecialized<TPixel>(new() { GeneralOptions = options }, stream, cancellationToken);

    /// <inheritdoc/>
    public override Image Decode(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
        => this.DecodeSpecialized(new() { GeneralOptions = options }, stream, cancellationToken);

    /// <inheritdoc/>
    public override Image<TPixel> DecodeSpecialized<TPixel>(BmpDecoderOptions options, Stream stream, CancellationToken cancellationToken)
    {
        Guard.NotNull(options, nameof(options));
        Guard.NotNull(stream, nameof(stream));

        Image<TPixel> image = new BmpDecoderCore(options).Decode<TPixel>(options.GeneralOptions.Configuration, stream, cancellationToken);

        Resize(options.GeneralOptions, image);

        return image;
    }

    /// <inheritdoc/>
    public override Image DecodeSpecialized(BmpDecoderOptions options, Stream stream, CancellationToken cancellationToken)
        => this.DecodeSpecialized<Rgba32>(options, stream, cancellationToken);
```

```c#
/// <summary>
/// The base class for all specialized image decoders.
/// </summary>
/// <typeparam name="T">The type of specialized options.</typeparam>
public abstract class ImageDecoderSpecialized<T> : ImageDecoder
{
    /// <summary>
    /// Decodes the image from the specified stream to an <see cref="Image{TPixel}"/> of a specific pixel type.
    /// </summary>
    /// <param name="options">The specialized decoder options.</param>
    /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
    /// <returns>The <see cref="Image{TPixel}"/>.</returns>
    /// <exception cref="ImageFormatException">Thrown if the encoded image contains errors.</exception>
    public abstract Image<TPixel> DecodeSpecialized<TPixel>(T options, Stream stream, CancellationToken cancellationToken)

    /// <summary>
    /// Decodes the image from the specified stream to an <see cref="Image"/> of a specific pixel type.
    /// </summary>
    /// <param name="options">The specialized decoder options.</param>
    /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
    /// <returns>The <see cref="Image{TPixel}"/>.</returns>
    /// <exception cref="ImageFormatException">Thrown if the encoded image contains errors.</exception>
    public abstract Image DecodeSpecialized(T options, Stream stream, CancellationToken cancellationToken);
}
```

```c#
/// <summary>
/// The base class for all image decoders.
/// </summary>
public abstract class ImageDecoder : IImageInfoDetector, IImageDecoder
{
    /// <inheritdoc/>
    public abstract IImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken);

    /// <inheritdoc/>
    public abstract Image<TPixel> Decode<TPixel>(DecoderOptions options, Stream stream, CancellationToken cancellationToken)

    /// <inheritdoc/>
    public abstract Image Decode(DecoderOptions options, Stream stream, CancellationToken cancellationToken);

    /// <summary>
    /// Performs a resize operation against the decoded image. If the target size is not set, or the image size
    /// already matches the target size, the image is untouched.
    /// </summary>
    /// <param name="options">The decoder options.</param>
    /// <param name="image">The decoded image.</param>
    protected static void Resize(DecoderOptions options, Image image)
    {
        if (ShouldResize(options, image))
        {
            ResizeOptions resizeOptions = new()
            {
                Size = options.TargetSize.Value,
                Sampler = KnownResamplers.Box,
                Mode = ResizeMode.Max
            };

            image.Mutate(x => x.Resize(resizeOptions));
        }
    }

    /// <summary>
    /// Determines whether the decoded image should be resized.
    /// </summary>
    /// <param name="options">The decoder options.</param>
    /// <param name="image">The decoded image.</param>
    /// <returns><see langword="true"/> if the image should be resized, otherwise; <see langword="false"/>.</returns>
    private static bool ShouldResize(DecoderOptions options, Image image)
    {
        if (options.TargetSize is null)
        {
            return false;
        }

        Size targetSize = options.TargetSize.Value;
        Size currentSize = image.Size();
        return currentSize.Width != targetSize.Width && currentSize.Height != targetSize.Height;
    }
}
```

JimBobSquarePants: `BmpDecoder` is now stateless but I think I'm ok with that.

antonfirsov: After a very quick look: 
- I don't see the need for `ImageDecoderSpecialized<T>` base type. Am I missing something?
- I would actually rename `DecodeSpecialized(BmpDecoderOptions ...)` to `Decode(BmpDecoderOptions ...)`. The argument in the overload makes it obvious it's a codec-specific implementation, I don't feel the need for the verbose name.

> `BmpDecoder` is now stateless but I think I'm ok with that.

If we go with that, we should explore how would a similar separation look like in encoders, and how would we feel about it. (Stateless encoder + option object.)

antonfirsov:
One more concern:

We should think a bit more about the specialized decoder cases from user perspective. https://github.com/SixLabors/ImageSharp/discussions/2091#discussioncomment-3036607 considers it done with the following:

```C#
new JpegDecoder().DecodeSpecialized(JpegDecoderOptions ...)
```

However:
- `DecodeSpecialized` has a `CancellationToken` argument which is a weird thing to be expected from the user on a synchronous API.
- There is no async variant for `Decode` which can lead to complains similar to #272
(And in general, `IImageDecoder` / `ImageDecoder` types were designed to be subclassed / implemented and not called directly.)

I see two solutions for this problem:
1. Keep `ImageDecoderSpecialized<T>` as proposed in https://github.com/SixLabors/ImageSharp/pull/2180#discussion_r935729010 and defining additional convenience and async overloads. Or equivalently: defining an `IImageDecoderSpecialized<T>` interface with extension methods
2. Move back specialized options to stateful decoders, and let the existing `Image.Load` and `Image.LoadAsync` logic deal with the problem.

src/ImageSharp/Formats/Bmp/BmpDecoderOptions.cs
antonfirsov: Actually, it can be useful as a potential solution to the problem described in https://github.com/SixLabors/ImageSharp/pull/2180#issuecomment-1203297049.

JimBobSquarePants: >I don't see why is the ImageDecoderSpecialized<T> base type needed. Am I missing something?

Yeah, I guess that could be simply an interface unless we implement overloads. The `ImageDecoder` actually handles the meat now. 

>I don't feel the need for the verbose name.

The reason it's verbose is so I can do this with the `JpegDecoder`. We need a method that allows IDCT resize only and creating a `JpegDecoderOptions` type as you've identified is not great. 

```c#
/// <summary>
/// Decoder for generating an image out of a jpeg encoded stream.
/// </summary>
public sealed class JpegDecoder : ImageDecoderSpecialized<DecoderOptions>
{
    /// <inheritdoc/>
    public override IImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
    {
        Guard.NotNull(stream, nameof(stream));

        using JpegDecoderCore decoder = new(options);
        return decoder.Identify(options.Configuration, stream, cancellationToken);
    }

    /// <inheritdoc/>
    public override Image<TPixel> Decode<TPixel>(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
    {
        Image<TPixel> image = this.DecodeSpecialized<TPixel>(options, stream, cancellationToken);

        Resize(options, image);

        return image;
    }

    /// <inheritdoc/>
    public override Image Decode(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
    {
        Image image = this.DecodeSpecialized(options, stream, cancellationToken);

        Resize(options, image);

        return image;
    }

    /// <inheritdoc/>
    /// <remarks>
    /// Unlike <see cref="IImageDecoder.Decode{TPixel}(DecoderOptions, Stream, CancellationToken)"/>, when
    /// <see cref="DecoderOptions.TargetSize"/> is passed, the codec may not be able to scale efficiently to
    /// the exact scale factor requested, so returns a size that approximates that scale.
    /// Upscaling is not supported, so the original size will be returned.
    /// </remarks>
    public override Image<TPixel> DecodeSpecialized<TPixel>(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
    {
        using JpegDecoderCore decoder = new(options);
        return decoder.Decode<TPixel>(options.Configuration, stream, cancellationToken);
    }

    /// <inheritdoc/>
    /// <remarks>
    /// Unlike <see cref="IImageDecoder.Decode(DecoderOptions, Stream, CancellationToken)"/>, when
    /// <see cref="DecoderOptions.TargetSize"/> is passed, the codec may not be able to scale efficiently to
    /// the exact scale factor requested, so returns a size that approximates that scale.
    /// Upscaling is not supported, so the original size will be returned.
    /// </remarks>
    public override Image DecodeSpecialized(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
        => this.DecodeSpecialized<Rgb24>(options, stream, cancellationToken);
}
```

JimBobSquarePants:
>Move back specialized options to stateful decoders, and let the existing `Image.Load` and `Image.LoadAsync` logic deal with the problem.

Every time I look at stateful decoders mixed with general options it looks ugly. plus I really want to avoid passing a decoder to either of the `Load/LoadAsync` methods. It's a footgun for the inexperienced.

Regarding that cancellation token.... Yeah. Would be great to avoid it but we'd need a way to make `Identify`, `Decode<T>`, and `Decode` hidden from general consumption callable only by the methods on `Image`. This isn't a new problem though, it's possible today to spin up a decoder and decode an image using the `Decode` method directly with that token exposed.

br3aker:
> has a `CancellationToken` argument which is a weird thing to be expected from the user on a synchronous API

Why? From MSDN:

> A [CancellationToken](https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-6.0) enables cooperative cancellation between threads, thread pool work items, or [Task](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task?view=net-6.0) objects.

Image decoding may be a very long-running operation. If user needs/wants to quit early - why not allow it rather than forcing users to wait for a result user no longer needs?

antonfirsov:
> A [CancellationToken](https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-6.0) enables cooperative cancellation **between** threads, thread pool work items, or [Task](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task?view=net-6.0) objects.

This wording implies to me that `CancellationToken` was designed for multi-threaded / asynchronous scenarios. (When you do a sync call there aren't really any threads, thread pool work items or tasks involved, apart from the single thread that's doing the call.) There aren't any great examples where synchronous / blocking API-s take `CancellationToken`. When dotnet/runtime#32125 was designed, one of the main concerns was that proposal being the precedence that unleashes the beast of synchronous API-s with a `CancellationToken`.

> One way to hide the methods of course is to remove the `IImageDecoder` and `IImageInfoDetector` interfaces and favor abstract base classes. 

I'm not sure if it's worth to bother with this, but we really don't want the cancellable sync methods to be the only overloads users can call. I would still consider exploring ways to expose facade methods similar to `Image.Load` and `Image.LoadAsync`, if we want to avoid the stateful decoders + general options mess. (The outcome might be worse than additional overloads right on the decoder or interface+extensions, but might worth a try IMO.)

src/ImageSharp/Formats/Bmp/BmpDecoderOptions.cs
antonfirsov: >  We need a method that allows IDCT resize only 
> ` JpegDecoder : ImageDecoderSpecialized<DecoderOptions>`

To me it looks very unfortunate to use the general `DecoderOptions` as specialized options just to tell that a user wants an `IDCT` resize only. If I would look for such API the last thing I would expect that I just simply need to call a method called `DecodeSpcialized`.

I would rather introduce `JpegDecoderOptions` with a `bool IdctResize` property, or simply keep the "IDCT resize only" feature out of the public API. It's a niche use-case users would only use if they are unhappy with our default `TargetSize` decoding and need better control on the Resize. (Which IMHO would mean that we are doing something wrong and maybe we should reconsider exposing `ResizeOptions`.)

antonfirsov:
> Every time I look at stateful decoders mixed with general options it looks ugly. plus I really want to avoid passing a decoder to either of the `Load/LoadAsync` methods. It's a footgun for the inexperienced.

We can also consider stuff like:

```C#
class Image
{
    public static Image DecodeBmp(BmpDecoderOptions options, ...);
}
```

JimBobSquarePants:
```c#
class Image
{
    public static Image DecodeBmp(BmpDecoderOptions options, ...);
}
```

Unfortunately I don't think that's an improvement over passing a decoder to `Image.Load(...)`. 

I really want clean separation of advanced decoder operations from the general API. I've spotted more instances where people naïvely decode files because the extension is wrong than I can handle and I want to obscure that option so that experts only use it.  

I'll have another crack at the API tomorrow night.

br3aker:
> This wording implies to me that CancellationToken was designed for multi-threaded / asynchronous scenarios.

I personally understand that wording as it's impossible to cancel synchronous operation without an additional thread (which is obvious of course). IMO it's sad that concept of cancellation tokens became tightly coupled with async calls over the years...

But I understand your point, most likely end users would prefer simplier sync API. Just got curious about the reasoning behind it.

src/ImageSharp/Formats/Bmp/BmpDecoderOptions.cs
antonfirsov: If we want maximum flexibility, I would consider something like this for Jpeg:

```C#
public enum JpegResizeMode
{
    Combined, // This is what we do today
    IdctOnly, // IDCT-only to nearest block scale
    ScaleOnly // Opt-out the IDCT part and only Resize. Can be useful in case of quality concerns.
}

public class JpegDecoderOptions
{
    public DecoderOptions GeneralOptions { get; set; }
    public JpegResizeMode ResizeMode { get; set; } = JpegResizeMode.Combined;
}
```

JimBobSquarePants: OK. I love this!

JimBobSquarePants:
To remove the cancellation token from the public API we need to remove `IImageDecoder` and use an abstract class.

antonfirsov:
I'm not suggesting to remove it from the Decoder API. That's the best way for us to implement the cancellation, so makes total sense to pass it to decoder implementations.

The concern I raised was simply:
- For a *caller* of our public API-s, that should not be the only overload users can call.
- Convenience overloads (or facades) and `async` variants are needed anyways.

JimBobSquarePants:
@antonfirsov I've tried to follow your suggestions as closely as I can so please take another look. The API is a lot more sane now IMO. I still need to add tests though.

antonfirsov:
Cool! Will do another round of review on Monday.

src/ImageSharp/Formats/Jpeg/JpegDecoderResizeMode.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    public enum JpegDecoderResizeMode
+    {
+        /// <summary>
+        /// Both <see cref="IdctOnly"/> and <see cref="ScaleOnly"/>.
+        /// </summary>
+        Combined,
+
+        /// <summary>
+        /// IDCT-only to nearest block scale.
```
br3aker, line 18: I suppose it's worth mentioning that quality of this mode is close to the 'Box' resampling/scaling.

src/ImageSharp/Formats/ImageDecoderUtilities.cs
```diff
@@ -7,12 +7,55 @@
...
+        /// <param name="options">The decoder options.</param>
+        /// <param name="image">The decoded image.</param>
+        public static void Resize(DecoderOptions options, Image image)
+        {
+            if (ShouldResize(options, image))
+            {
+                ResizeOptions resizeOptions = new()
+                {
+                    Size = options.TargetSize.Value,
+                    Sampler = KnownResamplers.Box,
```
br3aker, line 32: I believe we should make resampler type configurable to mimic MagicScaler functionality.

JimBobSquarePants: I guess so, could easily add it to the general options. My original thinking though was that if they wanted more control they could use the processor.

JimBobSquarePants: I've added it.

br3aker: > My original thinking though was that if they wanted more control they could use the processor.

Totally missed that possibility, gonna need some sleep.
But I think all-in-one API may still be superior than two-step API.
Would like to hear Anton on this subject.

antonfirsov: The the target-size decoder is a memory-optimized version of the two step decode-resize workflow. In an ideal world I would feature-match them, so we don't force the user to choose between perf and flexibility. However I understand that it also makes sense to keep `DecoderOptions` as simple as possible. I think adding Resampler makes sense, we'll se if users need more.

antonfirsov (approved):

antonfirsov (commented):
Approving the general direction, since I couldn't spot anything principally wrong after a quick look, but don't have time to do a full scale review now.

I'm leaving to a vacation tomorrow with no computer access at all, and will be back next Monday. Would prefer to do a final "nitpicking" review early next week if possible, we need to add tests anyways before merging. Sorry that it's going this slow ...

antonfirsov:
Almost forgot the most important question regarding the proposed design:

What are our plans for the encoders? Do we plan to remove the disparity? If yes how? If not: are we really OK decoder and encoder design being very different?

JimBobSquarePants:
>What are our plans for the encoders? Do we plan to remove the disparity? If yes, can we put together a rough API proposal to make sure we are going to the right direction? If not: are we really OK decoder and encoder design being very different?

I'm not sure yet. I think I can make them similar but I don't want to block the API improvements here figuring it out.

JimBobSquarePants:
OK.... Added a bunch of extra tests. 

Had another look at the encoder story and stateless with options looks like a sensible pattern we can adopt also. It allows us to ensure only registered encoders are used (currently it's possible to allow passing unregistered).

If I can get some eyes on this branch again please I'd be very grateful. I'm very keen to move on.

JimBobSquarePants:
I'm gonna push on with this. Lot's to do.

PR #2189, opened 2022-07-25
Remove obsolete code

JimBobSquarePants:
### Prerequisites

### Description

Removes all obsolete conditions, obsolete methods, and fixes all compiler warnings.


src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
```diff
@@ -84,19 +82,17 @@ public static class SRgbCompanding
...
-#if SUPPORTS_RUNTIME_INTRINSICS
             if (Avx2.IsSupported && vectors.Length >= 2)
             {
                 CompandAvx2(vectors, ExpandTable);
 
                 if (Numerics.Modulo2(vectors.Length) != 0)
                 {
                     // Vector4 fits neatly in pairs. Any overlap has to be equal to 1.
-                    Expand(ref MemoryMarshal.GetReference(vectors.Slice(vectors.Length - 1)));
+                    Expand(ref MemoryMarshal.GetReference(vectors[^1..]));
```
gfoidl, line 92: The old version may be more performant, unfortunately. This is tracked in https://github.com/dotnet/roslyn/issues/43598, and you can see the difference in [this sharplab](https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoBBmABM3AYVwG9NcL88BlABwhgB5gBPAFwFMA+XAWRQAUdBs3bdcAZ3owAlLgC8PKQwDaAPRQA6TQF0A3OUo5cwpq049eSIdNEXJ0uYocNN1ADZQAxhwHKYmgAyHDAA5mwAFrgAtLgoMgYYAL5AA=).

If it's not perf-sensitive, then got with the new code, as it's cleaner to read.
If Roslyn tackles that issue, perf will increase "for free".

src/ImageSharp/Formats/Jpeg/Components/Block8x8.cs
```diff
@@ -271,12 +267,7 @@ public void TransposeInplace()
             // row #6
             Swap(ref Unsafe.Add(ref elemRef, 55), ref Unsafe.Add(ref elemRef, 62));
 
-            static void Swap(ref short a, ref short b)
-            {
-                short tmp = a;
-                a = b;
-                b = tmp;
-            }
+            static void Swap(ref short a, ref short b) => (b, a) = (a, b);
```
gfoidl, line 270: Just as info:
Resulting machine code will be the same, but the IL-size is with the tuple-based approach a bit larger than with the local (old version).

I don't think it matters here, but potential drawbacks
* JIT may give up inlining if any budget may exceed earlier --> add `MethodImpl.AggressiveInlining` (or the ImageSharp-equivalent) to be on the safe side?
* on products where IL size matters this could have a negative effect (TBH I don't know if two bytes more or less count here, as the reduction overall is much bigger that this PR does, but just to point it out)

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1042,7 +1041,7 @@ private void ReadCompressedTextChunk(ImageMetadata baseMetadata, PngMetadata met
             }
 
             int zeroIndex = data.IndexOf((byte)0);
-            if (zeroIndex < PngConstants.MinTextKeywordLength || zeroIndex > PngConstants.MaxTextKeywordLength)
+            if (zeroIndex is < PngConstants.MinTextKeywordLength or > PngConstants.MaxTextKeywordLength)
```
gfoidl, line 1044: Just FYI: https://github.com/dotnet/roslyn/issues/60534

I'd leave the code as is, as it's basically the same as now just nicer to read. When Roslyn got that addressed the perf-goodness comes "for free".

src/ImageSharp/Formats/Webp/Lossy/QuantEnc.cs
```diff
@@ -77,9 +75,7 @@ public static void PickBestIntra16(Vp8EncIterator it, ref Vp8ModeScore rd, Vp8Se
 
                 if (mode == 0 || rdCur.Score < rdBest.Score)
                 {
-                    Vp8ModeScore tmp = rdCur;
-                    rdCur = rdBest;
-                    rdBest = tmp;
+                    (rdBest, rdCur) = (rdCur, rdBest);
```
gfoidl, line 78: Use the swap-method from above (where I commented) instead (needs to be made generic then)?
Or drop the swap-method and write the one-line directly?

I don't have a strong opinion what's better here, but it should be consistent.
Leaning towards swap-method, as if there's a super nice trick once on how to swap more efficiently only one place needs to be updated.

gfoidl (commented):
Changes per-se LGTM, but I expect there will be some perf-regression. See comment about Span.Slice vs Range operator.
Maybe you can write a comment on the Roslyn-issue to help them prioritize that one 😉

src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
JimBobSquarePants: 🤦‍♂️ This is so frustrating. Why introduce an analyser that promotes worse behaviour?

src/ImageSharp/Formats/Jpeg/Components/Block8x8.cs
JimBobSquarePants: Another poor analyser suggestion 😔

src/ImageSharp/Formats/Webp/Lossy/QuantEnc.cs
JimBobSquarePants: 👍 good idea

src/ImageSharp/Formats/Jpeg/Components/Block8x8.cs
gfoidl: I'm no compiler engineer*, but for me this pattern is easy to recognize, so "optimal" code should be emitted, even if Roslyn isn't a optimizing compiler (that's the role of the JIT), but regarding IL-size optimizations count towards Roslyn.

\* so I have somewhat limited understanding on how things work there

src/ImageSharp/Formats/Webp/Lossy/Vp8EncIterator.cs
```diff
@@ -754,7 +755,9 @@ public void MakeChroma8Preds()
 
         public void SwapOut()
         {
+#pragma warning disable IDE0180 // Use tuple to swap values
```
dlemstra, line 758: Maybe we could add an extra overload to `RuntimeUtility.Swap`?

src/ImageSharp/Common/Helpers/RuntimeUtility.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// Swaps the two references.
+        /// </summary>
+        /// <typeparam name="T">The type to swap.</typeparam>
+        /// <param name="a">The first item.</param>
+        /// <param name="b">The second item.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void Swap<T>(ref T a, ref T b)
+        {
+            // Tuple swap uses 2 more IL bytes
+#pragma warning disable IDE0180 // Use tuple to swap values
```
dlemstra, line 24: Maybe do this on file level instead of the two spots where this is happening?

src/ImageSharp/Formats/Webp/Lossy/Vp8EncIterator.cs
JimBobSquarePants: You mean one that takes a `T[]`? The problem I found was that

```c#
RuntimeUtility.Swap(this.YuvOut, this.YuvOut2);
```

Would not work and creating local field references requires a lot more code than we have already.

src/ImageSharp/Common/Helpers/RuntimeUtility.cs
JimBobSquarePants: Good shout

antonfirsov (commented):
Looks good, but somewhat concerned about `var` rules.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.Generated.tt
```diff
@@ -13,10 +13,6 @@
 
 // <auto-generated />
 
-<#
-// Note use of MethodImplOptions.NoInlining. We have tests that are failing on certain architectures when
```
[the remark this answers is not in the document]
JimBobSquarePants: Yep. All fixed now 😀

src/ImageSharp/Common/Helpers/SimdUtils.ExtendedIntrinsics.cs
```diff
@@ -160,12 +153,10 @@ internal static void ByteToNormalizedFloat(ReadOnlySpan<byte> source, Span<float
                     Vector<uint> w2 = ConvertToUInt32(f2);
                     Vector<uint> w3 = ConvertToUInt32(f3);
 
-                    Vector<ushort> u0 = Vector.Narrow(w0, w1);
-                    Vector<ushort> u1 = Vector.Narrow(w2, w3);
+                    var u0 = Vector.Narrow(w0, w1);
```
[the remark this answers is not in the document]
JimBobSquarePants: I’ve really struggled to get editorconfig to do what I want here (var rules are actually unchanged). This seems to fall under the “type is apparent” rule. There’s actually been a warning sitting in the code there for years, it just was ignored.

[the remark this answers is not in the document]
JimBobSquarePants: Why arrays aren’t considered apparent I simply do not know.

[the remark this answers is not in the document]
antonfirsov: I don't think it's "apparent" when a non-generic type returns a variable with a generic type of the same name. The lack of the type argument hurts readability, especially with API-s like `Vector.Narrow`. Is there a way to fine-tune this rule? What if we disable it entirely?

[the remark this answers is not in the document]
JimBobSquarePants: I'm gonna create an empty project and see if I can get the rules working using the VS editorconfig editor and a fresh file.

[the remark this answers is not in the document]
JimBobSquarePants: Here's var turned off everywhere. It's verbose but at least it's consistent.

image

Here's with "when type is apparent"

This one is annoying because the array declaration should be apparent. It's probably the closes to what we want though.

image

Here's "when type is apparent" and "built in types"

This one is annoying because the double and int primitives are not apparent in my opinion since they require the reader to know the syntax declaration rules on the right hand side.

image

Which do you prefer?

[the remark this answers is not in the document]
antonfirsov: I may be very oldscool, but for me the first one is the lesser evil. I wonder what do others think @SixLabors/core?

This rule/analyzer ([IDE0007?](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0007-ide0008)) really sucks. I wonder if there is a better way to raise awareness about it than posting on https://developercommunity.visualstudio.com/ which is a straight way to oblivion.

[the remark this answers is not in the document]
JimBobSquarePants: I agree and with the forcing of target type new we handle known types. I'll implement that.

There's actually 2 issues in Roslyn around this. (One closed as a duplicate)

https://github.com/dotnet/roslyn/issues/23714
https://github.com/dotnet/roslyn/issues/55722

I wish I could do this though. It's really annoying that I can't seem to find a way.
```c#
byte[] b = [10];
```

src/Directory.Build.props
```diff
@@ -21,6 +21,10 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <CodeAnalysisRuleSet>..\ImageSharp.ruleset</CodeAnalysisRuleSet>
```
dlemstra, line 25: I tend to only use forward slashes now because Windows now also supports that.

src/ImageSharp/Formats/Png/PngHeader.cs
```diff
@@ -116,7 +116,7 @@ public void Validate()
         /// <param name="buffer">The buffer to write to.</param>
         public void WriteTo(Span<byte> buffer)
         {
-            BinaryPrimitives.WriteInt32BigEndian(buffer.Slice(0, 4), this.Width);
+            BinaryPrimitives.WriteInt32BigEndian(buffer[..4], this.Width);
             BinaryPrimitives.WriteInt32BigEndian(buffer.Slice(4, 4), this.Height);
```
dlemstra, line 120: Did you keep this a `Slice` on purpose?

src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
```diff
@@ -668,62 +654,62 @@ public static void FTransformWht(Span<short> input, Span<short> output, Span<int
         // luma 16x16 prediction (paragraph 12.3).
         public static void EncPredLuma16(Span<byte> dst, Span<byte> left, Span<byte> top)
         {
-            DcMode(dst.Slice(I16DC16), left, top, 16, 16, 5);
-            VerticalPred(dst.Slice(I16VE16), top, 16);
-            HorizontalPred(dst.Slice(I16HE16), left, 16);
-            TrueMotion(dst.Slice(I16TM16), left, top, 16);
+            DcMode(dst[..], left, top, 16, 16, 5);
```
dlemstra, line 657: `I16DC16` seems to be missing here?

src/ImageSharp/PixelFormats/PixelImplementations/PixelOperations/Rgb24.PixelOperations.cs
```diff
@@ -34,7 +34,7 @@ internal partial class PixelOperations : PixelOperations<Rgb24>
                 int count = redChannel.Length;
                 GuardPackFromRgbPlanes(greenChannel, blueChannel, destination, count);
 
-                SimdUtils.PackFromRgbPlanes(configuration, redChannel, greenChannel, blueChannel, destination);
+                SimdUtils.PackFromRgbPlanes(redChannel, greenChannel, blueChannel, destination);
```
dlemstra, line 37: It looks like we can also remove the `configuration` argument in the next major release?

dlemstra (commented):
Made a couple extra commits and I think we are good to go now for a merge.

src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
JimBobSquarePants: I'll re read the change.

src/ImageSharp/Formats/Png/PngHeader.cs
JimBobSquarePants: It wasn't suggested by the IDE. There's situations where is wont be due to performance differences. Maybe this is one of them?

src/ImageSharp/Color/Color.cs
```diff
@@ -286,14 +287,10 @@ public TPixel ToPixel<TPixel>()
...
         /// <typeparam name="TPixel">The pixel type to convert to.</typeparam>
-        /// <param name="configuration">The configuration.</param>
         /// <param name="source">The source color span.</param>
         /// <param name="destination">The destination pixel span.</param>
         [MethodImpl(InliningOptions.ShortMethod)]
-        public static void ToPixel<TPixel>(
-            Configuration configuration,
-            ReadOnlySpan<Color> source,
-            Span<TPixel> destination)
+        public static void ToPixel<TPixel>(ReadOnlySpan<Color> source, Span<TPixel> destination)
```
dlemstra, line 293: This is a breaking change in the Api? Should we not limit this PR to only remove non breaking changes?

src/ImageSharp/ColorSpaces/Conversion/Implementation/Converters/HunterLabToCieXyzConverter.cs
```diff
@@ -9,15 +9,15 @@ namespace SixLabors.ImageSharp.ColorSpaces.Conversion
     /// <summary>
     /// Color converter between <see cref="HunterLab"/> and <see cref="CieXyz"/>
     /// </summary>
-    internal sealed class HunterLabToCieXyzConverter : CieXyzAndHunterLabConverterBase
+    internal class HunterLabToCieXyzConverter : CieXyzAndHunterLabConverterBase
```
dlemstra, line 12: Why no longer sealed?

src/ImageSharp/Color/Color.cs
antonfirsov: I added `Configuration` to the parameter list to enable the optimization of this method by utilizing the `PixelOperations<TPixel>.FromRgba64()` batch conversion method. The abstract `PixelOperations<TPixel>` methods take `Configuration` for the case the implementation decides to refer to `Configuration.MemoryAllocator` to allocate temporary buffers.

If we don't like this pattern, we can open an issue to re-evaluate the design, after carefully analyzing our current conversion methods, which no longer need special logic for .NET Framework, though we may need to account for non-Intel, non-coreclr targets.

But I would avoid making sneaky public API changes in a big refactor PR. Is this the only one ore is there more?

src/ImageSharp/ColorSpaces/Conversion/Implementation/Converters/HunterLabToCieXyzConverter.cs
JimBobSquarePants: Accident. Will reverse

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Let's re-add it for now then and add a TODO in the code so intent is known. 

>Is this the only one ore is there more?
Some of the methods on `ColorSpaceConverter` are now static. I want to keep them like that though.

src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
JimBobSquarePants: No need for a slice at all so just pass `dst`

src/ImageSharp/Formats/Jpeg/Components/Decoder/ArithmeticScanDecoder.cs
```diff
@@ -306,15 +306,16 @@ private ArithmeticStatistics CreateOrGetStatisticsBin(bool dc, int identifier, b
...
-            var statistic = new ArithmeticStatistics(dc, identifier);
+            ArithmeticStatistics statistic = new(dc, identifier);
             this.statistics.Add(statistic);
             return statistic;
         }
 
         private void ParseBaselineData()
         {
-            foreach (ArithmeticDecodingComponent component in this.components)
+            for (int i = 0; i < this.components.Length; i++)
```
dlemstra, line 316: Why was this changed? Performance?

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -751,7 +746,7 @@ private void ProcessApplicationHeaderMarker(BufferedReadStream stream, int remai
             stream.Read(this.temp, 0, JFifMarker.Length);
             remaining -= JFifMarker.Length;
 
-            JFifMarker.TryParse(this.temp, out this.jFif);
+            _ = JFifMarker.TryParse(this.temp, out this.jFif);
```
dlemstra, line 749: Should we not handle this?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1590,18 +1561,17 @@ private PngChunkType ReadChunkType()
...
-            {
-                PngThrowHelper.ThrowInvalidChunkType();
 
-                // The IDE cannot detect the throw here.
-                return default;
-            }
+            PngThrowHelper.ThrowInvalidChunkType();
+
+            // The IDE cannot detect the throw here.
+            return default;
```
dlemstra, line 1568: Do we still need this with the `DoesNotReturn` attribute?

src/ImageSharp/Formats/Webp/Lossy/WebpLossyDecoder.cs
```diff
@@ -252,18 +270,35 @@ private void ParseIntraMode(Vp8Decoder dec, int mbX)
...
                 }
             }
 
             // Hardcoded UVMode decision tree.
-            block.UvMode = (byte)(this.bitReader.GetBit(142) == 0 ? 0 :
-                           this.bitReader.GetBit(114) == 0 ? 2 :
-                           this.bitReader.GetBit(183) != 0 ? 1 : 3);
+            if (this.bitReader.GetBit(142) == 0)
+            {
+                // Hardcoded UVMode decision tree.
```
dlemstra, line 281: Do we really need this comment on every assignment?

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
```diff
@@ -64,11 +66,12 @@ private void GetThumbnail(uint offset)
                 return;
             }
 
-            var values = new List<IExifValue>();
+            List<IExifValue> values = new();
             this.ReadValues(values, offset);
 
-            foreach (ExifValue value in values)
+            for (int i = 0; i < values.Count; i++)
```
dlemstra, line 72: Also wondering why this was changed.

src/ImageSharp/Metadata/Profiles/ICC/MultiProcessElements/IccMatrixProcessElement.cs
```diff
@@ -52,13 +52,17 @@ public override bool Equals(IccMultiProcessElement other)
...
         /// <inheritdoc />
         public bool Equals(IccMatrixProcessElement other)
-        {
-            return this.Equals((IccMultiProcessElement)other);
-        }
+            => this.Equals((IccMultiProcessElement)other);
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+            => this.Equals(obj as IccMatrixProcessElement);
```
dlemstra, line 59: We could also directly do `IccMultiProcessElement` here?

src/ImageSharp/Metadata/Profiles/ICC/TagDataEntries/IccXyzTagDataEntry.cs
```diff
@@ -50,8 +47,13 @@ public override bool Equals(IccTagDataEntry other)
...
         /// <inheritdoc />
         public bool Equals(IccXyzTagDataEntry other)
-        {
-            return this.Equals((IccTagDataEntry)other);
-        }
+            => this.Equals((IccTagDataEntry)other);
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+            => this.Equals(obj as IccXyzTagDataEntry);
```
dlemstra, line 54: `IccTagDataEntry`?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: It shouldn’t (I don’t think) but I was still getting a warning 🤔

Maybe we need to turn on Nullable Reference Types in a future PR? (Which I would love)

src/ImageSharp/Formats/Webp/Lossy/WebpLossyDecoder.cs
JimBobSquarePants: I think so. I think the reference source is not hard coded and there’s scope for potential enhancement here

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
JimBobSquarePants: Because the compiler was complaining that an explicit cast was required for each item. The automatic fix was Linqs cast which was obviously poor from a perf perspective so a for loop was a better fix

src/ImageSharp/Metadata/Profiles/ICC/TagDataEntries/IccXyzTagDataEntry.cs
JimBobSquarePants: The cast is the correct type no?

src/ImageSharp/Metadata/Profiles/ICC/MultiProcessElements/IccMatrixProcessElement.cs
JimBobSquarePants: The cast is the correct type no?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
dlemstra: Would also love to have that enabled. Probably going to be another major PR 😉

src/ImageSharp/Metadata/Profiles/ICC/MultiProcessElements/IccMatrixProcessElement.cs
dlemstra: It is correct but the Equals that you are calling does a cast to `IccMultiProcessElement` so you could decide to directly cast instead. Same goes for my other comment.

JimBobSquarePants: Ah right, that depends on an implementation detail of the explicit equality comparer method (which I want to look into. Why are there separate types that do not vary by properties only property values?)

For safety I would always attempt a cast to the concrete type.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
JimBobSquarePants: Yeah, you're right. Now explicitly throwing `NotSupportedException` since we only support JFIF App0 markers.

JimBobSquarePants:
I'm gonna merge this now. All tests are passing, and I've covered any raised issues.

PR #2192, opened 2022-07-26
Relaxed bmp dimensions validation

smorokin:
### Prerequisites

### Description
Relaxed bmp header dimensions checks so now the maximum possible values are accepted.


src/ImageSharp/Formats/Bmp/BmpInfoHeader.cs
```diff
@@ -545,13 +555,19 @@ public void WriteV5Header(Span<byte> buffer)
 
         internal void VerifyDimensions()
         {
-            const int MaximumBmpDimension = 65535;
-
             if (this.Width > MaximumBmpDimension || this.Height > MaximumBmpDimension)
```
brianpopow, line 558: I dont think this check makes sense anymore since we are checking against size > 2147483647 and the width and height of an image is `int` and therefore cannot be greater then this. I think we should remove this `VerifyDimensions` now.

brianpopow: removed with 1acbfbe3

src/ImageSharp/Formats/Bmp/BmpInfoHeader.cs
```diff
@@ -66,6 +66,16 @@ internal struct BmpInfoHeader
         /// </summary>
         public const int HeaderSizeSize = 4;
 
+        /// <summary>
+        /// Maximum dimensions of a bitmap with or height: 2**31 - 1, since width and height are int32
+        /// </summary>
+        private const int MaximumBmpDimension = 2147483647;
```
dlemstra, line 72: This is not being used?

brianpopow: no, it's no longer used, it's now removed.

JimBobSquarePants (approved):
Thanks for finishing this off. 

When I first looked at this, I thought we'd be encoding using one of the older headers using unsigned width/height so thought it would be more complicated.

PR #2197, opened 2022-08-03
TGA Encoder/Decoder Improvements

brianpopow:
### Prerequisites

### Description

This is a follow up PR for #2172. It does a few things:
- Add tests for #2172
- Add checks to make sure enough data is available when decoding
- Change run length encoding to not always write RLE packets. This can be ineffective (add overhead) if there are no equal pixels.

### DIFF
--- a/src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
+++ b/src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
@@ -75,7 +75,7 @@ namespace SixLabors.ImageSharp.Formats.Tga
         /// <summary>
         /// Gets the dimensions of the image.
         /// </summary>
-        public Size Dimensions => new Size(this.fileHeader.Width, this.fileHeader.Height);
+        public Size Dimensions => new(this.fileHeader.Width, this.fileHeader.Height);
 
         /// <inheritdoc />
         public Image<TPixel> Decode<TPixel>(BufferedReadStream stream, CancellationToken cancellationToken)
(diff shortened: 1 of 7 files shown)


src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
```diff
@@ -624,8 +636,8 @@ private void ReadRle<TPixel>(int width, int height, Buffer2D<TPixel> pixels, int
                                 }
                                 else
                                 {
-                                    var alpha = alphaBits == 0 ? byte.MaxValue : bufferSpan[idx + 3];
-                                    color.FromBgra32(new Bgra32(bufferSpan[idx + 2], bufferSpan[idx + 1], bufferSpan[idx], (byte)alpha));
+                                    byte alpha = alphaBits == 0 ? byte.MaxValue : bufferSpan[idx + 3];
```
JimBobSquarePants, line 639: Why do we use Bgra here when the buffer is Rgba?

brianpopow: Because of the Intel byte ordering, the data needs to be interpreted as bgra. Quote from the [Spec](http://www.paulbourke.net/dataformats/tga/):

```
|  Image Data Field.                                         |
|                                                            |
|  This field specifies (width) x (height) pixels.  Each     |
|  pixel specifies an RGB color value, which is stored as    |
|  an integral number of bytes.                              |
|  The 2 byte entry is broken down as follows:               |
|  ARRRRRGG GGGBBBBB, where each letter represents a bit.    |
|  But, because of the lo-hi storage order, the first byte   |
|  coming from the file will actually be GGGBBBBB, and the   |
|  second will be ARRRRRGG. "A" represents an attribute bit. |
|  The 3 byte entry contains 1 byte each of blue, green,     |
|  and red.                                                  |
|  The 4 byte entry contains 1 byte each of blue, green,     |
|  red, and attribute.  For faster speed (because of the     |
|  hardware of the Targa board itself), Targa 24 images are  |
|  sometimes stored as Targa 32 images.                      |
--------------------------------------------------------------------------------
```

JimBobSquarePants: Thanks... I was thrown by the assignment order in the `Bgra32` constructor. Forgot the order was r,g,b,a.

JimBobSquarePants (approved):
LGTM. Thanks!

PR #2200, opened 2022-08-06
Refactor `EnumUtils` and make it AOT-friendly.

teo-tsirpanis:
### Prerequisites

- [ ] ~~I have provided test coverage for my change (where applicable)~~

### Description

The `EnumUtils.Parse` method used to use the non-generic `Enum.GetValues` method to do its job, which is unsafe and produces AOT warnings.
This PR refactors it to use `Enum.IsDefined` instead, which does not allocate.
I also added asserts to methods in `EnumUtils` that ensure we use them with int-sized enums.

### DIFF
--- a/src/ImageSharp/Common/Helpers/EnumUtils.cs
+++ b/src/ImageSharp/Common/Helpers/EnumUtils.cs
@@ -19,15 +19,14 @@ namespace SixLabors.ImageSharp
         /// <param name="defaultValue">The default value to return.</param>
         /// <returns>The <typeparamref name="TEnum"/>.</returns>
         public static TEnum Parse<TEnum>(int value, TEnum defaultValue)
-            where TEnum : Enum
+            where TEnum : struct, Enum
         {
-            foreach (TEnum enumValue in Enum.GetValues(typeof(TEnum)))
+            DebugGuard.IsTrue(Unsafe.SizeOf<TEnum>() == sizeof(int), "Only int-sized enums are supported.");
+
+            TEnum valueEnum = Unsafe.As<int, TEnum>(ref value);
+            if (Enum.IsDefined(valueEnum))
             {
-                TEnum current = enumValue;
-                if (value == Unsafe.As<TEnum, int>(ref current))
-                {
-                    return enumValue;
-                }
+                return valueEnum;
             }
 
             return defaultValue;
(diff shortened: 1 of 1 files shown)


src/ImageSharp/Common/Helpers/EnumUtils.cs
```diff
@@ -19,15 +20,14 @@ internal static class EnumUtils
         /// <param name="defaultValue">The default value to return.</param>
         /// <returns>The <typeparamref name="TEnum"/>.</returns>
         public static TEnum Parse<TEnum>(int value, TEnum defaultValue)
-            where TEnum : Enum
+            where TEnum : struct, Enum
         {
-            foreach (TEnum enumValue in Enum.GetValues(typeof(TEnum)))
+            Debug.Assert(Unsafe.SizeOf<TEnum>() == sizeof(int), "Only int-sized enums are supported.");
```
antonfirsov, line 25:
```suggestion
            DebugGuard.IsTrue(Unsafe.SizeOf<TEnum>() == sizeof(int), "Only int-sized enums are supported.");
```

We don't use `Debug.Assert`.

src/ImageSharp/Common/Helpers/EnumUtils.cs
```diff
@@ -41,8 +41,10 @@ public static TEnum Parse<TEnum>(int value, TEnum defaultValue)
         /// <param name="flag">The flag.</param>
         /// <returns>The <see cref="bool"/>.</returns>
         public static bool HasFlag<TEnum>(TEnum value, TEnum flag)
-          where TEnum : Enum
+          where TEnum : struct, Enum
         {
+            Debug.Assert(Unsafe.SizeOf<TEnum>() == sizeof(int), "Only int-sized enums are supported.");
```
antonfirsov, line 46:
```suggestion
            DebugGuard.IsTrue(Unsafe.SizeOf<TEnum>() == sizeof(int), "Only int-sized enums are supported.");
```

antonfirsov (commented):
Thanks!

src/ImageSharp/Common/Helpers/EnumUtils.cs
teo-tsirpanis: Fixed.

JimBobSquarePants:
Looks great. Thanks!

PR #2203, opened 2022-08-12
WIP: Fix issue capping xmp data at 65KB

### LINKED ISSUE #2202: Resizing from PNG to JPG is done poorly in 2.0.0 and higher. It worked alright in 1.0.4
### Prerequisites

### ImageSharp version

2.1.3

1.0.4

Windows 11

### .NET Framework version

6.0

### Description

The app I am working on needs to convert PNGs to JPGs as well as resize them.

Using the following code with the version 1.0.4 works (Images averaging 150 to 200Kb), the same code in 2.0.0 all the way to 2.1.3 (Images averaging 4.5 to 4.8Mb)

### Steps to Reproduce

byte[] inputFile <- Here goes the loaded image. My current code retrieves the images from an Azure Blob repo.

Image image = Image.Load<Rgba32>(inputFile);
image.Mutate(x => x.BackgroundColor(Color.White)); // I've tried with this line and without it, it causes no difference
image.Mutate(x => x.Resize(width, height));

int quality = 85;
while (quality > 0)
{
    using MemoryStream ms = new();
    await image.SaveAsJpegAsync(ms, new JpegEncoder { Quality = quality });
    if ((ms.Length / 1024) <= 150)
    {
        return (ms.ToArray(), null);
    }
    else
        quality -= 5;
}

brianpopow:
### Prerequisites

### Description

This PR fixes an issue capping the XMP data at 65KB.
Closes #2202

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
@@ -330,9 +330,6 @@ namespace SixLabors.ImageSharp.Formats.Jpeg
         /// Writes the XMP metadata.
         /// </summary>
         /// <param name="xmpProfile">The XMP metadata to write.</param>
-        /// <exception cref="ImageFormatException">
-        /// Thrown if the XMP profile size exceeds the limit of 65533 bytes.
-        /// </exception>
         private void WriteXmpProfile(XmpProfile xmpProfile)
         {
             if (xmpProfile is null)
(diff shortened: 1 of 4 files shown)


src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -352,26 +349,18 @@ private void WriteXmpProfile(XmpProfile xmpProfile)
             }
 
             int dataLength = data.Length;
-            int offset = 0;
 
-            while (dataLength > 0)
+            // The XMP data cannot be split in multiple segments, therefore cap the data at 65KB.
```
br3aker, line 353: [It actually can be split](https://github.com/adobe/xmp-docs/blob/master/XMPSpecifications/XMPSpecificationPart3.pdf):
> If the serialized XMP packet becomes larger than the 64K limit, you can divide it into a main portion
(StandardXMP) and an extended portion (ExtendedXMP), and store it in multiple JPEG marker segment.

brianpopow: Thanks for this info, I did not know that. My info came from https://dev.exiv2.org/projects/exiv2/wiki/The_Metadata_in_JPEG_files

Switching back to WIP.

br3aker: I think we can merge this PR as it is now to eliminate exceptions and provide split metadata implementation later.

brianpopow: After reviewing this again, I think the current implementation in `main` is correct: It splits already the XMP data over multiple APP1 markers. My assumption that it should be capped at 65KB was wrong. I think I will close this PR then.

If anyone else wants to double check this (looking at you @br3aker :-), feel free to do so. A second opinion is always helpful.

JimBobSquarePants: I think 65k is the cap for each segment. There's no limit to the actual size that I know of.

PR #2207, opened 2022-08-18
Fix Image.Identify/IdentifyAsync byte array/file overloads still referring to a stream argument in documentation

UnknownShadow200:
### Prerequisites

### Description

The `Image.Identify/IdentifyAsync` overloads that took byte arrays or file paths as the input argument still referred to them as a stream in their documentation

### DIFF
--- a/src/ImageSharp/Image.FromBytes.cs
+++ b/src/ImageSharp/Image.FromBytes.cs
@@ -41,7 +41,7 @@ namespace SixLabors.ImageSharp
         }
 
         /// <summary>
-        /// Reads the raw image information from the specified stream without fully decoding it.
+        /// Reads the raw image information from the given encoded byte array without fully decoding it.
         /// </summary>
         /// <param name="data">The byte array containing encoded image data to read the header from.</param>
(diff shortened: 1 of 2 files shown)


src/ImageSharp/Image.FromBytes.cs
```diff
@@ -41,7 +41,7 @@ public static IImageFormat DetectFormat(Configuration configuration, byte[] data
         }
 
         /// <summary>
-        /// Reads the raw image information from the specified stream without fully decoding it.
+        /// Reads the raw image information from the given encoded byte array without fully decoding it.
```
dlemstra, line 44: Are we also using `encoded` in other comments?

UnknownShadow200: For byte arrays, `IdentifyFormat` doesn't call them encoded in \<summary\> but `Load` does, so I went with the latter for consistency

Although the \<summary\> for streams/files doesn't mention `encoded` at all, so I can remove `encoded` from \<summary\> for byte arrays if preferable

JimBobSquarePants:
Replaced by #2317

Issue #814, opened 2019-01-21
Feature request: Add Median filter


dejanberic:
This is a simple filter which can be used to filter out noise.

Here is one paper on fast Median and Bilateral filtering:
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.93.1608&rep=rep1&type=pdf

michasacuer:
Hi, can i take and try to make this feature with Parallel tasks and Interop services? @JimBobSquarePants

michasacuer:
Referring to gitter talk witn @antonfirsov i take that issue. I will start implementing it in february

antonfirsov:
@michasacuer great news! Thanks! 👍

Jarak-Jakar:
Just in case it helps (probably not, but I don't think it can hurt to mention it) I was recently working on basic median filtering, i.e. simply taking the median of a square window.  You can see the most up-to-date version of the fastest program I came up with [here](https://github.com/Jarak-Jakar/ivcnz2018/blob/master/naive/Program.fs) - feel free to look at everything else in the repo, but I think that'll be by far the most useful part of it.  It is written in F# though, so might not be of as much use to you.  Moreover, I have undoubtedly done something silly, or somehow not used ImageSharp effectively.  I also didn't get around to implementing more interesting types of median filters.

Based on some quick profiling of my program, it appears to spend about 60% of its time in the `Array.sort` function.  I tried implementing some alternatives like Insertion Sort and Shell Sort (I did think about Radix sort, but I can't remember actually testing it), but nothing came close to being as fast as the built-in .NET sort once the arrays got bigger than about 10 elements.  Someone suggested that I move from sorting the array to using a selection algorithm like QuickSelect, but I never got around to implementing it/trying it out.

I hope that this helps 😃

michasacuer:
Thanks for help! I will start to implementing this after my exams (i think it will be int the next one, two weeks)

FransBouma:
Isn't this a 9 tap tent filter, as described by Smith? https://pdfs.semanticscholar.org/cdf7/1f6ea2178cfa210a30c5870c235f3acdc42a.pdf

JimBobSquarePants:
Fixed via #2219 !

PR #2219, opened 2022-08-30
Implement Median Blur processor

ynse01:
### Prerequisites

### Description

This PR aims to solve #814 by implementing a Median Blur processor using the .NET runtime Span.Sort(). Results are comparable to Gimp.


src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
```diff
@@ -0,0 +1,164 @@
...
+                    var index = 0;
+                    for (var w = 0; w < this.kernelSize; w++)
+                    {
+                        var j = yOffsets[baseYOffsetIndex + w];
+                        var row = this.sourcePixels.DangerousGetRowSpan(j);
+                        for (var z = 0; z < this.kernelSize; z++)
+                        {
+                            var k = xOffsets[baseXOffsetIndex + z];
+                            var pixel = row[k];
+                            kernelBuffer[index + z] = pixel.ToVector4();
```
JimBobSquarePants, line 81: Perhaps we should Make `ReadOnlyKernel` `Kernel` and add setters. That way you can use a `DenseMatrix<T>` and use all the existing APIs for Kernel access.

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
```diff
@@ -0,0 +1,164 @@
...
+                        }
+
+                        index += this.kernelSize;
+                    }
+
+                    targetBuffer[x - boundsLeft] = this.FindMedian3(kernelBuffer, xChannel, yChannel, zChannel, kernelCount);
```
JimBobSquarePants, line 87: Looks like we're not actually preserving the alpha here. In other blur operations we preserve the exact per-pixel alpha value of the source. See `ConvolutionProcessor<T>`

ynse01: I'm using the middle **input** pixel for the `W` channel, which boils down to a NO-OP. This preserves the W or alpha value. Added a comment to explain this.

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
ynse01: The `ConvolutionProcessor` uses a clever trick, where it builds up the target value row-by-row. This is allowed for `ReadOnlyKernel` type of `RowOperation` classes, where we're calculating weighted sums basically. These weighted sums can indeed be split arbitrary, as is done by clearing the target buffer before the loop and using `+=` in [line 152](https://github.com/SixLabors/ImageSharp/blob/af9bfe516eeded3e6a52a7e74cf531af86d00f56/src/ImageSharp/Processing/Processors/Convolution/ConvolutionProcessor%7BTPixel%7D.cs#L152)
However, the operation of finding a median, **cannot** be split up like this (or at least not without lots of caching). It mandates the traversal of the entire kernel in one pass, to get a single `Span` of pixel values to `Sort`. Hence the different logic here, compared to `ConvolutionProcessor`.

I'll update the code to use `Unsafe.Add` iso indexers.

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
```diff
@@ -0,0 +1,164 @@
...
+            int boundsWidth = this.bounds.Width;
+            int boundsRight = this.bounds.Right;
+            int kernelCount = this.kernelSize * this.kernelSize;
+            Span<Vector4> kernelBuffer = span.Slice(0, kernelCount);
+            Span<Vector4> channelVectorBuffer = span.Slice(kernelCount, kernelCount);
+            Span<Vector4> targetBuffer = span.Slice(kernelCount << 1, boundsWidth);
+
+            // Stack 4 channels of floats in the space of Vector4's.
+            Span<float> channelBuffer = MemoryMarshal.Cast<Vector4, float>(channelVectorBuffer);
+            var xChannel = channelBuffer.Slice(0, kernelCount);
```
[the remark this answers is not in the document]
ynse01: Actually, I did not get these as I'm using VSCode on Linux.

By default only Visual Studio will complain about the IDExxxx, or CodeStyle Rules. With .NET 5 you can also enable these in MsBuild which allows enforcement. See the `EnforceCodeStyleInBuild` option in [here](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview#enable-on-build). I'll leave the choice to configure this or not, up to the maintainers.

For now, I manually converted any `var` into the actual type.

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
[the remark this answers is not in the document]
ynse01: Because the `z` loops runs over neighboring rows, this buffer would need to contain at least several rows of the source image. 

Do you feel that is worth the memory?

JimBobSquarePants: Ah no, I didn't mean the processors themselves I just meant the `ConvolutionState` struct. It's designed to track all the offsets you're manually tracking yourself. `ReadOnlyKernel` could become `Kernel` with a read/write indexer (maybe ref based).

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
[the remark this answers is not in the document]
JimBobSquarePants: Interesting. I didn't know about that property. We could add it to the build props.

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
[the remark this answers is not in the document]
JimBobSquarePants: We can still read/convert the source row slice one at a time though if you're sampling one kernel row at a time?

src/ImageSharp/Processing/Processors/Convolution/MedianBlurProcessor{TPixel}.cs
```diff
@@ -0,0 +1,60 @@
...
+            MemoryAllocator allocator = this.Configuration.MemoryAllocator;
+            using Buffer2D<TPixel> targetPixels = allocator.Allocate2D<TPixel>(source.Width, source.Height);
+
+            source.CopyTo(targetPixels);
+
+            Rectangle interest = Rectangle.Intersect(this.SourceRectangle, source.Bounds());
+
+            // We use a rectangle with width set to 2 * kernelSize^2 + width, to allocate a buffer big enough
+            // for kernel source and target bulk pixel conversion.
+            Rectangle operationBounds = new Rectangle(interest.X, interest.Y, (2 * (kernelSize * kernelSize)) + interest.Width, interest.Height);
```
JimBobSquarePants, line 38: If you use 2x interest width here you can use the extra buffer space to allow bulk per-row `ToVector4()` against the source row in the operation.

```c#
new Rectangle(interest.X, interest.Y, (2 * (kernelSize * kernelSize)) + (2 * interest.Width), interest.Height);
```

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        public void Invoke(int y, Span<Vector4> span)
+        {
+            // Span has kernelSize^2 followed by bound width.
+            int boundsX = this.bounds.X;
+            int boundsWidth = this.bounds.Width;
+            int kernelCount = this.kernelSize * this.kernelSize;
+            Span<Vector4> kernelBuffer = span.Slice(0, kernelCount);
+            Span<Vector4> channelVectorBuffer = span.Slice(kernelCount, kernelCount);
+            Span<Vector4> targetBuffer = span.Slice(kernelCount << 1, boundsWidth);
```
JimBobSquarePants, line 54: Slice of a portion of the span here for the source bulk conversion buffer.

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+                for (int x = 0; x < boundsWidth; x++)
+                {
+                    int index = 0;
+                    ref int sampleColumnBase = ref state.GetSampleColumn(x);
+                    ref Vector4 target = ref Unsafe.Add(ref targetBase, x);
+                    for (int kY = 0; kY < state.Kernel.Rows; kY++)
+                    {
+                        int currentYIndex = Unsafe.Add(ref sampleRowBase, kY);
+                        Span<TPixel> sourceRow = this.sourcePixels.DangerousGetRowSpan(currentYIndex).Slice(boundsX, boundsWidth);
```
JimBobSquarePants, line 79: Use `PixelOperations` to convert the full source row here into our sliced source `Vector4` buffer.

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
```diff
@@ -0,0 +1,171 @@
...
+                        ref TPixel sourceRowBase = ref MemoryMarshal.GetReference(sourceRow);
+                        for (int kX = 0; kX < state.Kernel.Columns; kX++)
+                        {
+                            int currentXIndex = Unsafe.Add(ref sampleColumnBase, kX) - boundsX;
+                            TPixel pixel = Unsafe.Add(ref sourceRowBase, currentXIndex);
+                            state.Kernel.SetValue(index, pixel.ToVector4());
```
JimBobSquarePants, line 85: Now you can set the value straight from our source `Vector4` buffer

src/ImageSharp/Processing/Processors/Convolution/MedianBlurProcessor{TPixel}.cs
ynse01: Well, it's not that simple I think, as we need more rows for a single kernel.

I think I get your point about caching the `Vector4` converted pixels, I'll take the challenge of making it work!

ynse01: On the multiple source rows that Median needs, the conversion is now done with BulkOperations. I think this is what you meant.

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
ynse01: Done

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
ynse01: Done

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
ynse01: Done, while picking the correct source row.

src/ImageSharp/Processing/Processors/Convolution/MedianRowOperation{TPixel}.cs
ynse01: Made a separate MedianConvolutionState struct for this, as it's using `Vector4` iso `float`.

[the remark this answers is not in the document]
ynse01: Please see in the comments below how this is done.

JimBobSquarePants:
@ynse01 I'll have a good look at this ASAP. Just trying to get #2189 through first.

JimBobSquarePants (approved):
Looks great now! Thanks for persevering!

PR #2225, opened 2022-09-12
Make TestFile Image threadsafe.

### LINKED ISSUE #2178: Possible false positive memory leak reports in tests
### Prerequisites

### ImageSharp version

main branch

None

Windows 10

### .NET Framework version

net6

### Description

We have [`TestFile`](https://github.com/SixLabors/ImageSharp/blob/main/tests/ImageSharp.Tests/TestFile.cs) class for testing purposes. It works as a cache of some sort and has this line:

https://github.com/SixLabors/ImageSharp/blob/2c42e41ccffe7bf5123cdc0746707c07ca0e5523/tests/ImageSharp.Tests/TestFile.cs#L71

I think this is responsible for random false positive memory leaks in tests, e.g. what happens in this case if memory leaks diagnostics are on and if we would imagine this is the only test in out test suite?

```cs
[Theory]
public void Test(TestFile testFile)
{
    // testFile.CreateRgba32Image() creates 2 Image<Rgba32> objects
    // one is lazily initialized and stored in cache and one is created via Clone() call
    // and used in this test
    using (var image = testFile.CreateRgba32Image())
    {
        // Undisposed objects count should be 2 here
        this.Output.WriteLine(image.Size());
    }
    // Cloned image is disposed and TestFile instance image is not disposed
    // Thus undisposed count should be 1 here which can lead to false positive memory leak
}
```

And aren't these lines in `TestFile.cs`:

```cs
public byte[] Bytes => this.bytes ?? (this.bytes = File.ReadAllBytes(this.FullPath));

private Image<Rgba32> Image => this.image ?? (this.image = ImageSharp.Image.Load<Rgba32>(this.Bytes));
```

not thread safe?

`File.ReadAllBytes(this.FullPath)` can be called multiple times and won't leak as it returns managed array.
`this.image = ImageSharp.Image.Load<Rgba32>(this.Bytes)` can also be called multiple times but extra instance would just leak.

### Steps to Reproduce

This happens quite randomly so there's no 100% correct steps to reproduce it.

### Images

_No response_

JimBobSquarePants:
### Prerequisites

### Description

Fix #2178 

CC @br3aker

### DIFF
--- a/tests/ImageSharp.Tests/TestFile.cs
+++ b/tests/ImageSharp.Tests/TestFile.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.IO;
+using System.Threading;
 using SixLabors.ImageSharp.Advanced;
 using SixLabors.ImageSharp.Formats;
 using SixLabors.ImageSharp.PixelFormats;
@@ -29,7 +30,7 @@ namespace SixLabors.ImageSharp.Tests
         /// <summary>
         /// The image (lazy initialized value)
         /// </summary>
-        private Image<Rgba32> image;
+        private volatile Image<Rgba32> image;
 
         /// <summary>
         /// The image bytes
@@ -65,7 +66,28 @@ namespace SixLabors.ImageSharp.Tests
         /// <summary>
         /// Gets the image with lazy initialization.
         /// </summary>
-        private Image<Rgba32> Image => this.image ??= ImageSharp.Image.Load<Rgba32>(this.Bytes);
+        private Image<Rgba32> Image
+        {
+            get
+            {
+                Image<Rgba32> img = this.image;
+                if (img is null)
+                {
+                    Image<Rgba32> loadedImg = ImageSharp.Image.Load<Rgba32>(this.Bytes);
+                    img = Interlocked.CompareExchange(location1: ref this.image, value: loadedImg, comparand: null);
+                    if (img is not null)
+                    {
+                        loadedImg.Dispose();
+                    }
+                    else
+                    {
+                        img = loadedImg;
+                    }
+                }
+
+                return img;
+            }
+        }
 
         /// <summary>
         /// Gets the input image directory.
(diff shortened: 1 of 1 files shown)


tests/ImageSharp.Tests/TestFile.cs
```diff
@@ -65,7 +66,28 @@ public sealed class TestFile
...
         /// </summary>
-        private Image<Rgba32> Image => this.image ??= ImageSharp.Image.Load<Rgba32>(this.Bytes);
+        private Image<Rgba32> Image
+        {
+            get
+            {
+                Image<Rgba32> img = this.image;
+                if (img is null)
+                {
+                    Image<Rgba32> loadedImg = ImageSharp.Image.Load<Rgba32>(this.Bytes);
```
br3aker, line 76: Well, we can do a double check lock instead so image would never be loaded twice:
```cs
// this.image should be marked volatile still
if (this.image is null)
{
    lock(this.lockObject)
    {
        if (this.image is null)
        {
            this.image = ImageSharp.Image.Load<Rgba32>(this.Bytes);
        }
    }
}
return this.image;
```

I proposed `Interlocked` solution, I know, but `lock` solution is a lot more readable IMO and doesn't have a possible flaw of loading the same image twice. It's also isn't slow on reads due to locking as most of the time first `if` check would be true.

Maybe we should use it instead due to simplicity?

JimBobSquarePants: Definitely far more readable. I’m happy to switch out

tests/ImageSharp.Tests/TestFile.cs
```diff
@@ -29,7 +29,12 @@ public sealed class TestFile
         /// <summary>
         /// The image (lazy initialized value)
         /// </summary>
-        private Image<Rgba32> image;
+        private volatile Image<Rgba32> image;
+
+        /// <summary>
+        /// Used to ensure image loading is threadsafe.
+        /// </summary>
+        private readonly object syncLock;
```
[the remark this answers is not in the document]
JimBobSquarePants: Ha! Can't believe I missed that!

PR #2226, opened 2022-09-13
PNG filter improvements

saucecontrol:
### Prerequisites

### Description

Fixes a bug in the sums calculation for adaptive filtering:

https://github.com/SixLabors/ImageSharp/blob/ea97bac681f325e3ca4e1ecb87eb53d0048c1d25/src/ImageSharp/Formats/Png/Filters/AverageFilter.cs#L217

I can only guess this adjustment was compensating for a bug that existed in an earlier version of the code where it counted the row filter byte in the sum.  In any case, that is not happening now, so the adjustment at the end is taking away from a valid sum value.  Although this is generally harmless, it does mean that in the event of a tie for sum values (e.g. 0 in a solid color image), Paeth will win by virtue of its larger row filter number (-4 adjustment), making the resulting image needlessly expensive to decode.

Also tidies up the code and codegen for PNG filters in a few ways:
* Inlines the SIMD Paeth predictors for encoding, which were previously emitted as method calls for each input vector
* Simplifies fallback code for SSE2-only configs
* Eliminates some sign-extensions and extra arithmetic on row ref offsets
* Uses FilterTypes enum values for row filter byte instead of magic numbers.

I will follow up another time with some improvements to the SIMD decode filter implementations.  I have ideas 😉

### DIFF
--- a/src/ImageSharp/Common/Helpers/Numerics.cs
+++ b/src/ImageSharp/Common/Helpers/Numerics.cs
@@ -780,25 +780,13 @@ namespace SixLabors.ImageSharp
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static int ReduceSum(Vector128<int> accumulator)
         {
-            if (Ssse3.IsSupported)
-            {
-                Vector128<int> hadd = Ssse3.HorizontalAdd(accumulator, accumulator);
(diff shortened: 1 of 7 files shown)


JimBobSquarePants (approved):
Very cool. Much improved readability also. Thanks!

PR #2233, opened 2022-09-16
Use System.Half for HalfTypeHelper

JimBobSquarePants:
### Prerequisites

### Description

Replaces our incredibly complex custom packing code with `System.Half` plus casting.

### DIFF
--- a/src/ImageSharp/PixelFormats/HalfTypeHelper.cs
+++ b/src/ImageSharp/PixelFormats/HalfTypeHelper.cs
@@ -1,8 +1,7 @@
-﻿// Copyright (c) Six Labors.
+// Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
 
 using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
 
 namespace SixLabors.ImageSharp.PixelFormats;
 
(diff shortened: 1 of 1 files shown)


src/ImageSharp/PixelFormats/HalfTypeHelper.cs
```diff
@@ -19,126 +18,15 @@ internal static class HalfTypeHelper
...
-        public int I;
-
-        /// <summary>
-        /// The unsigned integer.
-        /// </summary>
-        [FieldOffset(0)]
-        public uint U;
-    }
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static float Unpack(ushort value) => (float)Unsafe.As<ushort, Half>(ref value);
```
teo-tsirpanis, line 31:
```suggestion
    internal static float Unpack(ushort value) => (float)BitConverter.UInt16BitsToHalf(value);
```

JimBobSquarePants: Interesting…. Why?

src/ImageSharp/PixelFormats/HalfTypeHelper.cs
```diff
@@ -19,126 +18,15 @@ internal static class HalfTypeHelper
...
-        }
-
-        if (e > 30)
-        {
-            return (ushort)(s | 0x7c00);
-        }
-
-        return (ushort)(s | (e << 10) | (m >> 13));
+        Half h = (Half)value;
+        return Unsafe.As<Half, ushort>(ref h);
```
teo-tsirpanis, line 22:
```suggestion
        return BitConverter.HalfToUInt16Bits((Half)value);
```

src/ImageSharp/PixelFormats/HalfTypeHelper.cs
teo-tsirpanis: It's the Lawful Good™ way to do it without involving `Unsafe`. I think Tanner had said that Unsafe.As is technically UB but don't remember for sure.

JimBobSquarePants: Cool. I'm up for that. Wasn't aware of the new APIs.

brianpopow (approved):
Much better now 👍

src/ImageSharp/PixelFormats/HalfTypeHelper.cs
tannergooding: Yep, `BitConverter` has "well-defined" behavior and will often generate better codegen.

PR #2235, opened 2022-09-17
Implement AutoLevel processor - Attempt 2

ynse01:
### Prerequisites

### Description

Implementation of #87

This PR can be interpreted as attempt 2 of the closed PR #1619. It tries to address the comments mentioned there. This processor ensures that the image remains RGB and no color distortion occurs. 

Some design decisions (comments welcome):
- Not touching the Alpha channel.
- Using BT.709 luminance, as this is being used in other Normalization processors already. Original PR talks about using LAB.
- Include the _sync channels_ option, as it was originally in [ImageMagick](https://github.com/ImageMagick/ImageMagick/blob/7ba113a0d26b6d2b008ce0fcf428050a84e5a0ac/MagickCore/histogram.c#L926).

Please comment if this is the correct functionality.

I'll address the bulk conversion To-do's in a later PR, as there are more (unrelated) classes that suffer from that issue.


JimBobSquarePants:
@ynse01 Thanks for having a crack at this. The code looks great 👍 

Do we have output references from ImageMagick we can compare our output with?

ynse01:
> @ynse01 Thanks for having a crack at this. The code looks great +1
> 
> Do we have output references from ImageMagick we can compare our output with?

Good point. Let me try and make an explicit Test for it.

ynse01:
The `HistogramEqualizationTests` seem to be unstable in CI on MacOS. I don't own a Mac, so I can't have a look in any detail.
Please advise.

JimBobSquarePants:
I’ll have a look. It’s likely not us. ImagMagick seems to return different results on Mac

JimBobSquarePants:
I pulled down the artifacts from the build at the bottom of this page here and compared to a local test run on my Windows machine. (I need to dig my Mac out of a box as I moved out recently)

https://github.com/SixLabors/ImageSharp/actions/runs/3117901063

Interestingly it appears that there is an issue but not with the Mac output. Look at the pixel comparison there. 
My output and the reference both suffer from what looks like overflow issues when calculating individual component values.

**Mac CI - Left, Windows Local Right**

image


**Reference Output - Left, Windows Local Right**

image

ynse01:
Thanks for the directions @JimBobSquarePants.
I found an off-by-one error in the code for the separate channel option and updated that reference image accordingly.

JimBobSquarePants (approved):
This looks great to me. Thanks!

Looking forward to a follow up PR to handle the bulk pixel bits.

I need to have a look at the other Histogram equalization methods tbh. As I recall they only work for grayscale images.

Issue #2231, opened 2022-09-16
Enable Nullable Reference Types


JimBobSquarePants:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

v3 alpha + 

### Other ImageSharp packages and versions

NA

### Environment (Operating system, version and so on)

NA

### .NET Framework version

NA

### Description

Following some initial investigation, it appears adding [Nullable Reference Types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references) to the solution is not as much work as I initially feared. 

- Only 824 errors were reported
- 119 can be ignored in AotCompilerTools by explicitly disabling the analysis there via `#nullable disable` 
- Most of the errors can be fixed by updating equality overrides which are missing attributes.

Enabling analysis requires adding the following to individual `.csproj` files

```xml
<NullableContextOptions>enable</NullableContextOptions>
<Nullable>enable</Nullable>
```
We should consider doing this for V3.

### Steps to Reproduce

NA

JimBobSquarePants:
Thanks for the update. I thought they were two separate properties.

tocsoft:
I'll drop the comment in here to, for high level general approach on how to take the task.

This issue will need to be tackled in a few phases.

1. **Enable nullability checks globally _(single PR)_**
    update `csproj`s but inline disable the check for any/all files that are reporting warning because of it.
    This enables us to get new code checked correctly and/or existing code that is already valid staying valid but doesn't require validating all the change in one big bang.
2. **Public API surface _(one or more PRs)_**
    for each part of the public api surface ensure the nullablity checks are enabled and issues fixed (limiting each PR to a single logic area where possible), for example each processor in turn and associated extensions.
3. **Internal APIs _(one or more PRs)_**
    general clean up, again in logic areas (JPEG Decoder, PNG Decoder etc) 
    After this stage no code should have nullable reference type checking disabled.

I believe this is the only logic approach to tacking this with the resource we have on the team for code reviews alone let alone apply all the fixes.

[...]

PR #2236, opened 2022-09-18
Enable Nullable ReferenceTypes

stefannikolei:
#2231

### DIFF
--- a/src/ImageSharp/Advanced/AdvancedImageExtensions.cs
+++ b/src/ImageSharp/Advanced/AdvancedImageExtensions.cs
@@ -99,7 +99,7 @@ public static class AdvancedImageExtensions
     /// <param name="source">The source image</param>
     /// <returns>Returns the bounds of the image</returns>
     private static Configuration GetConfiguration(IConfigurationProvider source)
-        => source?.Configuration ?? Configuration.Default;
+        => source.Configuration ?? Configuration.Default;
 
     /// <summary>
     /// Gets the representation of the pixels as a <see cref="IMemoryGroup{T}"/> containing the backing pixel data of the image
@@ -115,7 +115,7 @@ public static class AdvancedImageExtensions
     /// <exception cref="ArgumentNullException">Thrown when the <paramref name="source"/> in <see langword="null"/>.</exception>
     public static IMemoryGroup<TPixel> GetPixelMemoryGroup<TPixel>(this ImageFrame<TPixel> source)
         where TPixel : unmanaged, IPixel<TPixel>
-        => source?.PixelBuffer!.FastMemoryGroup.View ?? throw new ArgumentNullException(nameof(source));
+        => source.PixelBuffer!.FastMemoryGroup.View ?? throw new ArgumentNullException(nameof(source));
 
     /// <summary>
     /// Gets the representation of the pixels as a <see cref="IMemoryGroup{T}"/> containing the backing pixel data of the image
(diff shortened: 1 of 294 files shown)


src/ImageSharp/Metadata/Profiles/IPTC/IptcValue.cs
```diff
@@ -101,7 +101,7 @@ public string Value
                 if (this.Strict && value.Length > maxLength)
                 {
                     string cappedValue = value[..maxLength];
-                    valueBytes = this.encoding.GetBytes(cappedValue);
+                    valueBytes = this.encoding!.GetBytes(cappedValue);
```
JimBobSquarePants, line 104: I think we should avoid doing this. It looks like the `Encoding` backing field should be set to `Encoding.UTF8` by default.

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
```diff
@@ -148,7 +148,7 @@ public static Memory<TPixel> DangerousGetPixelRowMemory<TPixel>(this ImageFrame<
         Guard.MustBeGreaterThanOrEqualTo(rowIndex, 0, nameof(rowIndex));
         Guard.MustBeLessThan(rowIndex, source.Height, nameof(rowIndex));
 
-        return source.PixelBuffer.GetSafeRowMemory(rowIndex);
+        return source.PixelBuffer!.GetSafeRowMemory(rowIndex);
```
stefannikolei, line 151: @JimBobSquarePants Should I look for all places where I used ! and add an ArgumentNullException.ThrowIfNull?

stefannikolei:
@JimBobSquarePants I think you could have a first look. Let me know what you think and what parts should be changed.

tocsoft:
Hmm... there feels like there is way too much changed in this PR, this doesn't feel like the right way to tackly adding NRT into the codebase, it feel like it should be a much more targeted file by file/ section by section change, there is way to much expert knowledge required to apply some of these fixes globally, just deciding what shoudl be nullable Vs having more sensible defaults to just just changing `{get; set;}` to `{get; init;}`.

I would propose we close this huge PR (sorry about that @stefannikolei ) and instead replace it with a first PR of enable the csproj setting but inline opt out all files with warnings. This as a series of follow up PRs the various sections can be tackled.

This should me any new files/code will have it enabled by default and we can be more methodical in applying the right fixes as we go.

PS. We should also aim to tackle the public API surface first before diving into the internals.

JimBobSquarePants:
It's with a heavy heart that I have to agree with @tocsoft here. 

I started a review of the PR and while there is some low hanging fruit - `Equals( object)` - there are several places where we are going to have to significantly refactor the codebase in order to achieve safe nullability handling. 

This is my fault. I should have investigated further before encouraging anyone to attempt a single PR.

I'm so, so sorry that you have made so much effort here with so little result.

stefannikolei:
> It's with a heavy heart that I have to agree with @tocsoft here.
> 
> I started a review of the PR and while there is some low hanging fruit - `Equals( object)` - there are several places where we are going to have to significantly refactor the codebase in order to achieve safe nullability handling.
> 
> This is my fault. I should have investigated further before encouraging anyone to attempt a single PR.
> 
> I'm so, so sorry that you have made so much effort here with so little result.

Its ok. NRT is not an easy change. And it gets big really fast. And errors are there to be done :)

I would change it back to draft. Perhaps it is useful for anyone to look up when making the changes in smaller steps

JimBobSquarePants:
Thanks for understanding @stefannikolei

JimBobSquarePants:
Sorry I have broken so much! It was really important to get #2276 merged in as it was blocking development.

stefannikolei:
@JimBobSquarePants this is the old one. I will close it. To not confuse anyone. 

I will merge your changes in my branch and fix the problems

PR #2237, opened 2022-09-19
Replace ImageExtensions.Save.tt with Source Generator

CollinAlpert:
### Prerequisites

### Description

This PR removes the ImageExtensions.Save T4 template and replaces it with a Source Generator.
The generator (and all future generators) is placed in the `ImageSharp.Generators` project, since generators need to be in a separate project than the code they generate for.
I have also added `<LangVersion>10</LangVersion>` to the generator, in order to use features like file-scoped namespaces and such. It is completely safe to use a language version as high as the project which consumes the generator.

Everything still builds and it is still possible to navigate to source for the extension methods the generator replaces. Let me know what you think!


CollinAlpert:
I see the build is failing due to the fact that the `ImageSharp.Generators` project does not have a `net6.0` target. Where is this limitation coming from? SGs [must target netstandard2.0](https://github.com/dotnet/roslyn/issues/45162#issuecomment-645052356).

CollinAlpert:
I see! I will try to multi target the SG later and see what happens. Should work fine.

CollinAlpert:
That's fine, the generator will build and work correctly under .NET 6 as well. It just won't show up in IntelliSense, which doesn't matter for a CI build. This is why it is [discouraged](https://github.com/dotnet/roslyn/issues/47087#issuecomment-679280683) to use a higher version than `netstandard2.0`.

Sergio0694:
I don't understand the point of this PR. T4 generators are better suited for this purpose. Source generators are not a replacement for T4 generators and they're not inherently better, they're just meant for different use cases, and this doesn't look like one.

CollinAlpert:
@Sergio0694 This PR was meant as a demonstration for the ImageSharp maintainers so they could evaluate the maintenance costs of abandoning T4 templates.

Sergio0694:
Right, but I'm saying there's just no reason to do so at all. T4 templates are the right tool for this. Not to mention, using source generators with a local project reference is unsupported, and has broken tooling. If you clean the repository (eg. `git clean -fdx`) and open the solution, IntelliSense will be completely broken until you rebuild the source generator and close and reopen VS again. And even without this, using a source generator is just more maintenance effort for literally no gain here.

JimBobSquarePants:
Very curious to hear more expert opinion here. This is all very new to me!

Sergio0694:
> "can you elaborate what do you mean by this considering our use-cases?"

The advantage of source generators is that they can dynamically generate code based on what code you're using in your own project, but that really only makes sense when there's some kind of introspection being done on said code from the generator end. Say, you're using `System.Text.Json`, and the generator is inspecting your data models to generate serialization code for you based on those models. In your case instead, there's no introspection being done, and everything is just done in a generator initialization step. That is literally just a T4 template but with extra steps, and it's not really recommended because it just adds way more complexity for no benefit. Not to mention that, currently, there's really no proper tooling support for locally referenced generators, so working on them is a bit of a pain because IntelliSense is pretty much always broken.

The point is: source generators are _not_ meant to replace T4 templates, they're just meant for different scenarios. If you need to generate some fixed templated code, then a T4 template is a perfectly reasonable, and much simpler, choice. Eg. I'm heavily using T4 templates in ComputeSharp too to generate all the HLSL primitive projection types for C#, not generators.

In your case, eg. in those pixel operations files that Anton linked, that setup looks perfectly reasonable. If you wanted to make it more compact, you could just use an array of all pixel format types, and generate all the partial types in a single file 🙂

antonfirsov:
>  In your case instead, there's no introspection being done

This is how we implemented with T4, but *in theory* it would be more robust if we could enumerate & inspect the existing pixel types, and existing `Image` API-s in the generator instead -- at least this is how I imagined it before @Sergio0694 came to ruin my naïve world 😆

If I get the whole picture right, you basically say: use Source Generator if you *really* need a reusable tool to generate sources in a flexible manner. Use T4 if you want to generate code for a single project you control, otherwise you'll just end up adding *even more* complexity and maintenance burden.

I think we should stay with T4 then? Or are any counter-opinions in the community which are based on experience with bigger projects?

CollinAlpert:
Yeah, I definitely agree that SGs are meant for scenarios where more of the code is actually analyzed/introspected to generate the result. I was just under the impression that the tooling around T4 templates (across all operating systems and IDEs) is also not great and that this would cancel out the Source Generators' shortcomings.

Sergio0694:
> "Use T4 if you want to generate code for a single project you control, otherwise you'll just end up adding even more complexity and maintenance burden."

I would say that in some cases it might make sense to still use a source generator for internal use, as long as the benefit is worth the extra work, eg. if you have lots of different uses on random types where the introspection makes sense due to the high variety of scenarios that are supported, but even then it's something you should think about on a case by case basis.

One rule of thumb I can definitely give you though is: if your generator is only using `RegisterPostInitializationOutput`, then that's really just T4 with extra steps 😄

src/ImageSharp.Generators/ImageExtensionsSaveGenerator.cs
```diff
@@ -0,0 +1,165 @@
...
+
+[Generator(LanguageNames.CSharp)]
+internal class ImageExtensionsSaveGenerator : IIncrementalGenerator
+{
+    private const string FileHeader = @"// Copyright (c) Six Labors.
+// Licensed under the Six Labors Split License.
+
+// <auto-generated />";
+
+    private static readonly string[] ImageFormats =
```
JimBobSquarePants, line 18: Wouldn't it be better to have some sort of attribute on the encoder type that can be used to generate the methods instead?

That would make it more generator worthy would it not @Sergio0694 ?

It would be amazing if we could add that to a base class (ImageEncoder See #2269) and have these methods generated automatically without us ever having to touch this file again.

CollinAlpert: Yup, this would be possible and probably a better suited scenario for a Source Generator. You'd still have the issues around local projects with SGs which @Sergio0694 mentioned in https://github.com/SixLabors/ImageSharp/pull/2237#issuecomment-1252359522 and would need to decide if this is okay for you.

JimBobSquarePants: > Not to mention, using source generators with a local project reference is unsupported, and has broken tooling

This one? I'm still blown away by that. How does someone test such a thing if locally referencing the project is unsupported? It seems incredibly short sighted.

CollinAlpert: Yeah, to be honest I haven't noticed this as much. I rarely clean my repository like that.
I have multiple SGs in a local project at work and only get red squiggly lines when modifying the SG code, which rarely happens. Even then closing and reopening the solution fixes it. Other than that, all is good.

PR #2241, opened 2022-09-26
Extend row operation interfaces with buffer length method

ynse01:
### Prerequisites

### Description
Some implementation of `IRowOperation{T}` and `IRowIntervalOperation{T}` interfaces use buffer length other than the width of the bounds. Examples are most convolution processors.
This PR makes the required buffer length an explicit method in these interfaces. This introduces some boiler plate code when the `IRowOperation{T}` is using the default, but I think the readability of the code improves even there. When re-using `IRowOperation{T}` implementations in other processors, the _internal_ knowledge of the buffer length is no longer required. Last but not least, my personal trigger to start this PR: It eliminates contributors going back and forth between the RowOperation and the Processor when the buffer length is changed in the code.

Also, implemented buffered pixel conversions in Normalization processors using these extended interfaces.

I feel the existing test cover the changes in this PR already. Let me know if you feel otherwise.

### DIFF
--- a/src/ImageSharp/Advanced/IRowIntervalOperation{TBuffer}.cs
+++ b/src/ImageSharp/Advanced/IRowIntervalOperation{TBuffer}.cs
@@ -13,6 +13,13 @@ public interface IRowIntervalOperation<TBuffer>
     where TBuffer : unmanaged
 {
     /// <summary>
+    /// Return the minimal required number of items in the buffer passed on <see cref="Invoke" />.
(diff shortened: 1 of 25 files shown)


JimBobSquarePants:
I agree with you here @ynse01 

When I first implemented the convolution API updates that included the hack to share the buffer that was definite misuse of the existing API. This makes it much cleaner.

JimBobSquarePants (approved):

Issue #2198, opened 2022-08-04
Loading a GIF results in a bigger file


sescandell:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

2.1.3

### Other ImageSharp packages and versions

None, made a simple monofile sample repo in .Net6

### Environment (Operating system, version and so on)

Windows 11, .Net6 (but not limited to this framework)

### .NET Framework version

.Net4.8, netcore3.1, net6

### Description

Hi,

I made a dummy test: load a GIF file, then save it somewhere else without processing it. The final file has a bigger size than the original (more than twice its size, but variable depending on the source. I observed a 5x on another image).

Am I missing an option on save?

Thanks,

### Steps to Reproduce

Here is a fully sample working test (.Net6, dependency : Image.Sharp v.2.1.3)

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="SixLabors.ImageSharp" Version="2.1.3" />
  </ItemGroup>

  <ItemGroup>
    <None Update="sample.gif">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </None>
  </ItemGroup>

</Project>
```

```csharp
// program.cs
using SixLabors.ImageSharp;

Console.WriteLine("Cloning...");

Directory.CreateDirectory("results");
using var src = File.OpenRead("sample.gif");
File.Delete("results/output.gif");
using var fs = File.OpenWrite("results/output.gif");

using var image = await Image.LoadAsync(src);
await image.SaveAsGifAsync(fs);

await fs.FlushAsync();

Console.WriteLine("Done");
Console.WriteLine($"SrcSize: {src.Length / 1024}");
Console.WriteLine($"OutputSize: {fs.Length / 1024}");
Console.ReadKey();
```

Results sample:

> Cloning...
> Done
> SrcSize: 1819
> OutputSize: 3886

### Images

[...]

PR #2244, opened 2022-09-27
Fix and rewrite the Gif LzwEncoder

JimBobSquarePants:
### Prerequisites

### Description

While investigating #2198 (_Answers on a postcard there btw_) I noticed that our Gif LZW encoder was equal parts unreadable and incorrect (_we were maxing out our sub-block lengths a byte short - 254 vs 255 bytes_).

Rather than try to patch the existing encoder I decided to rewrite it to gain a better understanding of the algorithm and make it more readable. I think I've succeeded. 

I've not benchmarked this, but it should be faster too.

### DIFF
--- a/src/ImageSharp/Formats/Gif/LzwEncoder.cs
+++ b/src/ImageSharp/Formats/Gif/LzwEncoder.cs
@@ -190,7 +190,7 @@ internal ref struct HashLut
         if (index != -1)
         {
             ref int v = ref MemoryMarshal.GetReference(this.values);
-            value = Unsafe.Add(ref v, index);
+            value = Unsafe.Add(ref v, (uint)index);
             return true;
         }
 
--- a/src/ImageSharp/Formats/Gif/LzwEncoder.cs
+++ b/src/ImageSharp/Formats/Gif/LzwEncoder.cs
@@ -11,411 +11,219 @@ namespace SixLabors.ImageSharp.Formats.Gif;
 /// <summary>
 /// Encodes and compresses the image data using dynamic Lempel-Ziv compression.
 /// </summary>
-/// <remarks>
-/// Adapted from Jef Poskanzer's Java port by way of J. M. G. Elliott. K Weiner 12/00
-/// <para>
-/// GIFCOMPR.C       - GIF Image compression routines
-/// </para>
-/// <para>
-/// Lempel-Ziv compression based on 'compress'.  GIF modifications by
-/// David Rowley (mgardi@watdcsu.waterloo.edu)
-/// </para>
-/// GIF Image compression - modified 'compress'
-/// <para>
-/// Based on: compress.c - File compression ala IEEE Computer, June 1984.
-/// By Authors:  Spencer W. Thomas      (decvax!harpo!utah-cs!utah-gr!thomas)
-///              Jim McKie              (decvax!mcvax!jim)
-///              Steve Davies           (decvax!vax135!petsd!peora!srd)
-///              Ken Turkowski          (decvax!decwrl!turtlevax!ken)
-///              James A. Woods         (decvax!ihnp4!ames!jaw)
-///              Joe Orost              (decvax!vax135!petsd!joe)
-/// </para>
-/// </remarks>
-internal sealed class LzwEncoder : IDisposable
+internal ref struct LzwEncoder
 {
-    /// <summary>
-    /// 80% occupancy
-    /// </summary>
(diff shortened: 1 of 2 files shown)


antonfirsov:
Though not familiar with the algorithm, nor the old implementation, I hope I will have a chance to take a look on a train ride later. It might be worth to create a before/after benchmark with a *large* image.

src/ImageSharp/Formats/Gif/LzwEncoder.cs
```diff
@@ -11,411 +11,219 @@ namespace SixLabors.ImageSharp.Formats.Gif;
...
     public LzwEncoder(MemoryAllocator memoryAllocator, int colorDepth)
     {
-        this.initialCodeSize = Math.Max(2, colorDepth);
-        this.hashTable = memoryAllocator.Allocate<int>(HashSize, AllocationOptions.Clean);
-        this.codeTable = memoryAllocator.Allocate<int>(HashSize, AllocationOptions.Clean);
+        this.memoryAllocator = memoryAllocator;
+        this.minCodeSize = Math.Max(2, colorDepth);
+        this.clearCode = 1 << this.minCodeSize;
+        this.codeMask = this.clearCode - 1;
+        this.eoiCode = this.clearCode + 1;
```
gfoidl, line 37: Instead of having this a fields, that are initialized, they could be properties too.
* shrinks the size of this (huge) struct by 8 bytes
* the values are very use to compute and JIT will inline these for sure

src/ImageSharp/Formats/Gif/LzwEncoder.cs
```diff
@@ -11,411 +11,219 @@ namespace SixLabors.ImageSharp.Formats.Gif;
...
+        this.index = 0;
+    }
 
-        while (this.currentBits >= 8)
+    public int Capacity { get; }
+
+    public bool TryGetValue(int key, out int value)
+    {
+        int index = this.keys.IndexOf(key);
+        if (index != -1)
```
gfoidl, line 190:
```suggestion
        if (index >= 0)
```
Comparison to 0 is a tiny little bit better at cpu-level, as the "compare and jump" can be fused by the cpu to only one instruction executed.

src/ImageSharp/Formats/Gif/LzwEncoder.cs
```diff
@@ -11,411 +11,219 @@ namespace SixLabors.ImageSharp.Formats.Gif;
...
         {
-            this.AddCharacter((byte)(this.currentAccumulator & 0xFF), ref accumulatorsRef, outs);
-            this.currentAccumulator >>= 8;
-            this.currentBits -= 8;
+            ref int v = ref MemoryMarshal.GetReference(this.values);
+            value = Unsafe.Add(ref v, index);
```
gfoidl, line 193:
```suggestion
            value = Unsafe.Add(ref v, (uint)index);
```
To avoid the sign-extending move.

src/ImageSharp/Formats/Gif/LzwEncoder.cs
JimBobSquarePants: Nice. Didn't know that!

src/ImageSharp/Formats/Gif/LzwEncoder.cs
JimBobSquarePants: I was wondering. Tried a cast to `nint` but didn't consider `uint`

src/ImageSharp/Formats/Gif/LzwEncoder.cs
JimBobSquarePants: I wasn't actually aware that this made a difference. Do you know of any docs/blogs that chat about it?

I'll move most of them to the `Compress` method body and convert the allocator and min code size to properties.

I absolutely love it when you jump in and review like this. It's always a great learning experience!

JimBobSquarePants:
I'm gonna close this PR. My LUT strategy is idiotic (30x slower) and a simple dictionary is 2x slower than the original.

I don't want to waste more time attempting to optimize it.

src/ImageSharp/Formats/Gif/LzwEncoder.cs
gfoidl: Thanks ❤️ 

Unfortunately I don't have docs/blogs-links for you. I collected these infos by measuring, reading GH-issues, etc. So very spread around but some things I remember (sometimes 😉).

PR #2249, opened 2022-09-30
Fix position handling in BufferedReadStream

JimBobSquarePants:
### Prerequisites

### Description

Changes the behavior to prefill when the position is set out of the buffer range.

### DIFF
--- a/tests/ImageSharp.Tests/IO/BufferedReadStreamTests.cs
+++ b/tests/ImageSharp.Tests/IO/BufferedReadStreamTests.cs
@@ -334,9 +334,8 @@ public class BufferedReadStreamTests
         bufferedStream.Position = 10;
         bufferedStream.Position = 3;
 
-        // readValue is 25, but should be 3
-        int readValue = bufferedStream.ReadByte();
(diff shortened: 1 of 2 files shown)


tests/ImageSharp.Tests/IO/BufferedReadStreamTests.cs
```diff
@@ -235,34 +215,28 @@ public void BufferedStreamCanReadSubsequentMultipleByteSpanCorrectly(int bufferS
...
+        const int plusOne = 1;
+        int skip2 = bufferSize;
 
-                // Skip
-                reader.Skip(skip);
-                Assert.Equal(skip, reader.Position);
-                Assert.Equal(stream.Position, reader.Position);
+        // Skip
+        reader.Skip(skip);
+        Assert.Equal(skip, reader.Position);
```
JimBobSquarePants, line 227: Tests were bad here. It was wrong to expect the underlying position to match since we often go out of sync by filling the buffer.

brianpopow (approved):
Looks good, thanks for the fix

PR #2267, opened 2022-10-12
Ensure seekable streams can be read multiple times.

### LINKED ISSUE #2259: Loading image from a stream doesn't start at the current position
### Prerequisites

### ImageSharp version

2.1.3

n/a

Windows 11

### .NET Framework version

6.0.0

### Description

I have a file that consists of two concatenated images (ie one PNG file, and then a second PNG file).

When I load the first image, the position of the stream increases, but the second read reads from the beginning again. So, now matter how times I load an image, it's always the first image in the stream that gets loaded.

Here's my sample code:

```
using SixLabors.ImageSharp;


//var config = new Configuration(new PngConfigurationModule());
//config.ReadOrigin = ReadOrigin.Current; // this makes no difference

//first of all, read the stream sequentially

using var stream = new FileStream("red_and_blue.png", FileMode.Open, FileAccess.Read);

using var i = Image.Load(stream);

Console.WriteLine($"Stream position: {stream.Position}");

using var j = Image.Load(stream);
Console.WriteLine($"Stream position: {stream.Position}"); // THIS HASN'T MOVED

using var i2 = i.CloneAs<Rgb24>();
using var j2 = j.CloneAs<Rgb24>();

Console.WriteLine($"First colour: {i2[5, 5]}");

Console.WriteLine($"Second colour: {j2[5, 5]}"); // THIS SHOULD BE DIFFERENT

//instead, copy the rest of the stream

using var reader = new BinaryReader(stream);
var bytes = reader.ReadBytes((int)(stream.Length - stream.Position));
using var k = Image.Load(bytes);

using var k2 = k.CloneAs<Rgb24>();
Console.WriteLine($"Second colour (2nd attempt): {k2[5, 5]}"); 
```

### Steps to Reproduce

See above

### Images

The final image is made by running:

`copy /b red.png+blue.png red_and_blue.png`

blue
red
red_and_blue

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2259 

I wasn't able to replicate the issue with main in synchronous loading (likely due to the decoder refactor) but I was able to do so for our asynchronous load.

I addition I've included a more efficient fix for buffered stream position setting.

### DIFF
--- a/src/ImageSharp/IO/BufferedReadStream.cs
+++ b/src/ImageSharp/IO/BufferedReadStream.cs
@@ -172,7 +172,7 @@ internal sealed class BufferedReadStream : Stream
 
         // Too big for remaining buffer but less than entire buffer length
         // Copy to buffer then read from there.
-        if (this.readBufferIndex < 0 || (count + this.readBufferIndex > this.BufferSize))
+        if (this.readBufferIndex < 0 || (this.readBufferIndex > this.BufferSize - count))
         {
             return this.ReadToBufferViaCopySlow(buffer);
         }
--- a/src/ImageSharp/IO/BufferedReadStream.cs
+++ b/src/ImageSharp/IO/BufferedReadStream.cs
@@ -172,7 +172,7 @@ internal sealed class BufferedReadStream : Stream
 
         // Too big for remaining buffer but less than entire buffer length
         // Copy to buffer then read from there.
-        if (this.readBufferIndex < 0 || (this.readBufferIndex > this.BufferSize - count))
+        if ((uint)this.readBufferIndex > (uint)(this.BufferSize - count))
         {
             return this.ReadToBufferViaCopySlow(buffer);
         }
--- a/src/ImageSharp/IO/BufferedReadStream.cs
+++ b/src/ImageSharp/IO/BufferedReadStream.cs
@@ -60,7 +60,7 @@ internal sealed class BufferedReadStream : Stream
         }
 
         // This triggers a full read on first attempt.
-        this.readBufferIndex = this.BufferSize;
+        this.readBufferIndex = int.MinValue;
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


src/ImageSharp/IO/BufferedReadStream.cs
```diff
@@ -140,7 +141,7 @@ public override int ReadByte()
 
         // Our buffer has been read.
         // We need to refill and start again.
-        if (this.readBufferIndex > this.maxBufferIndex)
+        if (this.readBufferIndex < 0 || this.readBufferIndex > this.maxBufferIndex)
```
gfoidl, line 144:
```suggestion
        if ((uint)this.readBufferIndex > (uint)this.maxBufferIndex)
```
Saves one comparison, as with the uint-cast negative value will be re-interpreted as values > int.MaxValue.

I know this is for streams, so the cost of reading from the stream will be dominant, but I can't stop squeezing out perf 😉

src/ImageSharp/IO/BufferedReadStream.cs
```diff
@@ -186,7 +172,7 @@ public override int Read(Span<byte> buffer)
 
         // Too big for remaining buffer but less than entire buffer length
         // Copy to buffer then read from there.
-        if (count + this.readBufferIndex > this.BufferSize)
+        if (this.readBufferIndex < 0 || (count + this.readBufferIndex > this.BufferSize))
```
gfoidl, line 175: The `+` could overflow (at least in theory).
```suggestion
        if (this.readBufferIndex < 0 || (this.readBufferIndex > this.BufferSize - count))
```
Or to remove one check
```suggestion
        if ((uint)this.readBufferIndex > (uint)(this.BufferSize - count))
```

JimBobSquarePants: Should have committed the second one. Will do so later. Thanks!

Issue #2263, opened 2022-10-11
Can't open some TIFs files (System.NotSupportedException: The specified TIFF photometric interpretation is not supported: Separated)


aistvan:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

2.1.3

### Other ImageSharp packages and versions

No

### Environment (Operating system, version and so on)

Windows

### .NET Framework version

net6

### Description

I tried this lib at first time, with TIFF's and now, I have no clue, why I get tis exception.

`System.Private.CoreLib: Exception while executing function: ResizeImage. SixLabors.ImageSharp: The specified TIFF photometric interpretation is not supported: Separated.`

Does anyone have any solution for this?

### Steps to Reproduce

```
using (Image<Rgba32> input = Image.Load<Rgba32>(image, out format))
{
   ...
}
```

### Images

I only have copyrighted images unfortunately.

brianpopow:
Closing with #2268 merged

aistvan:
> Are you serious? Have you tried reading the exception? What is unclear in "The specified TIFF photometric interpretation is not supported"?

Yes, I'm serious. I think an image handling lib's one of the most important goal must be to support the image hanling itself.
And yes, I read the exception, but for me, who is not an "imaging" developer, it doesn't say anything.

> It's one of the million different variant's of tiff, someone needs to implement support for this.

Nice. I tried to keep myself away from tiff images, but we have marketing and they used it a lot.
I must support them. If ImageSharp doesn't work for me, I have to look for another lib. I don't want to write a new one.

JimBobSquarePants:
Do you pay for this? Does marketing?

aistvan:
> Do you pay for this? Does marketing?

No. Wife.

PR #2268, opened 2022-10-13
Add support for decoding tiff images with cmyk pixel data

brianpopow:
### Prerequisites

### Description

This PR adds support for decoding tiff images with cmyk pixel data.

Related issue: #2263


src/ImageSharp/Formats/Tiff/PhotometricInterpretation/CieLabTiffColor{TPixel}.cs
```diff
@@ -31,8 +31,8 @@ public override void Decode(ReadOnlySpan<byte> data, Buffer2D<TPixel> pixels, in
             for (int x = 0; x < pixelRow.Length; x++)
             {
                 float l = (data[offset] & 0xFF) * 100f * Inv255;
-                var lab = new CieLab(l, (sbyte)data[offset + 1], (sbyte)data[offset + 2]);
-                var rgb = ColorSpaceConverter.ToRgb(lab);
+                CieLab lab = new(l, (sbyte)data[offset + 1], (sbyte)data[offset + 2]);
```
JimBobSquarePants, line 34: I'm curious. Why are we casting to `sbyte` here? CieLab takes `float` for the constructor parameters. 

The `L` component ranges from `0-100`
Generally, the `A` and `B` components are unbounded though it's common to clamp them to `−128` to `127`

_Note. I just realized the property notes on the type are incorrect, saying `-100` to `100` for the `A` and `B` components._

https://en.wikipedia.org/wiki/CIELAB_color_space

JimBobSquarePants, line 34: I'm curious to how the data is encoded here. 

The CieLab  `A`, `B` values usually range from −128 to 127 when working with integer values.  
Casting to `sbyte` and using overflow is a neat trick but you have to be careful.

JimBobSquarePants (approved):
LGTM 👍

src/ImageSharp/Formats/Tiff/PhotometricInterpretation/CieLabTiffColor{TPixel}.cs
brianpopow: To cite the spec:

```
Tiff is defined such that each quantity be encoded with 8 bits. This provides 256 levels of L* lightness;
256 levels (+/- 127) of a*, and 256 levels (+/- 127) of b*.  
...
Limiting the theoretically unbounded a* and b*
ranges to +/- 127 allows encoding in 8 bits without eliminating any but the most
saturated self-luminous colors. It is anticipated that the rare specialized applications
requiring support of these extreme cases would be unlikely to use CIELAB
anyway. All object colors, in fact all colors within the theoretical MacAdam limits,
fall within the +/- 127 a*/b* range.
```

PR #2269, opened 2022-10-16
Normalize and cleanup encoders

JimBobSquarePants:
### Prerequisites

### Description

There's no point trying to copy the options pattern that we introduced in #2180 as the requirements are different.

- We DON'T want to allow passing specific DECODERS to the General API.
- We DO want to allow passing specific ENCODERS to the General API.

Instead, I've refactored the encoders to do the following

- Introduced 2 new abstract types `ImageEncoder` and `QuantizingImageEncoder` with shared, preconfigured options properties. Two new shared options properties `SkipMetadata` and `PixelSamplingStrategy` have been introduced into the respective types.
- Inlined the individual encoder options interfaces into the concrete implementations and removed any duplication (_no more separate mutable options types_)
- Made the individual encoder options immutable.
- Refactored each encoder to allow skipping metadata on encode - This allows us to encode images without metadata while preserving the metadata in the decoded image.
- Refactored each `QuantizingImageEncoder` instance to use the sampling strategy from the options. This applies to local and global palette generation.
- General cleanup in files I was altering.


src/ImageSharp/Formats/Jpeg/JpegEncoder.cs
```diff
@@ -1,25 +1,28 @@
...
 {
     /// <summary>
     /// Backing field for <see cref="Quality"/>.
     /// </summary>
     private int? quality;
 
-    /// <inheritdoc/>
+    /// <summary>
+    /// Gets the quality, that will be used to encode the image. Quality
+    /// index must be between 0 and 100 (compression from max to min).
```
br3aker, line 18: Should be between 1 and 100 as code itself throws on 0 :)

Maybe it's also wise to 'promote' [1..99] range in remarks as jpeg spec doesn't advise quality 100 it doesn't really preserve 'quality' and can lead to artifacts. Most likely I'll just bump this up in some of the PRs when I'll have time to actually commit.

br3aker (commented):
Don't have time for a proper look but it definitely looks interesting.

src/ImageSharp/Formats/Jpeg/JpegEncoder.cs
JimBobSquarePants: I didn't know that about quality 100! Weird that it does that. I wonder if we should simply clamp to 1-99?

br3aker: I don't think so, quality=100 is still technically correct according to the spec. I believe regular users treat quality as percent value, would be weird to be clamped at 99% from user perspective :)

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -1192,4 +1219,196 @@ private void SwapScanlineBuffers()
...
+            }
+            else if (bits is >= 5 and <= 7)
+            {
+                bits = 8;
+            }
+
+            bitDepth = bits;
+        }
+
+        if (Array.IndexOf(PngConstants.ColorTypes[colorType], bitDepth) == -1)
```
gfoidl, line 1339:
```suggestion
        if (Array.IndexOf(PngConstants.ColorTypes[colorType], bitDepth) < 0)
```
It's tiny little bit faster than `== -1` as the cpu can fuse the compare and jump into one instruction actually executed (the `test + jmp`-combo).

src/ImageSharp/Processing/Processors/Quantization/DefaultPixelSamplingStrategy.cs
```diff
@@ -101,4 +101,54 @@ Buffer2DRegion<TPixel> GetRow(int pos)
...
+            r = Math.Max(this.MinimumScanRatio, r); // always visit the minimum defined portion of the image.
+
+            Rational ratio = new(r);
+
+            int denom = (int)ratio.Denominator;
+            int num = (int)ratio.Numerator;
+
+            for (int pos = 0; pos < totalNumberOfRows; pos++)
+            {
+                int subPos = pos % denom;
```
gfoidl, line 140:
```suggestion
                Debug.Assert(denom > 0);
                int subPos = (int)((uint)pos % (uint)denom);
```
Emits a "better" divison as the sign doesn't need to be considered.
Same below.

Although in contrast of an IEnumerable that effect won't be big.
If this method is hot, maybe a custom struct-based enumerator has more impact on perf and allocations.

gfoidl (commented):
Just eye-rolled product code changes.

antonfirsov:
Will take a look at this within a day or two!

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -97,7 +103,8 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream, CancellationToken
             }
             else
             {
-                quantized = frameQuantizer.BuildPaletteAndQuantizeFrame(image.Frames.RootFrame, image.Bounds());
+                frameQuantizer.BuildPalette(this.pixelSamplingStrategy, image.Frames.RootFrame);
+                quantized = frameQuantizer.QuantizeFrame(image.Frames.RootFrame, image.Bounds());
```
antonfirsov, line 107: Do I understand it correctly that this introduces an optimization for single-frame gifs? Can't remember why didn't I add it originally single-frame, but would be nice to see a performance comparison. I assume there is no impact on the output quality?

src/ImageSharp/Processing/Processors/Quantization/DefaultPixelSamplingStrategy.cs
```diff
@@ -101,4 +103,55 @@ Buffer2DRegion<TPixel> GetRow(int pos)
             }
         }
     }
+
+    /// <inheritdoc />
+    public IEnumerable<Buffer2DRegion<TPixel>> EnumeratePixelRegions<TPixel>(ImageFrame<TPixel> frame)
```
antonfirsov, line 108: The multi-frame overload has some unit tests, might be worth to test this one too.

antonfirsov (commented):
My review assumes that we are OK with the general design direction, pointing out a few minor issues, while the PR looks mostly good.

However:
While I do understand the differences in the requirements, I'm still not happy with the decoder-encoder API asymmetry. It feels weird, and I still think we can do better. We should re-evaluate the tradeoffs of different decoder/encoder design strategies.

That should be another iteration/discussion though, since this PR contains way too many changes better to be merged sooner than later.

src/ImageSharp/Formats/IEncoderOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Gets the quantizer used to generate the color palette.
+    /// </summary>
+    IQuantizer Quantizer { get; init; }
+
+    /// <summary>
+    /// Gets the <see cref="IPixelSamplingStrategy"/> used for quantization when building color palettes.
+    /// </summary>
+    IPixelSamplingStrategy PixelSamplingStrategy { get; init; }
+}
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah, happy to use the base classes instead.

src/ImageSharp/Formats/ImageEncoder.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <inheritdoc/>
+    public abstract Task EncodeAsync<TPixel>(Image<TPixel> image, Stream stream, CancellationToken cancellationToken)
+        where TPixel : unmanaged, IPixel<TPixel>;
+}
+
+/// <summary>
+/// The base class for all image encoders that allow color palette generation via quantization.
+/// </summary>
+public abstract class QuantizingImageEncoder : ImageEncoder, IQuantizingEncoderOptions
```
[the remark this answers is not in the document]
JimBobSquarePants: I actually want to subclass in the decoders. Just trying to figure out a nice way.

src/ImageSharp/Image{TPixel}.cs
```diff
@@ -181,7 +173,7 @@ public new ImageFrameCollection<TPixel> Frames
     /// <exception cref="ArgumentOutOfRangeException">Thrown when the provided (x,y) coordinates are outside the image boundary.</exception>
     public TPixel this[int x, int y]
     {
-        [MethodImpl(InliningOptions.ShortMethod)]
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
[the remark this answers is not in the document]
JimBobSquarePants: I think we should be using the custom methods more sparingly on areas we're actively profiling. (I recall you saying similar a few years back now ;))

src/ImageSharp/Formats/Bmp/BmpEncoder.cs
```diff
@@ -2,48 +2,38 @@
...
-    public IQuantizer Quantizer { get; set; }
+    public bool SupportTransparency { get; init; }
 
     /// <inheritdoc/>
-    public void Encode<TPixel>(Image<TPixel> image, Stream stream)
-        where TPixel : unmanaged, IPixel<TPixel>
+    public override void Encode<TPixel>(Image<TPixel> image, Stream stream)
     {
-        var encoder = new BmpEncoderCore(this, image.GetMemoryAllocator());
+        BmpEncoderCore encoder = new(this, image.GetMemoryAllocator());
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah, but every time I open a file just now there's a heap of warnings from the new rules. I'm adopting the boy scout rule.

src/ImageSharp/Processing/Processors/Quantization/DefaultPixelSamplingStrategy.cs
```diff
@@ -79,14 +80,15 @@ public IEnumerable<Buffer2DRegion<TPixel>> EnumeratePixelRegions<TPixel>(Image<T
...
-            var ratio = new Rational(r);
+            Rational ratio = new(r);
 
             int denom = (int)ratio.Denominator;
             int num = (int)ratio.Numerator;
 
             for (int pos = 0; pos < totalNumberOfRows; pos++)
             {
-                int subPos = pos % denom;
+                Debug.Assert(denom > 0, "Denominator must be greater than zero.");
```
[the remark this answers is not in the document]
JimBobSquarePants: Sorry, I shouldn't have added this. `DebugGuard` is better, and it definitely should not be in a loop.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
JimBobSquarePants: Yeah. It is but it only affects 4K or greater images by default. I didn't introduce a change to the default dimension limit as I didn't want to affect quality on smaller images.

src/ImageSharp/Processing/Processors/Quantization/DefaultPixelSamplingStrategy.cs
JimBobSquarePants: Will do!

src/ImageSharp/Formats/IEncoderOptions.cs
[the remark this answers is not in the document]
JimBobSquarePants: Done

src/ImageSharp/Image{TPixel}.cs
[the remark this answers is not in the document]
antonfirsov: Makes sense for these 2 methods actually.

src/ImageSharp/Processing/Processors/Quantization/DefaultPixelSamplingStrategy.cs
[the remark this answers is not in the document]
antonfirsov: You changed only the other overload :)

```suggestion
                DebugGuard.MustBeGreaterThan(denom, 0, "Denominator must be greater than zero.");
```

antonfirsov (approved):

PR #2275, opened 2022-10-25
add some nullable ref types for image detection

SimonCropp:
My plan it to tackle this as small easily reviewable PRs

given the Big Bang https://github.com/SixLabors/ImageSharp/pull/2236 didnt seem to work

### DIFF
--- a/src/ImageSharp/Formats/Bmp/BmpImageFormatDetector.cs
+++ b/src/ImageSharp/Formats/Bmp/BmpImageFormatDetector.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
 
+#nullable enable
 using System.Buffers.Binary;
 
 namespace SixLabors.ImageSharp.Formats.Bmp;
@@ -14,7 +15,7 @@ public sealed class BmpImageFormatDetector : IImageFormatDetector
     public int HeaderSize => 2;
 
     /// <inheritdoc/>
-    public IImageFormat DetectFormat(ReadOnlySpan<byte> header)
+    public IImageFormat? DetectFormat(ReadOnlySpan<byte> header)
     {
         return this.IsSupportedFileFormat(header) ? BmpFormat.Instance : null;
     }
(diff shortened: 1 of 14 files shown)


JimBobSquarePants:
>My plan it to tackle this as small easily reviewable PRs

Definitely the best approach.

src/ImageSharp/Formats/Bmp/BmpImageFormatDetector.cs
```diff
@@ -14,7 +15,7 @@ public sealed class BmpImageFormatDetector : IImageFormatDetector
     public int HeaderSize => 2;
 
     /// <inheritdoc/>
-    public IImageFormat DetectFormat(ReadOnlySpan<byte> header)
+    public IImageFormat? DetectFormat(ReadOnlySpan<byte> header)
```
JimBobSquarePants, line 18: I cannot decide what the best pattern for nullable return types should be. I feel like anything that returns a nullable type should probably use a `TryXX(in, out..)` pattern (though I still haven't found a nice async pattern for that)

SimonCropp: > I feel like anything that returns a nullable type should probably use a TryXX(in, out..)

yeah. i was trying to do this in the least intrusive way

> though I still haven't found a nice async pattern for that

neither have I

SimonCropp: happy to add some TryXX overloads. but i would prefer to do those in a new PR

antonfirsov:
`DetectFormat` and `FindFormatByFileExtension` returning nullable is definitely OK.

Assuming we are 100% committed to finish #2231 for 3.0, we need to decide about the global strategy we want to take to refactor the entire API. The PR doesn't fit the approach recommended by @tocsoft's https://github.com/SixLabors/ImageSharp/issues/2231#issuecomment-1254967136 (changes a small-sub area without without step 1).

SimonCropp:
note that the tryxxx conversation should not be happening on this pr

JimBobSquarePants:
Thanks for this @SimonCropp and apologies for the slow response. 

I'm gonna close this in favour of the more general #2282 which allows us to complete step 1 of the global approach.

SimonCropp:
@JimBobSquarePants no worries at all

PR #2276, opened 2022-10-28
Improve Decoder/Encoder symmetry

JimBobSquarePants:
### Prerequisites

### Description

An attempt to make the decoder/encoder patterns more symmetric and also simplify implementation.

- Adds abstract `ImageDecoder` & `ImageEncoder` types which allow simplified synchronous implementations while ensuring streams are seekable.
- Renames `IImageDecoderSpecialized<T>` to `SpecializedImageDecoder<T>` and adds a `SpecializedImageDecoder<T>` implementation that inherits `ImageDecoder`.
- Adds `QuantizingEncoder` abstraction containing common properties and ensure that they are used across all implementing encoders.
- Simplifies decoder/encoder implementations by removing casting and introducing default method proxying by specialized decoders/encoders
- Expose default decoding resize method (_not sure about this one as we might end up implementing scanline block resize on decode_).
- Make `DecoderOptions` properties `init` only.

There are now two separate pipelines - Our implemented codecs use a synchronous pipeline where the asynchronous implementation is proxied. External implementations are now able to fully implement both sync and async pipelines or inherit our sync pipeline.


src/ImageSharp/Formats/SpecializedImageDecoder{T}.cs
```diff
@@ -0,0 +1,91 @@
...
+/// Acts as a base class for specialized image decoders.
+/// Specialized decoders allow for additional options to be passed to the decoder.
+/// Types that inherit this decoder are required to implement cancellable synchronous decoding operations only.
+/// </summary>
+/// <typeparam name="T">The type of specialized options.</typeparam>
+public abstract class SpecializedImageDecoder<T> : ImageDecoder, ISpecializedImageDecoder<T>
+    where T : ISpecializedDecoderOptions
+{
+    /// <inheritdoc/>
+    public Image<TPixel> Decode<TPixel>(T options, Stream stream)
```
JimBobSquarePants, line 18: All `ISpecializedImageDecoder<T>` methods are now available for usage and ensure seekable streams.

src/ImageSharp/Formats/QuantizingImageEncoder.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using SixLabors.ImageSharp.Processing;
+using SixLabors.ImageSharp.Processing.Processors.Quantization;
+
+namespace SixLabors.ImageSharp.Formats;
+
+/// <summary>
+/// Acts as a base class for all image encoders that allow color palette generation via quantization.
+/// </summary>
+public abstract class QuantizingImageEncoder : ImageEncoder
```
JimBobSquarePants, line 12: No need for an interface. This exists for normalization and ease of implementation.

src/ImageSharp/Image.FromStream.cs
```diff
@@ -574,34 +583,36 @@ public static Image Load(DecoderOptions options, Stream stream, out IImageFormat
...
-            T result = action(ms, ct);
-            if (stream.CanSeek && ms.Position != ms.Length)
+            if (configuration.ReadOrigin == ReadOrigin.Begin)
             {
-                stream.Position = ms.Position;
+                stream.Position = 0;
             }
 
-            return result;
+            return await Action(stream, stream.Position, cancellationToken).ConfigureAwait(false);
```
JimBobSquarePants, line 608: We now allow external implementations to be asynchronous and only copy when necessary.

antonfirsov (commented):
This looks definitely better than the initial version!

I've always seen encoders and decoders as types which are supposed to be implemented and only used through the our `Image.Load` / `Image.Save` infrastructure, and never directly. I think we should keep this design, since it greatly simplifies things both for us and the users.

If the main reason for changing this is the "specialized" decoding niche case, I think it's definitely not worth the extra layers being added. Instead, we can add an infra for the specialized decoding cases separated from the decoder types.

For example we can create API-s as following:

```C#
public class Image
{
    public static Image Load(JpegDecoderOptions options, Stream stream);
}
```

The only disadvantage of the above design is that 3rd party decoders won't be able to utilize the specialized infra, but it's such a niche (or better even non-existent) case, that I'm fine with it.

Or we can consider subclassing `JpegDecoderOptions : DecoderOptions`. This would greatly simplify the object model. The disadvantage is that a user can pass `JpegDecoderOptions` together with a PNG stream, but IMHO it doesn't sound that terrible. We can throw an exception or just ignore it.

A third solution would be to make decoders stateful, merging the decoder role with the "specialized options" role. We could then compose `IImageDecoder` from the general `DecoderOptions`. This would also remove the stateless decoder / stateful encoder asymmetry.

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,241 @@
...
+        T Action(Stream s, long position)
+        {
+            T result = action(s);
+
+            // Our buffered reads may have left the stream in an incorrect non-zero position.
+            // Reset the position of the seekable stream if we did not read to the end to allow additional reads.
+            if (stream.CanSeek && stream.Position != s.Position && s.Position != s.Length)
+            {
+                stream.Position = position + s.Position;
+            }
```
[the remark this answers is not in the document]
JimBobSquarePants: It's a fix for #2259 where 3 images were embedded in a stream one after the other. We have tests to cover it. I can add a reference to the issue in the code.

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,241 @@
...
+/// <summary>
+/// Acts as a base class for image decoders.
+/// Types that inherit this decoder are required to implement cancellable synchronous decoding operations only.
+/// </summary>
+public abstract class ImageDecoder : IImageDecoder
+{
+    /// <inheritdoc/>
+    public Image<TPixel> Decode<TPixel>(DecoderOptions options, Stream stream)
+        where TPixel : unmanaged, IPixel<TPixel>
+        => WithSeekableStream(
```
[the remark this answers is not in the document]
JimBobSquarePants: It's for when the decoder is called directly. Duplication is avoided because the decoder is always using a seekable stream when called via `Image.Load`

src/ImageSharp/Formats/IImageDecoder.cs
```diff
@@ -6,35 +6,68 @@
...
+    /// <summary>
+    /// Decodes the image from the specified stream to an <see cref="Image{TPixel}"/> of a specific pixel type.
+    /// </summary>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    /// <param name="options">The general decoder options.</param>
+    /// <param name="stream">The <see cref="Stream"/> containing image data.</param>
+    /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+    /// <returns>A <see cref="Task{Image}"/> representing the asynchronous operation.</returns>
+    /// <exception cref="ImageFormatException">Thrown if the encoded image contains errors.</exception>
+    public Task<Image<TPixel>> DecodeAsync<TPixel>(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
JimBobSquarePants: There's one library I know that has both sync and async decoders/encoders. TiffLibrary which has an [ImageSharp adapter](https://github.com/yigolden/TiffLibrary#adapter-for-sixlaborsimagesharp). 

I would be happy to drop async from both decode and encode and just use an extension to wrap the synchronous call though. `IImageDecoder` is a lot of work to implement.

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,241 @@
...
+    /// <exception cref="ImageFormatException">Thrown if the encoded image contains errors.</exception>
+    protected abstract IImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Performs a resize operation against the decoded image. If the target size is not set, or the image size
+    /// already matches the target size, the image is untouched.
+    /// </summary>
+    /// <param name="options">The decoder options.</param>
+    /// <param name="image">The decoded image.</param>
+    protected static void Resize(DecoderOptions options, Image image)
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah, that's a much better name.

JimBobSquarePants:
@antonfirsov I'm really struggling with a stateful decoder implementation that avoids being called directly via our general middleware. The absolute last pattern I ever want to allow again is for someone to do this. 

```c#
using var image = Image.Load("my.png", new PngDecoder());
```
It's a footgun IMO. I've answered quite a few questions on Stack Overflow and elsewhere, where people have attempted to pass explicit, incorrect decoders in an attempt to decode streams then wondered why it doesn't work. Sometimes it's not their fault - I've seen tons of images over the years working in digital agencies where the image has the wrong file extension. 

That's why I have tried to make it as clear as I can that specialized decoder options are an advanced, specialist operation that only individuals with the specific requirement to use those options should use. 

General users should never care about, nor require the specialization.

I see 3 levels of usage.

Simple.

```c#
using var image  = Image.Load(inStream);
```

Intermediate.

```c#
DecoderOptions options = new(){...};
using var image  = Image.Load(options, inStream);
```

Advanced.

```c#
bool isJpeg = // Some method that is guaranteed to the correct result, maybe via pre-stored mime type.
if(isJpeg)
{
    JpegDecoderOptions options = new(){...}
    using var image  = new JpegDecoder().Decoder(options, inStream);
}
```

The stateless nature of the decoder is there because we use the options in our `IImageDecoder` methods since we require the ability for the decoders to be initialized without constructor parameters. 

Maybe we enforce a decoder factory to `ImageFormatManager.SetDecoder(...)` that accepted the options and used them to construct instances?

We could then drop the decoder options from the `IImageDecoder` API?

However, that doesn't mean we would drop passing the general decoder options via the `Image.Load` middleware. Those general options give us a lot of power. 

I dunno, this is hard...

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,241 @@
...
+            }
+
+            return result;
+        }
+
+        // NOTE: We are explicitly not executing the action against the stream here as we do in WithSeekableStream() because that
+        // would incur synchronous IO reads which must be avoided in this asynchronous method. Instead, we will *always* run the
+        // code below to copy the stream to an in-memory buffer before invoking the action.
+
+        // TODO: Enable these optimizations. Several Decode_Cancellation tests fail if enables as no exception is thrown.
```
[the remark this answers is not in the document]
JimBobSquarePants: NM. Figured it out.

Webreaper:
For me intermediate and advanced are the same, but I see why you called them out separately. I like the beginner / intermediate options, and they make sense to me - the options container object is exactly what makes the code readable but gives extra flexibility. Is probably make the stream the first parameter though, and options second, as that seems more intuitive.

I presume for scaled decoding I'd pass the target size in the options object? But yeah, I'd be very happy with this.

JimBobSquarePants:
@Webreaper Thanks for your input. The main difference between Intermediate and Advanced is that you cannot pass `JpegDecoderOptions` to any of the `Image.Load(...)` methods. And yeah, for general scaled decoding you pass a `Size` via the options. The `JpegDecoder` will decode the image to the closest size then perform secondary resize for an exact match. To avoid this second resize you'd use the specialized options against the decoder directly. 

@antonfirsov I've been reviewing the 2.x branch and the build in `IImageDecoder` instances are essentially stateless there since they are registered in the `ImageFormatManager` against the private dictionary.

src/ImageSharp/Formats/IImageDecoder.cs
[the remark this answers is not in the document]
antonfirsov: Wow, that's nice! We should keep the async path then.

src/ImageSharp/Formats/ImageDecoder.cs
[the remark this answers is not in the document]
antonfirsov: For me this is a sign, that something is off with the design. Would be nice to figure out how to do it only once.

src/ImageSharp/Formats/ImageDecoder.cs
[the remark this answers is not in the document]
antonfirsov: I think I'm either missing something, or this code might be buggy.

If `stream.CanSeek == false`, we are creating a `ChunkedMemoryStream` and passing it to `Action`, but in that case the `stream.CanSeek && ... ` condition cannot be ever true, so we won't rewind anything, therefore `Action` is basically a dummy.

antonfirsov:
Note that for the mentioned StackOverflow case this PR wouldn't change much, it looks to me that the user would face the same exceptions with `Image.Load` and by using the decoder directly.

> `new JpegDecoder().Decode(options, inStream);`

What I mean is that introducing an API disparity is not necessary for the advanced case, if we can figure out how to push the specialized options through the `Image.Load` infra, though I admit it is hard. Another thing I'm not big fan of is to solve too many shared, format-agnostic concerns from within a decoder (stream conversion, stream rewinding, default async path). 

On the other hand with the introduction of `DecoderOptions` now there are too many arbitrary details a well-behaving decoder should conform to, including resizing. With this in mind, it looks like the complexity boost in the decoders is by-design so if you think *really* don't want to deal with the disparity thing, I'm not gonna stick to my design preferences forever :) But we need to test out the API-s in tests directly, to make sure they are usable in ways we intend them to be used & avoid obvious mistakes. One thing I noticed is that  currently there are no async overloads defaulting the `CancellationToken`.

src/ImageSharp/Formats/ImageDecoder.cs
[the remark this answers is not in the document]
JimBobSquarePants: It's really only a nice to have for our own decoders. I'm happy to remove it and leave the direct decoder usage pipeline throwing instead.

src/ImageSharp/Formats/ImageDecoder.cs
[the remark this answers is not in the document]
JimBobSquarePants: I'll have another read through; `ChunkedMemoryStream` can seek though. 

The issue only really happens because BufferedReadStream will often overshoot the expected position since it does a bulk read. We only use that stream in our internal decoders so I should move the fix to the extension method that uses it making it much more readable and only appear one time.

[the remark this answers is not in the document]
antonfirsov: Wherever it lands, I would unit test the logic, if possible. No sure `Issue_2259.png` is enough to exercise all the cases.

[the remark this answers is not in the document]
JimBobSquarePants: OK. There was no need for the logic in `Image.FromStream` so I've removed it. `ImageDecoder` is the only place now we perform the check which makes sense since we only use `BufferedReadStream` internally on decoders inheriting that type.

src/ImageSharp/Formats/ImageDecoder.cs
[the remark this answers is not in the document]
JimBobSquarePants: I think I want to keep this. The stream is always seekable if loaded via the `Image.Load` middleware and our checks in the decoder will avoid a duplicate operation and call the action directly.

When someone calls `ImageDecoder.Decode()` or the async equivalent directly our checks will avoid unnecessary copying and also ensure that the input stream is cloned to a seekable stream when required.

src/ImageSharp/Formats/IImageDecoder.cs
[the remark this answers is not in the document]
JimBobSquarePants: As a side note I've now decorated the public cancellation token parameters with a default value.

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,246 @@
...
+        {
+            return Action(stream, stream.Position);
+        }
+
+        Configuration configuration = options.Configuration;
+        using ChunkedMemoryStream memoryStream = new(configuration.MemoryAllocator);
+        stream.CopyTo(memoryStream, configuration.StreamProcessingBufferSize);
+        memoryStream.Position = 0;
+
+        return Action(memoryStream, 0);
```
antonfirsov, line 181: It seems to have the same issue I mentioned in my previous review:
`stream.CanSeek` is always false on this branch, therefore this is equivalent of the following:

```suggestion
        return action(memoryStream, 0);
```

(I don't mean to apply this suggestion, better to inline `Action`.)

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,246 @@
...
+        T Action(Stream s, long position, CancellationToken ct)
+        {
+            T result = action(s, ct);
+
+            // Issue #2259. Our buffered reads may have left the stream in an incorrect non-zero position.
+            // Reset the position of the seekable stream if we did not read to the end to allow additional reads.
+            if (stream.CanSeek && stream.Position != s.Position && s.Position != s.Length)
+            {
+                stream.Position = position + s.Position;
+            }
```
antonfirsov, line 207: Do we test this rewinding for both sync and async?

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,246 @@
...
+            throw new NotSupportedException("Cannot read from the stream.");
+        }
+
+        T Action(Stream s, long position, CancellationToken ct)
+        {
+            T result = action(s, ct);
+
+            // Issue #2259. Our buffered reads may have left the stream in an incorrect non-zero position.
+            // Reset the position of the seekable stream if we did not read to the end to allow additional reads.
+            if (stream.CanSeek && stream.Position != s.Position && s.Position != s.Length)
```
antonfirsov, line 204: `stream.CanSeek` is always true here.

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,246 @@
...
+    {
+        Guard.NotNull(options, nameof(options));
+        Guard.NotNull(stream, nameof(stream));
+
+        if (!stream.CanRead)
+        {
+            throw new NotSupportedException("Cannot read from the stream.");
+        }
+
+        T Action(Stream s, long position, CancellationToken ct)
```
antonfirsov, line 198: Naming: `PeformActionAndResetPosittion`?

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,246 @@
...
+            // Reset the position of the seekable stream if we did not read to the end to allow additional reads.
+            if (stream.CanSeek && stream.Position != s.Position && s.Position != s.Length)
+            {
+                stream.Position = position + s.Position;
+            }
+
+            if (ct.IsCancellationRequested)
+            {
+                throw new TaskCanceledException();
+            }
```
antonfirsov, line 212: I don't see value in this check after we effectively finished the decoding.

antonfirsov (commented):
I'm fine with the API shape now 👍

Since it's a feature now, I would really love to see direct tests for direct decoder usage, and we **really** need proper coverage (eg. method unit tests) for the untrivial logic implemented in the `WithSeekableStream(Async)` utilities.

src/ImageSharp/Formats/ImageEncoder.cs
```diff
@@ -1,43 +1,94 @@
...
+            ms.Position = 0;
+            await ms.CopyToAsync(stream, configuration.StreamProcessingBufferSize, cancellationToken)
+                    .ConfigureAwait(false);
+        }
+
+        Task DoEncodeAsync(Stream innerStream)
+        {
+            try
+            {
+                // TODO: Are synchronous IO writes OK? We avoid reads.
```
antonfirsov, line 80: If it's a memory stream yes, otherwise no, since it will result in IO calls potentially blocking ThreadPool threads. We make sure to use `ChunkedMemoryStream` here as well.

Maybe better to do it in a separate PR though to test it out properly like we did in #1316.

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,246 @@
...
+
+    /// <inheritdoc/>
+    public Image Decode(DecoderOptions options, Stream stream)
+        => WithSeekableStream(
+              options,
+              stream,
+              s => this.Decode(options, s, default));
+
+    /// <inheritdoc/>
+    public Task<Image<TPixel>> DecodeAsync<TPixel>(DecoderOptions options, Stream stream, CancellationToken cancellationToken = default)
```
antonfirsov, line 33: Custom decoders may not need to enforce rewrapping into `ChunkedMemoryStream`, especially TiffLibrary's pure async decoder.

Is our recommendation to implement `IImageDecoder` directly then? Would it be safe?

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,246 @@
...
+
+        if (stream is BufferedReadStream brs && brs.BaseStream is MemoryStream)
+        {
+            return Action(brs, brs.Position, cancellationToken);
+        }
+
+        if (stream is BufferedReadStream brs2 && brs2.BaseStream is ChunkedMemoryStream)
+        {
+            return Action(brs2, brs2.Position, cancellationToken);
+        }
```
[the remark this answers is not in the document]
JimBobSquarePants: You're right! `WebPDecoder` was being called by `WebpTiffCompression` which would have caused this, but I've since refactored it to use `WebPDecoderCore` so can remove.

src/ImageSharp/Formats/ImageDecoder.cs
antonfirsov: But then there is no way to `ScaleToTargetSize`.

We can make these methods virtual, but then there is no need to override the abstract ones. We can turn the abstract methods to virtual ones throwing `NotImplementedException`, but it would become confusing.

Any better ideas?

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,246 @@
...
+            return result;
+        }
+
+        // NOTE: We are explicitly not executing the action against the stream here as we do in WithSeekableStream() because that
+        // would incur synchronous IO reads which must be avoided in this asynchronous method. Instead, we will *always* run the
+        // code below to copy the stream to an in-memory buffer before invoking the action.
+        if (stream is MemoryStream ms)
+        {
+            return Action(ms, ms.Position, cancellationToken);
+        }
```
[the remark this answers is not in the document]
JimBobSquarePants: Because the version in `Image.FromStream` does not do the forced copy since it is designed to work with pure async implementations.

src/ImageSharp/Formats/ImageDecoder.cs
JimBobSquarePants: Not sure I understand. We want to call Action and the code is called from multiple branches.

src/ImageSharp/Formats/ImageDecoder.cs
JimBobSquarePants: Yep. The test image we have contains 3 encoded pngs. That requires 2 rewinds to read all of them. We test this in syn and async scenarios.

src/ImageSharp/Formats/ImageDecoder.cs
JimBobSquarePants: I'll see if our tests pass if I move it before. Basically, none of our decoders are currently checking for this and I didn't want to have to inject it into each one.

src/ImageSharp/Formats/ImageEncoder.cs
JimBobSquarePants: Will deal with it separately since I just moved the existing code to a new location.

src/ImageSharp/Formats/ImageDecoder.cs
JimBobSquarePants: Yeah, they would implement `IImageDecoder` directly since we don't expose any async virtuals in `ImageDecoder`

JimBobSquarePants: Re ScaleToTargetSize. My concern with moving it to a shared utility was that someone might end up implementing a per-scanline scaling operation.

It's tricky.

src/ImageSharp/Formats/ImageDecoder.cs
antonfirsov: After the code doesn't return here, `stream.CanSeek` will be always `false`:

https://github.com/SixLabors/ImageSharp/blob/8fdd6b0e6324e9ff405c4fc822d81ffe6cef051c/src/ImageSharp/Formats/ImageDecoder.cs#L171-L174

In that case `if (stream.CanSeek & ....` in `Action` doesn't make sense.

src/ImageSharp/Formats/ImageDecoder.cs
antonfirsov: > per-scanline scaling operation.

If we have the time, we definitely need to do that for 3.0 at least for JPEG+PNG. The implementation codein  `***DecoderCore` classes will be practically feeding scanlines to [`ResizeWorker`](https://github.com/SixLabors/ImageSharp/blob/7b403897dfee6433d7c39feb6fdef1f4dfe0949b/src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeWorker.cs) or whatever it gets refactored to. Not sure if that infrastructure will be mature enough to be exposed as a public API, that's a secondary priority IMO.

src/ImageSharp/Formats/ImageDecoder.cs
antonfirsov: Jpeg and tiff do:
https://github.com/SixLabors/ImageSharp/search?q=ThrowIfCancellationRequested

It's good to check for cancellation before calling the decoders, I see no point to do it afterwards.

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            // Issue #2259. Our buffered reads may have left the stream in an incorrect non-zero position.
+            // Reset the position of the seekable stream if we did not read to the end to allow additional reads.
+            if (stream.Position != s.Position && s.Position != s.Length)
+            {
+                stream.Position = position + s.Position;
+            }
+
+            // TODO: This is a hack. Our decoders do not check for cancellation requests.
+            // We need to fix this properly by implemented each decoder or allowing passing a token to BufferedReadStream
```
antonfirsov, line 209: You can't pass `CancellationToken` to the stream, because it's the sync path. We do cancellation checks in the loops processing scanlines/blocks etc.

If the check is here because of some tests, we should fix the tests instead.

src/ImageSharp/Formats/ImageDecoder.cs
JimBobSquarePants: OK. Let me double check the tests. Maybe we're pushing to a decoder that doesn't.

src/ImageSharp/Formats/ImageDecoder.cs
antonfirsov: [6978235](https://github.com/SixLabors/ImageSharp/pull/2276/commits/69782359ba4718d232c4cd62bc592e5065a7354e) removed the `stream.CanSeek` check in the sync path, effectively activating the rewinding logic, without any impact to our test result outcome. To me this is a sign that our functional coverage is incomplete.

JimBobSquarePants: Ah yes of course! We cannot reset anyway since the entire input stream is copied. Will update.

JimBobSquarePants: > [6978235](https://github.com/SixLabors/ImageSharp/pull/2276/commits/69782359ba4718d232c4cd62bc592e5065a7354e) removed the `stream.CanSeek` check in the sync path, effectively activating the rewinding logic, without any impact to our test result outcome. To me this is a sign that our functional coverage is incomplete.

That's not true. The CanSeek part was superfluous as the stream was always seekable at that point as you noted elsewhere.

JimBobSquarePants: Ah, hold on. For sync, no, for async yes. Will add a non-seekable input stream to cover.

JimBobSquarePants: Have updated the implementation and test both seekable and non-seekable scenarios

src/ImageSharp/Formats/ImageDecoder.cs
JimBobSquarePants: The [general tests](https://github.com/SixLabors/ImageSharp/blob/5834c39c41a4bfeaef5d02b43d1d818e3c5c467b/tests/ImageSharp.Tests/Image/ImageTests.Decode_Cancellation.cs) here are not great. They don't test any of the decoders rather a mocked infrastructure that doesn't contain any viable image data (the header is a `Guid`). This means that the checks have to be in `WithSeekableStream` in order for those tests to pass.

IMO this highlights the dangers of using mocks. I would remove every single instance of usage if I could find the time. 

It is possible to use `BufferedReadStream` btw, you can pass the cancellation token via the constructor.

JimBobSquarePants: OK. I've replaced those tests with ones that test `IdentifyAsync(...)` and `DecodeAsync(...)` for all our decoders.

`BufferedReadStream .Read(Span<byte> buffer)` now throws if the token is cancelled.

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        Configuration configuration = options.Configuration;
+        using ChunkedMemoryStream memoryStream = new(configuration.MemoryAllocator);
+        stream.CopyTo(memoryStream, configuration.StreamProcessingBufferSize);
+        memoryStream.Position = 0;
+
+        return action(memoryStream);
+    }
+
+    internal static async Task<T> WithSeekableStreamAsync<T>(
```
antonfirsov, line 184: This is not really the little brother of the `WithSeekableStream` or the `Image.WithSeekableStreamAsync` methods. The purpose is not just to make it seekable, but to ensure it's some sort of memory stream. Would be nice to reflect this in the name.

Would be also nice to add a comment, something like:
```C#
// We need to ensure to pass an in-memory stream to the synchronous decoder methods to avoid blocking IO calls in async code.
```

antonfirsov (commented):
Almost there! 👍

tests/ImageSharp.Tests/Image/ImageTests.Decode_Cancellation.cs
```diff
@@ -13,148 +13,68 @@ public class Decode_Cancellation : ImageLoadTestBase
...
-        [Theory]
-        [InlineData(false)]
-        [InlineData(true)]
-        public Task IdentifyWithFormatAsync_CustomConfiguration_Stream(bool isInputStreamSeekable)
-        {
-            this.isTestStreamSeekable = isInputStreamSeekable;
-            _ = Task.Factory.StartNew(this.DoCancel, TaskCreationOptions.LongRunning);
+                cts.Cancel();
+                pausedStream.Release();
+            });
```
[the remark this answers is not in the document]
JimBobSquarePants: I'm struggling to follow exactly what you mean here, and I don't want to mess up the implementation. You couldn't push this fix, could you?

src/ImageSharp/Formats/ImageDecoder.cs
JimBobSquarePants: `WithSeekableMemoryStreamAsync`

tests/ImageSharp.Tests/Image/ImageTests.Decode_Cancellation.cs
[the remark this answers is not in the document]
antonfirsov: Yeah, this isn't trivial stuff. I think I will have some time tomorrow.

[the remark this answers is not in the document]
JimBobSquarePants: Thanks! Really appreciate it.

tests/ImageSharp.Tests/Image/ImageTests.Decode_Cancellation.cs
```diff
@@ -13,148 +13,116 @@ public class Decode_Cancellation : ImageLoadTestBase
...
-            };
-
-            return Assert.ThrowsAsync<TaskCanceledException>(() => Image.LoadAsync(options, this.DataStream, this.cts.Token));
+                foreach (double p in percentages)
+                {
+                    data.Add(false, file, p);
+
+                    // Do not test "direct" decoder cancellation for Identify for percentages other than 0% to avoid fine-tuning the percentages.
+                    // Cancellation should happen before we read enough data to consider the stream "identified". This can be very early for some formats/files.
+                    if (!identify || p > 0)
```
antonfirsov, line 45:
```suggestion
                    if (!identify || p == 0)
```

JimBobSquarePants:
Saw this failure in the CI run but not locally. Gonna restart the check in case it was a GitHub performance glitch.

```
[xUnit.net 00:02:18.29]     LoadAsync_IsCancellable(useMemoryStream: False, file: "Webp/testpattern_opaque_small.png", percentageOfStreamReadToCancel: 0.9) [FAIL]
  Failed LoadAsync_IsCancellable(useMemoryStream: False, file: "Webp/testpattern_opaque_small.png", percentageOfStreamReadToCancel: 0.9) [32 s]
  Error Message:
   System.TimeoutException : The operation has timed out.
  Stack Trace:
     at SixLabors.ImageSharp.Tests.ImageTests.Decode_Cancellation.LoadAsync_IsCancellable(Boolean useMemoryStream, String file, Double percentageOfStreamReadToCancel) in /home/runner/work/ImageSharp/ImageSharp/tests/ImageSharp.Tests/Image/ImageTests.Decode_Cancellation.cs:line 122
--- End of stack trace from previous location ---
```

antonfirsov:
There are likely issues with cancellation over `ZlibInflateStream` on Unixes, though weirdly I wasn't able to reproduce it in my Linux VM. We don't have to solve everything in this PR, so skipped PNG tests for now in 0eadea23254a9cc916e314620aa9756ac5541deb. Really hope I can take a look at the GIF failures tomorrow.

antonfirsov:
@JimBobSquarePants the `.WaitAsync(TimeSpan.FromSeconds(30))` is a trick to avoid hanging the test run when `PausedStream` becomes stuck. We should give sufficient time though, otherwise the test will be flaky, depending on the server load.

JimBobSquarePants:
@antonfirsov Ah Ok yeah. It's `WaitAsync` that throws the exception. 

There's something seriously funky going on then. Should we be overriding `CopyAsync` in `PausedStream`? We seem to be mixing and matching `PausedMemoryStream` and `PausedStream` behaviors through the tests (JpegDecoderTests are mappsing a `MemoryStream` to `PausedStream`). Perhaps `PausedStream` should be `PauseFileStream`?

JimBobSquarePants:
https://github.com/SixLabors/ImageSharp/actions/runs/3654144968/jobs/6174307674#step:13:267

I think there's something wrong with the VM. This test passed fine at https://github.com/SixLabors/ImageSharp/pull/2276/commits/8fdd6b0e6324e9ff405c4fc822d81ffe6cef051c and we haven't touched anything that could be related since.

JimBobSquarePants:
@antonfirsov Figured out why the metadata test is failing as it's now replicable on my machine. `TestFile` has a private Image property that is created then cloned whenever we call `TestFile.CreateRgba32Image`. This image is never disposed. 

I'm actually amazed the tests haven't thrown for this before. 

I'm now seeing the png file issue on my local machine also so shall investigate.

JimBobSquarePants:
@antonfirsov Looks like your tests definitely fail on Windows too.

image

https://github.com/SixLabors/ImageSharp/actions/runs/3671278744/jobs/6206376200#step:14:338

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -0,0 +1,254 @@
...
+    private static async Task<T> CopyToMemoryStreamAndActionAsync<T>(
+        DecoderOptions options,
+        Stream stream,
+        Func<Stream, long, CancellationToken, Task<T>> action,
+        CancellationToken cancellationToken)
+    {
+        long position = stream.CanSeek ? stream.Position : 0;
+        Configuration configuration = options.Configuration;
+        using ChunkedMemoryStream memoryStream = new(configuration.MemoryAllocator);
+        await stream.CopyToAsync(memoryStream, configuration.StreamProcessingBufferSize, cancellationToken).ConfigureAwait(false);
```
JimBobSquarePants, line 250: @antonfirsov  What I don't understand about the test that is timing out is that the entire stream is copied here in the failing condition. This means we shouldn't even reach the actual decode operation before the position-based cancellation kicks in and an exception is thrown.

antonfirsov: The position based cancellation will kick in during the copy, since the copy is utilizing reads/writes since 3a7c4f442d22ba7370bed4b49b772f01e096f274. When `PausedMemoryStream` is used `WithSeekableMemoryStreamAsync` will threat it as a `MemoryStream` and avoid wrapping it.

We'll need to press the rerun a few more times to prove it, but I think 85dc0f3 fixed the test issues. I introduced the timeout to handle a scenario where I noticed (or imagined?) that the paused stream was hanging without actually trying to understand why did it hang. Anyways, good news is that I don't see this hang happening anymore.

Instead, we were seeing the tests timing out because a decoder run can actually take longer than 30 seconds on a heavily overloaded 1-core VM.

JimBobSquarePants: Our code shouldn't, (_at least I couldn't find a reason it should following refactoring_) be hanging since the wrapping since the sync wrapping handles exceptions and always returns a Task. 

Fingers crossed this has it! I'll trigger the run again now.

antonfirsov:
Re 2145794: the cache in `TestFile` was introduced very long time ago to speed-up overall test execution. Instead of removing it, I would either avoid `TestFile.GetImage()` in decoder tests, or introduce a `TestFile.GetImage(..., bool skipCache=false)` overload.

Alternatively, we can measure test run times with/without the cache, and drop it if we don't see much difference in test execution times, but I'm not big fan of making this many conceptual changes in a single PR.

JimBobSquarePants:
>Alternatively, we can measure test run times with/without the cache, and drop it if we don't see much difference in test execution times, but I'm not big fan of making this many conceptual changes in a single PR.

I haven't noticed any change in times at all actually.

src/ImageSharp/Formats/ImageDecoder.cs
JimBobSquarePants: That's 2 successful runs in a row. I'll trigger a 3rd now.

src/ImageSharp/Formats/SpecializedImageDecoder{T}.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// <remarks>
+    /// This method is designed to support the ImageSharp internal infrastructure and is not recommended for direct use.
+    /// </remarks>
+    /// <typeparam name="TPixel">The pixel format.</typeparam>
+    /// <param name="options">The specialized decoder options.</param>
+    /// <param name="stream">The <see cref="Stream" /> containing image data.</param>
+    /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+    /// <returns>The <see cref="Image{TPixel}" />.</returns>
+    /// <exception cref="ImageFormatException">Thrown if the encoded image contains errors.</exception>
+    protected abstract Image<TPixel> Decode<TPixel>(T options, Stream stream, CancellationToken cancellationToken)
```
antonfirsov, line 61: Here we still lack the option for defaulting `CancellationToken`.

antonfirsov (commented):
One thing I would consider is to introduce shared singletons for now stateless decoders, so

```C#
new JpegDecoder().Decode(new JpegDecoderOptions() { ...}, stream);
```

becomes

```C#
JpegDecoder.Shared.Decode(new JpegDecoderOptions() { ...}, stream);
```

src/ImageSharp/Formats/SpecializedImageDecoder{T}.cs
JimBobSquarePants: That happens on the public caller. This is the protected inner call.

src/ImageSharp/Formats/ImageDecoder.cs
antonfirsov: I also did it 3 times, should be good now hopefully.

src/ImageSharp/Formats/SpecializedImageDecoder{T}.cs
antonfirsov: Oops, commented on the wrong method.

If you type `new JpegDecoder().DecodeAsync( ...`  you won't see an overload that takes `JpegDecoderOptions` without also asking for a `CancellationToken`.

One reason why I aim to add tests that validate all public methods directly it helps avoiding such mistakes.

JimBobSquarePants: Ah right... That was intentional. My thought was that more often than not a cancellation token will be passed. I'll update.

JimBobSquarePants: Fixed. I've updated all tests calling those APIs directly to remove the default passed token.

JimBobSquarePants:
@antonfirsov I've introduced shared singletons for all our decoders and updated all references to them. I went with `Instance` though over `Shared` to match existing convention. 

Please let me know if there's anything else you consider blocking. I'm very keen to get this merged.

antonfirsov (approved):
Great work, thanks for the patience!

JimBobSquarePants:
Thanks for really pushing me here @antonfirsov Without your input I would have made many mistakes.

PR #2281, opened 2022-10-29
Enable Nullable

stefannikolei:
#2231 
First step. Enabled nullable analysis. Disabled for all files with an annotation


stefannikolei:
Dammit I will redo it..

PR #2289, opened 2022-11-07
Fix various Gif Decoder/Encoder behaviors. 

### LINKED ISSUE #2288: Gifs sometimes get optimized incorrectly
### Prerequisites

### ImageSharp version

v2.1.3

ImageSharp.Web 2.0.2

Linux and Mac

### .NET Framework version

6

### Description

When doing any processing on some gifs, the output isn't correct.

**Example 1**
This is the original gif:
51819_osvart_orkushow
When sending it through ImageSharp.Web with any query parameters, even inconsequential ones, the output gif is extremely slow (with `rmode=crop`):
imagesharp-51819_osvart_orkushow

**Example 2**
This is the original gif:
irishpublysing
When sending it through ImageSharp.Web with any query parameters, even inconsequential ones, the output gif has the first "V" in the smaller text almost turned into a "M" (with `rmode=crop`):
imagesharp-irishpublysing

### Steps to Reproduce

1. Run a gif through ImageSharp.Web
2. It may or may not have things that are incorrect

### Images

51819_osvart_orkushow
imagesharp-51819_osvart_orkushow

irishpublysing
imagesharp-irishpublysing

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2288 

This PR does a few things
- Ensures that Graphics Control Extension blocks are read and written correctly
- No longer clones the entire previous frame on decode and only write what we need. This was causing some gif sizes to explode.
- Fixes decode transparency flag handling.
- Ensures palette quantizer enables the whole palette
- Changes lower bit threshold for octree hack to ensure a transparent palette entry is maintained
- Allows mixing of global and local color tables per frame as per specification.
- Fixes the comparer to output the correct frames index.
- Fixes bad XMP handling

OptionalExtensionsShouldBeHandledProperly_Rgba32_issue_2288
OptionalExtensionsShouldBeHandledProperly_Rgba32_issue_2288_2

### DIFF
--- a/src/ImageSharp/Formats/Gif/GifDecoderCore.cs
+++ b/src/ImageSharp/Formats/Gif/GifDecoderCore.cs
@@ -514,7 +514,7 @@ internal sealed class GifDecoderCore : IImageDecoderInternals
         int descriptorBottom = descriptorTop + descriptor.Height;
         int descriptorLeft = descriptor.Left;
         int descriptorRight = descriptorLeft + descriptor.Width;
-        byte transparentIndex = this.graphicsControlExtension.TransparencyIndex;
+        byte transIndex = this.graphicsControlExtension.TransparencyIndex;
         int colorTableMaxIdx = colorTable.Length - 1;
 
         for (int y = descriptorTop; y < descriptorBottom && y < imageHeight; y++)
@@ -573,14 +573,17 @@ internal sealed class GifDecoderCore : IImageDecoderInternals
                 for (int x = descriptorLeft; x < descriptorRight && x < imageWidth; x++)
                 {
                     int rawIndex = Unsafe.Add(ref indicesRowRef, x - descriptorLeft);
-                    int transIndex = Numerics.Clamp(rawIndex, 0, Math.Max(transparentIndex, colorTableMaxIdx));
-                    int index = Numerics.Clamp(rawIndex, 0, colorTableMaxIdx);
-                    if (transparentIndex != transIndex)
+
+                    // Treat any out of bounds values as transparent.
+                    if (rawIndex > colorTableMaxIdx || rawIndex == transIndex)
                     {
-                        ref TPixel pixel = ref Unsafe.Add(ref rowRef, x);
-                        Rgb24 rgb = colorTable[index];
-                        pixel.FromRgb24(rgb);
+                        continue;
                     }
+
+                    int index = Numerics.Clamp(rawIndex, 0, colorTableMaxIdx);
+                    ref TPixel pixel = ref Unsafe.Add(ref rowRef, x);
+                    Rgb24 rgb = colorTable[index];
+                    pixel.FromRgb24(rgb);
                 }
             }
         }
--- a/src/ImageSharp/Formats/Gif/GifDecoderCore.cs
+++ b/src/ImageSharp/Formats/Gif/GifDecoderCore.cs
@@ -479,7 +479,7 @@ internal sealed class GifDecoderCore : IImageDecoderInternals
                 prevFrame = previousFrame;
             }
 
-            currentFrame = image.Frames.AddFrame(new ImageFrame<TPixel>(this.configuration, previousFrame.Width, previousFrame.Height));
+            currentFrame = image.Frames.CreateFrame();
 
             this.SetFrameMetadata(currentFrame.Metadata, false);
 
--- a/src/ImageSharp/Formats/Gif/GifDecoderCore.cs
+++ b/src/ImageSharp/Formats/Gif/GifDecoderCore.cs
@@ -165,6 +165,11 @@ internal sealed class GifDecoderCore : IImageDecoderInternals
             this.globalColorTable?.Dispose();
         }
 
+        if (image is null)
+        {
+            GifThrowHelper.ThrowNoData();
+        }
+
         return image;
     }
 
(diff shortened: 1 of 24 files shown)


src/ImageSharp/Processing/Processors/Quantization/OctreeQuantizer{TPixel}.cs
```diff
@@ -106,7 +106,7 @@ public void AddPaletteColors(Buffer2DRegion<TPixel> pixelRegion)
         // for higher bit depths. Lower bit depths will correctly reduce the palette.
         // TODO: Investigate more evenly reduced palette reduction.
```
JimBobSquarePants, line 107: I'd love to figure this out properly.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -554,7 +558,7 @@ private void ReadFrameColors<TPixel>(ref Image<TPixel> image, ref ImageFrame<TPi
             {
                 for (int x = descriptorLeft; x < descriptorRight && x < imageWidth; x++)
                 {
-                    int index = Numerics.Clamp(Unsafe.Add(ref indicesRowRef, x - descriptorLeft), 0, colorTableMaxIdx);
+                    int index = Numerics.Clamp(Unsafe.Add(ref indicesRowRef, x - descriptorLeft), 0, Math.Max(transIndex, colorTableMaxIdx));
```
JimBobSquarePants, line 561: Some images use a transparent index that is the same as the length of the palette.

saucecontrol: As far as I've seen, most viewers treat OOB transparent index as 0.  That's also the behavior written into the PNG spec.  Here's an example where 0 works and last index doesn't.

bad_transparent_idx

saucecontrol:
I just happen to have run a ton of GIFs through ImageSharp this week while testing my new GIFLIB wrapper.  Don't know if you've already fixed this, but this image fails to load in 2.1.3.  Issue is it has a truncated XMP doc embedded (just one 255 byte segment of it)

truncated_xmp

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
JimBobSquarePants: That's interesting. I found that the second gif in the referenced issue had a frame with a local table that contained 64 colors with a transparent index of 64. Browsers/Windows were rendering it just fine, but we were not.

image

Here's your input gif processed with this PR code.

OptionalExtensionsShouldBeHandledProperly_Rgba32_issue_2288_3

JimBobSquarePants: I use this. Stumbled upon it a few years back and it's proven incredibly useful

https://github.com/ata4/gifiddle

saucecontrol: Oh wow, that's great, thanks!

I see what's happening here now.  Before this fix, you're not setting a transparent color because it's out of range, but you're clamping the image values into palette range, so pixels meant to be transparent are taking the last color.  This was the result.

oob_color_is

Now you're treating the last palette color as transparent and clamping OOB image values to that color, making the transparency work, but also making any pixels that were supposed to be that last palette color transparent as well.  It's not creating artifacts as visible as skipping transparency, but it's also not completely correct.

In my implementation, I made the same mistake, but in the other direction.  I use shared code for all animated image formats, so I followed the PNG rules of treating OOB image values as the *first* color in the palette.  That meant I also had to treat invalid transparency index as meaning the first color was supposed to be transparent to get things to look right.  Hence my initial comment 😅

So yeah, the browser behavior is better.  Don't clamp the values but rather assume the palette is actually bigger than the image says.

JimBobSquarePants: Pushed a better fix now. I treat anything greater than the max palette index or equal to the flagged transparent index as transparent.

I only clamp following that check to ensure we don't hit OOB when reading the palette.

tests/ImageSharp.Tests/Formats/Tiff/TiffEncoderMultiframeTests.cs
```diff
@@ -30,7 +30,9 @@ public void TiffEncoder_EncodeMultiframe_WithPreview<TPixel>(TestImageProvider<T
 
     [Theory]
     [WithFile(TestImages.Gif.Receipt, PixelTypes.Rgb24)]
-    [WithFile(TestImages.Gif.Issues.BadDescriptorWidth, PixelTypes.Rgba32)]
+
+    // MAGICK decoder makes the same mistake we did and clones the proceeding frame overwriting the differences.
+    // [WithFile(TestImages.Gif.Issues.BadDescriptorWidth, PixelTypes.Rgba32)]
     public void TiffEncoder_EncodeMultiframe_Convert<TPixel>(TestImageProvider<TPixel> provider)
```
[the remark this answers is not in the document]
dlemstra: That's odd I am seeing this when I read that image inside Magick.NET and use the `Immediate Window`:

```
magickImageCollection [0].Width
500
magickImageCollection [1].Width
350
magickImageCollection [2].Width
332
magickImageCollection [3].Width
500
```

PR #2312, opened 2023-01-01
Remove some nullable disable comments

stefannikolei:
#2231

Removed the nullable disable in all places where fixing it was an easy task


stefannikolei:
After these only 10 files have the annotation `nullable disable`

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
```diff
@@ -27,7 +27,7 @@ public static IImageEncoder DetectEncoder(this Image source, string filePath)
         Guard.NotNull(filePath, nameof(filePath));
 
         string ext = Path.GetExtension(filePath);
-        IImageFormat format = source.GetConfiguration().ImageFormatsManager.FindFormatByFileExtension(ext);
+        IImageFormat? format = source.GetConfiguration().ImageFormatsManager.FindFormatByFileExtension(ext);
```
JimBobSquarePants, line 30: Ah... This is one of the areas I want to have a look at. Ideally this method should be rewritten to use the try-out pattern.

e.g 

``` c#
bool ImageFormatsManager.TryFindFormatByFileExtension(ext, [NotNullWhen(true)] out IImageFormat? format)
```

We'd return named tuples for async equivalents.

src/ImageSharp/Formats/Bmp/BmpImageFormatDetector.cs
```diff
@@ -15,7 +14,7 @@ public sealed class BmpImageFormatDetector : IImageFormatDetector
     public int HeaderSize => 2;
 
     /// <inheritdoc/>
-    public IImageFormat DetectFormat(ReadOnlySpan<byte> header)
+    public IImageFormat? DetectFormat(ReadOnlySpan<byte> header)
```
JimBobSquarePants, line 17: Same here. We'd convert to use `TryDetectFormat`

src/ImageSharp/Formats/ImageFormatManager.cs
```diff
@@ -93,7 +92,7 @@ public void AddImageFormat(IImageFormat format)
     /// </summary>
     /// <param name="extension">The extension to discover</param>
     /// <returns>The <see cref="IImageFormat"/> if found otherwise null</returns>
-    public IImageFormat FindFormatByFileExtension(string extension)
+    public IImageFormat? FindFormatByFileExtension(string extension)
```
JimBobSquarePants, line 95: All these would use `TryXX`

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
stefannikolei: Ok gonna rewrite that.

JimBobSquarePants: Legend. Thanks! It's something I very much wish we'd done from the start.

src/ImageSharp/Formats/ImageFormatManager.cs
```diff
@@ -91,8 +92,9 @@ public void AddImageFormat(IImageFormat format)
     /// For the specified file extensions type find the e <see cref="IImageFormat"/>.
     /// </summary>
     /// <param name="extension">The extension to discover</param>
-    /// <returns>The <see cref="IImageFormat"/> if found otherwise null</returns>
-    public IImageFormat? FindFormatByFileExtension(string extension)
+    /// <param name="format">The <see cref="IImageFormat"/> if found otherwise null</param>
+    /// <returns>False if no format was found</returns>
+    public bool TryFindFormatByFileExtension(string extension, [NotNullWhen(true)] out IImageFormat? format)
```
JimBobSquarePants, line 97: 👍

src/ImageSharp/Image.FromBytes.cs
```diff
@@ -27,7 +26,7 @@ public static IImageFormat DetectFormat(ReadOnlySpan<byte> data)
     /// <param name="data">The byte span containing encoded image data to read the header from.</param>
     /// <exception cref="ArgumentNullException">The options are null.</exception>
     /// <returns>The mime type or null if none found.</returns>
-    public static IImageFormat DetectFormat(DecoderOptions options, ReadOnlySpan<byte> data)
+    public static IImageFormat? DetectFormat(DecoderOptions options, ReadOnlySpan<byte> data)
```
JimBobSquarePants, line 29: We still need to convert this and above to use the Try...Out pattern

src/ImageSharp/Processing/Processors/Convolution/ConvolutionProcessorHelpers.cs
```diff
@@ -99,7 +98,7 @@ internal static float[] CreateGaussianSharpenKernel(int size, float weight)
     /// <param name="row">The resulting 1D row vector, if possible.</param>
     /// <param name="column">The resulting 1D column vector, if possible.</param>
     /// <returns>Whether or not <paramref name="matrix"/> was linearly separable.</returns>
-    public static bool TryGetLinearlySeparableComponents(this DenseMatrix<float> matrix, out float[] row, out float[] column)
+    public static bool TryGetLinearlySeparableComponents(this DenseMatrix<float> matrix, out float[]? row, out float[]? column)
```
JimBobSquarePants, line 101: We can add the `NotNullWhen` attributes here

src/ImageSharp/Image.FromBytes.cs
stefannikolei: Do you want to convert InternalDetectFormat from Image.Decode.cs to the Try Pattern or just return IImageFormat?

stefannikolei:
> After these only 10 files have the annotation `nullable disable`

Ok i have no clue how I searched. In total there a more nullable disable in the Solution.. Probably I searched while in a specific scope ... oO

src/ImageSharp/Image.FromBytes.cs
JimBobSquarePants: Ideally it should use the Try pattern all the way down.

JimBobSquarePants: Having a second look and we wouldn't be able to adopt the Try pattern there actually. The method is used within a lambda which doesn't allow ref or out parameters..

JimBobSquarePants (approved):
I think this is good to merge for now. Thanks!

I want to have a discussion about how to handle the various `Image.Load` APIs next.

Issue #1489, opened 2020-12-31
Add memory size to WrapMemory(void* ...);


vpenades:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

I've noticed the new [WrapMemory](https://github.com/SixLabors/ImageSharp/blob/cd1a25efeeb4b243cbfcfff8950fc215523077b7/src/ImageSharp/Image.WrapMemory.cs#L422) method allowing to initialize an image with a raw, low level pointer, which is great.

I would suggest to add a memory length argument, so the method would look like this:

```c#
WrapMemory<TPixel>(
            void* pointer,
            int byteLength, // byte length of the memory allocated
            int width,
            int height)
```

It might look redundant, since the underlaying code _does_ know which is the expected memory size, and the caller is responsible to allocate that amount of memory.

But in practice, we do know that things are not that easy, and in the worst case scenario it can happen that a developer might misunderstand or code an incorrect implementation that allocates _less_ memory than required, and by sheer luck, it doesn't crash and might seem the code is right.

Given that in general, anybody allocating low level memory _does know_ the amount of allocated memory, and most APIs I've seen that provide pointers, also provide the memory length in bytes, it seems reasonable that the WrapMemory method checks both the pointer not being null, and the memory length being enough to meet the Image needs.

So in essence, this argument is only used as an extra safety check, and given the dangerous nature of unmanaged code, it would be very useful to detect errors early.

### System Configuration

- ImageSharp version: 1.0.2
- Other ImageSharp packages and versions: -
- Environment (Operating system, version and so on): all
- .NET Framework version: all
- Additional information:

antonfirsov:
Makes sense to me. @Sergio0694 thoughts?

James-Jackson-South:
Let's move this to a discussion where it belongs.

JimBobSquarePants:
Let's move this to a discussion where it belongs.

JimBobSquarePants:
Closed via #2313

PR #2313, opened 2023-01-04
Add byteLength to WrapMemory

stefannikolei:
* Added byteLength parameter to WrapMemory method and check that the memory is at least the size of byteLength

See #1489


src/ImageSharp/Image.WrapMemory.cs
```diff
@@ -403,6 +403,7 @@ public abstract partial class Image
     /// <typeparam name="TPixel">The pixel type</typeparam>
     /// <param name="configuration">The <see cref="Configuration"/></param>
     /// <param name="pointer">The pointer to the target memory buffer to wrap.</param>
+    /// <param name="byteLength">The byte length of the memory allocated.</param>
```
JimBobSquarePants, line 406: You're gonna hate me but I think `bufferSizeInBytes` is a better parameter name as it's more consistant with our existing memory allocator APIs

JimBobSquarePants (approved):
Wonderful. Thanks!

PR #2316, opened 2023-01-11
Add ApiApprovalTests

stefannikolei:
* This test checks the public api of ImageSharp for changes.

Perhaps this is a small addition which can be a benefit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -227,3 +227,4 @@ artifacts/
 #lfs
 hooks/**
 lfs/**
+*.received.txt
--- a/tests/ImageSharp.Tests/ApiApprovalTests.cs
+++ b/tests/ImageSharp.Tests/ApiApprovalTests.cs
@@ -3,6 +3,7 @@
 
 using System.Diagnostics;
 using System.Reflection;
+using System.Runtime.Versioning;
 using PublicApiGenerator;
 using Shouldly;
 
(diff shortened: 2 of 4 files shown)


stefannikolei:
This PR is here to discuss this. Wanted to show how this could look.

Another way how to do this could be to adapt the way the dotnet team does it. With Microsoft.CodeAnalysis.PublicApiAnalyzers


But first... Do you want such a functionality in the code?

JimBobSquarePants:
@stefannikolei So what does this do exactly?

stefannikolei:
@JimBobSquarePants this test will fail when the public api of the project changes. 
When the test fails a diff tool will be opened with the changes.

JimBobSquarePants:
But what if we want the API to change? We're going through a fair amount of churn for v3 just now.

stefannikolei:
Yeah then the approved.txt must be updated. 

This is just an idea. Best timing for such a thing would be when v3 is ready. 
So afterwards you are able to spot breaking api changes easier because the test failed or the approved.txt was updated. 

I mentioned another implementation which comes from Microsoft and is used in the runtime and aspnetcore. They use an analyzer which will throw an error. 
There the advantage is that you have two files. One for the released api interface and one for the unreleased. Tho you can track the additions for a new major release.

JimBobSquarePants:
Aye, something to look at post V3 then. Very interesting!!

PR #2317, opened 2023-01-12
Remove ImageFormat out params and return on Metadata

JimBobSquarePants:
### Prerequisites

### Description

This PR replaces and build upon #2092. Touches #2231 

- [x] Expose `ImageInfo` type for easier decoder implementation.
- [x] Add `IImageInfo` to `Image.Metadata` as `IImageInfo?` via the `DecodedImageFormat` property and set on decode.
- [x] Remove API overloads using the `out IImageFormat` parameter and cleanup.
- [x] Normalizes `DetectFormat`, `Identify`, and `Load` APIs to all throw `UnknownImageFormatException` on failure.
- [x] Updates `ImageFormatManager.GetXX` methods to throw `UnknownImageFormatException` on failure.
- [x] Add tests.

@stefannikolei Apologies but I sent you the wrong direction with your previous PR. It's best we return the none nullable type and throw when something goes wrong with the aforementioned APIs.


stefannikolei:
@JimBobSquarePants all good. It's no wasted time. All are learning on this path.

src/ImageSharp/Formats/ImageDecoder.cs
```diff
@@ -17,49 +16,85 @@ public abstract class ImageDecoder : IImageDecoder
...
-              options,
-              stream,
-              s => this.Decode<TPixel>(options, s, default));
+    {
+        Image<TPixel> image = WithSeekableStream(
+            options,
+            stream,
+            s => this.Decode<TPixel>(options, s, default));
+
+        this.SetDecoderFormat(options.Configuration, image);
```
tocsoft, line 25: rather than calling this at every overload... could this not just happen inside the `Decode()` call above? well instead of calling the abstract `Decode<>()` directly create a `DecodeInternal<>()` (name?) that calls the abstract method and sets the decoder rather than have to repeat the call on every overload.

tests/ImageSharp.Tests/TestUtilities/Tests/TestImageProviderTests.cs
```diff
@@ -363,8 +363,11 @@ public static void DoTestThreadSafe(Action action)
             }
         }
 
-        protected override IImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
-            => this.Decode<Rgba32>(this.CreateDefaultSpecializedOptions(options), stream, cancellationToken);
+        protected override ImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
+        {
+            Image<Rgba32> image = this.Decode<Rgba32>(this.CreateDefaultSpecializedOptions(options), stream, cancellationToken);
```
tocsoft, line 368:
```suggestion
            using Image<Rgba32> image = this.Decode<Rgba32>(this.CreateDefaultSpecializedOptions(options), stream, cancellationToken);
```

tests/ImageSharp.Tests/TestUtilities/Tests/TestImageProviderTests.cs
```diff
@@ -403,8 +406,11 @@ public static void DoTestThreadSafe(Action action)
...
```
tocsoft, line 411:
```suggestion
            using Image<Rgba32> image = this.Decode<Rgba32>(this.CreateDefaultSpecializedOptions(options), stream, cancellationToken);
```

tests/ImageSharp.Tests/TestUtilities/ReferenceCodecs/MagickReferenceDecoder.cs
```diff
@@ -72,8 +72,11 @@ protected override Image<TPixel> Decode<TPixel>(DecoderOptions options, Stream s
     protected override Image Decode(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
         => this.Decode<Rgba32>(options, stream, cancellationToken);
 
-    protected override IImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
-        => this.Decode<Rgba32>(options, stream, cancellationToken);
+    protected override ImageInfo Identify(DecoderOptions options, Stream stream, CancellationToken cancellationToken)
+    {
+        Image<Rgba32> image = this.Decode<Rgba32>(options, stream, cancellationToken);
```
tocsoft, line 77: needs disposing
```suggestion
        using Image<Rgba32> image = this.Decode<Rgba32>(options, stream, cancellationToken);
```

src/ImageSharp/IImageInfo.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Six Labors.
+// Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
```
stefannikolei, line 2: I think this Interface and Image are not longer used in the code. Can probably get dropped

src/ImageSharp/Formats/ImageDecoder.cs
JimBobSquarePants: Not that I can see. Each of the sync/async calls are different calling the generic/non-generic types.

src/ImageSharp/IImageInfo.cs
JimBobSquarePants: I've been considering this also. 

`IImageInfo` is used in two places in `ImageInfoExtensions` to get the size and bounds.
`IImage` is only referenced by `Image` and is unused anywhere.

@tocsoft what do you think?

JimBobSquarePants: OK. So, I actually ran with this and in my opinion, things look much better. 

`ImageInfo` is now unsealed and `Image` inherits that.
 `Size Size` and `Rectangle Bounds` are now properties of `ImageInfo` not extension methods which remove the odd inconsistency between those extensions and the `Width`/`Height` properties,

tocsoft: seems fine to me, we seem to be dropping interfaces where we can

JimBobSquarePants: Yeah. I think we went a little heavy on them originally and it actually limited our ability to improve in some areas.

What are your thoughts on the rest of the PR. OK to merge?

tocsoft (approved):

PR #2320, opened 2023-01-17
Remove #nullable disable from ExifProfile

stefannikolei:
### Prerequisites

### Description
Resolves the nullable issues in the ExifProfile space #2231

@JimBobSquarePants Do you think we can tackle it like that, or do you have other opinions how we could/should do it?

I had a bit of a struggle with the Generic TValueType. I tried to introduce `were TValueType : struct`. But there are cases where `TValueType = string` --> zonk

I added an InvalidOperationException to 'private static Encoding JIS0208Encoding' to not expose Encoding? because that will never happen. What do you think about this? Wanna go this way or should I add an ! here to dismiss the warning

### DIFF
--- a/src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
+++ b/src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
@@ -276,9 +276,15 @@ internal class TiffDecoderCore : IImageDecoderInternals
     private void DecodeImageWithStrips<TPixel>(ExifProfile tags, ImageFrame<TPixel> frame, CancellationToken cancellationToken)
         where TPixel : unmanaged, IPixel<TPixel>
     {
-        int rowsPerStrip = tags.GetValue(ExifTag.RowsPerStrip) != null
-            ? (int)tags.GetValue(ExifTag.RowsPerStrip).Value
-            : TiffConstants.RowsPerStripInfinity;
+        int rowsPerStrip;
+        if (tags.TryGetValue(ExifTag.RowsPerStrip, out IExifValue<Number> value))
+        {
+            rowsPerStrip = (int)value.Value;
+        }
+        else
+        {
+            rowsPerStrip = TiffConstants.RowsPerStripInfinity;
+        }
 
         Array stripOffsetsArray = (Array)tags.GetValueInternal(ExifTag.StripOffsets).GetValue();
         Array stripByteCountsArray = (Array)tags.GetValueInternal(ExifTag.StripByteCounts).GetValue();
@@ -320,8 +326,18 @@ internal class TiffDecoderCore : IImageDecoderInternals
         int width = pixels.Width;
         int height = pixels.Height;
 
-        int tileWidth = (int)tags.GetValue(ExifTag.TileWidth).Value;
-        int tileLength = (int)tags.GetValue(ExifTag.TileLength).Value;
+        if (!tags.TryGetValue(ExifTag.TileWidth, out IExifValue<Number> valueWidth))
+        {
+            ArgumentNullException.ThrowIfNull(valueWidth);
+        }
+
+        if (!tags.TryGetValue(ExifTag.TileWidth, out IExifValue<Number> valueLength))
+        {
+            ArgumentNullException.ThrowIfNull(valueLength);
+        }
+
+        int tileWidth = (int)valueWidth.Value;
+        int tileLength = (int)valueLength.Value;
         int tilesAcross = (width + tileWidth - 1) / tileWidth;
         int tilesDown = (height + tileLength - 1) / tileLength;
 
@@ -775,8 +791,7 @@ internal class TiffDecoderCore : IImageDecoderInternals
     /// <returns>The image width.</returns>
     private static int GetImageWidth(ExifProfile exifProfile)
     {
-        IExifValue<Number> width = exifProfile.GetValue(ExifTag.ImageWidth);
-        if (width == null)
+        if (!exifProfile.TryGetValue(ExifTag.ImageWidth, out IExifValue<Number> width))
         {
             TiffThrowHelper.ThrowImageFormatException("The TIFF image frame is missing the ImageWidth");
         }
@@ -793,8 +808,7 @@ internal class TiffDecoderCore : IImageDecoderInternals
     /// <returns>The image height.</returns>
     private static int GetImageHeight(ExifProfile exifProfile)
     {
-        IExifValue<Number> height = exifProfile.GetValue(ExifTag.ImageLength);
-        if (height == null)
+        if (!exifProfile.TryGetValue(ExifTag.ImageLength, out IExifValue<Number> height))
         {
             TiffThrowHelper.ThrowImageFormatException("The TIFF image frame is missing the ImageLength");
         }
--- a/src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
+++ b/src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
@@ -793,7 +793,7 @@ internal class TiffDecoderCore : IImageDecoderInternals
     {
         if (!exifProfile.TryGetValue(ExifTag.ImageWidth, out IExifValue<Number> width))
         {
-            TiffThrowHelper.ThrowImageFormatException("The TIFF image frame is missing the ImageWidth");
+            TiffThrowHelper.ThrowInvalidImageContentException("The TIFF image frame is missing the ImageWidth");
         }
 
         DebugGuard.MustBeLessThanOrEqualTo((ulong)width.Value, (ulong)int.MaxValue, nameof(ExifTag.ImageWidth));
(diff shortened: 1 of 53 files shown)


src/ImageSharp/Metadata/Profiles/Exif/ExifProfile.cs
```diff
@@ -134,7 +136,7 @@ public IReadOnlyList<IExifValue> Values
     /// <returns>
     /// The <see cref="Image{TPixel}"/>.
     /// </returns>
-    public Image<TPixel> CreateThumbnail<TPixel>()
+    public Image<TPixel>? CreateThumbnail<TPixel>()
```
stefannikolei, line 139: Question if this should throw, or perhaps change it to TryCreateThumbnail?

src/ImageSharp/Metadata/Profiles/Exif/Values/ExifUcs2String.cs
```diff
@@ -22,11 +22,11 @@ private ExifUcs2String(ExifUcs2String value)
 
     public override ExifDataType DataType => ExifDataType.Byte;
 
-    protected override string StringValue => this.Value;
+    protected override string? StringValue => this.Value;
```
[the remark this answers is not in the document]
stefannikolei: Ok compiler tells me that it can be removed in the other places

src/ImageSharp/Metadata/Profiles/Exif/ExifProfile.cs
JimBobSquarePants: It certainly shouldn't throw as EXIF metadata is not guaranteed to contain a thumbnail. I would use `TryCreate`.

src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
```diff
@@ -27,7 +26,8 @@ internal static class ExifEncodedStringHelpers
 
     // 20932 EUC-JP Japanese (JIS 0208-1990 and 0212-1990)
     // https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding?view=net-6.0
-    private static Encoding JIS0208Encoding => CodePagesEncodingProvider.Instance.GetEncoding(20932);
+    private static Encoding JIS0208Encoding => CodePagesEncodingProvider.Instance.GetEncoding(20932) ??
+                                               throw new InvalidOperationException();
```
JimBobSquarePants, line 30: We should revisit how we use this. https://github.com/dotnet/runtime/issues/31229 says to always use `Encoding`

src/ImageSharp/Metadata/Profiles/Exif/ExifProfile.cs
```diff
@@ -161,9 +163,9 @@ public Image<TPixel> CreateThumbnail<TPixel>()
     /// <param name="tag">The tag of the exif value.</param>
     /// <returns>The value with the specified tag.</returns>
     /// <typeparam name="TValueType">The data type of the tag.</typeparam>
-    public IExifValue<TValueType> GetValue<TValueType>(ExifTag<TValueType> tag)
+    public IExifValue<TValueType>? GetValue<TValueType>(ExifTag<TValueType> tag)
```
JimBobSquarePants, line 166: This should probably be `TryGet` since we do not throw.

src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
```diff
@@ -320,8 +326,18 @@ private void DecodeImageWithTiles<TPixel>(ExifProfile tags, ImageFrame<TPixel> f
         int width = pixels.Width;
         int height = pixels.Height;
 
-        int tileWidth = (int)tags.GetValue(ExifTag.TileWidth).Value;
-        int tileLength = (int)tags.GetValue(ExifTag.TileLength).Value;
+        if (!tags.TryGetValue(ExifTag.TileWidth, out IExifValue<Number> valueWidth))
+        {
+            ArgumentNullException.ThrowIfNull(valueWidth);
```
JimBobSquarePants, line 331: This should be an `InvalidImageContentException`

src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
```diff
@@ -775,8 +791,7 @@ private int CalculateStripBufferSize(int width, int height, int plane = -1)
     /// <returns>The image width.</returns>
     private static int GetImageWidth(ExifProfile exifProfile)
     {
-        IExifValue<Number> width = exifProfile.GetValue(ExifTag.ImageWidth);
-        if (width == null)
+        if (!exifProfile.TryGetValue(ExifTag.ImageWidth, out IExifValue<Number> width))
         {
             TiffThrowHelper.ThrowImageFormatException("The TIFF image frame is missing the ImageWidth");
```
JimBobSquarePants, line 796: Just realized these are not using the correct exception type. It should be `InvalidImageContentException`

src/ImageSharp/Formats/Tiff/TiffEncoderCore.cs
```diff
@@ -164,8 +164,6 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream, CancellationToken
         {
             cancellationToken.ThrowIfCancellationRequested();
```
JimBobSquarePants, line 166: No idea what the snippet was supposed to be doing. Deleting it was wise.

src/ImageSharp/Metadata/Profiles/Exif/ExifProfile.cs
```diff
@@ -294,6 +314,7 @@ private void SyncResolution(ExifTag<Rational> tag, double resolution)
         this.SetValue(tag, newResolution);
     }
 
+    [MemberNotNull(nameof(values))]
```
JimBobSquarePants, line 317: Didn't know about this attribute. Cool!

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
```diff
@@ -20,7 +19,7 @@ public ExifReader(byte[] exifData)
     {
     }
 
-    public ExifReader(byte[] exifData, MemoryAllocator allocator)
+    public ExifReader(byte[] exifData, MemoryAllocator? allocator)
         : base(new MemoryStream(exifData ?? throw new ArgumentNullException(nameof(exifData))), allocator)
```
JimBobSquarePants, line 23: It looks like you've changed the wrong field here. `allocator` is not null `exifData` is maybe null

src/ImageSharp/Metadata/Profiles/Exif/ExifTagDescriptionAttribute.cs
```diff
@@ -29,10 +28,10 @@ public ExifTagDescriptionAttribute(object value, string description)
     /// <returns>
     /// The <see cref="string"/>.
     /// </returns>
-    public static string GetDescription(ExifTag tag, object value)
+    public static string? GetDescription(ExifTag tag, object? value)
```
JimBobSquarePants, line 31: `TryGetDescription` since we don't explicitly throw.

src/ImageSharp/Metadata/Profiles/Exif/ExifWriter.cs
```diff
@@ -179,7 +178,9 @@ private static IExifValue GetOffsetValue(List<IExifValue> ifdValues, List<IExifV
                 return ifdValues[index];
             }
 
-            ExifValue result = ExifValues.Create(offset);
+            ExifValue? result = ExifValues.Create(offset);
+            Guard.NotNull(result);
```
JimBobSquarePants, line 182: Shouldn't this be an `if` wrapping the add rather than throwing?

src/ImageSharp/Metadata/Profiles/Exif/ExifWriter.cs
```diff
@@ -219,7 +220,7 @@ private List<IExifValue> GetPartValues(ExifParts part)
 
     private static bool HasValue(IExifValue exifValue)
     {
-        object value = exifValue.GetValue();
+        object? value = exifValue.GetValue();
```
JimBobSquarePants, line 223: This reads like `IExifValue.GetValue` should use `TryGetValue`

src/ImageSharp/Metadata/Profiles/Exif/ExifReader.cs
JimBobSquarePants: Oh! I misread. We NEVER call a version of the reader that passes an allocator.

src/ImageSharp/Metadata/Profiles/Exif/ExifWriter.cs
JimBobSquarePants: Oh wow! Not refactoring that!!

JimBobSquarePants (approved):
LGTM. Thanks! 👍

PR #2324, opened 2023-01-23
Allow zero DPI and only throw at EOF when not enough data

### LINKED ISSUE #2315: Exception Input stream does not have enough bytes in version 2.1.1+
### Prerequisites

### ImageSharp version

2.1.1 - 2.1.3

2.1.1 - 2.1.3

Windows 10

### .NET Framework version

.NET 6

### Description

I found valid (from my point of view) JPG image that fails to load from starting from version 2.1.1.
2.1.0 and lower and also other image processing libs works fine with this image. Image attached.

### Steps to Reproduce

`using var image = Image.Load("image path");`

### Images

site137

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2315 

The image has two issues. 

1. Zero density unit values in the JFIF header.
2. Erroneous marker data at the end of the stream following the image data.

I've relaxed the decoder to cater for these cases.

 cc\ @br3aker

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -356,6 +356,18 @@ internal sealed class JpegDecoderCore : IRawJpegData, IImageDecoderInternals
                 // to uint to avoid sign extension.
                 if (stream.RemainingBytes < (uint)markerContentByteSize)
                 {
+                    if (metadataOnly && this.Metadata != null && this.Frame != null)
+                    {
+                        // We have enough data to decode the image, so we can stop parsing.
+                        return;
+                    }
+
+                    if (this.Metadata != null && this.Frame != null && spectralConverter.HasPixelBuffer())
+                    {
+                        // We have enough data to decode the image, so we can stop parsing.
+                        return;
+                    }
+
                     JpegThrowHelper.ThrowNotEnoughBytesForMarker(fileMarker.Marker);
                 }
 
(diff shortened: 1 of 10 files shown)


src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -356,6 +356,18 @@ internal void ParseStream(BufferedReadStream stream, SpectralConverter spectralC
                 // to uint to avoid sign extension.
                 if (stream.RemainingBytes < (uint)markerContentByteSize)
                 {
+                    if (metadataOnly && this.Metadata != null && this.Frame != null)
+                    {
+                        // We have enough data to decode the image, so we can stop parsing.
+                        return;
+                    }
+
+                    if (this.Metadata != null && this.Frame != null && spectralConverter.HasPixelBuffer())
```
br3aker, line 365: As we want to now if image was decoded at this point - we can use `Converted` of `SpectralConverter` as it's called right after **baseline** image is decoded. 

Two problems:

1. It's `protected`, not really a problem creating a public getter though
2. It's not set for progressive images as those are lazy-converted during `pixel buffer -> Image<T>` step which won't be called until parsing loop is finished, not the wisest solution from me

I'd probably refactor spectral converter so it could better describe its state.

br3aker (approved):
Looks good!

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
JimBobSquarePants: Yeah. This felt a little hacky. State matching would be better.

PR #2326, opened 2023-01-25
Remove nullable disable from ColorSpace and Color

stefannikolei:
### Prerequisites

### Description
Remove nullable disable from ColorSpace and Color

### DIFF
--- a/src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.CieXyz.cs
+++ b/src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.CieXyz.cs
@@ -13,7 +13,7 @@ public partial class ColorSpaceConverter
 {
     private static readonly HunterLabToCieXyzConverter HunterLabToCieXyzConverter = new();
 
-    private LinearRgbToCieXyzConverter linearRgbToCieXyzConverter;
+    private LinearRgbToCieXyzConverter? linearRgbToCieXyzConverter;
 
     /// <summary>
     /// Converts a <see cref="CieLab"/> into a <see cref="CieXyz"/>.
--- a/src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.CieXyz.cs
+++ b/src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.CieXyz.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
 
+using System.Collections.Concurrent;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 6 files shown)


src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.CieXyz.cs
```diff
@@ -13,7 +13,7 @@ public partial class ColorSpaceConverter
 {
     private static readonly HunterLabToCieXyzConverter HunterLabToCieXyzConverter = new();
 
-    private LinearRgbToCieXyzConverter linearRgbToCieXyzConverter;
+    private LinearRgbToCieXyzConverter? linearRgbToCieXyzConverter;
```
antonfirsov, line 16: This is a very strange field, changing it's value for different values of parameters passed to `GetLinearRgbToCieXyzConverter` and it's callers. If there are shared instances, they are not thread-safe. @JimBobSquarePants thoughts? Is it some optimization?

antonfirsov (commented):
Not an expert in the area, but we should be careful.

src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverterOptions.cs
```diff
@@ -44,7 +44,7 @@ public class ColorSpaceConverterOptions
     /// <summary>
     /// Gets or sets the chromatic adaptation method used. When <value>null</value>, no adaptation will be performed.
     /// </summary>
-    public IChromaticAdaptation ChromaticAdaptation { get; set; } = new VonKriesChromaticAdaptation();
+    public IChromaticAdaptation? ChromaticAdaptation { get; set; } = new VonKriesChromaticAdaptation();
```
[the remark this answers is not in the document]
stefannikolei: Perhaps we should also add `public bool PerformChromaticAdaptation { get; set; } = true` 
Then we can make this non nullable.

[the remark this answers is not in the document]
antonfirsov: No, I think nullability is fine here, even the docs state that, which I overlooked.

src/ImageSharp/ColorSpaces/Conversion/ColorSpaceConverter.CieXyz.cs
JimBobSquarePants: ColorSpaceConverter is not threadsafe. 

Whole thing needs a rework tbh. I copied the design, and it doesn't allow easy expansion or optimization.

JimBobSquarePants: It was an optimization hack. I've replaced it with a concurrent dictionary. Should be safe now.

JimBobSquarePants (approved):

PR #2319, opened 2023-01-15
Removed #nullable disabled

stefannikolei:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description
Removed `#nullable disable` I all files where nearly the only problem was in the Equals method

#2231

PR #2327, opened 2023-01-25
Remove nullable disable from Memory.Allocators

stefannikolei:
### Prerequisites

### Description
#2231

removes nullable disable from the Memory.Allocators space

### DIFF
--- a/src/ImageSharp/Memory/Allocators/Internals/Gen2GcCallback.cs
+++ b/src/ImageSharp/Memory/Allocators/Internals/Gen2GcCallback.cs
@@ -42,7 +42,7 @@ internal sealed class Gen2GcCallback : CriticalFinalizerObject
             // Execute the callback method.
             try
             {
-                if (this.callback1 is not null && !this.callback1(targetObj))
+                if (!this.callback1!(targetObj))
                 {
                     // If the callback returns false, this callback object is no longer needed.
                     this.weakTargetObj.Free();
@@ -63,7 +63,7 @@ internal sealed class Gen2GcCallback : CriticalFinalizerObject
             // Execute the callback method.
             try
             {
-                if (this.callback0 is not null && !this.callback0())
+                if (!this.callback0!())
                 {
                     // If the callback returns false, this callback object is no longer needed.
                     return;
--- a/src/ImageSharp/Memory/Allocators/Internals/Gen2GcCallback.cs
+++ b/src/ImageSharp/Memory/Allocators/Internals/Gen2GcCallback.cs
@@ -1,6 +1,5 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
-#nullable disable
 
 // Port of BCL internal utility:
 // https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Private.CoreLib/src/System/Gen2GcCallback.cs
@@ -15,8 +14,8 @@ namespace SixLabors.ImageSharp.Memory.Internals;
 /// </summary>
 internal sealed class Gen2GcCallback : CriticalFinalizerObject
 {
-    private readonly Func<bool> callback0;
-    private readonly Func<object, bool> callback1;
+    private readonly Func<bool>? callback0;
+    private readonly Func<object, bool>? callback1;
     private GCHandle weakTargetObj;
 
     private Gen2GcCallback(Func<bool> callback) => this.callback0 = callback;
@@ -32,7 +31,7 @@ internal sealed class Gen2GcCallback : CriticalFinalizerObject
         if (this.weakTargetObj.IsAllocated)
         {
             // Check to see if the target object is still alive.
-            object targetObj = this.weakTargetObj.Target;
+            object? targetObj = this.weakTargetObj.Target;
             if (targetObj == null)
             {
                 // The target object is dead, so this callback object is no longer needed.
@@ -43,7 +42,7 @@ internal sealed class Gen2GcCallback : CriticalFinalizerObject
             // Execute the callback method.
             try
             {
-                if (!this.callback1(targetObj))
+                if (this.callback1 is not null && !this.callback1(targetObj))
                 {
                     // If the callback returns false, this callback object is no longer needed.
                     this.weakTargetObj.Free();
@@ -64,7 +63,7 @@ internal sealed class Gen2GcCallback : CriticalFinalizerObject
             // Execute the callback method.
             try
             {
-                if (!this.callback0())
+                if (this.callback0 is not null && !this.callback0())
                 {
                     // If the callback returns false, this callback object is no longer needed.
                     return;
--- a/src/ImageSharp/Memory/Allocators/Internals/ManagedBufferBase.cs
+++ b/src/ImageSharp/Memory/Allocators/Internals/ManagedBufferBase.cs
@@ -43,5 +43,5 @@ internal abstract class ManagedBufferBase<T> : MemoryManager<T>
     /// Gets the object that should be pinned.
     /// </summary>
     /// <returns>The pinnable <see cref="object"/>.</returns>
-    protected abstract object GetPinnableObject();
+    protected abstract object? GetPinnableObject();
 }
--- a/src/ImageSharp/Memory/Allocators/Internals/ManagedBufferBase.cs
+++ b/src/ImageSharp/Memory/Allocators/Internals/ManagedBufferBase.cs
@@ -43,5 +43,5 @@ internal abstract class ManagedBufferBase<T> : MemoryManager<T>
     /// Gets the object that should be pinned.
     /// </summary>
     /// <returns>The pinnable <see cref="object"/>.</returns>
-    protected abstract object? GetPinnableObject();
+    protected abstract object GetPinnableObject();
 }
--- a/src/ImageSharp/Memory/Allocators/Internals/SharedArrayPoolBuffer{T}.cs
+++ b/src/ImageSharp/Memory/Allocators/Internals/SharedArrayPoolBuffer{T}.cs
@@ -3,6 +3,7 @@
 
 using System.Buffers;
 using System.Diagnostics;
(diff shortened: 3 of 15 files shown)


src/ImageSharp/Memory/Allocators/Internals/Gen2GcCallback.cs
```diff
@@ -43,7 +42,7 @@ private Gen2GcCallback(Func<object, bool> callback, object targetObj)
             // Execute the callback method.
             try
             {
-                if (!this.callback1(targetObj))
+                if (this.callback1 is not null && !this.callback1(targetObj))
```
antonfirsov, line 45: `callback1`should never be null here. The type was ported from [here](https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/shared/System/Gen2GcCallback.cs). We don't use `Debug.Assert` in our codebase, so this should be probably

```suggestion
                if (!this.callback1!(targetObj))
```

src/ImageSharp/Memory/Allocators/Internals/Gen2GcCallback.cs
```diff
@@ -64,7 +63,7 @@ private Gen2GcCallback(Func<object, bool> callback, object targetObj)
             // Execute the callback method.
             try
             {
-                if (!this.callback0())
+                if (this.callback0 is not null && !this.callback0())
```
antonfirsov, line 66:
```suggestion
                if (!this.callback0!())
```

src/ImageSharp/Memory/Allocators/Internals/ManagedBufferBase.cs
```diff
@@ -43,5 +43,5 @@ public override void Unpin()
     /// Gets the object that should be pinned.
     /// </summary>
     /// <returns>The pinnable <see cref="object"/>.</returns>
-    protected abstract object GetPinnableObject();
+    protected abstract object? GetPinnableObject();
```
antonfirsov, line 46: This should rather throw when there is no pinnable object present.

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroup{T}.cs
```diff
@@ -41,7 +41,7 @@ private MemoryGroup(int bufferLength, long totalLength)
     /// <inheritdoc />
     public bool IsValid { get; private set; } = true;
 
-    public MemoryGroupView<T> View { get; private set; }
+    public MemoryGroupView<T>? View { get; private set; }
```
antonfirsov, line 44: This should not be nullable. In what states is it `null`?

src/ImageSharp/Memory/Buffer2D{T}.cs
```diff
@@ -43,7 +43,7 @@ internal Buffer2D(MemoryGroup<T> memoryGroup, int width, int height)
     /// Gets the backing <see cref="IMemoryGroup{T}"/>.
     /// </summary>
     /// <returns>The MemoryGroup.</returns>
-    public IMemoryGroup<T> MemoryGroup => this.FastMemoryGroup.View;
+    public IMemoryGroup<T>? MemoryGroup => this.FastMemoryGroup.View;
```
antonfirsov, line 46: This should not be nullable.

src/ImageSharp/Memory/Allocators/Internals/SharedArrayPoolBuffer{T}.cs
```diff
@@ -22,7 +21,7 @@ public SharedArrayPoolBuffer(int lengthInElements)
         this.lifetimeGuard = new LifetimeGuard(this.Array);
     }
 
-    public byte[] Array { get; private set; }
+    public byte[]? Array { get; private set; }
```
antonfirsov, line 24: What are uses of this property outside of the class? Can make it private instead of making it nullable?

antonfirsov (changes requested):
We should be much more strict making things nullable, especially around public properties. Omitting null checks can lead to bugs. I wonder if there were similar functional changes in #2319 and #2319?

src/ImageSharp/Memory/Allocators/Internals/Gen2GcCallback.cs
stefannikolei: callback1 or callback0 can be null. It depends on the used constructor. It will just be a silent exception. I can replace it with the !

src/ImageSharp/Memory/Allocators/Internals/SharedArrayPoolBuffer{T}.cs
stefannikolei: It is used in MemoryGroupSpanCache:30

I can make it non nullable. But then I have to remove the null check in the dispose and the explicit setting it to null in the dispose method

src/ImageSharp/Memory/Allocators/Internals/Gen2GcCallback.cs
antonfirsov: The point is that if you add stuff like `this.callback1 is not null`, or turn `this.someField.Call()` to `this.someField?.Call() there no error will occur if a bug is introduced to the code later. We should be very careful with such changes, better not to make them unless it's 100% safe and future-proof.

src/ImageSharp/Memory/Allocators/Internals/SharedArrayPoolBuffer{T}.cs
antonfirsov: In that case `byte[]? Array` seems ok.

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroup{T}.cs
stefannikolei: I can dismiss this compiler error with null!; All the inheriting classes make sure in the constructor that View is set

src/ImageSharp/Memory/Allocators/Internals/ManagedBufferBase.cs
stefannikolei: What exception do you want me to throw?

antonfirsov: Depends on the states when the underlying field is null. If it's because of disposal, it should be `ObjectDisposedException`.

src/ImageSharp/Memory/Allocators/Internals/SharedArrayPoolBuffer{T}.cs
```diff
@@ -74,7 +79,6 @@ protected override void Release()
             // meaning likely a different bucket than it was rented from,
             // but this is PROBABLY better than not returning the arrays at all.
             ArrayPool<byte>.Shared.Return(this.array);
-            this.array = null;
```
antonfirsov, line 77: It's better to make `this.array` nullable so it's detached after a call to `Release`.

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroup{T}.Consumed.cs
```diff
@@ -48,6 +48,6 @@ IEnumerator<Memory<T>> IEnumerable<Memory<T>>.GetEnumerator()
             return ((IList<Memory<T>>)this.source).GetEnumerator();
         }
 
-        public override void Dispose() => this.View.Invalidate();
+        public override void Dispose() => this.View?.Invalidate();
```
antonfirsov, line 51: Now that `View` is non-nullable there is no need for `?`. Same the for changes below.

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroup{T}.Owned.cs
```diff
@@ -123,7 +123,7 @@ public override void DecreaseRefCounts()
 
         public override void RecreateViewAfterSwap()
         {
-            this.View.Invalidate();
+            this.View?.Invalidate();
```
antonfirsov, line 126:
```suggestion
            this.View.Invalidate();
```

src/ImageSharp/Memory/DiscontiguousBuffers/MemoryGroup{T}.Owned.cs
```diff
@@ -141,15 +141,15 @@ public override void Dispose()
                 return;
             }
 
-            this.View.Invalidate();
+            this.View?.Invalidate();
```
antonfirsov, line 144:
```suggestion
            this.View.Invalidate();
```

src/ImageSharp/Memory/Buffer2D{T}.cs
```diff
@@ -138,7 +138,7 @@ internal Memory<T> GetSafeRowMemory(int y)
     {
         DebugGuard.MustBeGreaterThanOrEqualTo(y, 0, nameof(y));
         DebugGuard.MustBeLessThan(y, this.Height, nameof(y));
-        return this.FastMemoryGroup.View.GetBoundedMemorySlice(y * (long)this.Width, this.Width);
+        return this.FastMemoryGroup.View!.GetBoundedMemorySlice(y * (long)this.Width, this.Width);
```
antonfirsov, line 141:
```suggestion
        return this.FastMemoryGroup.View.GetBoundedMemorySlice(y * (long)this.Width, this.Width);
```

src/ImageSharp/Memory/Buffer2DExtensions.cs
```diff
@@ -22,7 +22,7 @@ public static IMemoryGroup<T> GetMemoryGroup<T>(this Buffer2D<T> buffer)
         where T : struct
     {
         Guard.NotNull(buffer, nameof(buffer));
-        return buffer.FastMemoryGroup.View;
+        return buffer.FastMemoryGroup.View!;
```
antonfirsov, line 25:
```suggestion
        return buffer.FastMemoryGroup.View;
```

antonfirsov (approved):
Good job, thanks!

PR #2338, opened 2023-01-31
Remove nullable disable from Processing.Processors

stefannikolei:
### Prerequisites

### Description
#2231

Here I went the easy way and introduced some bangs!.

It is not easy to not use it, for example using the visitor pattern, I can not tell the compiler that the value is not null so I had to use the !

### DIFF
--- a/src/ImageSharp/Processing/Processors/Quantization/OctreeQuantizer{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Quantization/OctreeQuantizer{TPixel}.cs
@@ -26,7 +26,7 @@ public struct OctreeQuantizer<TPixel> : IQuantizer<TPixel>
     private readonly int maxColors;
     private readonly int bitDepth;
     private readonly Octree octree;
-    private IMemoryOwner<TPixel>? paletteOwner;
+    private IMemoryOwner<TPixel> paletteOwner;
     private ReadOnlyMemory<TPixel> palette;
     private EuclideanPixelMap<TPixel>? pixelMap;
     private readonly bool isDithering;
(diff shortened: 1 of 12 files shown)


src/ImageSharp/Processing/Processors/Quantization/OctreeQuantizer{TPixel}.cs
```diff
@@ -27,9 +26,9 @@ public struct OctreeQuantizer<TPixel> : IQuantizer<TPixel>
     private readonly int maxColors;
     private readonly int bitDepth;
     private readonly Octree octree;
-    private IMemoryOwner<TPixel> paletteOwner;
+    private IMemoryOwner<TPixel>? paletteOwner;
```
antonfirsov, line 29: Uninitialized instances of this struct are invalid.

```suggestion
    private IMemoryOwner<TPixel> paletteOwner;
```

Rendering `!` unnecessary in all `this.paletteOwner!` expressions below.

src/ImageSharp/Processing/Processors/Quantization/PaletteQuantizer{TPixel}.cs
```diff
@@ -21,7 +20,7 @@ namespace SixLabors.ImageSharp.Processing.Processors.Quantization;
 internal struct PaletteQuantizer<TPixel> : IQuantizer<TPixel>
     where TPixel : unmanaged, IPixel<TPixel>
 {
-    private EuclideanPixelMap<TPixel> pixelMap;
+    private EuclideanPixelMap<TPixel>? pixelMap;
```
antonfirsov, line 23:
```suggestion
    private EuclideanPixelMap<TPixel> pixelMap;
```

antonfirsov (commented):
Mostly good, with a few suggestions.

src/ImageSharp/Processing/Processors/Quantization/OctreeQuantizer{TPixel}.cs
stefannikolei: Ok then I have to remove the setting of null in dispose

JimBobSquarePants (approved):
Good job 👍

Some horrible internal code here that will need revisiting in the future.

PR #2339, opened 2023-02-01
Added L16 support to tiff encoder

ptasev:
### Prerequisites

### Description
Added L16 support to tiff encoder ([source discussion](https://github.com/SixLabors/ImageSharp/discussions/2333)).

Breaking Changes:
- When encoder BitsPerPixel is set to Bit16, the image will be encoded as L16 now instead of Rgb24
- When only PhotometricInterpretation is set to BlackIsZero or WhiteIsZero and source image pixel type has Bits/Px of 16 then the image will be encoded as L16 instead of L8
- When neither is set and source image pixel type has Bits/Px of 16 then the image will be encoded as L16 instead of Rgb24

### DIFF
--- a/src/ImageSharp/Formats/Tiff/Compression/HorizontalPredictor.cs
+++ b/src/ImageSharp/Formats/Tiff/Compression/HorizontalPredictor.cs
@@ -67,6 +67,11 @@ internal static class HorizontalPredictor
         {
             ApplyHorizontalPrediction8Bit(rows, width);
         }
+        else if (bitsPerPixel == 16)
+        {
+            // Assume rows are L16 grayscale since that's currently the only way 16 bits is supported by encoder
+            ApplyHorizontalPrediction16Bit(rows, width);
+        }
         else if (bitsPerPixel == 24)
         {
             ApplyHorizontalPrediction24Bit(rows, width);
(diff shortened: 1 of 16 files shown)


tests/ImageSharp.Tests/Formats/Tiff/TiffEncoderTests.cs
```diff
@@ -473,6 +513,7 @@ public void TiffEncoder_EncodeBiColor_WithModifiedHuffmanCompression_BlackIsZero
 
     [Theory]
     [WithFile(GrayscaleUncompressed, PixelTypes.L8, TiffPhotometricInterpretation.BlackIsZero, TiffCompression.PackBits)]
+    [WithFile(GrayscaleUncompressed, PixelTypes.L16, TiffPhotometricInterpretation.BlackIsZero, TiffCompression.PackBits)]
```
ptasev, line 516: I need to generate L16 versions of `GrayscaleUncompressed` and `Calliphora_GrayscaleUncompressed`. What's the recommended way to do so?

JimBobSquarePants: @brianpopow You generated the tiff files didn't you?

brianpopow: I usually use  `tiffcp` which comes with `libtiff` to convert tiff, but I dont think you can change the bit per pixel in the output. It will have the same as the input (maybe I am missing an option).
```
usage: tiffcp [options] input... output
```

If `tiffcp` does not work, I use `imagemagick`. You can install `imagemagick` via [chocolatey](https://chocolatey.org)
The command would be:
```
magick convert Calliphora_grayscale_uncompressed.tiff -depth 16 Calliphora_grayscale_uncompressed_16bit.tiff
```

To convert with compression, use the `-compress` flag.

To make sure the image is as expected, I again use a tool from `libtiff` called `tiffinfo`. It shows a nice summary of the tiff.

ptasev: Thanks! I was able to use `magick` for depth, and `tiffcp` for compression.

src/ImageSharp/Formats/Tiff/Compression/HorizontalPredictor.cs
```diff
@@ -67,6 +67,11 @@ public static void ApplyHorizontalPrediction(Span<byte> rows, int width, int bit
         {
             ApplyHorizontalPrediction8Bit(rows, width);
         }
+        else if (bitsPerPixel == 16)
+        {
+            // Assume rows are L16 grayscale since that's currently the only way 16 bits is supported by encoder
```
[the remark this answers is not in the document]
ptasev: Can you clarify how you'd like to support color? Do you mean support encoding color as 16 bits per pixel, or when the user tells the encoder to use 16 bits per pixel, we should determine somehow whether to use Rgb24 or L16?

Before this change, if the user set bits per pixel in the encoder to 16, the encoder would save as Rgb24. So color at 16 bits per pixel already isn't supported. We're changing the assumption now that the user meant L16 instead of Rgb24 when using the encoder setting of 16 bits per pixel. The new implementation would only become a breaking change if the user is explicitly setting this option in the encoder which didn't do exactly as they wanted anyway.

Supporting encoding color as 16 bits per pixel would become a bigger design/architecture challenge since we'll need to figure out which pixel format the user wants to use specifically (4444, 565, 5551, La16).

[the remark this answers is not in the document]
JimBobSquarePants: Sorry, I misread. I was thinking 16bits per component like png. 

I would assume that 16 bits per pixel can only mean luminance as you have.

JimBobSquarePants (approved):
This all looks great. Thanks! :+1:

PR #2343, opened 2023-02-05
Restrict trigger for ARM to PRs containing labels only 

JimBobSquarePants:
### Prerequisites

### Description

Trying to keep the ARM build runner checks to a minimum to reduce costs.

Following a combination of the following instructions

https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label#comment122159108_62331521

https://github.com/orgs/community/discussions/26253#discussioncomment-3250989


JimBobSquarePants:
First run with condition excludes ARM matrix entries. 

Now adding labels to ensure they are captured.

JimBobSquarePants:
@brianpopow It works!!

.github/workflows/build-and-test.yml
```diff
@@ -58,6 +61,10 @@ jobs:
             sdk: 6.0.x
             runtime: -x64
             codecov: false
+        exclude:
+          - isARM: false
```
brianpopow, line 65: @JimBobSquarePants: I am not sure that I understand the intention of the `exclude` here, does that mean only the arm runners should be executed and all others should not run?

JimBobSquarePants: The opposite. Exclude the ARM runners unless the label exists. 

I thought best to run the other runners at all times which helps us avoid any errors we potentially introduce with conditionals.

PR #2346, opened 2023-02-05
remove nullable disable from format gif

stefannikolei:
### Prerequisites

### Description
#2231

### DIFF
--- a/src/ImageSharp/Formats/Gif/GifEncoderCore.cs
+++ b/src/ImageSharp/Formats/Gif/GifEncoderCore.cs
@@ -2,7 +2,6 @@
 // Licensed under the Six Labors Split License.
 
 using System.Buffers;
-using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using SixLabors.ImageSharp.Advanced;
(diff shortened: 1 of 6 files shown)


src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -141,7 +141,7 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream, CancellationToken
     private void EncodeFrames<TPixel>(
         Stream stream,
         Image<TPixel> image,
-        IndexedImageFrame<TPixel> quantized,
+        IndexedImageFrame<TPixel>? quantized,
```
JimBobSquarePants, line 144: I haven't pulled this down to check locally but I think quantized is never null when calling this method. We set it to null but I don't think we need to.

stefannikolei: I also thought that. I changed it and the tests failed.

It can be null because the method is calling itself. The problem is, that quantized is passed from the outside for the first frame. For the other frames it is created.

This could be probably get a refactoring but probably way out of scope of this change

JimBobSquarePants: `EncodeFrames` calls `EncodeFrame` - Not the most penetrable code I know but I actually refactored it to make it more readable recently! 🤣

I managed to tweak it to work without nullifying.

JimBobSquarePants (approved):

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
stefannikolei: I thought I tried the exact same thing but got test failures afterwards. But could also been in another file. Not so sure anymore touched too much lately to remember^^

JimBobSquarePants: I've been thoroughly impressed with how well you've navigated the codebase. There's a LOT to take in!

PR #2348, opened 2023-02-05
Remove nullable disable from format tga

stefannikolei:
### Prerequisites

### Description
#2330

### DIFF
--- a/src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
+++ b/src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
@@ -259,7 +259,7 @@ internal sealed class TgaDecoderCore : IImageDecoderInternals
                     {
                         for (int x = width - 1; x >= 0; x--)
                         {
-                            this.ReadPalettedBgr24Pixel(stream, palette, colorMapPixelSizeInBytes, x, color, pixelRow);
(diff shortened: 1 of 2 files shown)


src/ImageSharp/Formats/Tga/TgaDecoderCore.cs
```diff
@@ -29,12 +29,12 @@ internal sealed class TgaDecoderCore : IImageDecoderInternals
     /// <summary>
     /// The metadata.
     /// </summary>
-    private ImageMetadata metadata;
+    private ImageMetadata metadata = null!;
```
JimBobSquarePants, line 32: Whoa! I didn't know this was possible.

JimBobSquarePants: Is it a good idea? We haven't used this strategy elsewhere.

stefannikolei: It is an easy way. In the decoder and encoder this is not null while processing because it is initialized in the public methods. 

To make it more future proof it is probably better to not do that. 

But then we need Guards in many places or some !

JimBobSquarePants: I think we should probably go the future proof way and be consistent with the other formats

JimBobSquarePants (approved):
Awesome.

PR #2352, opened 2023-02-08
Remove nullable disable from ChunkedMemoryStream

stefannikolei:
### Prerequisites

### Description
#2231

### DIFF
--- a/src/ImageSharp/IO/ChunkedMemoryStream.cs
+++ b/src/ImageSharp/IO/ChunkedMemoryStream.cs
@@ -47,8 +47,6 @@ internal sealed class ChunkedMemoryStream : Stream
     /// <param name="allocator">The memory allocator.</param>
     public ChunkedMemoryStream(MemoryAllocator allocator)
     {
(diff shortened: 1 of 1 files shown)


src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -555,11 +557,13 @@ private sealed class MemoryChunk : IDisposable
     {
         private bool isDisposed;
 
-        public IMemoryOwner<byte> Buffer { get; set; }
+        public MemoryChunk(IMemoryOwner<byte> buffer) => this.Buffer = buffer;
```
stefannikolei, line 560: At first I did not want to create a constructor for this because there was none..

I wanted to use the required keyword, but I could not use it because it is a C#11 feature.

That's why I went with the constructor

JimBobSquarePants: That's all cool. I had to do the same with `TextOptions` in the Fonts library.

JimBobSquarePants (approved):

PR #2353, opened 2023-02-09
Remove nullable disable from various files

stefannikolei:
### Prerequisites

### Description
#2231


src/ImageSharp/Common/Helpers/Guard.cs
```diff
@@ -17,13 +17,14 @@ internal static partial class Guard
     /// <typeparam name="TValue">The type of the value.</typeparam>
     /// <exception cref="ArgumentException"><paramref name="value"/> is not a value type.</exception>
     [MethodImpl(InliningOptions.ShortMethod)]
-    public static void MustBeValueType<TValue>(TValue value, string parameterName)
+    public static void SamplerMustBeValueType<TValue>(TValue value, [CallerArgumentExpression("value")] string? parameterName = null)
```
stefannikolei, line 20: I went in Renaming this method because it was just used for samplers to make sure that the implementation is a struct and not a class.

So I added IResampler as a Type constraint.

src/ImageSharp/Processing/Extensions/ProcessingExtensions.cs
```diff
@@ -150,14 +149,14 @@ public static Image Clone(this Image source, Action<IImageProcessingContext> ope
...
-    public static Image Clone(this Image source, Configuration configuration, Action<IImageProcessingContext> operation)
+    public static Image? Clone(this Image source, Configuration configuration, Action<IImageProcessingContext> operation)
     {
         Guard.NotNull(configuration, nameof(configuration));
         Guard.NotNull(source, nameof(source));
         Guard.NotNull(operation, nameof(operation));
         source.EnsureNotDisposed();
 
-        var visitor = new ProcessingVisitor(configuration, operation, false);
+        ProcessingVisitor visitor = new(configuration, operation, false);
```
stefannikolei, line 159: I could have also just changed it to visitor.ResultImage!

Can we always be sure that the visitor has a ResultImage after AcceptVisitor?

src/ImageSharp/ImageFrame{TPixel}.cs
```diff
@@ -342,8 +341,7 @@ protected override void Dispose(bool disposing)
 
         if (disposing)
         {
-            this.PixelBuffer?.Dispose();
-            this.PixelBuffer = null;
```
stefannikolei, line 346: There's a test to check if it is null.

Is that check mandatory?

src/ImageSharp/Common/Helpers/Guard.cs
JimBobSquarePants: Guard types are generally meant to be more general in nature. It's only used by the sampler now but could be used for other things in the future.

src/ImageSharp/Processing/Extensions/ProcessingExtensions.cs
JimBobSquarePants: Let me check

src/ImageSharp/ImageFrame{TPixel}.cs
JimBobSquarePants: Let me check

src/ImageSharp/Common/Helpers/Guard.cs
stefannikolei: I found a good solution ^^

src/ImageSharp/ImageFrame{TPixel}.cs
JimBobSquarePants: No. the pixel buffer is always assigned.

src/ImageSharp/Common/Helpers/Guard.cs
```diff
@@ -17,13 +16,14 @@ internal static partial class Guard
     /// <typeparam name="TValue">The type of the value.</typeparam>
     /// <exception cref="ArgumentException"><paramref name="value"/> is not a value type.</exception>
     [MethodImpl(InliningOptions.ShortMethod)]
-    public static void MustBeValueType<TValue>(TValue value, string parameterName)
+    public static void MustBeValueType<TValue>(TValue value, [CallerArgumentExpression("value")] string? parameterName = null)
```
JimBobSquarePants, line 19: This is clever!

src/ImageSharp/Processing/Extensions/ProcessingExtensions.cs
JimBobSquarePants: I changed the property to a method to make the temporal coupling more obvious. The visitor should always have a result image though.

JimBobSquarePants (approved):
Looks great. Thanks!

PR #2355, opened 2023-02-11
Fix mode score calculation for SSE2/AVX2 version

brianpopow:
### Prerequisites

### Description

This PR fixes an issue with the mode score calculation: The `SubtractSaturate` should be `_mm_subs_epi16` not `_mm_subs_epu8`


JimBobSquarePants (commented):
We’ll spotted! What affect does that have on encoding?

brianpopow:
> We’ll spotted! What affect does that have on encoding?

The encoder will still produce a valid webp image, if the score is calculated wrong, but it may be not as small as it could be when the score is calculated correctly.

JimBobSquarePants:
Thought so. Thanks!

JimBobSquarePants (approved):

PR #2357, opened 2023-02-16
Allow setting configuration property

JimBobSquarePants:
### Prerequisites

### Description

The property was internal when it shouldn't be.

### DIFF
--- a/src/ImageSharp/Formats/DecoderOptions.cs
+++ b/src/ImageSharp/Formats/DecoderOptions.cs
@@ -15,6 +15,11 @@ public sealed class DecoderOptions
 
     private uint maxFrames = int.MaxValue;
 
+    // Used by the FileProvider in the unit tests to set the configuration on the fly.
+#pragma warning disable SA1401 // Fields should be private
+    internal Configuration BackingConfiguration = Configuration.Default;
+#pragma warning restore SA1401 // Fields should be private
+
     /// <summary>
     /// Gets the shared default general decoder options instance.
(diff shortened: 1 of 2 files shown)


src/ImageSharp/Formats/DecoderOptions.cs
```diff
@@ -15,15 +15,21 @@ public sealed class DecoderOptions
 
     private uint maxFrames = int.MaxValue;
 
+    // Used by the FileProvider in the unit tests to set the configuration on the fly.
```
JimBobSquarePants, line 18: Yeah. I know. This is nasty but so is the file provider.

src/ImageSharp/Formats/DecoderOptions.cs
```diff
@@ -15,15 +15,21 @@ public sealed class DecoderOptions
...
+#pragma warning disable SA1401 // Fields should be private
+    internal Configuration BackingConfiguration = Configuration.Default;
+#pragma warning restore SA1401 // Fields should be private
```
antonfirsov, line 21: Why not a member `SetConfiguration()` method and a private field?

JimBobSquarePants: Yeah... I was trying to avoid adding to the method itself and overcomplicated things. I moved it but still had to suppress a warning as the IDE thinks incorrectly that it can be an auto-implemented property.

antonfirsov (approved):
I would expect these analyzers to be smarter 😞

PR #2359, opened 2023-02-17
Enable Avx2 optimizations on Porter-Duff operations.

### LINKED ISSUE #2340: Enable Avx2 optimizations on Porter-Duff operations.
### Discussed in https://github.com/SixLabors/ImageSharp/discussions/2328

<sup>Originally posted by **JimBobSquarePants** January 26, 2023</sup>
Given that #1433 is not achievable within a reasonable timescale for v3 and might well be better implemented in a pattern using [static virtual members in interfaces](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/static-virtual-interface-members) I'd like to propose that we enhance the existing `Vector4` based Porter-Duff implementations with `Vector256<float>` equivalents. 

This will give us a 2X performance improvement on supported devices over the existing implementation which we badly need if we are ever able to release a v1 of ImageSharp.Drawing.

We should be able to implement each separable method in the same fashion as the existing methods. The pixel blender implementations can call these. 

For example. The existing `Over` method...

```c#
public static Vector4 Over(Vector4 destination, Vector4 source, Vector4 blend)
{
    // calculate weights
    float blendW = destination.W * source.W;
    float dstW = destination.W - blendW;
    float srcW = source.W - blendW;

    // calculate final alpha
    float alpha = dstW + source.W;

    // calculate final color
    Vector4 color = (destination * dstW) + (source * srcW) + (blend * blendW);

    // unpremultiply
    color /= MathF.Max(alpha, 0.001F);
    color.W = alpha;

    return color;
}
```

...Could be implemented in the following manner.

```c#
public static Vector256<float> Over(Vector256<float> destination, Vector256<float> source, Vector256<float> blend)
{
    const int BlendAlphaControl = 0b_10_00_10_00;
    const int ShuffleAlphaControl = 0b_11_11_11_11;

    // calculate weights
    var sW = Avx.Shuffle(source, source, ShuffleAlphaControl);
    var dstW = Avx.Shuffle(destination, destination, ShuffleAlphaControl);
    var blendW = Avx.Multiply(sW, dstW);

    dstW = Avx.Subtract(dstW, blendW);
    var srcW = Avx.Subtract(sW, blendW);

    // calculate final alpha
    Vector256<float> alpha = Avx2.Add(dstW, sW);

    // calculate final color
    Vector256<float> color = Avx2.Add(Avx2.Add(Avx2.Multiply(destination, dstW), Avx2.Multiply(source, srcW)), Avx2.Multiply(blend, blendW));

    // unpremultiply
    Vector256<float> alphaEpsilon = Avx2.Max(alpha, Vector256.Create(0.001f));
    color = Avx2.Divide(color, alphaEpsilon);
    return Avx.Blend(color, alpha, BlendAlphaControl);
}
```

For someone with a good grasp of the SIMD APIs this should not be a huge undertaking.

Thoughts @antonfirsov @br3aker ??</div>

JimBobSquarePants:
### Prerequisites

### Description

Adds `Vector256<float>` implementations of the Porter-Duff functions to speed up our blending operations. Fixes #2340 

- [x] Complete function component set within `PorterDuffFunctions.cs`.
- [x] Update the `PorterDuffFunctions.Generated.tt` to include the `Vector256<float>` variants.
- [x] Update `DefaultPixelBlenders.Generated.tt` to use the `Vector256<float>` variants when appropriate.
- [x] Update the unit tests to use `RemoteExecutor`

@antonfirsov @tannergooding @saucecontrol @br3aker I'll take any help you can offer pointing me in the right direction to finish porting these over.


src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -98,10 +151,17 @@ public static Vector4 Darken(Vector4 backdrop, Vector4 source)
...
+
+    /// <summary>
+    /// Returns the result of the "Lighten" compositing equation.
+    /// </summary>
+    /// <param name="backdrop">The backdrop vector.</param>
+    /// <param name="source">The source vector.</param>
+    /// <returns>The <see cref="Vector256{Single}"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector256<float> Lighten(Vector256<float> backdrop, Vector256<float> source)
+        => Avx.Max(backdrop, source);
```
JimBobSquarePants, line 164: I don't know how to do `OverlayValueFunction` below.

br3aker: I'm a little rusty but something like this should work:

```cs
private static readonly Vector256<float> Half = Vector256.Create(0.5f);
private static readonly Vector256<float> One = Vector256.Create(0.5f);
private static readonly Vector256<float> Two = Vector256.Create(0.5f);

public static Vector256<float> OverlayValueFunction(Vector256<float> backdrop, Vector256<float> source)
{
    var left = Avx.Multiply(Avx.Multiply(Two, backdrop), source);
    var right = Avx.Subtract(One, Avx.Multiply(Avx.Multiply(Two, Avx.Subtract(One, source)), Avx.Subtract(One, backdrop)));

    var cmp = Avx.CompareGreaterThan(backdrop, Half);
    return Avx.BlendVariable(left, right, cmp);
}
```

SIMD implementation must calculate both cases but it's still a bit faster than scalar:

```
BenchmarkDotNet=v0.13.4, OS=Windows 11 (10.0.22621.1265)
Unknown processor
.NET SDK=7.0.100
  [Host]     : .NET 6.0.12 (6.0.1222.56807), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.12 (6.0.1222.56807), X64 RyuJIT AVX2


| Method |     Mean |    Error |   StdDev |
|------- |---------:|---------:|---------:|
| Scalar | 92.27 ns | 0.313 ns | 0.277 ns |
|    Avx | 35.93 ns | 0.356 ns | 0.333 ns |
```

JimBobSquarePants: Ah that's awesome. Thanks! That lets me complete the first objective.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -21,6 +21,12 @@ namespace SixLabors.ImageSharp.PixelFormats.PixelBlenders;
 /// </remarks>
 internal static partial class PorterDuffFunctions
 {
+    private const int BlendAlphaControl = 0b_10_00_10_00;
+    private const int ShuffleAlphaControl = 0b_11_11_11_11;
+    private static readonly Vector256<float> Vector256Half = Vector256.Create(0.5F);
+    private static readonly Vector256<float> Vector256One = Vector256.Create(1F);
+    private static readonly Vector256<float> Vector256Two = Vector256.Create(2F);
```
tannergooding, line 28: This is going to be worse than direct usage for .NET 5+

Starting in .NET 5 there was some special support added for `Vector###.Create(cns)` and `Vector###.Create(cns, ..., cns)`. These APIs will now generate a "method local constant" which makes it quite a bit more efficient than even a `static readonly` will be.

In .NET 5 this support only existed in the late phases of the JIT (lowering). That support was improved a bit more in .NET 6, 7, and now in 8 as well. Starting in 7 in particular we now have a direct node (`GT_CNS_VEC`) which allows other phases of the JIT to take advantage of this. In .NET 8 we've started adding constant folding support as well.

tannergooding: If you want to centralize such "constants", I'd recommend returning them from a property instead:
```csharp
private static Vector256<float> Vector256Half => Vector256.Create(0.5F);
private static Vector256<float> Vector256One => Vector256.Create(1F);
private static Vector256<float> Vector256Two => Vector256.Create(2F);
```

JimBobSquarePants: Oh that's very interesting. Will revert and look into other examples in the codebase.

What about multiple references to a local variable within a single method? Should I inline them?

```c#
public static Vector256<float> Screen(Vector256<float> backdrop, Vector256<float> source)
{
    Vector256<float> vOne = Vector256.Create(1F);
    return Avx.Subtract(vOne, Avx.Multiply(Avx.Subtract(vOne, backdrop), Avx.Subtract(vOne, source)));
}
```

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -179,6 +182,19 @@ public static Vector4 Overlay(Vector4 backdrop, Vector4 source)
...
+    /// Returns the result of the "Overlay" compositing equation.
+    /// </summary>
+    /// <param name="backdrop">The backdrop vector.</param>
+    /// <param name="source">The source vector.</param>
+    /// <returns>The <see cref="Vector4"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector256<float> Overlay(Vector256<float> backdrop, Vector256<float> source)
+    {
+        Vector256<float> color = OverlayValueFunction(backdrop, source);
+        return Avx.Min(Vector256One, Avx.Blend(color, Vector256<float>.Zero, BlendAlphaControl));
```
tannergooding, line 195: We're "missing" a JIT optimization here so swapping the parameter order for `vblendps` can be more efficient

Namely do `Avx.Blend(Vector256<float>.Zero, color, 0b_01_11_01_11);`. This is special to `Zero` and `AllBitsSet` since they aren't "normal" constants but are instead always generated into register. For other constants the original ordering you had would've been better.

tannergooding: I'll log a JIT issue to track ensuring we support commutativity for `Blend`

JimBobSquarePants: Awesome. Thanks!

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -205,6 +234,22 @@ public static Vector4 HardLight(Vector4 backdrop, Vector4 source)
...
+    /// <summary>
+    /// Helper function for Overlay and HardLight modes
+    /// </summary>
+    /// <param name="backdrop">Backdrop color element</param>
+    /// <param name="source">Source color element</param>
+    /// <returns>Overlay value</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector256<float> OverlayValueFunction(Vector256<float> backdrop, Vector256<float> source)
+    {
+        Vector256<float> left = Avx.Multiply(Avx.Multiply(Vector256Two, backdrop), source);
```
tannergooding, line 246: `Multiply(Two, backdrop)` is better as `Add(backdrop, backdrop)`. 

This is something I'm working on recognizing implicitly in the JIT as part of the constant folding support being added in .NET 8

tannergooding: You could then consider a helper method here:
```csharp
public static Vector256<float> MultiplyAddEstimate(Vector256<float> left, Vector256<float> right, Vector256<float> addend)
{
    if (Fma.IsSupported)
    {
        return Fma.MultiplyAdd(left, right, addend);
    }
    else
    {
        return Avx.Add(Avx.Multiply(left, right), addend);
    }
}
```

The reason that it's an "estimate" is that `FMA` does a more precise computation and gives a more accurate result. This can also subtly change the answer in some edge cases.

The most notably is that for `double`, `(1e308 * 2) - 1e308` returns `infinity`, but `FusedMultiplyAdd(1e308, 2, -1e308)` returns `1e300` (there is a similar scenario for `float`, but I don't remember the exact inputs off the top of my head).

Most other inputs are less drastic in their differences. They'll either produce the same result (with `Fma` typically being faster) or be off by 1 ULP (if you looked at the raw bits, the least significant bit would typically differ by 1).

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -243,12 +288,9 @@ public static Vector4 Over(Vector4 destination, Vector4 source, Vector4 blend)
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static Vector256<float> Over(Vector256<float> destination, Vector256<float> source, Vector256<float> blend)
     {
-        const int blendAlphaControl = 0b_10_00_10_00;
-        const int shuffleAlphaControl = 0b_11_11_11_11;
-
         // calculate weights
-        Vector256<float> sW = Avx.Shuffle(source, source, shuffleAlphaControl);
-        Vector256<float> dW = Avx.Shuffle(destination, destination, shuffleAlphaControl);
+        Vector256<float> sW = Avx.Shuffle(source, source, ShuffleAlphaControl);
```
tannergooding, line 294: Since it's the same input twice, using `Avx.Permute` should be preferred. The same shuffle control is used.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
JimBobSquarePants: We actually have a helper that does exactly that in `SimdUtils.HwIntrinsics.MultiplyAdd`

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
saucecontrol: `vpermilps` is 1 byte larger than `vshufps`.  The only reason to prefer it is that it supports a memory operand where shuffle doesn't, but that doesn't apply here.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -243,12 +288,9 @@ public static Vector4 Over(Vector4 destination, Vector4 source, Vector4 blend)
...
+        Vector256<float> dW = Avx.Shuffle(destination, destination, ShuffleAlphaControl);
         Vector256<float> blendW = Avx.Multiply(sW, dW);
 
         Vector256<float> dstW = Avx.Subtract(dW, blendW);
```
tannergooding, line 298: You should be able to build a `MultiplyAddNegatedEstimate` helper using `Fma.MultiplyAddNegate` since that does `-(a * b) + c`

JimBobSquarePants: I'd need [`Vector256.Negate<T>`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.intrinsics.vector256.negate?view=net-7.0) for that wouldn't I?

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -351,20 +389,17 @@ public static Vector4 In(Vector4 destination, Vector4 source)
...
     {
-        const int blendAlphaControl = 0b_10_00_10_00;
-        const int shuffleAlphaControl = 0b_11_11_11_11;
-
         // calculate alpha
-        Vector256<float> sW = Avx.Shuffle(source, source, shuffleAlphaControl);
-        Vector256<float> dW = Avx.Shuffle(destination, destination, shuffleAlphaControl);
+        Vector256<float> sW = Avx.Shuffle(source, source, ShuffleAlphaControl);
+        Vector256<float> dW = Avx.Shuffle(destination, destination, ShuffleAlphaControl);
         Vector256<float> alpha = Avx.Multiply(sW, dW);
```
tannergooding, line 397: You should be able to change this to:
```csharp
Vector256<float> alpha = Avx.Permute(Avx.Multiply(source, destination), ShuffleAlphaControl);
```

That will do a single shuffle/permute, rather than two of them. In general, if you can do ops and shuffle less after its "better" because shuffle ports are typically limited.

In this case its safe since all the operations are `element-wise`, so even though you're multiplying `source[i] * destination[i]`, the ones for `i = 0, 1, 2` don't matter and just get ignored in the subsequent permute.

It's possible there are similar opportunities in the shuffle ops in the methods above here, but they are "more complex" than this one so I didn't do any in depth analysis to look for the same simplification/optimizations in them

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
tannergooding: Not sure off the top of my head. In .NET 7/8 I'd expect us to be doing the right thing and recognizing them as a "common subexpression". I'd also expect the same for .NET 6, but the support wasn't quite as good and so manually hoisting it to a local may provide better results and shouldn't hurt on any of the target frameworks.

In general you can assume that most `Create(cns)` and `Create(cns, ..., cns)` come from memory and so using a local can help ensure there is only one memory access. The special consideration is `Zero` and `AllBitsSet` which are generated dynamically using a "zero cost" instruction instead.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
tannergooding: It's 1-byte larger but also preferred by Clang/GCC in practically all cases (both Intel and AMD have their instruction selection metadata set to prefer it).

There have been CPUs in the past (Knight's Landing most notably) where there was actual cost difference between them. Any CPU that decides to emulate YMM based operations via 2x 128-bit ops internally may likewise see worse performance.

It ultimately shouldn't make too much a difference, but I'd typically side with what Intel/AMD have made Clang/GCC emit.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
tannergooding: You'd use `Fma.MultiplyAddNegate(a, b, c)` and fallback to `Avx.Subtract(c, Avx.Multiply(a, b))` otherwise.

This is because `(x - y)` == `(-y + x)` and so you're just taking advantage of the specialized instruction `Fma` instruction that negates the multiplied result itself.

JimBobSquarePants:
Over, Atop, and Xor are currently broken. I'm not sure what I've done wrong as the individual functions themselves look correct.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
JimBobSquarePants: Yep. I see that. In this instance I use `blendW` further down so it looks like I'd be better off keeping as is since I would be introducing duplicate multiplications.

tests/ImageSharp.Benchmarks/PixelBlenders/PorterDuffBulkVsSingleVector.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        Vector256<float> result = default;
+        Vector256<float> opacity = Vector256.Create(.5F);
+        int count = this.backdrop.Length / 2;
+        for (int i = 0; i < count; i++)
+        {
+            result = PorterDuffFunctions.NormalSrcOver(Unsafe.Add(ref backdrop, i), Unsafe.Add(ref source, i), opacity);
+        }
+
+        return result;
```
JimBobSquarePants, line 66: Is there a bug in my benchmark? 

Numbers seem to good to be true.

```
.NET SDK=7.0.102


| Method |       Mean |   Error |  StdDev |
|------- |-----------:|--------:|--------:|
| Scalar | 2,176.5 ns | 4.48 ns | 3.97 ns |
|    Avx |   231.3 ns | 0.87 ns | 0.82 ns |

// * Hints *
Outliers
  PorterDuffBulkVsSingleVector.Scalar: Default -> 1 outlier  was  removed (2.20 us)

// * Legends *
  Mean   : Arithmetic mean of all measurements
  Error  : Half of 99.9% confidence interval
  StdDev : Standard deviation of all measurements
  1 ns   : 1 Nanosecond (0.000000001 sec)

// ***** BenchmarkRunner: End *****
// ** Remained 0 benchmark(s) to run **
Run time: 00:00:26 (26.39 sec), executed benchmarks: 2

Global total time: 00:00:47 (47.57 sec), executed benchmarks: 2
// * Artifacts cleanup *
```

jnyrup: I notice that both `backdrop` and `source` references `this.backdrop`?

JimBobSquarePants: Good catch! Doesn't affect the results though.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
tannergooding: https://github.com/dotnet/runtime/issues/82365

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
tannergooding: https://github.com/dotnet/runtime/issues/82366

tests/ImageSharp.Benchmarks/PixelBlenders/PorterDuffBulkVsSingleVector.cs
tannergooding: Just peaking at .NET 7 and changing the vector algorithm to be `V128` to simplify the diff a bit, then we have:

### Vector4
```asm
; Assembly listing for method C:NormalSrcOver(System.Numerics.Vector4,System.Numerics.Vector4,float):System.Numerics.Vector4
; Emitting BLENDED_CODE for X64 CPU with AVX - Windows
; optimized code
; rsp based frame
; partially interruptible
; No PGO data
; 0 inlinees with PGO data; 3 single block inlinees; 0 inlinees without PGO data

G_M000_IG01:                ;; offset=0000H
       4883EC18             sub      rsp, 24
       C5F877               vzeroupper 
       C5F8293424           vmovaps  xmmword ptr [rsp], xmm6
 
G_M000_IG02:                ;; offset=000CH
       498D400C             lea      rax, bword ptr [r8+0CH]
       C5E25900             vmulss   xmm0, xmm3, dword ptr [rax]
       C5FA1100             vmovss   dword ptr [rax], xmm0
       C4C1781000           vmovups  xmm0, xmmword ptr [r8]
       C5F828C8             vmovaps  xmm1, xmm0
       C5F81012             vmovups  xmm2, xmmword ptr [rdx]
       C5E8C6DAFF           vshufps  xmm3, xmm2, xmm2, -1
       C5F0C6E1FF           vshufps  xmm4, xmm1, xmm1, -1
       C5E259EC             vmulss   xmm5, xmm3, xmm4
       C5E25CDD             vsubss   xmm3, xmm3, xmm5
       C5DA5CF5             vsubss   xmm6, xmm4, xmm5
       C5E258E4             vaddss   xmm4, xmm3, xmm4
       C4E27918DB           vbroadcastss xmm3, xmm3
       C5E859D3             vmulps   xmm2, xmm2, xmm3
       C4E27918DE           vbroadcastss xmm3, xmm6
       C5F059CB             vmulps   xmm1, xmm1, xmm3
       C5E858C9             vaddps   xmm1, xmm2, xmm1
       C4E27918D5           vbroadcastss xmm2, xmm5
       C5F859C2             vmulps   xmm0, xmm0, xmm2
       C5F058C0             vaddps   xmm0, xmm1, xmm0
       C5F828CC             vmovaps  xmm1, xmm4
       C5F8101532000000     vmovups  xmm2, xmmword ptr [reloc @RWD00]
       C5E85FC9             vmaxps   xmm1, xmm2, xmm1
       C4E27918C9           vbroadcastss xmm1, xmm1
       C5F85EC1             vdivps   xmm0, xmm0, xmm1
       C4E37921C430         vinsertps xmm0, xmm0, xmm4, 48
       C5F81101             vmovups  xmmword ptr [rcx], xmm0
       488BC1               mov      rax, rcx
 
G_M000_IG03:                ;; offset=0088H
       C5F8283424           vmovaps  xmm6, xmmword ptr [rsp]
       4883C418             add      rsp, 24
       C3                   ret   
```

### Vector128

```asm
G_M000_IG01:                ;; offset=0000H
       4883EC18             sub      rsp, 24
       C5F877               vzeroupper 
       C5F8293424           vmovaps  qword ptr [rsp], xmm6
 
G_M000_IG02:                ;; offset=000CH
       C4C1791000           vmovupd  xmm0, xmmword ptr [r8]
       C4C1785909           vmulps   xmm1, xmm0, xmmword ptr [r9]
       C4E3790CC188         vblendps xmm0, xmm0, xmm1, -120
       C4C1791100           vmovupd  xmmword ptr [r8], xmm0
       C4C1791000           vmovupd  xmm0, xmmword ptr [r8]
       C5F828C8             vmovaps  xmm1, xmm0
       C5F91012             vmovupd  xmm2, xmmword ptr [rdx]
       C4E37904D8FF         vpermilps xmm3, xmm0, -1
       C4E37904E2FF         vpermilps xmm4, xmm2, -1
       C5E059EC             vmulps   xmm5, xmm3, xmm4
       C5D85CE5             vsubps   xmm4, xmm4, xmm5
       C5E05CF5             vsubps   xmm6, xmm3, xmm5
       C5D858DB             vaddps   xmm3, xmm4, xmm3
       C5E859D4             vmulps   xmm2, xmm2, xmm4
       C4E279B8D6           vfmadd231ps xmm2, xmm0, xmm6
       C4E271B8D5           vfmadd231ps xmm2, xmm1, xmm5
       C5E05F0530000000     vmaxps   xmm0, xmm3, xmmword ptr [reloc @RWD00]
       C5E85ED0             vdivps   xmm2, xmm2, xmm0
       C4E3690CC388         vblendps xmm0, xmm2, xmm3, -120
       C5F91101             vmovupd  xmmword ptr [rcx], xmm0
       488BC1               mov      rax, rcx
 
G_M000_IG03:                ;; offset=0071H
       C5F8283424           vmovaps  xmm6, qword ptr [rsp]
       4883C418             add      rsp, 24
       C3                   ret
```

The `V256` code is effectively identical to the `V128` code, just using `ymm` instead of `xmm` and we can assume is roughly 2x the speed.

For the differences between `Vector4` and `V128`...

tannergooding: The initial `source.W *= opacity` isn't great, it's generating:

```asm
lea      rax, bword ptr [r8+0CH]
vmulss   xmm0, xmm3, dword ptr [rax]
vmovss   dword ptr [rax], xmm0
vmovupd  xmm0, xmmword ptr [r8]
```

If `NormalSrcOver` was inlined and `source` was in register, it'd also be generating a spill before this:
```asm
vmovapd  xmmword ptr [r8], xmm0
```

That basically means it's doing:
* write source to memory
* get the address of `w`
* load `w` from memory and multiply by `opacity`
* store result into `w`
* load entire vector from memory

Compare this with the `V128` approach:
```asm
vmovupd  xmm0, xmmword ptr [r8]
vmulps   xmm1, xmm0, xmmword ptr [r9]
vblendps xmm0, xmm0, xmm1, 8
```

Which is:
* load source from memory
* load opacity from memory and multiply by source
* blend so that you get original source plus new W

If you instead did something roughly equivalent, making it `Vector4 opacity` and then did `source.W = (source * opacity).W;`, you get this instead:
```csharp
vmovupd  xmm0, xmmword ptr [r8]
vmulps   xmm1, xmm0, xmmword ptr [r9]
vshufps  xmm1, xmm1, xmm1, -1
vinsertps xmm0, xmm0, xmm1, 48
```

Which is nearly ideal and doing the same thing, there's just a stray `vshufps` that I need to remove on the JIT side >.>

tannergooding: The next 6 instructions are "the same" between the two algorithms, but there is a slight codegen pessimization from the JIT:

We have for `Vector4`:
```asm
vmovupd  xmmword ptr [r8], xmm0
vmovupd  xmm0, xmmword ptr [r8]
vmovaps  xmm1, xmm0
vmovupd  xmm2, xmmword ptr [rdx]
vshufps  xmm3, xmm2, xmm2, -1
vshufps  xmm4, xmm1, xmm1, -1
```

and for `V128`:
```asm
vmovupd  xmmword ptr [r8], xmm0
vmovupd  xmm0, xmmword ptr [r8]
vmovaps  xmm1, xmm0
vmovupd  xmm2, xmmword ptr [rdx]
vpermilps xmm3, xmm0, -1
vpermilps xmm4, xmm2, -1
```

They really just differ on whether `shuffle` or `permute` is picked. There is an initial spill and load of `source` which shouldn't be necessary, but the JIT is introducing due to some complexities around struct returns on Windows (it's caused by `Normal(backdrop, source)` being a "nop" effectively) and also due to required evaluation order of passed parameters.

There's not anything for you to do here, just describing the difference.

tannergooding: The last bit primarily comes down to the `Vector4` version again pulling out and operating on scalars, rather than treating it as a "single vector". There is also a difference in whether `Fma(a, b, c)` or `Add(Multiply(a, b), c)` is used, but that's "less important" and could still be handled with a tiny helper method.

Changing the `Vector4` implementation to more closely match `V128`, gives us:
```csharp
Vector4 sW = new Vector4(source.W);
Vector4 dW = new Vector4(destination.W);

// calculate weights
Vector4 blendW = sW * dW;
Vector4 dstW = dW - blendW;
Vector4 srcW = sW - blendW;

// calculate final alpha
Vector4 alpha = dstW + sW;

// calculate final color
Vector4 color = (destination * dstW);
color = MultiplyAdd(color, source, srcW);
color = MultiplyAdd(color, blend, blendW);

// unpremultiply
color /= Vector4.Max(alpha, new Vector4(Epsilon));
color.W = alpha.W;
return color;
```

This generates:
```asm
C5F0C6D9FF           vshufps  xmm3, xmm1, xmm1, -1
C4E27918DB           vbroadcastss xmm3, xmm3
C5E8C6E2FF           vshufps  xmm4, xmm2, xmm2, -1
C4E27918E4           vbroadcastss xmm4, xmm4
C5E059EC             vmulps   xmm5, xmm3, xmm4
C5D85CE5             vsubps   xmm4, xmm4, xmm5
C5E05CF5             vsubps   xmm6, xmm3, xmm5
C5D858DB             vaddps   xmm3, xmm4, xmm3
C5E859D4             vmulps   xmm2, xmm2, xmm4
C5F059CE             vmulps   xmm1, xmm1, xmm6
C5F058D2             vaddps   xmm2, xmm1, xmm2
C5F859C5             vmulps   xmm0, xmm0, xmm5
C5F858D2             vaddps   xmm2, xmm0, xmm2
C4E27918053C000000   vbroadcastss xmm0, dword ptr [reloc @RWD00]
C5E05FC0             vmaxps   xmm0, xmm3, xmm0
C5E85ED0             vdivps   xmm2, xmm2, xmm0
C5E0C6C3FF           vshufps  xmm0, xmm3, xmm3, -1
C4E36921D030         vinsertps xmm2, xmm2, xmm0, 48
C5F91111             vmovupd  xmmword ptr [rcx], xmm2
488BC1               mov      rax, rcx
```

This is quite a bit better, but still not "ideal". There are notably some unnecessary `broadcastss` after the first two `shufps` and an unnecessary `shufps` before the last `insertps` (same general thing as mentioned for the first scenario covering `source.W = (source * opacity).W`).

For those couple of remaining pessimizations there's nothing that ImageSharp could "improve" itself there for `Vector4`, its codegen fixes that the JIT will need to make for .NET 8

tannergooding: It could however "improve" it if it wanted.

Notably it could selectively use `vector4.AsVector128()` and `vector128.AsVector4()` to convert between `Vector4` and `Vector128<float>`. It could then specially handle these couple edge cases with small helper methods.

For example, you could define a helper:
```csharp
private static Vector4 PermuteW(Vector4 value)
{
    return Sse.Shuffle(value.AsVector128(), value.AsVector128(), 0xFF).AsVector4();
}
```

and then change the following:
```diff
-Vector4 sW = new Vector4(source.W); 
-Vector4 dW = new Vector4(destination.W);
+Vector4 sW = PermuteW(source);
+Vector4 dW = PermuteW(destination);
```

That then removes the two unnecessary `broadcastss`.

You could likewise create a helper for:
```csharp
private static Vector4 WithW(Vector4 value, Vector4 w)
{
    // TODO: Provide SSE fallback which uses "shuffle" - just pick XYZ from value and W from w
    return Sse41.Insert(value.AsVector128(), w.AsVector128(), 0b11_11_0000).AsVector4();
}
```

and do the following diff:
```diff
-source.W = (source * opacity).W;
+source = WithW(source, source * opacity);
```

also:
```diff
-color.W = alpha.W;
+color = WithW(color, alpha);
```

tannergooding: I'll make sure JIT issues are logged for the couple pessimizations on the JIT side so that manual workarounds aren't needed long term.

br3aker: Insanely detailed writeup, learnt a lot, thank you!

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -75,8 +112,19 @@ public static Vector4 Subtract(Vector4 backdrop, Vector4 source)
...
+    /// </summary>
+    /// <param name="backdrop">The backdrop vector.</param>
+    /// <param name="source">The source vector.</param>
+    /// <returns>The <see cref="Vector256{Single}"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector256<float> Screen(Vector256<float> backdrop, Vector256<float> source)
     {
-        return Vector4.One - ((Vector4.One - backdrop) * (Vector4.One - source));
+        Vector256<float> vOne = Vector256.Create(1F);
+        return Avx.Subtract(vOne, Avx.Multiply(Avx.Subtract(vOne, backdrop), Avx.Subtract(vOne, source)));
```
saucecontrol, line 127: Pretty much anywhere you have a multiply followed by add or multiply followed by subtract, you can use FMA.  This one would be

```cs
Fma.MultiplyAddNegated(Avx.Multiply(Avx.Subtract(vOne, backdrop), Avx.Subtract(vOne, source)), vOne);
```

With the appropriate helper for non-FMA, ofc.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -136,15 +212,44 @@ public static Vector4 HardLight(Vector4 backdrop, Vector4 source)
...
+    /// <param name="source">Source color element</param>
+    /// <returns>Overlay value</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector256<float> OverlayValueFunction(Vector256<float> backdrop, Vector256<float> source)
     {
-        return backdrop <= 0.5f ? (2 * backdrop * source) : 1 - (2 * (1 - source) * (1 - backdrop));
+        Vector256<float> vOne = Vector256.Create(1F);
+        Vector256<float> vTwo = Vector256.Create(2F);
+        Vector256<float> left = Avx.Multiply(Avx.Add(backdrop, backdrop), source);
+        Vector256<float> right = Avx.Subtract(vOne, Avx.Multiply(Avx.Multiply(vTwo, Avx.Subtract(vOne, source)), Avx.Subtract(vOne, backdrop)));
```
saucecontrol, line 249: Another MultiplyAddNegated opportunity here.  You can also remove the vTwo constant (which will be a memory load) by saving `Avx.Subtract(vOne, source)` into a temp and adding it to itself instead of multiplying by 2.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
```diff
@@ -238,6 +420,28 @@ public static Vector4 Out(Vector4 destination, Vector4 source)
...
+    /// <param name="destination">The destination vector.</param>
+    /// <param name="source">The source vector.</param>
+    /// <returns>The <see cref="Vector256{Single}"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector256<float> Out(Vector256<float> destination, Vector256<float> source)
+    {
+        // calculate alpha
+        Vector256<float> sW = Avx.Shuffle(source, source, ShuffleAlphaControl);
+        Vector256<float> dW = Avx.Shuffle(destination, destination, ShuffleAlphaControl);
+        Vector256<float> alpha = Avx.Multiply(Avx.Subtract(Vector256.Create(1F), dW), sW);
```
saucecontrol, line 435: Thought I saw Tanner comment on this yesterday, but it's disappeared.  You can save an instruction by doing the alpha shuffle after the multiply.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
saucecontrol: > (both Intel and AMD have their instruction selection metadata set to prefer it).

I don't know what this means.  The two instructions mostly do different things, with one overlapping case.  Can you point me at the metadata in question?  And any example where Clang does the substitution?

> There have been CPUs in the past (Knight's Landing most notably) where there was actual cost difference between them. Any CPU that decides to emulate YMM based operations via 2x 128-bit ops internally may likewise see worse performance for shuffle.

Ignoring Knight's Landing since it's not a real CPU, this argument doesn't make any sense to me.  This is an in-lane operation, so there's no reason splitting it in two would affect one and not the other.  In fact, the two existing architectures that do split AVX into 2x128 (Zen+ and Gracemont) have identical measured latency and throughput numbers.

Not that 1 byte here or there is especially worth arguing over, but since you've suggested it "should be preferred", I'd like to see something concrete.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
tannergooding: My comment should be here, hopefully it hasn't actually disappeared https://github.com/SixLabors/ImageSharp/pull/2359#discussion_r1111157263 😅

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
tannergooding: You can trivially see this in godbolt via:
```cpp
#include <immintrin.h>

__m256 Permute(__m256 x)
{
    return _mm256_shuffle_ps(x, x, 0xFF);
}
```

and that Clang and GCC both universally replace this with `vpermilps` regardless of ISAs or target `-march=...`.

I'll see if I can dig up, again, exactly where the instruction selection for this happens, but might take me a bit given the size/complexity of LLVM in general.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
saucecontrol: Ha, yeah, it just got marked as outdated so it doesn't show up in the code view anymore.  Saw it after I submitted the comments :)

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
saucecontrol: Well that's interesting.  At least trusty old MSVC does what it's told 😉

https://godbolt.org/z/3eTKGqchf

I'd appreciate it if you can find the reference for why they do it.  I've seen Clang make some odd decisions before, but with GCC doing it too there's got to be something.

tests/ImageSharp.Benchmarks/PixelBlenders/PorterDuffBulkVsSingleVector.cs
JimBobSquarePants: This is incredible  @tannergooding Thanks so much!

I'm gonna have a tinker and see where I can get. `NormalSrcOver` is currently 9.2X faster using AVX which is absolutely bonkers. CC @antonfirsov 

```
.NET SDK=7.0.102


| Method |       Mean |   Error |  StdDev | Ratio |
|------- |-----------:|--------:|--------:|------:|
| Scalar | 2,177.0 ns | 4.38 ns | 3.66 ns |  1.00 |
|    Avx |   233.5 ns | 1.25 ns | 1.17 ns |  0.11 |
```

JimBobSquarePants: MAJOR improvement for the Vector4 path. @tannergooding

I implemented the `WithW` and the things immediate dropped to nearly 4X faster. `PermuteW` brought it down even further.

The only thing I couldn't figure out was how to fallback to `Sse` for `WithW`. I'm not great with shuffling.

```
.NET SDK=7.0.102


| Method |     Mean |   Error |  StdDev | Ratio |
|------- |---------:|--------:|--------:|------:|
| Scalar | 448.6 ns | 0.99 ns | 0.88 ns |  1.00 |
|    Avx | 235.1 ns | 1.23 ns | 1.03 ns |  0.52 |
```

JimBobSquarePants:
OK. I think I'm happy here unless there are any objections. 

The old Vector4 path is ~5x faster than before.
The new AVX path is 9x faster than before.

src/ImageSharp/PixelFormats/PixelBlenders/PorterDuffFunctions.cs
JimBobSquarePants: I've implemented that now.

tests/ImageSharp.Benchmarks/PixelBlenders/PorterDuffBulkVsSingleVector.cs
tannergooding: > MAJOR improvement for the Vector4 path.

Awesome! Much closer to the 1.5-2x difference we'd otherwise expect

> The only thing I couldn't figure out was how to fallback to Sse for WithW.

Think you want:
```
Vector128<float> tmp = Sse.Shuffle(w.AsVector128(), value.AsVector128(), 0b00_10_00_11);
return Sse.Shuffle(value.AsVector128(), tmp, 0b00_10_01_00);
```

That will create `tmp` as `<w[3], w[0], value[2], value[0]>`
Then return `<value[0], value[1], tmp[2], tmp[0]>` (which is `<value[0], value[1], value[2], w[3]>`)

JimBobSquarePants: I would not have figured that one out. Thanks!

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -573,10 +573,30 @@ public static class HwIntrinsics
...
+        /// <remarks>ret = c - (a * b)</remarks>
+        /// <param name="a">The first vector to multiply.</param>
+        /// <param name="b">The second vector to multiply.</param>
+        /// <param name="c">The vector to add negated to the intermediate result.</param>
+        /// <returns>The <see cref="Vector256{T}"/>.</returns>
+        [MethodImpl(InliningOptions.ShortMethod)]
+        public static Vector256<float> MultiplyAddNegated(
+            in Vector256<float> a,
+            in Vector256<float> b,
+            in Vector256<float> c)
```
tannergooding, line 592: Particularly since this is inlined, I wouldn't expect you to need `in` here (or on the other examples above). 

It shouldn't actually make a difference since the JIT should inline things and elide the "address taken", but it's typically easier on the JIT if you don't mark them unnecessarily for "enregisterable" types.

saucecontrol: IIRC, without the `in` JIT could/would lose track of the fact memory load operands could be contained when inlining these small helpers, at least in netcoreapp3.1.  Don't know if or when that was fixed.

JimBobSquarePants: I haven't looked at the JIT but am seeing a 2% slowdown when running my benchmark when I remove them. Could be error but I'll leave them for now.

tests/ImageSharp.Benchmarks/PixelBlenders/PorterDuffBulkVsSingleVector.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    [Benchmark(Description = "Scalar", Baseline = true)]
+    public Vector4 OverlayValueFunction_Scalar()
+    {
+        Vector4 result = default;
+        for (int i = 0; i < this.backdrop.Length; i++)
+        {
+            result = PorterDuffFunctions.NormalSrcOver(this.backdrop[i], this.source[i], .5F);
```
antonfirsov, line 46: The only concern with these benchmarks is that - unlike [the real span code](https://github.com/SixLabors/ImageSharp/blob/c06da8c4bcf37cbdc327a90d38bedbbff08fca73/src/ImageSharp/PixelFormats/PixelBlenders/DefaultPixelBlenders.Generated.tt#L129) - they do not walk the `source` and `amount` buffers, meaning there are fewer cache misses. 

I would have created a parametric benchmark (N=2,16,128,512,1024) that uses the Span API-s, and run it against the code before/after the changes. Shouldn't significantly impact the result, but would paint a more nuanced picture about the improvements.

Anyways, amazing job!

JimBobSquarePants: They actually used to walk those buffers. I was surprised by that, so I replicated it in the benchmark.

https://github.com/SixLabors/ImageSharp/blob/f53c78d7570c06eca02030e24d9c104e20a5431c/src/ImageSharp/PixelFormats/PixelBlenders/DefaultPixelBlenders.Generated.tt#L87-L103

I have actually realized that the benchmarks were actually slightly generous to the old approach as I'd forgotten the clamp on `amount`

PR #2364, opened 2023-02-21
#2231 First steps for removing nullable disable in webp

stefannikolei:
### Prerequisites

### Description
#2231 Remove nullable disable from some classes in the webp format

### DIFF
--- a/src/ImageSharp/Formats/Webp/BitReader/Vp8BitReader.cs
+++ b/src/ImageSharp/Formats/Webp/BitReader/Vp8BitReader.cs
@@ -186,7 +186,7 @@ internal class Vp8BitReader : BitReaderBase
     {
         if (this.pos < this.bufferMax)
         {
-            ulong inBits = BinaryPrimitives.ReadUInt64LittleEndian(this.Data.Memory.Span.Slice((int)this.pos, 8));
+            ulong inBits = BinaryPrimitives.ReadUInt64LittleEndian(this.Data!.Memory.Span.Slice((int)this.pos, 8));
             this.pos += BitsCount >> 3;
             ulong bits = ByteSwap64(inBits);
             bits >>= 64 - BitsCount;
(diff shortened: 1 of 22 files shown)


src/ImageSharp/Formats/Webp/BitReader/Vp8BitReader.cs
```diff
@@ -186,7 +186,7 @@ private void LoadNewBytes()
     {
         if (this.pos < this.bufferMax)
         {
-            ulong inBits = BinaryPrimitives.ReadUInt64LittleEndian(this.Data.Memory.Span.Slice((int)this.pos, 8));
+            ulong inBits = BinaryPrimitives.ReadUInt64LittleEndian(this.Data!.Memory.Span.Slice((int)this.pos, 8));
```
JimBobSquarePants, line 189: There's some horrible temporal coupling here. I don't think the correct fix would be to add `!`. It may require much deeper refactoring.

stefannikolei: Yeah probably. And there are some more classes where it is not so easy to remove the #nullable disable

JimBobSquarePants: I'm gonna be looking at some optimization opportunities in those classes. I think we should leave this until that is done. 

We've covered all the public bits now anyway I think?

stefannikolei: ok I tried to ref it. Can you have a look?

stefannikolei: image
Those are the areas where nullable disable is still set

JimBobSquarePants: > ok I tried to ref it. Can you have a look?

Ah nice, will do!

JimBobSquarePants: Refactor looks good. I just pushed some additional cleanup.

JimBobSquarePants: Actions seems to be having a wobble. I'll merge in the morning.

PR #2369, opened 2023-02-24
Fix and normalize Vector4 UnPremultiply

JimBobSquarePants:
### Prerequisites

### Description
Our existing `UnPremultiply` methods suffered from an issue in that they allowed dividing values by zero. This was masked by our tests since the transform to byte would lead to `zero` in most cases but could cause issues when chaining certain operations requiring premultiplication using single precision pixel formats since interim values for the `XYZ` components would be `∞`.

The Porter-Duff operators contained a bad fix for this by dividing by the max of the `W` component vs an epsilon of `0.001F`. This would lead to out-of-range values if the X, Y, Z components had a non-zero value. This was generally masked by clamping but again could cause issues for single precision pixel formats.

The fix is simple. Compare the alpha component to `zero` and return the source value if so. We return the source rather than an explicit zeroed value for two reasons:

1. For all our operation this would be equivalent to `Vector4.Zero`
2. There are Tiff reference files in our tests suite that are encoded with associated alpha values of `zero` but have `non-zero` RGB component values. This allows us to maintain the integrity of those values. (_Note the reference decoder makes the same mistake we did before this fix_).

The fix does not affect the performance of `UnPremultiply`. Migrating the Porter-Duff operators to use the fixed method however yields an additional 1.55x performance boost over the work done in #2359 

This brings an overall performance boost of 14.4x over our v2 version of the operators.

```bash
.NET SDK=7.0.200
```

Porter-Duff NormalSrcOver

Main
| Method |     Mean |   Error |  StdDev | Ratio |
|------- |---------:|--------:|--------:|------:|
| Scalar | 416.1 ns | 0.86 ns | 0.80 ns |  1.00 |
|    Avx | 235.3 ns | 0.62 ns | 0.55 ns |  0.57 |

PR

| Method |     Mean |   Error |  StdDev | Ratio |
|------- |---------:|--------:|--------:|------:|
| Scalar | 354.6 ns | 0.56 ns | 0.50 ns |  1.00 |
|    Avx | 151.4 ns | 0.43 ns | 0.40 ns |  0.43 |

PremultiplyVector4

Main

|              Method |     Mean |    Error |   StdDev | Ratio |
|-------------------- |---------:|---------:|---------:|------:|
| PremultiplyBaseline | 31.16 us | 0.061 us | 0.054 us |  1.00 |
|         Premultiply | 22.95 us | 0.050 us | 0.044 us |  0.74 |

PR

|              Method |     Mean |    Error |   StdDev | Ratio |
|-------------------- |---------:|---------:|---------:|------:|
| PremultiplyBaseline | 31.17 us | 0.074 us | 0.066 us |  1.00 |
|         Premultiply | 22.94 us | 0.042 us | 0.039 us |  0.74 |

UnPremultiplyVector4

Main

|                Method |     Mean |     Error |    StdDev | Ratio |
|---------------------- |---------:|----------:|----------:|------:|
| UnPremultiplyBaseline | 1.539 us | 0.0023 us | 0.0020 us |  1.00 |
|         UnPremultiply | 1.265 us | 0.0015 us | 0.0013 us |  0.82 |

PR

|                Method |     Mean |     Error |    StdDev | Ratio |
|---------------------- |---------:|----------:|----------:|------:|
| UnPremultiplyBaseline | 1.538 us | 0.0031 us | 0.0029 us |  1.00 |
|         UnPremultiply | 1.266 us | 0.0013 us | 0.0012 us |  0.82 |

One thing I do not understand though is why is there such a difference between `Premultiply` vs `UnPremultiply`

The baseline implementations in the benchmark use virtually identical code and generate [virtually identical ASM](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgDkBXfGKASzFwBuGjWIBmJqQYBhBgG8aDJQwAOfAG7YMMJuSRMUDAAqx8HADYZeK8wE8AFLABmDAGowwGaIdwQOUMBgASkVlBWplSIYncwgtBgB3BgBeBl9/QJYAdWEIqKV0gJ0AKlSE3PyCvyLslMSK5QBfUKUW1Q0tHWZ9YkMAVQA7ExgzS2s7RxgXd09vNOrAkLylcMro2Pik1MLMnLbInZ0AejKG/MPa07bm6kagA).

@tannergooding would you have any insight? 

```c#
private static void Premultiply(ref Vector4 source)
{
    float w = source.W;
    source *= w;
    source.W = w;
}

private static void UnPremultiply(ref Vector4 source)
{
    float w = source.W;
    source /= w;
    source.W = w;
}
```

```asm
C.Premultiply(System.Numerics.Vector4 ByRef)
    L0000: vzeroupper
    L0003: vmovss xmm0, [rcx+0xc]
    L0008: vmovupd xmm1, [rcx]
    L000c: vbroadcastss xmm2, xmm0
    L0011: vmulps xmm1, xmm1, xmm2
    L0015: vmovupd [rcx], xmm1
    L0019: vmovss [rcx+0xc], xmm0
    L001e: ret

C.UnPremultiply(System.Numerics.Vector4 ByRef)
    L0000: vzeroupper
    L0003: vmovss xmm0, [rcx+0xc]
    L0008: vmovupd xmm1, [rcx]
    L000c: vbroadcastss xmm2, xmm0
    L0011: vdivps xmm1, xmm1, xmm2
    L0015: vmovupd [rcx], xmm1
    L0019: vmovss [rcx+0xc], xmm0
    L001e: ret
```

### DIFF
--- a/src/ImageSharp/Common/Helpers/Numerics.cs
+++ b/src/ImageSharp/Common/Helpers/Numerics.cs
@@ -474,8 +474,10 @@ internal static class Numerics
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static void Premultiply(ref Vector4 source)
     {
-        Vector4 alpha = PermuteW(source);
-        source = WithW(source * alpha, alpha);
+        // Load into a local variable to prevent accessing the source from memory multiple times.
+        Vector4 src = source;
+        Vector4 alpha = PermuteW(src);
+        source = WithW(src * alpha, alpha);
     }
 
     /// <summary>
--- a/src/ImageSharp/Common/Helpers/Numerics.cs
+++ b/src/ImageSharp/Common/Helpers/Numerics.cs
@@ -496,8 +496,8 @@ internal static class Numerics
             while (Unsafe.IsAddressLessThan(ref vectorsBase, ref vectorsLast))
             {
                 Vector256<float> source = vectorsBase;
-                Vector256<float> multiply = Avx.Permute(source, ShuffleAlphaControl);
-                vectorsBase = Avx.Blend(Avx.Multiply(source, multiply), source, BlendAlphaControl);
+                Vector256<float> alpha = Avx.Permute(source, ShuffleAlphaControl);
+                vectorsBase = Avx.Blend(Avx.Multiply(source, alpha), source, BlendAlphaControl);
                 vectorsBase = ref Unsafe.Add(ref vectorsBase, 1);
             }
 
@@ -610,7 +610,7 @@ internal static class Numerics
     {
         if (Sse.IsSupported)
         {
-            return Sse.Shuffle(value.AsVector128(), value.AsVector128(), 0b11111111).AsVector4();
+            return Sse.Shuffle(value.AsVector128(), value.AsVector128(), ShuffleAlphaControl).AsVector4();
         }
 
         return new(value.W);
--- a/src/ImageSharp/Common/Helpers/Numerics.cs
+++ b/src/ImageSharp/Common/Helpers/Numerics.cs
@@ -538,6 +538,8 @@ internal static class Numerics
             return;
         }
 
+        // Divide source by alpha if alpha is nonzero, otherwise set all components to match the source value
+        // Blend the result with the alpha vector to ensure that the alpha component is unchanged
         source = WithW(source / alpha, alpha);
     }
 
--- a/src/ImageSharp/Common/Helpers/Numerics.cs
+++ b/src/ImageSharp/Common/Helpers/Numerics.cs
@@ -474,21 +474,8 @@ internal static class Numerics
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static void Premultiply(ref Vector4 source)
     {
-        float w = source.W;
-        source *= w;
-        source.W = w;
-    }
-
-    /// <summary>
-    /// Reverses the result of premultiplying a vector via <see cref="Premultiply(ref Vector4)"/>.
-    /// </summary>
-    /// <param name="source">The <see cref="Vector4"/> to premultiply</param>
-    [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public static void UnPremultiply(ref Vector4 source)
-    {
-        float w = source.W;
-        source /= w;
-        source.W = w;
+        Vector4 alpha = PermuteW(source);
+        source = WithW(source * alpha, alpha);
     }
 
     /// <summary>
@@ -498,7 +485,7 @@ internal static class Numerics
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static void Premultiply(Span<Vector4> vectors)
     {
-        if (Avx2.IsSupported && vectors.Length >= 2)
+        if (Avx.IsSupported && vectors.Length >= 2)
         {
             // Divide by 2 as 4 elements per Vector4 and 8 per Vector256<float>
             ref Vector256<float> vectorsBase = ref Unsafe.As<Vector4, Vector256<float>>(ref MemoryMarshal.GetReference(vectors));
@@ -507,7 +494,7 @@ internal static class Numerics
             while (Unsafe.IsAddressLessThan(ref vectorsBase, ref vectorsLast))
             {
                 Vector256<float> source = vectorsBase;
-                Vector256<float> multiply = Avx.Shuffle(source, source, ShuffleAlphaControl);
+                Vector256<float> multiply = Avx.Permute(source, ShuffleAlphaControl);
                 vectorsBase = Avx.Blend(Avx.Multiply(source, multiply), source, BlendAlphaControl);
                 vectorsBase = ref Unsafe.Add(ref vectorsBase, 1);
             }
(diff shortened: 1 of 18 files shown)


tannergooding:
> One thing I do not understand though is why is there such a difference between Premultiply vs UnPremultiply

Floating-point division is quite a bit slower than multiplication (although not quite as bad as integral division).

`divps` takes around `11-12 cycles` to execute (its latency) and has about `3 cycles` before another non-dependent instruction can execute on the same port (reciprocal throughput).

`mulps` on the other hand takes `3-4 cycles` to execute and has about `0.5 cycles` before another non-dependent instruction can execute on the same port.

----------

This is notably another case where you'd probably see some benefit in using the `PermuteW` and `WithW` helpers.

What you have now is roughly:
```asm
C.Premultiply(System.Numerics.Vector4 ByRef)
    L0000: vzeroupper                   ;
    L0003: vmovss xmm0, [rcx+0xc]       ; 4-7 cycles;  port 2 or 3
    L0008: vmovupd xmm1, [rcx]          ; 4-7 cycles;  port 2 or 3
    L000c: vbroadcastss xmm2, xmm0      ; 1 cycle;     port 5
    L0011: vmulps xmm1, xmm1, xmm2      ; 4 cycles;    port 0 or 1
    L0015: vmovupd [rcx], xmm1          ; 4-10 cycles; port 2 or 3 or 7 + port 4
    L0019: vmovss [rcx+0xc], xmm0       ; 4-10 cycles; port 2 or 3 or 7 + port 4
    L001e: ret                          ;
```

Given dependencies, parallel dispatch, and other bits this will take around `17-39` cycles.

However, if you change the implementation to something like:
```csharp
private static void Premultiply1(ref Vector4 source)
{
    var vsource = source.AsVector128();

    Vector4 w = PermuteW(vsource);
    vsource *= w;
    source = WithW(vsource, w);
}
```

We instead generate:
```asm
    L0000: vzeroupper                           ;
    L0003: vmovupd xmm0, [rcx]                  ; 4-7 cycles;  port 2 or 3
    L0007: vpermilps xmm1, xmm0, 0xff           ; 1 cycle;     port 5
    L000d: vmulps xmm0, xmm0, xmm1              ; 4 cycles;    port 0 or 1
    L0011: vinsertps xmm0, xmm0, xmm1, 0xf0     ; 1 cycle;     port 5
    L0017: vmovupd [rcx], xmm0                  ; 4-10 cycles; port 2 or 3 or 7 + port 4
    L001b: ret                                  ;
```

Which is around `14-23` cycles instead, which avoids the extra memory accesses and uses less overall ports.

src/ImageSharp/Common/Helpers/Numerics.cs
```diff
@@ -474,21 +474,8 @@ private static void ClampImpl<T>(Span<T> span, T min, T max)
...
-    /// </summary>
-    /// <param name="source">The <see cref="Vector4"/> to premultiply</param>
-    [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public static void UnPremultiply(ref Vector4 source)
-    {
-        float w = source.W;
-        source /= w;
-        source.W = w;
+        Vector4 alpha = PermuteW(source);
+        source = WithW(source * alpha, alpha);
```
tannergooding, line 478: You'll want to load `source` into a local here.

```suggestion
        Vector4 src = source;

        Vector4 alpha = PermuteW(src);
        source = WithW(src * alpha, alpha);
```

The reasoning is that since `source` is a `ref`, each access ends up coming from memory. However, we're using the value multiple times (twice given the implementation) and so accessing it twice is unnecessary.

src/ImageSharp/Common/Helpers/Numerics.cs
```diff
@@ -532,24 +519,49 @@ public static void Premultiply(Span<Vector4> vectors)
...
+    public static void UnPremultiply(ref Vector4 source, Vector4 alpha)
+    {
+        if (alpha == Vector4.Zero)
+        {
+            return;
+        }
+
+        // Divide source by alpha if alpha is nonzero, otherwise set all components to match the source value
+        // Blend the result with the alpha vector to ensure that the alpha component is unchanged
+        source = WithW(source / alpha, alpha);
```
tannergooding, line 543: Branching in SIMD code is "generally" undesirable.

It's probably worth comparing this to an implementation that does `CompareEqual` + `Blend` instead (much like you're already doing for the `Avx2` path).

You should be able to use the `AsVector128()` again. With `Sse.CompareEqual` and using `Sse41.BlendVariable`.

If `Sse41` isn't supported (which is unlikely since its been around since 2007), you can replace it with `Sse.Or(Sse.And(left, mask), Sse.AndNot(mask, right))` 

-- That is: `(left & mask) | (right & ~mask)`
-- `Sse.AndNot(x, y)` does `~x & y`, despite its name implying `x & ~y`; a weird quirk of the instruction

src/ImageSharp/Common/Helpers/Numerics.cs
tannergooding: -- Since it's `AggressiveInlining`, its possible the JIT is already optimizing this almost equivalently. However, it might also make it more likely to spill the value since its "address taken" with multiple uses.

src/ImageSharp/Common/Helpers/Numerics.cs
tannergooding: -- `left` here represents the value to use if the comparison result (`mask`) was true for that element
-- `right` correspondingly is the value to use if it was `false`

JimBobSquarePants: Would you believe the version I had was actually faster!?

UnPremultiply using Vector4 and branching.

|                Method |     Mean |     Error |    StdDev | Ratio |
|---------------------- |---------:|----------:|----------:|------:|
| UnPremultiplyBaseline | 1.545 us | 0.0070 us | 0.0066 us |  1.00 |
|         UnPremultiply | 1.827 us | 0.0039 us | 0.0037 us |  1.18 |
|     UnPremultiplyBulk | 1.267 us | 0.0033 us | 0.0027 us |  0.82 |

UnPremultiply using Sse41and branching.

|                Method |     Mean |     Error |    StdDev | Ratio |
|---------------------- |---------:|----------:|----------:|------:|
| UnPremultiplyBaseline | 1.546 us | 0.0077 us | 0.0068 us |  1.00 |
|         UnPremultiply | 2.058 us | 0.0043 us | 0.0036 us |  1.33 |
|     UnPremultiplyBulk | 1.268 us | 0.0025 us | 0.0022 us |  0.82 |

JimBobSquarePants:
> > One thing I do not understand though is why is there such a difference between Premultiply vs UnPremultiply
> 
> Floating-point division is quite a bit slower than multiplication (although not quite as bad as integral division).
> 
> `divps` takes around `11-12 cycles` to execute (its latency) and has about `3 cycles` before another non-dependent instruction can execute on the same port (reciprocal throughput).
> 
> `mulps` on the other hand takes `3-4 cycles` to execute and has about `0.5 cycles` before another non-dependent instruction can execute on the same port.
> 

Thanks for all the info there, much appreciated.

What I'm seeing though is that the division is faster. A lot faster. 1.538 us vs 31.17 us. It makes no sense to me.

tannergooding:
> What I'm seeing though is that the division is faster. A lot faster. 1.538 us vs 31.17 us. It makes no sense to me.

Ah yes. I misread the numbers initially.

This isn't clear to me either and I'd probably end up doing a profile using `Intel VTune` or `AMD uProf` so I could get instruction level timing information at this point.

My gut instinct would be a difference in input data or in how BDN ir measuring the two profiles.

src/ImageSharp/Common/Helpers/Numerics.cs
JimBobSquarePants: Yeah. We'd only ever use in for single operations or for overlaps in the AVX run.

PR #2371, opened 2023-02-25
Rewrite ColorMatrix

stefannikolei:
### Prerequisites

### Description

Adopts the pattern introduced in .NET 8 by https://github.com/dotnet/runtime/pull/80091 for our `ColorMatrix` type.

A comfortable 2X speedup over main.

```
.NET SDK=7.0.200
```

Main

|         Method |     Mean |     Error |    StdDev | Ratio | RatioSD |
|--------------- |---------:|----------:|----------:|------:|--------:|
|      Transform | 8.413 us | 0.1002 us | 0.0888 us |  1.00 |    0.00 |
| Transform_Span | 8.406 us | 0.1590 us | 0.1488 us |  1.00 |    0.02 |

PR

|         Method |     Mean |     Error |    StdDev | Ratio | RatioSD |
|--------------- |---------:|----------:|----------:|------:|--------:|
|      Transform | 4.135 us | 0.0819 us | 0.1148 us |  1.00 |    0.00 |
| Transform_Span | 3.145 us | 0.0539 us | 0.0504 us |  0.76 |    0.02 |


JimBobSquarePants:
Ah... Bit of a problem here. 

The compiled binary we ship will never contain the .NET 7 code. We target .NET 6 only and everything else will be stripped out.

stefannikolei:
> Ah... Bit of a problem here.
> 
> The compiled binary we ship will never contain the .NET 7 code. We target .NET 6 only and everything else will be stripped out.

Then you will also have problems with LossyUtils. There's also Conditional code.

JimBobSquarePants:
> > Ah... Bit of a problem here.
> > The compiled binary we ship will never contain the .NET 7 code. We target .NET 6 only and everything else will be stripped out.
> 
> Then you will also have problems with LossyUtils. There's also Conditional code.

Yeah. I did spot that after it was merged but it's only a couple of lines which serves as a reminder for the future, so I left it. This is a lot of extra code.

JimBobSquarePants:
Is it just the `UnscopedRefAttribute` we're missing?

If so we can polyfill it. 

See https://github.com/Sergio0694/PolySharp/blob/615f6ccb3e59644e0e16752571c0db81ce780c5d/src/PolySharp.SourceGenerators/EmbeddedResources/System.Diagnostics.CodeAnalysis.UnscopedRefAttribute.cs

https://github.com/Sergio0694/PolySharp/blob/db53c4a2b3e931e0bf2489b0e405979421e44b64/tests/PolySharp.Tests/LanguageFeatures.cs#L127

stefannikolei:
> Is it just the `UnscopedRefAttribute` we're missing?
> 
> If so we can polyfill it.
> 
> See https://github.com/Sergio0694/PolySharp/blob/615f6ccb3e59644e0e16752571c0db81ce780c5d/src/PolySharp.SourceGenerators/EmbeddedResources/System.Diagnostics.CodeAnalysis.UnscopedRefAttribute.cs
> 
> https://github.com/Sergio0694/PolySharp/blob/db53c4a2b3e931e0bf2489b0e405979421e44b64/tests/PolySharp.Tests/LanguageFeatures.cs#L127

Yeah I think so. I will try.

JimBobSquarePants:
If this passes, that's awesome! Only thing missing then is the transform.

https://github.com/dotnet/runtime/pull/80091/files#diff-4b222986f29a2b43c247d90ef0dba0c56b067f839a300583b7ed8ca0ac14d246R636-R647

Our equivalent is in ColorNumerics.

stefannikolei:
It seems that we have a problem with this implementation on windows :(

tests/ImageSharp.Tests/Primitives/ColorMatrixTests.cs
```diff
@@ -75,10 +76,10 @@ public void ColorMatrixMultiply()
         m.M44 = (value1.M41 * value2.M14) + (value1.M42 * value2.M24) + (value1.M43 * value2.M34) + (value1.M44 * value2.M44);
 
         // Fifth row
-        m.M51 = (value1.M51 * value2.M11) + (value1.M52 * value2.M21) + (value1.M53 * value2.M31) + (value1.M54 * value2.M41) + value2.M51;
-        m.M52 = (value1.M51 * value2.M12) + (value1.M52 * value2.M22) + (value1.M53 * value2.M32) + (value1.M54 * value2.M52) + value2.M52;
-        m.M53 = (value1.M51 * value2.M13) + (value1.M52 * value2.M23) + (value1.M53 * value2.M33) + (value1.M54 * value2.M53) + value2.M53;
-        m.M54 = (value1.M51 * value2.M14) + (value1.M52 * value2.M24) + (value1.M53 * value2.M34) + (value1.M54 * value2.M54) + value2.M54;
+        m.M51 = (value1.M51 * value2.M11) + (value1.M52 * value2.M21) + (value1.M53 * value2.M31) + (value1.M54 * value2.M41);
```
JimBobSquarePants, line 79: What am I missing here. Why did this change?

JimBobSquarePants:
Th tests only fail on windows because we're not doing a comparison check on other platforms.

tests/ImageSharp.Tests/Primitives/ColorMatrixTests.cs
stefannikolei: Probably because I did something wrong and misunderstood the logic ;) 
Gonna check it.

JimBobSquarePants: Use `Matrix3x2` as a reference. That's what I did the first time round 👍

stefannikolei: I think I just forget adding the values of the 5th row

stefannikolei: I think I fixed it

JimBobSquarePants: Yep. Seems to be working. I just pushed an additional fix to optimize the transformation. You were adding each element separately when it could be a single instruction. 

I'd love to see a benchmark, if possible, compared to main? It's just about midnight here so I won't be able to do it.

stefannikolei: Gonna have a try setting up a benchmark. Do you have a hint where I should start?

JimBobSquarePants: There’s a benchmark project in the solution. I’d copy bits from the PremultiplyVector4 benchmark into a new file and directly test ColorNumerics.

stefannikolei: This is a Benchmark over the Multiply code which is in the unit tests

``` ini

BenchmarkDotNet=v0.13.0, OS=macOS 13.2.1 (22D68) [Darwin 22.3.0]
Intel Core i9-9880H CPU 2.30GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=7.0.200
  Job-HOQWTB : .NET 6.0.14 (6.0.1423.7309), X64 RyuJIT

Runtime=.NET 6.0  Arguments=/p:DebugType=portable  

```

Before:

|         Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
|      Transform | 34.36 ns | 0.370 ns | 0.346 ns |     - |     - |     - |         - |
| Transform_Span | 61.50 ns | 1.248 ns | 3.374 ns |     - |     - |     - |         - |

After:

|         Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
|      Transform | 30.11 ns | 0.215 ns | 0.180 ns |     - |     - |     - |         - |
| Transform_Span | 44.29 ns | 0.787 ns | 0.737 ns |     - |     - |     - |         - |

JimBobSquarePants (approved):
Thanks for doing this. A lovely speedup to add to the library.

PR #2373, opened 2023-02-27
Automatically push the build to NuGet on a tagged release

JimBobSquarePants:
### Prerequisites

### Description

Adds configuration to automatically push the build to NuGet on a tagged release.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -198,10 +198,10 @@ jobs:
 
       - name: Feedz Publish
         shell: pwsh
-        run: dotnet nuget push .\artifacts\*.nupkg -k ${{secrets.FEEDZ_TOKEN}} -s https://f.feedz.io/sixlabors/sixlabors/nuget/index.json -ss https://f.feedz.io/sixlabors/sixlabors/symbols
+        run: dotnet nuget push .\artifacts\*.nupkg -k ${{secrets.FEEDZ_TOKEN}} -s https://f.feedz.io/sixlabors/sixlabors/nuget/index.json -ss https://f.feedz.io/sixlabors/sixlabors/symbols --skip-duplicate
 
       - name: NuGet Publish
         if: ${{ startsWith(github.ref, 'refs/tags/') }}
         shell: pwsh
-        run: dotnet nuget push .\artifacts\*.nupkg -k ${{secrets.NUGET_TOKEN}} -s https://api.nuget.org/v3/index.json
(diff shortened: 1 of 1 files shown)


SkyeHoefling (approved):
This change makes a lot of sense to me. When I publish to multiple package feeds I typically set it up in separate tasks like this. Since you told me this is new for you I am going to explain what it looks like this does to me to ensure you got what you want.

* The step `Feedz Publish` will always publish regardless of tasks when this build runs
* The step `NuGet Publish` will only publish if you add a new tag, regardless of the tag name. This would include any Release Client or preview builds that may be numbered `1.0.0-rc1` or `1.0.0-preview-1`.

I think this PR captures what we discussed on twitter and the description and should do what you are looking to do.

## Additional Ideas
Some things that I do with this technique is tag any release including final releases and preview releases. I publish them to feeds slightly differently and need additional steps to build out the logic tree.

.github/workflows/build-and-test.yml
```diff
@@ -198,7 +198,10 @@ jobs:
 
       - name: Feedz Publish
         shell: pwsh
-        run: |
-          dotnet nuget push .\artifacts\*.nupkg -k ${{secrets.FEEDZ_TOKEN}} -s https://f.feedz.io/sixlabors/sixlabors/nuget/index.json
-          dotnet nuget push .\artifacts\*.snupkg -k ${{secrets.FEEDZ_TOKEN}} -s https://f.feedz.io/sixlabors/sixlabors/symbols
-        # TODO: If github.ref starts with 'refs/tags' then it was tag push and we can optionally push out package to nuget.org
+        run: dotnet nuget push .\artifacts\*.nupkg -k ${{secrets.FEEDZ_TOKEN}} -s https://f.feedz.io/sixlabors/sixlabors/nuget/index.json -ss https://f.feedz.io/sixlabors/sixlabors/symbols
```
MeikTranel, line 201: I'm pretty sure this will publish twice now once for the push on `main` and one for the tag build. I'm not sure how `MinVer` resolves the version on the push on main build, but theres a chance your pipeline could fail because either the CI or the tag build has already published to Feedz and you are trying a second time with the same version.

JimBobSquarePants: MinVer is great and will actually produce 2 different tags.

1. A minor for main
2. The tag value for the tag build.

JimBobSquarePants:
OK. I'm gonna give this a pop.

PR #2400, opened 2023-03-13
Add arm64 intrinsics for cmyk converter

stefannikolei:
### Prerequisites

### Description

Added Simd for CMYK. I added ARM64 because there was no matching method in arm32

Benchmarks:
|          Method |     Mean |   Error |  StdDev | Ratio | RatioSD |
|---------------- |---------:|--------:|--------:|------:|--------:|
|          Scalar | 288.7 ns | 1.75 ns | 1.64 ns |  1.00 |    0.00 |
|     SimdVector8 | 125.4 ns | 1.74 ns | 1.63 ns |  0.43 |    0.01 |
|   SimdVectorAvx |       NA |      NA |      NA |     ? |       ? |
| SimdVectorArm64 | 108.1 ns | 0.56 ns | 0.43 ns |  0.37 |    0.00 |

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.CmykArm64.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.CmykArm64.cs
@@ -80,7 +80,7 @@ internal abstract partial class JpegColorConverterBase
                 Vector128<float> ytmp = AdvSimd.Subtract(scale, Unsafe.Add(ref srcB, i));
                 Vector128<float> ktmp = AdvSimd.Min(ctmp, AdvSimd.Min(mtmp, ytmp));
 
-                Vector128<float> kMask = AdvSimd.CompareEqual(ktmp, scale);
+                Vector128<float> kMask = AdvSimd.Negate(AdvSimd.CompareEqual(ktmp, scale));
 
                 ctmp = AdvSimd.And(AdvSimd.Arm64.Divide(AdvSimd.Subtract(ctmp, ktmp), AdvSimd.Subtract(scale, ktmp)), kMask);
                 mtmp = AdvSimd.And(AdvSimd.Arm64.Divide(AdvSimd.Subtract(mtmp, ktmp), AdvSimd.Subtract(scale, ktmp)), kMask);
--- a/src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.CmykArm64.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.CmykArm64.cs
@@ -1,0 +1,96 @@
+// Copyright (c) Six Labors.
+// Licensed under the Six Labors Split License.
+
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.Arm;
+using System.Runtime.Intrinsics.X86;
+
+namespace SixLabors.ImageSharp.Formats.Jpeg.Components;
+
+internal abstract partial class JpegColorConverterBase
+{
+    internal sealed class CmykArm64 : JpegColorConverterArm
+    {
+        public CmykArm64(int precision)
+            : base(JpegColorSpace.Cmyk, precision)
+        {
+        }
+
+        /// <inheritdoc/>
+        public override void ConvertToRgbInplace(in ComponentValues values)
+        {
+            ref Vector128<float> c0Base =
+                ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(values.Component0));
+            ref Vector128<float> c1Base =
+                ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(values.Component1));
+            ref Vector128<float> c2Base =
+                ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(values.Component2));
+            ref Vector128<float> c3Base =
+                ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(values.Component3));
+
+            // Used for the color conversion
+            var scale = Vector128.Create(1 / (this.MaximumValue * this.MaximumValue));
+
+            nint n = (nint)(uint)values.Component0.Length / Vector128<float>.Count;
+            for (nint i = 0; i < n; i++)
+            {
+                ref Vector128<float> c = ref Unsafe.Add(ref c0Base, i);
+                ref Vector128<float> m = ref Unsafe.Add(ref c1Base, i);
+                ref Vector128<float> y = ref Unsafe.Add(ref c2Base, i);
+                Vector128<float> k = Unsafe.Add(ref c3Base, i);
+
+                k = AdvSimd.Multiply(k, scale);
+                c = AdvSimd.Multiply(c, k);
+                m = AdvSimd.Multiply(m, k);
+                y = AdvSimd.Multiply(y, k);
+            }
+        }
+
+        /// <inheritdoc/>
+        public override void ConvertFromRgb(in ComponentValues values, Span<float> rLane, Span<float> gLane, Span<float> bLane)
+            => ConvertFromRgb(in values, this.MaximumValue, rLane, gLane, bLane);
+
+        public static void ConvertFromRgb(in ComponentValues values, float maxValue, Span<float> rLane, Span<float> gLane, Span<float> bLane)
+        {
+            ref Vector128<float> destC =
+                ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(values.Component0));
+            ref Vector128<float> destM =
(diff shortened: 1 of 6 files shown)


antonfirsov:
The 1-2% differences don't seem to be horrible, but @stefannikolei do you have an opportunity to check the output images visually just to be on the safe side?

Also, I would make sure the increased tolerance only applies to ARM, and only for the images where it's necessary. We can define conditional percentage values right in `TheoryData` definitions:

https://github.com/SixLabors/ImageSharp/blob/5283d7736f3546c67b63d8a95d26830bb83f4836/tests/ImageSharp.Tests/Formats/Jpg/JpegEncoderTests.cs#L49-L54

stefannikolei:
@antonfirsov For me the visual test failed. 😞 
EncodeBaseline_Interleaved_Rgb24_cmyk_Cmyk-Q40
EncodeBaseline_NonInterleavedMode_Rgb24_cmyk_Cmyk-Q40

br3aker:
Don't have an ARM with me but I'll try to take a look on this today, looks like a clamping issue to me.

stefannikolei:
> Don't have an ARM with me but I'll try to take a look on this today, looks like a clamping issue to me.

If you want I can give you temp access to my arm vm.

antonfirsov:
We should also check RGB conversion output from #2397.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
```diff
@@ -20,7 +20,7 @@ public class JpegColorConverterTests
 
     private const int TestBufferLength = 40;
 
-    private const HwIntrinsics IntrinsicsConfig = HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX;
+    private const HwIntrinsics IntrinsicsConfig = HwIntrinsics.AllowAll | HwIntrinsics.DisableHWIntrinsic;
```
antonfirsov, line 23: I wonder why don't these tests fail with `JpegColorConverterArm64`?

stefannikolei:
> We should also check RGB conversion output from #2397.

The RGB pictures look good to me. 
EncodeBaseline_Interleaved_Rgb24_CalliphoraPartial_Rgb-Q100

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
brianpopow: Me too, we need to re-review the tests, they should fail.

src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.CmykArm64.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            nint n = (nint)(uint)values.Component0.Length / Vector128<float>.Count;
+            for (nint i = 0; i < n; i++)
+            {
+                Vector128<float> ctmp = AdvSimd.Subtract(scale, Unsafe.Add(ref srcR, i));
+                Vector128<float> mtmp = AdvSimd.Subtract(scale, Unsafe.Add(ref srcG, i));
+                Vector128<float> ytmp = AdvSimd.Subtract(scale, Unsafe.Add(ref srcB, i));
+                Vector128<float> ktmp = AdvSimd.Min(ctmp, AdvSimd.Min(mtmp, ytmp));
+
+                Vector128<float> kMask = AdvSimd.CompareEqual(ktmp, scale);
```
brianpopow, line 83: The AVX version has `Avx.CompareNotEqual` here, could that be the problem?

stefannikolei: Good catch. I will have a Look. Can not remember changing it while porting 🫣

stefannikolei: I tried debugging the code on VS 2022 but it just hangs when I step to the next line :(

This line should not make any difference in my eyes. Because it is ConvertFromRgb. The code which we need to look into is ConvertToRgb? Or am I wrong?

br3aker: `ConvertFromRgb` is used during encoding from ImageSharp to actual jpeg image.
`ConvertToRgb` is used during decding from jpeg to ImageSharp.
Problem can be in any of them.

stefannikolei: The tests work now :) Any chance to get the test output from the test run?

I changed it from Negate to Not. That makes sense :)

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
antonfirsov: @stefannikolei or @brianpopow  since you have a HW - can you try injecting an exception into the `CmykArm64` methods and running `dotnet test -f net6.0 --filter FullyQualifiedName~JpegColorConverterTests`? If it gets hit, there's probably a bug in the tests since it they didn't show the regression, if not, it's probably a bug in `FeatureTestRunner`.

stefannikolei: those failed: 
ImageSharp\tests\ImageSharp.Tests\Formats\Jpg\JpegColorConverterTests.cs:line 81

But I also saw all xxxFromVector methods fail because it is not supported on the hardware(vm)

src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.CmykArm64.cs
stefannikolei: I ran the tests. The images look good now

brianpopow:
@stefannikolei I just tried to build your branch on my ARM device, but I am getting build errors. I pretty sure those build errors are not related to your changes. I am getting compiler error `CS8170` in `ColorMatrix.impl.cs`. 

I am using dotnet sdk 6.0.113 on ubuntu. What version are you using? Any idea howto fix this?

stefannikolei:
> @stefannikolei I just tried to build your branch on my ARM device, but I am getting build errors. I pretty sure those build errors are not related to your changes. I am getting compiler error `CS8170` in `ColorMatrix.impl.cs`. 
> 
> 
> 
> I am using dotnet sdk 6.0.113 on ubuntu. What version are you using? Any idea howto fix this?

I am using 7.0.xxx that's why the dotnet6 arm builds/test is disabled. Cause the image only had dotnet6 and in dotnet6 sdk the implementation of colormatrix does not build. 

This is discussed in the rgb PR

JimBobSquarePants:
We should update the readme to say we need the NET7 SDK

src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.CmykArm64.cs
brianpopow: I can confirm, the image looks good now.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
brianpopow: @antonfirsov The `FeatureTestRunner` seems to work as expected. Some path of the intrinsics methods (`ConvertFromRgb`) were not covered by tests. That's why the error was not detected. I have added tests for those.

brianpopow: I have decided to revert the `IntrinsicsConfig` back to `HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX2`. I think it was a mistake to change it. The `IntrinsicsConfig` is only used by the Vector color conversion tests. On x64 disabling AVX2 makes sense to test SSE here.

This means on ARM the test will be run twice, once 'HwIntrinsics.AllowAll` and once `HwIntrinsics.DisableAVX2'. I dont know howto change it so the second run will be avoided, but I also think its not a big deal. If anyone has a better idea let me know.

stefannikolei: Perhaps we can use this lib https://github.com/AArnott/Xunit.SkippableFact and just skip the not supported runs. 

So we can define the desired ones (avx/sse/arm) as a theory.

JimBobSquarePants:
I'm happy here if everyone else is.

brianpopow (approved):
I also think, this good to be merged now.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
brianpopow: I dont think its really worth the extra effort only to skip an addional test run on ARM. We maybe need to re-evaluate once we add more ARM code / tests.

PR #2409, opened 2023-03-22
Port GrayscalConverter to Arm

stefannikolei:
Port Grayscaleconverter to ARM

Benchmarks:

```md
|        Method |      Mean |    Error |   StdDev | Ratio | RatioSD |
|-------------- |----------:|---------:|---------:|------:|--------:|
|        Scalar | 118.76 ns | 0.499 ns | 0.443 ns |  1.00 |    0.00 |
| SimdVectorAvx |        NA |       NA |       NA |     ? |       ? |
| SimdVectorArm |  53.57 ns | 0.410 ns | 0.364 ns |  0.45 |    0.00 |

```

### DIFF
--- a/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
+++ b/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
@@ -580,7 +580,7 @@ internal static partial class SimdUtils
                 return AdvSimd.Add(AdvSimd.Multiply(vm0, vm1), va);
             }
 
-            return Avx.Add(Avx.Multiply(vm0, vm1), va);
+            return Sse.Add(Sse.Multiply(vm0, vm1), va);
         }
 
         /// <summary>
--- a/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
+++ b/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
@@ -4,6 +4,7 @@
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.Arm;
 using System.Runtime.Intrinsics.X86;
 using SixLabors.ImageSharp.PixelFormats;
 
@@ -549,6 +550,34 @@ internal static partial class SimdUtils
             if (Fma.IsSupported)
             {
                 return Fma.MultiplyAdd(vm1, vm0, va);
+            }
+
+            return Avx.Add(Avx.Multiply(vm0, vm1), va);
+        }
+
+        /// <summary>
+        /// Performs a multiplication and an addition of the <see cref="Vector128{Single}"/>.
+        /// TODO: Fix. The arguments are in a different order to the FMA intrinsic.
+        /// </summary>
+        /// <remarks>ret = (vm0 * vm1) + va</remarks>
+        /// <param name="va">The vector to add to the intermediate result.</param>
+        /// <param name="vm0">The first vector to multiply.</param>
+        /// <param name="vm1">The second vector to multiply.</param>
+        /// <returns>The <see cref="Vector256{T}"/>.</returns>
+        [MethodImpl(InliningOptions.AlwaysInline)]
+        public static Vector128<float> MultiplyAdd(
+            in Vector128<float> va,
+            in Vector128<float> vm0,
+            in Vector128<float> vm1)
+        {
+            if (Fma.IsSupported)
+            {
+                return Fma.MultiplyAdd(vm1, vm0, va);
+            }
+
+            if (AdvSimd.IsSupported)
+            {
+                return AdvSimd.Add(AdvSimd.Multiply(vm0, vm1), va);
             }
 
             return Avx.Add(Avx.Multiply(vm0, vm1), va);
--- a/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
+++ b/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
@@ -566,9 +566,9 @@ internal static partial class SimdUtils
         /// <returns>The <see cref="Vector256{T}"/>.</returns>
         [MethodImpl(InliningOptions.AlwaysInline)]
         public static Vector128<float> MultiplyAdd(
-            in Vector128<float> va,
-            in Vector128<float> vm0,
-            in Vector128<float> vm1)
+            Vector128<float> va,
+            Vector128<float> vm0,
+            Vector128<float> vm1)
         {
             if (Fma.IsSupported)
             {
--- a/src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.GrayScaleArm.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.GrayScaleArm.cs
@@ -27,8 +27,8 @@ internal abstract partial class JpegColorConverterBase
             // Used for the color conversion
             var scale = Vector128.Create(1 / this.MaximumValue);
 
-            nint n = values.Component0.Length / Vector128<float>.Count;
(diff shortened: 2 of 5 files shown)


src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -554,6 +555,34 @@ public static class HwIntrinsics
...
+            {
+                return Fma.MultiplyAdd(vm1, vm0, va);
+            }
+
+            if (AdvSimd.IsSupported)
+            {
+                return AdvSimd.Add(AdvSimd.Multiply(vm0, vm1), va);
+            }
+
+            return Avx.Add(Avx.Multiply(vm0, vm1), va);
```
gfoidl, line 583:
```suggestion
            return Sse.Add(Sse.Multiply(vm0, vm1), va);
```

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -554,6 +555,34 @@ public static class HwIntrinsics
...
+        /// <remarks>ret = (vm0 * vm1) + va</remarks>
+        /// <param name="va">The vector to add to the intermediate result.</param>
+        /// <param name="vm0">The first vector to multiply.</param>
+        /// <param name="vm1">The second vector to multiply.</param>
+        /// <returns>The <see cref="Vector256{T}"/>.</returns>
+        [MethodImpl(InliningOptions.AlwaysInline)]
+        public static Vector128<float> MultiplyAdd(
+            in Vector128<float> va,
+            in Vector128<float> vm0,
+            in Vector128<float> vm1)
```
gfoidl, line 571:
```suggestion
            Vector128<float> va,
            Vector128<float> vm0,
            Vector128<float> vm1)
```
The `in` modifier isn't needed here. `in` is a readonly-reference, thus the address of the arguments must be exposed (i.e. passed via stack and not via registers). Here the JIT will inline the method, so that "address exposure" won't happen. To make JIT's job easier just don't specify `in` here.

src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.GrayScaleArm.cs
```diff
@@ -0,0 +1,68 @@
...
+        public override void ConvertToRgbInplace(in ComponentValues values)
+        {
+            ref Vector128<float> c0Base =
+                ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(values.Component0));
+
+            // Used for the color conversion
+            var scale = Vector128.Create(1 / this.MaximumValue);
+
+            nint n = values.Component0.Length / Vector128<float>.Count;
+            for (nint i = 0; i < n; i++)
```
gfoidl, line 31: With https://github.com/SixLabors/ImageSharp/pull/2401 there's a change to `nuint`, so please use that here too.
```suggestion
            nuint n = (uint)values.Component0.Length / (uint)Vector128<float>.Count;
            for (nuint i = 0; i < n; i++)
```
Note: I need to update the length-division in the other PR too (will write the reason for this over there).

src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.GrayScaleArm.cs
```diff
@@ -0,0 +1,68 @@
...
+            ref Vector128<float> srcBlue =
+                ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(bLane));
+
+            // Used for the color conversion
+            var f0299 = Vector128.Create(0.299f);
+            var f0587 = Vector128.Create(0.587f);
+            var f0114 = Vector128.Create(0.114f);
+
+            nint n = values.Component0.Length / Vector128<float>.Count;
+            for (nint i = 0; i < n; i++)
```
gfoidl, line 57:
```suggestion
            nuint n = (uint)values.Component0.Length / (uint)Vector128<float>.Count;
            for (nuint i = 0; i < n; i++)
```

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
gfoidl: > Why do you want to change this to sse?

These methods are defined at `Sse`.
`Avx : Sse2 : Sse` as hierarchy chain (excerpt) thus here the inherited member is used.
Personally I prefere using the type that defined the members, so I see "requirement is SSE" (and AVX isn't required for this method).

> be on pair with the other methods

Ah, that's what I missed here. Leaving the decision to you or the team.
Consistency is good. Maybe these should be cleaned up (changed to SSE where possible, but not as part of this PR).

stefannikolei: Ty for the explanation. Yeah I'll wait for @JimBobSquarePants 🙂

JimBobSquarePants: Yeah, you always want to use the lowest supported syntax. The IDE usually reminds me to do so when I haven't, dunno if that's Roslynator or built in analyzers.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
JimBobSquarePants: Yeah, that one's on me.

JimBobSquarePants (approved):
Great speed up! All these additions are going to be a gamechanger for ARM

JimBobSquarePants:
Looks like something in #2401 has changed the result of some of the output on ARM.

brianpopow:
> Looks like something in #2401 has changed the result of some of the WebP output on ARM.

Just stating the obvious: It seems impossible to me, that the changes from this PR could be responsible for WebP test failures.
@stefannikolei If you run the webp tests on you machine, do you see any test failing? 
You can run only Webp tests with: `dotnet test -c Release -f net6.0 --filter "Format=Webp"`

stefannikolei:
WebpDecoderTests
        WebpDecoder_CanDecode_Lossless_WithThreeTransforms
         WebpDecoder_CanDecode_Lossless_WithThreeTransforms<Rgba32>(provider: Webp/bike_lossless.webp[Rgba32])
         WebpDecoder_CanDecode_Lossless_WithThreeTransforms<Rgba32>(provider: Webp/color_cache_bits_11.webp[Rgba32])
        WebpDecoder_Decode_Resize
         WebpDecoder_Decode_Resize<Rgba32>(provider: Webp/bike_lossless.webp[Rgba32])


Yeah this tests fail

JimBobSquarePants:
> > Looks like something in #2401 has changed the result of some of the WebP output on ARM.
> 
> Just stating the obvious: It seems impossible to me, that the changes from this PR could be responsible for WebP test failures. @stefannikolei If you run the webp tests on you machine, do you see any test failing? You can run only Webp tests with: `dotnet test -c Release -f net6.0 --filter "Format=Webp"`

@brianpopow  #2401 is the low hanging fruit PR. WebP is touched there. I'm doing another code review as we didn't label the PR to catch ARM issues.

WebpDecoder_CanDecode_Lossless_WithThreeTransforms_Rgba32_color_cache_bits_11

gfoidl:
I also have a look at this right now.

gfoidl:
Hm, so far I can't spot the bug.

State so far:
* AVX --> ✔️
* ARM --> 🐛 
* SSE disabled --> scalar --> 🐛 

So I assume the bug isn't specific to ARM, rather it's anywhere in the scalar-path (that is hit by ARM too, but not with SSE/AVX).
Will search and debug further to smash the bug down....

gfoidl:
I found something, will send a PR in a moment (just checking if a similar bug doesn't exist elsewhere).

JimBobSquarePants:
OK. Let's get this merged as the changes are covered. Thanks

Issue #2411, opened 2023-03-23
System.ExecutionEngineException when trying to save image as webp


claustsen:
### Prerequisites

- [X] I have bought a Commercial License
- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

3.0.0

### Other ImageSharp packages and versions

None

### Environment (Operating system, version and so on)

Windows 11

### .NET Framework version

dotnet 6.0.15 and dotnet 7.0.4

### Description

When running the below code on attached image I get a System.ExecutionEngineException when n reaches 2. Have tried multiple quality settings where it also fails. Does not fail on loss less encoding.

It fails with the following stacktrace
```
Fatal error. Internal CLR error. (0x80131506)
   at System.Buffer._Memmove(Byte ByRef, Byte ByRef, UIntPtr)
   at System.Buffer.Memmove(Byte ByRef, Byte ByRef, UIntPtr)
   at System.IO.Strategies.BufferedFileStreamStrategy.WriteSpan(System.ReadOnlySpan`1<Byte>, System.ArraySegment`1<Byte>)
   at System.IO.FileStream.Write(System.ReadOnlySpan`1<Byte>)
   at SixLabors.ImageSharp.Formats.Webp.BitWriter.BitWriterBase.WriteAlphaChunk(System.IO.Stream, System.Span`1<Byte>, Boolean)
   at SixLabors.ImageSharp.Formats.Webp.BitWriter.Vp8BitWriter.WriteEncodedImageToStream(System.IO.Stream, SixLabors.ImageSharp.Metadata.Profiles.Exif.ExifProfile, SixLabors.ImageSharp.Metadata.Profiles.Xmp.XmpProfile, SixLabors.ImageSharp.Metadata.Profiles.Icc.IccProfile, UInt32, UInt32, Boolean, System.Span`1<Byte>, Boolean)
[...]

PR #2412, opened 2023-03-24
Webp: Dont use using statement for encodedAlphaData

brianpopow:
### Prerequisites

- [ ] I have provided test coverage for my change (where applicable) Not sure howto provide a test for this issue.

### Description

This PR changes the disposing of the encodedAlphaData to be in a try/finally block instead of using statement.
This is related to Issue #2411

### DIFF
--- a/src/ImageSharp/Formats/Webp/Lossy/Vp8Encoder.cs
+++ b/src/ImageSharp/Formats/Webp/Lossy/Vp8Encoder.cs
@@ -345,30 +345,6 @@ internal class Vp8Encoder : IDisposable
         int expectedSize = this.Mbw * this.Mbh * averageBytesPerMacroBlock;
         this.bitWriter = new Vp8BitWriter(expectedSize, this);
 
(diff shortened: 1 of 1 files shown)


src/ImageSharp/Formats/Webp/Lossy/Vp8Encoder.cs
```diff
@@ -405,16 +406,23 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream)
...
-            exifProfile,
-            xmpProfile,
-            metadata.IccProfile,
-            (uint)width,
-            (uint)height,
-            hasAlpha,
-            alphaData[..alphaDataSize],
-            this.alphaCompression && alphaCompressionSucceeded);
+        try
+        {
```
tocsoft, line 410: your going to want to start this `try` before the `if (hasAlpha)` line otherwise anything that fails before the `WriteEncodedImageToStream` line will cause the memory to leak

PR #2413, opened 2023-03-24
Fixed wrong division hack

gfoidl:
### Prerequisites

### Description

In https://github.com/SixLabors/ImageSharp/pull/2401 I introduced a bug (sorry for that) -- this PR fixes that.
The bit-hack for divison by powers of two is only valid for positive numerators, and in the failed testcase (cf. https://github.com/SixLabors/ImageSharp/pull/2409#issuecomment-1482798035) it was negative. As showcase see [this sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoayPAZwwAIBLPZgExgBsMBDJgF4mcAMwB2ANzVqRMgE4AFF14CA9ExIBKaVTlKV/JgD5jTMjqA===).

**Note to maintainers**: please run CI on ARM too (as I understand that's not on by default, so that bug even came into main-branch)

### DIFF
--- a/src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
+++ b/src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs
@@ -1456,7 +1456,7 @@ internal static unsafe class LosslessUtils
     }
 
     [MethodImpl(InliningOptions.ShortMethod)]
-    private static int AddSubtractComponentHalf(int a, int b) => (int)Clip255((uint)(a + ((a - b) >> 1)));  // >> 1 is bit-hack for / 2
+    private static int AddSubtractComponentHalf(int a, int b) => (int)Clip255((uint)(a + ((a - b) / 2)));
 
     [MethodImpl(InliningOptions.ShortMethod)]
     private static int AddSubtractComponentFull(int a, int b, int c) => (int)Clip255((uint)(a + b - c));


gfoidl:
While hunting this bug I tried several things (although in hindsight it's quite obvious that the wrong use of the bit-hack is the bug), amongst others disabling SSE with this https://github.com/SixLabors/ImageSharp/blob/0a1f05b746a6cf2e56b863bdc39705db20dff76a/.runsettings#L2
```xml
<?xml version="1.0" encoding="utf-8" ?>
<RunSettings>
    <RunConfiguration>
        <!--Used in conjunction with ActiveIssueAttribute to skip tests with known issues-->
        <TestCaseFilter>category!=failing</TestCaseFilter>

        <EnvironmentVariables>
            <DOTNET_EnableSSE>0</DOTNET_EnableSSE>
        </EnvironmentVariables>
    </RunConfiguration>
</RunSettings>
```
and more tests started to fail -- even when going back before https://github.com/SixLabors/ImageSharp/pull/2401 (e.g. to 5283d7736f3546c67b63d8a95d26830bb83f4836).
To reproduce together with that runsettings one can try this [playlist.zip](https://github.com/SixLabors/ImageSharp/files/11066186/playlist.zip) (for filtered tests).

To me it looks like there are latent bugs in the scalar code-pathes. Can anyone confirm? Should I open a separate issue to handle these (potential) failures?

brianpopow:
> In https://github.com/SixLabors/ImageSharp/pull/2401 I introduced a bug (sorry for that) -- this PR fixes that.

No need to be sorry here, it happens, no problem.

> Note to maintainers: please run CI on ARM too (as I understand that's not on by default, so that bug even came into main-branch)

ARM CI runner only runs when the PR is tagged with ARM. That's a compromise we need to make, because it costs us money to run those ARM CI runner. I know that's not ideal, but usually changes to ARM code was very rare in the past.

> and more tests started to fail -- even when going back before https://github.com/SixLabors/ImageSharp/pull/2401 (e.g. to https://github.com/SixLabors/ImageSharp/commit/5283d7736f3546c67b63d8a95d26830bb83f4836).
To reproduce together with that runsettings one can try this [playlist.zip](https://github.com/SixLabors/ImageSharp/files/11066186/playlist.zip) (for filtered tests).
To me it looks like there are latent bugs in the scalar code-pathes. Can anyone confirm? Should I open a separate issue to handle these (potential) failures?

Yes please open a issue for that, we should investigate that. We have a code coverage report, I will look into the report tomorrow and see if we would have known that its not covered.

One thing I would like to have is a simple unit test for `AddSubtractComponentHalf` to cover this issue. Just pick some random input values for a and b and make sure the right result is produced.

gfoidl:
> simple unit test for `AddSubtractComponentHalf` to cover this issue

It's a private method https://github.com/SixLabors/ImageSharp/blob/ca21b6d53d4443413664d1dc6de9192e5199eec9/src/ImageSharp/Formats/Webp/Lossless/LosslessUtils.cs#L1459

So for my PoV
* it's not worth it to expose it public / internal just for that test
* as we've seen it should be covered by tests -- but I'll open an issue for the other point (see above) and test-coverage will be handled in that issue for this piece of code too

> <sub>No need to be sorry here, it happens, no problem.</sub>

<sub>I know, but my own quality standard shouldn't allow such a mistake 😁</sub>

brianpopow:
> > simple unit test for AddSubtractComponentHalf to cover this issue

> It's a private method

Ah, yes your right, I overlooked that it's private.

brianpopow (approved):
@gfoidl thanks for looking into this issue and providing the fix.

PR #2416, opened 2023-03-26
Fix RgbScalar

stefannikolei:
### Prerequisites

### Description
I ran the tests with SSE Disabled. And I found this bug... 

Before fixing it I had 151 failing tests after it only 53 tests are failing. On a quick check all other failures are 'Image difference is over threshold!'

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverterBase.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverterBase.cs
@@ -229,7 +229,7 @@ internal abstract partial class JpegColorConverterBase
             return new RgbScalar(precision);
         }
 
-        return new GrayscaleScalar(precision);
+        return new RgbScalar(precision);
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


brianpopow:
We need a test to cover this issue. I will try to help here.

There are more color converter path's which are not covered by tests: [ColorConverters](https://app.codecov.io/github/SixLabors/ImageSharp/commit/d7cd46f503e0ffc9fc0149ab1f6f53de775ec8df/tree/src/ImageSharp/Formats/Jpeg/Components/ColorConverters). We need not add more unit tests there (@stefannikolei: not necessary you, btw).

brianpopow:
I have added tests now to cover all path for [JpegColorConverterBase](https://app.codecov.io/github/SixLabors/ImageSharp/commit/d7cd46f503e0ffc9fc0149ab1f6f53de775ec8df/blob/src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverterBase.cs)

I think its ready for review now.

antonfirsov (approved):
LGTM for the fix.

I wonder now however, why do we have `CmykArm64`, while we don't have `YCbCrArm64`. Doesn't `CmykArm64` and `CmykVector` have essentially the same performance?

antonfirsov:
> Ycbcr is implemented in https://github.com/SixLabors/ImageSharp/pull/2417

Ah for some reason I thought that was already merged 😄 The with those numbers it makes sense thanks!

PR #2418, opened 2023-03-26
Improvements around `fixed`

gfoidl:
### Prerequisites

### Description

* avoid length-check in pinning spans
* don't use a dummy fixed sized buffer to set the size of the `Block8x8` struct
* in `Block8x8F.Load` avoid unnecessary copying of data -- read it directly from the source

### DIFF
--- a/src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
+++ b/src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
@@ -167,7 +167,7 @@ public static class SRgbCompanding
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static unsafe void CompandAvx2(Span<Vector4> vectors, float[] table)
     {
-        fixed (float* tablePointer = &table[0])
+        fixed (float* tablePointer = &MemoryMarshal.GetArrayDataReference(table))
         {
             var scale = Vector256.Create((float)Scale);
             Vector256<float> zero = Vector256<float>.Zero;
@@ -199,7 +199,7 @@ public static class SRgbCompanding
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static unsafe void CompandScalar(Span<Vector4> vectors, float[] table)
     {
-        fixed (float* tablePointer = &table[0])
+        fixed (float* tablePointer = &MemoryMarshal.GetArrayDataReference(table))
         {
             Vector4 zero = Vector4.Zero;
             var scale = new Vector4(Scale);
--- a/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
+++ b/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
@@ -644,7 +644,7 @@ internal static partial class SimdUtils
             ReadOnlySpan<byte> source,
             Span<float> dest)
         {
-            fixed (byte* sourceBase = source)
+            fixed (byte* sourceBase = &MemoryMarshal.GetReference(source))
             {
                 if (Avx2.IsSupported)
                 {
--- a/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
+++ b/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
@@ -222,7 +222,7 @@ internal static partial class SimdUtils
                 ref Vector256<float> destBase =
                     ref Unsafe.As<float, Vector256<float>>(ref MemoryMarshal.GetReference(dest));
 
-                nint n = (nint)(uint)(dest.Length / Vector256<float>.Count);
+                nint n = (nint)dest.Vector256Count<float>();
                 nint m = Numerics.Modulo4(n);
                 nint u = n - m;
 
@@ -392,7 +392,7 @@ internal static partial class SimdUtils
                 ref Vector128<byte> destBase =
                     ref Unsafe.As<byte, Vector128<byte>>(ref MemoryMarshal.GetReference(dest));
 
-                nuint n = (uint)source.Length / (uint)Vector128<byte>.Count;
+                nuint n = source.Vector128Count<byte>();
 
                 for (nuint i = 0; i < n; i += 3)
                 {
@@ -455,7 +455,7 @@ internal static partial class SimdUtils
                 ref Vector128<byte> destBase =
                     ref Unsafe.As<byte, Vector128<byte>>(ref MemoryMarshal.GetReference(dest));
 
-                nuint n = (uint)source.Length / (uint)Vector128<byte>.Count;
+                nuint n = source.Vector128Count<byte>();
 
                 for (nuint i = 0, j = 0; i < n; i += 3, j += 4)
                 {
@@ -499,7 +499,7 @@ internal static partial class SimdUtils
                 ref Vector128<byte> destBase =
                     ref Unsafe.As<byte, Vector128<byte>>(ref MemoryMarshal.GetReference(dest));
 
-                nuint n = (uint)source.Length / (uint)Vector128<byte>.Count;
+                nuint n = source.Vector128Count<byte>();
 
                 for (nuint i = 0, j = 0; i < n; i += 4, j += 3)
                 {
@@ -679,7 +679,7 @@ internal static partial class SimdUtils
                 {
                     VerifySpanInput(source, dest, Vector256<byte>.Count);
 
-                    nuint n = (uint)dest.Length / (uint)Vector256<byte>.Count;
+                    nuint n = dest.Vector256Count<byte>();
 
                     ref Vector256<float> destBase =
                         ref Unsafe.As<float, Vector256<float>>(ref MemoryMarshal.GetReference(dest));
@@ -712,7 +712,7 @@ internal static partial class SimdUtils
                     // Sse
                     VerifySpanInput(source, dest, Vector128<byte>.Count);
 
-                    nuint n = (uint)dest.Length / (uint)Vector128<byte>.Count;
+                    nuint n = dest.Vector128Count<byte>();
 
                     ref Vector128<float> destBase =
                         ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(dest));
@@ -811,7 +811,7 @@ internal static partial class SimdUtils
             {
                 VerifySpanInput(source, dest, Vector256<byte>.Count);
 
-                nuint n = (uint)dest.Length / (uint)Vector256<byte>.Count;
+                nuint n = dest.Vector256Count<byte>();
 
                 ref Vector256<float> sourceBase =
                     ref Unsafe.As<float, Vector256<float>>(ref MemoryMarshal.GetReference(source));
@@ -850,7 +850,7 @@ internal static partial class SimdUtils
                 // Sse
                 VerifySpanInput(source, dest, Vector128<byte>.Count);
 
-                nuint n = (uint)dest.Length / (uint)Vector128<byte>.Count;
+                nuint n = dest.Vector128Count<byte>();
 
                 ref Vector128<float> sourceBase =
                     ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(source));
@@ -893,7 +893,7 @@ internal static partial class SimdUtils
             ref Vector256<byte> bBase = ref Unsafe.As<byte, Vector256<byte>>(ref MemoryMarshal.GetReference(blueChannel));
             ref byte dBase = ref Unsafe.As<Rgb24, byte>(ref MemoryMarshal.GetReference(destination));
 
-            nuint count = (uint)redChannel.Length / (uint)Vector256<byte>.Count;
+            nuint count = redChannel.Vector256Count<byte>();
 
             ref byte control1Bytes = ref MemoryMarshal.GetReference(PermuteMaskEvenOdd8x32);
             Vector256<uint> control1 = Unsafe.As<byte, Vector256<uint>>(ref control1Bytes);
@@ -965,7 +965,7 @@ internal static partial class SimdUtils
             ref Vector256<byte> bBase = ref Unsafe.As<byte, Vector256<byte>>(ref MemoryMarshal.GetReference(blueChannel));
             ref Vector256<byte> dBase = ref Unsafe.As<Rgba32, Vector256<byte>>(ref MemoryMarshal.GetReference(destination));
 
-            nuint count = (uint)redChannel.Length / (uint)Vector256<byte>.Count;
+            nuint count = redChannel.Vector256Count<byte>();
             ref byte control1Bytes = ref MemoryMarshal.GetReference(PermuteMaskEvenOdd8x32);
             Vector256<uint> control1 = Unsafe.As<byte, Vector256<uint>>(ref control1Bytes);
             var a = Vector256.Create((byte)255);
--- a/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
+++ b/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
@@ -4,6 +4,7 @@
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.Arm;
 using System.Runtime.Intrinsics.X86;
 using SixLabors.ImageSharp.PixelFormats;
 
@@ -552,6 +553,34 @@ internal static partial class SimdUtils
             }
 
             return Avx.Add(Avx.Multiply(vm0, vm1), va);
+        }
+
+        /// <summary>
+        /// Performs a multiplication and an addition of the <see cref="Vector128{Single}"/>.
+        /// TODO: Fix. The arguments are in a different order to the FMA intrinsic.
+        /// </summary>
+        /// <remarks>ret = (vm0 * vm1) + va</remarks>
+        /// <param name="va">The vector to add to the intermediate result.</param>
+        /// <param name="vm0">The first vector to multiply.</param>
+        /// <param name="vm1">The second vector to multiply.</param>
+        /// <returns>The <see cref="Vector256{T}"/>.</returns>
+        [MethodImpl(InliningOptions.AlwaysInline)]
+        public static Vector128<float> MultiplyAdd(
+            Vector128<float> va,
+            Vector128<float> vm0,
+            Vector128<float> vm1)
+        {
+            if (Fma.IsSupported)
+            {
+                return Fma.MultiplyAdd(vm1, vm0, va);
+            }
+
+            if (AdvSimd.IsSupported)
+            {
+                return AdvSimd.Add(AdvSimd.Multiply(vm0, vm1), va);
+            }
+
+            return Sse.Add(Sse.Multiply(vm0, vm1), va);
         }
 
         /// <summary>
--- a/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
+++ b/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
@@ -673,7 +673,7 @@ internal static partial class SimdUtils
             ReadOnlySpan<byte> source,
             Span<float> dest)
         {
-            fixed (byte* sourceBase = &MemoryMarshal.GetReference(source))
+            fixed (byte* sourceBase = source)
             {
                 if (Avx2.IsSupported)
                 {
--- a/src/ImageSharp/Formats/Jpeg/Components/Block8x8.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Block8x8.cs
@@ -15,24 +15,13 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components;
 /// 8x8 matrix of <see cref="short"/> coefficients.
 /// </summary>
 // ReSharper disable once InconsistentNaming
-[StructLayout(LayoutKind.Explicit)]
-internal unsafe partial struct Block8x8
+[StructLayout(LayoutKind.Explicit, Size = 2 * Size)]
+internal partial struct Block8x8
 {
     /// <summary>
     /// A number of scalar coefficients in a <see cref="Block8x8F"/>
     /// </summary>
     public const int Size = 64;
-
-#pragma warning disable IDE0051 // Remove unused private member
-    /// <summary>
-    /// A placeholder buffer so the actual struct occupies exactly 64 * 2 bytes.
-    /// </summary>
-    /// <remarks>
-    /// This is not used directly in the code.
-    /// </remarks>
-    [FieldOffset(0)]
-    private fixed short data[Size];
-#pragma warning restore IDE0051
 
     /// <summary>
     /// Gets or sets a <see cref="short"/> value at the given index
--- a/src/ImageSharp/Formats/Jpeg/Components/Block8x8.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Block8x8.cs
@@ -63,9 +63,10 @@ internal partial struct Block8x8
 
     public static Block8x8 Load(Span<short> data)
     {
-        Unsafe.SkipInit(out Block8x8 result);
-        result.LoadFrom(data);
-        return result;
+        DebugGuard.MustBeGreaterThanOrEqualTo(data.Length, Size, "data is too small");
+
+        ref byte src = ref Unsafe.As<short, byte>(ref MemoryMarshal.GetReference(data));
+        return Unsafe.ReadUnaligned<Block8x8>(ref src);
     }
 
     /// <summary>
@@ -93,9 +94,10 @@ internal partial struct Block8x8
     /// </summary>
     public void CopyTo(Span<short> destination)
     {
-        ref byte selfRef = ref Unsafe.As<Block8x8, byte>(ref this);
-        ref byte destRef = ref MemoryMarshal.GetReference(MemoryMarshal.Cast<short, byte>(destination));
-        Unsafe.CopyBlockUnaligned(ref destRef, ref selfRef, Size * sizeof(short));
+        DebugGuard.MustBeGreaterThanOrEqualTo(destination.Length, Size, "destination is too small");
+
+        ref byte destRef = ref Unsafe.As<short, byte>(ref MemoryMarshal.GetReference(destination));
+        Unsafe.WriteUnaligned(ref destRef, this);
     }
 
     /// <summary>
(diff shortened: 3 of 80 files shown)


src/ImageSharp/ColorSpaces/Companding/SRgbCompanding.cs
```diff
@@ -167,7 +167,7 @@ public static float Compress(float channel)
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static unsafe void CompandAvx2(Span<Vector4> vectors, float[] table)
     {
-        fixed (float* tablePointer = &table[0])
+        fixed (float* tablePointer = &MemoryMarshal.GetArrayDataReference(table))
```
gfoidl, line 170: This avoids the bound-checks.
See [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4a9Jq049+ggJI8ZEMfKhKV6mpoDMDbrmwAzGE1IMhDGgG8aDAKZ7YhQGAFlyAApgAE8MGABtAF0GABNsDGwASm9qQIY/PPzA514EGFSGaLiYACoGMQwoBgBeNIzs/2KAwu7ugDEICEjGqCyNIu6AXy6Amcnc/OJg0LDSavjk9syc2YK9/NLyyo26hqbWhgAydMyEuiSsg8DevvzB4dHx57m9+fy9ssmKtbKctrdsosXj8GEcKlVYvF6qNLlcwjB8NAYmFsFBcAALbCSFgAcRgGAAglAoNgYgARDpsGCuWBcFSRCFZJ4LfKvN4BD4jJrfHmBf5ivZ7BLojD4iCpAziSSRGVyhVKgDyjV4EC4uBYADkIEZJLwuGaAOZZJJ7MRQXgKDJuEIMQWIs5fPY+Pp/GhTIA===) for example.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -644,7 +644,7 @@ public static class HwIntrinsics
             ReadOnlySpan<byte> source,
             Span<float> dest)
         {
-            fixed (byte* sourceBase = source)
+            fixed (byte* sourceBase = &MemoryMarshal.GetReference(source))
```
gfoidl, line 647: This avoids the length-check.

```c#
fixed (byte* sourceBase = source)
```
is compiled to (via pattern matching)
```c#
fixed (byte* ptr = &span.GetPinnableReference())
```
and `Span<T>.GetPinnableReference` does a length-check which we don't need here, so we can omit it.

See [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4a9Jq049+ggJI8ZEMfKhKV6mpoDMDbrmwAzGE1IMhDGgG8aDAKZ7YhQGAFlyAAo2GGwAEwB5LkkAT1kxbC4AHmAUjBgAPgZcDK4ASm9qQIY/KurA514EGDiGSNz8gCoGMQwoBgBeYtKy/3qA2vHxgDEICEjeqDKNOvGAXzGAjdXK6uJg0LDSaNjE5LTSnLzC4cyKzZqH6sbm1vbr7sXBhgAyMJh8NAUmFsFBcAALbCSFgAcRgGBirlgXBUkRKd1GO2qkym1Vm80WyyegW21VJgQeAG1/hhwRA4gZxJJIjS6QymQlerwIFxcCwAHIQIySXhcUUAczKAF0HmIoLwFNh8kxQvj3l0en1Mdipg9tmsgA=) for example.

src/ImageSharp/Formats/Jpeg/Components/Block8x8.cs
```diff
@@ -15,25 +15,14 @@ namespace SixLabors.ImageSharp.Formats.Jpeg.Components;
 /// 8x8 matrix of <see cref="short"/> coefficients.
 /// </summary>
 // ReSharper disable once InconsistentNaming
-[StructLayout(LayoutKind.Explicit)]
-internal unsafe partial struct Block8x8
+[StructLayout(LayoutKind.Explicit, Size = 2 * Size)]
```
gfoidl, line 18: The unused fixed sized buffer to specify the size of the struct isn't neede, as the size can be set directly.

[Codegen](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8YUCAAcAymIBuvMDFwBuGjWIBmJqQYBhBjQDeNBqYa9RULtgA2DAEI2IYANYAOBG4YBZcgAoASgYAXgA+BgATGAAzbA4bDFVqM3NLaztHZ3dPAH0AQR9SQJDwqNj4xJoAXzVqAG0pDCgOMAwAGWwATwgODD8O7t6AaQsIlgBRBAkbBV4MNAYpXgAvGBCGHQAqRZWYAIBdGgsMMXSGCWwoPlsGXCaWjAcnVw83IxMzTQZILjvUx6Wq3WqCSNWoNAa91aAx6fRhw1GEymMzAcwORzSN24uGw0TWFyuvBud2arSeWVe+XeyU+Wh+f2OOyBwQYINqKTqADFeDAbBEAPLRaK4GB9OjommmCRQXiybAnBjRXgIGARW4AC2gjwi8uwDV2+1BNCAA===) for creating the struct is also a bit better -- only zeroing the memory, instead of stack-cookie + zeroing.
Though that won't matter much -- the reduction of C# code is the motivation for this change.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
```diff
@@ -101,24 +101,17 @@ internal partial struct Block8x8F : IEquatable<Block8x8F>
         set => this[((uint)y * 8) + (uint)x] = value;
     }
 
-    public static Block8x8F Load(Span<float> data)
-    {
-        Block8x8F result = default;
-        result.LoadFrom(data);
```
gfoidl, line 107: Going via instance method needs to copy two times, which can be avoided.

In [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgDkBXfGKASzFwBuGvSasAShwB2GXtxYBhCPgAOvADY8AyjwBu/GEJGNmLSTLkwWASRk8IKnVH1hDw6jV52oU7OqakDABiEBAMNADeNAwxTADMDABC6hBgANYAHAgZQQwAIjAAZtgc6hgAFFoq2FIAPIUp2BgAfAwAJk3YAJQMALytyamZ2UEsADIQ2G3lHRjd7rHxSSnpWTkA+gAK4gzbldV1DZMt7Z09/ctDa0Fb4uOT07PzNAC+NDQA2loYUBxgGGNsABPCAcCqAkFggDSXjaLAAoggVOp+LwMF0ALqeby+fzVKCyPwMXA/P4YS6rEaRaKxYgJZhICnDHIMCZTfY1eqNE5PLo0mJRaiLRaDSks2C4Urk3rtIolMoLYUxCVS+5TIJQZQzM6KpXEADsDBVCv5DDeHiFsQ+AFkYBgABYQNrWVTqcq2h1Ol3IgDyKlkECkuBYAEEAOZhiW4Xi6GC2FFSLxhzGmulMFCsh4arVVTlHJqtXCgqCuPmWgWmxawQoMYBAjAwYl9I1FBgAVSD2EKVhDuC5xzQtfrMGa5WrDFt+GgQOt2CguHtfhYAHE7eIijwYFJXOUixwSzAul1dcLx3WG+1m+OO7guz2+6LmUFB+eR2PWw7eLhj+9y8Kb3eigOECj7vjWbSDuOuCDlovAAF6NgAVMS8EwBAhTlPm6I/n+5qLKmCSQEG5JeOSsEIc2qDuKaHxBLwMDqG0PqFIUuB2uUdApn+aYAGowP80AZjxdBjCetH0YxzGsex5BIFxiy8fxGCCQwwniNRf7iQxTEsWxFRxKQ8m0gkfECVAQnkKJNF0dpUl6eUKAZEZMSKWZFnqb+ixaZJunsagzlLKZynmappBWZpNk+dJFQZJxWLcSZSkqTxpAeRaXmRTp0XlAAnHJ8UKYlbmqXE4UZRJWX2eQ5CGQVxmqUlIU8XEaXWRVdkyaQTl1S5RXBUJKBlVamUdRU5AoCgAWuf1qkoK1EXtb5Y1IHFBENcVPEAKxDTE3mVTJ+r5WtQXJZt83lbZS3lOQOW1cdjVCUgO0MHto3lKQdDdfdG1IOdw2LdlpCkJNPWBQ9qn6s9r1XcDq0JetM08fqaXmp83y/P8EKguCwLYzCUhwoiyKoui8WkTwuIMPihL+CSGPko+1y3NSmkeo6zquu6drs966h+gGQahhGUYxnGUgJkmU30rJTJM9smbsrmhzcq0vKmoKSrKq2r7EiWV6tgB3ahn2WEvsOo7jpO06zvOi7qCua4brA24wNqcxHieCmGobVjrlMHZ+LwYZSDAbS1IzIy3Bbra4CWOGLHhsRrURJIMKRDDkY2MpUZ5/2XdlcOFQjyUiWJI1XbJUvF01dB/bt5fZQZVcnU1lllwD9mOc34M8eQdcvQ39n+aD03JWF7f5/ZsXdxtqUaRdUX2XlM+I6VE+LzJNUr8lLXz3nG9jV129NYN6/7WNE3HwN/fQ9lsmF/VLdCdtZ9veQh1X6pZ17/XHcybdn8eJPVfjDT6gDfo/wHn/CoQMQbfURpDEBgMUAP16tXISyN3AvCAA=) it's hard to see (no method names, just the address of the call), so here from the JIT directly:

<details>
  <summary>JIT assembly</summary>

```asm
; Assembly listing for method Foo:Default(System.Span`1[float]):Block8x8F
; Emitting BLENDED_CODE for X64 CPU with AVX - Windows
; Tier-1 compilation
; optimized code
; rsp based frame
; partially interruptible
; No PGO data
; 0 inlinees with PGO data; 4 single block inlinees; 0 inlinees without PGO data

G_M000_IG01:                ;; offset=0000H
       56                   push     rsi
       4881EC00010000       sub      rsp, 256
       488BF1               mov      rsi, rcx

G_M000_IG02:                ;; offset=000BH
       488B12               mov      rdx, bword ptr [rdx]
       488D0C24             lea      rcx, bword ptr [rsp]
       41B800010000         mov      r8d, 256
       E80330AF5F           call     CORINFO_HELP_MEMCPY
       488BCE               mov      rcx, rsi
       488D1424             lea      rdx, bword ptr [rsp]
       41B800010000         mov      r8d, 256
       E8F12FAF5F           call     CORINFO_HELP_MEMCPY
       488BC6               mov      rax, rsi

G_M000_IG03:                ;; offset=0032H
       4881C400010000       add      rsp, 256
       5E                   pop      rsi
       C3                   ret

; Total bytes of code 59

; Assembly listing for method Foo:PR(System.Span`1[float]):Block8x8F_PR
; Emitting BLENDED_CODE for X64 CPU with AVX - Windows
; Tier-1 compilation
; optimized code
; rsp based frame
; partially interruptible
; No PGO data
; 0 inlinees with PGO data; 3 single block inlinees; 0 inlinees without PGO data

G_M000_IG01:                ;; offset=0000H
       56                   push     rsi
       488BF1               mov      rsi, rcx

G_M000_IG02:                ;; offset=0004H
       488B12               mov      rdx, bword ptr [rdx]
       488BCE               mov      rcx, rsi
       41B800010000         mov      r8d, 256
       E80B2FAF5F           call     CORINFO_HELP_MEMCPY
       488BC6               mov      rax, rsi

G_M000_IG03:                ;; offset=0018H
       5E                   pop      rsi
       C3                   ret

; Total bytes of code 26
```
</details>

gfoidl (commented):
Notes for review.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
antonfirsov: Do such (out-of-hot-loop) length checks have a measurable impact on perf? The new code feels bloated, I'm not in favor of such changes if there are no tangible gains.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
gfoidl: It depends what's done with the instance of `Block8x8F`. In [this example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgDkBXfGKASzFwBuGvSasAShwB2GXtxYBhCPgAOvADY8AyjwBu/GEJGNmLSTLkwWASRk8IKnVH1hDw6jV52oU7OqakDABiEBAMNADeNAwxTADMDABqMGAY0CgMACIwAGbYHOoYABRaKthSADw56hDYGAB8DAAmddgAlAwAvI0AQjVgANYAHAhDQSwAMrVNRS0Y7SyJdBPusfFJKWlQGdl5BRikJWWV1bUNza1t0bFR1GtrfRCDI2MMwF3NufmFq/cxwJNpkEoMpZpdfn9iAB2N6LZYQhgAX2uMWICWSqXSDAACuIjuUqjU6o05u0ur1+sNRkEAPq4wHYGaktpwlY0ZEeagAbS0GCgHFSE2wAE8IBxikLReKANJeJosACiCBU6n4vAwbQAup5vL5/GUoLI/AxcHyBRgGI9ntTIij1swkJbKS8ggwpoz8SciedmXbbn8Ylaqa9YLh9h8ml99gi1qH9gymsDQcyY7FoQw4z87Ry7VyALIwDAACwgTWsqnURQLxdL5ZVAHkVLIIFJcCwAIIAc07odwvF0MFsqqkXk7WrtaKYGXdiZB+E9hLOjVwYqgriudxudtjuTewowMBNH1gOQYAFVW9gclZ27hF3U0HuD/UiieGAX8NBhXnsFBcEW/BYABxQtxFyHgYCkVwihXDg1xgNo2lTGI32AfdDyaY9dwvXArxvO8gxdR80OfV9d2LXhcCQmht1iHC8MUBxhSDMjTyaR831wR8tF4AAvQ8ACoTT4mAIByIpTjqRCEQ5NYJwSSBWwtLwLR4/iPlQdxcyCXgYHUJp6xyHJcELIo6HHTdUXRTYsSWNlLIYLkdL0gyjJM4pyCQCy1knDEtgyJZxC0hynN0/TDOM0y4lIby02szFtiSch7LWUKXIi9yihQIZYqsjYEoC8ggpokLnPCtzTNQXL1j82zSBS2I0vKyLiiGcztQc3ybMSxJSGKzlUrK1yWqKABOLyOp8+L/KSOIGpiJrhsy8hyBiya4vymbEjifrtLCpbTNWnL1ry2qepQebHKGjLDpQFBqq6gqkhQXbSv2m6PKQdr5M22yAFZLsWj6inIKEJp+s6Ar+17BveiqPNGtaIe6gKkEB674aKUg6GO5GnsSJAYcajGRtIUh7pOmqUaSKF0bh0mUG+zrptsqF+pzbleX5QURTFCVeZlOVFWVVUwHVCyVJ4PUGANI1/FNbmLUI6k6XEW0QurEsywrKtCy1ut1EbZtWw7bte37QcpGHUcHoSB0nSeYNaVxN1pgXSSfUuP1aJQ3cSMPXA1yw096OvDs7w94j0JfN8Py/H8/wA9RgNA8DYCgmAwXmaSfaYGFQ6sMDGQvPxeE7KQYCaCplbGVWY93QOwGohzZNiH7FNNBgVIYNTD06BhNJK2H0sxpmpt+nr4T2keRs822J4CugiYWknMui+fIaSumZ8y7KN+pxIiuC4fmsyqrKcerb6oRIHMba/f8b64/ifpzLxofra5pv1fDtWj/bJ2s/Fer9DqkFxszBez1t6n1uhTPGW0XpAKuiAz6Y8Nqb0SADb+KCQZg3/j1aGSDb6z0Rvg1G0CDrFGxuA8eGDCZEJ/lQ8mZCaYUOBuTNBp0D5s3cIiIAA=) `Default2` shows no difference, as the JIT is smart enough to figure out what's going on. Namely: load the data onto the stack, then operate with the instance method of that type and use data on the stack.
But as soon the `Block8x8F` is passed around a copy needs to be made. Thus for me the instance `LoadFrom` has the potential for a perf-trap, and should be removed.

Besides that from an API point of view having an instance method to load (or better "fill") the block with data is a bit strange for an value-type*. Here a static method seems the right fit, like `int.Parse`, `Vector128.Load`, etc.
As it's an internal type this change seems fine from this view.

\* for reference types this is pretty much OK

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
gfoidl: For this method in HwIntrinsics I don't think it's really measurable in (micro-) benchmarks. But it reduces the machine code a bit and the length-check is just not needed at all here.

I'm more on the eager side to squeeze out unnecessary instructions.
But when you want me this change to revert I'm fine with that too.

> The new code feels bloated

The C# code is bloated. That's the obvious downside.

antonfirsov: We typically call these methods for entire image scanlines (->100's or 1000's of elements in the span). I would assume an extra bounds check at the beginning of the method is negligible, unless there are benchmarks proving it otherwise.

> The C# code is bloated.

When you turn it into semi-C, yes. But you don't always have to; we need to make informed, individual decisions.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
antonfirsov: > Besides that from an API point of view having an instance method to load (or better "fill") the block with data is a bit strange 

Note that `Block8x8F` is really huge. If I remember correctly, in early Jpeg code there were cases when existing on-stack blocks were being reused by repopulating it. If it's no longer the case in the current code, I'm happy with the change.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
gfoidl: OK - I'll revert.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.cs
gfoidl: Checked it -- all usages are safe. `Block8x8F` is either passed by reference, and only copied where needed (e.g. constructors).

>  existing on-stack blocks were being reused by repopulating

When I understand it right, then this is a [simple](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgDkBXfGKASzFwBuGvSasAShwB2GXtxYBhCPgAOvADY8AyjwBu/GEJGNmLSTLkwWASRk8IKnVH1hDw6jV52oU7OqakDABiEBAMNADeNAwxTADMTCgMACIwAGbYHOoYABRaKthSADxp6hDYGAB8DAAmFdhoDF4YDFIAlNGxUdSxvQwAQmVgANYAHAijQQzAQ8MMALy16ZnZ7n2xnetp0Aw5zU0LDHSCB0WtJwDUF7wdPesM3fd9MxAjLAAy5TVBUMo5dRhsG01k9YgBVXAwHKwNLTWbAzZ9AC+iJRdxiiOICWISQACuI8gViqVylVavVGvt2ojHvdtlBdvteIdjqdzgwrjdEb1aaCBrNxpMAPr4uGvOaLQbiwVBEXiD5ff71BHo0EQqEwsUjFVPNG9PUbVUAbQAsjAMAALCA1ayqdQ5M2W622lTqADyKlkECkuHYEFs6i8XgA5m0ALqIlR8XQVGCJBjq6HpfnSiZTF7ah4GjHGx1Wm12h3m/Mu92e3je31sf1SQNSEPhyPR2PxxOaqUjGVyrXDNpZmhomhGrQYKAcMAYd7YACeEA4uSns/nAGkvDUWABRBCu/i8DCN6jNHi+fwFKCyPwMXCj8ctDtjNORTHY8hIFOdtMMT7YGqEwolMoKmqAEgRpblYnvGUGFgXAshaRYamWOCQXWGC4IVH8fj+ECdXuYgAHZoMMZDURoRFTWLZ1CzzKjXQ9L0fRYABBYNgxg3BeF0GAAyDKRQwjVUsXjb9vl+fA/2JQCyVwOcoFcW4eXAmJNWAacMDjXBDk1MEfWwNIrCY3AANJRpVPUyok1hM18GgacTWwKBcAtPwWAAcXNcR0h4GApFcHIZI4OSYDaXDUOTMy4xqLTkx03A9IMozILTUy1JgCzNUtXhcBVJSE10/TFAcad70s2pGk1XBGi0XgAC84wAKivWqYAgNIchJCoQpQhhs3CQSEkgH0Wn2aq6sOVB3HIoJeBgdQajdNI0khXI6APXohIANRgCdoCSDa6HebqjWm2b5sW5aclfNbYk27aMF2hh9vESbjROuaFqW80cjiUhrpiW6dqgPbyEOqaZve86vpQUY/viR67oejbyGesjXvBs7PtyVBYYB+6gce0hQbR06Pou0ZVoE9aEi2wG9tIFGPGJiHMZyABOJAcephH8Y2uIid6Y70dJr7yHIX7KZurnaceuIGbBknIdyMWYYl/6pbxvaUH52JBYVlnyBQFBOfh6WNpQOWmYxi7Xwp58TY1x6AFZtZiXXmet/COdVuGaYdjbHYtgW3qtkXWfFu3fcRpAXYYN2Q9yUg6BViPub2pBA514PhYT0gje93HEfwmO4+znJc9t/r7cLhnB2oYcbwnRc5wXGdm9XKR1y3HcwD3A8jx8S8zwvfxrzHCd3wfYV8SfXNKILV0iydeeywY30WLYwwOK4nj6z443mDfJKp/EL9FXyf8OrJHCwNVXoVNSq85Oi2FYvi5ijMvlLzNK6zbPsxznLqDch5LysBfJQhwqFPoBE8pxQKp5H8Ok/C8GDFIGANQihH1lPidKyZcBySgb1O2g1rxNBkAwUacZFgTVRkHIWiscgVyplXHmB0jpZwYVdfO6tEZ0Azq7DhLMfrG0jjzEG7D6Es2hiI1Oj1kYvToXrC62NuEsLpsXQRZMmGSzUQTfhsdNFfXZjI02fMJFKJFmLExftZYKMzpI62pBk6V1EZrDRDiRaG2sYjc2diBEeKVkgbRatdH+3cRYpWntvE8wDn4gxATLph2iWncJ7svqJ2ccw1xj105xJLgw0gudkmPSLuYtJCcUDBJ9rIja+Ea5AA===) example of it.

Both variants work and produce the same machine-code.
(It's because the compilers hoists the loop-local block variable, and as it's a value-type the stack space needs to be reserved anyway).
Even if there are [two blocks](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgDkBXfGKASzFwBuGvSasAShwB2GXtxYBhCPgAOvADY8AyjwBu/GEJGNmLSTLkwWASRk8IKnVH1hDw6jV52oU7OqakDABiEBAMNADeNAwxTADMTCgMACIwAGbYHOoYABRaKthSADxp6hDYGAB8DAAmFdhoDF4YDFIAlNGxUdSxvQwAQmVgANYAHAijQQzAQ8OMALy16ZnZ7n2xgxAj45PTs+QMizXLWRhr6+E9F2nQDDnNTYcMdIKPRa2vANSfvB1X690Ln0Zls5iwADLlGpBKDKHJ1DDYNrnIExEEjViQ7DQ2H4eH1ZGdVEMACquBgOVgaT2oLohP+qLJFKpNIx9KBAF8iQwuQzucQEsQkgAFcR5ArFUrlKq1eqNB7tbmA663e4yR6LF5vD4Mb6/bm9ZXEzbbCZBAD6otZcyeJrGZst4ghUPxiPZxIGsx2Fqt6OGB0Wdu9jud2NdSJRjPJlPS1rpkaBTJj1L95Hd615vUzsW5AG0ALIwDAACwgNWsqnUOULJbLFZU6gA8ipZBApLh2BBbOovF4AOZtAC63JUfF0FRgiVJ0ZZQbN1raDAi2ZieZrpfLlerRY39abLd4bY7bC7Uh7Un7Q5HY4nU6Ts69Dt9s0Xy5ovJoua0GCgHDAGHBbAAE8IA4XJAJAsCAGkvBqFgAFEEAbfheAwK9qGaHhfH8AooFkPwGFwH8/xaOdJkiflBXIJBPVBb0GCxGpxUKEoygqaoESRJUDQ2R9dlgXBTieY4MlOBNYgE05QxxOFOPTPpiAAdgYSTVm5D8GQLHc6y3dcdIbZtW3bFgAEE+z7ATcF4XQYG7XspAHYcGQFKdGJhOF8hYqV2MI0CoFcP5DR4mIWWAICMEnXAnhZEl22wNIrBM3BWOlRowoiypkwYQt8GgID82wKBcGLPwWAAcSLcR0h4GApFcHJcD8gL5N6ULwsnGpotjWLcHixLkrIoI0vazKWRLXhcHpYLSTihLFAcIC7SympGhZXBGi0XgAC9JwAKkI7aYAgNIcm8tD5JXS5ehcyB2xaB5Np2p5UHcPMgl4GB1BqRs0jSclcjpJzroSAA1GB/2gJIQbocFI1zd7Pu+37/pyaj0OBhgwYhqAobocRXs0hGvp+v6ixyOJSHR2IXKxjBIcx8hYbej7ieRsmUFGKmYhp8G6Zxhn8ZoZnEZJlHUC5+JMd5+mQdIJnCZZpHSdyUZAcoqXsah0hBY8BWRbZ3IAE4kAlnnNcxuJ5d6eHFdFsnyHISmgep0Hpf5kG4h14XWeV1HSE553udd82QZQK3Yht/XffIFAUFN4O+ahlAvb1n2UeotXnITmWAFZw5iSO0/txSTcDyXadzlPraJpX08Np31Yr92kHzhhC9rsnSDoAPG7dqGkCriOa7t3JSFIOOy7NxPMcU1v25HnJx8zjGm6hxSdY0r9iP/CDQPA4C95gqQ4MQ5CwFQ9DMJ8AjcPw/wiN/f9aNNSZHQozS9M3Btt1rL/9yMjsZkLKGCsjZOyF4HLxzEDRQajoGIuk8pKNiMo5LcQZK1WM6VIr+S6tSHqfVTLJTOsNDKWUcp5QKkVEq6hyqVWqrAOqFI5ItWpspfBc0qrYlin4XgfYpAwBqEUWBopRqxlwP5C63J1a3SIk0dUj1JyLBekLVOHcAZQNXpjGGcNh4G1RqXXuIc8YE2rrbPRFMNF9wZnPXRvsOaWJDuQQeBdbFiwnoY6essbFmN9qrBxnjtYmKHj4lGxt/Ey0tjokJ9tHbhPdp7IJLjom5Edj3LOGtPFhyiVHdOsc4lJ2cW3Vx9skDLxdhk3O3icnFwMekzRIMc6FPnno8g9d8mYxbtkouo9u7tJBgPRJRTkmL3Hn02eXS1GLxQGUoOFT3br3cByIAA===) read then it's the same machine code.

I'll see if the C# code here can be simplified.

JimBobSquarePants:
Is this ready for another review?

gfoidl:
> Is this ready for another review?

Yes, please.

JimBobSquarePants (approved):
Looks good. Thanks!

PR #2419, opened 2023-03-26
Create vector constants inline and not via ROS<byte>

gfoidl:
### Prerequisites

### Description

On .NET 6 onwards (which is relevant here) it's better for codegen to define vector constant inline (including to have helper-methods for this).
* `private static readonly VectorXYZ`-fields produce less ideal code with more indirections
* reading the vector constant from a ROS<byte> (which may refer directly to assembly's data segment) results in some overhead

Now the vector constants can be read from memory directly.

<sub>PS: I don't have any other planned PRs -- right now weather is too bad / good enough to dig through this code 😉</sub>

### DIFF
--- a/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
+++ b/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
@@ -194,7 +194,7 @@ internal static partial class SimdUtils
         {
             if (Ssse3.IsSupported)
             {
-                int remainder = source.Length & (Vector128<byte>.Count * 4 - 1);    // bit-hack for modulo
+                int remainder = source.Length & ((Vector128<byte>.Count * 4) - 1);    // bit-hack for modulo
 
                 int sourceCount = source.Length - remainder;
                 int destCount = (int)((uint)sourceCount * 3 / 4);
--- a/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
+++ b/src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
@@ -13,33 +13,38 @@ internal static partial class SimdUtils
 {
     public static class HwIntrinsics
     {
-        public static ReadOnlySpan<byte> PermuteMaskDeinterleave8x32 => new byte[] { 0, 0, 0, 0, 4, 0, 0, 0, 1, 0, 0, 0, 5, 0, 0, 0, 2, 0, 0, 0, 6, 0, 0, 0, 3, 0, 0, 0, 7, 0, 0, 0 };
+        [MethodImpl(MethodImplOptions.AggressiveInlining)] // too much IL for JIT to inline, so give a hint
+        public static Vector256<int> PermuteMaskDeinterleave8x32() => Vector256.Create(0, 0, 0, 0, 4, 0, 0, 0, 1, 0, 0, 0, 5, 0, 0, 0, 2, 0, 0, 0, 6, 0, 0, 0, 3, 0, 0, 0, 7, 0, 0, 0).AsInt32();
 
-        public static ReadOnlySpan<byte> PermuteMaskEvenOdd8x32 => new byte[] { 0, 0, 0, 0, 2, 0, 0, 0, 4, 0, 0, 0, 6, 0, 0, 0, 1, 0, 0, 0, 3, 0, 0, 0, 5, 0, 0, 0, 7, 0, 0, 0 };
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Vector256<uint> PermuteMaskEvenOdd8x32() => Vector256.Create(0, 0, 0, 0, 2, 0, 0, 0, 4, 0, 0, 0, 6, 0, 0, 0, 1, 0, 0, 0, 3, 0, 0, 0, 5, 0, 0, 0, 7, 0, 0, 0).AsUInt32();
 
-        public static ReadOnlySpan<byte> PermuteMaskSwitchInnerDWords8x32 => new byte[] { 0, 0, 0, 0, 1, 0, 0, 0, 4, 0, 0, 0, 5, 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0, 6, 0, 0, 0, 7, 0, 0, 0 };
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Vector256<uint> PermuteMaskSwitchInnerDWords8x32() => Vector256.Create(0, 0, 0, 0, 1, 0, 0, 0, 4, 0, 0, 0, 5, 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0, 6, 0, 0, 0, 7, 0, 0, 0).AsUInt32();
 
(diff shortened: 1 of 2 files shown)


src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -221,7 +226,7 @@ public static class HwIntrinsics
                 ref Vector256<float> destBase =
                     ref Unsafe.As<float, Vector256<float>>(ref MemoryMarshal.GetReference(dest));
 
-                nint n = (nint)(uint)(dest.Length / Vector256<float>.Count);
+                nint n = (nint)((uint)dest.Length / (uint)Vector256<float>.Count);
```
gfoidl, line 229: I've forgotten these in the other PR (I think because they're kept as `nint` due the operation in L231).

stefannikolei:
@gfoidl On fire

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
antonfirsov: I wonder if we should create a helper method for these? Feels very noisy seeing this before all our SIMD loops.

```C#
internal static partial class Numerics
{
    // Not sure about the name
    public static nint VectorCount<T>(Span<T> dest, int vectorSize) => (nint)((uint)dest.Length / (uint)vectorSize);
}
```

So then we can do:

```C#
nint n = Numerics.VectorCount(dest, Vector256<float>.Count);
```

antonfirsov: @JimBobSquarePants meh or yeah?

gfoidl: I like the idea.
But it should be generic to account for different vector types. 

Maybe as extension to span?

antonfirsov: > But it should be generic to account for different vector types

I wish there was an `IIntrinsicVector<T>` interface with a static `Count` property but there isn't, and we are not on `net7.0+` anyways :) So I'm just passing `Vector256<float>.Count` as a parameter in my recommendation, meaning that there is no way to check if the `Vector256<T>` and `Span<T>` element types are the same. Not ideal, but still better than writing out the verbose `uint` magic everywhere IMHO.

gfoidl: > no way to check if the `Vector256<T>` and `Span<T>` element types are the same

They don't need to match. Reading e.g. `Vector128<float>` out of a `ROS<byte>` is pretty common.

So for the generic variant there could be [two generic arguments](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8ovLrl5hcAbho0AxFw4AbFdmAqYDACa9cGrfNoBmJqQYAxCBAY0A3jQbOmZ5eIwMAggAoAygAO2FwAPMAAnhgwAHwMuEFcAJQMALyx8cEsAGowYBjQAMIQ3BjkpAAcYZEwaAwAZioQ2BjRPoly1AC+xh4wUFzYKkzkSOYMALLhAKIIURK8EBIOTi7EZsyj7jwMOXmFxTxllQAqtce7+VCtGAAW+gyBwSHH6QmJKy4uAO43fdrnuUuDBAcQwUA4eQ+DEc1E+n2IAHYGD4OB5EhkuCwADIwLgAc1uDAA9MjUTxEhdoEdnpSriwiiUOp9ul0gA=). Here I don't like that the type of the span must be repeated. That should be infered (by the compiler).

In [your](https://github.com/SixLabors/ImageSharp/pull/2419#discussion_r1149799246) I don't like that `VectorXYZ<Type>.Count` needs to be typed.

What do you think of [this approach](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8ovLrl5hcAbho0AxFw4AbFdmAqYDACa9cGrfNoBmJqQYAxCBAY0A3jQbOmZ5eIwMAggAoAygAO2FwAPMAAnhgwAHwMuEFcAJQMALyx8cEsAGowYBjQAMIQ3BjkpAAcIQBmKhDYGNE+iXLULq4M7jwMAEL+CWGRMXEJyWnDmTl5hcU8pACsSNW19Y3NNAC+xh4wUFzYKkzkSOYMALLhAKIIURK8EBIOTi7EZszHnZ6T+VBFJWWVABUvtBGhgABb6BiBYIDKLpEZPNrOADuYJ22iBuW+DBAcQwUA4eURSLGPg4HkSGS4LAAMjAuABzcEMAD0DDJFOBUH+IUxUyg0RYvx4LTaxJeh3e5K6XOFGHmiz531BENwUP6ETh4ySxLaqPRDCV0BxeIJRNaSOcpOlGEpCVp9KZYNZ7JtiS5Ct5XMFcpa6yAA==) (IIRC it's only ROS<byte>/Span<byte>)?

antonfirsov: > What do you think of [this approach](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8ovLrl5hcAbho0AxFw4AbFdmAqYDACa9cGrfNoBmJqQYAxCBAY0A3jQbOmZ5eIwMAggAoAygAO2FwAPMAAnhgwAHwMuEFcAJQMALyx8cEsAGowYBjQAMIQ3BjkpAAcIQBmKhDYGNE+iXLULq4M7jwMAEL+CWGRMXEJyWnDmTl5hcU8pACsSNW19Y3NNAC+xh4wUFzYKkzkSOYMALLhAKIIURK8EBIOTi7EZszHnZ6T+VBFJWWVABUvtBGhgABb6BiBYIDKLpEZPNrOADuYJ22iBuW+DBAcQwUA4eURSLGPg4HkSGS4LAAMjAuABzcEMAD0DDJFOBUH+IUxUyg0RYvx4LTaxJeh3e5K6XOFGHmiz531BENwUP6ETh4ySxLaqPRDCV0BxeIJRNaSOcpOlGEpCVp9KZYNZ7JtiS5Ct5XMFcpa6yAA==) (IIRC it's only ROS<byte>/Span<byte>)?

I like this option the most. We can have one overload for each span type.

gfoidl: Super.

I prefer doing this in a separete PR, as it doesn't match with this PR (this location is a left-over from the other PR) and there a quite a few places that need to be touched.
Fine?

Anyway I'll do this change later this afternoon (here or in a new PR).

gfoidl: --> https://github.com/SixLabors/ImageSharp/pull/2422

JimBobSquarePants (approved):
I'm really curious to benchmark the library now vs v3.0.0. There's been lots of improvements!

gfoidl:
> curious to benchmark the library now vs v3.0.0

So am I too. Memory, and especially "static footprint" should be better. The latter is important for server usage, especially for serverless computing (less startup-cost, less memory consumption).

PR #2422, opened 2023-03-29
Added and used Numerics.Vector{128|256}Count extension

gfoidl:
### Prerequisites

### Description

Instead of typing 
```c#
nuint n = (uint)dest.Length / (uint)Vector256<float>.Count;
```
everytime, there's now a helper that shortens this to
```c#
nuint n = dest.Vector256Count<float>();
```
Cf. https://github.com/SixLabors/ImageSharp/pull/2419#discussion_r1149799246

### DIFF
--- a/tests/ImageSharp.Tests/TestUtilities/FeatureTesting/FeatureTestRunner.cs
+++ b/tests/ImageSharp.Tests/TestUtilities/FeatureTesting/FeatureTestRunner.cs
@@ -62,7 +62,7 @@ public static class FeatureTestRunner
             ProcessStartInfo processStartInfo = new();
             if (intrinsic.Key != HwIntrinsics.AllowAll)
             {
-                processStartInfo.Environment[$"COMPlus_{intrinsic.Value}"] = "0";
+                processStartInfo.Environment[$"DOTNET_{intrinsic.Value}"] = "0";
 
                 RemoteExecutor.Invoke(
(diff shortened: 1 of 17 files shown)


tests/ImageSharp.Tests/TestUtilities/FeatureTesting/FeatureTestRunner.cs
```diff
@@ -62,7 +62,7 @@ public static T Deserialize<T>(string value)
             ProcessStartInfo processStartInfo = new();
             if (intrinsic.Key != HwIntrinsics.AllowAll)
             {
-                processStartInfo.Environment[$"COMPlus_{intrinsic.Value}"] = "0";
+                processStartInfo.Environment[$"DOTNET_{intrinsic.Value}"] = "0";
```
gfoidl, line 65: `COMPlus` prefix changed with .NET 6 to be `DOTNET`.
The COMPlus is still supported, but may get obsolete in future versions.

JimBobSquarePants: Thanks for that. One of those things we'd have only found out once all our tests break.

JimBobSquarePants (approved):
I love these kinds of changes. So much better! I'm assuming you want it merged before #2419 ?

PR #2429, opened 2023-04-02
Add AdvSimd in ComponentProcessor

stefannikolei:
### Prerequisites

### Description
Added Arm intrinsics in the ComponentProcessor. 

Only SumHorizontal is not ported to Arm.

### DIFF
--- a/src/ImageSharp/Formats/Jpeg/Components/Encoder/ComponentProcessor.cs
+++ b/src/ImageSharp/Formats/Jpeg/Components/Encoder/ComponentProcessor.cs
@@ -5,6 +5,7 @@ using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.Arm;
 using System.Runtime.Intrinsics.X86;
 using SixLabors.ImageSharp.Memory;
 
@@ -201,10 +202,22 @@ internal class ComponentProcessor : IDisposable
 
                 // Spans are guaranteed to be multiple of 8 so no extra 'remainder' steps are needed
                 nuint count = target.Vector256Count<float>();
-                var multiplierVector = Vector256.Create(multiplier);
+                Vector256<float> multiplierVector = Vector256.Create(multiplier);
                 for (nuint i = 0; i < count; i++)
                 {
                     Unsafe.Add(ref targetVectorRef, i) = Avx.Multiply(Unsafe.Add(ref targetVectorRef, i), multiplierVector);
+                }
+            }
+            else if (AdvSimd.IsSupported)
+            {
+                ref Vector128<float> targetVectorRef = ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(target));
+
+                // Spans are guaranteed to be multiple of 8 so no extra 'remainder' steps are needed
+                nuint count = target.Vector128Count<float>();
+                Vector128<float> multiplierVector = Vector128.Create(multiplier);
+                for (nuint i = 0; i < count; i++)
+                {
+                    Unsafe.Add(ref targetVectorRef, i) = AdvSimd.Multiply(Unsafe.Add(ref targetVectorRef, i), multiplierVector);
                 }
             }
             else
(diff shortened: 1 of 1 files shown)


src/ImageSharp/Formats/Jpeg/Components/Encoder/ComponentProcessor.cs
```diff
@@ -128,6 +129,18 @@ static void SumVertical(Span<float> target, Span<float> source)
                     Unsafe.Add(ref targetVectorRef, i) = Avx.Add(Unsafe.Add(ref targetVectorRef, i), Unsafe.Add(ref sourceVectorRef, i));
                 }
             }
+            else if (AdvSimd.IsSupported)
+            {
+                ref Vector128<float> targetVectorRef = ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(target));
+                ref Vector128<float> sourceVectorRef = ref Unsafe.As<float, Vector128<float>>(ref MemoryMarshal.GetReference(source));
+
+                // Spans are guaranteed to be multiple of 8 so no extra 'remainder' steps are needed
```
gfoidl, line 137: Instead of the comment I'd prefer a Debug.Assert (or the DebugGuard).
* makes it self-documenting
* in case of failure, it crashes (with Debug builds)

br3aker: @stefannikolei if you'd replace comments with asserts - can you also replace them for Avx path please? :)

stefannikolei: Sure. That's was just copy pasted 😅

antonfirsov: We don't use `Debug.Assert`, so if anything it should be `DebugGuard`. I think in this case the comment adds a bit more info, so I wouldn't delete it even if there is an assertion.

JimBobSquarePants:
>Only SumHorizontal is not ported to Arm.

Are you hitting an API wall here?

stefannikolei:
> >Only SumHorizontal is not ported to Arm.
> 
> 
> 
> Are you hitting an API wall here?

I had no clue how to implement Permutate6x64 to Arm

JimBobSquarePants:
> > > Only SumHorizontal is not ported to Arm.
> > 
> > 
> > Are you hitting an API wall here?
> 
> I had no clue how to implement Permute4x64 to Arm

Yeah, that's got me stumped also. @tannergooding is there anything we can do here?

stefannikolei:
> > > > Only SumHorizontal is not ported to Arm.
> 
> > > 
> 
> > > 
> 
> > > Are you hitting an API wall here?
> 
> > 
> 
> > I had no clue how to implement Permute4x64 to Arm
> 
> 
> 
> Yeah, that's got me stumped also. @tannergooding is there anything we can do here?

Another Api which I have no clue how to port is Sse.Shuffle. 

If I knew how to write that in AdvSimd, then I could go forward porting other methods.

JimBobSquarePants:
Thanks Tanner!! That's gonna take me a few reads to get my head around 🤣

tannergooding:
No worries, happy to provide additional suggestions and/or review if needed. Feel free to tag me :)

stefannikolei:
> Depends on exactly how you're permuting the 4x doubles, but you can at worst use 2x VectorTableLookupExtension today.

Am I missing something? VectorTableLookupExtensions only support byte and sbyte. Unfortunately this code (SumHorizontal) is working on floats.

tannergooding:
Yes, because it does things bytewise, much as `Ssse3.Shuffle`. You have to move bytes in groups of `4` if you want to do it for `float`/`int`/`uint`.

That is, if you wanted to pick `3, 2, 1, 0`, you'd use a shuffle mask of `12, 13, 14, 15, 8, 9, 10, 11, 4, 5, 6, 7, 0, 1, 2, 3`. Basically take the `index` scale by `size` (4 in this case), and grab the next `size` entries. -- This assumes you started with `float` and are needing to shuffle as `byte`. If you're instead starting with `bytes` and want `floats`, you may have to account for endianness as well.

JimBobSquarePants:
@stefannikolei Yeah I'm happy for that to be separate. There's a lot of figuring out to do to implement.

>Do you know of any Benchmarks which cover the ComponentProcessor?

Not to my knowledge no.

JimBobSquarePants (approved):

PR #2446, opened 2023-05-04
Adding QOI support

LuisAlfredo92:
### Prerequisites

### Description
First of all, I'm sorry for being so late to add this, I've been extremely busy :(
Also I took some time to see the code to understand it and do it with the same structure

Then, I've add some files to work with qoi, at this moment it only identifies qoi images and returns the ImageInfo, but I want to make sure that I'm doing this correctly before doing the rest of functions. I've followed the Stylecop rules with Visual Studio Intellisense.

The test images were taken from the [official website](https://qoiformat.org/) in the [Test images in QOI and PNG format](https://qoiformat.org/qoi_test_images.zip) link.

This is about #2403 

Could you review it please?


tests/ImageSharp.Benchmarks/Program.cs
LuisAlfredo92, line 1: Ooooohh I forgot to change this one

I'll remove it in the next commit, I add it to test the identification

tests/ImageSharp.Benchmarks/Codecs/Qoi/IdentifyQoi.cs
LuisAlfredo92, line 1: I added a PNG in the tests to check the exceptions

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,94 @@
...
+    public Size Dimensions { get; }
+
+    public Image<TPixel> Decode<TPixel>(BufferedReadStream stream, CancellationToken cancellationToken)
+        where TPixel : unmanaged, IPixel<TPixel> => throw new NotImplementedException();
+
+    public ImageInfo Identify(BufferedReadStream stream, CancellationToken cancellationToken)
+    {
+        ImageMetadata metadata = new();
+
+        byte[] magicBytes = new byte[4], widthBytes = new byte[4], heightBytes = new byte[4];
```
antonfirsov, line 52:
```suggestion
        Span<byte> magicBytes = stackalloc byte[4];
        Span<byte> widthBytes = stackalloc byte[4];
        Span<byte> heightBytes = stackalloc byte[4];
```

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        // Read magic bytes
+        int read = stream.Read(magicBytes);
+        if (read != 4 || !magicBytes.SequenceEqual(QoiConstants.Magic.ToArray()))
+        {
+            throw new InvalidImageContentException("The image is not a QOI image");
```
antonfirsov, line 58:
```suggestion
            throw new InvalidImageContentException("The image is not a valid QOI image.");
```

Since there are 4 more occurrences, it would make sense to refactor this to a helper method:

```C#
[DoesNotReturn]
private static ThrowInvalidImageContentException() => throw new InvalidImageContentException("The image is not a valid QOI image.");
```

tests/ImageSharp.Benchmarks/Codecs/Qoi/IdentifyQoi.cs
```diff
@@ -0,0 +1,30 @@
...
+    private string TestImageFullPath => Path.Combine(TestEnvironment.InputImagesDirectoryFullPath, this.TestImage);
+
+    [Params(TestImages.Qoi.TestCardRGBA, TestImages.Qoi.TestCard, TestImages.Qoi.QoiLogo, TestImages.Qoi.EdgeCase, TestImages.Png.Bike)]
+    public string TestImage { get; set; }
+
+    [GlobalSetup]
+    public void ReadImages() => this.qoiBytes ??= File.ReadAllBytes(this.TestImageFullPath);
+
+    [Benchmark]
+    public ImageInfo Identify()
```
antonfirsov, line 25: This benchmark doesn't give us much info, since there is no baseline to compare to, and IMHO benchmarking Identify is not really valuable in general.

The only benchmark I would find really interesting here is one that compares decoding a larger QOI to the decoding of a PNG (by the system decoder and the ImageSharp PNG decoder). I wouldn't bother with the rest.

antonfirsov (commented):
Since you already implemented `Identify`, your next step is probably to make sure it actually works. Here's an example test, you want an equivalent for Qoi:

https://github.com/SixLabors/ImageSharp/blob/7d0bbbdf9388009c22ad0382a4d49ca8bdb700c0/tests/ImageSharp.Tests/Formats/Png/PngDecoderTests.cs#L338-L355

(I recommend to validate image size and stuff like `QoiChannels` and `QoiColorSpace` there.)

To make sure the format is recognized by `Image.Identify` and `Image.Load` and triggers your decoder, you also need to implement and register a "configuration module":

https://github.com/SixLabors/ImageSharp/blob/7d0bbbdf9388009c22ad0382a4d49ca8bdb700c0/src/ImageSharp/Configuration.cs#L217-L225

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,94 @@
...
+    public DecoderOptions Options { get; }
+
+    public Size Dimensions { get; }
+
+    public Image<TPixel> Decode<TPixel>(BufferedReadStream stream, CancellationToken cancellationToken)
+        where TPixel : unmanaged, IPixel<TPixel> => throw new NotImplementedException();
+
+    public ImageInfo Identify(BufferedReadStream stream, CancellationToken cancellationToken)
+    {
+        ImageMetadata metadata = new();
```
antonfirsov, line 50: You'd need to implement a `QoiMetadata` and expose stuff like `QoiChannels` and `QoiColorSpace` there.

LuisAlfredo92:
Thanks for the reviewing, I'll correct everything and tell when I start working on the other functions 👍

src/ImageSharp/Formats/Qoi/QoiConstants.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Text;
+
+namespace SixLabors.ImageSharp.Formats.Qoi;
+
+internal static class QoiConstants
+{
+    /// <summary>
+    /// Gets the bytes that indicates the image is QOI
+    /// </summary>
+    public static ReadOnlySpan<byte> Magic => Encoding.UTF8.GetBytes("qoif");
```
gfoidl, line 13: This property calls `GetBytes` (and allocates) on every invocation.
For .NET 7 this could be written as
```suggestion
    public static ReadOnlySpan<byte> Magic => "qoif"u8;
```
and before as
```suggestion
    private static readonly byte[] s_magic = Encoding.UTF8.GetBytes("qoif");
    public static ReadOnlySpan<byte> Magic => s_magic;
```

src/ImageSharp/Formats/Qoi/QoiConstants.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    /// <summary>
+    /// The list of mimetypes that equate to a QOI.
+    /// See https://github.com/phoboslab/qoi/issues/167
+    /// </summary>
+    public static readonly IEnumerable<string> MimeTypes = new[] { "image/qoi", "image/x-qoi", "image/vnd.qoi" };
```
gfoidl, line 19: Should the type be `string[]` instead of `IEnumerable<string>`?
I.e. `string[]` allows for easier searching (if needed).

src/ImageSharp/Formats/Qoi/QoiConstants.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    /// <summary>
+    /// The list of file extensions that equate to a QOI.
+    /// </summary>
+    public static readonly IEnumerable<string> FileExtensions = new[] { "qoi" };
```
gfoidl, line 24: Same question.

src/ImageSharp/Formats/Qoi/QoiHeader.cs
```diff
@@ -0,0 +1,45 @@
...
+        this.Width = width;
+        this.Height = height;
+        this.Channels = channels;
+        this.ColorSpace = colorSpace;
+    }
+
+    /// <summary>
+    /// Magic bytes "qoif"
+    /// </summary>
+    public byte[] Magic { get; } = Encoding.UTF8.GetBytes("qoif");
```
gfoidl, line 24:
```suggestion
    public static byte[] Magic { get; } = Encoding.UTF8.GetBytes("qoif");
```

src/ImageSharp/Formats/Qoi/QoiConstants.cs
LuisAlfredo92: We're using .NET 6 as far as I know so I'll go with the second option

src/ImageSharp/Formats/Qoi/QoiConstants.cs
LuisAlfredo92: Well I used the PngConstatns as example so I though it had to be that way, but if there's no problem I'd change it for an array

LuisAlfredo92:
Hello again, I can finally work in this again 🎉 

I've made the changes you told me, however, I didn't find how to run the tests, so I only added the configurations but didn't test them

Soon I'll try to add the Decode and Encode options

antonfirsov:
> I didn't find how to run the tests.

You can use [dotnet test](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-test) from the command line:

```
 dotnet test -f net6.0 --filter FullyQualifiedName~QoiDecoderTests
```

Or if you are using Visual Studio, they should appear in the Test Explorer window.

LuisAlfredo92:
> There are some some StyleCop complaints breaking the build since we use `TreatWarningsAsErrors`: https://github.com/SixLabors/ImageSharp/actions/runs/4885169437/jobs/8735504561?pr=2446
> 
> When you will move on to the actual Decode implementation, you want to base your decoder tests on this:
> 
> https://github.com/SixLabors/ImageSharp/blob/7d0bbbdf9388009c22ad0382a4d49ca8bdb700c0/tests/ImageSharp.Tests/Formats/Pbm/PbmDecoderTests.cs#L81-L99
> 
> First just make sure visually that `DebugSave` is saving the right thing. `CompareToReferenceOutput` would need your reference png to be in the right directory with the right name, we can explain that later.

Hello again, I need some help with the tests

When I try to test the Qoi decoder, when it tries to get the extension, qoi image format never appears:
imagen

However, during debugging, it actually is correctly added:
imagen

So I don't know if I must add something else on the code or tests

Just in case, my qoi configuration module adds only a decoder and format detector
imagen

LuisAlfredo92:
Nevermind, I found where the configuration is done
imagen

JimBobSquarePants:
> Nevermind, I found where the configuration is done imagen

Sorry, didn't see this. Glad you figured it out. Any other questions please let me know!

LuisAlfredo92:
Hello!
I've finished the decoder and it passes all the tests 🎉 I've also taken into consideration the edge case and now it works perfectly, however, I'm not sure if I followed correctly the standards of the project, could you check it please?

I'll be working on the encoder now

LuisAlfredo92:
Hello again
I've finished encoder and decoder

Before adding the extensions (like SaveAsQoi) I'd like to ask for a review to see if I'm not doing something wrong

Both encoder and decoder have passed all the tests

LuisAlfredo92:
Hello again
I've finished encoder, decoder and extensions, and all of them have passed all the tests, but we should discuss some things about the encoder (if it's correct to allow file to start with a QOI_OP_RUN) and decoder (throw an exception when the end of stream doesn't end with seven zero-bytes and one byte with value 1). I wrote some comments if you want to check

I'd like to ask for a review to see if I'm not doing something wrong

JimBobSquarePants:
Amazing! I'll start reviewing ASAP.

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,287 @@
...
+
+        this.ProcessPixels(stream, pixels);
+
+        return image;
+    }
+
+    /// <inheritdoc />
+    public ImageInfo Identify(BufferedReadStream stream, CancellationToken cancellationToken)
+    {
+        ImageMetadata metadata = new();
```
antonfirsov, line 73: If we can efficiently parse out `QoiChannels` and `QoiColorSpace` without reading the pixel data, this should be `QoiMetadata`, and we should do the metadata parsing here. Tests also need to be updated to validate it works.

antonfirsov (commented):
Apologies, I started this review a couple of days ago but forgot to post it.

tests/ImageSharp.Tests/Formats/Qoi/QoiEncoderTests.cs
```diff
@@ -0,0 +1,35 @@
...
+    private static void Encode<TPixel>(TestImageProvider<TPixel> provider)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        using Image<TPixel> image = provider.GetImage();
+        using MemoryStream stream = new();
+        QoiEncoder encoder = new();
+        image.Save(stream, encoder);
+        stream.Position = 0;
+        using Image<TPixel> encodedImage = (Image<TPixel>)Image.Load(stream);
+        ImageComparingUtils.CompareWithReferenceDecoder(provider, encodedImage);
```
antonfirsov, line 33: We could avoid using our own decoder in this test if we do this instead:

```suggestion
        using Image<TPixel> image = provider.GetImage(new MagickReferenceDecoder());
        using MemoryStream stream = new();
        QoiEncoder encoder = new();
        image.Save(stream, encoder);
        stream.Position = 0;
        using Image<TPixel> encodedImage = (Image<TPixel>)Image.Load(stream);
        ImageComparer.Exact.CompareImages(image, encodedImage);
```

Cool that ImageMagick already supports QOI!

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
```diff
@@ -0,0 +1,206 @@
...
+    {
+        // Get metadata
+        Span<byte> width = stackalloc byte[4];
+        Span<byte> height = stackalloc byte[4];
+        BinaryPrimitives.WriteUInt32BigEndian(width, (uint)image.Width);
+        BinaryPrimitives.WriteUInt32BigEndian(height, (uint)image.Height);
+        QoiChannels qoiChannels = image.PixelType.BitsPerPixel == 24 ? QoiChannels.Rgb : QoiChannels.Rgba;
+
+        // I need to check this, how do I check it with the pixel type or metadata of the original image?
+        const QoiColorSpace qoiColorSpace = QoiColorSpace.SrgbWithLinearAlpha;
```
antonfirsov, line 45: We probably want to make this configurable by properties exposed on `QoiEncoder`. Check `PngEncoder` as an example,  eg:

https://github.com/SixLabors/ImageSharp/blob/2738b3ab5793bddb96f3a5317d46afa0529220ce/src/ImageSharp/Formats/Png/PngEncoderCore.cs#L1257

antonfirsov (commented):
Great job so far!

tests/ImageSharp.Tests/TestImages.cs
```diff
@@ -1040,4 +1040,16 @@ public static class Pbm
...
+    public static class Qoi
+    {
+        public const string Dice = "Qoi/dice.qoi";
+        public const string EdgeCase = "Qoi/edgecase.qoi";
+        public const string Kodim10 = "Qoi/kodim10.qoi";
+        public const string Kodim23 = "Qoi/kodim23.qoi";
+        public const string QoiLogo = "Qoi/qoi_logo.qoi";
+        public const string TestCard = "Qoi/testcard.qoi";
+        public const string TestCardRGBA = "Qoi/testcard_rgba.qoi";
+        public const string Wikipedia008 = "Qoi/wikipedia_008.qoi";
```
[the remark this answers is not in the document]
LuisAlfredo92: Ooh yeah, it's malformed, I'll check the encoder then

It should look like this
Decode_Rgba32_wikipedia_008

tests/ImageSharp.Tests/Formats/Qoi/QoiEncoderTests.cs
LuisAlfredo92: I've just did it but it still passes all the tests, the wikipedia_008.qoi image doesn't throw an error

I'm doing this on windows, maybe OS is involved?

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
LuisAlfredo92: I've done the changes, however, now `WriteHeader(Image image, Stream stream)` method can't be static but `WritePixels` and `WriteEndOfStream` are, isn't there any problem with that?

tests/ImageSharp.Tests/Formats/Qoi/QoiEncoderTests.cs
antonfirsov: It fails for me on Windows, and it also looks wrong with the viewer I linked in https://github.com/SixLabors/ImageSharp/pull/2446#discussion_r1246840717. Is it possible that the image you have locally differs from the one [you pushed to the repo](https://github.com/SixLabors/ImageSharp/blob/6245666d61cfc2d79981b81806b3d253d9bc8b07/tests/Images/Input/Qoi/wikipedia_008.qoi) for some reason?

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
antonfirsov: No, that's an implementation detail and it's fine.

tests/ImageSharp.Tests/Formats/Qoi/QoiEncoderTests.cs
LuisAlfredo92: > Is it possible that the image you have locally differs from the one [you pushed to the repo](https://github.com/SixLabors/ImageSharp/blob/6245666d61cfc2d79981b81806b3d253d9bc8b07/tests/Images/Input/Qoi/wikipedia_008.qoi) for some reason?

Omg
I've just done some some tests and yeah, **it's malformed**
imagen

It's 120 bytes smaller than the original file, I'll search how do re-upload a file

tocsoft: could be a https://github.com/SixLabors/ImageSharp/blob/main/.gitattributes thing...if git is treating like text then it might be trying to replace newlines etc.

LuisAlfredo92: > could be a https://github.com/SixLabors/ImageSharp/blob/main/.gitattributes thing...if git is treating like text then it might be trying to replace newlines etc.

Thanks!
It looks like it's fixed now 😄

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,287 @@
...
+        // See https://github.com/phoboslab/qoi/issues/258
+        int pixelArrayPosition = GetArrayPosition(previousPixel);
+        previouslySeenPixels[pixelArrayPosition] = previousPixel;
+
+        for (int i = 0; i < this.header.Height; i++)
+        {
+            for (int j = 0; j < this.header.Width; j++)
+            {
+                byte operationByte = (byte)stream.ReadByte();
+                byte[] pixelBytes;
```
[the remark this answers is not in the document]
antonfirsov: Next level trick would be to create the span around the `readPixel` so it points to the memory area of the `readPixel` variable:

```
Rgba32 readPixel = default;
Span<byte> pixelBytes = MemoryMarshal.CreateSpan(ref Unsafe.As<Rgba32, byte>(ref readPixel), 4);
```

This way you could remove the following call in the RGBA branch would look like:

```diff
- readPixel = new Rgba32(pixelBytes[0], pixelBytes[1], pixelBytes[2], pixelBytes[3]);
```

While the RGB branch would simplify to this if I'm not mistaken:

```diff
- readPixel = previousPixel with { R = pixelBytes[0], G = pixelBytes[1], B = pixelBytes[2] };
+ readPixel.A = previousPixel.A;
```

But I haven't tested this second idea, you should apply it with care and know what's happening!

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,287 @@
...
+            }
+        }
+
+        if (stream.ReadByte() != 1)
+        {
+            ThrowInvalidImageContentException();
+        }
+    }
+
+    private static int GetArrayPosition(Rgba32 pixel) => ((pixel.R * 3) + (pixel.G * 5) + (pixel.B * 7) + (pixel.A * 11)) % 64;
```
antonfirsov, line 286:
```suggestion
    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    private static int GetArrayPosition(Rgba32 pixel) => (((pixel.R * 3) + (pixel.G * 5) + (pixel.B * 7) + (pixel.A * 11))) & 11;
```
Even better if you add a `Modulo64` method to our `Numerics` class.

Formula: https://stackoverflow.com/a/19760152

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,287 @@
...
+                                    currentBlue = (byte)((diffBlueDG - 8 + (diffGreen - 32) + previousPixel.B) % 256);
+                                readPixel = previousPixel with { R = currentRed, B = currentBlue, G = currentGreen };
+                                pixel.FromRgba32(readPixel);
+                                pixelArrayPosition = GetArrayPosition(readPixel);
+                                previouslySeenPixels[pixelArrayPosition] = readPixel;
+                                break;
+
+                            // Repeating the previous pixel 1..63 times
+                            case QoiChunk.QoiOpRun:
+                                byte repetitions = (byte)(operationByte & 0b00111111);
```
antonfirsov, line 236: It's better to leave this be an `int`:

```suggestion
                                int repetitions = operationByte & 0b00111111;
```

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,287 @@
...
+                            // Get one pixel from the difference (-2..1) of the previous pixel
+                            case QoiChunk.QoiOpDiff:
+                                byte redDifference = (byte)((operationByte & 0b00110000) >> 4),
+                                    greenDifference = (byte)((operationByte & 0b00001100) >> 2),
+                                    blueDifference = (byte)(operationByte & 0b00000011);
+                                readPixel = previousPixel with
+                                {
+                                    R = (byte)((previousPixel.R + (redDifference - 2)) % 256),
+                                    G = (byte)((previousPixel.G + (greenDifference - 2)) % 256),
+                                    B = (byte)((previousPixel.B + (blueDifference - 2)) % 256)
```
antonfirsov, line 211: You can apply the modulo trick I commented later. `mod 256` should be equivalent to `& 15˙ if I'm not mistaken. Would be good to have a `Numerics.Modulo256` method.

antonfirsov (commented):
Some optimization ideas for the Decoder. Eliminating the per-operation allocation is a must-have, apply the rest depending on the time you are fine spending on the PR. Haven't looked at the encoder yet.

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,287 @@
...
+
+                            default:
+                                ThrowInvalidImageContentException();
+                                return;
+                        }
+
+                        break;
+                }
+
+                pixels[j, i] = pixel;
```
antonfirsov, line 266: Avoid use this indexer!
Take `Span<TPixel> row = pixels.DangerousGetRowSpan(i);` in the outer for loop. Then you can do `row[j] = pixel`.

Same for the encoder!

tests/ImageSharp.Tests/Formats/Qoi/QoiEncoderTests.cs
JimBobSquarePants: You'll need to do this fix upstream at https://github.com/SixLabors/SharedInfrastructure

Once that is merged then you would update the submodule 

```
git submodule foreach git pull origin main
```

Otherwise, every time you run a local build the `.gitattributes` file is overwritten by the shared one.

antonfirsov: Opened https://github.com/SixLabors/SharedInfrastructure/pull/39.

JimBobSquarePants: And merged.

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
LuisAlfredo92: Isn't it `& 63` to do a modulo 64?

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
LuisAlfredo92: Something funny is that initially, when creating both functions, I used the `[j, i]` indexer because I didn't understand `pixels.DangerousGetRowSpan(i);`, but now that I've been so deep into the project, I wonder why I didn't do that before xD

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
LuisAlfredo92: I'm sorry I haven't done this, I don't understand what do I have to do

Do I create an `ImageMetadata` in `QoiDecoder`? Or an `QoiMetadata`?
Or `QoiChannels` and `QoiColorSpace` variables in the `QoiDecoder`?

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
JimBobSquarePants: Yeah

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
JimBobSquarePants: You would create and assign `QoiMetadata` in your decoder then add that to the `ImageMetadata`

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
JimBobSquarePants: Even I do that sometimes.

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
antonfirsov: Sorry, for some reason I thought `QoiMetadata` is a subclass of `ImageMetaData`, so my comment was super confusing. 

This is how it actually works:
`ImageMetaData` is a container of format-specific metadata objects. The decoder is expected to fill the metadata object ( in your case `QoiMetadata`) belonging to the format in both `Decode` and `Identify`. You can ask `ImageMetaData.GetQoiMetadata()` to "get or create" your format-specific metadata object which you can then fill from the the data you parse out to `QoiHeader`. Here's how PNG creates its' metadata object:

https://github.com/SixLabors/ImageSharp/blob/2738b3ab5793bddb96f3a5317d46afa0529220ce/src/ImageSharp/Formats/Png/PngDecoderCore.cs#L147-L148

tests/ImageSharp.Tests/Formats/Qoi/QoiEncoderTests.cs
JimBobSquarePants: @LuisAlfredo92 You need to run this command now in your branch to update the submodule

https://github.com/SixLabors/ImageSharp/pull/2446#discussion_r1247237463

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,291 @@
...
+        byte operationByte;
+        Rgba32 readPixel = default;
+        Span<byte> pixelBytes = MemoryMarshal.CreateSpan(ref Unsafe.As<Rgba32, byte>(ref readPixel), 4);
+        TPixel pixel = default;
+
+        for (int i = 0; i < this.header.Height; i++)
+        {
+            for (int j = 0; j < this.header.Width; j++)
+            {
+                Span<TPixel> row = pixels.DangerousGetRowSpan(i);
```
antonfirsov, line 165: Querying the row span for each pixel actually performs worse than the indexer. I assume you are querying it here and not in the outer loop because of the `QoiOpRun` logic.  Can't you instead "forward" the `row` variable in that branch so you don't penalize the other cases?

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    [DoesNotReturn]
+    private static void ThrowInvalidImageContentException()
+        => throw new InvalidImageContentException("The image is not a valid QOI image.");
+
+    private void ProcessPixels<TPixel>(BufferedReadStream stream, Buffer2D<TPixel> pixels)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        Rgba32[] previouslySeenPixels = new Rgba32[64];
```
antonfirsov, line 149: I though I already posted this:

```suggestion
        using IMemoryOwner<Rgba32> previouslySeenPixelsBuffer = this.memoryAllocator.Allocate<Rgba32>(64);
        Span<Rgba32> previouslySeenPixels = previouslySeenPixelsBuffer.GetSpan();
```

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
LuisAlfredo92: Oh no no no, my bad, I didn't move it to the correct line

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
LuisAlfredo92: I guess this won't be possible

The reference says that the array must be zero-initialized, and I though there was no problem since pixels will be overwritten but there's this error:

Due to `memoryAllocator`, the array isn't zero-initialized
imagen

And if an decoded image starts with (0, 0, 0, 0) color (like `qoi_logo.qoi`), and it's encoded to qoi, then the first byte will be a 0b00000000, that means it's an QOI_OP_INDEX that requires the first position ([0]) of the previously seen pixels, but this won't be that color
imagen

If the array is zero-initialized, then it's okay, it takes a (0, 0, 0, 0) color, however, using this implementation takes garbage of the previous decoding, so we must set everything to zero

I tried to use `previouslySeenPixelsBuffer.Dispose();` but it doesn't clear the array for the next decoding, is there any more efficient way to create the array? Or do I roll-back to `Rgba32[] previouslySeenPixels = new Rgba32[64];`?

LuisAlfredo92:
Okay, I've done everything you've asked

Is there anything else I should do?

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        read = stream.Read(heightBytes);
+        if (read != 4)
+        {
+            ThrowInvalidImageContentException();
+        }
+
+        // These numbers are in Big Endian so we have to reverse them to get the real number
+        uint width = BinaryPrimitives.ReadUInt32BigEndian(widthBytes),
+            height = BinaryPrimitives.ReadUInt32BigEndian(heightBytes);
```
antonfirsov, line 116: I guess there is no rule for this in StyleCop , but we don't use the multi-assignment in declarations.

```suggestion
        uint width = BinaryPrimitives.ReadUInt32BigEndian(widthBytes);
        uint height = BinaryPrimitives.ReadUInt32BigEndian(heightBytes);
```

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,290 @@
...
+                            case QoiChunk.QoiOpIndex:
+                                readPixel = previouslySeenPixels[operationByte];
+                                pixel.FromRgba32(readPixel);
+                                break;
+
+                            // Get one pixel from the difference (-2..1) of the previous pixel
+                            case QoiChunk.QoiOpDiff:
+                                byte redDifference = (byte)((operationByte & 0b00110000) >> 4),
+                                    greenDifference = (byte)((operationByte & 0b00001100) >> 2),
+                                    blueDifference = (byte)(operationByte & 0b00000011);
```
antonfirsov, line 206:
```suggestion
                                byte redDifference = (byte)((operationByte & 0b00110000) >> 4);
                                byte greenDifference = (byte)((operationByte & 0b00001100) >> 2);
                                byte blueDifference = (byte)(operationByte & 0b00000011);
```

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,290 @@
...
+                            // Get green difference in 6 bits and red and blue differences
+                            // depending on the green one
+                            case QoiChunk.QoiOpLuma:
+                                byte diffGreen = (byte)(operationByte & 0b00111111),
+                                    currentGreen = (byte)Numerics.Modulo256(previousPixel.G + (diffGreen - 32)),
+                                    nextByte = (byte)stream.ReadByte(),
+                                    diffRedDG = (byte)(nextByte >> 4),
+                                    diffBlueDG = (byte)(nextByte & 0b00001111),
+                                    currentRed = (byte)Numerics.Modulo256(diffRedDG - 8 + (diffGreen - 32) + previousPixel.R),
+                                    currentBlue = (byte)Numerics.Modulo256(diffBlueDG - 8 + (diffGreen - 32) + previousPixel.B);
```
antonfirsov, line 227:
```suggestion
                                byte diffGreen = (byte)(operationByte & 0b00111111),
                                byte currentGreen = (byte)Numerics.Modulo256(previousPixel.G + (diffGreen - 32));
                                byte nextByte = (byte)stream.ReadByte();
                                byte diffRedDG = (byte)(nextByte >> 4);
                                byte diffBlueDG = (byte)(nextByte & 0b00001111);
                                byte currentRed = (byte)Numerics.Modulo256(diffRedDG - 8 + (diffGreen - 32) + previousPixel.R);
                                byte currentBlue = (byte)Numerics.Modulo256(diffBlueDG - 8 + (diffGreen - 32) + previousPixel.B);
```

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,290 @@
...
+                $"The image has an invalid size: width = {width}, height = {height}");
+        }
+
+        int channels = stream.ReadByte();
+        if (channels is -1 or (not 3 and not 4))
+        {
+            ThrowInvalidImageContentException();
+        }
+
+        PixelTypeInfo pixelType = new(8 * channels);
```
antonfirsov, line 129: This seems to be unused.

src/ImageSharp/Formats/Qoi/QoiDecoderCore.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        // Create Image object
+        ImageMetadata metadata = new()
+        {
+            DecodedImageFormat = QoiFormat.Instance,
+            HorizontalResolution = this.header.Width,
+            VerticalResolution = this.header.Height,
+            ResolutionUnits = PixelResolutionUnit.AspectRatio
+        };
+        Image<TPixel> image = new(this.configuration, (int)this.header.Width, (int)this.header.Height, metadata);
```
antonfirsov, line 59: This `metadata` object also needs a `QoiMetadata` to be added.

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
```diff
@@ -0,0 +1,222 @@
...
+                }
+                else
+                {
+                    // else, we check if the difference is less than -2..1
+                    // Since it wasn't found on the previously seen pixels, we save it
+                    previouslySeenPixels[pixelArrayPosition] = currentRgba32;
+
+                    sbyte diffRed = (sbyte)(currentRgba32.R - previousPixel.R),
+                        diffGreen = (sbyte)(currentRgba32.G - previousPixel.G),
+                        diffBlue = (sbyte)(currentRgba32.B - previousPixel.B);
```
antonfirsov, line 146: Same in encoder regarding variable assignments.

tests/ImageSharp.Tests/Formats/Qoi/QoiDecoderTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    [Theory]
+    [WithFile(TestImages.Qoi.Dice, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.EdgeCase, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.Kodim10, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.Kodim23, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.QoiLogo, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.TestCard, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.TestCardRGBA, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.Wikipedia008, PixelTypes.Rgba32)]
+    public void Decode<TPixel>(TestImageProvider<TPixel> provider)
```
antonfirsov, line 41: We need at least a few test cases to validate `QoiMetadata` for the encoder & decoder.

tests/ImageSharp.Tests/Formats/Qoi/QoiEncoderTests.cs
```diff
@@ -0,0 +1,36 @@
...
+    [WithFile(TestImages.Qoi.Dice, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.EdgeCase, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.Kodim10, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.Kodim23, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.QoiLogo, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.TestCard, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.TestCardRGBA, PixelTypes.Rgba32)]
+    [WithFile(TestImages.Qoi.Wikipedia008, PixelTypes.Rgba32)]
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("CodeQuality", "IDE0051:Quitar miembros privados no utilizados", Justification = "Function implicitly in tests")]
+    private static void Encode<TPixel>(TestImageProvider<TPixel> provider)
```
antonfirsov, line 25:
```suggestion
    public static void Encode<TPixel>(TestImageProvider<TPixel> provider)
```

antonfirsov (commented):
I see some optimization opportunities in JpegEncoder too, but focusing on functionality first.

tests/ImageSharp.Tests/Formats/Qoi/QoiDecoderTests.cs
LuisAlfredo92: I'm not sure how could we determine the channels and color space when encoding image. They are purely informative and they don't affect how pixels are encoded, so there's no way to know if an image is RGA or RGBA

I though we could check if any pixel alpha is different to 255, then the channels will be RGBA, but some images (like `edgecase.qoi` or `testcard.qoi`) are fully opaque but their header say they're RGBA

Also I'm not sure how to create the `QoiMetadata` inside the decoder, so I though encoding every image with _RGBA_ channels (value 4 in enum) and _sRGB with linear alpha_ (value 0 in enum) since most images have that metadata

LuisAlfredo92: Researching, I found https://github.com/phoboslab/qoi/issues/37, I guess it can give us more information since those bytes are purely informative but I didn't find how to decide what value it'll have

So I guess we could use the `PixelType` from the provider to decide it, what do you think?

antonfirsov: You can do a separate round-trip test: encode then re-decode with ImageSharp qoi codecs, and check the resulting `image.Metadata.GetQoiMetadata()`. This would test the encoder+decoder `QoiMetadata` handling functionality at the same time, so also address https://github.com/SixLabors/ImageSharp/pull/2446#discussion_r1251050456.

antonfirsov: Just noticed you did this already in [ef78f98](https://github.com/SixLabors/ImageSharp/pull/2446/commits/ef78f98eb3d686638f9da867f8896b286a722e10), so nevermind :)

LuisAlfredo92: So that was correct?
I though we couldn't use any of the data from the tests :o

tests/ImageSharp.Tests/Formats/Qoi/QoiEncoderTests.cs
```diff
@@ -0,0 +1,36 @@
...
+using SixLabors.ImageSharp.Formats.Qoi;
+using SixLabors.ImageSharp.PixelFormats;
+using SixLabors.ImageSharp.Tests.TestUtilities.ImageComparison;
+using SixLabors.ImageSharp.Tests.TestUtilities.ReferenceCodecs;
+
+namespace SixLabors.ImageSharp.Tests.Formats.Qoi;
+
+[Trait("Format", "Qoi")]
+[ValidateDisposedMemoryAllocations]
+public class QoiEncoderTests
```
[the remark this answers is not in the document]
LuisAlfredo92: The custom `QoiEncoder` would be `MagicReferenceDecoder` like in `DecoderTests`?

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
```diff
@@ -0,0 +1,222 @@
...
+        Buffer2D<TPixel> pixels = image.Frames[0].PixelBuffer;
+
+        for (int i = 0; i < pixels.Height; i++)
+        {
+            Span<TPixel> row = pixels.DangerousGetRowSpan(i);
+            for (int j = 0; j < row.Length && i < pixels.Height; j++)
+            {
+                // We get the RGBA value from pixels
+                TPixel currentPixel = row[j];
+                currentPixel.ToRgba32(ref currentRgba32);
```
antonfirsov, line 86: These conversions can be done much faster in batches thanks to SIMD. A pattern we commonly do is to convert the entire image row into a temporary working buffer. You would need to do something like:

```C#
        using IMemoryOwner<Rgba32> rgbaRowBuffer = this.memoryAllocator.Allocate<Rgba32>(pixels.Width);
        Span<Rgba32> rgbaRow = rgbaRowBuffer.GetSpan();

        for (int i = 0; i < pixels.Height; i++)
        {
            Span<TPixel> row = pixels.DangerousGetRowSpan(i);
            PixelOperations<TPixel>.Instance.ToRgba32(this.PASS_CONFIGURATION_TO_THE_ENCODER, row, rgbaRow);
            for (int j = 0; j < row.Length && i < pixels.Height; j++)
            {
                currentRgba32 = rgbaRow[j];
```

And also update `rgbaRow` in the QOI_OP_RUN.

tests/ImageSharp.Benchmarks/Codecs/Qoi/IdentifyQoi.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using BenchmarkDotNet.Attributes;
+using SixLabors.ImageSharp.Formats;
+using SixLabors.ImageSharp.Formats.Qoi;
+using SixLabors.ImageSharp.Tests;
+
+namespace SixLabors.ImageSharp.Benchmarks.Codecs.Qoi;
+
+[Config(typeof(Config.ShortMultiFramework))]
+public class IdentifyQoi
```
antonfirsov, line 12: Let's delete this benchmark since it doesn't really compare anything.

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
```diff
@@ -0,0 +1,222 @@
...
+                    // If so, we do a QOI_OP_DIFF
+                    if (diffRed is >= -2 and <= 1 &&
+                        diffGreen is >= -2 and <= 1 &&
+                        diffBlue is >= -2 and <= 1 &&
+                        currentRgba32.A == previousPixel.A)
+                    {
+                        // Bottom limit is -2, so we add 2 to make it equal to 0
+                        byte dr = (byte)(diffRed + 2);
+                        byte dg = (byte)(diffGreen + 2);
+                        byte db = (byte)(diffBlue + 2);
```
antonfirsov, line 157: Unless you need to clamp the numbers, can't you `int` instead of `sbyte` and `byte` until the point when you actually need to write a `byte` to the stream or a assign it to a value within a pixel? That would get rid of the conversions, making things faster.

This suggestion applies to all code in both the Decoder & Encoder.

antonfirsov (commented):
Almost there! A few more suggestions for perf.

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
LuisAlfredo92: Oh okay, I though I had to save memory as much as I could, but yeah, I also prefer use ints to avoid conversions xD

tests/ImageSharp.Tests/Formats/Qoi/QoiEncoderTests.cs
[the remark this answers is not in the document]
antonfirsov: By "custom QoiEncoder" I meant 

```C#
QoiEncoder encoder = new()
        {
            Channels = channels,
            ColorSpace = colorSpace
        };
```
This is good now with the latest changes.

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
antonfirsov: > I had to save memory

The JIT will map these variables to CPU registers. Memory is a concern when you work with larger buffers, not small local variables.

LuisAlfredo92:
Changes done
I'm pretty exited to see these changes merged into the main project 😄

src/ImageSharp/Formats/ImageExtensions.Save.cs
```diff
@@ -9,6 +9,7 @@
 using SixLabors.ImageSharp.Formats.Jpeg;
 using SixLabors.ImageSharp.Formats.Pbm;
 using SixLabors.ImageSharp.Formats.Png;
+using SixLabors.ImageSharp.Formats.Qoi;
```
antonfirsov, line 12: One last important thing I almost missed ...

This file is auto generated. If you change & save the `ImageExtensions.Save.tt` in Visual Studio, it will generate `ImageExtensions.Save.cs` for you. There is an extension for VS Code and Rider as well.

Wanted to push the necessary changes for you, but looks like I don't have the rights. You can just grab this commit if you want: https://github.com/SixLabors/ImageSharp/commit/2700f8b40e2841fad3e615380cc7213255727981

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
```diff
@@ -0,0 +1,230 @@
...
+using System.Runtime.CompilerServices;
+using SixLabors.ImageSharp.Memory;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Formats.Qoi;
+
+/// <summary>
+/// Image encoder for writing an image to a stream as a QOi image
+/// </summary>
+public class QoiEncoderCore : IImageEncoderInternals
```
antonfirsov, line 15:
```suggestion
internal class QoiEncoderCore : IImageEncoderInternals
```

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
```diff
@@ -0,0 +1,230 @@
...
+    /// The configuration instance for the encoding operation.
+    /// </summary>
+    private readonly Configuration configuration;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="QoiEncoderCore"/> class.
+    /// </summary>
+    /// <param name="encoder">The encoder with options.</param>
+    /// <param name="memoryAllocator">The <see cref="MemoryAllocator" /> to use for buffer allocations.</param>
+    public QoiEncoderCore(QoiEncoder encoder, MemoryAllocator memoryAllocator, Configuration configuration)
```
antonfirsov, line 37: If you run a release build locally, it should catch these:

https://github.com/SixLabors/ImageSharp/actions/runs/5449320272/jobs/9936168476?pr=2446#step:13:67

src/ImageSharp/Formats/ImageExtensions.Save.cs
LuisAlfredo92: Sorry, I only forked it :(

I'll add all the changes you told me in one commit because I'm not sure how to get that commit into my fork, so I'll add it manually but thanks

src/ImageSharp/Formats/Qoi/QoiEncoderCore.cs
LuisAlfredo92: Ooh yeah, I forgot to check StyleCop on Release

LuisAlfredo92:
Everything's done 🎉

LuisAlfredo92:
Is there anything else I could do?

src/ImageSharp/Formats/Qoi/QoiMetadata.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public QoiMetadata()
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="QoiMetadata"/> class.
+    /// </summary>
+    /// <param name="other">The metadata to create an instance from.</param>
+    public QoiMetadata(QoiMetadata other)
```
antonfirsov, line 22:
```suggestion
    private QoiMetadata(QoiMetadata other)
```

antonfirsov (approved):
Looks good to merge now. Thanks a lot for the nice work, and really appreciate the patience for the long review!

LuisAlfredo92:
Thanks to you all to review the changes so constantly and I'm sorry it took so long, I posted the idea when I was about to get back to school and I was able to do it until semester finished xD

JimBobSquarePants:
Adding an entirely new image format is no small task! Really well done and thanks so much for your efforts!

LuisAlfredo92:
When you publish the 3.1.0 version, could I add ImageSharp it to the [phoboslab/qoi](https://github.com/phoboslab/qoi) repository README? Maybe on tools or implementations

antonfirsov:
Yeah I also wanted to call out that we need to get into that hall of fame after the release.

PR #2453, opened 2023-05-11
bugfix in Rational.cs, when both Nominator and Denominator equals to 0

tsardaryanCamenAI:
### Prerequisites

### Description
In cases, when you're trying to create Rational.FromDouble(0) it generates an invalid Rational item.

### DIFF
--- a/src/ImageSharp/Primitives/LongRational.cs
+++ b/src/ImageSharp/Primitives/LongRational.cs
@@ -131,6 +131,11 @@ internal readonly struct LongRational : IEquatable<LongRational>
     /// <param name="bestPrecision">Whether to use the best possible precision when parsing the value.</param>
     public static LongRational FromDouble(double value, bool bestPrecision)
     {
+        if (value == 0.0)
+        {
+            return new LongRational(0, 1);
+        }
+
         if (double.IsNaN(value))
         {
             return new LongRational(0, 0);
@@ -152,10 +157,6 @@ internal readonly struct LongRational : IEquatable<LongRational>
         double val = Math.Abs(value);
         double df = numerator / (double)denominator;
         double epsilon = bestPrecision ? double.Epsilon : .000001;
-
-        if(val < epsilon) {
-            return new LongRational(0, 1);
-        }
 
         while (Math.Abs(df - val) > epsilon)
         {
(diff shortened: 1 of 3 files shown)


src/ImageSharp/Primitives/Rational.cs
```diff
@@ -74,6 +74,11 @@ public Rational(double value, bool bestPrecision)
 
         this.Numerator = (uint)rational.Numerator;
         this.Denominator = (uint)rational.Denominator;
+        
+        if(this.Numerator == 0 && this.Denominator == 0)
```
JimBobSquarePants, line 78: If there's an issue here it will have to be fixed in [`LongRational`](https://github.com/SixLabors/ImageSharp/blob/7d0bbbdf9388009c22ad0382a4d49ca8bdb700c0/src/ImageSharp/Primitives/LongRational.cs#L132)

tsardaryanCamenAI: if you want to update GPS location of ExifInformation, it accepts Retional[]. That's why I suggest this change. Anyway if you are creating Rational.FromDouble(0) Denominator shouldn't be 0.

dlemstra: That's the correct fix. Can you also add some unit tests for this to make sure this stays fixed?

tsardaryanCamenAI: yes, sure will do that and will push all as a different pull request. You can close this one.

JimBobSquarePants: No need to open a new PR. That’s just confusing. Push to this one please.

dlemstra: And you can also use a force push to update this PR with a new set of commits.

src/ImageSharp/Primitives/LongRational.cs
```diff
@@ -153,6 +153,10 @@ public static LongRational FromDouble(double value, bool bestPrecision)
         double df = numerator / (double)denominator;
         double epsilon = bestPrecision ? double.Epsilon : .000001;
 
+        if(val < epsilon) {
```
JimBobSquarePants, line 156: Shouldn't the fix be the following inserted at 149?

```c#
if (value == 0)
{
    return new LongRational(0, 1);
}
```

Issue #2404, opened 2023-03-17
Support for palette access when reading indexed image formats


Paul-JanPauptit:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

3.0.0

### Other ImageSharp packages and versions

-

### Environment (Operating system, version and so on)

Windows, any

### .NET Framework version

6.

### Description

I am processing 4-bpp indexed images (16 colors) for retro & fantasy platform development (TIC80 in particular). For that we often need access to the original palette of the image. 

I understand this is not a primary use case for ImageSharp, but it would be great if ImageSharp offered just that... and like @JimBobSquarePants mentioned on twitter, "why not submit an issue". 😄

### Steps to Reproduce

Load an indexed image. Try to access the palette.

JimBobSquarePants:
@Paul-JanPauptit What format are your images in?

Completed
- [x] gif
- [x] png

For future
- [x] bmp
- [x] tiff
- [x] ico
- [x] cur

Paul-JanPauptit:
@JimBobSquarePants I am currently in a bit of crunch so my demoscene toolbuilding is on hold for a couple of weeks, but I wanted to say that I really appreciate that you have been adding this functionality. 

I don't have a set workflow yet, I think centering it around PNG should work perfectly. Basically being able to read out the palette and pixels should solve all my problems.

JimBobSquarePants:
Happy to help!

I'll eventually port bitmap and maybe tiff across to do it also but will leave that for future releases as it gets complicated.

Beyley:
Would this also extend to storing the palette used by a quantize operation on an Image for later access?

Im trying to write a custom ImageSharp encoder for the LittleBigPlanet PSP custom texture format "MIP", which uses an 8bit indexed pixel format. currently to my knowledge, after i run a quantize with a "max colors" of 256, i need to loop over every pixel to fill out a palette myself, it would be nice if after doing this
![image](https://github.com/SixLabors/ImageSharp/assets/34501060/b4708297-aa74-45e6-8a91-f802d4939ae5)
[...]

PR #2455, opened 2023-05-15
Preserve Gif color palettes and deduplicate frame pixels.

### LINKED ISSUE #635: GIF Encoder: Exploit transparency

### LINKED ISSUE #2198: Loading a GIF results in a bigger file

### LINKED ISSUE #2450: Resizing animated GIFs causes quality loss after 3.0.0
### Prerequisites

### ImageSharp version

3.0.1

None

Windows 11

### .NET Framework version

7

### Description

Using ImageSharp to resize animated GIFs produces different results on 2.1.4 versus 3.0.0 and 3.0.1.
The output size is much smaller on 3.x, but the quality is a lot worse.

### Steps to Reproduce

This is the sample code I've used to generate the samples:

```csharp
var image = Image.Load("input.gif");

image.Mutate(x =>
{
    x.Resize(250, 0);
});

image.Save("resized.gif");
```

### Images

Original:
input

Resized using 2.1.4:
resized-2 1 4

Resized using 3.0.1:
resized-3 0 1

JimBobSquarePants:
### Prerequisites

### Description

This PR does a few things. 
1. Reverts an issue introduced in #2289 where a naive attempt to reduce encoded output led to a loss of image data when gifs have been previously optimized.
2. Introduces a new SIMD accelerated (_Avx2, Sse2, Arm64_) algorithm that deduplifies the frame index buffers following quantization to allow much better encoder compression.
3. Introduces a necessary breaking change to the gif metadata and encoder (_Next release will be v3.1_) that allows taking advantage of these optimizations. Without these changes some gifs could encode at nearly 7x the input size using the default settings.

Fixes #635
Fixes #2450
Fixes #2198 
Touches #2404 

The new deduplifier does a comparison of the quantized pixel indexes using the previous and current frame. If duplicates are detected in the current frame, then the indices are replaced with the indez of the background or transparent color. Given the increased repeats this enables better compression.

This functionality is limited to gifs encoded using global color palettes as we would require extra work to detect duplicates in local color tables and produce a map of index offsets which could be cumbersome (not to say this might come in a following PR in the future) .

Screenshot 2023-05-20 000820


JimBobSquarePants:
> An API diff would be nice (for both the PR and the release notes) to make it obvious what specific API-s are we breaking.

Yep aggreed. I'll get on that.

JimBobSquarePants:
@SixLabors/core Would it be possible to get some eyes on this please?

antonfirsov:
Really hoping to be able to take a look later this week.

JimBobSquarePants:
> Really hoping to be able to take a look later this week.

Thanks!

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -332,7 +344,7 @@ private void ReadApplicationExtension(BufferedReadStream stream)
             if (subBlockSize == GifConstants.NetscapeLoopingSubBlockSize)
             {
                 stream.Read(this.buffer.Span, 0, GifConstants.NetscapeLoopingSubBlockSize);
-                this.gifMetadata!.RepeatCount = GifNetscapeLoopingApplicationExtension.Parse(this.buffer.Span.Slice(1)).RepeatCount;
+                this.gifMetadata!.RepeatCount = GifNetscapeLoopingApplicationExtension.Parse(this.buffer.Span[1..]).RepeatCount;
```
gfoidl, line 347:
```suggestion
                this.gifMetadata!.RepeatCount = GifNetscapeLoopingApplicationExtension.Parse(this.buffer.Span.Slice(1)).RepeatCount;
```
Perf-wise `Slice` is [better](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunHiUGAMQgQAFAGUADtgB2AHmABPDDAB8DXNJkBKLj07UeRjVoYBeE7IB04gDYBLMDFHltAbj3cAvp4a/6TAIAQthQEloKymqWOr4GxjyasuYxANrkVlYAuh6GPD7UXkA==) than the range-operator. Unfortunately Roslyn compiles "bad" IL here, which the JIT doesn't optimize away.

So either
* keep Slice
* wait for Roslyn to be fixed (which takes some years already for that issue)

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -412,16 +577,27 @@ private static void WriteCommentSubBlock(Stream stream, ReadOnlySpan<char> comme
...
     /// <param name="transparencyIndex">The index of the color in the color palette to make transparent.</param>
     /// <param name="stream">The stream to write to.</param>
-    private void WriteGraphicalControlExtension(GifFrameMetadata metadata, int transparencyIndex, Stream stream)
+    private void WriteGraphicalControlExtension(GifFrameMetadata? metadata, int transparencyIndex, Stream stream)
     {
+        bool hasTransparency;
+        if (metadata is null)
+        {
+            metadata = new();
+            hasTransparency = transparencyIndex > -1;
```
gfoidl, line 586:
```suggestion
            hasTransparency = transparencyIndex >= 0;
```
Comparisons with `0` a tad faster and allows the cpu itself to optimize (fuse) the instruction.
As the change is simple and code as readable as the current one, I'd do such a change.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -629,6 +629,33 @@ private static Vector256<byte> ShuffleMaskShiftAlpha() => Vector256.Create((byte
...
+        /// The high bit of each corresponding <paramref name="mask"/> byte determines the selection.
+        /// If the high bit is set the element of <paramref name="left"/> is selected.
+        /// The element of <paramref name="right"/> is selected otherwise.
+        /// </summary>
+        /// <param name="left">The left vector.</param>
+        /// <param name="right">The right vector.</param>
+        /// <param name="mask">The mask vector.</param>
+        /// <returns>The <see cref="Vector256{T}"/>.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Vector128<byte> BlendVariable(in Vector128<byte> left, in Vector128<byte> right, in Vector128<byte> mask)
```
gfoidl, line 643:
```suggestion
        public static Vector128<byte> BlendVariable(Vector128<byte> left, Vector128<byte> right, Vector128<byte> mask)
```
`in` modifier is equivalent to `readonly ref`, so the vectors can't be passed in registers, rather they need to be passed via memory which is slower.
Due the _AggressiveInlining_ this method is likely to be inlined, so the JIT can figure out that no passing via memory is needed and can it can pass these vectors via registers -- but: that's not guaranteed, and the JIT has more (unnecessary) work to do.

I see this similar to a method like `int Add(int a, int b);` where one can assume that `a`, and `b` are passed by value and via registers at cpu-level. No need to put some emphasis on the read-only nature of `a`, and `b`. Some goes for vectors.

---

Side note: starting with .NET 7 this could just be `Vector128.ConditionalSelect` (I know we have a .NET 6 target, but I mention it for completness)

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
JimBobSquarePants: I'm just taking the hit here as we have elsewhere. It's annoying but worth the readability improvements.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
JimBobSquarePants: Done

JimBobSquarePants:
@gfoidl While I have you attention I'm wondering if you have any ideas here. 

I want to do a single-pass top-down inspection of the 2D buffer to determine the smallest rectangle we can use to represent the frame by excluding all outer pixels that match the background color. 

With scalar code this is easy enough but I'm wondering if it would be possible to accelerate it somehow using `BitOperations.LeadingZeroCount` to allow me to look for the 1st and last instance in a single per-row run.

Any ideas?

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
JimBobSquarePants: Done

gfoidl:
Ey, here I'm missing some background knowledge, so I can't follow along 100%. Can you please show me a test, sample, or the scalar code on what should be done? Then I can look into and play around with a vectorized solution.
(We can also do this in a separate issue, just tag me and I'll have a look).

gfoidl:
Thanks for the extra information, now I see what should be done. I've put that on my todo list, and hope that next week I can give this a stab.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -170,88 +227,299 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream, CancellationToken
...
+            PixelOperations<TPixel>.Instance.ToVector4(this.configuration, backgroundFrame.DangerousGetPixelRowMemory(y).Span, background, PixelConversionModifiers.Scale);
+            PixelOperations<TPixel>.Instance.ToVector4(this.configuration, sourceFrame.DangerousGetPixelRowMemory(y).Span, source, PixelConversionModifiers.Scale);
+
+            ref Vector256<float> backgroundBase = ref Unsafe.As<Vector4, Vector256<float>>(ref MemoryMarshal.GetReference(background));
+            ref Vector256<float> sourceBase = ref Unsafe.As<Vector4, Vector256<float>>(ref MemoryMarshal.GetReference(source));
+            ref Vector256<float> resultBase = ref Unsafe.As<Vector4, Vector256<float>>(ref MemoryMarshal.GetReference(result));
+
+            uint x = 0;
+            int remaining = background.Length;
+            if (Avx.IsSupported && remaining >= 2)
```
JimBobSquarePants, line 367:
```suggestion
            if (Axv2.IsSupported && remaining >= 2)
```

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -170,88 +227,299 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream, CancellationToken
...
+            PixelOperations<TPixel>.Instance.ToVector4(this.configuration, backgroundFrame.DangerousGetPixelRowMemory(y).Span, background, PixelConversionModifiers.Scale);
+            PixelOperations<TPixel>.Instance.ToVector4(this.configuration, sourceFrame.DangerousGetPixelRowMemory(y).Span, source, PixelConversionModifiers.Scale);
+
+            ref Vector256<float> backgroundBase = ref Unsafe.As<Vector4, Vector256<float>>(ref MemoryMarshal.GetReference(background));
+            ref Vector256<float> sourceBase = ref Unsafe.As<Vector4, Vector256<float>>(ref MemoryMarshal.GetReference(source));
+            ref Vector256<float> resultBase = ref Unsafe.As<Vector4, Vector256<float>>(ref MemoryMarshal.GetReference(result));
+
+            uint x = 0;
+            int remaining = background.Length;
+            if (Axv2.IsSupported && remaining >= 2)
```
JimBobSquarePants, line 367:
```suggestion
            if (Avx2.IsSupported && remaining >= 2)
```

gfoidl:
@JimBobSquarePants FYI regarding

> I've put that on my todo list, and hope that next week I can give this a stab.

Go a business work-item which I put above this one in my todo 😉 
Hoping that I can look into this at the end of the week.

JimBobSquarePants:
> @JimBobSquarePants FYI regarding
> 
> > I've put that on my todo list, and hope that next week I can give this a stab.
> 
> Go a business work-item which I put above this one in my todo 😉 Hoping that I can look into this at the end of the week.

Thanks.. I think I've simplified it enough to maybe make it possible.

JimBobSquarePants:
253 of the changed files here are reference images. If I could get someone to have a look at the actual code, I'd really appreciate it.

gfoidl:
My take on vectorizing the TrimTransparentPixels is in https://github.com/SixLabors/ImageSharp/pull/2500 (to be merged into this PR / branch).

JimBobSquarePants:
@SixLabors/core Can I please get some eyes on this? I'd like to keep moving as I have a huge backlog of work to do.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -170,88 +227,418 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream, CancellationToken
...
+    private void EncodeFirstFrame<TPixel>(
+        Stream stream,
+        GifFrameMetadata? metadata,
+        IndexedImageFrame<TPixel> quantized,
+        int transparencyIndex)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        this.WriteGraphicalControlExtension(metadata, transparencyIndex, stream);
+
+        Buffer2DRegion<byte> region = ((IPixelSource)quantized).PixelBuffer.GetRegion();
```
antonfirsov, line 239: Nit: do we need the region here? I don't see us slicing down the full buffer.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
antonfirsov: I don't think this is a critical path, simpler looking code is better.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -682,21 +701,30 @@ private void SetFrameMetadata(ImageFrameMetadata metadata)
...
         {
             GifFrameMetadata gifMeta = metadata.GetGifMetadata();
             gifMeta.ColorTableMode = GifColorTableMode.Local;
-            gifMeta.ColorTableLength = this.imageDescriptor.LocalColorTableSize;
+
+            Color[] colorTable = new Color[this.imageDescriptor.LocalColorTableSize];
+            ref Rgb24 localBase = ref MemoryMarshal.GetReference(MemoryMarshal.Cast<byte, Rgb24>(this.currentLocalColorTable!.GetSpan()[..this.currentLocalColorTableSize]));
+            for (int i = 0; i < colorTable.Length; i++)
+            {
+                colorTable[i] = new Color(Unsafe.Add(ref localBase, (uint)i));
```
antonfirsov, line 716: Since this is not a critical path, I wouldn't go unsafe here and just use span indexers.

src/ImageSharp/Formats/Gif/GifMetadata.cs
```diff
@@ -45,9 +50,15 @@ private GifMetadata(GifMetadata other)
     public GifColorTableMode ColorTableMode { get; set; }
 
     /// <summary>
-    /// Gets or sets the length of the global color table if present.
+    /// Gets or sets the global color table, if any.
+    /// </summary>
+    public ReadOnlyMemory<Color>? GlobalColorTable { get; set; }
```
antonfirsov, line 55: Is color resolution different from 24 bits ever expected according to the gif standard? If not, wouldn't it be better to go with `ReadOnlyMemory<Rgb24>` and avoiding the conversions?

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -91,16 +105,43 @@ public void Clear(ReadOnlyMemory<TPixel> palette)
...
+            }
+            else
+            {
+                Unsafe.SkipInit(out TPixel pixel);
+                pixel.FromScaledVector4(Vector4.Zero);
+                match = pixel;
+            }
+
+            return index;
+        }
```
antonfirsov, line 139: This is now penalizing all quantizers and decoders while only gif is using `transparentIndex`. Can't we  keep the transparency detection logic in `GifDecoderCore`?

antonfirsov (commented):
A couple of recommendations and a few general comments.

src/ImageSharp/Formats/Png/Filters/PaethFilter.cs
```diff
@@ -50,7 +50,7 @@ public static void Decode(Span<byte> scanline, Span<byte> previousScanline, int
     }
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah, we need to start reviewing this. Need good tooling to inspect the codegen though.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -629,6 +629,33 @@ private static Vector256<byte> ShuffleMaskShiftAlpha() => Vector256.Create((byte
...
+                return Sse41.BlendVariable(left, right, mask);
+            }
+            else if (Sse2.IsSupported)
+            {
+                return Sse2.Or(Sse2.And(right, mask), Sse2.AndNot(mask, left));
+            }
+
+            // Use a signed shift right to create a mask with the sign bit.
+            Vector128<short> signedMask = AdvSimd.ShiftRightArithmetic(mask.AsInt16(), 7);
+            return AdvSimd.BitwiseSelect(signedMask, right.AsInt16(), left.AsInt16()).AsByte();
```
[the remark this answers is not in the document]
JimBobSquarePants: This has become a bit of a dumping ground over the years. My plan for v4 is a mass simplification of intrinsics utilizing things like `Vector128<T>` to allow us to implement better cross chipset approaches. We can do a tidy up at the same time.

src/ImageSharp/Formats/Gif/GifMetadata.cs
```diff
@@ -45,9 +50,15 @@ private GifMetadata(GifMetadata other)
...
+
+    /// <summary>
+    /// Gets or sets the index at the <see cref="GlobalColorTable"/> for the background color.
+    /// The background color is the color used for those pixels on the screen that are not covered by an image.
     /// </summary>
-    public int GlobalColorTableLength { get; set; }
+    public byte BackgroundColor { get; set; }
```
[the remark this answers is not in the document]
JimBobSquarePants: Yes, much better, will change.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
JimBobSquarePants: `WriteImageData` requires a region as we slice down the buffer. I'll rewrite it though so it accepts a buffer and a rectangle.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
JimBobSquarePants: I just wanted to avoid the secondary indexer on the raw `Rgb24` span.

src/ImageSharp/Formats/Gif/GifMetadata.cs
JimBobSquarePants: It doesn't and I did consider using `Rgb24` originally but thought that `Color` gave us greater portability on the API surface. For example, someone could now easily create a pallet quantizer from this color table.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
JimBobSquarePants: I couldn't figure out a way to do it. Given that the slow code is actually the loop through the palette I thought this addition check outside the loop would be ok. 

I actually think Bmp will end up using it eventually when we finally expose a decoded palette there.

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
antonfirsov: But is there a measurable perf gain from avoiding the indexer here? This method is only called once per frame on a relatively small (<256) set of data. On the other hand, each usage of `Unsafe` reduces readability and may introduce a serious bug in case we accidentally create a buffer overflow as we refactor the code in the future. IMO usage of `Unsafe`  should be always justified (=hottest hot path) instead of being the default way of indexing things.

src/ImageSharp/Formats/Gif/GifMetadata.cs
antonfirsov: If we think such use-cases will be common it might be justified, but then I would extend the xml doc describing that `GlobalColorTable` is 24bit. For me it still feels better to expose the raw gif data as-is, and instead provide helper API-s to make work with `PaletteQuantizer` easier.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
antonfirsov: "Slow" actually means we run it every time a new value is added to `ColorDistanceCache` which should be still quite common for an image with a wide original palette. I also don't see an easy way to optimize this, we should probably introduce multiple variants of `EuclideanPixelMap<TPixel>` and `PalettQuantizer<TPixel>` in the future.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -91,16 +105,43 @@ public void Clear(ReadOnlyMemory<TPixel> palette)
...
         int index = 0;
         float leastDistance = float.MaxValue;
+
+        if (this.transparentIndex >= 0 && rgba == default)
+        {
+            // We have explicit instructions. No need to search.
+            index = this.transparentIndex;
+            this.cache.Add(rgba, (byte)index);
+
+            if (index >= 0 && index < this.Palette.Length)
```
antonfirsov, line 127: Given that in this branch `index == this.transparentIndex >= 0`, the `index >= 0` condition seems to be unnecessary.

Regarding `index < this.Palette.Length`, shouldn't we make this an invariant rule instead, and `Guard` it in the constructor and `SetTransparentIndex`?

JimBobSquarePants: I'm not sure I follow. If you have time, could you push to the branch?

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
JimBobSquarePants: I've been thinking about this and compared to the actual iteration the `>=0` check should be barely detectable. The comparison operator won't even kick in.

antonfirsov: The comparison is indeed cheap, the problem is not with that. We have a lot of code added to the method  which may lead to more frequent L1 instruction cache misses. Branches can also hurt CPU pipelining.

Since `GetQuantizedColor` is being called per-pixel, the method should be as slim as possible, ideally containing no branching except the pixelmap lookup. We should be able to achieve this (and greatly reduce the memory footprint for pixel types without an alpha!) by separating different `OctreeQuantizer<T>` setups to separate classes.

If we could look into this together with optimization opportunities in `ErrorDitherer`, we could achieve significant improvements in the quantization space. Not a small amount of work though.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
antonfirsov: Will do it later this week.

JimBobSquarePants: Thanks!

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
JimBobSquarePants: Yeah true. I'll not touch this until you have a chance to make changes. We can look at greater optimization opportunities for v4

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
antonfirsov: b48dbc5087de9bd98dc8ed0f478550e22d4e1ff9

JimBobSquarePants: Ah yeah, that's much better! Not sure what I was thinking when I wrote mine! 🤣

JimBobSquarePants:
@antonfirsov I think I've covered everything now.

antonfirsov (approved):
LGTM, thanks for the patience!

PR #2468, opened 2023-05-30
Update Directory.Build.targets

Pashkinz92:
Updated SharpZip package version to  1.4.2

There is malware

SharpZipLib (or #ziplib) is a Zip, GZip, Tar and BZip2 library. Prior to version 1.3.3, a TAR file entry `../evil.txt` may be extracted in the parent directory of `destFolder`. This leads to arbitrary file write that may lead to code execution. The vulnerability was patched in version 1.3.3.


Pashkinz92:
Of course :-) I Sorry, I missed that requirement

PR #2471, opened 2023-06-07
Fix #2467 bmp encoding issue for BMP with 1 bit per pixel and more pixels per row than divisible by 8.

### LINKED ISSUE #2467: BMP with 1 bitcount (B/W) corrupts after save
### Prerequisites

### ImageSharp version

3.0.1 

>&gt;= 2.0.0 (tested up to 3.0.2-alpha.0.17)

Ubuntu-latest & Windows-latest

### .NET Framework version

7.0.0

### Description

I have a bmp file that is 10x10 pixels. The bmp file has a bitcount of 1, and thus a color pallet of 2 colors, namely: FFFFFF & 000000.

When loading the bmp, the pixeldata is loaded correctly, but when saving the bmp, the image somehow becomes corrupt.

When you save the image by first copying the pixeldata into a new `Image` object, the resulting image is correct. Indicating that the bug has to do with the original metadata messing somehow with the encoding process.

The error bug occurs since ImageSharp version 2.0.0.

Input image:
Original image
original-large

Corrupt image:
Corrupt image
corrupt-large

Test project: https://github.com/synercoder/BitmapBug/blob/main/ImageSharp.BmpBug/BmpSavedWrong.cs
Test run showing matrix of os, mode & ImageSharp versions: https://github.com/synercoder/BitmapBug/actions/runs/5094509469

### Steps to Reproduce

Use the attached image (`datamatrix.bmp`), and execute the following code:
```
using (var dmImage = Image.Load<Rgb24>("datamatrix.bmp"))
using (var fs = File.OpenWrite("corrupt.bmp")))
{
    dmImage.SaveAsBmp(fs);
}
```

The resulting `corrupt.bmp` is visibly (zoom in, it is 10x10 pixels) different on the right side of the image.

### Images

[datamatrix.zip](https://github.com/SixLabors/ImageSharp/files/11579625/datamatrix.zip)

synercoder:
### Prerequisites

### Description
Fix for issue #2467.

If you have a BMP with 1 bit per pixel, and more pixels per row than divisible by 8, the last byte of the row will need to be padded with extra zeros.

So if you have a black/white image of 10 pixels in a row, and the row looks like: 
`bwwbbbwbww`
The bytes should look like:"
`0x01100010 0x11000000`

However, the current implementation of Write1BitPixelData will when it reaches the last byte looks like:
```C#
if (quantizedPixelRow.Length % 8 != 0)
{
    int startIdx = quantizedPixelRow.Length - 7;
    endIdx = quantizedPixelRow.Length;
    Write1BitPalette(stream, startIdx, endIdx, quantizedPixelRow);
}
```
So `quantizedPixelRow.Length` will be 10, resulting in a `startIdx` of 10-7 = 3
It will then write pixels to the byte starting at index 3 again, meaning that the last bytes looks like:
`0x00010110`, **the last byte will always contain the last 7 pixels, instead of the amount of pixels remaining**.

This results in that when the image in opened again, the row bytes looks like:
`0x01100010 0x00010110`

If you check the images in #2467, you can see that the corrupted image shows column 4 & 5 again in column 9 & 10.

### The fix
The fix is to read the last X pixels from the `quantizedPixelRow` when X equals the remaining pixels instead of always the last 7 pixels. So the `startIdx` needs to be adjusted:
From:
`int startIdx = quantizedPixelRow.Length - 7;`
To: 
`int startIdx = quantizedPixelRow.Length - (quantizedPixelRow.Length % 8);`

### Small note
I am unsure of my test name, so I named it after my issue. I hope this is fine.

### DIFF
--- a/src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
+++ b/src/ImageSharp/Formats/Bmp/BmpEncoderCore.cs
@@ -668,7 +668,7 @@ internal sealed class BmpEncoderCore : IImageEncoderInternals
 
             if (quantizedPixelRow.Length % 8 != 0)
             {
-                int startIdx = quantizedPixelRow.Length - 7;
+                int startIdx = quantizedPixelRow.Length - (quantizedPixelRow.Length % 8);
                 endIdx = quantizedPixelRow.Length;
                 Write1BitPalette(stream, startIdx, endIdx, quantizedPixelRow);
             }
--- a/tests/ImageSharp.Tests/Formats/Bmp/BmpEncoderTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Bmp/BmpEncoderTests.cs
@@ -370,6 +370,30 @@ public class BmpEncoderTests
         TestBmpEncoderCore(provider, bitsPerPixel);
     }
 
+    [Theory]
+    [WithFile(BlackWhitePalletDataMatrix, PixelTypes.Rgb24, BmpBitsPerPixel.Pixel1)]
+    public void Encode_Issue2467<TPixel>(TestImageProvider<TPixel> provider, BmpBitsPerPixel bitsPerPixel)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        using Image<TPixel> image = provider.GetImage();
+
+        using var reencodedStream = new MemoryStream();
+        var encoder = new BmpEncoder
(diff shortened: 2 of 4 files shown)


JimBobSquarePants (approved):
Legend!

I would have been wasting time looking for issues in the quantizer. Thanks for providing a fix.

PR #2473, opened 2023-06-12
Use more accurate bit for alpha lookup.

### LINKED ISSUE #2469: Unexpected artefacts when quantizing images
### Prerequisites

### ImageSharp version

3.0.1

NA

Windows, Linux

### .NET Framework version

6

### Description

We are currently experiencing strange artifacts when quantizing images.

### Steps to Reproduce

Load image.
Quantize the image using KnownQuantizers.Wu
The saved output will have unexpected artifacts.

Below is a test program I created to replicate the problem. It does not happen on all images only on some. I have attached a image that it happens on. If you compare the original to the quantized version you will note a grey triangular artifact from right of centre to the right middle of the image.

```

using (var i = Image.Load("Orig.png"))
{
    i.Mutate(x => x.Quantize(KnownQuantizers.Wu));
    await i.SaveAsPngAsync("Resized.png", new PngEncoder()
    {
        CompressionLevel = PngCompressionLevel.BestCompression,
    });
}
```
Original Image
Orig
Quantized Image
Resized

### Images

Orig
Resized

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2469 

Our pixel map accuracy wasn't good enough for alpha comparison, so we were leaving artifacts.

This increases memory from 2.26MB to 4.45MB but not only do we fix our issue, as a result of the increased accuracy, we're also getting less collisions in our RGB components.


src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -149,13 +149,13 @@ private static int DistanceSquared(Rgba32 a, Rgba32 b)
...
     /// providing enough match accuracy.
-    /// Entry count is currently limited to 1185921 entries (2371842 bytes ~2.26MB).
+    /// Entry count is currently limited to 2335905 entries (4671810 bytes ~4.45MB).
     /// </para>
     /// </remarks>
     private unsafe struct ColorDistanceCache : IDisposable
     {
         private const int IndexBits = 5;
-        private const int IndexAlphaBits = 5;
+        private const int IndexAlphaBits = 6;
```
antonfirsov, line 158: This could be zero when `TPixel` doesn't have an alpha. It might be worth to eventually implement some trick to have two versions of EuclideanPixelMap, so we save memory when working with RGB pngs and gifs.

JimBobSquarePants: Good point. We could easily preserve much more accurate matches in less space.

PR #2481, opened 2023-06-21
Fix for B&W pbm images with width not dividable by 8

### LINKED ISSUE #2477: Images saved using Pbm Binary Byte format (P4) are corrupted.
### Prerequisites

### ImageSharp version

3.0.1

3.0.1

Windows 10

### .NET Framework version

6.0

### Description

Images saved using Pbm Binary Byte format (P4) are corrupted. 
It seems like the file written is too short and if you try to open it in GIMP it will say it's corrupted.
Reloading the saved file and saving to a different format you can see it is visibly corrupted. 
I have tried a few different images and they all have the same issue. 

### Steps to Reproduce

`
var image = Image.Load("input.png");
var encoder = new PbmEncoder()
{
    ColorType = PbmColorType.BlackAndWhite,
    ComponentType = PbmComponentType.Byte,
    Encoding = PbmEncoding.Binary
};
image.Save("output.pbm", encoder);
var output = Image.Load("output.pbm");
output.SaveAsPng("output.png");

`

### Images

input
output
[output.zip](https://github.com/SixLabors/ImageSharp/files/11785960/output.zip)
[outputImageMagick.zip](https://github.com/SixLabors/ImageSharp/files/11786038/outputImageMagick.zip)

ynse01:
### Prerequisites

### Description
This fixes #2477 by handling Black & White binary images with a width not dividable by 8 correctly. The handling of the padding bits was incorrect and not according to the specification.

Thanks for reporting this bug !

### DIFF
--- a/src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
+++ b/src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
@@ -152,7 +152,6 @@ internal class BinaryDecoder
     {
         int width = pixels.Width;
         int height = pixels.Height;
-        int startBit = 0;
         MemoryAllocator allocator = configuration.MemoryAllocator;
         using IMemoryOwner<L8> row = allocator.Allocate<L8>(width);
         Span<L8> rowSpan = row.GetSpan();
@@ -162,23 +161,11 @@ internal class BinaryDecoder
             for (int x = 0; x < width;)
             {
                 int raw = stream.ReadByte();
-                int bit = startBit;
-                startBit = 0;
-                for (; bit < 8; bit++)
+                for (int bit = 0; bit < 8; bit++)
                 {
                     bool bitValue = (raw & (0x80 >> bit)) != 0;
                     rowSpan[x] = bitValue ? black : white;
                     x++;
-                    if (x == width)
-                    {
-                        startBit = (bit + 1) & 7; // Round off to below 8.
-                        if (startBit != 0)
-                        {
-                            stream.Seek(-1, System.IO.SeekOrigin.Current);
-                        }
-
-                        break;
-                    }
                 }
             }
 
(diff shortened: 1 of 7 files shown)


src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
```diff
@@ -162,23 +161,12 @@ private static void ProcessBlackAndWhite<TPixel>(Configuration configuration, Bu
             for (int x = 0; x < width;)
             {
                 int raw = stream.ReadByte();
-                int bit = startBit;
-                startBit = 0;
-                for (; bit < 8; bit++)
+                int stopBit = Math.Min(8, width - x);
```
JimBobSquarePants, line 164: Should be a nice perf boost here also.

JimBobSquarePants:
Thanks so much for doing this @ynse01 !

The code looks great, however it doesn't look like the reference image was included in GIT LFS for some reason. 

```basjh
git lfs track --filename [file path]
```
Should do the trick.

I'd do it for you, but LFS doesn't work well with GitHub when pushing a file against a fork.

src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
ynse01: Indeed, solving a bug by simplifying the code

ynse01:
> I'd do it for you, but LFS doesn't work well with GitHub when pushing a file against a fork.

Please do so, as I tried the command and as a result I got modification in the _.gitattributes_ file as a result. Doesn't look desirable, as the file extensions are already mentioned, so I didn't commit those modifications.

JimBobSquarePants:
I just pulled your branch down and cut the file to see what git was saying. It looks like the GitHub UI was playing silly buggers because the image is definitely already tracked.

image

This looks great then. Thanks for your help!

JimBobSquarePants (approved):

Issue #2451, opened 2023-05-09
"System.NotSupportedException: ccitt extensions are not supported" when loading some tifs


joaovsoares:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

3.0.2-alpha.0.14

### Other ImageSharp packages and versions

No

### Environment (Operating system, version and so on)

Linux

### .NET Framework version

NET 6

### Description

When loading some Tiffs, we received this exception while decoding.
Other windows applications can open the file.

BTW: I have bought a Commercial License (couldnt find the checkbox to inform).

```
System.NotSupportedException: ccitt extensions are not supported.
   at SixLabors.ImageSharp.Formats.Tiff.Compression.Decompressors.T6TiffCompression.Decode2DScanline(T6BitReader bitReader, Boolean whiteIsZero, CcittReferenceScanline referenceScanline, Span`1 scanline)
   at SixLabors.ImageSharp.Formats.Tiff.Compression.Decompressors.T6TiffCompression.Decompress(BufferedReadStream stream, Int32 byteCount, Int32 stripHeight, Span`1 buffer, CancellationToken cancellationToken)
   at SixLabors.ImageSharp.Formats.Tiff.TiffDecoderCore.DecodeStripsChunky[TPixel](ImageFrame`1 frame, Int32 rowsPerStrip, Span`1 stripOffsets, Span`1 stripByteCounts, CancellationToken cancellationToken)
   at SixLabors.ImageSharp.Formats.Tiff.TiffDecoderCore.DecodeImageWithStrips[TPixel](ExifProfile tags, ImageFrame`1 frame, CancellationToken cancellationToken)
   at SixLabors.ImageSharp.Formats.Tiff.TiffDecoderCore.DecodeFrame[TPixel](ExifProfile tags, CancellationToken cancellationToken)
   at SixLabors.ImageSharp.Formats.Tiff.TiffDecoderCore.Decode[TPixel](BufferedReadStream stream, CancellationToken cancellationToken)
[...]

PR #2483, opened 2023-06-24
Tiff: ccitt compression, fix for issue #2451

brianpopow:
### Prerequisites

- [ ] I have provided test coverage for my change (where applicable) (no test image available, which can be used publicly)

### Description

This PR fixes #2541 as discussed [here](https://github.com/SixLabors/ImageSharp/issues/2451#issuecomment-1544618277):
If we encounter a Extensions1D or 2D code, we read the next 5 bits and if it turns out to be a EOL and not a Extensions1D or 2D, we return that. Otherwise throw a NotSupported Exception for the Extension1D/2D code.

### DIFF
--- a/src/ImageSharp/Formats/Tiff/Compression/Decompressors/T6BitReader.cs
+++ b/src/ImageSharp/Formats/Tiff/Compression/Decompressors/T6BitReader.cs
@@ -127,7 +127,7 @@ internal sealed class T6BitReader : T4BitReader
                         this.Code = Len7Code0000000;
 
                         // We do not support Extensions1D codes, but some encoders (scanner from epson) write a premature EOL code,
-                        // which at this point cannot be distinguished from a distinguish, because we read the data bit by bit.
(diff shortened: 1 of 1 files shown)


src/ImageSharp/Formats/Tiff/Compression/Decompressors/T6BitReader.cs
```diff
@@ -125,13 +125,29 @@ public bool ReadNextCodeWord()
                     if (value == Len7Code0000000.Code)
                     {
                         this.Code = Len7Code0000000;
-                        return false;
+
+                        // We do not support Extensions1D codes, but some encoders (scanner from epson) write a premature EOL code,
+                        // which at this point cannot be distinguished from a distinguish, because we read the data bit by bit.
```
JimBobSquarePants, line 130:
```suggestion
                        // which at this point cannot be distinguished from the marker, because we read the data bit by bit.
```

JimBobSquarePants (approved):
Thanks for doing this. 👍

PR #2485, opened 2023-06-26
Expose and conserve the color palette for indexed png images.

JimBobSquarePants:
### Prerequisites

### Description

**This builds upon #2455 and should be merged after.**

Changes the following in `PngMetadata`:
- Exposes a new `ReadonlyMemory<Color> ColorTable` property.
- Merges `TransparentRgb24`, `TransparentRgb48`, `TransparentL8` and, `TransparentL16` into a new nullable `TransParentColor Color?` property.
- Removes `HasTransparency` since we now know that from the palette or transparent color.

On index Png encode, as with Gif if a palette is found in the metadata the quantizer will use that instead.


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -924,10 +919,44 @@ private void ProcessInterlacedDefilteredScanline<TPixel>(ReadOnlySpan<byte> defi
...
+        }
+
+        if (alpha.Length > 0)
+        {
+            // The alpha chunk may contain as many transparency entries as there are palette entries
+            // (more than that would not make any sense) or as few as one.
+            for (int i = 0; i < alpha.Length; i++)
+            {
+                ref Color color = ref colorTable[i];
+                color = color.WithAlpha(alpha[i] / 255F);
```
gfoidl, line 949: If perf is important here, then
```suggestion
                color = color.WithAlpha(alpha[i] * (1F / 255F));
```
will produce a [mul instead of div](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunGAzADYRsGBgFlyACgCWAO1HYAlAwC8APgbYGAegakArPr4BuLj3oNBw0WNIz5m5es0MAVA0nk+OvYb6LTagBfIA==).

Even if perf isn't critical here I think that I'd change the code, as
* it's almost as readable
* it's not only about the perf of this piece of code, but rather for the whole system, as the cpu-ports can be utilized better when there's no division clogging it

Note: it's Roslyn to blame here for not being able to emit `1 / 255F` automatically -- though the JIT could do that optimization too, but it has more stringent timing constraints...

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -875,7 +875,7 @@ private void WriteGammaChunk(Stream stream)
             // 4-byte unsigned integer of gamma * 100,000.
             uint gammaValue = (uint)(this.gamma * 100_000F);
 
-            BinaryPrimitives.WriteUInt32BigEndian(this.chunkDataBuffer.Span.Slice(0, 4), gammaValue);
+            BinaryPrimitives.WriteUInt32BigEndian(this.chunkDataBuffer.Span[..4], gammaValue);
```
gfoidl, line 878: FYI: the `Slice` here was IIRC intentional, as it produces better code than the slice. There are issues in Roslyn and runtime to care about that, but for .NET 8 these aren't resolved (when I'm not mistaken).

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -1175,7 +1177,7 @@ private void WriteChunk(Stream stream, PngChunkType type, Span<byte> data, int o
 
         stream.Write(buffer);
 
-        uint crc = Crc32.Calculate(buffer.Slice(4)); // Write the type buffer
+        uint crc = Crc32.Calculate(buffer[4..]); // Write the type buffer
```
gfoidl, line 1180: Same here.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
saucecontrol: This would not be a valid optimization for either compiler to make, as the results are not the same (due to 1/255f not being precisely representable in float).  See https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoIgBgAIiyA6AGQEs8BHAbmqIDMjEvQDC9AN7V6M4SKlVZS+gDNoMAIZgAFvQAUnDPQBu9TvQCieHAFsYUDcAA2MZgCUNeAOYw9tNPQkAKxBAJShktLK0ewq+qYA9IEhcQCEALwm9ABU+mQJwUEq4VHRZYxkAJx6ACQARBLGAL4gkonJRU30GW05eQUpoU11oXyKZU2l9JNUTUA===

For that reason, unless it's a hot path, this optimization loses precision for no measurable benefit.

gfoidl: > as the results are not the same

Yep, but the error is in the region of 1e-8 or less ([demo](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoIgBgAIiyA6AGQEs8BHAbmqIDMjEvQDC9AN7V6M4SKlVZS+gDNoMAIZgAFvQAUnDPQBu9TvQCieHAFsYUDcAA2MZgCUNeAOYw9tNPQkAKxBAJShktLK0ewq+qYA9IEhcQCEALwm9ABU+mQJwUEq4VHRZYxkAJx6ACQARBLGAL4gkonJRU30GW05eQUpoU0AOhiN9EmFcXBZuXr5U0N1oXyKZU2l9BtUTUA===))

saucecontrol: In that value range, yes, but the magnitude of the error is relative to the magnitude of the value -- that's how floating point works.  Whether the precision loss is acceptable in any given case is a question of the trade-off between performance (~12 cycles for `divss` vs ~4 for `mulss`) and precision.

I'm not familiar enough with the code to know whether that trade-off is worthwhile here (though it looks questionable).  Mostly, I was commenting on your suggestion that Roslyn and RyuJIT were somehow missing an optimization one of them could have made automatically, which is incorrect.

gfoidl: > which is incorrect

You are right (and I have to apologize here by Roslyn).
I had a look what GCC does (before writing the first comment on this) and it emitted the `mul` -- but I had `-Ofast` compiler switch set which disregards strict standards compliance (as I re-read just now). With `-O{2,3}` it's a `div`.

Thanks for the heads up!

src/ImageSharp/Formats/Png/PngEncoderCore.cs
JimBobSquarePants: We're just taking the hit here. One day Roslyn will catch up, but we need greater consistency to ease the (_massive_) maintenance burden. I'm hoping come v4/.NET8 with the SIMD migration improvements I have planned we won't notice codegen issues.

JimBobSquarePants:
@antonfirsov The changes here are based upon the changes made to the gif format API. If you have a chance, I'd love to get your feedback please.

JimBobSquarePants:
@SixLabors/core  Can I get a review here please.

JimBobSquarePants:
> @JimBobSquarePants will do it by EOW if that's good enough.

Thanks! Appreciate it.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -924,10 +919,44 @@ private void ProcessInterlacedDefilteredScanline<TPixel>(ReadOnlySpan<byte> defi
...
+        {
+            return;
+        }
+
+        Color[] colorTable = new Color[palette.Length / Unsafe.SizeOf<Rgb24>()];
+        ReadOnlySpan<Rgb24> rgbTable = MemoryMarshal.Cast<byte, Rgb24>(palette);
+        for (int i = 0; i < colorTable.Length; i++)
+        {
+            colorTable[i] = new Color(rgbTable[i]);
+        }
```
antonfirsov, line 940: Nit: I see several loops like this one in codecs now. We can maybe define a `public static Color.FromPixel<TPixel>(Span<TPixel>, Span<Color>)`  analogous to the bulk `ToPixel<TPixel>(...)` method.

tests/ImageSharp.Tests/Formats/Png/PngEncoderTests.cs
```diff
@@ -449,44 +449,15 @@ public void Encode_PreserveTrns(string imagePath, PngBitDepth pngBitDepth, PngCo
...
+        Assert.True(inMeta.TransparentColor.HasValue);
 
         using MemoryStream memStream = new();
         input.Save(memStream, PngEncoder);
         memStream.Position = 0;
         using Image<Rgba32> output = Image.Load<Rgba32>(memStream);
         PngMetadata outMeta = output.Metadata.GetPngMetadata();
-        Assert.True(outMeta.HasTransparency);
-
-        switch (pngColorType)
```
antonfirsov, line 461: We should validate `metadata.PngBitDepth/PngColorType` against the expected vaules `pngBitDepth` & `pngColorType`.

antonfirsov (commented):
A few suggestions, but mostly LGTM.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: Without knowing the component precision of the `TPixel` we don't know which boxed pixel to assign to. We could do it in v4 though if we added additional information to `PixelTypeInfo` https://github.com/SixLabors/ImageSharp/discussions/2534#discussioncomment-7171432

tests/ImageSharp.Tests/Formats/Png/PngEncoderTests.cs
JimBobSquarePants: Done

src/ImageSharp/Formats/Png/PngDecoderCore.cs
antonfirsov: Ah yeah this indeed goes well beyond this PR, nevermind then.

antonfirsov (approved):

PR #2500, opened 2023-07-27
Vectorize TrimTransparentPixels in GifEncoderCore

gfoidl:
### Prerequisites

### Description

See https://github.com/SixLabors/ImageSharp/pull/2455#issuecomment-1613901985

A simple benchmark -- just for the inner loop -- yields:
```
|     Method |      Mean |    Error |   StdDev | Ratio |
|----------- |----------:|---------:|---------:|------:|
|    Default | 102.33 ns | 2.073 ns | 2.973 ns |  1.00 |
| Vectorized |  16.53 ns | 0.065 ns | 0.055 ns |  0.16 |
```
This is measured with .NET 7, but the codegen for .NET 6 is very similar.

<details>
  <summary>benchmark code</summary>

```c#
using System.Runtime.CompilerServices;
using System.Runtime.InteropServices;
using System.Runtime.Intrinsics;

Bench bench = new();
bench.Setup();
Console.WriteLine(bench.Default());
Console.WriteLine(bench.Vectorized());

#if !DEBUG
BenchmarkDotNet.Running.BenchmarkRunner.Run<Bench>();
#endif

public class Bench
{
    private byte[] _rowSpan = null!;
    private byte _trimmableIndex;

    [GlobalSetup]
    public void Setup()
    {
        _rowSpan = new byte[100];
        _rowSpan.AsSpan().Fill(42);
        _rowSpan.AsSpan(25, 9).Clear();

        _trimmableIndex = 42;
    }

    [Benchmark(Baseline = true)]
    public (int left, int right, bool isTransparentRow) Default()
    {
        Span<byte> rowSpan = _rowSpan;
        byte trimmableIndex = _trimmableIndex;

        int left = int.MaxValue;
        int right = int.MinValue;
        bool isTransparentRow = true;

        for (int x = 0; x < rowSpan.Length; ++x)
        {
            if (rowSpan[x] != trimmableIndex)
            {
                isTransparentRow = false;
                left = Math.Min(left, x);
                right = Math.Max(right, x);
            }
        }

        if (left == int.MaxValue)
        {
            left = 0;
        }

        if (right == int.MinValue)
        {
            right = rowSpan.Length;
        }

        return (left, right, isTransparentRow);
    }

    [Benchmark]
    public (int left, int right, bool isTransparentRow) Vectorized()
    {
        Span<byte> rowSpan = _rowSpan;
        byte trimmableIndex = _trimmableIndex;

        int left = int.MaxValue;
        int right = int.MinValue;
        bool isTransparentRow = true;

        ref byte rowPtr = ref MemoryMarshal.GetReference(rowSpan);
        nint rowLength = (nint)(uint)rowSpan.Length;
        nint x = 0;

        if (Vector128.IsHardwareAccelerated && rowLength >= Vector128<byte>.Count)
        {
            Vector256<byte> trimmableVec256 = Vector256.Create(trimmableIndex);

            if (Vector256.IsHardwareAccelerated && rowLength >= Vector256<byte>.Count)
            {
                do
                {
                    Vector256<byte> vec = Vector256.LoadUnsafe(ref rowPtr, (nuint)x);
                    Vector256<byte> notEquals = ~Vector256.Equals(vec, trimmableVec256);

                    if (notEquals != Vector256<byte>.Zero)
                    {
                        isTransparentRow = false;
                        uint mask = notEquals.ExtractMostSignificantBits();
                        nint start = x + (nint)uint.TrailingZeroCount(mask);

                        nint end = (nint)uint.LeadingZeroCount(mask);
                        // end is from the end, but we need the index from the beginning
                        end = x + Vector256<byte>.Count - 1 - end;

                        left = Math.Min(left, (int)start);
                        right = Math.Max(right, (int)end);
                    }

                    x += Vector256<byte>.Count;
                }
                while (x <= rowLength - Vector256<byte>.Count);
            }

            Vector128<byte> trimmableVec = Vector256.IsHardwareAccelerated
                ? trimmableVec256.GetLower()
                : Vector128.Create(trimmableIndex);

            while (x <= rowLength - Vector128<byte>.Count)
            {
                Vector128<byte> vec = Vector128.LoadUnsafe(ref rowPtr, (nuint)x);
                Vector128<byte> notEquals = ~Vector128.Equals(vec, trimmableVec);

                if (notEquals != Vector128<byte>.Zero)
                {
                    isTransparentRow = false;
                    uint mask = notEquals.ExtractMostSignificantBits();
                    nint start = x + (nint)uint.TrailingZeroCount(mask);

                    nint end = (nint)uint.LeadingZeroCount(mask) - Vector128<byte>.Count;
                    // end is from the end, but we need the index from the beginning
                    end = x + Vector128<byte>.Count - 1 - end;

                    left = Math.Min(left, (int)start);
                    right = Math.Max(right, (int)end);
                }

                x += Vector128<byte>.Count;
            }
        }

        for (; x < rowLength; ++x)
        {
            if (Unsafe.Add(ref rowPtr, x) != trimmableIndex)
            {
                isTransparentRow = false;
                left = Math.Min(left, (int)x);
                right = Math.Max(right, (int)x);
            }
        }

        if (left == int.MaxValue)
        {
            left = 0;
        }

        if (right == int.MinValue)
        {
            right = (int)rowLength;
        }

        return (left, right, isTransparentRow);
    }
}
```
</details>

### DIFF
--- a/src/ImageSharp/Formats/Gif/GifEncoderCore.cs
+++ b/src/ImageSharp/Formats/Gif/GifEncoderCore.cs
@@ -435,11 +435,11 @@ internal sealed class GifEncoderCore : IImageEncoderInternals
                     {
                         Vector256<byte> vec = Vector256.LoadUnsafe(ref rowPtr, (nuint)x);
                         Vector256<byte> notEquals = ~Vector256.Equals(vec, trimmableVec256);
+                        uint mask = notEquals.ExtractMostSignificantBits();
 
-                        if (notEquals != Vector256<byte>.Zero)
+                        if (mask != 0)
                         {
                             isTransparentRow = false;
-                            uint mask = notEquals.ExtractMostSignificantBits();
                             nint start = x + (nint)uint.TrailingZeroCount(mask);
                             nint end = (nint)uint.LeadingZeroCount(mask);
 
@@ -463,11 +463,11 @@ internal sealed class GifEncoderCore : IImageEncoderInternals
                 {
                     Vector128<byte> vec = Vector128.LoadUnsafe(ref rowPtr, (nuint)x);
                     Vector128<byte> notEquals = ~Vector128.Equals(vec, trimmableVec);
+                    uint mask = notEquals.ExtractMostSignificantBits();
 
-                    if (notEquals != Vector128<byte>.Zero)
+                    if (mask != 0)
                     {
                         isTransparentRow = false;
-                        uint mask = notEquals.ExtractMostSignificantBits();
                         nint start = x + (nint)uint.TrailingZeroCount(mask);
                         nint end = (nint)uint.LeadingZeroCount(mask) - Vector128<byte>.Count;
 
@@ -493,11 +493,11 @@ internal sealed class GifEncoderCore : IImageEncoderInternals
                         Vector256<byte> vec = Unsafe.ReadUnaligned<Vector256<byte>>(ref Unsafe.Add(ref rowPtr, x));
                         Vector256<byte> notEquals = Avx2.CompareEqual(vec, trimmableVec256);
                         notEquals = Avx2.Xor(notEquals, Vector256<byte>.AllBitsSet);
+                        int mask = Avx2.MoveMask(notEquals);
 
-                        if (!Avx.TestZ(notEquals, notEquals))
+                        if (mask != 0)
                         {
                             isTransparentRow = false;
-                            int mask = Avx2.MoveMask(notEquals);
                             nint start = x + (nint)(uint)BitOperations.TrailingZeroCount(mask);
                             nint end = (nint)(uint)BitOperations.LeadingZeroCount((uint)mask);
 
@@ -522,11 +522,11 @@ internal sealed class GifEncoderCore : IImageEncoderInternals
                     Vector128<byte> vec = Unsafe.ReadUnaligned<Vector128<byte>>(ref Unsafe.Add(ref rowPtr, x));
                     Vector128<byte> notEquals = Sse2.CompareEqual(vec, trimmableVec);
                     notEquals = Sse2.Xor(notEquals, Vector128<byte>.AllBitsSet);
+                    int mask = Sse2.MoveMask(notEquals);
 
-                    if (!Sse41.TestZ(notEquals, notEquals))
+                    if (mask != 0)
                     {
                         isTransparentRow = false;
-                        int mask = Sse2.MoveMask(notEquals);
                         nint start = x + (nint)(uint)BitOperations.TrailingZeroCount(mask);
                         nint end = (nint)(uint)BitOperations.LeadingZeroCount((uint)mask) - Vector128<byte>.Count;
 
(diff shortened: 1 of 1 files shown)


src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -412,23 +412,142 @@ private static Buffer2DRegion<byte> TrimTransparentPixels(Buffer2D<byte> buffer,
...
+                Vector256<byte> trimmableVec256 = Vector256.Create(trimmableIndex);
+
+                if (Vector256.IsHardwareAccelerated && rowLength >= Vector256<byte>.Count)
+                {
+                    do
+                    {
+                        Vector256<byte> vec = Vector256.LoadUnsafe(ref rowPtr, (nuint)x);
+                        Vector256<byte> notEquals = ~Vector256.Equals(vec, trimmableVec256);
+
+                        if (notEquals != Vector256<byte>.Zero)
```
gfoidl, line 439: At the moment I don't have any idea on how to make this branchless.
`isTransparentRow` could be tracked in a vector, but `left` and `right` not, as there's a mismatch of vector-types, namely `byte` and `int`. 

A quite complicated approach would be to use `VectorXYZ<byte>` and track the left and right -- but just before these can overflow merge it back to the scalar `left`, `right` and start over. But I guess the book-keeping is more work, so I'm not sure if this is actually faster. For sure the code gets painful.

src/ImageSharp/Formats/Gif/GifEncoderCore.cs
```diff
@@ -412,23 +412,142 @@ private static Buffer2DRegion<byte> TrimTransparentPixels(Buffer2D<byte> buffer,
...
+
+                        left = Math.Min(left, (int)start);
+                        right = Math.Max(right, (int)end);
+                    }
+
+                    x += Vector128<byte>.Count;
+                }
+            }
+#endif
+            for (; x < rowLength; ++x)
```
gfoidl, line 544: The remainder could be handled vectorized too, by shifting the mask of the most significant bits around by the count of elements left in the final vector. 
I tried this somewhere else, the cost for that book-keeping isn't negligible, so didn't do this here and now (maybe I'll try this later).

gfoidl (commented):
Some notes for review.

JimBobSquarePants:
Oof! Look at those numbers!! Thanks so much for looking at this. I'm going have a good dig through it to wrap my head round what you have done.

gfoidl:
😃 
I think the easiest way to understand / check is to use the benchmark-code (see top-comment) in a simple console app and step with the debugger through it (maybe change the size of the `_rowSpan` to 20 or that like.
Calculation of the correct index for `end` is the strangest part IMO.

JimBobSquarePants:
This is fantastic stuff. I figured theoretically after reading some of the source for `Span.IndexOf` that masking with bit counting would be the vectorized solution I just had no idea how I'd actually implement it.

Tip of the cap to you sir.

gfoidl:
Thanks for the kind words ❤️

PR #2501, opened 2023-08-01
Prevent crafted DOS attack.

JimBobSquarePants:
### Prerequisites

### Description

It's possible to cause the library to hang with a crafted byte array targetting the PBM decoder. This PR fixes that vulnarability. All other formats correctly test for the end of the stream.

### DIFF
--- a/src/ImageSharp/Formats/Pbm/BufferedReadStreamExtensions.cs
+++ b/src/ImageSharp/Formats/Pbm/BufferedReadStreamExtensions.cs
@@ -28,7 +28,7 @@ internal static class BufferedReadStreamExtensions
                 {
                     innerValue = stream.ReadByte();
                 }
-                while (innerValue != 0x0a);
+                while (innerValue is not 0x0a and not -0x1);
 
                 // Continue searching for whitespace.
                 val = innerValue;
--- a/tests/ImageSharp.Tests/Formats/Pbm/PbmMetadataTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Pbm/PbmMetadataTests.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
 
+using SixLabors.ImageSharp.Formats;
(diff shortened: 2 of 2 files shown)


tocsoft (approved):

src/ImageSharp/Formats/Pbm/BufferedReadStreamExtensions.cs
```diff
@@ -28,7 +28,7 @@ public static void SkipWhitespaceAndComments(this BufferedReadStream stream)
                 {
                     innerValue = stream.ReadByte();
                 }
-                while (innerValue != 0x0a);
+                while (innerValue is not 0x0a and not -0x1);
```
ynse01, line 31: Should a similar check be added also in line 52 ?

antonfirsov:
```C#
int current = stream.ReadByte() - 0x30;
if ((uint)current > 9)
{
    break;
}
```
In case of EOF `current == -1 -0x30 == -0x31`. `(uint)-0x31 == SomeHighNum > 9`.

Nevertheless, a test case might be nice, these infinite loops are dangerous. 💀

JimBobSquarePants: Yeah, that's why I left it. We actually enter that loop in ReadDecimal with this test, but I'll double check before merging.

JimBobSquarePants: We cover this code with the test so I'm happy. I'll open another PR after this one to backport the fix to 2.X

antonfirsov (approved):
We shall merge and publish this, the faster the better. It also needs a 2.1 backport, which I would be happy to do, but I have no computer access until Monday.

JimBobSquarePants:
> We shall merge and publish this, the faster the better. It also needs a 2.1 backport, which I would be happy to do, but I have no computer access until Monday.

Done #2509

PR #2511, opened 2023-08-12
APNG support

Poker-sang:
### Prerequisites

### Description
Implement APNG decoding and encoding based on PNG


src/ImageSharp/Formats/Png/APngDisposeOperation.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Six Labors.
+// Licensed under the Six Labors Split License.
+
+namespace SixLabors.ImageSharp.Formats.Png;
+
+/// <summary>
+/// Specifies how the output buffer should be changed at the end of the delay (before rendering the next frame).
+/// </summary>
+public enum APngDisposeOperation
```
JimBobSquarePants, line 9: `PngDisposalMethod` to normalize with Gif.

src/ImageSharp/Formats/Png/APngFrameMetadata.cs
```diff
@@ -0,0 +1,94 @@
...
+// Licensed under the Six Labors Split License.
+
+using SixLabors.ImageSharp.Formats.Png.Chunks;
+
+namespace SixLabors.ImageSharp.Formats.Png;
+
+/// <summary>
+/// Provides APng specific metadata information for the image frame.
+/// </summary>
+public class APngFrameMetadata : IDeepCloneable
```
JimBobSquarePants, line 11: `PngFrameMetadata`

src/ImageSharp/Formats/Png/Chunks/APngAnimationControl.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Six Labors.
+// Licensed under the Six Labors Split License.
+
+using System.Buffers.Binary;
+
+namespace SixLabors.ImageSharp.Formats.Png.Chunks;
+
+internal record APngAnimationControl(
```
JimBobSquarePants, line 8: This reads like it should be a readonly struct.

src/ImageSharp/Formats/Png/Chunks/APngAnimationControl.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// <summary>
+    /// Gets the number of frames
+    /// </summary>
+    public int NumberFrames { get; } = NumberFrames;
+
+    /// <summary>
+    /// Gets the number of times to loop this APNG.  0 indicates infinite looping.
+    /// </summary>
+    public int NumberPlays { get; } = NumberPlays;
```
JimBobSquarePants, line 22: `RepeatCount` same as Gif.

src/ImageSharp/Formats/Png/Chunks/APngFrameControl.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        if (this.Height <= 0)
+        {
+            throw new NotSupportedException($"Invalid Height. Expected > 0. Was '{this.Height}'.");
+        }
+
+        if (this.XOffset + this.Width > hdr.Width)
+        {
+            throw new NotSupportedException($"Invalid XOffset or Width. The sum > PngHeader.Width. Was '{this.XOffset + this.Width}'.");
```
JimBobSquarePants, line 109: Can we move this to a throwhelper?

src/ImageSharp/Formats/Png/Chunks/PngPhysical.cs
```diff
@@ -10,11 +10,11 @@ namespace SixLabors.ImageSharp.Formats.Png.Chunks;
 /// <summary>
 /// The pHYs chunk specifies the intended pixel size or aspect ratio for display of the image.
 /// </summary>
-internal readonly struct PhysicalChunkData
+internal readonly struct PngPhysical
```
JimBobSquarePants, line 13: Not sure why you renamed this.

src/ImageSharp/Formats/Png/Chunks/APngFrameControl.cs
```diff
@@ -0,0 +1,172 @@
...
+
+using System.Buffers.Binary;
+
+namespace SixLabors.ImageSharp.Formats.Png.Chunks;
+
+internal readonly struct APngFrameControl
+{
+    public const int Size = 26;
+
+    public APngFrameControl(
```
JimBobSquarePants, line 12: `FrameControl` is fine.

src/ImageSharp/Formats/Png/Chunks/APngAnimationControl.cs
JimBobSquarePants, line 8: `AnimationControl` is fine, we don't expose the type and it's namespaced.

src/ImageSharp/Formats/Png/MetadataExtensions.cs
```diff
@@ -14,7 +15,23 @@ public static partial class MetadataExtensions
...
     /// <returns>The <see cref="PngMetadata"/>.</returns>
-    public static PngMetadata GetPngMetadata(this ImageMetadata metadata) => metadata.GetFormatMetadata(PngFormat.Instance);
+    public static PngMetadata GetPngMetadata(this ImageMetadata source) => source.GetFormatMetadata(PngFormat.Instance);
+
+    /// <summary>
+    /// Gets the aPng format specific metadata for the image frame.
+    /// </summary>
+    /// <param name="source">The metadata this method extends.</param>
+    /// <returns>The <see cref="APngFrameMetadata"/>.</returns>
+    public static APngFrameMetadata GetAPngFrameMetadata(this ImageFrameMetadata source) => source.GetFormatMetadata(PngFormat.Instance);
```
JimBobSquarePants, line 27: You can drop the `A` prefix.

src/ImageSharp/Formats/Png/PngConstants.cs
```diff
@@ -80,24 +80,5 @@ internal static class PngConstants
...
-        (byte)'d',
-        (byte)'o',
-        (byte)'b',
-        (byte)'e',
-        (byte)'.',
-        (byte)'x',
-        (byte)'m',
-        (byte)'p'
-    };
+    public static ReadOnlySpan<byte> XmpKeyword => "XML:com.adobe.xmp"u8;
```
JimBobSquarePants, line 83: I absolutely love that we will be able to do this but I don't think we should change the language version to allow it.

src/ImageSharp/Formats/Png/PngEncoder.cs
```diff
@@ -18,7 +17,12 @@ public class PngEncoder : QuantizingImageEncoder
 
         // We set the quantizer to null here to allow the underlying encoder to create a
         // quantizer with options appropriate to the encoding bit depth.
-        this.Quantizer = null;
+        this.Quantizer = null!;
+
+    /// <summary>
+    /// Gets whether the file is a simple PNG.
+    /// </summary>
+    public bool? IsSimplePng { get; init; }
```
JimBobSquarePants, line 25: No need for this. The encoder should automatically encode multiple frames as animated png.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -26,7 +25,7 @@ internal sealed class PngEncoderCore : IImageEncoderInternals, IDisposable
     /// <summary>
     /// The maximum block size, defaults at 64k for uncompressed blocks.
     /// </summary>
-    private const int MaxBlockSize = 65535;
+    private const int MaxBlockSize = (1 << 16) - 1;
```
JimBobSquarePants, line 28: I guess...

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -28,6 +28,11 @@ namespace SixLabors.ImageSharp.Formats.Png;
 /// </summary>
 internal sealed class PngDecoderCore : IImageDecoderInternals
 {
+    /// <summary>
+    /// Indicate whether the file is a simple PNG.
+    /// </summary>
+    private bool isSimplePng;
```
JimBobSquarePants, line 34: I don't think you need this field.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -490,6 +597,27 @@ private void InitializeImage<TPixel>(ImageMetadata metadata, out Image<TPixel> i
         this.scanline = this.configuration.MemoryAllocator.Allocate<byte>(this.bytesPerScanline, AllocationOptions.Clean);
     }
 
+    /// <summary>
+    /// Initializes the image and various buffers needed for processing
+    /// </summary>
+    /// <typeparam name="TPixel">The type the pixels will be</typeparam>
+    /// <param name="frameControl">The frame control information for the frame</param>
+    /// <param name="image">The image that we will populate</param>
+    private void InitializeFrame<TPixel>(APngFrameControl frameControl, Image<TPixel> image, out ImageFrame<TPixel> frame)
```
JimBobSquarePants, line 606: Missing parameter

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -264,6 +329,14 @@ public ImageInfo Identify(BufferedReadStream stream, CancellationToken cancellat
                         case PngChunkType.Header:
                             this.ReadHeaderChunk(pngMetadata, chunk.Data.GetSpan());
                             break;
+                        case PngChunkType.AnimationControl:
```
JimBobSquarePants, line 332: When decoding multi-frame images the decoder should respect the `DecoderOptions.MaxFrames` property.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -647,38 +779,47 @@ private void DecodePixelData<TPixel>(DeflateStream compressedStream, ImageFrame<
...
+
         Buffer2D<TPixel> imageBuffer = image.PixelBuffer;
         while (true)
         {
             int numColumns = Adam7.ComputeColumns(width, pass);
 
             if (numColumns == 0)
             {
-                pass++;
+                ++pass;
```
JimBobSquarePants, line 800: Not sure why you reversed this. Less intuitive when incrementing.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -264,6 +329,14 @@ public ImageInfo Identify(BufferedReadStream stream, CancellationToken cancellat
                         case PngChunkType.Header:
                             this.ReadHeaderChunk(pngMetadata, chunk.Data.GetSpan());
                             break;
+                        case PngChunkType.AnimationControl:
+                            if (this.isSimplePng || this.animationControl is not null)
```
JimBobSquarePants, line 333: I don't think we need the additional checks here. A standard png will not have this chunk.

Do you have examples of APNGs with multiple animation controls?

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -365,9 +465,9 @@ public ImageInfo Identify(BufferedReadStream stream, CancellationToken cancellat
             }
 
             EOF:
-            if (this.header.Width == 0 && this.header.Height == 0)
+            if (this.header is { Width: 0, Height: 0 })
```
JimBobSquarePants, line 468: I can't stand this syntax. Pattern matching should be used to simplify readability, not obfuscate it.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -585,22 +714,25 @@ private void ReadScanlines<TPixel>(PngChunk chunk, ImageFrame<TPixel> image, Png
     private void DecodePixelData<TPixel>(DeflateStream compressedStream, ImageFrame<TPixel> image, PngMetadata pngMetadata, CancellationToken cancellationToken)
         where TPixel : unmanaged, IPixel<TPixel>
     {
-        while (this.currentRow < this.header.Height)
+        int currentRow = Adam7.FirstRow[0];
+        int currentRowBytesRead = 0;
+        int height = image.Metadata.TryGetAPngFrameMetadata(out APngFrameMetadata? frameMetadata) ? frameMetadata.Height : this.header.Height;
+        while (currentRow < height)
```
JimBobSquarePants, line 720: All frames within a decoded image must be the same size. The decoded data should be offset against the frame. See the gif decoder.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -160,22 +162,84 @@ public Image<TPixel> Decode<TPixel>(BufferedReadStream stream, CancellationToken
...
+                            {
+                                this.InitializeFrame(lastFrameControl.Value, image, out currentFrame);
+                            }
+
+                            this.currentStream.Position += 4;
+                            this.ReadScanlines(
+                                chunk.Length - 4,
+                                currentFrame,
+                                pngMetadata,
+                                () =>
```
JimBobSquarePants, line 216: This feels very messy. I think it needs reworking. Separate methods are fine.

src/ImageSharp/ImageSharp.csproj
```diff
@@ -13,6 +13,7 @@
     <PackageTags>Image Resize Crop Gif Jpg Jpeg Bitmap Pbm Png Tga Tiff WebP NetCore</PackageTags>
     <Description>A new, fully featured, fully managed, cross-platform, 2D graphics API for .NET</Description>
     <Configurations>Debug;Release</Configurations>
+    <LangVersion>preview</LangVersion>
```
JimBobSquarePants, line 16: We don't want to change this just now. It limits contribution.

JimBobSquarePants (changes requested):
Thanks for this, I really appreciate your efforts!

 I've tried to offer guidance as to what need to change. Please let me know if anything doesn't make sense.

src/ImageSharp/Formats/Png/Chunks/PngPhysical.cs
Poker-sang: I was trying to make all trunks type under the same naming style. Says PngHeader, PngTextData, **PngPhysical**, APngAnimationControl, APngFrameControl. Maybe I should rename them to Header, TextData, Physical, AnimationControl, FrameControl?

Poker-sang:
Thanks for your review! It is rigorous and easy for me to understand. I'm sorry for accidentally formatting some code that confused you (e.g. x++ to ++x). Now it's all reverted.

src/ImageSharp/Formats/Png/PngScanlineProcessor.cs
```diff
@@ -298,12 +309,23 @@ internal static class PngScanlineProcessor
         }
         else
         {
-            PixelOperations<TPixel>.Instance.FromRgb24Bytes(configuration, scanlineSpan, rowSpan, header.Width);
+            ReadOnlySpan<Rgb24> source = MemoryMarshal.Cast<byte, Rgb24>(scanlineSpan)[..frameControl.Width];
```
JimBobSquarePants, line 312: Why not slice both spans? Your scanline can only be <= to header width.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -682,8 +685,9 @@ private int CalculateScanlineLength(int width)
     /// <param name="image"> The pixel data.</param>
     /// <param name="pngMetadata">The png metadata</param>
     /// <param name="getData">A delegate to get more data from the inner stream for <see cref="ZlibInflateStream"/>.</param>
+    /// <param name="frameControl">The frame control</param>
     /// <param name="cancellationToken">The cancellation token.</param>
-    private void ReadScanlines<TPixel>(int chunkLength, ImageFrame<TPixel> image, PngMetadata pngMetadata, Func<int> getData, CancellationToken cancellationToken)
+    private void ReadScanlines<TPixel>(int chunkLength, ImageFrame<TPixel> image, PngMetadata pngMetadata, Func<int> getData, FrameControl frameControl, CancellationToken cancellationToken)
```
JimBobSquarePants, line 674: Pass `FrameControl` via `in` to avoid copying.

Poker-sang: Do I need to put `in` before all `FrameControl`'s? Same as here.

src/ImageSharp/Formats/Png/PngEncoder.cs
```diff
@@ -18,7 +17,7 @@ public class PngEncoder : QuantizingImageEncoder
 
         // We set the quantizer to null here to allow the underlying encoder to create a
         // quantizer with options appropriate to the encoding bit depth.
-        this.Quantizer = null;
+        this.Quantizer = null!;
```
antonfirsov, line 20: If `QuantizingImageEncoder.Quantizer == null` is a valid state, shouldn't it be nullable?

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -101,12 +100,12 @@ internal sealed class PngEncoderCore : IImageEncoderInternals, IDisposable
     /// <summary>
     /// The raw data of previous scanline.
     /// </summary>
-    private IMemoryOwner<byte> previousScanline;
+    private IMemoryOwner<byte> previousScanline = null!;
```
antonfirsov, line 103: Shouldn't this field be nullable instead?

src/ImageSharp/Formats/Png/PngEncoder.cs
Poker-sang: hmm yes, but `QuantizingImageEncoder.Quantizer` is also used in encoders in many other formats. In order not to change more code, I suppressed the nullable notification here.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
Poker-sang: Most of the usages of these fields are not null. It is possible to be null only when dispose. So I also used null-forgiving here for the sake of code brevity.

JimBobSquarePants: You can remove the explicit set to null then from disposal. GC will clean up.

src/ImageSharp/Formats/Png/PngEncoder.cs
JimBobSquarePants: It shouldn't be nullable anywhere really, however we're unable to lazily set the quantizer to the required bitdepth without it being nullable here. Best just leaving nullability checks disabled in this file for now. The core encoder deals with nullability. 

I want to have a good look at the quantizers again for V4.

Poker-sang: Then we can disable nullable in the `QuantizingImageEncoder.cs`. This allows us to use nullable in the `Encoder` and avoid property `Quantizer`'s null checking.

JimBobSquarePants: You can't do that because that breaks the ability to use a palette quantizer in the presence of a color palette in the metadata #2455  If someone explicitly sets a quantizer they are showing intent.

Poker-sang: I am sorry. I will revert the enable disable in `PngEncoder.cs`.

JimBobSquarePants: Nothing to apologize for! 🙂

JimBobSquarePants:
@Poker-sang Sorry this is taking so long to merge. There's a bit of a backlog we need to shift first to ensure there are no functional conflicts.

Poker-sang:
It's all right. Take your time please.

JimBobSquarePants:
@Poker-sang Thanks again for your patience. Following the merge of #2485 there's a few conflicts here. Would you be able to update your PR to include the updates from main? If not, I can have a try.

Poker-sang:
I can have a try. But I'm sorry I can't handle it for I'm busy these days. If you're not in a hurry you can wait a few days for me. ☺️ Or you can handle it the way you want if it doesn't bother you.

JimBobSquarePants:
I'm having a go at merging main back into this PR now.

JimBobSquarePants:
@Poker-sang I've merged the code plus performed some cleanup. We're getting pretty close now, I think. 

Things to note:
- I removed the offsets/widths from the frame metadata. When decoding we should decode to an offset within the full frame and the metadata values become meaningless as soon as we transform the image. We should, in the future, be able to optimize the encoding process by clipping the frames to their non-transparent bounds and encoding only the remaining data with appropriate control offsets (not required for this PR).
- I optimized the code somewhat, we were allocating far too much when cloning and writing the frame data.
- I've renamed some things in the metadata and will have a thinking about renaming others to match gif naming convention. I'm also thinking we should use our Rational type for the delay.

It doesn't seem to be that we are using the blending information at all currently during decode. We'll have to implement that, maybe gif can be used as a reference (I imagine the approach is very similar). 

We definitely need more samples and comprehensive tests. 

There's a few here.
http://littlesvr.ca/apng/samples.html

But maybe we can use this to make more.
https://sourceforge.net/projects/apngasm/

src/ImageSharp/Formats/Png/Chunks/AnimationControl.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    /// <summary>
+    /// Gets the number of frames
+    /// </summary>
+    public int NumberFrames { get; }
+
+    /// <summary>
+    /// Gets the number of times to loop this APNG.  0 indicates infinite looping.
+    /// </summary>
+    public int NumberPlays { get; }
```
JimBobSquarePants, line 26: RepeatCount

JimBobSquarePants:
> Thank you so much for helping me resolve the conflict. Your code is very clear and impressive! Regarding the points you mentioned, I think I can explain my previous considerations:
> 
> * You removed the width and offset fields, but I'm concerned that this will result in images that are unclipped and large.
> * The naming and type of the original field was something I just wanted to keep in line with the standard documentation. Therefore, I thought it would be fine to change it to some other more understandable name. I also think it is necessary to change the type of the field `Delay` to `Rational` and to be consistent with the GIF.
> * We do need more APNG test samples. I previously referenced the images in https://philip.html5.org/tests/apng/tests.html. Not all of them are available to us, but I think some will help.
> * And since several formats of animated images are implemented. Maybe the type `ImageFrame` needs some slight refactoring to facilitate the handling of animated images.

Happy to help! 

To answer your questions.

The reason they were removed is that they're useful only for placing the data correctly on decode. 
During decode (for gif) we use those offsets to place pixel data into the frame buffer. You can end up with something like this.

```
+------------------------------------------+
|                                          |
|                                          |
|                                          |
|     x,y                                  |
|      +---------------------------+       |
|      |                           |       |
|      |                           |       |
|      |                           |       |
|      |                           |       |
|      |                           | h     |
|      |                           |       |
|      |                           |       |
|      |                           |       |
|      |                           |       |
|      +---------------------------+       |
|                   w                      |
|                                          |
|                                          |
|                                          |
|                                          |
+------------------------------------------+
```

Any transform applied to that image (resize, skew, rotate, etc) will adjust those positions and we have no mechanism by which we can ensure they can be updated accurately. Imagine if a 3rd party wrote a general warping algorithm, there's simply no way to keep track. In addition, any drawing operations could overflow those original bounds. 

The only way to handle this is to keep all frames the same size and place the pixel data in the buffer on decode, we then discard the metadata and, if possible, recalculate it on encode by clipping the quantized frame buffer based on the presence of transparent pixels. You can see that in action in our gif encoder.

https://github.com/SixLabors/ImageSharp/blob/d93bc6cca2383095c2a08db85861b88a1c94c43c/src/ImageSharp/Formats/Gif/GifEncoderCore.cs#L403

Now hopefully we can introduce a general-purpose quantizing utility which will allow us to apply the same approach to all quantized buffers, but I need to have a think about that as we need to cater for multple transparent palette entry support in a performant manner. For the time being it's best to simply encode the whole buffer.

Naming is hard 😄 We tend to use the specification names for internal types but always attempt to normalize property names for public types for API consistency.

I'm curious to hear what changes you deem required for `ImageFrame` as I don't see anything currently that is an additional requirement for animated formats though am always up for good ideas.

Poker-sang:
hmmm say, almost all animated image formats. Frames have a `delay` property. Maybe we can give the `Delay` as a property to the type `ImageFrame` instead of using metadata?

JimBobSquarePants:
> > I'm curious to hear what changes you deem required for `ImageFrame` as I don't see anything currently that is an additional requirement for animated formats though am always up for good ideas.
> 
> hmmm say, almost all animated image formats. Frames have a `delay` property. Maybe we can give the `Delay` as a property to the type `ImageFrame` instead of using metadata? (Or create a property in the base type `ImageFrameMetadata`)

Ah no, think of `ImageFrame` as merely an in-memory representation of the pixel data with no superflous properties. A `Delay` property would not make sense for an image containing a single frame. 

We've deliberately kept any metadata separate to highlight this and to allow unlimited metadata variations.

JimBobSquarePants:
I need to investigate what is wrong with the encoder. There's two tests failing.

Poker-sang:
> Ah no, think of `ImageFrame` as merely an in-memory representation of the pixel data with no superflous properties. A `Delay` property would not make sense for an image containing a single frame. 
> 
> We've deliberately kept any metadata separate to highlight this and to allow unlimited metadata variations. 

If we convert gif to apng, webp or other animated image formats, the delay information may be lost. Or we need to write appropriate logic for each metadata. 🤔

JimBobSquarePants:
Yes, you would also lose any disposal or blend mode information

Poker-sang:
Ok I see. Thanks for explaining.

JimBobSquarePants:
@Poker-sang I've added support for alpha blending and disposal.

I think it's correct based on how we work with gif/webp but I could do with a second pair of eyes. We desperately need to add test images now to validatethe implementation so if you could add that, it would be great!

JimBobSquarePants:
@Poker-sang Can you give me collaboration access to your fork? There's an open issue with Git LFS which prevents me pushing images to your fork without those permissions.

https://github.com/git-lfs/git-lfs/issues/5001

Poker-sang:
Sure. When I'm cloning a repository lately, filtering content always takes me a long time. Is it caused by too many test images?

JimBobSquarePants:
> Sure. When I'm cloning this repo lately, `Filtering Content` always takes me a long time. Is it caused by too many test images?

I'm not sure what you mean by `Filtering Content`.

JimBobSquarePants:
Ah yeah, there's a lot of LFS files. 

You should only be cloning the repo 1 time really though and using feature branches for your PRs. Every time someone with a fork clones the repo it actually uses up my LFS bandwidth and ends up costing me money if I go over my pre-purchased allowance.

Poker-sang:
Oh I see. I'll pay attention to it.

JimBobSquarePants:
I’ve started adding the tests locally.

Poker-sang:
APNG Specification says:
> If the default image is the first frame:
> ```
> Sequence number    Chunk
> (none)             `acTL`
> 0                  `fcTL` first frame
> (none)             `IDAT` first frame / default image
> 1                  `fcTL` second frame
> 2                  first `fdAT` for second frame
> 3                  second `fdAT` for second frame
> ....
> ```
> If the default image is not part of the animation:
> ```
> Sequence number    Chunk
> (none)             `acTL`
> (none)             `IDAT` default image
> 0                  `fcTL` first frame
> 1                  first `fdAT` for first frame
> 2                  second `fdAT` for first frame
> ....
> ```
Seems like a simple and not very useful feature, do we need to implement it? Says adding an option in the `Encoder` to exclude the first frame.

JimBobSquarePants:
I think that just means the first frame control is optional. The code I have locally handles that.

Poker-sang:
I am so happy to hear that. I am so grateful that you have helped me so much!♥

src/ImageSharp/Formats/Png/Chunks/AnimationControl.cs
```diff
@@ -0,0 +1,47 @@
...
+    public AnimationControl(int numberFrames, int numberPlays)
+    {
+        this.NumberFrames = numberFrames;
+        this.NumberPlays = numberPlays;
+    }
+
+    /// <summary>
+    /// Gets the number of frames
+    /// </summary>
+    public int NumberFrames { get; }
```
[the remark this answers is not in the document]
Poker-sang: Do we have to fix this before merging? Or just resolve conversation?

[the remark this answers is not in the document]
JimBobSquarePants: Yeah we need to fix this and others. I planned on doing it today

JimBobSquarePants (approved):
I'm happy with this now. Thanks @Poker-sang for a great addition to the library!

PR #2514, opened 2023-08-17
Expose non-nullable configuration to remove AOT limiting null check

### LINKED ISSUE #2486: JPEG/WebP/... part of the executable even if all I do is save a PNG
### Prerequisites

### ImageSharp version

3.0.1

?

Windows 11

### .NET Framework version

?

### Description

As requested on Twitter: https://twitter.com/James_M_South/status/1673489691956625410

When compiling https://github.com/pjmlp/ppm2png/tree/2b77cfc103f48bbcdce0512f6182153d6d4d8644/csharp with PublishAot, the produced executable is rather large. The reason why it's large can be inspected with [Sizoscope](https://github.com/MichalStrehovsky/sizoscope). I made this change to the above repo in an attempt to make the size smaller but it didn't help much:

```diff
diff --git a/csharp/ImageWriter.cs b/csharp/ImageWriter.cs
index 89af11c..5b4b3eb 100644
--- a/csharp/ImageWriter.cs
+++ b/csharp/ImageWriter.cs
@@ -36,7 +36,7 @@ public static class ImageWriter
     /// <param name="height">The image height in pixels.</param>
     public static void SaveImage(string filename, ReadOnlySpan<byte> buffer, int width, int height)
     {
-        using var image = Image.LoadPixelData<Rgb24>(buffer, width, height);
+        using var image = Image.LoadPixelData<Rgb24>(new Configuration(), buffer, width, height);
         using var stream = new System.IO.FileStream(filename, FileMode.Create);
         var pngEncoder = new SixLabors.ImageSharp.Formats.Png.PngEncoder() {
              ColorType = SixLabors.ImageSharp.Formats.Png.PngColorType.Rgb
diff --git a/csharp/ppm2png.csproj b/csharp/ppm2png.csproj
index c079a6e..cc877f5 100644
--- a/csharp/ppm2png.csproj
+++ b/csharp/ppm2png.csproj
@@ -2,7 +2,7 @@

   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <RestoreAdditionalProjectSources>https://www.myget.org/F/sixlabors/api/v3/index.json</RestoreAdditionalProjectSources>
```

I'm attaching intermediate files that can be opened with Sizoscope: 
[native.zip](https://github.com/SixLabors/ImageSharp/files/11876513/native.zip) but feel free to make your own.

One immediate issue is that this app also includes support for WebP/JPEG/etc. even though I don't need it.

Sizoscope root cause analysis for `WebpEncoder..ctor` points to `Configuration.Default` accessed from here:

image

There might be more issues.

### Steps to Reproduce

Clone the repo and build the project. Add switches described in the sizoscope repo if you want to inspect with sizoscope.

### Images

_No response_

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2486

As discussed in the linked issue `AdvancedImageExtensions.GetConfiguration()` causes the AOT compiler to include all image unused formats due to the null check and assignment of `Configuration.Default`. 

Since that null check is unavoidable without an API change, I decided to make it by making `IConfigurationProvider` public and exposing the property directly.

While developers should be careful to not edit the configuration singleton, I do not consider accessing the configuration an advanced scenario since we already provide means to construct and pass custom configuration to our types.

### DIFF
--- a/src/ImageSharp/Advanced/AdvancedImageExtensions.cs
+++ b/src/ImageSharp/Advanced/AdvancedImageExtensions.cs
@@ -27,11 +27,11 @@ public static class AdvancedImageExtensions
         Guard.NotNull(filePath, nameof(filePath));
 
         string ext = Path.GetExtension(filePath);
-        if (!source.GetConfiguration().ImageFormatsManager.TryFindFormatByFileExtension(ext, out IImageFormat? format))
+        if (!source.Configuration.ImageFormatsManager.TryFindFormatByFileExtension(ext, out IImageFormat? format))
         {
             StringBuilder sb = new();
             sb = sb.AppendLine(CultureInfo.InvariantCulture, $"No encoder was found for extension '{ext}'. Registered encoders include:");
-            foreach (IImageFormat fmt in source.GetConfiguration().ImageFormats)
+            foreach (IImageFormat fmt in source.Configuration.ImageFormats)
             {
                 sb = sb.AppendFormat(CultureInfo.InvariantCulture, " - {0} : {1}{2}", fmt.Name, string.Join(", ", fmt.FileExtensions), Environment.NewLine);
             }
@@ -39,13 +39,13 @@ public static class AdvancedImageExtensions
             throw new UnknownImageFormatException(sb.ToString());
         }
 
-        IImageEncoder? encoder = source.GetConfiguration().ImageFormatsManager.GetEncoder(format);
+        IImageEncoder? encoder = source.Configuration.ImageFormatsManager.GetEncoder(format);
 
         if (encoder is null)
         {
             StringBuilder sb = new();
             sb = sb.AppendLine(CultureInfo.InvariantCulture, $"No encoder was found for extension '{ext}' using image format '{format.Name}'. Registered encoders include:");
-            foreach (KeyValuePair<IImageFormat, IImageEncoder> enc in source.GetConfiguration().ImageFormatsManager.ImageEncoders)
+            foreach (KeyValuePair<IImageFormat, IImageEncoder> enc in source.Configuration.ImageFormatsManager.ImageEncoders)
             {
                 sb = sb.AppendFormat(CultureInfo.InvariantCulture, " - {0} : {1}{2}", enc.Key, enc.Value.GetType().Name, Environment.NewLine);
             }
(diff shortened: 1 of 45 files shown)


src/ImageSharp/Advanced/AdvancedImageExtensions.cs
```diff
@@ -76,30 +76,6 @@ public static void AcceptVisitor(this Image source, IImageVisitor visitor)
     public static Task AcceptVisitorAsync(this Image source, IImageVisitorAsync visitor, CancellationToken cancellationToken = default)
         => source.AcceptAsync(visitor, cancellationToken);
 
-    /// <summary>
-    /// Gets the configuration for the image.
-    /// </summary>
-    /// <param name="source">The source image.</param>
-    /// <returns>Returns the configuration.</returns>
-    public static Configuration GetConfiguration(this Image source)
```
tocsoft, line 84: we should probably leave these in despite then no longer being needed for backwards compatability, anyone with a dependency on this public API will break otherwise, shoudl mark them `Obsolete` instead

tocsoft (commented):
seems fine... only concern is the unnesorcery breaking of the public API. (`AdvancedImageExtensions.GetConfiguration(...)` overloads.

yes we are no longer dependent on them, but they are public api and some else probably is, anyone with custom encoders/decoders for example.

src/ImageSharp/Advanced/AdvancedImageExtensions.cs
antonfirsov: There are several breaking changes coming to 3.1 already without obsoletion eg. #2455 or #2485. I see little point using Obsolete selectively.

The net amount of changes between 3.0.* -> 3.1.0 seems to be much bigger than between 2.0.* -> 2.1.0 though. Maybe we should rename 3.1 to 4.0 to avoid confusion?

JimBobSquarePants: My plan with major versions has always been to support latest the .NET LTS version. There are breaking changes (_unfortunately more than I'd like_), but they typically only affect scenarios involving advanced behavior.

JimBobSquarePants:
@MichalStrehovsky I finally bit the bullet and installed the .NET 8 RC so I could use Sizoscope. With these changes and a tweak to the calling code from the sample project in #2486 we achieve much better results.

```c#
/// <summary>
/// Saves the given buffer using the filename extension as format.
/// </summary>
/// <param name="filename">The full pathname for the filename being written</param>
/// <param name="buffer">The image data stored as RGB buffer.</param>
/// <param name="width">The image width in pixels.</param>
/// <param name="height">The image height in pixels.</param>
public static void SaveImage(string filename, ReadOnlySpan<byte> buffer, int width, int height)
{
    // Pass custom Configuration to avoid loading all default formats.
    using var image = Image.LoadPixelData<Rgb24>(new(), buffer, width, height);
    using var stream = new FileStream(filename, FileMode.Create);
    var pngEncoder = new PngEncoder()
    {
        ColorType = PngColorType.Rgb
    };

    image.SaveAsPng(stream, pngEncoder);
}
```
image

We could probably drop the size by another 40-45kb if we refactored the way the quantizer is loaded in the png encoder, for some reason the compiler cannot trim out the construction of that type but that's beyond the scope of this PR.

tocsoft (approved):

MichalStrehovsky:
> @MichalStrehovsky I finally bit the bullet and installed the .NET 8 RC so I could use Sizoscope. With these changes and a tweak to the calling code from the sample project in https://github.com/SixLabors/ImageSharp/issues/2486 we achieve much better results.

Nice! How big is the resulting sample project? I'm guessing this is a ~30% size reduction? IIRC this app was about 5 MB before.

JimBobSquarePants:
> Nice! How big is the resulting sample project? I'm guessing this is a ~30% size reduction? IIRC this app was about 5 MB before.

The executable? Less than half if I'm following the Twitter thread correctly. 

image

MichalStrehovsky:
> > Nice! How big is the resulting sample project? I'm guessing this is a ~30% size reduction? IIRC this app was about 5 MB before.
> 
> The executable? Less than half if I'm following the Twitter thread correctly.
> 
> image

Better than expected! Looks like C#/.NET matches the [Go size](https://twitter.com/progtools/status/1672514510484021251) now!

JimBobSquarePants:
Fantastic stuff! I'll be looking at Sizoscope a lot for v4 to see if we can design the library internals to be as trimmable as possible. 😁

PR #2516, opened 2023-08-23
Handle EOF in Jpeg bit reader when data is bad to prevent DOS attack.

JimBobSquarePants:
### Prerequisites

### Description

This specially crafted file has had the start of frame marker altered to change the dimensions to values (_65503x65503_) that far exceed the encoded data. Nothing seems to be able to open this but with a very small change we actually can and quickly too. 

It's better that we allow decoding than explicitly throwing because there is no way to tell the difference between a maliciously crafted file with complete data from a damaged file with limited data and we already work very hard to enable decoding the latter.

I've also added a fix to allow us to actually parallel process images of this size. I discovered the iterator suffered from an overflow issue during attempts to determine what the maximum dimensions of a file Windows can handle (_65500x65500_) as it cannot decode a reencoded of the full dimensions with any built in applications.

I've created a branch (release/v3.0.2) based of the v3.0.1 tagged commit and cherry-picked #2501. Once merged I'll publish a release based upon that branch I'll backport this fix into release/v2.x for publishing.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -9,6 +9,7 @@ on:
   pull_request:
     branches:
       - main
+      - release/*
     types: [ labeled, opened, synchronize, reopened ]
 jobs:
   Build:
--- a/src/ImageSharp/Advanced/ParallelRowIterator.cs
+++ b/src/ImageSharp/Advanced/ParallelRowIterator.cs
@@ -115,7 +115,7 @@ public static partial class ParallelRowIterator
         int width = rectangle.Width;
         int height = rectangle.Height;
 
-        int maxSteps = DivideCeil(width * height, parallelSettings.MinimumPixelsProcessedPerTask);
+        int maxSteps = DivideCeil(width * (long)height, parallelSettings.MinimumPixelsProcessedPerTask);
         int numOfSteps = Math.Min(parallelSettings.MaxDegreeOfParallelism, maxSteps);
         MemoryAllocator allocator = parallelSettings.MemoryAllocator;
         int bufferLength = Unsafe.AsRef(operation).GetRequiredBufferLength(rectangle);
@@ -180,7 +180,7 @@ public static partial class ParallelRowIterator
         int width = rectangle.Width;
         int height = rectangle.Height;
 
-        int maxSteps = DivideCeil(width * height, parallelSettings.MinimumPixelsProcessedPerTask);
+        int maxSteps = DivideCeil(width * (long)height, parallelSettings.MinimumPixelsProcessedPerTask);
         int numOfSteps = Math.Min(parallelSettings.MaxDegreeOfParallelism, maxSteps);
 
         // Avoid TPL overhead in this trivial case:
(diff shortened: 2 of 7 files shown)


src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegBitReader.cs
```diff
@@ -212,7 +212,7 @@ public bool FindNextMarker()
     private int ReadStream()
     {
         int value = this.badData ? 0 : this.stream.ReadByte();
-        if (value == -1)
+        if (value == -1 || this.stream.Position >= this.stream.Length)
```
JimBobSquarePants, line 215: >Seeking to any location beyond the length of the stream is supported.

https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.position?view=net-7.0#remarks

JimBobSquarePants: Turns out the fix is bad. It's somehow preventing complete data reading.

.github/workflows/build-and-test.yml
```diff
@@ -9,6 +9,7 @@ on:
   pull_request:
     branches:
       - main
+      - release/*
```
JimBobSquarePants, line 12: Allow the trigger of the build for PRs against release branches.

tests/ImageSharp.Tests/Formats/Jpg/JpegDecoderTests.cs
```diff
@@ -314,4 +314,14 @@ public void Issue2315_DecodeWorks<TPixel>(TestImageProvider<TPixel> provider)
         image.DebugSave(provider);
         image.CompareToOriginal(provider);
     }
+
+    [Theory]
+    [WithFile(TestImages.Jpeg.Issues.HangBadScan, PixelTypes.L8)]
+    public void DecodeHang<TPixel>(TestImageProvider<TPixel> provider)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        using Image<TPixel> image = provider.GetImage(JpegDecoder.Instance);
```
JimBobSquarePants, line 323: There's no way to test this without actually decoding the image so I've tried to use as little memory as possible.

src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegBitReader.cs
JimBobSquarePants: Proper fix implemented now.

src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegBitReader.cs
```diff
@@ -212,7 +212,7 @@ public bool FindNextMarker()
     private int ReadStream()
     {
         int value = this.badData ? 0 : this.stream.ReadByte();
-        if (value == -1)
+        if (value == -1 || (this.badData && this.data == 0 && this.stream.Position >= this.stream.Length))
```
antonfirsov, line 215: Can describe the stuff behind `||` in a comment?

antonfirsov (approved):
LGTM, with a few recommendations. I assume we also want to land this change to `main`?

src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegBitReader.cs
JimBobSquarePants: Done

src/ImageSharp/Advanced/ParallelRowIterator.cs
```diff
@@ -50,7 +50,7 @@ public static void IterateRows<T>(Configuration configuration, Rectangle rectang
         int width = rectangle.Width;
         int height = rectangle.Height;
 
-        int maxSteps = DivideCeil(width * height, parallelSettings.MinimumPixelsProcessedPerTask);
+        int maxSteps = DivideCeil(width * (long)height, parallelSettings.MinimumPixelsProcessedPerTask);
```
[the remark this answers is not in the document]
JimBobSquarePants: I've kept it in for simplicities sake but added tests.

JimBobSquarePants:
I'll squash these commits when merging so I should be able to PR the v3.0.2 branch against main once done but if not, I'll be able to checrry pick the commit and PR that.

antonfirsov (approved):
:shipit:

PR #2519, opened 2023-08-25
Fix #2518

### LINKED ISSUE #2518: AccessViolationException during OilPaint
### Prerequisites

### ImageSharp version

3.0.1

None

Windows 10 and Debian GNU/Linux trixie/sid

### .NET Framework version

7.0

### Description

I have a small program that applies effects to multiple images in parallel, and after a small change I found that my program crashes quite often.

The exception produced is (Debug mode):
```cs
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at SixLabors.ImageSharp.Memory.MemoryGroup`1+Owned[[SixLabors.ImageSharp.PixelFormats.RgbaVector, SixLabors.ImageSharp, Version=3.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13]].Dispose()
   at SixLabors.ImageSharp.Memory.Buffer2D`1[[SixLabors.ImageSharp.PixelFormats.RgbaVector, SixLabors.ImageSharp, Version=3.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13]].Dispose()
   at SixLabors.ImageSharp.ImageFrame`1[[SixLabors.ImageSharp.PixelFormats.RgbaVector, SixLabors.ImageSharp, Version=3.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13]].Dispose(Boolean)
   at SixLabors.ImageSharp.ImageFrame.Dispose()
   at SixLabors.ImageSharp.ImageFrameCollection`1[[SixLabors.ImageSharp.PixelFormats.RgbaVector, SixLabors.ImageSharp, Version=3.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13]].Dispose(Boolean)
   at SixLabors.ImageSharp.ImageFrameCollection.Dispose()
   at SixLabors.ImageSharp.Image`1[[SixLabors.ImageSharp.PixelFormats.RgbaVector, SixLabors.ImageSharp, Version=3.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13]].Dispose(Boolean)
   at SixLabors.ImageSharp.Image.Dispose()
   at OilPaintRepro.Program.AttemptRepro(Int32)
   at System.Threading.Tasks.Parallel+<>c__DisplayClass19_0`1[[System.__Canon, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].<ForWorker>b__1(System.Threading.Tasks.RangeWorker ByRef, Int32, Boolean ByRef)
   at System.Threading.Tasks.TaskReplicator+Replica.Execute()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(System.Threading.Thread, System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(System.Threading.Tasks.Task ByRef, System.Threading.Thread)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool+WorkerThread.WorkerThreadStart()
   ```

### Steps to Reproduce

```cs
using SixLabors.ImageSharp;
using System.Threading.Tasks;

namespace OilPaintRepro
{
    public static class Program
    {
        public static void Main(string[] args)
        {
            Parallel.For(0, int.MaxValue, AttemptRepro);
        }

        public static void AttemptRepro(int attempt)
        {
            using var image = new Image<RgbaVector>(1920, 1200, new(127, 191, 255));
            using var oiled = image.Clone();
            image.Mutate(ctx => ctx.OilPaint());
        }
    }
}
```
It can takes a few seconds to a few minutes depending on your luck.

### Images

No images needed

antonfirsov:
### Prerequisites

### Description
Applying a (hopefully correct) bin buffer size, and eliminating `Unsafe` usage in a less critical processor we did not attempt to optimize with other safer techniques.

#### main
```
|     Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------- |---------:|--------:|--------:|------:|------:|------:|----------:|
| DoOilPaint | 201.7 ms | 4.01 ms | 5.07 ms |     - |     - |     - |     18 KB |
```

#### PR
```
|     Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------- |---------:|--------:|--------:|------:|------:|------:|----------:|
| DoOilPaint | 214.5 ms | 3.61 ms | 3.20 ms |     - |     - |     - |     18 KB |
```

I believe the 6% cost is worth the readability and memory safety benefits. We should keep `Unsafe` for the hot path in critical stuff (codecs, resize etc.). If we want to optimize this processor, we can try other techniques first, eg. vectorization.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Effects/OilPaintingProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Effects/OilPaintingProcessor{TPixel}.cs
@@ -35,13 +35,14 @@ internal class OilPaintingProcessor<TPixel> : ImageProcessor<TPixel>
     /// <inheritdoc/>
     protected override void OnFrameApply(ImageFrame<TPixel> source)
     {
+        int levels = Math.Clamp(this.definition.Levels, 1, 255);
         int brushSize = Math.Clamp(this.definition.BrushSize, 1, Math.Min(source.Width, source.Height));
 
         using Buffer2D<TPixel> targetPixels = this.Configuration.MemoryAllocator.Allocate2D<TPixel>(source.Size());
 
         source.CopyTo(targetPixels);
 
-        RowIntervalOperation operation = new(this.SourceRectangle, targetPixels, source.PixelBuffer, this.Configuration, brushSize >> 1, this.definition.Levels);
+        RowIntervalOperation operation = new(this.SourceRectangle, targetPixels, source.PixelBuffer, this.Configuration, brushSize >> 1, levels);
         ParallelRowIterator.IterateRowIntervals(
             this.Configuration,
             this.SourceRectangle,
@@ -117,7 +118,7 @@ internal class OilPaintingProcessor<TPixel> : ImageProcessor<TPixel>
                 Span<TPixel> sourceRowPixelSpan = this.source.DangerousGetRowSpan(y);
                 Span<TPixel> sourceRowAreaPixelSpan = sourceRowPixelSpan.Slice(this.bounds.X, this.bounds.Width);
 
-                PixelOperations<TPixel>.Instance.ToVector4(this.configuration, sourceRowAreaPixelSpan, sourceRowAreaVector4Span);
+                PixelOperations<TPixel>.Instance.ToVector4(this.configuration, sourceRowAreaPixelSpan, sourceRowAreaVector4Span, PixelConversionModifiers.Scale);
 
                 for (int x = this.bounds.X; x < this.bounds.Right; x++)
                 {
@@ -141,7 +142,7 @@ internal class OilPaintingProcessor<TPixel> : ImageProcessor<TPixel>
                             int offsetX = x + fxr;
                             offsetX = Numerics.Clamp(offsetX, 0, maxX);
 
-                            Vector4 vector = sourceOffsetRow[offsetX].ToVector4();
+                            Vector4 vector = sourceOffsetRow[offsetX].ToScaledVector4();
 
                             float sourceRed = vector.X;
                             float sourceBlue = vector.Z;
@@ -161,9 +162,9 @@ internal class OilPaintingProcessor<TPixel> : ImageProcessor<TPixel>
                             }
                         }
 
-                        float red = MathF.Abs(redBinSpan[maxIndex] / maxIntensity);
-                        float blue = MathF.Abs(blueBinSpan[maxIndex] / maxIntensity);
-                        float green = MathF.Abs(greenBinSpan[maxIndex] / maxIntensity);
+                        float red = redBinSpan[maxIndex] / maxIntensity;
+                        float blue = blueBinSpan[maxIndex] / maxIntensity;
+                        float green = greenBinSpan[maxIndex] / maxIntensity;
                         float alpha = sourceRowVector4Span[x].W;
 
                         targetRowVector4Span[x] = new Vector4(red, green, blue, alpha);
@@ -172,7 +173,7 @@ internal class OilPaintingProcessor<TPixel> : ImageProcessor<TPixel>
 
                 Span<TPixel> targetRowAreaPixelSpan = this.targetPixels.DangerousGetRowSpan(y).Slice(this.bounds.X, this.bounds.Width);
 
-                PixelOperations<TPixel>.Instance.FromVector4Destructive(this.configuration, targetRowAreaVector4Span, targetRowAreaPixelSpan);
+                PixelOperations<TPixel>.Instance.FromVector4Destructive(this.configuration, targetRowAreaVector4Span, targetRowAreaPixelSpan, PixelConversionModifiers.Scale);
             }
         }
     }
--- a/src/ImageSharp/Processing/Processors/Effects/OilPaintingProcessor{TPixel}.cs
+++ b/src/ImageSharp/Processing/Processors/Effects/OilPaintingProcessor{TPixel}.cs
@@ -43,10 +43,17 @@ internal class OilPaintingProcessor<TPixel> : ImageProcessor<TPixel>
         source.CopyTo(targetPixels);
 
         RowIntervalOperation operation = new(this.SourceRectangle, targetPixels, source.PixelBuffer, this.Configuration, brushSize >> 1, levels);
-        ParallelRowIterator.IterateRowIntervals(
+        try
+        {
+            ParallelRowIterator.IterateRowIntervals(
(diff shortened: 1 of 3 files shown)


src/ImageSharp/Processing/Processors/Effects/OilPaintingProcessor{TPixel}.cs
```diff
@@ -97,19 +98,19 @@ public void Invoke(in RowInterval rows)
              * are cleared for each loop iteration, to avoid the repeated allocation for each processed pixel. */
             using IMemoryOwner<Vector4> sourceRowBuffer = this.configuration.MemoryAllocator.Allocate<Vector4>(this.source.Width);
             using IMemoryOwner<Vector4> targetRowBuffer = this.configuration.MemoryAllocator.Allocate<Vector4>(this.source.Width);
-            using IMemoryOwner<float> bins = this.configuration.MemoryAllocator.Allocate<float>(this.levels * 4);
+            using IMemoryOwner<float> bins = this.configuration.MemoryAllocator.Allocate<float>(this.levels * 256 * 4);
```
JimBobSquarePants, line 101: Shouldn't we be clamping this in the extension method? I'm trying to wrap my head around how this actually works again.

JimBobSquarePants (commented):
Thanks for looking at this. Agreed regarding the indexers. 

How did you derive 256 as a multiplier?

JimBobSquarePants:
Here's the article it was based upon.

https://softwarebydefault.com/2013/06/29/oil-painting-cartoon-filter/

antonfirsov:
Yeah, according to the article we should divide by `255` to get the intensity bin index instead of defining 255x more buckets.

```C#
int currentIntensity = (int)MathF.Round((sourceBlue + sourceGreen + sourceRed) / 3F * this.levels / 255f);
```

But with this tests are failing and the output looks like a simple blur. Unfortunately I cannot go deeper understanding right now.

JimBobSquarePants:
I’ll have a look tonight

src/ImageSharp/Processing/Processors/Effects/OilPaintingProcessor{TPixel}.cs
```diff
@@ -140,29 +142,29 @@ public void Invoke(in RowInterval rows)
                             int offsetX = x + fxr;
                             offsetX = Numerics.Clamp(offsetX, 0, maxX);
 
-                            Vector4 vector = sourceOffsetRow[offsetX].ToVector4();
+                            Vector4 vector = Vector4.Clamp(sourceOffsetRow[offsetX].ToScaledVector4(), Vector4.Zero, Vector4.One);
```
antonfirsov, line 145: Unfortunately, `ToScaledVector4` doesn't guarantee that the vector will be clamped to `0..1` since a pixel implementation is free to return anything. (In fact `RgbaVector` doesn't clamp!)

With clamping I was able to undo the buffer overallocation in [99771d6](https://github.com/SixLabors/ImageSharp/pull/2519/commits/99771d61a87de9448698aa1368979d5d23e983c6).

I think the PR should be good now if you are also fine with the changes @JimBobSquarePants .

JimBobSquarePants: Shouldn’t we fix RgbaVector? ScaledVector is supposed to guarantee 0…1

JimBobSquarePants: Just had a look at several of our pixel formats. None of the sanitize the range in construction. We should fix that across the board.

antonfirsov: > None of the sanitize the range in construction

I wonder which are the pixel types where `ToScaledVector4()` could go out of range? To me it looks like it can happen only with floating point ones (`RgbaVector`, `HalfVector2|4`, `HalfSingle` ...) Do you see any other ones?

Sanitizing in the constructor is insufficient since (1) there are public fields/properties to manipulate the values (2) low-level `MemoryMarshal.Cast`-like manipulation is completely normal when working with pixel buffers and it can also push values out of range. If we want a guarantee that `ToScaledVector4()` implementations always return values in `0..1`, clipping should happen in `ToScaledVector4()`.

However, even if guarantee this in our own pixel types, arbitrary user pixel types can still potentially violate the contract. If it only leads to an `IndexOutOfRangeException`, we can blame it on the user. However, if we have unsafe indexing depending on the the values returned from `ToScaledVector4()` (like we used to do in OilPaint) we have a serious bug. We *may* have a similar problem here:

https://github.com/SixLabors/ImageSharp/blob/e81efa3ca649116282e3f58e76abeecbd625e482/src/ImageSharp/Processing/Processors/Normalization/AutoLevelProcessor%7BTPixel%7D.cs#L151-L153

My recommendation would be to revisit these processors before doing anything else.

JimBobSquarePants: All of the pixel formats accepting a `float` lack sanitation so `Short2`, `Short4`, and the `NormalizedXX` variants would be on the list also.

Scaled vector shouldn't clip. It should normalize the scale of the min max values of the pixel format. The mutability of the pixel formats makes sanitation difficult but as you say, avoiding unnecessary unsafe code turns a trivial exception to a serious one so we should definitely review and fix any processors that could be affected.

Regarding this PR. That clamp is going to hurt performance and _should_ be unnecessary given the correct `Vector4` range. I'd say we revert it and merge before continuing to sanitize the processors.

antonfirsov: See [ac0c1f9](https://github.com/SixLabors/ImageSharp/pull/2519/commits/ac0c1f903d7aa8583bf2eb4e29975f0a88bd5989). With aéé  the confusion this PR started with, it would make sense to squash on merge.

JimBobSquarePants (approved):
Yeah, this will do for now until we figure out a way to clamp pixel ranges in an efficient manner.

PR #2522, opened 2023-08-30
Handle EOF in Jpeg bit reader when data is bad to prevent DOS attack.

JimBobSquarePants:
### Prerequisites


src/ImageSharp/Formats/Pbm/BufferedReadStreamExtensions.cs
```diff
@@ -28,7 +28,7 @@ public static void SkipWhitespaceAndComments(this BufferedReadStream stream)
                 {
                     innerValue = stream.ReadByte();
                 }
-                while (innerValue != 0x0a);
+                while (innerValue is not 0x0a and not -0x1);
 
                 // Continue searching for whitespace.
                 val = innerValue;
```
Nikolais-afk, line 34: Fff

PR #2535, opened 2023-09-13
Fix for issue 2504 - IPTC and ICC profile information being lost during TIFF file save

JeffP134:
### Prerequisites

### Description
When testing adding and writing out IPTC tag data, it was found that the data was being lost during the save process for TIFF files.  It was also determined that the ICC profile data was also being lost.

During debugging, it was found that the Exif and XMP profiles were coming from the root frame only, and the IPTC and ICC profiles were coming from the image and not the root frame.

The changes introduced by this pull request split the writing of the image's metadata from each frame's metadata.  Also, rather than only using the profile data from the root frame, the metadata for each frame is written instead.  If multiple frames in the TIFF file have different metadata settings, that should preserve the data for each frame.

A new unit test was added to store and retrieve IPTC and ICC profile data, to verify that they were being saved and loaded correctly.

### DIFF
--- a/src/ImageSharp/Formats/Tiff/TiffEncoderCore.cs
+++ b/src/ImageSharp/Formats/Tiff/TiffEncoderCore.cs
@@ -157,6 +157,7 @@ internal sealed class TiffEncoderCore : IImageEncoderInternals
         long ifdMarker = WriteHeader(writer, buffer);
 
         Image<TPixel> metadataImage = image;
+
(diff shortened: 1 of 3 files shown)


JeffP134:
Stylecop extra line issue should now be fixed.

The code duplication was bugging me, so I refactored the code to extract out common methods as well.

JimBobSquarePants (approved):
This all looks fantastic! Thanks so much for your help here. 👍

PR #2549, opened 2023-10-10
Switch to Bottom as upper bound for Quantizer Rect

mrange:
Hi.

I was trying out Quantizer with custom rects. I couldn't get it to work and checked the code. To me it seems to be a typo where the `interest.Height` was used rather than `interest.Bottom`. The x-dimension `interest.Left` and `interest.Right` is used which makes sense to me.

I did the change and it worked for me but then when I tried running the tests I got multiple image comparison failures.

So perhaps this is the intended behavior and I misunderstood the API. Or it is a good fix and oracles should be regenerated (but I don't know how).

I don't really know what is the right choice here so I thought I create a draft pull request even though the tests fails and you can tell me what do or just close it if it works as intended.

Regards,.


mrange:
Read the code a bit closer today and it seems I need to think it through one more. Closing for now. Still think there's something wrong with original (but could be wrong)

mrange:
Thinking...

JimBobSquarePants:
@mrange There is an issue, but a fix needs to be applied in two places. I'm working on the fix now.

PR #2550, opened 2023-10-13
Tiff decoding robustness improvements

JimBobSquarePants:
### Prerequisites

### Description

Adds additional checks to the decoder infrastructure to enable us to better handle corrupt files. 

Some light formatting cleanup also.

### DIFF
--- a/src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
+++ b/src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
@@ -161,6 +161,11 @@ internal sealed class ZlibInflateStream : Stream
             bytesToRead = Math.Min(count - totalBytesRead, this.currentDataRemaining);
             this.currentDataRemaining -= bytesToRead;
             bytesRead = this.innerStream.Read(buffer, offset, bytesToRead);
+            if (bytesRead == 0)
+            {
+                return totalBytesRead;
+            }
+
             totalBytesRead += bytesRead;
         }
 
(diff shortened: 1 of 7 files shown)


src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
```diff
@@ -161,29 +161,25 @@ public override int Read(byte[] buffer, int offset, int count)
...
+            if (bytesRead == 0)
+            {
+                return totalBytesRead;
+            }
+
             totalBytesRead += bytesRead;
         }
 
         return totalBytesRead;
     }
```
antonfirsov, line 173: Since we are already touching this class: we should check the return value of `this.innerStream.Read` in `InitializeInflateStream` when it's skipping 4 bytes.

antonfirsov (approved):
One suggestion, otherwise LGTM.

src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
antonfirsov: Pushed the change together with another check.

tests/ImageSharp.Tests/Formats/Tiff/TiffDecoderTests.cs
```diff
@@ -666,6 +665,33 @@ public void TiffDecoder_CanDecode_Fax4CompressedWithStrips<TPixel>(TestImageProv
...
+    [WithFile(Tiled0000023664, PixelTypes.Rgba32)]
+    public void TiffDecoder_CanDecode_TiledWithBadZlib<TPixel>(TestImageProvider<TPixel> provider)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        using Image<TPixel> image = provider.GetImage(TiffDecoder.Instance);
+
+        // ImageMagick cannot decode this image.
+        image.DebugSave(provider);
+        image.CompareToReferenceOutput(
+            ImageComparer.Exact,
```
antonfirsov, line 690: Isn't there uninialized memory garbage after the exit criteria has been reached? If so, this test might be flaky. Or is there some logic in tiff that guarantees those scanlines it to be black in tiff?

(If not, clearing it out would be beyond the scope of this fix.)

JimBobSquarePants: Each frame is decoded into a clean buffer.

image

src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
JimBobSquarePants: Good changes. Thanks!

tests/ImageSharp.Tests/Formats/Tiff/TiffDecoderTests.cs
antonfirsov: I would prefer it to somehow create frames with unintialized buffers, but that's a topic for another day.

JimBobSquarePants: Me too but I didn’t want to risk greater changes. There’s too much going on in the Tiff decoder I don’t fully understand

PR #2551, opened 2023-10-14
PBM decoder robustness improvements and BufferedReadStream observability

antonfirsov:
### Prerequisites

### Description

Add additonal checks to handle corrupt files better in the PBM decoder.

Also extend `BufferedReadStream` to monitor the number of times it has been made to hit EOF by read calls. This improves the testability of decoder behavior. The perf impact of the `BufferedReadStream` change is [within the margin of error](https://gist.github.com/antonfirsov/456c748e5bf1dc52c41f1e416560e4f1).

### DIFF
--- a/src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
+++ b/src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
@@ -71,7 +71,11 @@ internal class BinaryDecoder
 
         for (int y = 0; y < height; y++)
         {
-            stream.Read(rowSpan);
+            if (stream.Read(rowSpan) == 0)
+            {
+                return;
+            }
+
             Span<TPixel> pixelSpan = pixels.DangerousGetRowSpan(y);
             PixelOperations<TPixel>.Instance.FromL8Bytes(
                 configuration,
@@ -93,7 +97,11 @@ internal class BinaryDecoder
 
         for (int y = 0; y < height; y++)
         {
-            stream.Read(rowSpan);
+            if (stream.Read(rowSpan) == 0)
+            {
+                return;
+            }
+
             Span<TPixel> pixelSpan = pixels.DangerousGetRowSpan(y);
             PixelOperations<TPixel>.Instance.FromL16Bytes(
                 configuration,
@@ -115,7 +123,11 @@ internal class BinaryDecoder
 
         for (int y = 0; y < height; y++)
         {
-            stream.Read(rowSpan);
+            if (stream.Read(rowSpan) == 0)
+            {
+                return;
+            }
+
             Span<TPixel> pixelSpan = pixels.DangerousGetRowSpan(y);
             PixelOperations<TPixel>.Instance.FromRgb24Bytes(
                 configuration,
@@ -137,7 +149,11 @@ internal class BinaryDecoder
 
         for (int y = 0; y < height; y++)
         {
-            stream.Read(rowSpan);
+            if (stream.Read(rowSpan) == 0)
+            {
+                return;
+            }
+
             Span<TPixel> pixelSpan = pixels.DangerousGetRowSpan(y);
             PixelOperations<TPixel>.Instance.FromRgb48Bytes(
                 configuration,
@@ -161,6 +177,11 @@ internal class BinaryDecoder
             for (int x = 0; x < width;)
             {
                 int raw = stream.ReadByte();
+                if (raw < 0)
+                {
+                    return;
+                }
+
                 int stopBit = Math.Min(8, width - x);
                 for (int bit = 0; bit < stopBit; bit++)
                 {
--- a/src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
+++ b/src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
@@ -145,18 +145,18 @@ internal sealed class PbmDecoderCore : IImageDecoderInternals
                 throw new InvalidImageContentException("Unknown of not implemented image type encountered.");
         }
 
-        if (!stream.TrySkipWhitespaceAndComments() ||
-            !stream.TryReadDecimal(out int width) ||
-            !stream.TrySkipWhitespaceAndComments() ||
-            !stream.TryReadDecimal(out int height) ||
-            !stream.TrySkipWhitespaceAndComments())
+        if (!stream.SkipWhitespaceAndComments() ||
+            !stream.ReadDecimal(out int width) ||
+            !stream.SkipWhitespaceAndComments() ||
+            !stream.ReadDecimal(out int height) ||
+            !stream.SkipWhitespaceAndComments())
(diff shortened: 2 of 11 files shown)


tests/ImageSharp.Tests/Formats/Pbm/PbmMetadataTests.cs
```diff
@@ -83,12 +83,9 @@ public void Identify_DetectsCorrectComponentType(string imagePath, PbmComponentT
...
     [Fact]
-    public void Identify_HandlesCraftedDenialOfServiceString()
+    public void Identify_EofInHeader_ThrowsInvalidImageContentException()
     {
         byte[] bytes = Convert.FromBase64String("UDEjWAAACQAAAAA=");
-        ImageInfo info = Image.Identify(bytes);
-        Assert.Equal(default, info.Size);
-        Configuration.Default.ImageFormatsManager.TryFindFormatByFileExtension("pbm", out IImageFormat format);
-        Assert.Equal(format!, info.Metadata.DecodedImageFormat);
+        Assert.Throws<InvalidImageContentException>(() => Image.Identify(bytes));
```
antonfirsov, line 89: It doesn't really matter but I don't think decoding makes much sense when the EOF is right in the header.

JimBobSquarePants: Yeah, I'm happy with that.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
```diff
@@ -144,14 +145,22 @@ private void ProcessHeader(BufferedReadStream stream)
                 throw new InvalidImageContentException("Unknown of not implemented image type encountered.");
         }
 
-        stream.SkipWhitespaceAndComments();
-        int width = stream.ReadDecimal();
-        stream.SkipWhitespaceAndComments();
-        int height = stream.ReadDecimal();
-        stream.SkipWhitespaceAndComments();
+        if (!stream.SkipWhitespaceAndComments() ||
```
JimBobSquarePants, line 148: I really like this!

JimBobSquarePants: Only thing I would suggest is to change the method names to use `Try` prefix to match convention.

src/ImageSharp/Formats/Pbm/PlainDecoder.cs
```diff
@@ -65,20 +65,30 @@ private static void ProcessGrayscale<TPixel>(Configuration configuration, Buffer
...
 
+        bool eofReached = false;
         for (int y = 0; y < height; y++)
         {
             for (int x = 0; x < width; x++)
             {
-                byte value = (byte)stream.ReadDecimal();
-                stream.SkipWhitespaceAndComments();
-                rowSpan[x] = new L8(value);
+                stream.ReadDecimal(out int value);
```
JimBobSquarePants, line 73: Shouldn't we test here?

JimBobSquarePants: I guess it doesn't matter since we're simply assigning 0.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
JimBobSquarePants: I've pushed those changes.

JimBobSquarePants (approved):
Very nice. Liking the EOF counter changes!

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
JimBobSquarePants: Ha, didn't realize you'd explicitly removed them! I can revert at a later time if you feel strongly about it.

antonfirsov: I really prefer it to be the other way.

By convention, `TryDoSomething(...) == false` should indicate that the operation failed as a whole. This is not the case with `TryReadDecimal(out value) == false` when it happens at the very las digit in the file: the decoding of the digit was succesful (we should threat the result as valid!) just the file reached an EOF.

I like following such conventions to the letter, because ambiguity can lead to a misunderstandings and programming mistakes.

src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
```diff
@@ -71,7 +71,11 @@ private static void ProcessGrayscale<TPixel>(Configuration configuration, Buffer
 
         for (int y = 0; y < height; y++)
         {
-            stream.Read(rowSpan);
+            if (stream.Read(rowSpan) == 0)
```
antonfirsov, line 74: This is lame actually. I should have sliced down `rowSpan` with the result of `stream.Read()`, for the pixel conversion done later, or used the condition `stream.Read(rowSpan) < rowSpan.Length` (the latter isstricter, but simpler).

This has little practical relevance (`BuffferedReadStream` is hitting EOF twice & `FromL8Bytes` is converting some memory garbage for broken files), so no need to fix it now, but the code looks stupid :)

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
JimBobSquarePants: I’m not precious about it (and happy to revert) but there’s something a little clunky about the method there. We’re really returning two things, the decimal and the stream state. I’ve seen something in the runtime I’m sure that handles it better (Maybe inside guid parsing code)

src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
JimBobSquarePants: We can improve it with follow up.

src/ImageSharp/Formats/Pbm/PlainDecoder.cs
antonfirsov: No, when it's the last digit of the file the value is valid! For simplicity, I'm ignoring the retval here and letting the `SkipWhitespaceAndComments` call below to detect EOF.

src/ImageSharp/Formats/Pbm/PbmDecoderCore.cs
antonfirsov: I considered returning a `(bool, int)` tuple, but that would lead to much more complicated code on the callsites, since we couldn't do `!stream.SkipWhitespaceAndComments() || ...`. Defining the names without the `Try` prefix + documenting the behavior was the best idea I was able to come up with.

src/ImageSharp/Formats/Pbm/BinaryDecoder.cs
antonfirsov: #2552

PR #2561, opened 2023-10-17
Png- Do not attempt to read data for chunks of length 0.

JimBobSquarePants:
### Prerequisites

### Description

Png was attempting to read chunk data for chunks of length `0`, for example `EOF`. This lead, to multiple attempts to read past the end of a stream.


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -1569,6 +1570,11 @@ private void SkipChunkDataAndCrc(in PngChunk chunk)
     [MethodImpl(InliningOptions.ShortMethod)]
     private IMemoryOwner<byte> ReadChunkData(int length)
     {
+        if (length == 0)
```
JimBobSquarePants, line 1573: We avoid the read here because Png still has to do a CRC check which if does not occur leads to failure during reading of chained Png images within a single stream.

antonfirsov (commented):
The behavior with these changes is still odd: we run through the stream with ~500 attempts to read a 0-byte chunk. It doesn't look like the correct/optimal thing to do for such an input.

This code doesn't look ok to me:
https://github.com/SixLabors/ImageSharp/blob/d93bc6cca2383095c2a08db85861b88a1c94c43c/src/ImageSharp/Formats/Png/PngDecoderCore.cs#L1470-L1479

If  `length > (this.currentStream.Length - this.currentStream.Position)` because the file is truncated, we just start interpreting the following (type/data) bytes as length until `length` becomes 0.

JimBobSquarePants:
> The behavior with these changes is still odd: we run through the stream with ~500 attempts to read a 0-byte chunk. It doesn't look like the correct/optimal thing to do for such an input.
> 
> This code doesn't look correct to me:
> 
> https://github.com/SixLabors/ImageSharp/blob/d93bc6cca2383095c2a08db85861b88a1c94c43c/src/ImageSharp/Formats/Png/PngDecoderCore.cs#L1470-L1479
> 
> If `length > (this.currentStream.Length - this.currentStream.Position)` because the file is truncated, we just start interpreting the following (type/data) bytes as length until `length` becomes 0. If we really don't want to throw/abort here, why not to process the chunk in the truncated form instead?
> 
> _Edit: I guess there was some corrupt input this was logic was introduced for but is this really the best we can do?_

Is it the best we can do? 

We differ from libpng here. We continue reading after parsing a length greater than 2147483647 while libpng will throw.

https://github.com/glennrp/libpng/blob/f2294569cfe43f87c9f72d35545e8de1126eb4b9/pngrutil.c#L22-L30  

libpng doesn't seem to do the inner check.
https://github.com/glennrp/libpng/blob/libpng16/pngrutil.c#L156-L192

However, it seems to freely loop like we do. I'm not sure there's a better way to detect chunks while skipping corrupted sections.
https://github.com/glennrp/libpng/blob/f2294569cfe43f87c9f72d35545e8de1126eb4b9/pngread.c#L106-L262

JimBobSquarePants:
> // Not a valid chunk so try again until we reach a known chunk.

I recked our suite of tests while disabling that section and we only had the one failing image. I don't think we'll hit that path often at all and it's the only way I think that we can be skip bad sections.

Issue #2565, opened 2023-10-19
Bug ERROR Unknown App0 Marker - Expected JFIF ProfileResolver OSQid


JoseEliasSantos:
Working with camera images, I ended up getting the error: System.NotSupportedException: 'Unknown App0 Marker - Expected JFIF.'. Looking at the latest issues I noticed that I was having a similar problem, I verified that the bytes referring to the maker of the file in question are: '\n', '[', 'I', 'D' and ' '. I added the ProfileResolver class to the OSQidMaker property and with that I was successful in loading the image, I also added a test with the image that generated the mentioned error.

_Originally posted by @JoseEliasSantos in https://github.com/SixLabors/ImageSharp/discussions/2564_

JoseEliasSantos:
I can't push with my corrected branch getting the error remote: Permission to SixLabors/ImageSharp.git denied to JoseEliasSantos.

JimBobSquarePants:
You must fork a repository, make the changes in your clone and then open a Pull Request. 

https://docs.github.com/en/get-started/quickstart/fork-a-repo

How did you create this issue without using the predefined template?

PR #2566, opened 2023-10-19
[fix]: Fixed Unknown App0 Marker add profileResolver missing

JoseEliasSantos:
### Prerequisites

### Description
Fixed discussion and issue bug
`https://github.com/SixLabors/ImageSharp/discussions/2564`
`https://github.com/SixLabors/ImageSharp/issues/2565`


src/ImageSharp/Formats/Jpeg/Components/Decoder/JFifMarker.cs
```diff
@@ -73,7 +73,8 @@ public static bool TryParse(ReadOnlySpan<byte> bytes, out JFifMarker marker)
     {
         // Some images incorrectly use JFXX as the App0 marker (Issue 2478)
         if (ProfileResolver.IsProfile(bytes, ProfileResolver.JFifMarker)
-            || ProfileResolver.IsProfile(bytes, ProfileResolver.JFxxMarker))
+            || ProfileResolver.IsProfile(bytes, ProfileResolver.JFxxMarker)
+            || ProfileResolver.IsProfile(bytes, ProfileResolver.OSQidMaker))
```
JimBobSquarePants, line 77: While this works, I'm not sure it's the correct fix. 

I think we might be able to simply remove the throw when the marker doesn't match a JFIF one and skip the remaining bytes. 

What is OSQidMaker? Can you please supply more information?

src/ImageSharp/Formats/Jpeg/Components/Decoder/ProfileResolver.cs
```diff
@@ -24,6 +24,14 @@ internal static class ProfileResolver
         (byte)'J', (byte)'F', (byte)'X', (byte)'X', (byte)'\0'
     };
 
+    /// <summary>
+    /// Gets the \n[ID or 10 91 73 68 32 specific markers.
+    /// </summary>
+    public static ReadOnlySpan<byte> OSQidMaker => new[]
+    {
+        (byte)'\n', (byte)'[', (byte)'I', (byte)'D', (byte)' '
+    };
```
gfoidl, line 33:
```suggestion
    public static ReadOnlySpan<byte> OSQidMaker => new[] { (byte)'\n', (byte)'[', (byte)'I', (byte)'D', (byte)' ' };
```
I'd like to have it written as one-liner, because when we are able to use ut8-literals it's easier to find & replace these.

src/ImageSharp/Formats/Jpeg/Components/Decoder/JFifMarker.cs
JoseEliasSantos: In fact, in my research I didn't find markers that specified such a conversion starting with these bytes, but I have images of a camera that always has these bytes at the beginning. I also had success in all tests when ignoring this if but I don't know what more problems this could cause. I'm running a fork of the lib in my application and I don't get any more errors

src/ImageSharp/Formats/Jpeg/Components/Decoder/ProfileResolver.cs
JoseEliasSantos: It's a good suggestion, but the other properties of the class already follow this same pattern, so I chose to keep it.

src/ImageSharp/Formats/Jpeg/Components/Decoder/JFifMarker.cs
JimBobSquarePants: The JFIF JFXX headers contain metadata and we can actually operate without them. Your images actually contain a JFIF header in an APP0 block just after the camera junk. 

I can’t accept the fix as-is because the values are non-standard so we should try a more robust solution as suggested

JimBobSquarePants:
Hi @JoseEliasSantos I can see you updated your PR from main. As commented, we won't be able to proceedhttps://github.com/SixLabors/ImageSharp/pull/2566#discussion_r1372248254 with the fix you have applied. If you are not able to make the suggested changes, that's OK. I can close this and create a new PR.

JimBobSquarePants (approved):
I've made the required changes here to make JFIF marker optional. Thanks for your assistance here 🙂

PR #2579, opened 2023-11-08
Add ICO and CUR file decoder.

frg2089:
### Prerequisites

### Description
- Make the Bmp Decoder Core to support ICO's data.
- :tada: Add ICO Encoder.
- Add ICO Decoder.
- Add ICO Image Format Detector.
- :tada: Add CUR Encoder.
- Add CUR Decoder.
- Add CUR Image Format Detector.

### Some Questions

> [!WARNING]
> We don't support bmp palettes yet.


Poker-sang:
Even though there is no magic head for ICOs, should we still provide `IcoImageFormatDetector` as well? In order to support `Image.Load<TPixel>()`.

src/ImageSharp/Image.cs
```diff
@@ -25,12 +25,14 @@ public abstract partial class Image : IDisposable, IConfigurationProvider
     /// <param name="pixelType">The pixel type information.</param>
     /// <param name="metadata">The image metadata.</param>
     /// <param name="size">The size in px units.</param>
-    protected Image(Configuration configuration, PixelTypeInfo pixelType, ImageMetadata metadata, Size size)
+    /// <param name="forceFramesSize">force frames size.</param>
+    protected Image(Configuration configuration, PixelTypeInfo pixelType, ImageMetadata metadata, Size size, bool forceFramesSize = true)
```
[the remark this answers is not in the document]
JimBobSquarePants: No. This will break multiple processors. You should always draw smaller pixel data over the larger frame.

JimBobSquarePants:
> Even though there is no magic head for ICOs, should we still provide `IcoImageFormatDetector` as well? In order to support `Image.Load<TPixel>()`.
> 
> And should we name the classes with `Ico` instead of `Icon`?

How do you identify without leading bytes?

Poker-sang:
Maybe we can combine an `ICODir` and an `entry`, and think of it as the header of an ICO file.

Or we just provide an decoder, but this decoder will not be used automatically.

frg2089:
> Maybe we can combine an `ICODir` and an `entry`, and think of it as the header of an ICO file.
> 
> Or we just provide a decoder, but this decoder will not be used without specification.

I think we can do without providing `IconImageFormatDetector`.
Because nobody cares if unknown streams are ICOs. They are usually not transmitted without filename.
So we should just provide a decoder and use it directly.

like this
```csharp
using SixLabors.ImageSharp.Formats.Icon;

async using var fs = File.OpenRead(@"xxx.ico");
var ico = IconDecoder.Instance.Decode(new (), fs);
```

src/ImageSharp/Image.cs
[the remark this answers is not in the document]
frg2089: Could we make a special image type.
Some special image formats allow storing multiple images of different sizes inside them, which we can decode into a `MultiImage<TPixel>`.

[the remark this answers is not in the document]
JimBobSquarePants: No. The frames must remain the same dimensions. Encoders need to handle clipping on save.

Poker-sang:
> We actually have an open discussion supporting tiffs with multisize frames. It would be good to split any changes to the tiff decoder out so we can draw the pixel data on the frame.

@JimBobSquarePants Yes. I have already reverted all changes to tiff files.

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// Specifies number of colors in the color palette. Should be 0 if the image does not use a color palette.
+    /// </summary>
+    public byte ColorCount { get; set; }
+
+    /// <summary>
+    /// Gets or sets Planes field. <br />
+    /// In ICO format: Specifies color planes. Should be 0 or 1. <br />
+    /// In CUR format: Specifies the horizontal coordinates of the hotspot in number of pixels from the left.
+    /// </summary>
+    public ushort Planes { get; set; }
```
JimBobSquarePants, line 85: What does Planes actually mean? Should it be stored in metadata?

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// In ICO format: Specifies bits per pixel. <br />
+    /// In CUR format: Specifies the vertical coordinates of the hotspot in number of pixels from the top.
+    /// </summary>
+    public ushort BitCount { get; set; }
+
+    /// <summary>
+    /// Gets or sets BytesInRes field. <br />
+    /// Specifies the size of the image's data in bytes
+    /// </summary>
+    public uint BytesInRes { get; set; }
```
JimBobSquarePants, line 98: Is this useful? I don't think so. What if the size/format is changed?

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// <summary>
+    /// Gets or sets ImageOffset field. <br />
+    /// Specifies the offset of BMP or PNG data from the beginning of the ICO/CUR file
+    /// </summary>
+    public uint ImageOffset { get; set; }
```
JimBobSquarePants, line 104: Why do we care about this?

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// <inheritdoc/>
+    public IDeepCloneable DeepClone() => new IcoFrameMetadata(this);
+
+    internal void SetMetadata(in IcoDirEntry metadata)
```
JimBobSquarePants, line 109: You used `FromXX` syntax in png. I'd like to be consistent.

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// In CUR format: Specifies the horizontal coordinates of the hotspot in number of pixels from the left.
+    /// </summary>
+    public ushort Planes { get; set; }
+
+    /// <summary>
+    /// Gets or sets BitCount field. <br />
+    /// In ICO format: Specifies bits per pixel. <br />
+    /// In CUR format: Specifies the vertical coordinates of the hotspot in number of pixels from the top.
+    /// </summary>
+    public ushort BitCount { get; set; }
```
JimBobSquarePants, line 92: Ambiguity like this makes me want to split the formats out into separate ones with shared utility methods.

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// Gets or sets Height field. <br />
+    /// Specifies image height in pixels. Can be any number between 0 and 255. Value 0 means image height is 256 pixels.
+    /// </summary>
+    public byte Height { get; set; }
+
+    /// <summary>
+    /// Gets or sets ColorCount field. <br />
+    /// Specifies number of colors in the color palette. Should be 0 if the image does not use a color palette.
+    /// </summary>
+    public byte ColorCount { get; set; }
```
JimBobSquarePants, line 78: Why not store the palette itself like gif/png?

src/ImageSharp/Formats/Ico/IcoMetadata.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="IcoMetadata"/> class.
+    /// </summary>
+    /// <param name="reserved">always be <see langword="0" /></param>
+    /// <param name="type">
+    /// ICO file is <see langword="1" />.
+    /// CUR file is <see langword="2" />.
+    /// </param>
+    /// <param name="count">Image Count</param>
+    public IcoMetadata(ushort reserved, ushort type, ushort count)
```
JimBobSquarePants, line 28: I definitely think we should split the formats out now. This is unwise.

src/ImageSharp/Formats/Ico/IcoMetadata.cs
```diff
@@ -0,0 +1,70 @@
...
+        this.Reserved = metadata.Reserved;
+        this.Type = metadata.Type;
+        this.Count = metadata.Count;
+    }
+
+    /// <summary>
+    /// Gets or sets Reserved field. <br />
+    /// Must always be <see langword="0" />
+    /// </summary>
+    public ushort Reserved { get; set; }
```
JimBobSquarePants, line 46: If it's always 0 then we never need to store it.

src/ImageSharp/Formats/Ico/IcoMetadata.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// <see langword="1" /> for icon (.ICO) image, <br />
+    /// <see langword="2" /> for cursor (.CUR) image. <br />
+    /// Other values are invalid.
+    /// </summary>
+    public ushort Type { get; set; }
+
+    /// <summary>
+    /// Gets or sets Image Count.
+    /// </summary>
+    public ushort Count { get; set; }
```
JimBobSquarePants, line 59: The count of what? Frames? We don't need this.

JimBobSquarePants (changes requested):
Thanks for this, your efforts are really appreciated! There's a lot to discuss. 

Some notes...

I don't think it's a good idea to merge the two formats. We're creating unnecessary ambiguity. If you look at format metadata, we avoid exposing unconstrained configuration values. 

It's imperative that the pixel data is drawn on the frame and not resized. Position 0,0, seems sane as an origin. Any other behavior such as resizing or introducing new image types or editing existing signifies a major breaking change for the library.

We need an encoder. I can't accept code against main without one as we need to ensure that we have covered encoding requirements in the format design.

I'll be honest and say this won't get merged before V4. There's a lot to do to ensure that the decoder is safe (I haven't checked bounds handling and EOF detection) and the design practical.

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
```diff
@@ -0,0 +1,119 @@
...
+        this.BytesInRes = metadata.BytesInRes;
+        this.ImageOffset = metadata.ImageOffset;
+        this.BmpMetadata = metadata.BmpMetadata;
+        this.PngMetadata = metadata.PngMetadata;
+    }
+
+    /// <summary>
+    /// Gets or sets BmpMetadata.
+    /// </summary>
+    public Bmp.BmpMetadata? BmpMetadata { get; set; }
```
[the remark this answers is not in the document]
frg2089: But I don't know how I can make the file metadata into frame metadata in any other way.

src/ImageSharp/Formats/Ico/IcoConstants.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </remarks>
+    public static readonly IEnumerable<string> MimeTypes = new[]
+    {
+        // IANA-registered
+        "image/vnd.microsoft.icon",
+
+        // ICO & CUR types used by Windows
+        "image/x-icon",
+
+        // "image/x-win-cursor",
```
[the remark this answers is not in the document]
frg2089: I don't think application/octet-stream is a MIME Type exclusively for the .cur format. any binary stream can be application/octet-stream, so I didn't list it here.

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
frg2089: Because the BmpMetadata does not support color palettes yet.
https://github.com/SixLabors/ImageSharp/blob/afe21339afb60061de8a942e1262cac230e6b6a2/src/ImageSharp/Formats/Bmp/BmpMetadata.cs#L41

src/ImageSharp/Formats/Ico/IcoConstants.cs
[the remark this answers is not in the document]
JimBobSquarePants: It's not. It's a fallback for uncategorized file types. There doesn't appear to be a registered mime-type for cur so we would use this one.

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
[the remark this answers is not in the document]
JimBobSquarePants: You store the bmp/png metadata separately as their own entries the same way as the explicit decoders do. In IcoFrameMetadata you would use an enum to specify which format is used for encoding.

```
IcoCompression
{
    Bmp,

    Png
}
```

src/ImageSharp/Formats/Ico/IcoFrameMetadata.cs
JimBobSquarePants: We should store and reuse the palette when available like we do for Gif/Png. It's been on my TODO list for a while. 

https://github.com/SixLabors/ImageSharp/issues/2404

However, this should be a separate PR.

src/ImageSharp/Formats/Ico/Common/IcoLikeDecoderCore.cs
```diff
@@ -0,0 +1,158 @@
...
+            _ = stream.Seek(this.icoDirEntries[i].ImageOffset, SeekOrigin.Begin);
+
+            int size = stream.Read(flag);
+            _ = stream.Seek(-size, SeekOrigin.Current);
+
+            bool isPng = flag.SequenceEqual(Png.PngConstants.HeaderBytes);
+
+            Image<TPixel> img = this.GetDecoder(isPng).Decode<TPixel>(stream, cancellationToken);
+            IcoAssert.NotSquare(img.Size);
+            frames.Add((img, isPng, i));
```
JimBobSquarePants, line 47: We should be decoding each image in full here so we can handle early EOF. 

I would add an internal method to `ImageFrameCollection<T>` called `ExtractFrameUnsafe()` (_unsafe because it can leave an image containing a single frame in an unusable state_) which allows removing a frame from an image at a given index. You can then dispose of the parent safely. 

A major advantage to this other than the lack of copying (_and why I keep pushing you to use the original metadata_)  is that you already have the bmp or png metadata in the image frame. You only now need to add your ico/cur metadata

src/ImageSharp/Formats/Ico/Common/IcoLikeDecoderCore.cs
```diff
@@ -0,0 +1,158 @@
...
+            ImageFrame<TPixel> source = i.Image.Frames.RootFrameUnsafe;
+            for (int h = 0; h < source.Height; h++)
+            {
+                source.PixelBuffer.DangerousGetRowSpan(h).CopyTo(target.PixelBuffer.DangerousGetRowSpan(h));
+            }
+
+            IcoFrameMetadata frameMetadata;
+            if (i.IsPng)
+            {
+                frameMetadata = target.Metadata.GetIcoMetadataWithPng();
```
JimBobSquarePants, line 67: No need for these extensions. See above

src/ImageSharp/Formats/Ico/Common/IcoLikeDecoderCore.cs
```diff
@@ -0,0 +1,158 @@
...
+            if (stream.Read(new Span<byte>(ptr, size.Value)) != size)
+            {
+                throw new EndOfStreamException();
+            }
+
+            return size.Value;
+        }
+    }
+
+    private IImageDecoderInternals GetDecoder(bool isPng)
```
JimBobSquarePants, line 127: No need for this use an explicit call to the correct decoder when required.

src/ImageSharp/Formats/Ico/Common/IcoLikeFrameCompression.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Six Labors.
+// Licensed under the Six Labors Split License.
+
+namespace SixLabors.ImageSharp.Formats.Ico.Common;
+
+/// <summary>
+/// IcoFrameCompression
+/// </summary>
+public enum IcoLikeFrameCompression
```
JimBobSquarePants, line 9: Use separate types. That they use the same compression approach is an implementation detail.

src/ImageSharp/Formats/Ico/Common/IcoLikeFrameMetadata.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// <summary>
+    /// Gets or sets Width field. <br />
+    /// Specifies image width in pixels. Can be any number between 0 and 255. Value 0 means image width is 256 pixels.
+    /// </summary>
+    public byte EncodingWidth { get; set; }
+
+    /// <inheritdoc/>
+    public abstract IDeepCloneable DeepClone();
+
+    internal void FormIcoDirEntry(in IcoDirEntry metadata)
```
JimBobSquarePants, line 86: From

src/ImageSharp/Formats/Ico/IcoBmpFrameMetadata.cs
```diff
@@ -0,0 +1,67 @@
...
+
+using SixLabors.ImageSharp.Formats.Bmp;
+using SixLabors.ImageSharp.Formats.Ico.Common;
+
+namespace SixLabors.ImageSharp.Formats.Ico;
+
+/// <summary>
+/// Provides Ico specific bmp metadata information for the image.
+/// </summary>
+public class IcoBmpFrameMetadata : IcoFrameMetadata, IDeepCloneable<IcoBmpFrameMetadata>, IBmpMetadata
```
JimBobSquarePants, line 12: You don't need this type.

src/ImageSharp/Formats/Ico/IcoImageFormatDetector.cs
```diff
@@ -0,0 +1,31 @@
...
+/// </summary>
+public sealed class IcoImageFormatDetector : IImageFormatDetector
+{
+    /// <inheritdoc/>
+    public int HeaderSize { get; } = 4;
+
+    /// <inheritdoc/>
+    public unsafe bool TryDetectFormat(ReadOnlySpan<byte> header, [NotNullWhen(true)] out IImageFormat? format)
+    {
+        fixed (byte* ptr = &header.GetPinnableReference())
```
JimBobSquarePants, line 19: This seems dangerous. You're working too hard to maintain DRY when you should be treating each image format as an entirely separate type.

src/ImageSharp/Formats/Ico/IcoPngFrameMetadata.cs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /// <inheritdoc/>
+    public override IcoLikeFrameCompression Compression { get; } = IcoLikeFrameCompression.Png;
+
+    /// <summary>
+    /// Gets or sets the number of bits per sample or per palette index (not per pixel).
+    /// Not all values are allowed for all <see cref="ColorType"/> values.
+    /// </summary>
+    public PngBitDepth? BitDepth { get; set; }
```
JimBobSquarePants, line 47: This should just have the compression type and the encoding dimensions

JimBobSquarePants (changes requested):
You've gone the opposite direction to how I hoped I'm afraid here. Please let me know if you need further guidance and I'll try my best to explain.

frg2089:
> You've gone the opposite direction to how I hoped I'm afraid here. Please let me know if you need further guidance and I'll try my best to explain.

I refactored the code.
Now it doesn't have the IcoPngFrameMetadata, IcoBmpFrameMetadata and the corresponding Cur variants.
We can use GetFormatMetadata(XXXFormat.Instance) (which isn't there yet) to get specific metadata.
If I'm not doing this the right way, I'm probably not understanding what you're really thinking.

JimBobSquarePants:
I'm going to write some code to your fork

src/ImageSharp/Formats/Icon/IconDecoderCore.cs
```diff
@@ -0,0 +1,207 @@
...
+                    pngMetadata = x.Image.Metadata.GetPngMetadata();
+                }
+
+                // Bmp does not contain frame specific metadata.
+                target.Metadata.SetFormatMetadata(PngFormat.Instance, target.Metadata.GetPngFrameMetadata());
+            }
+            else if (x.Index == 0)
+            {
+                bmpMetadata = x.Image.Metadata.GetBmpMetadata();
+            }
```
frg2089, line 94: I think the BitsPerPixel property is important and will be used by the encoder.
Such as the flutter.ico
It has 4-bit 8-bit and 32-bit images of the same size.
We must tell the encoder to encode using the same BitsPerPixel.

src/ImageSharp/Formats/Icon/IconDecoderCore.cs
```diff
@@ -0,0 +1,207 @@
...
+        // Copy the format specific metadata to the image.
+        if (bmpMetadata != null)
+        {
+            result.Metadata.SetFormatMetadata(BmpFormat.Instance, bmpMetadata);
+        }
+
+        if (pngMetadata != null)
+        {
+            result.Metadata.SetFormatMetadata(PngFormat.Instance, pngMetadata);
+        }
```
frg2089, line 114: I think the PNG data may not be the only one. ICO file may have 2 and more PNG data.
So, we should store the PngMetadata in the FrameMetadata instead of the ImageMetadata.

frg2089 (commented):
I have some questions about how to store PngMetadata and BmpMetadata.

We should store BitsPerPixel property for encoder.
The ICO file may have more than one PNG data.

JimBobSquarePants:
> I have some questions about how to store PngMetadata and BmpMetadata.
> 
> We should store BitsPerPixel property for encoder. The ICO file may have more than one PNG data.

I’d store that in the Ico/Cur frame metadata. I’d make it an enum though (it’s a fixed range based on available bmp/png fixes)

src/ImageSharp/Formats/Icon/IconDecoderCore.cs
JimBobSquarePants: They’re just for easy conversation really. See below re bits per pixel. You may want to capture palette data also

src/ImageSharp/Formats/Icon/IconDecoderCore.cs
JimBobSquarePants: No we’d tell the encoder to encode each frame independently. Your ico/cut frame metadata should contain the bits per pixel property

JimBobSquarePants:
Apologies for the radio silence over the last few weeks. I've been very busy fixing up the animation behavior for v3.1

frg2089:
Do we have some new ways to save frames of different sizes yet?

JimBobSquarePants:
> Do we have some new ways to save frames of different sizes yet?

You encode the size stored in the metadata. Pixel buffers can be constrained to regions

src/ImageSharp/Formats/Icon/IconDecoderCore.cs
```diff
@@ -0,0 +1,226 @@
...
+{
+    private IconDir fileHeader;
+
+    public DecoderOptions Options { get; } = options;
+
+    public Size Dimensions { get; private set; }
+
+    protected IconDir FileHeader { get => this.fileHeader; private set => this.fileHeader = value; }
+
+    protected IconDirEntry[] Entries { get; private set; } = [];
```
frg2089, line 22: SA1010
Should I avoid this syntax?

JimBobSquarePants: You should be fine now that I've updated the StyleCop version.

JimBobSquarePants:
@frg2089 Thanks for all your effort so far. It's been tricky for me to review as I've just moved house and have no real internet just now. I'll be able to do a proper review next week though.

JimBobSquarePants:
I'm going to have a look at how we manage the sizes in the encoding options. I want to come up with the best possible design so I will look for inspiration in other libraries.

JimBobSquarePants:
@Poker-sang @frg2089 Thanks for your patience.

I think this is in a really good place now. I can do any additional changes (potentially palette handling) once merged. 

The only thing I'd like to add is some more tests to ensure we can encode and decode the files. I scoured the internet and managed to find some test files here. 

https://github.com/SystemRage/Iconolatry/tree/master/test_decode

I cannot add the image to your fork due to LFS permissions issues with GitHub but would be very grateful if you were able to do so.

frg2089:
> The only thing I'd like to add is some more tests to ensure we can encode and decode the files. I scoured the internet and managed to find some test files here.

Done!

JimBobSquarePants (approved):
Let's get this merged! Thanks very much for your contribution!! 🤩

PR #2588, opened 2023-11-24
Add smart animated format conversion.

JimBobSquarePants:
### Prerequisites

### Description

**Note: Most of the file changes here are new references images that were generated to cater for changes in the color distance calculator that improved lookup accuracy.** 

This PR adds some internal tooling that allows us to convert between animated image formats without the user having to specify the mechanics for each frame to ensure converting between formats returns an equivalent result.

For v4 I plan to update the metadata formats to introduce public bridging types and methods to create smart defaults when switching formats but for now I had to use an internal, non-extensible approach. 

For v3.1 we will support bi-directional conversion between gif, png, and webp formats. Full color table mapping is performed between formats when present as well as duration, and disposal, blend modes where applicable.

Here's an example of an automatic  conversion.

Gif (45.2KB)

Encode_AnimatedFormatTransform_FromGif_Rgba32_giphy

Png (33.3KB)

Encode_AnimatedFormatTransform_FromGif_Rgba32_giphy

In addition to the conversion functionality all 3 animated formats have been enhanced with shared methodology that allows the optimization of the frames to only include changed data. This prevents the potential explosion in file size (multi MBs) that can be caused when re-encoding the frames that are coalesced during decode.

This example from gifiddle demonstrates the active encoded area of a frame following that optimization.

image

For many gifs this actually leads to an improvement over the incoming file size. For example, the original of the gif above is 52.3KB


JimBobSquarePants:
@antonfirsov @tocsoft I need your help if you have time? 

I'm seeing allocation cleanup failure reports on Linux/Mac and I can recreate them on my old MacBook. Thing is... I've touched nothing related to memory management nor does this test touch code I've changed with my PR. 

Is there any way to easily diagnose the cause? I tried using dotmemory Unit in Rider but it won't give the option to run the test in the manner that allows profiling memory.

Running this test in isolation is enough to cause failure.
[[xUnit.net 00:00:27.79]     TiffDecoder_CanDecode_JpegCompressed<Rgba32>(provider: Tiff/twain-rgb-jpeg-with-bogus-ycbcr-subsampling.tiff[Rgba32]) [FAIL]](https://github.com/SixLabors/ImageSharp/actions/runs/6987623519/job/19014351749?pr=2588#step:15:54)

src/ImageSharp/Formats/AnimationUtilities.cs
```diff
@@ -0,0 +1,190 @@
...
+
+            bool hasRowDiff = false;
+            int i = 0;
+            uint x = 0;
+            int length = current.Length;
+            int remaining = current.Length;
+            if (Avx2.IsSupported && remaining >= 2)
+            {
+                while (remaining >= 2)
+                {
```
gfoidl, line 80: So the condition `remaining >= 2` is check twice. Thus either make a `do-while` loop or keep the `while` but as
```suggestion
                while (Avx2.IsSupported && remaining >= 2)
                {
```
. The JIT will constant eleminate `Avx2.IsSupported`.
Personally I'd go with the `do-while` as it reads more natural then.

src/ImageSharp/Formats/AnimationUtilities.cs
```diff
@@ -0,0 +1,190 @@
...
+        // Webp requires even bounds
+        if (clampingMode == ClampingMode.Even)
+        {
+            bounds.Width = Math.Min(resultFrame.Width, bounds.Width + (bounds.X & 1));
+            bounds.Height = Math.Min(resultFrame.Height, bounds.Height + (bounds.Y & 1));
+            bounds.X = Math.Max(0, bounds.X - (bounds.X & 1));
+            bounds.Y = Math.Max(0, bounds.Y - (bounds.Y & 1));
+        }
+
+        return new(hasDiff, bounds);
```
gfoidl, line 166:
```suggestion
        return (hasDiff, bounds);
```
For the value tuples explicit `new` isn't needed.

(I rather find the `new` here confusing, because I'm think what type was it -- in contrast w/o the new it's like "ah, a value tuple")

src/ImageSharp/Formats/AnimationUtilities.cs
```diff
@@ -0,0 +1,190 @@
...
+        for (int y = 0; y < destBuffer.Height; y++)
+        {
+            Span<TPixel> sourceRow = sourceBuffer.DangerousGetRowSpan(y);
+            Span<TPixel> destRow = destBuffer.DangerousGetRowSpan(y);
+            sourceRow.CopyTo(destRow);
+        }
+    }
+}
+
+#pragma warning disable SA1201 // Elements should appear in the correct order
```
gfoidl, line 183: Is this needed?

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -200,24 +215,16 @@ public ColorDistanceCache(MemoryAllocator allocator)
         }
 
         [MethodImpl(InliningOptions.ShortMethod)]
-        public void Add(Rgba32 rgba, byte index)
+        public readonly void Add(Rgba32 rgba, byte index)
```
gfoidl, line 218: I don't think the `readonly` is correct here, as it mutates the field `tablePointer`.

Below in `TryGetValue` it's fine, no mutation happens.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -226,18 +233,20 @@ public bool TryGetValue(Rgba32 rgba, out short match)
         /// Clears the cache resetting each entry to empty.
         /// </summary>
         [MethodImpl(InliningOptions.ShortMethod)]
-        public void Clear() => this.table.GetSpan().Fill(-1);
+        public readonly void Clear() => this.table.GetSpan().Fill(-1);
```
gfoidl, line 236: I thinks it's not readonly.

stefannikolei:
> @antonfirsov @tocsoft I need your help if you have time?
> 
> I'm seeing allocation cleanup failure reports on Linux/Mac and I can recreate them on my old MacBook. Thing is... I've touched nothing related to memory management nor does this test touch code I've changed with my PR.
> 
> Is there any way to easily diagnose the cause? I tried using dotmemory Unit in Rider but it won't give the option to run the test in the manner that allows profiling memory.
> 
> Running this test in isolation is enough to cause failure. [[xUnit.net 00:00:27.79] TiffDecoder_CanDecode_JpegCompressed(provider: Tiff/twain-rgb-jpeg-with-bogus-ycbcr-subsampling.tiff[Rgba32]) [FAIL]](https://github.com/SixLabors/ImageSharp/actions/runs/6987623519/job/19014351749?pr=2588#step:15:54)
> 
> EDIT.... OK. It was failing locally but now it suddenly isn't. Could it be a memory pressure thing triggered by my additional tests?

I tried running main first on my Mac. But not even there all tests go green.

         Encode_AnimatedLossy<Rgba32>(provider: Webp/leo_animated_lossy.webp[Rgba32])

SixLabors.ImageSharp.ImageFormatException
Read unexpected webp chunk data
   at SixLabors.ImageSharp.Formats.Webp.WebpThrowHelper.ThrowImageFormatException(String errorMessage) in /Users/stefannikolei/projects/github/sixlabors/ImageSharp/src/ImageSharp/Formats/Webp/WebpThrowHelper.cs:line 14

       CanLimitHwIntrinsicFeaturesWithIntrinsicsParam
       CanLimitHwIntrinsicFeaturesWithSerializableAndIntrinsicsParams
       CanLimitHwIntrinsicFeaturesWithSerializableParam

Microsoft.DotNet.RemoteExecutor.RemoteExecutionException
Remote process failed with an unhandled exception.
Child exception:
  Xunit.Sdk.FalseException: Assert.False() Failure
Expected: False
Actual:   True

stefannikolei:
I tried debugging it. 

In my eyes (looking at the failing png test) The problem starts to manifest here:

image

In all other iterations the width was 32 and resulted in bytesPerFrameScanLine = 129. Here the width is 10

JimBobSquarePants:
> following tests are failing on my machine
> 
> ```
>      GifDecoderTests
>      Decode_VerifyRootFrameAndFrameCount<Rgba32>(provider: Gif/cheers.gif[Rgba32], expectedFrameCount: 93)
>      Decode_VerifyRootFrameAndFrameCount<Rgba32>(provider: Gif/issues/issue403_baddescriptorwidth.gif[Rgba32], expectedFrameCount: 36)
>      Issue1668_InvalidColorIndex<Rgba32>(provider: Gif/issues/issue1668_invalidcolorindex.gif[Rgba32])
>    PngEncoderTests
>      Encode_APng<Rgba32>(provider: Png/animated/apng.png[Rgba32])     
>     TiffDecoderTests
>      DecodeMultiframe<Rgba32>(provider: Tiff/multipage_deflate_withPreview.tiff[Rgba32])
>      DecodeMultiframe<Rgba32>(provider: Tiff/multipage_lzw.tiff[Rgba32])
>      TiffDecoder_CanDecode_JpegCompressed<Rgba32>(provider: Tiff/twain-rgb-jpeg-with-bogus-ycbcr-subsampling.tiff[Rgba32])
> ```
> 
> I tried debugging it.
> 
> In my eyes (looking at the failing png test) The problem starts to manifest here:
> 
> image
> In all other iterations the width was 32 and resulted in bytesPerFrameScanLine = 129. Here the width is 10
> 
> A naive fix (the unit test passes after this:
> 
> image

Well spotted! Yes, that's definitely a bug. The interlaced version of the method was doing the correct thing.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
JimBobSquarePants: Interesting... I only changed these because the IDE suggested it. 

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0251

JimBobSquarePants: Actually, just asked in the Discord server about this and got this reply which was how I thought it worked.

"readonly only concerns itself with whether it modifies this, not if it is transitively readonly"

gfoidl: Aya, I've forgotten that Roslyn only sees "this" and doesn't look further.
Thanks for the update!

src/ImageSharp/Formats/AnimationUtilities.cs
JimBobSquarePants: Yeah, I wanted to keep the enum in the same file as the class, but Stylecop doesn't like it.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
```diff
@@ -184,14 +184,14 @@ private static float DistanceSquared(Rgba32 a, Rgba32 b)
...
-    /// Entry count is currently limited to 4601025 entries (8MB).
+    /// Entry count is currently limited to 2335905 entries (4MB).
     /// </para>
     /// </remarks>
     private unsafe struct ColorDistanceCache : IDisposable
     {
         private const int IndexRBits = 5;
         private const int IndexGBits = 5;
-        private const int IndexBBits = 6;
+        private const int IndexBBits = 5;
```
gfoidl, line 194: Is there a test that unveils that "bug" (was this a bug?)?

gfoidl (approved):
Code-wise LGTM.

To the implementation itself I can't say too much as I'm missing knowledge here.

src/ImageSharp/Processing/Processors/Quantization/EuclideanPixelMap{TPixel}.cs
JimBobSquarePants: I was something I just caught doing a visual test which is hard to test for. 

The cache was only filled to a maximum value of ~96% which led to matching where it shouldn't. This led to transparent areas in the frame following deduplication. Originally, I was going to simply add more precision for a single component (blue) which fixed the issue, but I discovered that with the new index determinaiton code the cache fills to a maximum of ~98% and the problem simply went away.

PR #2601, opened 2023-12-03
Promote PixelTypeInfo to Pixel

stefannikolei:
Fixes #2534


src/ImageSharp/PixelFormats/PixelOperations{TPixel}.cs
```diff
@@ -32,7 +31,7 @@ public partial class PixelOperations<TPixel>
     /// Gets the pixel type info for the given <typeparamref name="TPixel"/>.
     /// </summary>
     /// <returns>The <see cref="PixelTypeInfo"/>.</returns>
-    public virtual PixelTypeInfo GetPixelTypeInfo() => LazyInfo.Value;
+    public static PixelTypeInfo GetPixelTypeInfo() => TPixel.GetPixelTypeInfo();
```
JimBobSquarePants, line 34: I'm assuming this is the reason you added the exclusion to the ruleset? 

Is there an advantage to making this static? If so, I'd rather the exclusion was in the source so we are forced to consider introduction of static fields elsewhere.

src/ImageSharp/Formats/PixelTypeInfo.cs
```diff
@@ -12,42 +12,37 @@ namespace SixLabors.ImageSharp.Formats;
...
     /// </summary>
-    public PixelAlphaRepresentation? AlphaRepresentation { get; }
-
-    internal static PixelTypeInfo Create<TPixel>()
-        where TPixel : unmanaged, IPixel<TPixel>
-        => new(Unsafe.SizeOf<TPixel>() * 8);
+    public PixelAlphaRepresentation? AlphaRepresentation { get; init; }
 
-    internal static PixelTypeInfo Create<TPixel>(PixelAlphaRepresentation alpha)
+    internal static PixelTypeInfo Create<TPixel>(byte componentCount, PixelAlphaRepresentation pixelAlphaRepresentation)
```
JimBobSquarePants, line 40: I'm wondering whether we can remove this? I want to add the additional enum https://github.com/SixLabors/ImageSharp/discussions/2534#discussioncomment-7171432 which means adding an additional property to the constructor.

src/ImageSharp/PixelFormats/IPixel.cs
```diff
@@ -14,6 +15,12 @@ namespace SixLabors.ImageSharp.PixelFormats;
 public interface IPixel<TSelf> : IPixel, IEquatable<TSelf>
     where TSelf : unmanaged, IPixel<TSelf>
 {
+    /// <summary>
+    /// Gets the The pixel type information.
```
JimBobSquarePants, line 19: `the The`

src/ImageSharp/PixelFormats/IPixel.cs
```diff
@@ -14,6 +15,12 @@ namespace SixLabors.ImageSharp.PixelFormats;
...
+    /// </summary>
+    /// <returns>PixelTypeInfo</returns>
```
JimBobSquarePants, line 21: Use `<see/>` for return types.

src/ImageSharp/PixelFormats/PixelImplementations/A8.cs
```diff
@@ -55,6 +56,12 @@ public partial struct A8 : IPixel<A8>, IPackedVector<byte>
     [MethodImpl(InliningOptions.ShortMethod)]
     public static bool operator !=(A8 left, A8 right) => !left.Equals(right);
 
+    /// <summary>
+    /// Gets the The pixel type information.
+    /// </summary>
+    /// <returns>PixelTypeInfo</returns>
```
JimBobSquarePants, line 62: You should use `/// <inheritdoc/>` for each implementation.

src/ImageSharp/Formats/PixelTypeInfo.cs
JimBobSquarePants: Maybe just add the additional parameter as having to duplicate `Unsafe.SizeOf<TPixel>() * 8` is worse.

src/ImageSharp/PixelFormats/PixelImplementations/A8.cs
stefannikolei: I first did that. But SA1648 is giving a false positive here. Should I disable it globally or in each file?

src/ImageSharp/Formats/PixelTypeInfo.cs
JimBobSquarePants: @stefannikolei I'm going to add this for you today.

stefannikolei: Ah. To be honest.. totally forgot about that comment

JimBobSquarePants:
@stefannikolei I've made my changes, will need to do a sense check in the morning before merging as it's late now.

src/ImageSharp/Formats/PixelTypeInfo.cs
```diff
@@ -4,50 +4,51 @@
...
+    /// Gets the count of the color components
     /// </summary>
-    /// <param name="bitsPerPixel">Color depth, in number of bits per pixel.</param>
-    /// <param name="alpha">The pixel alpha transparency behavior.</param>
-    public PixelTypeInfo(int bitsPerPixel, PixelAlphaRepresentation alpha)
-    {
-        this.BitsPerPixel = bitsPerPixel;
-        this.AlphaRepresentation = alpha;
-    }
+    public byte ComponentCount { get; init; }
```
antonfirsov, line 29: We should consider changing this to `int`.
```suggestion
    public int ComponentCount { get; init; }
```

Unless there is a reason to compress data for optimization (~the type is expected to be instantiated very frequently), it is generally recommended to prefer `int` over other integral types in .NET APIs. For example, `IPEndPoint.Port` is an `int`, while technically it could be an `ushort`. The reason is that this removes the need for conversion which is both expensive and complicated.

On the other hand, I noticed that we use non-`int` integrals in other descriptor types, eg. `PngMetadata`. I wish I noticed this earlier.

src/ImageSharp/Formats/PixelTypeInfo.cs
```diff
@@ -4,50 +4,51 @@
...
 
     /// <summary>
-    /// Gets color depth, in number of bits per pixel.
+    /// Gets the pixel component precision.
     /// </summary>
-    public int BitsPerPixel { get; }
+    public PixelComponentPrecision? ComponentPrecision { get; init; }
```
antonfirsov, line 34: AFAIK there might be pixel types with assymmetric component precision. (I remember briefly touching this topic in an old discussion with Clinton.)

How would this model handle such pixel types?

src/ImageSharp/Color/Color.cs
```diff
@@ -124,7 +125,9 @@ public static Color FromPixel<TPixel>(TPixel pixel)
         {
             return new((L16)(object)pixel);
         }
-        else if (Unsafe.SizeOf<TPixel>() <= Unsafe.SizeOf<Rgba32>())
+
```
antonfirsov, line 128: While not critical, I would prefer an abstraction that can get rid of the special casing above this line - by moving it to `IPixel` / `PixelOperations` code.

src/ImageSharp/Formats/PixelTypeInfo.cs
JimBobSquarePants: I’ve often wondered if there was a rule. For v4 we can definitely fix this.

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Me also. In this case I think we should ditch the specialisation and box anything greater than byte. `ColorFromPixel` is not area for high performance operations.

On `IPixel` I would, if we can, drop all the various `FromXX` specialisation there also and ensure a good optimised bulk operations for all `Rgba32` compatible pixel formats. That may not be possible though.

src/ImageSharp/Formats/PixelTypeInfo.cs
antonfirsov: I looked at the usages of the property, and the only way we use it is the `info.ComponentPrecision <= PixelComponentPrecision.Byte` comparison, meaning that the only thing we need this for is to determine if the precision is below or above a certain treshold, and the exact precision information is practically unused. Unless we think there are use cases (proven by actual user user need) to observe *the precision of the highest precision component*, IMO we should try to avoid exposing such a property and explore alternative options to solve this.

src/ImageSharp/Formats/PixelTypeInfo.cs
JimBobSquarePants: All good. I've fixed the property for this PR, we can look at others.

src/ImageSharp/Formats/PixelTypeInfo.cs
JimBobSquarePants: My thought process is that users can implement custom performance optimized processors based upon this implementation. If they know the precision, they could operate over `byte` vs `Vector4`. 

The use case that I've currently implemented fixes a bug we had also.

`Rgba1010102` uses a `uint` as its packed value which is the same size as `Rgba32` but we lose precision for the RGB components if we convert to it as we were since they require 10 bit precision.

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Well remembered, but those are really short loops (max 256) in real terms. I'd hate to limit our flexibility and extensibility based on introducing a requirement there. 

Ideally, I want to introduce as much agnostity as possible in our APIs.

src/ImageSharp/Formats/PixelTypeInfo.cs
antonfirsov: I see that `ComponentCount` is a similar informative property. I didn't notice this fact in the #2534 discussion. The way I view this now is that we are attempting to add a *Rich Pixel Type Descriptor* feature. IMO there are a couple of important points we should consider before fully committing to it:
- To really get this right, it might be better to expose a `struct/class ComponentInfo` and a property in `PixelTypeInfo` that enumerates the ComponentInfos, so each component can tell its precision separately instead of telling the maximum (for which I'm not sure if there is a use-case except the conversion implementation this PR is proposing). As far as I can recall that old discussion with Clinton, WIC has such a model but I may be wrong.
- Note that this feature idea doesn't necessarily overlap with the `Pixel <=> Color` conversion. I don't see how can we implement efficient bulk conversion using these descriptor information only. In fact, if our goal is to solve the `Pixel <=> Color` conversion thing, it might be better to promote the conversion (& bulk conversion) methods to `PixelOperations<T>`.
- Considering that this isn't super trivial to get right, I'm really curious if we have enough evidence (=enough users) proving that the  *Rich Pixel Type Descriptor* feature is worth the headache.

src/ImageSharp/Color/Color.cs
antonfirsov: Boxing is very ugly, even if done only `256x` typically, I would try to avoid it if there are alternatives.

src/ImageSharp/Formats/PixelTypeInfo.cs
antonfirsov: (Your comment somehow didn't appear before me posting the previous one.)

> My thought process is that users can implement custom performance optimized processors based upon this implementation. 

Do you have any specific examples?

JimBobSquarePants: >Considering that this isn't super trivial to get right, I'm really curious if we have enough evidence (=enough users) proving that the Rich Pixel Type Descriptor feature is worth the headache.

We get zero feedback on this stuff. I'm just trying to create something that gives us as much knowledge as possible. 

For example, We, could be using both the `PixelComponentPrecision` and `PixelAlphaRepresentation` properties in encoders to determine color types and bit depth no explicit encoding options are passed. This helps solve handling of introduced transparency in a processing operation.

JimBobSquarePants: Here's the info from WIC

https://learn.microsoft.com/en-us/windows/win32/wic/-wic-codec-native-pixel-formats

And here's how it's dealt with in MagicScaler

https://github.com/saucecontrol/PhotoSauce/blob/v0.14.0/src/MagicScaler/Core/PixelFormats.cs

We're missing a color type enum which would give us a fairly robust model that can allow encoders to make sensible decisions.

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Maybe we just back Color with `Vector4`? it already takes up the same amount of space and means we would be able to use bulk operations.

src/ImageSharp/Formats/PixelTypeInfo.cs
JimBobSquarePants: Here's an idea for how we can handle the pixel color type. We can combine them in any order to show component order but also test against them to determine encoding color type. 

```c#
[Flags]
public enum PixelColorType
{
  Red = 1,
  Green = 2,
  Blue = 4,
  Alpha = 8,
  Grayscale = 16
}
```

Defining them.

``` c#
var rgb = PixelColorType.Red | PixelColorType.Green | PixelColorType.Blue;
var  bgr = PixelColorType.Blue | PixelColorType.Green | PixelColorType.Red;
var la = PixelColorType.Grayscale | PixelColorType.Alpha;
```

Testing them.

```c#
info.ColorType.HasFlag(PixelColorType.Blue | PixelColorType.Green | PixelColorType.Red | PixelColorType.Alpha);
```

saucecontrol: > * To really get this right, it might be better to expose a `struct/class ComponentInfo` and a property in `PixelTypeInfo` that enumerates the ComponentInfos, so each component can tell its precision separately instead of telling the maximum (for which I'm not sure if there is a use-case except the conversion implementation this PR is proposing). As far as I can recall that old discussion with Clinton, WIC has such a model but I may be wrong.

You are quite correct.  I don't currently have this in my implementation because I haven't built my own converters for packed formats yet, but WIC does define per-channel masks so that you can build generic converters.  https://learn.microsoft.com/en-us/windows/win32/api/wincodec/nf-wincodec-iwicpixelformatinfo-getchannelmask

JimBobSquarePants: Whoa! How does that work?

saucecontrol: It's no different than what you've already got for decoding/converting BMP (https://github.com/SixLabors/ImageSharp/blob/main/src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs#L1129) since it also allows arbitrary channel depths and masks.

WIC allows you to define/register your own optimized converters for formats you register, but otherwise it falls back to a generic one that just works off the defined masks.

src/ImageSharp/Color/Color.cs
antonfirsov: The reason we introduced the ability to define (higher than `Vector4`) arbitrary precison pixel types was this feature request: https://github.com/SixLabors/ImageSharp.Drawing/discussions/165

I like the fact that ImageSharp supports such use-cases.

src/ImageSharp/Formats/PixelTypeInfo.cs
antonfirsov: > It's no different than what you've already got for decoding/converting BMP

Except that the channel count is not limited to 4!

> so that you can build generic converters

Our "almost generic (tm)" way of doing this is to push everything through `Vector4`. Defining a pixel type in ImageSharp means implementing conversion to/from Vector4 and virtually all of our processors depend on Vector4-convertibility. We should decide what to do with the cases that don't fit this model. Maybe there are some limited applications for them, even though such images are for example "not resizable".

@saucecontrol I wonder if there are any image formats that can encode `GUID_WICPixelFormat144bpp8ChannelsAlpha`? What are real-life applications of such wild stuff? :)

src/ImageSharp/Color/Color.cs
JimBobSquarePants: Then use `Vector4` as the default and  box anything bigger. People expected double precision are in for a rough time with the library since almost everything is converted to use single precision at some point.

src/ImageSharp/Formats/PixelTypeInfo.cs
saucecontrol: In general, it's a pretty safe assumption anything weird defined by WIC is to support the JXR/HD Photo codec https://learn.microsoft.com/en-us/windows/win32/wic/-wic-codec-native-pixel-formats#jpeg-xr-native-codec 😄 

I think Microsoft meant JXR to be all things to all people and implemented a lot of things that never actually got used.  In the case of the 6+ channel formats, I assume those were meant to replace some specialized uses of TIFF, since TIFF also allows an arbitrary number of channels.  The only real world use I can think of for those formats is print workflows since printers may have any number of inks.

WIC does support up to 5 channels in its TIFF codec, for CMYKA.  That's a real-world format, although even Photoshop will warn you that most software doesn't support it when you save in that format.  Here's a sample for reference:

[cmyka.zip](https://github.com/SixLabors/ImageSharp/files/13665156/cmyka.zip)

Windows photo viewer will open that one and display the transparency correctly, but its color management doesn't support CMYKA, so the colors will display incorrectly.  Using the Little-CMS plugin, I'm able to convert those correctly in MagicScaler, but other than Photoshop, I don't know offhand what else works.  Probably ImageMagick and GIMP?

JimBobSquarePants: How about this? It stores up to 8 components with a precision of 255 bits per component. 

Usage 

image


```c#
public readonly struct PixelPrecision
{
    private readonly int count;
    private readonly long precisionData;

    private PixelPrecision(int count, long precisionData)
    {
        this.count = count;
        this.precisionData = precisionData;
    }

    public int ComponentCount => this.count;

    public static PixelPrecision Create(int count, params int[] precision)
    {
        if (precision.Length != count || precision.Length > 8)
        {
            throw new ArgumentException($"Cannot have more than 8 components and must match {count}.");
        }

        long precisionData = 0;
        for (int i = 0; i < precision.Length; i++)
        {
            if (precision[i] is < 0 or > 255)
            {
                throw new ArgumentException("Precision must be between 0 and 255.");
            }

            precisionData |= ((long)precision[i]) << (8 * i);
        }

        return new PixelPrecision(count, precisionData);
    }

    public int GetComponentPrecision(int componentIndex)
    {
        if (componentIndex < 0 || componentIndex >= this.count)
        {
            throw new ArgumentOutOfRangeException($"index must be greater than or equal to 0 and less than {this.count}.");

        }

        return (int)((this.precisionData >> (8 * componentIndex)) & 0xFF);
    }

    public int GetMaximumComponentPrecision()
    {
        int maxPrecision = 0;
        for (int i = 0; i < this.count; i++)
        {
            int componentPrecision = (int)((this.precisionData >> (8 * i)) & 0xFF);
            if (componentPrecision > maxPrecision)
            {
                maxPrecision = componentPrecision;
            }
        }

        return maxPrecision;
    }
}
```

JimBobSquarePants: Damn. Just realized `GUID_WICPixelFormat144bpp8ChannelsAlpha` has 9 channels!!! I guess I could use 2 backing stores?

JimBobSquarePants: Here is as a class supporting up to 16 components. Each component precision maxes out at 255bits. 

I'm allowing such high precision because of https://github.com/SixLabors/ImageSharp.Drawing/discussions/165 
otherwise, I'd limit it to 16 bits and using a single backing store.

```c#
public sealed class PixelPrecision
{
    private readonly long precisionData1;
    private readonly long precisionData2;

    private PixelPrecision(int count, long precisionData1, long precisionData2)
    {
        this.ComponentCount = count;
        this.precisionData1 = precisionData1;
        this.precisionData2 = precisionData2;
    }

    public int ComponentCount { get; }

    public static PixelPrecision Create(int count, params int[] precision)
    {
        if (precision.Length != count || precision.Length > 16)
        {
            throw new ArgumentException($"Count must match the length of precision array and cannot exceed 16.");
        }

        long precisionData1 = 0;
        long precisionData2 = 0;
        for (int i = 0; i < precision.Length; i++)
        {
            if (precision[i] is < 0 or > 255)
            {
                throw new ArgumentException("Precision must be between 0 and 255.");
            }

            if (i < 8)
            {
                precisionData1 |= ((long)precision[i]) << (8 * i);
            }
            else
            {
                precisionData2 |= ((long)precision[i]) << (8 * (i - 8));
            }
        }

        return new PixelPrecision(count, precisionData1, precisionData2);
    }

    public int GetComponentPrecision(int componentIndex)
    {
        if (componentIndex < 0 || componentIndex >= this.ComponentCount)
        {
            throw new ArgumentOutOfRangeException($"Component index must be between 0 and {this.ComponentCount - 1} inclusive.");
        }

        long selectedPrecisionData = componentIndex < 8 ? this.precisionData1 : this.precisionData2;
        return (int)((selectedPrecisionData >> (8 * (componentIndex & 7))) & 0xFF);
    }

    public int GetMaximumComponentPrecision()
    {
        int maxPrecision = 0;
        for (int i = 0; i < this.ComponentCount; i++)
        {
            int componentPrecision = GetComponentPrecision(i);
            if (componentPrecision > maxPrecision)
            {
                maxPrecision = componentPrecision;
            }
        }

        return maxPrecision;
    }
}
```

saucecontrol: That's neat!  I think you might have to be able to account for padding bits, so I'm not sure you can get away with not having a mask per channel.  I'm trying to think of tricky examples... I've seen some BMPs that have gaps in the channel masks, but those were of the stress-test variety, not real world.  You'll have real-world formats like B10G10R10, which is 32-bit with 2 bits padding, though.

JimBobSquarePants: I'm not sure that's required for our use cases which is what precision can I safly store this pixel in. 

We can tell the actual size of the pixel format from the info since the types themselves must handle padding. `Unsafe.Sizeof<B10G10R10>()` should be the same size `Unsafe.Sizeof<Rgba1010102>()` which is `4`.

Theoretically we could add padding to the info. I'm assuming though that world is not completely mad and padding is always at the end?

saucecontrol: Right, I don't think you'll see a real-world case where `BitsPerPixel - Sum(ChannelPrecision)` doesn't give you the padding bits, which *should* always be at the end.  Just trying to come up with reasons explicit masks might be beneficial, and weirdo pixel formats with interior padding are at least a possibility.

The other thing I guess that's useful in an explicit channel mask is that it also communicates the channel order.  I'm not sure if WIC actually uses that information in that way, but I suppose it might.  They don't expose channel order in any other way in their format definitions.  It's just in the name and in the mask.

JimBobSquarePants: I think I'll go with real world for now. If someone wants something really weird, they can find a really weird library.

JimBobSquarePants: One thing that does concern me though is the the performance of the parent type. We removed all the lazy loading as the type was to be a small and fast struct but now, I'm concerned about overheads.

src/ImageSharp/Color/Color.cs
JimBobSquarePants: I've updated `Color` to use `Vector4` to avoid boxing. I've also removed the implicit casting for random pixel formats in favour of explicit methods.

src/ImageSharp/Formats/PixelTypeInfo.cs
JimBobSquarePants: Type has been introduced as `PixelComponentInfo`

JimBobSquarePants (approved):
I'm happy with this now. Would love a second or third opinion though.

saucecontrol (approved):
Really nice improvement!

PR #2609, opened 2023-12-08
Correctly calculate foreground Rectangle when negative background location is set (3.1 fix)

### LINKED ISSUE #2608: Bug with negative background location while drawing an image
We incorrectly calculate the updated area of interest of the foreground image when applying a negative offset to the background drawing location.

> I checked the source code and I think I see the issue around these lines: https://github.com/SixLabors/ImageSharp/blob/3d69f21ec2fe6a0e274316bafa3d1e1fab06dc83/src/ImageSharp/Processing/Processors/Drawing/DrawImageProcessor%7BTPixelBg%2CTPixelFg%7D.cs#L89
>
> The `foregroundRectangle.Width` is adjusted by the background location X value if it was negative. This is resulting in cropping off the _right_ side of the top image, not the left. Same for the Y direction just below that.
> 
> I created a test project to demonstrate the difference to make sure I wasn't going crazy, but here is an illustration:
> 
> imagesharp-3 0 2 Above: 3.0.2 behavior. The top/left of the top image is lost (expected due to negative coordinates).
>
> imagesharp-3 1-intended
> Above: what I believe the code in 3.1 is intended to do. The output would remain the same. I believe adjusting the rectangle's X an Y (in addition to removing it from the Width and Height) could have this effect. I am not an expert at what's happening below there, though, so I could be wrong.
> 
> imagesharp-3 1-actual
>Above: what is actually happening. The top image is indeed painted in the same box, but it originates from the top left corner and cuts off the bottom/right of the image instead of the top/left.
>
> Thanks again for your time!
>
>_Originally posted by @aceoft in https://github.com/SixLabors/ImageSharp/issues/2603#issuecomment-1845710788_

tocsoft:
### Prerequisites

### Description

This fixes #2608 which incorrectly handles the a negative drawing target/background location when drawing an image, ensuing the correct portion of the foreground image is retained and never cropped incorrectly.

The foreground rectangle is supposed to be in the foreground images coordinate space where are the background location is in the background images coordinate space, so we need to offset the x, y by the amount we shrink it by when unifying them to determine the portion of the foreground we are interested in.

This still does not change the rules for the foregrounds rectangle as that should not  support a negative x,y but only effects the background/target location we will draw the final cropped foreground image to.

### DIFF
--- a/src/ImageSharp/Processing/Processors/Drawing/DrawImageProcessor{TPixelBg,TPixelFg}.cs
+++ b/src/ImageSharp/Processing/Processors/Drawing/DrawImageProcessor{TPixelBg,TPixelFg}.cs
@@ -87,12 +87,14 @@ internal class DrawImageProcessor<TPixelBg, TPixelFg> : ImageProcessor<TPixelBg>
         if (this.BackgroundLocation.X < 0)
         {
             foregroundRectangle.Width += this.BackgroundLocation.X;
+            foregroundRectangle.X -= this.BackgroundLocation.X;
             left = 0;
         }
 
         if (this.BackgroundLocation.Y < 0)
         {
             foregroundRectangle.Height += this.BackgroundLocation.Y;
+            foregroundRectangle.Y -= this.BackgroundLocation.Y;
             top = 0;
         }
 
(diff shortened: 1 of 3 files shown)


JimBobSquarePants (approved):
LGTM. Thanks!!

PR #2610, opened 2023-12-08
Handle case when Foreground image overhangs bottom of background image during DrawImage Call (3.1 target)

### LINKED ISSUE #2603: Calling DrawImage with top image that overhangs bottom image in Y direction fails with exception
### Prerequisites

### ImageSharp version

3.1

SixLabors.ImageSharp.Drawing 2.0.1

Win10 22H2

### .NET Framework version

7

### Description

After upgrading the ImageSharp nuget package today, several of my unit tests started failing. They are all related to overlaying images that hang over the bottom edge of the bottom image.

The exception being thrown is:

`System.AggregateException: One or more errors occurred. (Specified argument was out of the range of valid values. (Parameter 'DangerousGetRowSpan(100). Y was out of range. Height=100')) ---> System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'DangerousGetRowSpan(100). Y was out of range. Height=100')`

It only seems to happen when a `Y` value is larger than the height of the bottom image.

I'm just getting started with ImageSharp and have been impressed by the amount of thought and hard work that has been put into the library. Thank you, and hopefully I provided everything we need here.

### Steps to Reproduce

Test method to reproduce the exception:

```csharp
public void TestOverlayImageOffEdge()
{
	using var bottom = new Image<Rgb24>(100, 100, Color.Black);
	using var top = new Image<Rgb24>(100, 100, Color.White);

	// OK (expected)
	bottom.Mutate(i => i.DrawImage(top, new Point(0, 0), 1f));

	// OK (expected, overhangs X direction by 1 pixel)
	bottom.Mutate(i => i.DrawImage(top, new Point(1, 0), 1f));

	// throws System.AggregateException
	bottom.Mutate(i => i.DrawImage(top, new Point(0, 1), 1f));
}
```

I had a quick look at the source code and wondered if providing a source rectangle would help -- it did. The below line appears to be a workaround.

```csharp
// OK
bottom.Mutate(i => i.DrawImage(top, new Point(0, 1), new Rectangle(0, 0, 100, 99), 1f));
```

### Images

_No response_

tocsoft:
### Prerequisites

### Description
Prevents the `ArgumentOutOfRangeException` when the area of interest of the foreground image will be drawn past the edge of the bounds of the background image.

fixes #2603 
related to #2609

Targeting 3.1 release branch

### DIFF
--- a/src/ImageSharp/Processing/Processors/Drawing/DrawImageProcessor{TPixelBg,TPixelFg}.cs
+++ b/src/ImageSharp/Processing/Processors/Drawing/DrawImageProcessor{TPixelBg,TPixelFg}.cs
@@ -96,14 +96,14 @@ internal class DrawImageProcessor<TPixelBg, TPixelFg> : ImageProcessor<TPixelBg>
             top = 0;
         }
 
-        if (left + foregroundRectangle.Width > source.Width)
+        if (left > source.Width - foregroundRectangle.Width)
         {
             // will overhange, lets trim it down
             int diff = (left + foregroundRectangle.Width) - source.Width;
             foregroundRectangle.Width -= diff;
         }
 
-        if (top + foregroundRectangle.Height > source.Height)
+        if (top > source.Height - foregroundRectangle.Height)
         {
             // will overhange, lets trim it down
             int diff = (top + foregroundRectangle.Height) - source.Height;
--- a/src/ImageSharp/Processing/Processors/Drawing/DrawImageProcessor{TPixelBg,TPixelFg}.cs
+++ b/src/ImageSharp/Processing/Processors/Drawing/DrawImageProcessor{TPixelBg,TPixelFg}.cs
@@ -96,6 +96,20 @@ internal class DrawImageProcessor<TPixelBg, TPixelFg> : ImageProcessor<TPixelBg>
             top = 0;
         }
 
+        if (left + foregroundRectangle.Width > source.Width)
+        {
+            // will overhange, lets trim it down
+            int diff = (left + foregroundRectangle.Width) - source.Width;
+            foregroundRectangle.Width -= diff;
+        }
+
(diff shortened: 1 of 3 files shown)


JimBobSquarePants:
Thanks for fixing this. 

We'll need to PR this against release/v3.1.x here also as well as update the ImageBrush to allow negative offsetting since it does the same thing.

tocsoft:
👍 i'll rebaseline the PRs against the 3.1 branch this weekend.

Regarding `ImageBrush` I was just looking over the code and I don't see how that would effected by these issues at all, this one is about the final blending stage and that's handled differently over there, and the other doesn't makes sense for textures either... I'm fairly sure that these changes should only effect the explicit `DrawImage` api.

JimBobSquarePants:
Thanks, appreciate it. See here for what I mean with the ImageBrush. It does the same reset of the position when given negative location. That’s why I exposed the rectangle constructor. 

https://github.com/SixLabors/ImageSharp.Drawing/discussions/305

src/ImageSharp/Processing/Processors/Drawing/DrawImageProcessor{TPixelBg,TPixelFg}.cs
```diff
@@ -96,6 +96,20 @@ protected override void OnFrameApply(ImageFrame<TPixelBg> source)
             top = 0;
         }
 
+        if (left + foregroundRectangle.Width > source.Width)
```
gfoidl, line 99:
```suggestion
        if (left > source.Width - foregroundRectangle.Width)
```
To prevent possible overflow.

tocsoft:
managed to reduce it to a single pair of clamping operations to solve the ArgumentOutOfRange exception.

JimBobSquarePants (approved):
Nice work here!

PR #2623, opened 2023-12-14
Adjusted the casing of the Webp format name

jscarle:
I adjusted the casing the Webp format name. It is the only format that has a mixed case name. All other formats are in upper case. This helps eliminate the need for case insensitive matching of format names (which is not always possible).


JimBobSquarePants:
Well spotted. Thanks!

PR #2633, opened 2023-12-29
Add support for HEIF based images

ynse01:
### Prerequisites

### Description
This PR is an attempt to start with #1320 

Implemented decoding and encoding of images based on a [HEIF](https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format) (ISO/IEC 23008-12:2022) container. These include (amongst others): [HEIC](https://www.adobe.com/creativecloud/file-types/image/raster/heic-file.html), [HIF](https://www.cameralabs.com/jpeg-vs-heif-canon-hif-comparison/) and [AVIF](https://en.wikipedia.org/wiki/AVIF).

Please note this PR does **NOT** add any new compression algorithm to ImageSharp. For now it will take a JPEG thumbnail as its pixel source. 
Needless to say, much more work is needed after this PR to reach the goal. I'm try to help there also.

Please do let me know any comments, also on whether to support the HEIF file format in the first place.


JimBobSquarePants:
Oh wow! Thanks! 🙏 I’ll have a good dig through asap!!

dlemstra:
I noticed that you used the x265 encoder. This is licensed under the GPL and cannot be used in this project.

ynse01:
> I noticed that you used the x265 encoder. This is licensed under the GPL and cannot be used in this project.

True, and that is why I used the Nokia repository as reference for my implementation, which has a license which I guess is compatible with ImageSharp. I did not start on writing an HVC Codec, due to patent concerns.

Nokia also includes a Patent waiver in its license. AVIF is supposed to be patent free, even its Codec. For AVIF I used SVT-AV1 as a reference, which has a 3-clause BSD license.

With all this, I hope the reference implementations that are mentioned now, can be used to create an AVIF Codec implementation in ImageSharp.

src/ImageSharp/Formats/Heif/Heif4CharCode.cs
```diff
@@ -0,0 +1,261 @@
...
+/// <summary>
+/// Supported 4 character codes for use in HEIF images.
+/// </summary>
+[GeneratedCode("T4", "")]
+public enum Heif4CharCode : uint
+{
+    /// <summary>
+    /// File Type.
+    /// </summary>
+    ftyp = 0x66747970U,
```
JimBobSquarePants, line 19: These will all need to use the correct casing.

src/ImageSharp/Formats/Heif/Heif4CharCode.cs
```diff
@@ -0,0 +1,261 @@
...
+// <auto-generated />
+
+using System.CodeDom.Compiler;
+
+namespace SixLabors.ImageSharp.Formats.Heif;
+
+/// <summary>
+/// Supported 4 character codes for use in HEIF images.
+/// </summary>
+[GeneratedCode("T4", "")]
```
JimBobSquarePants, line 13: I think `T4` here should be `TextTemplatingFileGenerator`. We don't use this elsewhere though (should we be? docs aren't great)

src/ImageSharp/Formats/Heif/HeifEncoderCore.cs
```diff
@@ -0,0 +1,334 @@
...
+        stream.Write(buf[..bytesWritten]);
+
+        stream.Write(data);
+    }
+
+    private static async Task<byte[]> CompressPixels<TPixel>(Image<TPixel> image, CancellationToken cancellationToken)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        using MemoryStream stream = new();
+        JpegEncoder encoder = new()
```
JimBobSquarePants, line 327: Is there any means to configure quality?

src/ImageSharp/Formats/Heif/HeifItemLink.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Six Labors.
+// Licensed under the Six Labors Split License.
+
+namespace SixLabors.ImageSharp.Formats.Heif;
+
+/// <summary>
+/// Link between <see cref="HeifItem"/> instances within the same HEIF file.
+/// </summary>
+public class HeifItemLink(Heif4CharCode type, uint sourceId)
```
JimBobSquarePants, line 9: I don't think these and others need to be public.

JimBobSquarePants (commented):
I need some guidance with the licensing implications here. 

@dlemstra ImageMagick uses libheif but has a custom license. How does that work?

src/ImageSharp/Formats/Heif/Heif4CharCode.cs
ynse01: OK, went for automaticaly generated PascalCasing for now, if this needs to be nicer, can be done with third element.

src/ImageSharp/Formats/Heif/Heif4CharCode.cs
ynse01: Corrected

src/ImageSharp/Formats/Heif/HeifItemLink.cs
ynse01: Indeed, made 3 classes internal

src/ImageSharp/Formats/Heif/HeifEncoderCore.cs
ynse01: Legacy JPEG is meant as a thumbnail format in HEIF. The primary formats is still to be added (Av1 in our case).
The current situation, with JPEG as main image is a bit strange, it would be more efficient to store the JPEG directly in its native format.

Av1 will have its own set of quality settings.

If you feel we need to set the quality of the thumbnail in the end situation, I'm happy to add that feature now already.

ynse01:
Not sure why one of the `RowIntervalTests` is failing, didn't touched that part of the code...

src/ImageSharp/Formats/Heif/HeifDecoderCore.cs
```diff
@@ -0,0 +1,684 @@
...
+    }
+
+    private static void SkipBox(BufferedReadStream stream, long boxLength)
+        => stream.Skip((int)boxLength);
+
+    private Span<byte> ReadIntoBuffer(BufferedReadStream stream, long length)
+    {
+        if (length <= this.buffer.Length)
+        {
+            stream.Read(this.buffer, 0, (int)length);
```
antonfirsov, line 641: `Stream.Read/ReadAsync(buffer)` is not guranteed to read `buffer.Length` bytes. Ignoring the return value may lead to bugs (#1707) or even worse, critical security vulnerabilities (#2550 fixed an infinite loop that could be triggered by premature EOF).

The return value should be **always** checked.

src/ImageSharp/Formats/Heif/HeifDecoderCore.cs
```diff
@@ -0,0 +1,684 @@
...
+            return this.buffer;
+        }
+        else
+        {
+            Span<byte> temp = new byte[length];
```
antonfirsov, line 646: This should be allocated/deallocated by the `MemoryAllocator`.

src/ImageSharp/Formats/Heif/HeifDecoderCore.cs
ynse01: Corrected all occurrences.

src/ImageSharp/Formats/Heif/HeifDecoderCore.cs
ynse01: Fixed by solution for previous comment

ynse01:
@JimBobSquarePants Let me know if you still want to go ahead with this PR.

src/ImageSharp/Formats/Heif/HeifEncoderCore.cs
JimBobSquarePants: No need then thanks.

JimBobSquarePants:
> @JimBobSquarePants Let me know if you still want to go ahead with this PR.

Hi @ynse01 I absolutely do want to continue but we'll need to get AV1 format together before I can merge the code. Thanks!

ynse01:
> > @JimBobSquarePants Let me know if you still want to go ahead with this PR.
> 
> Hi @ynse01 I absolutely do want to continue but we'll need to get AV1 format together before I can merge the code. Thanks!

AV1 is quite complex with all its options, especially for implementing an efficient Encoder.

Would it be easier for you, if we focus first on getting the pure HEIF container support in, including all metadata (ICC, Exif and XMP) working ?
AV1 compression could than be a separate next step.

JimBobSquarePants:
My concern here would be merging an incomplete format. We’ll have to make all the parts internal if you want to do things in that order.

ynse01:
I think I understand what you're trying to tell me:
From a user perspective, parsing only metadata out of an image is not considered _supporting_ a format. Users expect pixels out of an image.
Which would mean I guess, that the minimal viable PR to be merged in would be something like a first image round-trip, using 1 chosen set of parameters. That would allow for adding the other (many) variations relatively quickly and keep up with user's expectations.

The downside I see with this approach, is that the internal design of the implementation doesn't get any feedback until the very end, making design changes a bit more painful. Not to mention the size of the first PR ...

Now that I know you would like to have the AVIF format supported, I'm motivated again to continue to implement it.

@JimBobSquarePants Let me know your thoughts on the approach, order and intermediate goals of AVIF support.

JimBobSquarePants:
> I think I understand what you're trying to tell me: From a user perspective, parsing only metadata out of an image is not considered _supporting_ a format. Users expect pixels out of an image. Which would mean I guess, that the minimal viable PR to be merged in would be something like a first image round-trip, using 1 chosen set of parameters. That would allow for adding the other (many) variations relatively quickly and keep up with user's expectations.
> 
> The downside I see with this approach, is that the internal design of the implementation doesn't get any feedback until the very end, making design changes a bit more painful. Not to mention the size of the first PR ...
> 
> Now that I know you would like to have the AVIF format supported, I'm motivated again to continue to implement it.
> 
> @JimBobSquarePants Let me know your thoughts on the approach, order and intermediate goals of AVIF support.

Yes exactly. If we ship with support for a format, it must support the full decode/encode process for one compression format. Our users expect that now as that's how we've always worked. 

I would implement a single compression (_the most common if you know that_) algorithm then iterate once we know that is complete. Addition compression algorithms can be separate PRs. 

Regarding reviewing the code. This is why it's good to have the PR open early so we can talk through changes as they're written. I try to read changes as soon as they appear in PRs but if there's ever anything you explicitly want me to look at just tag me. 

I must say though, the quality of code you are delivering is very high!!

ynse01:
Sounds like a plan !

The user community of AVIF encoders seems to be quite scattered, can't identify a dominant force there. Feel free to provide real-life images, so I can direct the support towards these.

For now, I plan for the following initial limitations to AV1 compression:
- 8 bit depth only
- Color pixels only
- Base Profile for now
- Square block sizes
- _Split_ partition type only, to keep the blocks square
- _DCT_ transform type only
- Limited selection of Prediction modes
- Limited subsampling options
- No filters
- No alpha
- No sequences / animation

I do plan to take both lossy and lossless into account from the start.

@JimBobSquarePants In the meantime, you could help my local workflow, by adding the relevant file extensions to GIT LFS in Shared-infrastructure please:
```
*.heic              filter=lfs diff=lfs merge=lfs -text
*.hif               filter=lfs diff=lfs merge=lfs -text
*.avif              filter=lfs diff=lfs merge=lfs -text
```

JimBobSquarePants:
> Sounds like a plan !
> 
> The user community of AVIF encoders seems to be quite scattered, can't identify a dominant force there. Feel free to provide real-life images, so I can direct the support towards these.
> 
> For now, I plan for the following initial limitations to AV1 compression:
> 
> * 8 bit depth only
> * Color pixels only
> * Base Profile for now
> * Square block sizes
> * _Split_ partition type only, to keep the blocks square
> * _DCT_ transform type only
> * Limited selection of Prediction modes
> * Limited subsampling options
> * No filters
> * No alpha
> * No sequences / animation
> 
> I do plan to take both lossy and lossless into account from the start.
> 
> @JimBobSquarePants In the meantime, you could help my local workflow, by adding the relevant file extensions to GIT LFS in Shared-infrastructure please:
> 
> ```
> *.heic              filter=lfs diff=lfs merge=lfs -text
> *.hif               filter=lfs diff=lfs merge=lfs -text
> *.avif              filter=lfs diff=lfs merge=lfs -text
> ```

That all sounds great. Sorry about the delay, I've been very swamped. 

If you use the following command in your fork you will get those changes. 

```
git submodule foreach git pull origin main
```

src/ImageSharp/Formats/Heif/HeifDecoderCore.cs
```diff
@@ -0,0 +1,705 @@
...
+        Span<byte> boxBuffer = boxMemory.GetSpan();
+        byte version = boxBuffer[0];
+        int bytesRead = 4;
+        this.primaryItem = ReadUInt16Or32(boxBuffer, version != 0, ref bytesRead);
+    }
+
+    private void ParseItemProperties(BufferedReadStream stream, long boxLength)
+    {
+        // Cannot use Dictionary here, Properties can have multiple instances with the same key.
+        List<KeyValuePair<Heif4CharCode, object>> properties = new();
```
brianpopow, line 406: @ynse01 shouldn't those properties be bound to an item? Otherwise they will be lost here after parsing or do I miss something?

brianpopow:
@ynse01 The `Av1BitStreamReader` seems to work as it should, I was able to parse `Av1CodecConfiguration` with it d8e7078b. Maybe I can help with this PR, but I still in the process of reading the spec, there is a lot to grasp.

src/ImageSharp/Formats/Heif/HeifDecoderCore.cs
ynse01: Sure they are connected, this is hanlded by the `ipma` box or `ParsePropertyAssociation()` in the code.

ynse01:
> @ynse01 The `Av1BitStreamReader` seems to work as it should, I was able to parse `Av1CodecConfiguration` with it [d8e7078](https://github.com/SixLabors/ImageSharp/commit/d8e7078bb994fb197aca2d27d2468d930b9f2f54). Maybe I can help with this PR, but I still in the process of reading the spec, there is a lot to grasp.

There sure is a lot of grasp for AVIF. 

The parsing of `ObuSequenceHeader` seems to work OK indeed, but somewhere in either `ObuFrameHeader` or in the handling of TileGroups there is a bug. The way AVIF (or technically Av1) intertwines BitStream coding with Symbol coding is pretty hard to debug.

brianpopow:
> I fixed the bug :-) Tests are passing again.
> 
> @brianpopow There are a few TODO items in the Heif container area, if you like you could have a look at:
 ...
> I'll dive deeper into the Symbol coding and parsing the coefficients of the TileInfoBlocks.

Good job in finding the bug! I had some time on the weekend too look into your code. It's impressive howmuch is already done with the OBU reader. I thought maybe I could start with something small, maybe implementing parsing `5.9.14. Segmentation params` as a start?

ynse01:
> > I fixed the bug :-) Tests are passing again.
> > @brianpopow There are a few TODO items in the Heif container area, if you like you could have a look at:
> > ...
> > I'll dive deeper into the Symbol coding and parsing the coefficients of the TileInfoBlocks.
> 
> Good job in finding the bug! I had some time on the weekend too look into your code. It's impressive howmuch is already done with the OBU reader. I thought maybe I could start with something small, maybe implementing parsing `5.9.14. Segmentation params` as a start?

I see the _implement-another-spec-section_-virus has gotten to you also. If you like more of these, feel free to do so.
As a goal, we could support images that do not use the _reduced still picture header_ feature and explicitly contain every OBU module out there. The test image _Irvine_CA.avif_ is such an example, in `ObuFrameHeaderTest` you can uncomment it and work your way through the the OBU headers.

ynse01:
@brianpopow I'm following the code path of the SVT-AV1 library, which seems to deviate quite a bit in the latest Syntax's. From Residual and down, I don't think the spec and the code map 1-to-1.
As the SVT-AV1 library has extensive intrinsic parts, I would like to leverage those and stick close to their implementation.

brianpopow:
> @brianpopow I'm following the code path of the SVT-AV1 library, which seems to deviate quite a bit in the latest Syntax's. From Residual and down, I don't think the spec and the code map 1-to-1. As the SVT-AV1 library has extensive intrinsic parts, I would like to leverage those and stick close to their implementation.

OK, thanks for the update, I was following the spec very closely. I will take a look into SVT-AV1 library. I have seen that I have implemented GetPlaneResidualSize() in 13173a39, which you already have done in a another place, that was a waste of time, I have overlooked that.

brianpopow:
> @brianpopow I'm following the code path of the SVT-AV1 library, which seems to deviate quite a bit in the latest Syntax's.

I tried to look into the SVT-AV1 repository, but I have some understanding issues, @ynse01 maybe you can help me understand what you are using as the reference code base. Just for clarification, when you say SVT-AV1 library you mean this repo: https://gitlab.com/AOMediaCodec/SVT-AV1 right? If so, I only see an encoder here, where is the decoder part?

ynse01:
> > @brianpopow I'm following the code path of the SVT-AV1 library, which seems to deviate quite a bit in the latest Syntax's.
> 
> I tried to look into the SVT-AV1 repository, but I have some understanding issues, @ynse01 maybe you can help me understand what you are using as the reference code base. Just for clarification, when you say SVT-AV1 library you mean this repo: https://gitlab.com/AOMediaCodec/SVT-AV1 right? If so, I only see an encoder here, where is the decoder part?

Indeed, that is the repository that I'm used as reference. The decoder has been **removed** recently, in [this](https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/bd70bf1a3940965e20f33607d12b8f46a3b197db) commit. As I've used an earlier version of that exact decoder as the basis for my code here. This is a bit of a worry from a maintenance point of view.

The SVT-AV1 library is still one of the more efficient encoders out there and its license is permissive. Hopefully we can learn a couple of tricks from them on the encoder.

brianpopow:
Hi @ynse01, I still have some problems understanding howto decode a single image with the SVT-AV1 library.

I am using now the tagged version `v2.1.0`. I can see now there is an `SvtAv1DecApp`, but that seems to be for video files, if I am not mistaken. Is it not possible to decode just a single avif file? This would make comparing our implementation with the reference much easier, but I struggle to understand how this can be done.

ynse01:
@brianpopow Running the decoder app on an AVIF file will not work. I don't think SVT-AV1 has the HEIF container parsing capabilities. It seems to use the IVF file format, but it supports more, see [here](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/v2.1.0/Source/App/DecApp/EbFileUtils.h?ref_type=tags#L40). It appears to me that IVF can contain an AV1 bitstream although its meant for VP8. And yes, I know this whole container / bitstream and their naming is pretty confusing. 

To summarize the terminology: Heif is the ISO base container (also known as ISOBMFF), where AV1 is just one of the supported bitstream formats. AVIF is to be interpreted here as an AV1 bitstream inside a Heif container, where the bitstream contains exactly **one** intra frame. With the same terminology, Section 1 of [AVIF spec](https://storage.googleapis.com/downloads.aomedia.org/assets/pdf/AVIF-v1.1.0.pdf) also touched on this briefly.

The workflow I used to "debug" SVT-AV1 is considerably less user friendly from what you suggest. I mainly used modified unit tests from their own test suite, to get the information that I need. Alternatively, when you convert the AVIF file into IVF, you could get it visualized in a debugger.

brianpopow:
@ynse01 I have noticed an issue in the bitstream reader in `ReadLiteral()`. When you read to the end of  the stream the byte position would be advanced, resulting in an Index out of range issue. I have added a test demonstrating the issue: 62728b6f

Also when there is only 4 bytes of data available, the initialization would also result in an index out of range issue, because the `AdvanceToNextWord` would be called twice in the constructor. I did noticed this when decoding example image `jpeg444_xnconvert.avif` when av1CodecConfiguration was parsed in `HeifDecoderCore` -> `ParsePropertyContainer()`

I have tried to adjust the `ReadLiteral()` implementation to fix that, but I did not want to override your code without asking for your opinion first. I have added a new `Av1BitStreamReader2` to demonstrate this. It's adapted from [libgav1](https://chromium.googlesource.com/codecs/libgav1), because I found the implementation of Svt-AV1 too complicated.

ynse01:
@brianpopow I like your new `Av1BitStreamReader`, for 3 reasons:

- Fixes existing bug with reading large integers
- Doesn't need extra padding bytes at the end
- Has less jumps, which is more efficient with modern (Intel) CPU's

So, replaced the implementation. I did rework your suggested code a bit and implemented switching to Symbol reading again.

JimBobSquarePants:
@ynse01 Just a heads up. I'm going to be merging PRs #2751 #2762 very soon which contain some breaking changes to our codec APIs. If you have any questions about how to correctly implement the changes following merge, please let me know.

src/ImageSharp/Formats/Heif/Av1/Transform/Forward/Av1Dct4ForwardTransformer.cs
```diff
@@ -35,51 +68,71 @@ public static void TransformSse(ref Vector128<int> input, ref Vector128<int> out
         Vector128<int> v0, v1, v2, v3;
 
         int endidx = 3 * columnNumber;
-        s0 = Sse41.Add(input, Unsafe.Add(ref input, endidx));
-        s3 = Sse41.Subtract(input, Unsafe.Add(ref input, endidx));
+        s0 = Sse2.Add(input, Unsafe.Add(ref input, endidx));
```
JimBobSquarePants, line 71: If possible, we should look to use `Vector128` opreators and methods here that should abstract the instrinsic support.

I'm a little nervous also about the lack of contraints here. Are they strictly defined in the caller?

src/ImageSharp/Formats/Heif/Av1/Transform/Forward/Av1Dct4ForwardTransformer.cs
```diff
@@ -10,7 +10,39 @@ namespace SixLabors.ImageSharp.Formats.Heif.Av1.Transform.Forward;
...
+        step3 = HalfBtf(cospi[48], output3, -cospi[16], output2, cosBit);
+
+        // stage 3
+        output = step0;
+        output1 = step2;
+        output2 = step1;
+        output3 = step3;
+    }
 
     public void TransformAvx2(ref Vector256<int> input, ref Vector256<int> output, int cosBit, int columnNumber)
```
JimBobSquarePants, line 47: I would try to normalize these. Having specific public Avx2/SSe methods is something we want to remove from the codebase.

ynse01: I intend to leave 2 code paths, the scalar path and an AVX2 path. Thinking out loud for putting in an exception for the 4x4 case, as it operates on a very narrow strip of memory.

src/ImageSharp/Formats/Heif/Av1/Transform/Forward/Av1Dct4ForwardTransformer.cs
ynse01: The constraints are handled inside the `Av1BlockDecoder.DeriveBlockPointers` method. Please note that the `Av1FrameBuffer` is over allocating, such that the last blocks still fit in entirely (even if the resulting image is smaller). 

Please note, the `Transform` method will be called twice for every 16 pixels, for each color channel. In other words, this is bound to become a performance hotspot !

JimBobSquarePants: The operators and methods are hardware accelerated though and provide additional support for AdvSimd and PackedSimd

src/ImageSharp/Formats/Heif/Av1/Transform/Forward/Av1Dct4ForwardTransformer.cs
JimBobSquarePants: My concern here is with implementation detail leaking into the API shape. Im slowly refactoring the library to ensure internal method callers have no requirement to know the level of hardware support present. The runtime is moving that way also.

ynse01: I guess you mean to have something like this:
``` cs
internal class Av1Dct4ForwardTransformer : IAv1ForwardTransformer
{
  public void Transform(...)
  {
    {
      TransformSse(...);
    }
    else
    {
      TransformScalar();
    }
  }

  private void TransformScalar(...) 
  {
    ...
  }

  private void TransformSse(...)
  {
    ...
  }
}
```

JimBobSquarePants: Yeah exactly. Thanks!

ynse01:
@JimBobSquarePants Since a few days the Test runs on (at least the Linux) CI servers fails on a missing `libgdiplus` library file. Is this an issue specific to this branch? 
Asking as I didn't change any of the related code.

stefannikolei:
> @JimBobSquarePants Since a few days the Test runs on (at least the Linux) CI servers fails on a missing `libgdiplus` library file. Is this an issue specific to this branch? 
> 
> Asking as I didn't change any of the related code.

This is because of usage of Ubuntu:latest. The base image changed and libgdiplus is missing. 

I have a fix added to my open pr

JimBobSquarePants:
> > @JimBobSquarePants Since a few days the Test runs on (at least the Linux) CI servers fails on a missing `libgdiplus` library file. Is this an issue specific to this branch?
> > Asking as I didn't change any of the related code.
> 
> This is because of usage of Ubuntu:latest. The base image changed and libgdiplus is missing.
> 
> I have a fix added to my open pr
> 
> This commit fixes the problem [0a7cf55](https://github.com/SixLabors/ImageSharp/commit/0a7cf55fe0c14283242c3a69f8a61cd2e24fd56e)

Thanks guys, yeah spotted this, just taking a few days break but will get fixes merged soon. Happy New Year!

stefannikolei:
Enjoy your deserved break

jaddie:
Not sure if its suitable to do so or not, but putting this comment on as a appreciation of the effort but also a vote for avif support and anyone seeing this who is after it can thumbs this up to show their desire too as it would be great as the storage saving is fantastic as a result of avif vs jpg for sure along with the wide support existing now.

axtox:
is this still active? really waiting for this feature!

ynse01:
> is this still active? really waiting for this feature!

I'm still here. But after several attempts to crack the Entropy Encoding, I figured I need a break to start with a fresh mind.
So, unfortunately, this PR will take a little bit more time to complete than expected.

axtox:
> > is this still active? really waiting for this feature!
> 
> I'm still here. But after several attempts to crack the Entropy Encoding, I figured I need a break to start with a fresh mind. So, unfortunately, this PR will take a little bit more time to complete than expected.

I am now adopting [ImageSharp.Heif](https://github.com/hey-red/ImageSharp.Heif) nuget package, but it's out of date and does not work with iPhone >15 HEIF formats because it is using libheif version [1.15.1](https://www.nuget.org/packages/LibHeif.Native), meanwhile the latest is [1.20.2](https://github.com/strukturag/libheif/releases/tag/v1.20.2). I even tried to build C++ native libs that they are using - no luck. Check out the [issue in their repo](https://github.com/hey-red/ImageSharp.Heif/issues/4) if you're curious.

maybe you can peek some ideas on cracking the Entropy Encoding from the original libd265/libheif authors repo from [here](https://github.com/strukturag/libde265) (libd265, the h.265 codec for image and video processing, written in C++), or [here](https://github.com/strukturag/libheif) (libheif, written in C++ uses libd265 for HEIC decode and other libs to encode and decode formats), or maybe [here](https://github.com/0xC0000054/libheif-sharp) (libheif-sharp, written with C# and uses C++ libheif API calls)?

both C# [ImageSharp.Heif](https://github.com/hey-red/ImageSharp.Heif) and [libheif-sharp](https://github.com/0xC0000054/libheif-sharp) projects are using native C++ libs API's that are built and packed as nuget package with C++ binaries

maybe that'll help! thank you for your effort, looking forward to hear any updates from you.

jmderivco:
I have implemented native integration with the [libavif](https://github.com/AOMediaCodec/libavif) lib to encode and decode AVIF images. I used the approach that I saw in [BrycensRanch/NeoSolve.ImageSharp.AVIF](https://github.com/BrycensRanch/NeoSolve.ImageSharp.AVIF) repo.

It works pretty well. The encode / decode process is just very slow because of the overhead of integrating with a native lib. That being said, the advantages of using AVIF is totally worth it. The compression file size / image quality trade-off is super impressive. If you pre-compute your AVIF images and / or have a good caching strategy, this is a decent stop-gap solution until ImageSharp can do the encoding / decoding via managed code.

timoschraa:
@JimBobSquarePants / @dlemstra Any updates on this?

JimBobSquarePants:
Hey @ynse01 

Could you please do me a favour and grant me full write access to your fork? I want to help get this over the line and Git LFS can cause problems on GitHub.

Thanks!

ynse01:
> Hey @ynse01
> 
> Could you please do me a favour and grant me full write access to your fork? I want to help get this over the line and Git LFS can cause problems on GitHub.
> 
> Thanks!

Sure, would love some help !

I've send you the invite to collaborate in my fork.

winscripter:
@JimBobSquarePants I noticed there's an H.265/HEVC codec in this fork.

I thought it might be helpful to flag this. This codec is covered by several patent pools and is **not** royalty-free. A license must be obtained to distribute an implementation of this codec to users. ⚖️

AV1 is royalty-free though, no concern about that codec.

JimBobSquarePants:
> @JimBobSquarePants I noticed there's an H.265/HEVC codec in this fork.
> 
> I thought it might be helpful to flag this. This codec is covered by several patent pools and is **not** royalty-free. A license must be obtained to distribute an implementation of this codec to users. ⚖️
> 
> AV1 is royalty-free though, no concern about that codec.

Fixed. I mistakenly let AI have a crack there. I shouldn't have been surprised it didn't give a toss about patents. Thanks!

Issue #2067, opened 2022-03-17
Does Jpeg format need comment marker support?


JimBobSquarePants:
### Discussed in https://github.com/SixLabors/ImageSharp/discussions/2061

<div type='discussions-op-text'>

<sup>Originally posted by **br3aker** March 13, 2022</sup>
Current jpeg implementation lacks of COM marker support, does it need to support it? Should be easy to implement as this marker is just an array of bytes - itu spec leaves 'interpretation to the application', decoding API should not try to decode it as it can be anything from literal byte array to some weird text encoding.

So it can be a 'good first issue' or I can work on it a bit later.</div>

prabhavmehra:
any updates on this? I was looking to use it from ImageSharp (as I do use it for bunch of other stuff) or go about implementing it myself.

br3aker:
I think this place is better for a small guide than that discussion @prabhavmehra.

First of all, COM marker is already hooked up in the decoding loop [here](https://github.com/SixLabors/ImageSharp/blob/main/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs#L483)

The only thing left to do is to actually parse given stream which is skipped right now using this line: `stream.Skip(markerContentByteSize);`.

And the final question is where to store parsed results. Results should be saved into [JpegMetadata class](https://github.com/SixLabors/ImageSharp/blob/d93bc6cca2383095c2a08db85861b88a1c94c43c/src/ImageSharp/Formats/Jpeg/JpegMetadata.cs). AFAIR we decided to implement it as an `ICollection<char[]>` property backed by a `List<char[]>`.

The only thing that bothers me is what should we do if there are no COM markers in given JPEG, leave new property as null or an empty ICollection...

JimBobSquarePants:
How about a nullable `Memory<char>`?

tocsoft:
I would like to add in here that I feel reading `COM` markers should be explicitly opt-in. (could be exposed as a new property on `JpegDecoderOptions`)
The writing side of things however can be left as implicitly opt-in by the fact we have the metadata set or not.

I feel it needs to be opt-in for a couple of reasons.
1. It feels like the usage would be relatively niche for our general users.
2. We don't want to be penalizing general users/usage having to keep hold of more memory when most users would never want it.
3. It would now cause larger file outputs as we would start currying the data in & then out for those users that don't even know this thing exists.

br3aker:
> How about a nullable Memory<char>?

[...]

PR #2641, opened 2024-01-07
Add JPEG COM marker support

RobertMut:
### Prerequisites

### Description
This PR adds support for COM marker in JPEG as mentioned in #2067 and #2061

Implemented:
-  Methods for jpeg encoder and decoder
- `ICollection<Memory<char>>` metadata property (#2067),
-  SetComment (#2061),
-  GetComment (#2061),
-  ClearComments(#2061)


JimBobSquarePants:
Thanks for this! I'll review it asap.

tests/ImageSharp.Tests/Formats/Jpg/JpegEncoderTests.Metadata.cs
```diff
@@ -154,6 +154,50 @@ public void Encode_PreservesQuality(string imagePath, int quality)
         }
     }
 
+    [Theory]
+    [WithFile(TestImages.Jpeg.Issues.Issue2067_CommentMarker, PixelTypes.Rgba32)]
+    public void Encode_PreservesComments<TPixel>(TestImageProvider<TPixel> provider)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        // arrange
+        using var input = provider.GetImage(JpegDecoder.Instance);
```
JimBobSquarePants, line 163: Please favour explicit types and target typed new expressions

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
```diff
@@ -101,6 +103,11 @@ public int Quality
     /// </remarks>
     public bool? Progressive { get; internal set; }
 
+    /// <summary>
+    /// Gets the comments.
+    /// </summary>
+    public ICollection<Memory<char>>? Comments { get; }
```
JimBobSquarePants, line 109: I don't think this should be nullable. Collections should always have a backing value in a publi API. I would use `IList` also to prevent all the casting and maintain consistency with the `PngMetadata.TextDataProperty`

On that note we could use a specific struct `JpegComData` which would contain the backing memory and can have static `FromString(string value)`

```c#
public readonly struct JpegComData
{
	public static JpegComData FromString(string value) => new(value.AsMemory());

	public JpegComData(ReadOnlyMemory<char> value) 
		=> this.Value = value;
	
	
	public ReadOnlyMemory<char> Value { get; }

	public override string ToString() => this.Value.ToString();
}
```

This would also remove the need for any extension methods.

JimBobSquarePants (changes requested):
Thanks for submitting this and sorry for the slow review. I think we can improve upon our initial discussed design, so I've added comments.

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
JimBobSquarePants: @br3aker What do you think?

br3aker: Hi!
I'm in favor of an empty collection instead of a null one - it's better from a user perspective.
Not so sure about `JpegComData`, do we really need to expose comments as `ReadOnlyMemory<char>` instead of a plain string?

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -167,6 +171,47 @@ private void WriteJfifApplicationHeader(ImageMetadata meta, Span<byte> buffer)
         this.outputStream.Write(buffer, 0, 18);
     }
 
+    /// <summary>
+    /// Writes comment
+    /// </summary>
+    /// <param name="metadata">The image metadata.</param>
+    private void WriteComment(JpegMetadata metadata)
+    {
+        if (metadata.Comments is { Count: 0 })
```
br3aker, line 180: Can't we just use `metadata.Comments.Length == 0`?

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -167,6 +171,47 @@ private void WriteJfifApplicationHeader(ImageMetadata meta, Span<byte> buffer)
...
+        {
+            return;
+        }
+
+        // Length (comment strings lengths) + (comments markers with payload sizes)
+        int commentsBytes = metadata.Comments.Sum(x => x.Length) + (metadata.Comments.Count * 4);
```
br3aker, line 186: I'm against using LINQ at all. While this won't slow down jpeg codec overall it's slower than calculating sum yourself :)

src/ImageSharp/Formats/Jpeg/MetadataExtensions.cs
```diff
@@ -17,4 +18,37 @@ public static partial class MetadataExtensions
...
     /// <returns>The <see cref="JpegMetadata"/>.</returns>
     public static JpegMetadata GetJpegMetadata(this ImageMetadata metadata) => metadata.GetFormatMetadata(JpegFormat.Instance);
+
+    /// <summary>
+    /// Sets the comment in <see cref="JpegMetadata"/>
+    /// </summary>
+    /// <param name="metadata">The metadata this method extends.</param>
+    /// <param name="index">The index of comment to be inserted to.</param>
+    /// <param name="comment">The comment string.</param>
+    public static void SetComment(this JpegMetadata metadata, int index, string comment)
```
br3aker, line 28: Do we really need such methods if comments are exposed as `IList`?
`IList` has all this methods so users can do whatever they want.

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
JimBobSquarePants: I thought we couldn't guarantee that the data was always an ASCII string and that's why we had a collection of char?

src/ImageSharp/Formats/Jpeg/MetadataExtensions.cs
JimBobSquarePants: My thoughts exactly.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
br3aker: Ah yes, too big comment data must be scattered amongst multiple markers as I recall correctly.

src/ImageSharp/Formats/Jpeg/JpegMetadata.cs
br3aker: > I thought we couldn't guarantee that the data was always an ASCII string and that's why we had a collection of char?

True, if I recall correctly jpeg specification doesn't specify anything about comment section contents.
Why would we need `FromString(string value)` method though?

JimBobSquarePants: Sorry @br3aker I was confinced I'd replied to this!!

I'm just covering the most common use case, ASCII strings, in a nice manner. Saves using `String.AsMemory()` which feels a little clunky.

RobertMut:
I've made changes as you mentioned in PR comments. I left them open, because I don't know what the usual practice is for closing them (by the reviewer or contributor) on your repo. 
I hope we're on the same page.
Feel free to comment if I have missed/misunderstood something or if there is a better way to do this.

src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
```diff
@@ -515,6 +516,22 @@ public void Dispose()
...
+    /// </summary>
+    /// <param name="stream">The input stream.</param>
+    /// <param name="markerContentByteSize">The remaining bytes in the segment block.</param>
+    private void ProcessComMarker(BufferedReadStream stream, int markerContentByteSize)
+    {
+        Span<byte> temp = new byte[markerContentByteSize];
+        JpegMetadata metadata = this.Metadata.GetFormatMetadata(JpegFormat.Instance);
+
+        stream.Read(temp);
+        string comment = Encoding.ASCII.GetString(temp);
```
JimBobSquarePants, line 530: We're assuming here that the input is always ASCII when we do not actually know that. I would create a `char[]` cast it to `byte[]`, read the stream and pass it to the `JpegComData` constructor.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -167,6 +171,54 @@ private void WriteJfifApplicationHeader(ImageMetadata meta, Span<byte> buffer)
...
+        }
+
+        for (int i = 0; i < metadata.Comments.Count; i++)
+        {
+            string comment = metadata.Comments[i].ToString();
+
+            if (comment.Length > maxCommentLength)
+            {
+                string splitComment = comment.Substring(maxCommentLength, comment.Length - maxCommentLength);
+                metadata.Comments.Insert(i + 1, JpegComData.FromString(splitComment));
```
JimBobSquarePants, line 194: We should not be changing the metadata here. Encoding should leave the image unchanged.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -167,6 +171,54 @@ private void WriteJfifApplicationHeader(ImageMetadata meta, Span<byte> buffer)
...
+
+                // We don't want to keep the extra bytes
+                comment = comment.Substring(0, maxCommentLength);
+            }
+
+            int commentLength = comment.Length + 4;
+
+            Span<byte> commentSpan = new byte[commentLength];
```
JimBobSquarePants, line 202: You should be able to avoid the allocation here by using a single buffer of the maximum length plus markers outside of the loop and slicing.

src/ImageSharp/Formats/Jpeg/JpegEncoderCore.cs
```diff
@@ -167,6 +171,54 @@ private void WriteJfifApplicationHeader(ImageMetadata meta, Span<byte> buffer)
...
+            // Beginning of comment ff fe
+            markers[0] = JpegConstants.Markers.XFF;
+            markers[1] = JpegConstants.Markers.COM;
+
+            // Write payload size
+            int comWithoutMarker = commentLength - 2;
+            payloadSize[0] = (byte)((comWithoutMarker >> 8) & 0xFF);
+            payloadSize[1] = (byte)(comWithoutMarker & 0xFF);
+
+            Encoding.ASCII.GetBytes(comment, payload);
```
JimBobSquarePants, line 216: We don't know it's ASCII. I would cast and write.

JimBobSquarePants (approved):
@RobertMut I reworked your implementation to remove the extra allocations. Thanks for your help here!

PR #2645, opened 2024-01-15
Modernize and optimize pixel format operations across platforms.

### LINKED ISSUE #2232: Performance optimization opportunities in common pixel formats.
### Prerequisites

### ImageSharp version

v3 alpha + 

NA

NA

### .NET Framework version

NA

### Description

As described [here](https://github.com/SixLabors/ImageSharp/pull/2230#issuecomment-1248800429) there are several performance opportunities can be implemented in many of our pixel format types. This should be fairly low hanging fruit with good return.

Notably on .NET 6/7, you could make this even more efficient by doing something like:
```csharp
public void Pack(Vector4 vector)
{
    vector *= MaxBytes;
    vector += Half;
    vector = Vector4.Clamp(vector, Vector4.Zero, MaxBytes);

    Vector128<byte> result = Sse2.ConvertToVector128Int32WithTruncation(vector.AsVector128()).AsByte();
    // In .NET 7+ the above can be `result = Vector128.ConvertToInt32(vector.AsVector128()).AsByte()` so it works on Arm64 too

    R = result.GetElement(0);
    G = result.GetElement(4);
    B = result.GetElement(8);
    A = result.GetElement(12);
}
```

This converts all 4 elements at once and then extracts the truncated bytes directly:
```asm
vzeroupper
vmovupd xmm0, [0x7ffd160105c0]
vmovaps xmm1, xmm0
vmulps xmm1, xmm1, [rdx]
vmovupd [rdx], xmm1
vmovupd xmm1, [0x7ffd160105d0]
vaddps xmm1, xmm1, [rdx]
vmovupd [rdx], xmm1
vmovupd xmm1, [rdx]
vxorps xmm2, xmm2, xmm2
vmaxps xmm1, xmm1, xmm2
vminps xmm0, xmm1, xmm0
vmovupd [rdx], xmm0
vcvttps2dq xmm0, [rdx]
vpextrb eax, xmm0, 0
mov [rcx+2], al
vpextrb eax, xmm0, 4
mov [rcx+1], al
vpextrb eax, xmm0, 8
mov [rcx], al
vpextrb eax, xmm0, 0xc
mov [rcx+3], al
ret
```
* We'll also be improving the codegen around `vpextrb` more in the future so it can be just `vpextrb [rcx+2], xmm0, 0` instead of `vpextrb eax, xmm0, 0` followed by `mov [rcx+2], al`.

You can also optimize in .NET 6+ by directly using `Vector128.Create()`. This creates a method local constant and avoids the static initializer entirely:
```csharp
    private static Vector4 MaxBytes => Vector128.Create(255f).AsVector4();
    private static Vector4 Half => Vector128.Create(0.5f).AsVector4();
```

### Steps to Reproduce

NA

### Images

_No response_

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2232 and #594

- [x] Updates API to use static interface methods for improved usability and extensibility.
- [x] Optimizes and modernizes SIMD pixel shuffle operations, adding ARM support.
- [ ] ~Optimizes and modernizes SIMD Vector4 packing operations.~ I'm leaving this for a follow up PR

I don't expect anyone to reasonably find the time to analyze this line-by-line as it's a massive piece of work. However, changes on the whole are simplistic and the developer benefit massive. 

### API change Elevator Pitch

The abomination that is.
```c#
TPixel pixel = default;
pixel.FromScaledVector(vector)
```

Becomes, as nature intended.

```c#
TPixel pixel = TPixel.FromScaledVector(vector);
```

All other conversion methods follow suit where applicable.


src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /// <summary>
+    /// Shifts a 128-bit value right by a specified number of bytes while shifting in zeros.
+    /// </summary>
+    /// <param name="value">The value to shift.</param>
+    /// <param name="numBytes">The number of bytes to shift by.</param>
+    /// <returns>The <see cref="Vector128{Byte}"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector128<byte> ShiftRightBytesInVector(Vector128<byte> value, [ConstantExpected(Max = (byte)15)] byte numBytes)
```
JimBobSquarePants, line 110: There's something not right in one of the methods below on Arm64.

@tannergooding  @saucecontrol Would either of you be able to do a quick readthrough and set me right?

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -0,0 +1,174 @@
...
+    public static Vector128<byte> ShiftLeftBytesInVector(Vector128<byte> value, [ConstantExpected(Max = (byte)15)] byte numBytes)
+    {
+        if (Sse2.IsSupported)
+        {
+            return Sse2.ShiftLeftLogical128BitLane(value, numBytes);
+        }
+
+        if (AdvSimd.IsSupported)
+        {
+            return AdvSimd.ExtractVector128(Vector128<byte>.Zero, value, numBytes);
```
saucecontrol, line 142:
```suggestion
            return AdvSimd.ExtractVector128(Vector128<byte>.Zero, value, 16 - numBytes);
```

This is the only thing that looks amiss to me.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: Figured it out. It was left shift. I'd forgotten to offset.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: Yeah that's it.

src/ImageSharp/Common/Helpers/Numerics.cs
```diff
@@ -61,6 +60,12 @@ public static int LeastCommonMultiple(int a, int b)
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static nint Modulo4(nint x) => x & 3;
 
+    /// <summary>
+    /// Calculates <paramref name="x"/> % 4
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static nuint Modulo4(nuint x) => x & 3;
```
tannergooding, line 67: `x % 4` should already get optimized to `x & 3` by the JIT since `x` is `nuint` (and therefore definitely unsigned)

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -140,154 +165,168 @@ private static Vector256<byte> ShuffleMaskShiftAlpha() => Vector256.Create((byte
...
+                    ref Vector512<float> vd0 = ref Unsafe.Add(ref destinationBase, i);
 
-                    vd0 = Avx.Permute(vs0, control);
-                    Unsafe.Add(ref vd0, 1) = Avx.Permute(Unsafe.Add(ref vs0, 1), control);
-                    Unsafe.Add(ref vd0, 2) = Avx.Permute(Unsafe.Add(ref vs0, 2), control);
-                    Unsafe.Add(ref vd0, 3) = Avx.Permute(Unsafe.Add(ref vs0, 3), control);
+                    vd0 = Vector512Utilities.Shuffle(vs0, control);
+                    Unsafe.Add(ref vd0, (nuint)1) = Vector512Utilities.Shuffle(Unsafe.Add(ref vs0, (nuint)1), control);
+                    Unsafe.Add(ref vd0, (nuint)2) = Vector512Utilities.Shuffle(Unsafe.Add(ref vs0, (nuint)2), control);
+                    Unsafe.Add(ref vd0, (nuint)3) = Vector512Utilities.Shuffle(Unsafe.Add(ref vs0, (nuint)3), control);
```
tannergooding, line 266: Worth noting that you could do this like the following:
```csharp
Vector512Utilities.Shuffle(Vector512.LoadUnsafe(ref sourceBase), control).StoreUnsafe(ref destinationBase).
Vector512Utilities.Shuffle(Vector512.LoadUnsafe(ref sourceBase, 1), control).StoreUnsafe(ref destinationBase, 1).
Vector512Utilities.Shuffle(Vector512.LoadUnsafe(ref sourceBase, 2), control).StoreUnsafe(ref destinationBase, 2).
Vector512Utilities.Shuffle(Vector512.LoadUnsafe(ref sourceBase, 3), control).StoreUnsafe(ref destinationBase, 3).
```

It shouldn't result in any codegen differences, but avoids needing to manipulate the `byref` and uses direct hardware intrinsic APIs, rather than the Unsafe helpers, so can be easier to read/understand at least IMO.

tannergooding (approved):
LGTM, just left a couple comments on newer APIs that might help with readability

src/ImageSharp/PixelFormats/IPixel.cs
```diff
@@ -14,135 +14,151 @@ namespace SixLabors.ImageSharp.PixelFormats;
 public interface IPixel<TSelf> : IPixel, IEquatable<TSelf>
     where TSelf : unmanaged, IPixel<TSelf>
 {
-    /// <summary>
-    /// Gets the pixel type information.
-    /// </summary>
-    /// <returns>The <see cref="PixelTypeInfo"/>.</returns>
-#pragma warning disable CA1000
-    static abstract PixelTypeInfo GetPixelTypeInfo();
```
stefannikolei, line 22: Just curious. Is there a specific reason to kick it from the interface and only have it in the implementations? Probably it was never called from the interface and only from specific implementations?

src/ImageSharp.ruleset
```diff
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <RuleSet Name="ImageSharp" ToolsVersion="17.0">
   <Include Path="..\shared-infrastructure\sixlabors.ruleset" Action="Default" />
-  <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.Features" RuleNamespace="Microsoft.CodeAnalysis.CSharp.Features">
+    <Rule Id="IDE0290" Action="None" />
```
tannergooding, line 5: This should also be suppressible via `.editorconfig` using `csharp_style_prefer_primary_constructors = false:none` (or a different severity level if you want to block use of primary constructors altogether, such as `silent`, `suggestion`, `warning`, or `error`)

JimBobSquarePants: Thanks. Yeah, I’ll do that upstream. Not sold on primary constructors. The syntax doesn’t sit right with me.

src/ImageSharp/PixelFormats/IPixel.cs
JimBobSquarePants: Actually, this is just temporary. I had to hack to get the solution to build because we were misusing ApproximateFloatComparer in the tests.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
JimBobSquarePants: `sourceBase` is `Vector512<float>` so `Vector512.LoadUnsafe(ref sourceBase, i + 0)` would yield `Vector512<Vector512<float>>`. I could use `float` as the base ref type but that complicates the offsetting which confuses me and causes me to make mistakes.

src/ImageSharp/Common/Helpers/Numerics.cs
JimBobSquarePants: Thanks yeah, I just kept it the same for consistency.

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
```diff
@@ -492,7 +468,7 @@ private void UncompressRle4(BufferedReadStream stream, int w, Span<byte> buffer,
                         int max = cmd[1];
                         int bytesToRead = (int)(((uint)max + 1) / 2);
 
-                        Span<byte> run = bytesToRead <= 128 ? scratchBuffer.Slice(0, bytesToRead) : new byte[bytesToRead];
+                        Span<byte> run = bytesToRead <= 128 ? scratchBuffer[..bytesToRead] : new byte[bytesToRead];
```
stefannikolei, line 471: I am curious if there is a Analyzer to enforce this style. I know that Rider has a editoconfig setting for this

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
```diff
@@ -185,7 +185,7 @@ public ImageInfo Identify(BufferedReadStream stream, CancellationToken cancellat
     {
         uint frameCount = 0;
         ImageFrameMetadata? previousFrame = null;
-        List<ImageFrameMetadata> framesMetadata = new();
+        List<ImageFrameMetadata> framesMetadata = [];
```
stefannikolei, line 188: This could be enforced by this editorconfig rule https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0028?pivots=dotnet-8-0

stefannikolei (approved):

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
stefannikolei: If you want I can look at enforcing this style in a seperate pr

src/ImageSharp/Formats/Bmp/BmpDecoderCore.cs
tannergooding: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0057

You can find most of the language style rules here: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/language-rules

src/ImageSharp/Formats/Gif/GifDecoderCore.cs
JimBobSquarePants: That would be nice thanks. The IDE has been making the suggestion for me, that and range usage also.

src/ImageSharp/PixelFormats/IPixel.cs
```diff
@@ -14,135 +14,150 @@ namespace SixLabors.ImageSharp.PixelFormats;
...
-#pragma warning restore CA1000
-
     /// <summary>
     /// Creates a <see cref="PixelOperations{TPixel}"/> instance for this pixel type.
     /// This method is not intended to be consumed directly. Use <see cref="PixelOperations{TPixel}.Instance"/> instead.
     /// </summary>
     /// <returns>The <see cref="PixelOperations{TPixel}"/> instance.</returns>
     PixelOperations<TSelf> CreatePixelOperations();
```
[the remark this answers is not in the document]
JimBobSquarePants: Good catch, updated.

src/ImageSharp/PixelFormats/IPixel.cs
```diff
@@ -14,135 +14,150 @@ namespace SixLabors.ImageSharp.PixelFormats;
...
     /// <summary>
-    /// Expands the pixel into a <see cref="Vector4"/> which is specific to the current pixel type.
-    /// The vector components are typically expanded in least to greatest significance order.
+    /// Initializes the pixel instance from an <see cref="Abgr32"/> value.
     /// </summary>
-    /// <returns>The <see cref="Vector4"/>.</returns>
-    Vector4 ToVector4();
+    /// <param name="source">The <see cref="Abgr32"/> value.</param>
+    /// <returns>The <typeparamref name="TSelf"/>.</returns>
+    static abstract TSelf FromAbgr32(Abgr32 source);
```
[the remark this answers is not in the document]
JimBobSquarePants: I've updated the benchmarks in `PixelConversion_ConvertFromRgba32` to test the static byval implementation vs byref and byval for `TestRgba`, `TestArgb`, and `TestRgbaVector`. All benchmarks look good with the JIT doing a great job of normalizing the differences. 

```
  BenchmarkDotNet v0.13.10, Windows 11 (10.0.22631.3007/23H2/2023Update/SunValley3)
  .NET SDK 8.0.200-preview.23624.5
    [Host]     : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2
    DefaultJob : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2
```

**PixelConversion_ConvertFromRgba32_Compatible**

  | Method      | Count | Mean       | Error   | StdDev  | Ratio |
  |------------ |------ |-----------:|--------:|--------:|------:|
  | ByRef       | 256   |   103.4 ns | 0.52 ns | 0.46 ns |  1.00 |
  | ByVal       | 256   |   103.3 ns | 1.48 ns | 1.38 ns |  1.00 |
  | StaticByVal | 256   |   104.0 ns | 0.36 ns | 0.30 ns |  1.01 |
  | FromBytes   | 256   |   201.8 ns | 1.30 ns | 1.15 ns |  1.95 |
  | Inline      | 256   |   106.6 ns | 0.40 ns | 0.34 ns |  1.03 |
  |             |       |            |         |         |       |
  | ByRef       | 2048  |   771.5 ns | 3.68 ns | 3.27 ns |  1.00 |
  | ByVal       | 2048  |   769.7 ns | 3.39 ns | 2.83 ns |  1.00 |
  | StaticByVal | 2048  |   773.2 ns | 3.95 ns | 3.50 ns |  1.00 |
  | FromBytes   | 2048  | 1,555.3 ns | 9.24 ns | 8.19 ns |  2.02 |
  | Inline      | 2048  |   799.5 ns | 5.91 ns | 4.93 ns |  1.04 |

**PixelConversion_ConvertFromRgba32_Permuted_RgbaToArgb**

  | Method                         | Count | Mean        | Error     | StdDev   | Ratio | RatioSD |
  |------------------------------- |------ |------------:|----------:|---------:|------:|--------:|
  | ByRef                          | 256   |   203.48 ns |  3.318 ns | 3.104 ns |  1.00 |    0.00 |
  | ByVal                          | 256   |   201.46 ns |  2.242 ns | 1.872 ns |  0.99 |    0.02 |
  | StaticByVal                    | 256   |   201.45 ns |  0.791 ns | 0.701 ns |  0.99 |    0.02 |
  | FromBytes                      | 256   |   200.76 ns |  1.365 ns | 1.140 ns |  0.99 |    0.01 |
  | InlineShuffle                  | 256   |   221.65 ns |  2.104 ns | 1.968 ns |  1.09 |    0.02 |
  | PixelConverter_Rgba32_ToArgb32 | 256   |    26.23 ns |  0.277 ns | 0.231 ns |  0.13 |    0.00 |
  |                                |       |             |           |          |       |         |
  | ByRef                          | 2048  | 1,561.54 ns | 11.208 ns | 8.751 ns |  1.00 |    0.00 |
  | ByVal                          | 2048  | 1,554.26 ns |  9.607 ns | 8.517 ns |  1.00 |    0.01 |
  | StaticByVal                    | 2048  | 1,562.48 ns |  8.937 ns | 8.360 ns |  1.00 |    0.01 |
  | FromBytes                      | 2048  | 1,552.68 ns |  7.445 ns | 5.812 ns |  0.99 |    0.01 |
  | InlineShuffle                  | 2048  | 1,711.28 ns |  7.559 ns | 6.312 ns |  1.10 |    0.01 |
  | PixelConverter_Rgba32_ToArgb32 | 2048  |    94.43 ns |  0.363 ns | 0.322 ns |  0.06 |    0.00 |

**PixelConversion_ConvertFromRgba32_RgbaToRgbaVector**

  | Method      | Count | Mean       | Error    | StdDev   | Ratio | RatioSD |
  |------------ |------ |-----------:|---------:|---------:|------:|--------:|
  | ByRef       | 256   |   448.5 ns |  4.86 ns |  4.06 ns |  1.00 |    0.00 |
  | ByVal       | 256   |   447.0 ns |  1.55 ns |  1.21 ns |  1.00 |    0.01 |
  | StaticByVal | 256   |   447.4 ns |  1.67 ns |  1.30 ns |  1.00 |    0.01 |
  |             |       |            |          |          |       |         |
  | ByRef       | 2048  | 3,577.7 ns | 53.80 ns | 47.69 ns |  1.00 |    0.00 |
  | ByVal       | 2048  | 3,590.5 ns | 43.59 ns | 36.40 ns |  1.00 |    0.02 |
  | StaticByVal | 2048  | 3,604.6 ns | 16.19 ns | 14.36 ns |  1.01 |    0.01 |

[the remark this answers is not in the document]
JimBobSquarePants: @antonfirsov Are there any additional concerns here or can I move on? I'm keen to keep moving.

src/ImageSharp/PixelFormats/IPixel.cs
```diff
@@ -14,14 +14,14 @@ namespace SixLabors.ImageSharp.PixelFormats;
 public interface IPixel<TSelf> : IPixel, IEquatable<TSelf>
     where TSelf : unmanaged, IPixel<TSelf>
 {
+#pragma warning disable CA1000 // Do not declare static members on generic types
```
tannergooding, line 17: I'm going to open an issue for this on dotnet/roslyn-analyzers. Declaring static members on generic types is kind-of the purpose of `static abstracts in interfaces` 😆

tannergooding: Looks like we have an issue already: https://github.com/dotnet/roslyn-analyzers/issues/6424

Sergio0694: I think the reason that exists is to avoid unnecessary binary size bloat for when you eg. accidentally declare members that don't actually depend on type parameters at all. We had a few of those in ImageSharp, where the bokeh blur processor had a couple static fields with some readonly buffers (like, constants), and a bunch of pixel agnostic processing methods. We saved a nice amount of binary size moving those out to a non-generic type. It would be nice if the analyzer was smarter and just avoided warning in cases where you do in fact use the type parameter(s), but kept warning otherwise.

tannergooding: Nah, the analyzer exists not being of binary bloat but because `MyClass.Foo(x)` (given `Foo<T>(T x)`) works with generic inference while `MyClass<T>.Foo(x)` does not and since static members historically couldn't be `abstract` it didn't make sense to ever have them there.

So the analyzer really just hasn't been updated to account for the existence of `static abstract in interfaces`. Once it's been made aware, I'd expect it suggest such members be explicitly implemented on the type and to defer to the `MyClass.Foo<T>(...)` implementation.

JimBobSquarePants: I’ve upvoted on the linked issue anyway

JimBobSquarePants:
Let's merge this. It's awesome.

PR #2654, opened 2024-02-02
Cleanup SimdUtils

JimBobSquarePants:
### Prerequisites

### Description

The plan here is to be able to remove `ExtendedIntrinsics` and `FallbackIntrinsics128` by normalizing the `HwIntrinsics` implementation adding `Vector512` support and extending `Vector128` support to include ARM.

Any and all input is welcome. 

CC @tannergooding @saucecontrol @gfoidl @br3aker


src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -38,17 +43,18 @@ public static class HwIntrinsics
...
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static Vector256<uint> PermuteMaskShiftAlpha8x32() => Vector256.Create(
-            0, 0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 4, 0, 0, 0,
-            5, 0, 0, 0, 6, 0, 0, 0, 3, 0, 0, 0, 7, 0, 0, 0).AsUInt32();
-#pragma warning restore SA1003, SA1116, SA1117 // Parameters should be on same line or separate lines
+        public static Vector256<uint> PermuteMaskShiftAlpha8x32()
+            => Vector256.Create(
+                0, 0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 4, 0, 0, 0,
+                5, 0, 0, 0, 6, 0, 0, 0, 3, 0, 0, 0, 7, 0, 0, 0).AsUInt32();
```
saucecontrol, line 55:
```suggestion
        public static Vector256<uint> PermuteMaskShiftAlpha8x32() => Vector256.Create(0u, 1, 2, 4, 5, 6, 3, 7);
```

This is another one that started out as a `ROS<byte>` but can be simplified now.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -881,17 +887,24 @@ public static ushort TrailingZeroCount(ushort value)
...
-            DebugGuard.IsTrue(source.Length == dest.Length, nameof(source), "Input spans must be of same length!");
+            DebugGuard.IsTrue(source.Length == destination.Length, nameof(source), "Input spans must be of same length!");
 
-            if (Avx2.IsSupported || Sse2.IsSupported)
+            if (Avx512BW.IsSupported || Avx2.IsSupported || Sse2.IsSupported || AdvSimd.IsSupported)
             {
                 int remainder;
-                if (Avx2.IsSupported)
+
+                if (Avx512BW.IsSupported)
```
saucecontrol, line 903: This one's probably ok, but you'll want to consider bypassing the AVX-512 path on older hardware like Skylake-X, because of the [well-known downclocking issues](https://lemire.me/blog/2018/09/07/avx-512-when-and-how-to-use-these-new-instructions/).  

`Vector512.IsHardwareAccelerated` will return false on these processors, even though they support the AVX-512 instructions, so you can check both conditions to make sure the code only runs where AVX-512 is supported *and* fast.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -62,13 +62,25 @@ public static bool SupportsShiftByte
...
+        if (AdvSimd.IsSupported)
+        {
+#pragma warning disable CA1857 // A constant is expected for the parameter
+            Vector128<float> result = Vector128.Create(AdvSimd.Extract(vector, (byte)(control & 0x3)));
+            result = AdvSimd.Insert(result, 1, AdvSimd.Extract(vector, (byte)((control >> 2) & 0x3)));
+            result = AdvSimd.Insert(result, 2, AdvSimd.Extract(vector, (byte)((control >> 4) & 0x3)));
+            result = AdvSimd.Insert(result, 3, AdvSimd.Extract(vector, (byte)((control >> 6) & 0x3)));
+#pragma warning restore CA1857 // A constant is expected for the parameter
+            return result;
+        }
```
saucecontrol, line 82: This is going to be a perf killer in most cases.  It'll be worth testing any path that uses this to make sure it doesn't regress to slower than scalar when Arm starts taking a SIMD path it skipped before.

src/ImageSharp/Common/Helpers/Vector256Utilities.cs
```diff
@@ -73,6 +82,34 @@ public static Vector256<byte> Shuffle(Vector256<byte> vector, Vector256<byte> in
...
     }
 
+    /// <summary>
+    /// Performs a conversion from a 256-bit vector of 8 single-precision floating-point values to a 256-bit vector of 8 signed 32-bit integer values.
+    /// Rounding is equivalent to <see cref="MidpointRounding.ToEven"/>.
+    /// </summary>
+    /// <param name="vector">The value to convert.</param>
+    /// <returns>The <see cref="Vector256{Int32}"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector256<int> ConvertToInt32RoundToEven(Vector256<float> vector)
```
saucecontrol, line 92: Personally, I'd avoid helpers like this for now.  As of .NET 8, only AVX hardware supports hardware-accelerated Vector256 at all, so this is just that with more steps.  You may be able to expand that when Arm SVE becomes available in .NET 9 (or 10 if you stick to the current pattern of targeting LTS), but one would hope by that time, you'll have a BCL-provided xplat helper you can use to replace any explicit AVX calls.

That is to say, at the time this becomes actually useful, you'll have to make changes anyway.

Same goes for Vector512, ofc.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -881,17 +887,24 @@ public static ushort TrailingZeroCount(ushort value)
...
         internal static void NormalizedFloatToByteSaturateReduce(
             ref ReadOnlySpan<float> source,
-            ref Span<byte> dest)
+            ref Span<byte> destination)
         {
-            DebugGuard.IsTrue(source.Length == dest.Length, nameof(source), "Input spans must be of same length!");
+            DebugGuard.IsTrue(source.Length == destination.Length, nameof(source), "Input spans must be of same length!");
 
-            if (Avx2.IsSupported || Sse2.IsSupported)
+            if (Avx512BW.IsSupported || Avx2.IsSupported || Sse2.IsSupported || AdvSimd.IsSupported)
```
saucecontrol, line 899:
```suggestion
            if (Sse2.IsSupported || AdvSimd.IsSupported)
```

`Avx512BW.IsSupported` and `Avx2.IsSupported` strictly imply `Sse2.IsSupported`, so this is redundant.

saucecontrol (commented):
I see there's a test failure on Arm64.  Nothing jumps out as a cause, but I left some comments anyway :)

src/ImageSharp/Common/Helpers/Vector512Utilities.cs
```diff
@@ -75,6 +82,34 @@ public static Vector512<byte> Shuffle(Vector512<byte> vector, Vector512<byte> in
...
+
+        if (Avx.IsSupported)
+        {
+            Vector256<int> lower = Avx.ConvertToVector256Int32(vector.GetLower());
+            Vector256<int> upper = Avx.ConvertToVector256Int32(vector.GetUpper());
+            return Vector512.Create(lower, upper);
+        }
+
+        Vector512<float> sign = vector & Vector512.Create(-0.0f);
+        Vector512<float> val_2p23_f32 = sign | Vector512.Create(8388608.0f);
```
gfoidl, line 107: Can you make `8388608.0f` a named constant and use that in the other VectorXyzUtilities too?
So it's defined three times.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: I tossed a coin over this one. My thought was I was making enough savings elsewhere in calling method to justify the cost here. I need to see what the codegen is like, but I have no means to do so currently.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
JimBobSquarePants: Foot meet gun!
I'll add the additional checks.

src/ImageSharp/Common/Helpers/Vector256Utilities.cs
JimBobSquarePants: The hope is that bythe time 10 comes out I can replace `Vector256Utilities.XX` with `Vector256.XX` and it will be a simple job to refactor. Plus, the consistency feeds my soul.

src/ImageSharp/Common/Helpers/Vector512Utilities.cs
JimBobSquarePants: I'll have a look. Not sure where to put it.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
saucecontrol: I assumed codegen would just be the 4 extracts and 3 inserts (which would make the shuffle a minimum of 8 cycles vs the 1 it takes on x86), but it's actually worse.

image

JimBobSquarePants: 🤮I'lll um, i'll remove that then.

JimBobSquarePants: I wish that you could define constants from a value given a parameter is a constant. I bet the codegen would be much better then. 

```diff
- public static Vector128<float> Shuffle(Vector128<float> vector, [ConstantExpected] byte control)
+ public static Vector128<float> Shuffle(Vector128<float> vector, const byte control)
```

Then...

```c#
const byte w = (byte)((control >> 6) & 0x3)
```

saucecontrol: I would have thought the fact that it's inlined and the control byte *is* a constant would have been enough, so I'm quite surprised at that codegen.  I'll have a look in the morning and see if I can figure out what's up.

tannergooding: Looks to be an issue with where the constant folding happens.

The inlininer is happening and the operands seen are effectively:
```
               [000000] ----------- arg0                    +--*  LCL_VAR   simd16<System.Runtime.Intrinsics.Vector128`1> V00 arg0         
               [000006] ----------- arg1                    +--*  CNS_INT   int    0
               [000007] ----------- arg2                    +--*  LCL_VAR   simd16<System.Runtime.Intrinsics.Vector128`1> V00 arg0         
               [000010] ----------- arg3                    \--*  AND       int   
               [000008] -----------                            +--*  CNS_INT   int    228
               [000009] -----------                            \--*  CNS_INT   int    3
```

We're likely missing an attempt to constant fold `AND` which is blocking it, which is unfortunate.

tannergooding: Going to see if its a trivial fix I can get in...

src/ImageSharp/Common/Helpers/Vector512Utilities.cs
tannergooding: Worth noting that `8388608` is `2^23`, which is the boundary value at which fractional data can no longer be represented (only whole integers at or above this boundary).

src/ImageSharp/Common/Helpers/Vector256Utilities.cs
saucecontrol: > the consistency feeds my soul

I feel this 😄

My concern would be that the fallback code here is either
1) Never used, so just a maintenance burden with no payoff, or
2) Used accidentally, resulting in a perf drop over the optimized `Vector128` path

Perhaps if you really do want to hide all intrinsics behind helpers like this, it would be better to throw PNSE outside the one optimized implementation.  Be aware, though, adding these extra layers over intrinsics can hinder inlining when they're used in larger methods.

When refactoring time comes, most of these will hopefully just be `Avx2.XX` -> `Vector256.XX` anyway.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
tannergooding: https://github.com/dotnet/runtime/pull/97901 fixes this case (and similar scenarios) in particular

JimBobSquarePants: Hero!

src/ImageSharp/Common/Helpers/Vector512Utilities.cs
JimBobSquarePants: Never found a place to put this.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -772,126 +784,164 @@ public static ushort TrailingZeroCount(ushort value)
...
         internal static void NormalizedFloatToByteSaturateReduce(
             ref ReadOnlySpan<float> source,
-            ref Span<byte> dest)
+            ref Span<byte> destination)
         {
-            DebugGuard.IsTrue(source.Length == dest.Length, nameof(source), "Input spans must be of same length!");
+            DebugGuard.IsTrue(source.Length == destination.Length, nameof(source), "Input spans must be of same length!");
 
-            if (Avx2.IsSupported || Sse2.IsSupported)
+            if ((Vector512.IsHardwareAccelerated && Avx512BW.IsSupported) ||
```
saucecontrol, line 934: This check won't prevent you running the AVX-512 code on the slow processors, because those will pass the `Vector256.IsHardwareAccelerated && Avx2.IsSupported` on the next line and enter this block, where they'll also pass the `Avx512BW.IsSupported` check below and use `Vector512` anyway.

In fact, this entire tangle of checks passes on the lowest common denominator, which is `Sse2.IsSupported || AdvSimd.IsSupported`.  That would seem to be unnecessarily restrictive though, becuase you have xplat fallback code for the `Vector128` path.  I don't know if `Vector128.IsHardwareAccelerated` reports true for wasm on .NET 8, but if it does now or in .NET 9, you'd be getting that for free if you loosen this check to just `Vector128.IsHardwareAccelerated`.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -749,17 +752,26 @@ public static ushort TrailingZeroCount(ushort value)
...
+            ref Span<float> destination)
         {
-            DebugGuard.IsTrue(source.Length == dest.Length, nameof(source), "Input spans must be of same length!");
+            DebugGuard.IsTrue(source.Length == destination.Length, nameof(source), "Input spans must be of same length!");
 
-            if (Avx2.IsSupported || Sse2.IsSupported)
+            if ((Vector512.IsHardwareAccelerated && Avx512F.IsSupported) ||
+                Avx2.IsSupported ||
+                Sse2.IsSupported ||
+                AdvSimd.IsSupported)
```
saucecontrol, line 767:
```suggestion
            if (Vector128.IsHardwareAccelerated)
```

The check as written won't prevent you running the AVX-512 code on the slow processors, because those will still pass the `Avx2.IsSupported` check and enter this block, where they'll also pass the `Avx512F.IsSupported` check below and use `Vector512` anyway.

This entire expression boils down to the lowest common denominator, which is `Sse2.IsSupported || AdvSimd.IsSupported`.  That would seem to be unnecessarily restrictive though, because your `Vector128` path is 100% xplat intrinsics.  I don't know if `Vector128.IsHardwareAccelerated` reports true for wasm on .NET 8, but if it does now or in .NET 9, you'd be getting that for free if you loosen this check to just `Vector128.IsHardwareAccelerated`.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -749,17 +752,26 @@ public static ushort TrailingZeroCount(ushort value)
...
             {
                 int remainder;
-                if (Avx2.IsSupported)
+                if (Avx512F.IsSupported)
```
saucecontrol, line 770:
```suggestion
                if (Vector512.IsHardwareAccelerated && Avx512F.IsSupported)
```

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -772,126 +784,164 @@ public static ushort TrailingZeroCount(ushort value)
...
         /// Implementation is based on MagicScaler code:
         /// https://github.com/saucecontrol/PhotoSauce/blob/b5811908041200488aa18fdfd17df5fc457415dc/src/MagicScaler/Magic/Processors/ConvertersFloat.cs#L80-L182
         /// </remarks>
         internal static unsafe void ByteToNormalizedFloat(
             ReadOnlySpan<byte> source,
-            Span<float> dest)
+            Span<float> destination)
         {
-            fixed (byte* sourceBase = source)
+            if (Avx512F.IsSupported)
```
saucecontrol, line 808:
```suggestion
            if (Vector512.IsHardwareAccelerated && Avx512F.IsSupported)
```

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -772,126 +784,164 @@ public static ushort TrailingZeroCount(ushort value)
...
+
+                    ref Vector256<float> d = ref Unsafe.Add(ref destinationBase, i * 4);
+
+                    d = f0;
+                    Unsafe.Add(ref d, 1) = f1;
+                    Unsafe.Add(ref d, 2) = f2;
+                    Unsafe.Add(ref d, 3) = f3;
+                }
+            }
+            else if (Sse2.IsSupported || AdvSimd.IsSupported)
```
saucecontrol, line 873:
```suggestion
            else
```

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -772,126 +784,164 @@ public static ushort TrailingZeroCount(ushort value)
...
-            ref Span<byte> dest)
+            ref Span<byte> destination)
         {
-            DebugGuard.IsTrue(source.Length == dest.Length, nameof(source), "Input spans must be of same length!");
+            DebugGuard.IsTrue(source.Length == destination.Length, nameof(source), "Input spans must be of same length!");
 
-            if (Avx2.IsSupported || Sse2.IsSupported)
+            if ((Vector512.IsHardwareAccelerated && Avx512BW.IsSupported) ||
+                (Vector256.IsHardwareAccelerated && Avx2.IsSupported) ||
+                (Vector128.IsHardwareAccelerated && (Sse2.IsSupported || AdvSimd.IsSupported)))
```
saucecontrol, line 936:
```suggestion
            if (Vector128.IsHardwareAccelerated)
```

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -772,126 +784,164 @@ public static ushort TrailingZeroCount(ushort value)
...
             {
                 int remainder;
-                if (Avx2.IsSupported)
+
+                if (Avx512BW.IsSupported)
```
saucecontrol, line 940:
```suggestion
                if (Vector512.IsHardwareAccelerated && Avx512BW.IsSupported)
```

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -906,36 +956,70 @@ public static ushort TrailingZeroCount(ushort value)
...
         /// Implementation is based on MagicScaler code:
         /// https://github.com/saucecontrol/PhotoSauce/blob/b5811908041200488aa18fdfd17df5fc457415dc/src/MagicScaler/Magic/Processors/ConvertersFloat.cs#L541-L622
         /// </remarks>
         internal static void NormalizedFloatToByteSaturate(
             ReadOnlySpan<float> source,
-            Span<byte> dest)
+            Span<byte> destination)
         {
-            if (Avx2.IsSupported)
+            if (Avx512BW.IsSupported)
```
saucecontrol, line 980:
```suggestion
            if (Vector512.IsHardwareAccelerated && Avx512BW.IsSupported)
```

saucecontrol (approved):
Looks good aside from the tangle of `IsSupported` checks 👍

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
saucecontrol: BTW, I'm not 100% sure you want to include the `Vector512.IsHardwareAccelerated` restriction on all of these.  You do have the option of restricting on `IsSupported` only and telling users who don't want to use AVX-512 on the slower hardware to disable it with `DOTNET_EnableAVX512F=0`.  I haven't made this decision for my own libs yet.

src/ImageSharp/Common/Helpers/SimdUtils.Convert.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static void ConvertByteToNormalizedFloatRemainder(ReadOnlySpan<byte> source, Span<float> destination)
+    {
+        ref byte sBase = ref MemoryMarshal.GetReference(source);
+        ref float dBase = ref MemoryMarshal.GetReference(destination);
+
+        for (int i = 0; i < source.Length; i++)
+        {
+            Unsafe.Add(ref dBase, i) = Unsafe.Add(ref sBase, i) / 255f;
```
gfoidl, line 60:
```suggestion
            Unsafe.Add(ref dBase, (uint)i) = Unsafe.Add(ref sBase, (uint)i) / 255f;
```

src/ImageSharp/Common/Helpers/SimdUtils.Convert.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static void ConvertNormalizedFloatToByteRemainder(ReadOnlySpan<float> source, Span<byte> destination)
+    {
+        ref float sBase = ref MemoryMarshal.GetReference(source);
+        ref byte dBase = ref MemoryMarshal.GetReference(destination);
+        for (int i = 0; i < source.Length; i++)
```
gfoidl, line 69:
```suggestion
        ref byte dBase = ref MemoryMarshal.GetReference(destination);
        
        for (int i = 0; i < source.Length; i++)
```
For code style to have the same blank line as above.

src/ImageSharp/Common/Helpers/SimdUtils.Convert.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            Unsafe.Add(ref dBase, i) = ConvertToByte(Unsafe.Add(ref sBase, i));
```
gfoidl, line 71:
```suggestion
            Unsafe.Add(ref dBase, (uint)i) = ConvertToByte(Unsafe.Add(ref sBase, (uint)i));
```

gfoidl (approved):
Modulo some minor things and the feedback from @saucecontrol: LGTM

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
JimBobSquarePants: Is there one I can add? Would be great to do so.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
JimBobSquarePants: I think I'll stick and include the check. I want my users to get no surprises.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
JimBobSquarePants: Ah yeah, this is much smarter, nice to know about wasm too.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
saucecontrol: There's `Vector128.Narrow` (truncating), plus `Min` and `Max`, so you could put something together.  It's a bit awkward because you really want to go int->byte, and splitting the helpers to plug in for each of the narrowing steps separately would be a lot of extra codegen.  A single clamp to byte range followed by double narrow would be slightly cleaner.

JimBobSquarePants:
Thanks for the review chaps. I've actioned the lot now. Will merge on successful build.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
JimBobSquarePants: Thanks. I think I'll leave it for now. Don't want to spend too long figuring it out.

JimBobSquarePants:
Stating to see some improvement in numbers here. 

```
BenchmarkDotNet v0.13.11, Windows 11 (10.0.22631.3085/23H2/2023Update/SunValley3)
.NET SDK 8.0.101
  [Host]   : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  ShortRun : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI

WarmupCount=5
```

| Method                              | Mean      | Error     | StdDev   | Ratio | Gen0      | Gen1      | Gen2      | Allocated  | Alloc Ratio |
|------------------------------------ |----------:|----------:|---------:|------:|----------:|----------:|----------:|-----------:|------------:|
| 'System.Drawing Load, Resize, Save' | 368.56 ms | 12.636 ms | 1.955 ms |  1.00 |         - |         - |         - |   13.23 KB |        1.00 |
| 'ImageFlow Load, Resize, Save'      | 270.28 ms | 12.142 ms | 3.153 ms |  0.74 |  500.0000 |  500.0000 |  500.0000 | 4642.53 KB |      351.00 |
| 'ImageSharp Load, Resize, Save'     | 114.87 ms |  2.408 ms | 0.625 ms |  0.31 |         - |         - |         - | 1314.02 KB |       99.35 |
| 'ImageSharp TD Load, Resize, Save'  |  75.31 ms |  1.923 ms | 0.499 ms |  0.20 |  166.6667 |         - |         - | 1307.22 KB |       98.83 |
| 'ImageMagick Load, Resize, Save'    | 393.97 ms |  7.115 ms | 1.101 ms |  1.07 |         - |         - |         - |   54.48 KB |        4.12 |
| 'ImageFree Load, Resize, Save'      | 235.16 ms |  2.376 ms | 0.617 ms |  0.64 | 6000.0000 | 6000.0000 | 6000.0000 |    95.8 KB |        7.24 |
| 'MagicScaler Load, Resize, Save'    |  66.67 ms |  0.873 ms | 0.135 ms |  0.18 |         - |         - |         - |   45.42 KB |        3.43 |
| 'SkiaSharp Load, Resize, Save'      | 136.42 ms |  2.833 ms | 0.736 ms |  0.37 |         - |         - |         - |   88.15 KB |        6.66 |
| 'NetVips Load, Resize, Save'        | 122.52 ms |  7.601 ms | 1.974 ms |  0.33 |         - |         - |         - |   50.95 KB |        3.85 |

PR #2706, opened 2024-03-28
Limit all memory allocations in the MemoryAllocator layer

### LINKED ISSUE #2696: File (detected as bmp) fails to open, consumes a lot of memory
### Prerequisites

### ImageSharp version

3.1.3

N/A

Windows 11

### .NET Framework version

.net 8

### Description

The attached (zipped) file, 126 bytes long, gives an out-of-memory error when I try to open it with the following code:

```
using FileStream stream = File.OpenRead("00271-imageworsener-leftshift");
SixLabors.ImageSharp.Image.Load(stream);
```

I believe there is a problem with the image dimensions - BmpDecoderCore.ReadImageHeaders sets this.infoHeader.Width and this.infoHeader.Height to very high values which tries to create a huge image in BmpDecoderCore.Decode().

### Steps to Reproduce

See description.

### Images

[00271-imageworsener-leftshift.zip](https://github.com/SixLabors/ImageSharp/files/14583779/00271-imageworsener-leftshift.zip)

antonfirsov:
### Prerequisites

### Description

Fixes #2696.

This is an alternative take for #2704. Differences:
- The discontiguous limit is configurable in Megabytes on `MemoryAllocatorOptions` (together with the pool size).
    - On 64bit it defaults to 4GB, equivalent of `32k x 32k` of `Rgba32` which is more conservative than the limit in #2704. For me this feels somewhat safer while I don't see many downsides. Note that this is the absolute hardcoded limit in WIC, while newer [skia builds have a hardcoded 2GB limit](https://github.com/google/skia/commit/e262e688247792e9a7e31355a3e00315e606c0b3) (this hasn't been merged to in SkiaSharp 2.* yet)
    - On 32bit it defaults to 1GB which is a significant portion of the overall virtual memory limit of 4GB. Maybe we can make it even lower.
- Single contiguous buffers have a limit of 1GB (or the discontiguous limit, if it's lower). This limit is not configurable separately in order to make things simpler for users. I don't see any risks from that nor the need to make it as low as in #2704. The RLE path in BMP and TIFF decoders actually allocates temporary contiguous buffers of `width * height` sizes so a 64MB limit would be easy to exceed with sane images.
- It also significantly reduces `MaxAllocationAttempts` so we don't block too long if an OOM is still being hit.

The PR is missing tests and PNG tweak from #2704. Will finalize it after agreeing on the direction.

### DIFF
--- a/src/ImageSharp/Memory/Allocators/MemoryAllocator.cs
+++ b/src/ImageSharp/Memory/Allocators/MemoryAllocator.cs
@@ -2,6 +2,8 @@
 // Licensed under the Six Labors Split License.
 
 using System.Buffers;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
 
 namespace SixLabors.ImageSharp.Memory;
 
@@ -10,6 +12,8 @@ namespace SixLabors.ImageSharp.Memory;
 /// </summary>
 public abstract class MemoryAllocator
 {
+    private const int OneGigabyte = 1 << 30;
+
     /// <summary>
     /// Gets the default platform-specific global <see cref="MemoryAllocator"/> instance that
     /// serves as the default value for <see cref="Configuration.MemoryAllocator"/>.
@@ -19,6 +23,12 @@ public abstract class MemoryAllocator
     /// to change the default allocator used by <see cref="Image"/> and it's operations.
     /// </summary>
     public static MemoryAllocator Default { get; } = Create();
+
+    internal long MemoryGroupAllocationLimitBytes { get; private set; } = Environment.Is64BitProcess ?
+        4L * OneGigabyte :
+        OneGigabyte;
+
+    internal int SingleBufferAllocationLimitBytes { get; private set; } = OneGigabyte;
 
     /// <summary>
     /// Gets the length of the largest contiguous buffer that can be handled by this allocator instance in bytes.
@@ -30,16 +40,24 @@ public abstract class MemoryAllocator
     /// Creates a default instance of a <see cref="MemoryAllocator"/> optimized for the executing platform.
     /// </summary>
     /// <returns>The <see cref="MemoryAllocator"/>.</returns>
-    public static MemoryAllocator Create() =>
-        new UniformUnmanagedMemoryPoolMemoryAllocator(null);
+    public static MemoryAllocator Create() => Create(default);
 
     /// <summary>
     /// Creates the default <see cref="MemoryAllocator"/> using the provided options.
     /// </summary>
     /// <param name="options">The <see cref="MemoryAllocatorOptions"/>.</param>
     /// <returns>The <see cref="MemoryAllocator"/>.</returns>
-    public static MemoryAllocator Create(MemoryAllocatorOptions options) =>
-        new UniformUnmanagedMemoryPoolMemoryAllocator(options.MaximumPoolSizeMegabytes);
+    public static MemoryAllocator Create(MemoryAllocatorOptions options)
+    {
+        UniformUnmanagedMemoryPoolMemoryAllocator allocator = new(options.MaximumPoolSizeMegabytes);
+        if (options.AllocationLimitMegabytes.HasValue)
+        {
+            allocator.MemoryGroupAllocationLimitBytes = options.AllocationLimitMegabytes.Value * 1024 * 1024;
+            allocator.SingleBufferAllocationLimitBytes = (int)Math.Min(allocator.SingleBufferAllocationLimitBytes, allocator.MemoryGroupAllocationLimitBytes);
+        }
+
+        return allocator;
+    }
 
     /// <summary>
     /// Allocates an <see cref="IMemoryOwner{T}" />, holding a <see cref="Memory{T}"/> of length <paramref name="length"/>.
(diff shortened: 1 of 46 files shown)


src/ImageSharp/Memory/Allocators/MemoryAllocator.cs
```diff
@@ -20,6 +24,12 @@ public abstract class MemoryAllocator
     /// </summary>
     public static MemoryAllocator Default { get; } = Create();
 
+    internal long MemoryGroupAllocationLimitBytes { get; private set; } = Environment.Is64BitProcess ?
+        4L * OneGigabyte :
+        OneGigabyte;
+
+    internal int SingleBufferAllocationLimitBytes { get; private set; } = OneGigabyte;
```
antonfirsov, line 31: Actually, there is no good reason to set this to 1GB. Contiguous buffers have a natural limit of `int.MaxValue ~ 2GB` so I think we should set the contiguous limit to `min(int.MaxValue, MemoryAllocatorOptions.AllocationLimitMegabytes * 1MB)`. That would get rid of all obscure logic and make the semantics of `MemoryAllocatorOptions.AllocationLimitMegabytes` very obvious.

JimBobSquarePants: Yeah. I like that. The only risk we have with this stuff is in our decoders which can be easily mitigated with some care (PNG tweaks)

src/ImageSharp/Memory/Allocators/MemoryAllocatorOptions.cs
```diff
@@ -27,4 +28,22 @@ public struct MemoryAllocatorOptions
             this.maximumPoolSizeMegabytes = value;
         }
     }
+
+    /// <summary>
+    /// Gets or sets a value defining the maximum (discontiguous) buffer size that can be allocated by the allocator in Megabytes.
+    /// <see langword="null"/> means platform default: 1GB on 32-bit processes, 4GB on 64-bit processes.
+    /// </summary>
+    public int? AllocationLimitMegabytes
```
JimBobSquarePants, line 36: You got it in the options! Nice!

src/ImageSharp/Memory/Allocators/MemoryAllocator.cs
```diff
@@ -30,16 +40,24 @@ public abstract class MemoryAllocator
...
     /// Creates the default <see cref="MemoryAllocator"/> using the provided options.
     /// </summary>
     /// <param name="options">The <see cref="MemoryAllocatorOptions"/>.</param>
     /// <returns>The <see cref="MemoryAllocator"/>.</returns>
-    public static MemoryAllocator Create(MemoryAllocatorOptions options) =>
-        new UniformUnmanagedMemoryPoolMemoryAllocator(options.MaximumPoolSizeMegabytes);
+    public static MemoryAllocator Create(MemoryAllocatorOptions options)
+    {
+        UniformUnmanagedMemoryPoolMemoryAllocator allocator = new(options.MaximumPoolSizeMegabytes);
+        if (options.AllocationLimitMegabytes.HasValue)
```
JimBobSquarePants, line 53: For V4 I think we should introduce `UniformUnmanagedMemoryPoolOptions` which has the standard options as a property.

src/ImageSharp/Memory/InvalidMemoryOperationException.cs
```diff
@@ -24,4 +26,8 @@ public InvalidMemoryOperationException(string message)
     public InvalidMemoryOperationException()
     {
     }
+
+    [DoesNotReturn]
+    internal static void ThrowAllocationOverLimitException(long length, long limit) =>
```
JimBobSquarePants, line 31: I like this sitting here.

src/ImageSharp/Memory/Allocators/SimpleGcMemoryAllocator.cs
```diff
@@ -19,6 +20,13 @@ public override IMemoryOwner<T> Allocate<T>(int length, AllocationOptions option
     {
         Guard.MustBeGreaterThanOrEqualTo(length, 0, nameof(length));
 
+        int lengthInBytes = length * Unsafe.SizeOf<T>();
```
JimBobSquarePants, line 23: For V4 I'd like to have this in `MemoryAllocator` so we don't have to repeat the check.

src/ImageSharp/Memory/Allocators/MemoryAllocator.cs
```diff
@@ -69,10 +87,31 @@ public virtual void ReleaseRetainedResources()
...
     /// <exception cref="InvalidMemoryOperationException">Thrown when 'blockAlignment' converted to bytes is greater than the buffer capacity of the allocator.</exception>
-    internal virtual MemoryGroup<T> AllocateGroup<T>(
+    internal MemoryGroup<T> AllocateGroup<T>(
         long totalLength,
         int bufferAlignment,
         AllocationOptions options = AllocationOptions.None)
         where T : struct
-        => MemoryGroup<T>.Allocate(this, totalLength, bufferAlignment, options);
+    {
+        long totalLengthInBytes = totalLength * Unsafe.SizeOf<T>();
```
JimBobSquarePants, line 96: I think it might be possible to overflow this. I think one of our tests that allocates a 256 byte struct does so.

JimBobSquarePants (commented):
That's much neater than my version. I say go with this. 👍

src/ImageSharp/Memory/Allocators/MemoryAllocatorOptions.cs
tocsoft, line 36: Is this consistent with how we talk about memory across the public api? do we talk about megabytes anywhere else in the public api or is it all bytes?

we want to make sure we are being consistent on our units of measure in the public api surface.

antonfirsov: The other property on `MemoryAllocatiorOptions` is `MaximumPoolSizeMegabytes`. When I designed the [MemoryAllocator configuration API](https://docs.sixlabors.com/articles/imagesharp/memorymanagement.html#configuring-the-pool-size) it seemed like Megabytes are easier to reason about when talking about limits.

src/ImageSharp/Memory/Allocators/MemoryAllocator.cs
antonfirsov: Actually, `AllocatorOptions` is specific to `UniformUnmanagedMemoryAllocator`. I don't think anyone ever would need another allocator IRL. The design keeps it pluggable just in case + `SimpleGcMemoryAllocator` can be useful for diagnostic purposes since managed allocations are easy to track with standard .NET tools. But `SimpleGcMemoryAllocator` should not be used in production.

To secure users who still use it (mostly legacy code designed to workaround 1.* issues), I applied the default limits, but you cannot configure `SimpleGcMemoryAllocator` with `MemoryAllocatorOptions`.

JimBobSquarePants (approved):
I'm happy with this @antonfirsov and have created a backport PR for V2.

 I'll wait for your final read-through to merge though.

tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
```diff
@@ -558,4 +558,16 @@ public void BmpDecoder_CanDecode_Os2BitmapArray<TPixel>(TestImageProvider<TPixel
...
+    [Theory]
+    [WithFile(Issue2696, PixelTypes.Rgba32)]
+    public void BmpDecoder_ThrowsException_Issue2696<TPixel>(TestImageProvider<TPixel> provider)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        InvalidImageContentException ex = Assert.Throws<InvalidImageContentException>(() =>
+            {
+                using Image<TPixel> image = provider.GetImage(BmpDecoder.Instance);
+            });
+        Assert.IsType<InvalidMemoryOperationException>(ex.InnerException);
```
antonfirsov, line 571: This made me find that the file wasn't the actual file from #2696 here, leading to an exception in header parsing instead of an allocation failure. Fixed in [d1cc651](https://github.com/SixLabors/ImageSharp/pull/2706/commits/d1cc651e19def81df66779985a50f9372db332bc).

Issue #2708, opened 2024-03-28
APNG Decoder incorrectly handles frame offsets and dispose previous with blend over


SpaceCheetah:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

3.1.3

### Other ImageSharp packages and versions

None

### Environment (Operating system, version and so on)

Windows 11

### .NET Framework version

.net8.0

### Description

For APNG, the decoder has various significant issues handling certain cases.
For instance, in APNG, for compression, frames are allowed to be smaller than the overall image with an offset. The decoder currently does not take that offset into account.
Additionally, there seem to be some issues with dispose previous, blend over.
Both of these issues are represented in the test image below.

### Steps to Reproduce

```
using SixLabors.ImageSharp;

Image image = Image.Load("test.png");
for (int i = 0; i < image.Frames.Count; i++) {
    image.Frames.CloneFrame(i).SaveAsPng($"{i}.png");
}
```

### Images

Original:
![test](https://github.com/SixLabors/ImageSharp/assets/30188201/dda89586-c244-4eec-addd-8d6463002d46)
Outputted frames:
![0](https://github.com/SixLabors/ImageSharp/assets/30188201/f1a9aadb-d56b-4cf6-85b1-81f6b7d9fb4e)
![1](https://github.com/SixLabors/ImageSharp/assets/30188201/f93a4222-6406-47ac-a608-dc497071723f)
![2](https://github.com/SixLabors/ImageSharp/assets/30188201/afe8dcd3-4f26-4a34-a85d-0e7e70117563)
![3](https://github.com/SixLabors/ImageSharp/assets/30188201/b315eeea-e766-416d-94bd-27d77cc97fa2)
![4](https://github.com/SixLabors/ImageSharp/assets/30188201/c33dc27b-710d-41aa-ae22-896da94a000c)

SpaceCheetah:
I'll be looking into fixing the decoder (and probably encoder) soon, but first.

Should each of the frames from the decoder really be the end result, rendered frame? If so, giving any frame, in many cases, requires rendering all previous frames.

[...]

PR #2710, opened 2024-03-29
Fix animated png handling (issue #2708)

SpaceCheetah:
### Prerequisites

### Description

Identified and fixed the issues mentioned in issue #2708.
Fix: ProcessInterlacedPaletteScanline now respects frameControl.XOffset
Fix: PngDecoderCore handling of PngDisposalMethod.RestoreToBackground and RestoreToPrevious, for specification compliance.
Metadata change: Added DefaultImageAnimated to PngMetadata. The apng format allows for the first frame to not be part of the animation sequence, which needs to be visible to users.
Fix: Decoding and encoding when the default image is not animated

### DIFF
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -254,7 +254,7 @@ internal sealed class PngDecoderCore : IImageDecoderInternals
 
                             break;
                         case PngChunkType.Data:
-                            pngMetadata.DefaultImageAnimated = currentFrameControl != null;
+                            pngMetadata.AnimateRootFrame = currentFrameControl != null;
                             currentFrameControl ??= new((uint)this.header.Width, (uint)this.header.Height);
                             if (image is null)
                             {
@@ -271,7 +271,7 @@ internal sealed class PngDecoderCore : IImageDecoderInternals
                                 this.ReadNextDataChunk,
                                 currentFrameControl.Value,
                                 cancellationToken);
-                            if (pngMetadata.DefaultImageAnimated)
+                            if (pngMetadata.AnimateRootFrame)
                             {
                                 previousFrame = currentFrame;
                                 previousFrameControl = currentFrameControl;
--- a/src/ImageSharp/Formats/Png/PngDecoderCore.cs
+++ b/src/ImageSharp/Formats/Png/PngDecoderCore.cs
@@ -234,8 +234,7 @@ internal sealed class PngDecoderCore : IImageDecoderInternals
                                 PngThrowHelper.ThrowMissingFrameControl();
                             }
 
-                            previousFrameControl ??= new((uint)this.header.Width, (uint)this.header.Height);
-                            this.InitializeFrame(previousFrameControl.Value, currentFrameControl.Value, image, previousFrame, out currentFrame);
+                            this.InitializeFrame(previousFrameControl, currentFrameControl.Value, image, previousFrame, out currentFrame);
 
                             this.currentStream.Position += 4;
                             this.ReadScanlines(
@@ -255,7 +254,7 @@ internal sealed class PngDecoderCore : IImageDecoderInternals
 
                             break;
                         case PngChunkType.Data:
-
+                            pngMetadata.DefaultImageAnimated = currentFrameControl != null;
                             currentFrameControl ??= new((uint)this.header.Width, (uint)this.header.Height);
                             if (image is null)
                             {
@@ -272,9 +271,12 @@ internal sealed class PngDecoderCore : IImageDecoderInternals
                                 this.ReadNextDataChunk,
                                 currentFrameControl.Value,
                                 cancellationToken);
+                            if (pngMetadata.DefaultImageAnimated)
+                            {
+                                previousFrame = currentFrame;
+                                previousFrameControl = currentFrameControl;
+                            }
 
-                            previousFrame = currentFrame;
-                            previousFrameControl = currentFrameControl;
                             break;
                         case PngChunkType.Palette:
                             this.palette = chunk.Data.GetSpan().ToArray();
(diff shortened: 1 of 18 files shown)


JimBobSquarePants:
Note. We'll want to backport this to target the release/v3.1.x branch also.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -638,25 +645,25 @@ private void InitializeImage<TPixel>(ImageMetadata metadata, FrameControl frameC
...
         ImageFrame<TPixel>? previousFrame,
         out ImageFrame<TPixel> frame)
         where TPixel : unmanaged, IPixel<TPixel>
     {
-        // We create a clone of the previous frame and add it.
-        // We will overpaint the difference of pixels on the current frame to create a complete image.
-        // This ensures that we have enough pixel data to process without distortion. #2450
         frame = image.Frames.AddFrame(previousFrame ?? image.Frames.RootFrame);
 
-        // If the first `fcTL` chunk uses a `dispose_op` of APNG_DISPOSE_OP_PREVIOUS it should be treated as APNG_DISPOSE_OP_BACKGROUND.
```
JimBobSquarePants, line 653: This comment and above are important.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
```diff
@@ -238,22 +260,17 @@ public void Encode<TPixel>(Image<TPixel> image, Stream stream, CancellationToken
                 }
 
                 // Each frame control sequence number must be incremented by the number of frame data chunks that follow.
-                frameControl = this.WriteFrameControlChunk(stream, frameMetadata, bounds, (uint)i + increment);
```
JimBobSquarePants, line 241: There's a TODO in `WriteFrameControlChunk` where we need to set the X/Y offset. I'm not sure how your tests can pass without it being implemented.

SpaceCheetah: I'm not seeing this TODO? Neither of the two TODOs in PngEncoderCore have anything to do with that, as far as I can tell. Where is this? WriteFrameControlChunk does write x/y offsets.

I didn't have a test for it, but I just added one locally and it made a visually correct image, and looking at the metadata in TweakPNG it did use frame offsets. Not exactly the same ones as the original, but that's just an encoding difference.
It also set off ImageComparer.Exact, with the largest frame difference at 0.0024%. I think this is inherent to the current process of decoding and encoding, since compositing happens during decoding.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -246,11 +245,16 @@ public Image<TPixel> Decode<TPixel>(BufferedReadStream stream, CancellationToken
                                 currentFrameControl.Value,
                                 cancellationToken);
 
-                            previousFrame = currentFrame;
-                            previousFrameControl = currentFrameControl;
+                            // if current frame dispose is restore to previous, then from future frame's perspective, it never happened
+                            if (currentFrameControl.Value.DisposeOperation != PngDisposalMethod.RestoreToPrevious)
```
JimBobSquarePants, line 249: Are you sure about this? We need to ensure we create the entire frame by drawing the current frame over the previous.

Can you please check this against the gif decoder?

src/ImageSharp/Formats/Png/PngEncoderCore.cs
JimBobSquarePants: Ah yes, it's fixed in the main branch but will need to be fixed in the v3 branch.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
SpaceCheetah: This is what the GIF decoder does too. When the disposal method is RestoreToPrevious, it sets previousFrame to the value it already had (GifDecoderCore lines 522 and 621). Otherwise, it sets previousFrame to the current frame, updating it for the next one (line 625).

src/ImageSharp/Formats/Png/PngEncoderCore.cs
SpaceCheetah: Looks like it was already fixed in release/3.1.x, by commit 7890008

SpaceCheetah:
This can be backported to release/3.1.x with a simple cherry-pick. I can make a PR for that if you want.

src/ImageSharp/Formats/Png/PngEncoderCore.cs
JimBobSquarePants: Yeah. Sorry, I was looking at a separate old branch.

JimBobSquarePants:
> This can be backported to release/3.1.x with a simple cherry-pick. I can make a PR for that if you want.

That would be amazing if you can!

src/ImageSharp/Formats/Png/PngMetadata.cs
```diff
@@ -83,6 +84,11 @@ private PngMetadata(PngMetadata other)
     /// </summary>
     public uint RepeatCount { get; set; } = 1;
 
+    /// <summary>
+    ///  Gets or sets a value indicating whether the default image is shown as part of the animated sequence
+    /// </summary>
+    public bool DefaultImageAnimated { get; set; } = true;
```
JimBobSquarePants, line 90: Let's call this `AnimateRootFrame`. That matches `ImageFrameCollection.RootFrame` naming convention.

JimBobSquarePants (commented):
This looks great. Thanks for your perseverance. 

If we can change the property name, then I will approve this for merging. 

Thanks!

SpaceCheetah:
Realized that while this fixes the decoder, the encoder has the same issue with not respecting disposal method.
It would be best to overhaul the encoder to choose the best blend method and disposal method.

JimBobSquarePants:
Can you please check the code at https://github.com/SixLabors/ImageSharp/pull/2882

As far as I’m aware we’ve covered anything missing there.

SpaceCheetah:
Checked that and it doesn't fix it, I'll try to get a minimal reproduction. Would you prefer a separate issue?

JimBobSquarePants:
I’d like to get any fixes incorporated into that PR. Could you please add a comment there with the relevant detail. Thanks!

PR #2739, opened 2024-05-20
Simplify Color Space Conversion APIs

JimBobSquarePants:
### Prerequisites

### Description

Note: I don't expect anyone to thoroughly review every file. It's far too big. 

See https://github.com/SixLabors/ImageSharp/discussions/2531 for the original details. I've improved on the original design there by reducing the generic parameters required and removing the requirement for marker interfaces. 

Replaces `ColorSpaceConverter` with `ColorProfileConverter`. The new generic API brings several advantages.

- Simplifies and normalizes usage across single and multiple input APIs.
- Makes the API extensible (No more N*) implementations.
- Fixes several conversion accuracy issues.
- Improves performance, especially around companding. 

There's potential here now also for reusing the API to allow for converting between ICC profiles. We can use the information to create dynamically calculated working space inputs.

The new API is faster than the old implementation and alternative libraries.

```
BenchmarkDotNet v0.13.10, Windows 11 (10.0.22631.3593/23H2/2023Update/SunValley3)
.NET SDK 8.0.300-preview.24203.14
  [Host]     : .NET 8.0.4 (8.0.424.16909), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.4 (8.0.424.16909), X64 RyuJIT AVX2
```

CieXyz => CieLab

| Method                   | Mean     | Error    | StdDev   | Ratio | RatioSD |
|------------------------- |---------:|---------:|---------:|------:|--------:|
| 'Colourful Convert'      | 50.85 ns | 0.720 ns | 0.638 ns |  1.00 |    0.00 |
| 'ImageSharp Convert Old' | 64.99 ns | 0.875 ns | 0.819 ns |  1.28 |    0.02 |
| 'ImageSharp Convert'     | 43.62 ns | 0.864 ns | 1.061 ns |  0.86 |    0.02 |

CieXyz => HunterLab

| Method               | Mean     | Error    | StdDev   | Ratio | RatioSD |
|--------------------- |---------:|---------:|---------:|------:|--------:|
| 'Colourful Convert'  | 25.30 ns | 0.517 ns | 1.335 ns |  1.00 |    0.00 |
| 'ImageSharp Convert' | 13.51 ns | 0.244 ns | 0.217 ns |  0.57 |    0.03 |

CieXyz => Lms

| Method               | Mean      | Error     | StdDev    | Median    | Ratio | RatioSD |
|--------------------- |----------:|----------:|----------:|----------:|------:|--------:|
| 'Colourful Convert'  | 28.653 ns | 0.6023 ns | 1.6487 ns | 27.979 ns |  1.00 |    0.00 |
| 'ImageSharp Convert' |  8.215 ns | 0.1884 ns | 0.2821 ns |  8.123 ns |  0.29 |    0.02 |

Rgb Chromatic Adaption

| Method             | Mean     | Error   | StdDev   | Ratio | RatioSD |
|------------------- |---------:|--------:|---------:|------:|--------:|
| 'Colourful Adapt'  | 347.4 ns | 6.95 ns | 12.70 ns |  1.00 |    0.00 |
| 'ImageSharp Adapt' | 261.8 ns | 3.35 ns |  2.97 ns |  0.77 |    0.03 |


src/ImageSharp/ColorProfiles/KnownIlluminants.cs
```diff
@@ -12,60 +12,60 @@ namespace SixLabors.ImageSharp.ColorSpaces;
...
 /// Descriptions taken from: http://en.wikipedia.org/wiki/Standard_illuminant
 /// </remarks>
-public static class Illuminants
+public static class KnownIlluminants
 {
     /// <summary>
     /// Incandescent / Tungsten
     /// </summary>
-    public static readonly CieXyz A = new CieXyz(1.09850F, 1F, 0.35585F);
+    public static readonly CieXyz A = new(1.09850F, 1F, 0.35585F);
```
gfoidl, line 20: Maybe it's better to have this as properties like
```suggestion
    public static CieXyz A => new(1.09850F, 1F, 0.35585F);
```
because (I didn't measure it here, so only theory / feeling):
* with the static readonly it's a memory load (potentially not from cache, but from RAM, so hard to measure in micro-benchmarks)
* with the property it's a simple assignment of three floats
* the JIT sees what's going on, so can potentially optimize the use of the three floats (which isn't that easy w/ loading the struct from memory)

Note: that would be a change in the public API.

src/ImageSharp/ColorProfiles/CieLab.cs
```diff
@@ -0,0 +1,182 @@
...
+/// Represents a CIE L*a*b* 1976 color.
+/// <see href="https://en.wikipedia.org/wiki/Lab_color_space"/>
+/// </summary>
+public readonly struct CieLab : IProfileConnectingSpace<CieLab, CieXyz>
+{
+    /// <summary>
+    /// D50 standard illuminant.
+    /// Used when reference white is not specified explicitly.
+    /// </summary>
+    public static readonly CieXyz DefaultWhitePoint = KnownIlluminants.D50;
```
gfoidl, line 19: See the other comment about property / field. I won't comment on these later on.

src/ImageSharp/ColorProfiles/CieLab.cs
```diff
@@ -0,0 +1,182 @@
...
+        this.L = vector.X;
+        this.A = vector.Y;
+        this.B = vector.Z;
+    }
+
+    /// <summary>
+    /// Gets the lightness dimension.
+    /// <remarks>A value usually ranging between 0 (black), 100 (diffuse white) or higher (specular white).</remarks>
+    /// </summary>
+    public readonly float L { get; }
```
gfoidl, line 53: Here `readonly` isn't strictly needed, as the whole struct is readonly.

src/ImageSharp/ColorProfiles/CieLab.cs
```diff
@@ -0,0 +1,182 @@
...
+    /// <inheritdoc/>
+    public override bool Equals(object? obj) => obj is CieLab other && this.Equals(other);
+
+    /// <inheritdoc/>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public bool Equals(CieLab other) =>
+        this.L.Equals(other.L)
+        && this.A.Equals(other.A)
+        && this.B.Equals(other.B);
+
```
gfoidl, line 104:
```suggestion
        new Vector3(this.L, this.A, this.B) == new Vector3(other.L, other.A, other.B);
```
I don't know if this is faster, but at least it's less [machine code](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgDkBXfGKASzFwG4BYAKDMq0ASgwB2GVsxoBhCPgAOrADYsAyiwBu7GFz4DqNEeMkwaASXFtRudrt56AzOVjYAJhFFKAnuVwYoDGAY5NKsMAAy2MB8AN585AmUTgBmShDYweHkMeQA5jAYnOQAvvGJxClpGeQAgtl5BUWlvIlJ5KnpwQBC9fmFJWUJg+QA2gCyBQAWEK5mCkoAFBMY07PzAPLyEh64NDW5ubC4NhowFkqsope5AJQAusMV5MAQEErkAKIAjgzYSrgLUIRKLkCArFg3cgAXgAfMNWitWLtwjRvr9/gswZMWDRwjd4YkAGSE8iI3Y1VE/P4ArE4mr4lqtIkksk0LqU9E08FQNk3HiMoYC0bLVZzeSLEUzMVKTbbax7A5HE5nTyXa73R5OF5vT5U/5UQFhSLAUHcyGwgkJUQwADu5AAajAgtAHAtWeE0KTJki9p7WV1zVDyNa7Y7nVBXbSeR7TdieTVPVHefzikA==).

gfoidl (commented):
No thorough review, just what jumped into my eye while looking at the first files shown in GH.

src/ImageSharp/ColorProfiles/CieLab.cs
JimBobSquarePants: I don't actually need this property. Was missed during the cleanup.

src/ImageSharp/ColorProfiles/CieLab.cs
JimBobSquarePants: Yep. The machine code diff is much larger than expected and, looking at the instructions, SIMDified.

src/ImageSharp/ColorProfiles/KnownIlluminants.cs
JimBobSquarePants: I've changed the pattern to use the auto property with initializer. That's the pattern we've used elsewhere. 

The types are immutable, and I think this approach should be best for more tightly constrained memory environments.

src/ImageSharp/ColorProfiles/Cmyk.cs
```diff
@@ -157,8 +157,5 @@ public override bool Equals(object? obj)
     /// <inheritdoc/>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public bool Equals(Cmyk other)
-        => this.C.Equals(other.C)
-        && this.M.Equals(other.M)
-        && this.Y.Equals(other.Y)
-        && this.K.Equals(other.K);
+        => new Vector4(this.C, this.M, this.Y, this.K) == new Vector4(other.C, other.M, other.Y, other.K);
```
gfoidl, line 160: Another idea: see [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgDkBXfGKASzFwBuGvSasAShwB2GXtxYBhCPgAOvADY8AyjwBu/GEJGNmLSTLkwWASRk8IKnVH1hDw6iIDMDWNgAmEFLqAJ4MuBhQHGAYDAr4wQDWNADeNAzpTN4AZuoQ2DEKDMkMAOYwGIIMAL5pGcTZufkMALJFpeWVNdQZmQw5eTEAmm1lFdW16fV9jTEA0iMd490ZEwwA2s3lABYQftaq6gAUmxg7ewcA8iqygbgsAIIlJbC4uLy6MLbqvFI/JQCUAF1VlNgBAIOoGABRACOHGw6lwhziiQYEFOPH+qx6AF4AHwMKQwADuDAAajBotAUIdTrw7go0Aw6XdmkyWSxBuytvSWLN/gwcTjCSTyZSMNTDuitjxFEzpbK2WiMVBOfKVXz/u4ujQ1loIlEMAAZbDBCAcDCHE1mi2zH5+Fg6OEwCwIoFeHwwfyBEJhA3RWLxBIAfSo1FSy0mDQGsQWYy6PSm/SarWKo06IOjTWGacWCbqWbmcYzkYYqxUfF0+RgYqpUBQDHuuApdYb+IYAFUpLhsFkrE2ADwokPkJktiX1vGHWBZTvd3v93DiGBZQ4/Zk83D/LWrVYbba7fYqI4nM5H9RXG7dh5PF5vD5fH5/d2l0HgyGw+GI5FB0PKmVQFipa4gSHJNuO1KCsKCqquB4rUtqQA==) with the private `AsVector4` property.
This overlays the vector with the struct (is OK for float, as there's no padding in between), so the creation of the vector can be avoided.

That should work for other places (Vector3) too.

JimBobSquarePants: Oh that's cleaver yeah!

JimBobSquarePants:
I think I'm gonna merge this tomorrow.

Issue #2743, opened 2024-05-30
AccessViolationException when loading corrupt GIF


Erik-White:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

3.1.4

### Other ImageSharp packages and versions

N/A

### Environment (Operating system, version and so on)

Windows 11

### .NET Framework version

.NET 8.0.4

### Description

Opening the attached GIF causes an AccessViolatonException, and then a long hang (up to an hour) while the process tries to read GC information.

```
CoreCLR Version: 8.0.424.16909
.NET Version: 8.0.4
Description: The process was terminated due to an unhandled exception.
Exception Info: System.AccessViolationException: Arg_AccessViolationException
Stack:
   at System.Resources.ResourceManager.GetFirstResourceSet(System.Globalization.CultureInfo)
   at System.Resources.ResourceManager.GetString(System.String, System.Globalization.CultureInfo)
   at System.SR.InternalGetResourceString(System.String)
   at System.SR.GetResourceString(System.String)
   at System.Exception.ToString()
   at System.Buffers.SharedArrayPool`1[[System.Char, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].Rent(Int32)
   at System.Runtime.CompilerServices.DefaultInterpolatedStringHandler..ctor(Int32, Int32)
   at System.SR.InternalGetResourceString(System.String)
   at System.SR.GetResourceString(System.String)
   at System.AccessViolationException..ctor()
   at System.Resources.ResourceManager.GetFirstResourceSet(System.Globalization.CultureInfo)
   at System.Resources.ResourceManager.GetString(System.String, System.Globalization.CultureInfo)
   at System.SR.InternalGetResourceString(System.String)
   at System.SR.GetResourceString(System.String)
   at System.AccessViolationException..ctor()
[...]

Issue #2012, opened 2022-02-18
Fail to Load a Gif Image with v2.0.0 but works with v1.0.4 on the same machine.


paulyipkh:
### ImageSharp version

2.0.0

### Other ImageSharp packages and versions

1.0.4

### Environment (Operating system, version and so on)

Windows 10

### .NET Framework version

.NET 5

### Description

I came across a gif image https://www.muhammadniaz.net/wp-content/uploads/2013/06/Stronghold-Crusader-Extreme-Cover.gif which loaded correctly with ImageSharp v1.0.4 but failed with v2.0.0 with an unmanaged exception which cannot be caught by the try-catch

### Steps to Reproduce

```#
try
{
    var client = new HttpClient();
    string url = "https://www.muhammadniaz.net/wp-content/uploads/2013/06/Stronghold-Crusader-Extreme-Cover.gif";
    var bytes = await client.GetByteArrayAsync(url);
    var image = SixLabors.ImageSharp.Image.Load(bytes, out var format);
    Console.WriteLine(format);
}
catch (Exception e)
{
    Console.WriteLine(e.Message);
    Console.WriteLine(e.StackTrace);
    Debugger.Break();
}
```

### Images

![Stronghold-Crusader-Extreme-Cover](https://user-images.githubusercontent.com/12554293/154626625-91f99622-ea29-4ab5-889f-4e74f5b35d35.gif)

JimBobSquarePants:
Confirmed. Memory access violation in `LzwDecoder`. Will compare `GifDecoderCore` in 1.0.4 with the one from main to see what I can spot. It's likely got something to do with our new XMP functionality as I've already  checked the missing colortable fix.

JimBobSquarePants:
It's definitely the XMP application extension handling. I've fixed the reader so far but it looks like the writer does not match the specification also and is not writing the data in subblocks..

JimBobSquarePants:
>Here is another gif file which even fails the v1.0.4

That image is actually corrupt. The minimum code specified is greater than the specification allows. 

Windows Explorer can't open it, nor can Photos, Paint, or Paint.NET.

Edge can though, as can VS Code, returning it as a 276 x 49135 transparent gif which is actually incorrect according to the specification because there's no Graphic Control Extension present in the image and thus no transparency flag.  

I can update the decoder to simply ignore frame indices data if it's corrupt. This will return a 276 x 49135 black gif.

paulyipkh:
[...]

PR #2744, opened 2024-05-31
GifDecoder: Limit lzw bits to a maximum of 12 bits

brianpopow:
### Prerequisites

### Description

The image provided in #2743 has some invalid lzwCode length. This PR changes the gif lzw decoder to raise an exception when the lzwCode is larger then 12 Bits.

### DIFF
--- a/src/ImageSharp/Formats/Gif/LzwDecoder.cs
+++ b/src/ImageSharp/Formats/Gif/LzwDecoder.cs
@@ -20,6 +20,11 @@ internal sealed class LzwDecoder : IDisposable
     private const int MaxStackSize = 4096;
 
     /// <summary>
+    /// The maximum bits for a lzw code.
+    /// </summary>
+    private const int MaximumLzwBits = 12;
+
+    /// <summary>
     /// The null code.
     /// </summary>
     private const int NullCode = -1;
@@ -73,7 +78,7 @@ internal sealed class LzwDecoder : IDisposable
         // It is possible to specify a larger LZW minimum code size than the palette length in bits
         // which may leave a gap in the codes where no colors are assigned.
         // http://www.matthewflickinger.com/lab/whatsinagif/lzw_image_data.asp#lzw_compression
-        if (minCodeSize < 2 || clearCode > MaxStackSize)
+        if (minCodeSize < 2 || minCodeSize > MaximumLzwBits || clearCode > MaxStackSize)
         {
             // Don't attempt to decode the frame indices.
             // Theoretically we could determine a min code size from the length of the provided
@@ -245,7 +250,7 @@ internal sealed class LzwDecoder : IDisposable
         // It is possible to specify a larger LZW minimum code size than the palette length in bits
         // which may leave a gap in the codes where no colors are assigned.
         // http://www.matthewflickinger.com/lab/whatsinagif/lzw_image_data.asp#lzw_compression
-        if (minCodeSize < 2 || clearCode > MaxStackSize)
+        if (minCodeSize < 2 || minCodeSize > MaximumLzwBits || clearCode > MaxStackSize)
         {
             // Don't attempt to decode the frame indices.
             // Theoretically we could determine a min code size from the length of the provided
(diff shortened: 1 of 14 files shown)


src/ImageSharp/Formats/Gif/LzwDecoder.cs
```diff
@@ -73,7 +78,7 @@ public void DecodePixels(int minCodeSize, Buffer2D<byte> pixels)
         // It is possible to specify a larger LZW minimum code size than the palette length in bits
         // which may leave a gap in the codes where no colors are assigned.
         // http://www.matthewflickinger.com/lab/whatsinagif/lzw_image_data.asp#lzw_compression
-        if (minCodeSize < 2 || clearCode > MaxStackSize)
+        if (minCodeSize < 2 || minCodeSize > MaximumLzwBits || clearCode > MaxStackSize)
```
gfoidl, line 81: To save a comparison:
```suggestion
        if ((uint)minCodeSize - 2 > MaximumLzwBits - 2 || clearCode > MaxStackSize)
```
(due the `Unsafe.Add` below I guess this method is hot enough...)

src/ImageSharp/Formats/Gif/LzwDecoder.cs
```diff
@@ -245,7 +250,7 @@ public void SkipIndices(int minCodeSize, int length)
...
```
gfoidl, line 253:
```suggestion
        if ((uint)minCodeSize - 2 > MaximumLzwBits - 2 || clearCode > MaxStackSize)
```
or move this check into a helper, than it's more self-documenting what's going on.

brianpopow:
There are some undisposed buffers when throwing the exception. I have trouble figuring out what buffers are not disposed here.

```
[xUnit.net 00:00:01.10]     IssueTooLargeLzwBits<Rgba32>(provider: Gif/issues/issue_2743.gif[Rgba32]) [FAIL]
  Fehler IssueTooLargeLzwBits<Rgba32>(provider: Gif/issues/issue_2743.gif[Rgba32]) [2 ms]
  Fehlermeldung:
   Expected 0 undisposed buffers but found 7
```

JimBobSquarePants:
Shouldn't we be clamping rather than throwing? The provided image can be opened in the browser.

brianpopow:
> Shouldn't we be clamping rather than throwing?

The specification says 12 bits is the maximum code length, see spec-gif89a.txt in the gif folder. Appendix F, Compression, section 4:

```
4. The output codes are of variable length, starting at <code size>+1 bits per
code, up to 12 bits per code. This defines a maximum code value of 4095
(0xFFF). Whenever the LZW code value would exceed the current code length, the
code length is increased by one. The packing/unpacking of these codes must then
be altered to reflect the new code length.
```

ImageMagick raises an error with this image, so I thought we should, too. I am not sure what will happen when we clamp.

I am more in favor of throwing an exception, because > 12 bits is against the spec.

>  The provided image can be opened in the browser.

The first 7 images are fine, the issue happens with the 8th image.

JimBobSquarePants:
@brianpopow The best approach here is to actually adopt our standard of attempting to decode as much as possible. This way we don't throw we simply return and preserve the data for all the previous frames.  

I don't know why I chose to throw for #2012 but that was a bad choice.

src/ImageSharp/Formats/Gif/LzwDecoder.cs
brianpopow: thanks for the suggestions @gfoidl, but I do not see any difference when I execute gif decoder benchmark.

I do not want to sacrifice the readability here, if it does not really improve the performance.

brianpopow:
> @brianpopow The best approach here is to actually adopt our standard of attempting to decode as much as possible. This way we don't throw we simply return and preserve the data for all the previous frames.

Ok, makes sense. I think this then ready for a final review.

brianpopow:
> I don't know why I chose to throw for https://github.com/SixLabors/ImageSharp/issues/2012 but that was a bad choice.

@JimBobSquarePants should we change that here to also return instead of throwing a exception?

JimBobSquarePants:
> > I don't know why I chose to throw for #2012 but that was a bad choice.
> 
> @JimBobSquarePants should we change that here to also return instead of throwing a exception?

I've already made the change. It was the same conditional check in the LZWDecoder.

JimBobSquarePants (approved):

PR #2748, opened 2024-06-08
Add QuadDistortion to ProjectiveTransformBuilder

Socolin:
### Prerequisites

### Description

Add 2 new methods: `PrependQuadDistortion` and `AppendQuadDistortion` to `ProjectiveTransformBuilder`.

Those method allow to apply a distortion on an image by specifying the requested coordinate for each corner.

Example: 

```cs
using (var img1 = Image.Load("cat-original.png"))
{
	var topLeft = new Point(25, 25);
	var topRight = new Point(200, 75);
	var bottomRight = new Point(225, 175);
	var bottomLeft = new Point(38, 225);

	img1.Mutate(c => c.Transform(new ProjectiveTransformBuilder()
		.AppendQuadDistort(topLeft, topRight, bottomRight, bottomLeft)));
	img1.Save("output-small2.png");
}
```

Source
image
Result
image


JimBobSquarePants:
Thanks @Socolin for this, it is very interesting! I'll have to put my math's hat on to look at this. 

I'm curious. Is this potentially an alternative (better) approach to `ProjectiveTransformBuilder.PrependTaper`?

src/ImageSharp/Common/Helpers/GaussianEliminationSolver.cs
```diff
@@ -0,0 +1,89 @@
...
+namespace SixLabors.ImageSharp.Common.Helpers;
+
+/// <summary>
+/// Represents a solver for systems of linear equations using the Gaussian Elimination method.
+/// This class applies Gaussian Elimination to transform the matrix into row echelon form and then performs back substitution to find the solution vector.
+/// This implementation is based on: https://www.algorithm-archive.org/contents/gaussian_elimination/gaussian_elimination.html
+/// </summary>
+/// <typeparam name="TNumber">The type of numbers used in the matrix and solution vector. Must implement the <see cref="INumber{TNumber}"/> interface.</typeparam>
+internal static class GaussianEliminationSolver<TNumber>
+    where TNumber : INumber<TNumber>
```
JimBobSquarePants, line 14: I think we should be looking for other opportunities to use `INumber<TSelf>` etc in the code to reduce duplication as we slowly modernize this however, for this internal type will we ever not use `float`?

src/ImageSharp/Common/Helpers/GaussianEliminationSolver.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// The matrix passed to this method must be a square matrix.
+    /// If the matrix is singular (i.e., has no unique solution), an <see cref="NotSupportedException"/> will be thrown.
+    /// </remarks>
+    public static void Solve(TNumber[][] matrix, TNumber[] result)
+    {
+        TransformToRowEchelonForm(matrix, result);
+        ApplyBackSubstitution(matrix, result);
+    }
+
+    private static void TransformToRowEchelonForm(TNumber[][] matrix, TNumber[] result)
```
JimBobSquarePants, line 32: I wonder if we can use `DenseMatrix<T>`

src/ImageSharp/Common/Helpers/QuadDistortionHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+        PointF p2 = new(rectangle.X + rectangle.Width, rectangle.Y);
+        PointF p3 = new(rectangle.X + rectangle.Width, rectangle.Y + rectangle.Height);
+        PointF p4 = new(rectangle.X, rectangle.Y + rectangle.Height);
+
+        PointF q1 = topLeft;
+        PointF q2 = topRight;
+        PointF q3 = bottomRight;
+        PointF q4 = bottomLeft;
+
+        // @formatter:off
```
JimBobSquarePants, line 38: I take it this is a JetBrains thing? I'd rather avoid IDE specific commands and just use the default formatting.

Socolin: That's right, this is to keep the matrix easy to read. I can remove them and either just reformat it or keep the formatting as is and avoid applying auto reformat on this part manually.

src/ImageSharp/Common/Helpers/GaussianEliminationSolver.cs
Socolin: When I started I did not know if it was only float or not, so I tried with `INumber`. It was fun to discover those new things :)
I don't mind replacing it with `float`. I'll let you decide this, just let me know if this stay this way or if we change. (I don't know yet if ImageSharp is using double/decimal somewhere, from what I saw it was mainly `float` and `int`.

I also tried when, I was doing some benchmark, to replace TNumber with float, and the performances were the same.

src/ImageSharp/Common/Helpers/GaussianEliminationSolver.cs
JimBobSquarePants: Dense matrix is actually a 1D buffer with 2D API

src/ImageSharp/Common/Helpers/GaussianEliminationSolver.cs
JimBobSquarePants: ImageSharp uses `float` everywhere as we use `Vector4` as a pixel interop. I think using float here would be best.

src/ImageSharp/Common/Helpers/GaussianEliminationSolver.cs
Socolin: Yes I know that, but in the algorithm there is are a lots swap between rows. And swapping 2 pointer is much faster than swapping 8 values.

```
            (matrix[pivotRow], matrix[maxIndex]) = (matrix[maxIndex], matrix[pivotRow]);
            (result[pivotRow], result[maxIndex]) = (result[maxIndex], result[pivotRow]);
```

Socolin:
FYI: The performance test I did 

[TestPerformanceGaussianElimination.zip](https://github.com/user-attachments/files/16169385/TestPerformanceGaussianElimination.zip)

- `Matrix` is using `MathNet.Numerics` Nuget
- `MatrixCustom` is the implementation included in this PR
- `MatrixCustomFloat` is the same implementation with float instead of a generic
- `Matrix8x8` is using `TNumber[,]` instead of `TNumber[][]` with static size of 8x8

```

BenchmarkDotNet v0.13.12, Ubuntu 22.04.4 LTS (Jammy Jellyfish)
AMD Ryzen 9 7950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK 8.0.302
  [Host]     : .NET 8.0.6 (8.0.624.26715), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 8.0.6 (8.0.624.26715), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI


| Method            | Mean     | Error   | StdDev  |
|------------------ |---------:|--------:|--------:|
| Matrix            | 526.6 ns | 1.02 ns | 0.96 ns |
| MatrixCustom      | 237.0 ns | 3.31 ns | 3.09 ns |
| MatrixCustomFloat | 234.1 ns | 4.64 ns | 4.34 ns |
| Matrix8x8         | 293.5 ns | 0.32 ns | 0.28 ns |
```

src/ImageSharp/Common/Helpers/GaussianEliminationSolver.cs
Socolin: Ok, I'll change that

src/ImageSharp/Common/Helpers/GaussianEliminationSolver.cs
JimBobSquarePants: Yeah, I get your point there. Let's keep it as is. 👍

JimBobSquarePants:
@Socolin I'll see if I can find some time over the next few days to pull down your fork and help reimplement it into the latest codebase. I'm keen to add this functionality.

Socolin:
> @Socolin I'll see if I can find some time over the next few days to pull down your fork and help reimplement it into the latest codebase. I'm keen to add this functionality.

I finally have some time again.
I'll push changes related to the today. If it's easier for you to take over this PR, feel free to do it :)

Socolin:
I rebased the branch and pushed the changes

Socolin:
I just saw I'm getting an error because `ProjectiveTransformBuilder Append` take 2 arguments now, should this be computed twice ? Since this operation is taking some processing time, I'm not sure about what to do here.

JimBobSquarePants:
> I just saw I'm getting an error because `ProjectiveTransformBuilder Append` take 2 arguments now, should this be computed twice ? Since this operation is taking some processing time, I'm not sure about what to do here.

The operation takes nanoseconds so I'm not concerned by that.

JimBobSquarePants:
Hi @Socolin would it be possible for you to give me full write access to your fork so I can complete this PR for you? 

There's a [known issue](https://github.com/git-lfs/git-lfs/issues/5001) with Git LFS and GitHub which means I cannot update the reference images without full write access. 

Thanks!

Socolin:
I did grant you access to the repo, can you confirm it's all right ?

JimBobSquarePants:
> I did grant you access to the repo, can you confirm it's all right ?

Perfect. Thank you!

JimBobSquarePants:
I think this is good to go now. Thanks @Socolin for submitting the solution for this. It's a fantastic addition to the library!

PR #2760, opened 2024-06-29
Fix off-by-one error when centering a transform.

### LINKED ISSUE #2753: Rotation of AffineTransform seems to offset the image by 1 pixel
### Prerequisites

### ImageSharp version

3.1.4

ImageSharp.Drawing 2.1.3

Ubuntu 22.04, AMD Ryzen 9 7950X

### .NET Framework version

8.0

### Description

When using the rotation through the AffineTransformBuilder the result is not the same as the `.Rotate()` there seems to be 1 pixel offset.

Here the sample code of what I would expect
```cs
using (var img = new Image<Rgba32>(100, 100, Color.DimGray))
{
	img.Mutate(c => c
		.Rotate(180)
		.Fill(Color.Aqua, new RectangleF(49, 49, 2, 2))
	);
	img.Save("rotation-rotate.png");
}
```
Which give this result:
image

### Steps to Reproduce

When I do the same with the `.Transform()` I'm getting this:

```cs
using (var img = new Image<Rgb24>(100, 100, Color.DimGray))
{
	img.Mutate(c => c
		.Transform(new  AffineTransformBuilder().AppendRotationDegrees(180))
		.Fill(Color.Aqua, new RectangleF(49, 49, 2, 2))
	);
	img.Save("rotation-center.png");
}
```

image
image

Another example with non centered rotation

```cs
for (int i = 0; i <= 10; i++)
{

	using (var img = new Image<Rgba32>(100, 100, Color.DimGray))
	{
		img.Mutate(c => c
			.Transform(new  AffineTransformBuilder().AppendTranslation(new Vector2(100, 100)))
			.Transform(new  AffineTransformBuilder().AppendRotationDegrees(i * 18, new Vector2(100, 100)))
			.Fill(Color.Aqua, new RectangleF(99, 99, 2, 2))
		);
		img.Save($"rotation-{i}.png");
	}
}
```
image
image

The rotation seems to be centered 1 pixel too much toward the bottom-right

### Images

_No response_

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2753 

Fixes the off-by-one error introduced when rotating or skewing images.


src/ImageSharp/Processing/AffineTransformBuilder.cs
```diff
@@ -310,6 +323,32 @@ public Matrix3x2 BuildMatrix(Rectangle sourceRectangle)
...
+    /// </exception>
+    /// <returns>The <see cref="Size"/>.</returns>
+    public Size GetTransformedSize(Rectangle sourceRectangle)
+    {
+        Size size = sourceRectangle.Size;
+
+        // Translate the origin matrix to cater for source rectangle offsets.
+        Matrix3x2 matrix = Matrix3x2.CreateTranslation(-sourceRectangle.Location);
+
+        foreach (Func<Size, Matrix3x2> factory in this.boundsMatrixFactories)
```
tocsoft, line 343:
```suggestion
        Matrix3x2 matrix = Matrix3x2.CreateTranslation(-sourceRectangle.Location);
        matrix.Translation += new Vector2(1);

        foreach (Func<Size, Matrix3x2> factory in this.transformMatrixFactories)
        {
```

I believe this change should prevent the need of having the 2nd collection of `Func`s as I think matrix's for sizing should originate at (+1,+1) from the location of the transform matrix.

JimBobSquarePants: You’ve got me second guessing myself now but I’m not sure that would work. For the most part we want to return the transformed extents based on the explicitly provided information (translate/rotate/scale x, y). It’s only when we perform our own maths to calculate the centre point we go astray. That matrix must be zero based but can then not be used for bounds calculation.

JimBobSquarePants: Confirmed. This example would add an additional unwanted pixel to the right side of the image with the suggested changes.

image

tocsoft (approved):

src/ImageSharp/Processing/AffineTransformBuilder.cs
tocsoft: thanks for validating, just want to make sure we hadn't missed something more simple, this PR looks fine to me then. 👍

PR #2770, opened 2024-07-12
Backport 2759 to 2.1.x

antonfirsov:
### Prerequisites

### Description
Implemented it by a full port of 3.1.x Gif Decoder, so it contains improvements and behavioral changes made since then. It also decodes some stuff 2.1 API doesn't expose.

I didn't fully understand what was the JPEG change about, but the 2.1 decoder detects it as an invalid image. Added the test as a skipped test.


JimBobSquarePants (approved):
Thanks for doing this!

I think incorporating the full decoder was probably wise here. There were likely other critical issues missed over time that should be included. 

The jpeg issue in v3.1.x was that the chrominance quality was being determined as zero. This led to the encoder trying to divide by that value. I'm fine with decoder throwing an exception here.

PR #2787, opened 2024-08-01
Add tagname to debugger visualization for Exif- and Iptc-values, to facilitate easier debugging and discovery

lassevk:
### Prerequisites

- ~[ ] I have provided test coverage for my change (where applicable)~

### Description
The base class for ExifValue and IptcValue has been adorned with a DebuggerDisplayAttribute, so that looking at the collection of metadata values in the debugger is easier to understand. Instead of just the value being displayed, the tag for the value is also displayed, making it easier to see what the values mean, find the right one, and discover what is available.

The change does not impact runtime, as these attributes are only discovered and used by the various debuggers, as such there are no tests. Tests would be difficult to implement as well, considering one would have to mimick a debugger to evaluate and verify that the displayed format is working as intended, both now and into the future.


JimBobSquarePants:
Thanks @lassevk I'll pull this down and review ASAP

lassevk:
To test, you can check the original discussion for the two test methods I used, just put a breakpoint inside and check the in-scope image and review the metadata collections to see how it looks.

JimBobSquarePants:
@lassevk I finally got the time to test this. It's a great addition. Thanks!!

PR #2793, opened 2024-08-15
WIP - Speed improvements to resize convolution (no vpermps w/ FMA)

### LINKED ISSUE #1515: Pre-duplicate kernels in ResizeKernelMap for faster FMA convolution

JimBobSquarePants:
### Prerequisites

### Description
Fixes #1515

This is a replacement for #1518 by @Sergio0694 with most of the work based upon his implementation. I've modernized some of the code and added `Vector512` support also. 

> ### Description
> Follow up to #1513. This PR does a couple things:
> 
> - Switch the resize kernel processing to `float`
> - Add an AVX2 vectorized method to normalize the kernel
> - Vectorize the kernel copy when not using FMA, using `Span<T>.CopyTo` instead
> - Remove the permute8x32 when using FMA, by creating a convolution kernel of 4x the size
> 
> ### Resize convolution codegen diff
> 
> Before:
> 
> ```asm
> vmovsd xmm2, [rax]
> vpermps ymm2, ymm1, ymm2
> vfmadd231ps ymm0, ymm2, [r8]
> ```
> 
> After:
> 
> ```asm
> vmovupd ymm2, [r8]
> vfmadd231ps ymm0, ymm2, [rax] 
> ```

Resize tests currently have four failing tests with minor differences while the ResizeKernelMap has 3 single failing tests. Turning off the periodic kernel map fixes the kernel map failing tests so that is somehow related (I have no idea why). 

I would like to hopefully get issues fixed and merge this because performance in the [Playground Benchmarks](https://github.com/bleroy/core-imaging-playground) looks really, really good so if anyone can spare some time to either provide assistance or point me in the right direction. please let me know. 

CC @antonfirsov @saucecontrol 

```
BenchmarkDotNet v0.13.11, Windows 11 (10.0.22631.4037/23H2/2023Update/SunValley3)
.NET SDK 8.0.400
  [Host]   : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  ShortRun : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI

WarmupCount=5
```
| Method                              | Mean      | Error     | StdDev    | Ratio | RatioSD | Gen0      | Gen1      | Gen2      | Allocated  | Alloc Ratio |
|------------------------------------ |----------:|----------:|----------:|------:|--------:|----------:|----------:|----------:|-----------:|------------:|
| 'System.Drawing Load, Resize, Save' | 423.16 ms | 86.348 ms | 22.424 ms |  1.00 |    0.00 |         - |         - |         - |   13.23 KB |        1.00 |
| 'ImageFlow Load, Resize, Save'      | 280.57 ms | 19.780 ms |  5.137 ms |  0.66 |    0.04 |  500.0000 |  500.0000 |  500.0000 | 4642.65 KB |      351.01 |
| 'ImageSharp Load, Resize, Save'     | 121.74 ms |  5.141 ms |  1.335 ms |  0.29 |    0.01 |         - |         - |         - | 1320.98 KB |       99.87 |
| 'ImageSharp TD Load, Resize, Save'  |  78.83 ms |  4.658 ms |  1.210 ms |  0.19 |    0.01 |  166.6667 |         - |         - | 1309.11 KB |       98.98 |
| 'ImageMagick Load, Resize, Save'    | 411.67 ms | 18.311 ms |  4.755 ms |  0.97 |    0.05 |         - |         - |         - |   54.48 KB |        4.12 |
| 'ImageFree Load, Resize, Save'      | 244.69 ms | 16.343 ms |  2.529 ms |  0.58 |    0.03 | 6000.0000 | 6000.0000 | 6000.0000 |   95.97 KB |        7.26 |
| 'MagicScaler Load, Resize, Save'    |  72.05 ms |  3.984 ms |  1.035 ms |  0.17 |    0.01 |         - |         - |         - |   63.85 KB |        4.83 |
| 'SkiaSharp Load, Resize, Save'      | 146.21 ms | 21.342 ms |  5.542 ms |  0.35 |    0.03 |         - |         - |         - |   88.15 KB |        6.66 |
| 'NetVips Load, Resize, Save'        | 131.72 ms | 14.788 ms |  3.840 ms |  0.31 |    0.02 |         - |         - |         - |   50.95 KB |        3.85 |

### DIFF
--- a/src/ImageSharp/Common/Helpers/Numerics.cs
+++ b/src/ImageSharp/Common/Helpers/Numerics.cs
@@ -1106,7 +1106,39 @@ internal static class Numerics
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static void Normalize(Span<float> span, float sum)
     {
-        if (Vector256.IsHardwareAccelerated)
+        if (Vector512.IsHardwareAccelerated)
+        {
+            ref float startRef = ref MemoryMarshal.GetReference(span);
+            ref float endRef = ref Unsafe.Add(ref startRef, span.Length & ~15);
+            Vector512<float> sum512 = Vector512.Create(sum);
+
+            while (Unsafe.IsAddressLessThan(ref startRef, ref endRef))
+            {
+                Unsafe.As<float, Vector512<float>>(ref startRef) /= sum512;
+                startRef = ref Unsafe.Add(ref startRef, (nuint)16);
+            }
+
+            if ((span.Length & 15) >= 8)
+            {
+                Unsafe.As<float, Vector256<float>>(ref startRef) /= sum512.GetLower();
+                startRef = ref Unsafe.Add(ref startRef, (nuint)8);
+            }
+
+            if ((span.Length & 7) >= 4)
+            {
+                Unsafe.As<float, Vector128<float>>(ref startRef) /= sum512.GetLower().GetLower();
+                startRef = ref Unsafe.Add(ref startRef, (nuint)4);
+            }
+
+            endRef = ref Unsafe.Add(ref startRef, span.Length & 3);
+
+            while (Unsafe.IsAddressLessThan(ref startRef, ref endRef))
+            {
+                startRef /= sum;
+                startRef = ref Unsafe.Add(ref startRef, (nuint)1);
+            }
+        }
+        else if (Vector256.IsHardwareAccelerated)
         {
             ref float startRef = ref MemoryMarshal.GetReference(span);
             ref float endRef = ref Unsafe.Add(ref startRef, span.Length & ~7);
--- a/src/ImageSharp/Common/Helpers/Numerics.cs
+++ b/src/ImageSharp/Common/Helpers/Numerics.cs
@@ -1097,4 +1097,51 @@ internal static class Numerics
     public static nuint Vector512Count<TVector>(int length)
         where TVector : struct
         => (uint)length / (uint)Vector512<TVector>.Count;
+
+    /// <summary>
+    /// Normalizes the values in a given <see cref="Span{T}"/>.
+    /// </summary>
+    /// <param name="span">The sequence of <see cref="float"/> values to normalize.</param>
+    /// <param name="sum">The sum of the values in <paramref name="span"/>.</param>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static void Normalize(Span<float> span, float sum)
+    {
+        if (Vector256.IsHardwareAccelerated)
+        {
+            ref float startRef = ref MemoryMarshal.GetReference(span);
+            ref float endRef = ref Unsafe.Add(ref startRef, span.Length & ~7);
+            Vector256<float> sum256 = Vector256.Create(sum);
+
+            while (Unsafe.IsAddressLessThan(ref startRef, ref endRef))
+            {
+                Unsafe.As<float, Vector256<float>>(ref startRef) /= sum256;
+                startRef = ref Unsafe.Add(ref startRef, (nuint)8);
+            }
+
+            if ((span.Length & 7) >= 4)
+            {
+                Unsafe.As<float, Vector128<float>>(ref startRef) /= sum256.GetLower();
+                startRef = ref Unsafe.Add(ref startRef, (nuint)4);
+            }
+
+            endRef = ref Unsafe.Add(ref startRef, span.Length & 3);
+
+            while (Unsafe.IsAddressLessThan(ref startRef, ref endRef))
+            {
+                startRef /= sum;
+                startRef = ref Unsafe.Add(ref startRef, (nuint)1);
+            }
+        }
+        else
+        {
+            ref float startRef = ref MemoryMarshal.GetReference(span);
+            ref float endRef = ref Unsafe.Add(ref startRef, span.Length);
+
+            while (Unsafe.IsAddressLessThan(ref startRef, ref endRef))
+            {
+                startRef /= sum;
+                startRef = ref Unsafe.Add(ref startRef, (nuint)1);
+            }
+        }
+    }
 }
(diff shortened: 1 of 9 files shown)


src/ImageSharp/Common/Helpers/Vector512Utilities.cs
```diff
@@ -110,6 +110,39 @@ public static Vector512<int> ConvertToInt32RoundToEven(Vector512<float> vector)
...
+    /// If FMA is not supported, the operation is performed as a separate multiplication and addition. This might lead
+    /// to a minor difference in precision compared to the fused operation, particularly in cases where numerical accuracy
+    /// is critical.
+    /// </para>
+    /// </remarks>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector512<float> MultiplyAddEstimate(Vector512<float> a, Vector512<float> b, Vector512<float> c)
+
+        // Don't actually use FMA as it requires many more instruction to extract the
+        // upper and lower parts of the vector and then recombine them.
```
saucecontrol, line 143: What's this about?  When inlined, a helper wrapping `Avx512F.FusedMultiplyAdd` will compile to a single instruction.  One of:

```
Instruction: vfmadd132ps zmm, zmm, zmm
             vfmadd213ps zmm, zmm, zmm
             vfmadd231ps zmm, zmm, zmm
CPUID Flags: AVX512F
```

src/ImageSharp/Common/Helpers/Numerics.cs
```diff
@@ -1097,4 +1097,83 @@ public static nuint Vector512Count<TVector>(this Span<float> span)
...
+        else
+        {
+            ref float startRef = ref MemoryMarshal.GetReference(span);
+            ref float endRef = ref Unsafe.Add(ref startRef, span.Length);
+
+            while (Unsafe.IsAddressLessThan(ref startRef, ref endRef))
+            {
+                startRef /= sum;
+                startRef = ref Unsafe.Add(ref startRef, (nuint)1);
+            }
```
saucecontrol, line 1176:
```suggestion
            for (int i = 0; i < span.Length; i++)
            {
                span[i] /= sum;
            }
```

JIT has no problem eliding bounds checks for simple forward loops over spans/arrays, so the simpler code is just as good in places like this.

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
```diff
@@ -160,17 +210,35 @@ public Vector4 ConvolveCore(ref Vector4 rowStartRef)
...
+                bufferStart++;
+                valuesStart = ref Unsafe.Add(ref valuesStart, (nuint)1);
+            }
+        }
+        else
+        {
+            for (int i = 0; i < this.Length; i++)
+            {
+                this.Values[i] = values[i];
+            }
```
saucecontrol, line 241:
```suggestion
            values.CopyTo(this.Values);
```

saucecontrol:
Brain's not fully awake yet today, but I'll give the maths a look soon.

saucecontrol:
It looks to me like the only differences are due to the change to single precision for kernel normalization and for calculation of the distances passed to the interpolation function.  You'll definitely give up some accuracy there, and I'm not sure it's worth it since the kernels only have to be built once per resize.  You can see [here](https://github.com/SixLabors/ImageSharp/commit/b745ffaf5e2c7c3529f0c82573ed899fcdbeab5b) that @antonfirsov changed the precision to double from the initial implementation some years back.

Since the periodic kernel map relies on each repetition of the kernel weights being *exact*, I can see how precision loss might lead to some differences when compared with a separate calculation per interval.  I've actually never looked at your implementation of the kernel map before, and now my curiosity is piqued because I arrived at something similar myself, but my implementation calculates each kernel window separately, and only replaces that interval with the periodic version if they match exactly.  Part of this was due to a lack of confidence in the maths on my part, as I only discovered the periodicity of the kernel weights by observation and kind of intuited my way to a solution.

@antonfirsov would you mind filling in some gaps on the theory behind your periodic kernel map implementation?  Did you use some paper or other implementation as inspiration, or did you arrive at it observationally like I did?

JimBobSquarePants:
I thought, I'd update this to match latest main. I don't _quite_ understand what is happening with the sampling here and I'm not sure it's worth me taking the time to figure it out. @antonfirsov if you do have any insight I'd appreciate it otherwise I think I'll scrap this.

image

antonfirsov:
> @antonfirsov if you do have any insight I'd appreciate 

Not without going deep down the rabbit hole :(

JimBobSquarePants:
> > @antonfirsov if you do have any insight I'd appreciate
> 
> Not without going deep down the rabbit hole :(

I thought that might be the case. I'll leave this hanging around for a bit longer, but I don't know if it's worth it. I can do a few smaller things instead (like vectorize normalize.)

JimBobSquarePants:
@Sergio0694 Only took me 5 years to figure out the precision issue!! 😛

PR #2803, opened 2024-09-03
Changed all occurenceas of var

stefannikolei:
To have a consistent code base I changed all usages of var and changed the editorconfig rule to error.

Do you want this things to be on error in the editorconfig?


JimBobSquarePants:
Thanks @stefannikolei 

I'm into this but we need to make sure two things are in place:

1. editorconfig changes need to be made upstream in [SharedInfrastructure](https://github.com/SixLabors/SharedInfrastructure) or else they simply get overwritten on build.
2. We need to ensure we are enforcing [Target-typed `new` expressions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/target-typed-new) so we're not adding noise. 

For example:

```c#
RowInterval rows = new RowInterval(yMin, yMax); // ❌

RowInterval rows = new(yMin, yMax); // ✔️
```

The following rule will require updating to `error`
```
csharp_style_implicit_object_creation_when_type_is_apparent = true:warning
```

In addition, we should add a new rule to enforce collection expressions with the same severity.

[Use collection initializers or expressions (IDE0028)](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0028?pivots=dotnet-8-0)

stefannikolei:
I tried to do that. I also Affen some Settins vor Rider. These Look good. Some of the editorconfig settings are not supported in rider. Therefore I could not really test those.

stefannikolei:
Some build errors are false positives. A bug fix for StyleCop.Analyzer is already implemented but not yet released.
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3893

JimBobSquarePants:
@stefannikolei I've introduced a bunch of conflicts here. Do you want me to fix them or would you have time?

stefannikolei:
Just let it sit until the StyleCopAnalyzers thing is figured out. 

Most changes were made by tooling and not by hand.

PR #2812, opened 2024-09-22
cleanup projects and environment variables

kasperk81:
cleanup projects and use modern environment variables.

### DIFF
--- a/src/ImageSharp.ruleset
+++ b/src/ImageSharp.ruleset
@@ -1,4 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <RuleSet Name="ImageSharp" ToolsVersion="17.0">
   <Include Path="..\shared-infrastructure\sixlabors.ruleset" Action="Default" />
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.NetAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.CSharp.NetAnalyzers">
+    <Rule Id="CA2022" Action="Info" />
+  </Rules>
 </RuleSet>
--- a/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
+++ b/src/ImageSharp/Formats/Jpeg/JpegDecoderCore.cs
@@ -16,7 +16,6 @@ using SixLabors.ImageSharp.Metadata.Profiles.Exif;
 using SixLabors.ImageSharp.Metadata.Profiles.Icc;
 using SixLabors.ImageSharp.Metadata.Profiles.Iptc;
 using SixLabors.ImageSharp.Metadata.Profiles.Xmp;
-using SixLabors.ImageSharp.PixelFormats;
 
 namespace SixLabors.ImageSharp.Formats.Jpeg;
 
@@ -1473,7 +1472,7 @@ internal sealed class JpegDecoderCore : ImageDecoderCore, IRawJpegData
 
             this.Frame.ComponentOrder[i / 2] = (byte)componentIndex;
 
-            IJpegComponent component = this.Frame.Components[componentIndex];
+            JpegComponent component = this.Frame.Components[componentIndex];
 
             // 1 byte: Huffman table selectors.
             // 4 bits - dc
--- a/src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
+++ b/src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
@@ -667,12 +667,12 @@ internal static unsafe class Vp8Encoding
 
         // V block.
         dst = dst[8..];
-        if (top != default)
+        if (!top.IsEmpty)
         {
             top = top[8..];
         }
 
-        if (left != default)
+        if (!left.IsEmpty)
         {
             left = left[16..];
         }
@@ -701,7 +701,7 @@ internal static unsafe class Vp8Encoding
 
     private static void VerticalPred(Span<byte> dst, Span<byte> top, int size)
     {
-        if (top != default)
+        if (!top.IsEmpty)
         {
             for (int j = 0; j < size; j++)
             {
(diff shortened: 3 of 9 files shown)


tests/ImageSharp.Tests/ImageSharp.Tests.csproj
```diff
@@ -2,26 +2,14 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <TargetFrameworks>net8.0</TargetFrameworks>
     <DebugSymbols>True</DebugSymbols>
     <AssemblyName>SixLabors.ImageSharp.Tests</AssemblyName>
     <Platforms>AnyCPU;x64;x86;ARM64</Platforms>
     <RootNamespace>SixLabors.ImageSharp.Tests</RootNamespace>
     <Configurations>Debug;Release</Configurations>
   </PropertyGroup>
```
JimBobSquarePants, line 12: We shouldn’t be removing this, instead we should be multi targeting to include .NET 9

JimBobSquarePants (commented):
Thanks for the updates. I’m in Europe on holiday just now so cannot review the changes properly but will do so when I’m back in Australia (3 days)

tests/ImageSharp.Tests/ImageSharp.Tests.csproj
JimBobSquarePants: Thanks. We'll need to update the YAML file `build-and-test.yml` now to ensure the preview framework is installed with the additional workflows.

You can use the version from the release/v3.1.x branch as a guide. 

https://github.com/SixLabors/ImageSharp/blob/4584377879f8e83cd351425c665a9b691047804b/.github/workflows/build-and-test.yml

JimBobSquarePants:
Looks like .NET 9 introduces a bunch of new required fixes. A good thing but also a chore.

JimBobSquarePants:
I'll pull down your branch and compare the output. It's likely we can simply adjust that tolerance.

JimBobSquarePants:
I was right about the zlib changes. The underlying library .NET uses has changed. It actually decodes an extra 2 pixels in the busted image!

https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-9/libraries#compression-with-zlib-ng

src/ImageSharp.ruleset
```diff
@@ -1,4 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <RuleSet Name="ImageSharp" ToolsVersion="17.0">
   <Include Path="..\shared-infrastructure\sixlabors.ruleset" Action="Default" />
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.NetAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.CSharp.NetAnalyzers">
+    <Rule Id="CA2022" Action="Info" />
```
eerhardt, line 5: For my learning, why was this analyzer downgraded from the default `Warning` to `Info`? The issues that this analyzer flags are typically bugs when reading from Streams.

JimBobSquarePants: The warnings were in areas of our code where it wasn’t really applicable. Our decoders are designed to handle degenerate images and in each warning instance we had pre-allocated an empty buffer to read to. If we don’t get a full read then those buffers remain partially unfillled and the decoding process will return when we look for the next chunk.

eerhardt: Hmm, I'm wondering if we have a missed use case / bug in our analyzer then. Do you happen to have a link to example code that it flagged? Maybe it just couldn't understand the pattern.

JimBobSquarePants: I don't think the analyzer will ever be possible to follow the code well enough here.

Take this for example. In `ReadFrame` in the GIF decoder we attempt to read a local color palette from a stream. The analyzer will warn us that we might not get the full read though.

https://github.com/SixLabors/ImageSharp/blob/2d7cf48561765b2a4fc8db6c38c043d676ce5760/src/ImageSharp/Formats/Gif/GifDecoderCore.cs#L432

In this instance we simply do not care as the following method `ReadFrameColors` reads and checks the next byte. If it's invalid then we do nothing.

https://github.com/SixLabors/ImageSharp/blob/2d7cf48561765b2a4fc8db6c38c043d676ce5760/src/ImageSharp/Formats/Gif/GifDecoderCore.cs#L533-L534 

We also do a sense check and break when performing the next iteration to find additional frames so the decoder will quit after decoding as much information as possible.

https://github.com/SixLabors/ImageSharp/blob/2d7cf48561765b2a4fc8db6c38c043d676ce5760/src/ImageSharp/Formats/Gif/GifDecoderCore.cs#L155-L156

So, our code is safe, we could perhaps break earlier but we would end up complicating the code in order to do so. `Info` will work in our scenarios as it will prompt us to check the code but not inhibit our ability to work.

eerhardt: Taking a quick look, I believe the reason why the code is safe is because `BufferedReadStream` guarantees it will fill out the entire passed in buffer on each call to `.Read`.

https://github.com/SixLabors/ImageSharp/blob/2d7cf48561765b2a4fc8db6c38c043d676ce5760/src/ImageSharp/IO/BufferedReadStream.cs#L176-L195

`ReadToBufferDirectSlow` and `ReadToBufferViaCopySlow` read from the underlying stream in a loop until it fills the buffer or hits the end of the stream.

https://github.com/SixLabors/ImageSharp/blob/2d7cf48561765b2a4fc8db6c38c043d676ce5760/src/ImageSharp/IO/BufferedReadStream.cs#L347-L357

https://github.com/SixLabors/ImageSharp/blob/2d7cf48561765b2a4fc8db6c38c043d676ce5760/src/ImageSharp/IO/BufferedReadStream.cs#L320-L326

https://github.com/SixLabors/ImageSharp/blob/2d7cf48561765b2a4fc8db6c38c043d676ce5760/src/ImageSharp/IO/BufferedReadStream.cs#L270-L288

PR #2828, opened 2024-10-22
Rewrite ChunkedMemoryStream

### LINKED ISSUE #2806: WebP and TIFF encoders produce invalid images when writing to HttpContext.Response.Body
### Prerequisites

### ImageSharp version

3.1.5

N/A

Windows 10 19044.2604, 64-bit

### .NET Framework version

.NET 8

### Description

Encoding an `Image` as WebP to a `Microsoft.AspNetCore.Http.HttpContext` (a non-seekable `Stream`) produces an image that cannot be displayed in the browser (Firefox 129.0.2, Chrome 128.0.6613.114) and has a different length and contents from the same `Image` written to a file. Writing an `Image` as TIFF in the same way produces a file which cannot be displayed in the Windows image viewer and contains no valid IFDs. All other formats currently supported by ImageSharp produce valid images when written in this way. Tested with both `Image<Rgba32>` and `Image<Rgb24>`. I would guess it affects other non-seekable `Stream`s with these formats as well.

### Steps to Reproduce

Here is a code sample reproducing the problem. A complete project can be found here: https://github.com/csboling/imagesharp-webp-http

```c#
using SixLabors.ImageSharp;

var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();

app.MapGet(
    "/download.webp",
    async ctx =>
    {
        using (Image<Rgba32> image = new(1, 1))
        {
            ctx.Response.ContentType = "image/webp";
            await image.SaveAsWebpAsync(ctx.Response.Body);
        }
    });

app.Run();
```

### Images

This ZIP contains example 1x1 pixel RGBA32 images encoded to WebP and TIFF to an `HttpContext.Response`, then downloaded with Chrome (invalid.webp, invalid.tif) as well as images encoded directly to a file (valid.webp, valid.tif).

[examples.zip](https://github.com/user-attachments/files/16911652/examples.zip)

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2806 

~`ChunkedMemoryStream` contained multiple writing bugs and was too costly to fix/maintain relative to performance benefits so I'm just ditching it.~

- Complete rewrite of `ChunkedMemoryStream` to simplify it and fix numerous bugs. 
- Ensure ImageEncoder uses the chunked stream when encoding non-seekable streams.


antonfirsov:
> relative to performance benefits

IMO the memory benefits were significant. With the current decoder & encoder design, non seekable streams are always fully  buffered into memory. A switch to `MemoryStream` will reintroduce significant GC allocations for large inputs all around the library.

Recommendations:
- In case there is no bug in stream writing, do not remove `ChunkedMemoryStream` for decoders.
- Instead of using `MemoryStream` for encoders, implement our own non-chunked MemoryStream that still uses `MemoryAllocator`. Although [buffers over 4MB](https://github.com/SixLabors/ImageSharp/blob/21ec18e1b1186c3f1d549269271177d4516b3ba5/src/ImageSharp/Memory/Allocators/UniformUnmanagedMemoryPoolMemoryAllocator.cs#L15) won't be pooled, in typical scenarios encoded files are smaller.

JimBobSquarePants:
> > relative to performance benefits
> 
> IMO the memory benefits were significant. With the current decoder & encoder design, non seekable streams are always fully buffered into memory. A switch to `MemoryStream` will reintroduce significant GC allocations for large inputs all around the library, this will be a noticable regression for users who deal with http or other kinds of network streams.
> 
> Recommendations:
> 
> * In case there is no bug in stream writing, do not remove `ChunkedMemoryStream` for decoders, since the largest streams are typically the input ones.
> * Instead of using `MemoryStream` for encoders, implement our own non-chunked MemoryStream that still uses `MemoryAllocator`. Although [buffers over 4MB](https://github.com/SixLabors/ImageSharp/blob/21ec18e1b1186c3f1d549269271177d4516b3ba5/src/ImageSharp/Memory/Allocators/UniformUnmanagedMemoryPoolMemoryAllocator.cs#L15) won't be pooled, in typical scenarios encoded files are smaller.

Thanks for the review @antonfirsov I've instead chosen to completely rewrite the `ChunkedMemoryStream` to simplify the implementation. It's much easier to maintain now!

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -94,247 +57,133 @@ public override long Position
...
-            base.Dispose(disposing);
+            return -1;
         }
-    }
 
-    /// <inheritdoc/>
-    public override void Flush()
-    {
+        _ = this.Read(this.singleByteBuffer, 0, 1);
+        return MemoryMarshal.GetReference<byte>(this.singleByteBuffer);
```
mgravell, line 105: since the `byte[]` path goes via AsSpan(), why not use the span approach directly?

```
Span<byte> buffer = stackalloc byte[1];
return Read(buffer) == 1 ? buffer[0] : -1
```

ideally with `SkipLocalsInit` enabled

(you can't do this for the async path, though)

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -94,247 +57,133 @@ public override long Position
...
-            chunkSize = this.writeOffset;
+            remaining = count;
         }
 
+        int bytesToRead = (int)remaining;
         int bytesRead = 0;
-        int offset = 0;
-        int count = buffer.Length;
-        while (count > 0)
+        while (bytesToRead != 0 && this.currentChunk != this.memoryChunkBuffer.Length)
```
mgravell, line 145: I don't know enough about the underlying implementation here; if this is doing additional downstream reads, you might prefer to exit after the first read (you're only required to read "some" data - you don't need to fill the supplied data, just return at least 1 byte or 0 for EOF); if all the data is *already* loaded, I wonder whether your `memoryChunkBuffer` is duplicating the innards of `ReadOnlySequence<byte>` - that already has all the `Slice`, `CopyTo` etc you might want; just a suggestion, though (I can help you grok `ReadOnlySequence<T>` if you're not already familiar)

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -94,247 +57,133 @@ public override long Position
...
-                return -1;
-            }
-
-            this.readChunk = this.readChunk.Next;
-            this.readOffset = 0;
-            chunkBuffer = this.readChunk.Buffer;
-        }
-
-        return chunkBuffer.GetSpan()[this.readOffset++];
+        MemoryMarshal.Write(this.singleByteBuffer, ref value);
```
mgravell, line 182: ditto `stackalloc`; or possibly even the more exotic:

```
byte b = 0;
var span = MemoryMarshal.CreateSpan(ref b, 1);
```

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -344,157 +193,203 @@ public override void Write(byte[] buffer, int offset, int count)
...
-        chunkBuffer.GetSpan()[this.writeOffset++] = value;
+        this.position += bytesRead;
     }
 
     /// <summary>
-    /// Copy entire buffer into an array.
+    /// Writes the stream contents to a byte array, regardless of the <see cref="Position"/> property.
     /// </summary>
-    /// <returns>The <see cref="T:byte[]"/>.</returns>
+    /// <returns>A new <see cref="T:byte[]"/>.</returns>
```
mgravell, line 312: Note that this is very inefficient; I would suggest trying to deprecate this kind of API - especially if we *can* use ROS, for example;

```
[Obsolete("prefer " + nameof(AsReadOnlySequence)] public byte[] ToArray() => AsReadOnlySequence().ToArray();
public ReadOnlySequence<byte> AsReadOnlySequence() => /* magic happens */
```

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -12,44 +14,23 @@ namespace SixLabors.ImageSharp.IO;
 /// Chunks are allocated by the <see cref="MemoryAllocator"/> assigned via the constructor
 /// and is designed to take advantage of buffer pooling when available.
 /// </summary>
-internal sealed class ChunkedMemoryStream : Stream
+public class ChunkedMemoryStream : Stream
```
colgreen, line 17: Consider reverting this back to being sealed. I.e. is this class designed to be sub-classed? If not then sealing can allow the JITter to make certain optimisations around the method calling of the virtual/override methods

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -12,44 +14,23 @@ namespace SixLabors.ImageSharp.IO;
...
-    private MemoryChunk? memoryChunk;
-
-    // The total number of allocated chunks
-    private int chunkCount;
-
-    // The length of the largest contiguous buffer that can be handled by the allocator.
-    private readonly int allocatorCapacity;
-
-    // Has the stream been disposed.
+    private long length;
```
colgreen, line 22: Sidenote: IMO it's fairly conventional in C# land to prefix private field names with an underscore, to allow easy distinction from local variables, and to avoid excessive use of `this.`. I assume this is your personal preference, but thought I'd mention it as IMO it is somewhat non-idiomatic C#.

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -12,44 +14,23 @@ namespace SixLabors.ImageSharp.IO;
...
+    private long position;
+    private int currentChunk;
```
colgreen, line 24: These names could be considered a little misleading/confusing.

E.g. `currentChuck` is an index into `memoryChunkBuffer`, so I think maybe call it `currentChunkIdx`.

Whereas `currentChunkIdx` is an offset/index /within/ the current chunk, so maybe call it `intraChunkByteIdx`, `chunkByteIdx`, or `checkByteOffset`? etc.

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -507,48 +402,82 @@ private void EnsureNotDisposed()
...
+            this.memoryChunks.Add(chunk);
+            this.Length += chunk.Length;
+        }
+
+        public void Dispose()
         {
-            chunk.Dispose();
-            chunk = chunk.Next;
+            this.Dispose(true);
+            GC.SuppressFinalize(this);
```
colgreen, line 443: I think it's not necessary to call `GC.SuppressFinalize(this)` in a sealed class with no finalizer. This would be to cover sub-types that have a finalizer (in scenarios where there is no finalizer defined directly on the type).

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants: Thanks, but I'd rather stick to using the language as designed rather than using conventions carried over from C. 

The `this` keyword provides important context IMO and encouraged consistancy throughout a codebase.

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants: Yeah.. good point. I've opted for `bufferIndex` and `chunkIndex` as an improvement.

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants: Ah... I'd completely forgotten about that method!

Rather than decorating a public mthod with `[SkipLocalsInit]` i've opted for the following:

```c#
/// <inheritdoc/>
public override int ReadByte()
{
    Unsafe.SkipInit(out byte b);
    return this.Read(MemoryMarshal.CreateSpan(ref b, 1)) == 1 ? b : -1;
}
```

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants:
```c#
/// <inheritdoc/>
public override void WriteByte(byte value)
    => this.Write(MemoryMarshal.CreateSpan(ref value, 1));
```

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants: Unfortunately, we need this to when reading XMP data for the V3 build however I wish to rewrite the XMPProfile type for V4 to avoid passing arrays around. Once this is merged to the V3 branch I'll upstream and make additional changes.

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants: Force of habit. Well spotted!

src/ImageSharp/IO/ChunkedMemoryStream.cs
JimBobSquarePants: Yeah, looking at `MemoryChunkBuffer` there is definitely some overlap. Ideally, I should be tracking the buffer and chunk indexes internally within that class. 

However, this type needs to be expandable on-demand which AFAIK is not possible with `ReadonlySequence<T>`.

mgravell: The underlying buffer-chain is as mutable as you want it to be (it is your chain, ultimately); if you want to resize it, that is usually as simple as simply creating a new ROS (which is a lightweight struct just tracking the start and end), specifying the new bounds. The chin bits aren't trivial, but not too complex. I guess if what you already have works well, it might be overkill to touch it, though.

src/ImageSharp/IO/ChunkedMemoryStream.cs
antonfirsov: Looks much better!

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -507,48 +402,82 @@ private void EnsureNotDisposed()
...
-#pragma warning disable IDE1006 // Naming Styles
-        const int _128K = 1 << 17;
-        const int _4M = 1 << 22;
-        return i < 16 ? _128K * (1 << (int)((uint)i / 4)) : _4M;
-#pragma warning restore IDE1006 // Naming Styles
+        public IEnumerator<MemoryChunk> GetEnumerator()
+            => ((IEnumerable<MemoryChunk>)this.memoryChunks).GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator()
+            => ((IEnumerable)this.memoryChunks).GetEnumerator();
```
antonfirsov, line 450: I don't see any code enumerating this with `foreach`, so `IEnumerable` implementation can be deleted.

src/ImageSharp/IO/ChunkedMemoryStream.cs
antonfirsov: @colgreen the project's preferred coding style is based on [Framework design guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/), and on StyleCop recommendations. The guidelines [explicitly prohibit](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields) prefixing variables.

tests/ImageSharp.Tests/IO/ChunkedMemoryStreamTests.cs
```diff
@@ -30,7 +30,7 @@ public class ChunkedMemoryStreamTests
     [Fact]
```
antonfirsov, line 30: Is it possible to extend these tests to stress the corner(?) cases which were buggy in the previous implementation.

tests/ImageSharp.Tests/IO/ChunkedMemoryStreamTests.cs
```diff
@@ -107,8 +107,8 @@ public void MemoryStream_ReadByteBufferTest(int length)
     [InlineData(DefaultSmallChunkSize * 16)]
     public void MemoryStream_ReadByteBufferSpanTest(int length)
```
antonfirsov, line 108: I would also include lengths over `DefaultSmallChunkSize * 16` and make `buffer` a parameter. Would test cases, when `buffer.Length > DefaultSmallChunkSize`.

tests/ImageSharp.Tests/IO/ChunkedMemoryStreamTests.cs
```diff
@@ -167,18 +167,20 @@ public void MemoryStream_WriteToTests()
     [Fact]
     public void MemoryStream_WriteToSpanTests()
```
antonfirsov, line 168: Write tests could stress more cases with different sizes. See my comment on the read tests.

src/ImageSharp/IO/ChunkedMemoryStream.cs
antonfirsov, line 145: Even though current code looks good and `Slice` would throw for a negative number, I find this safer to maintain. Same for `Write`.

```suggestion
        while (bytesToRead >= 0 && this.currentChunk != this.memoryChunkBuffer.Length)
```

antonfirsov (commented):
- Previous reviewers had good points.
- Added some nitpicking.
- Improving funcitonal coverage would be valuable.

Looks good otherwise.

src/ImageSharp/IO/ChunkedMemoryStream.cs
colgreen: The guidelines state `Internal and private fields are not covered by guidelines,` (I believe those guidelines are primarily related to public API surface, rather than private/internal naming). The underscore prefix for private fields is very common in my experience, e.g. it's used widely in Microsoft .NET repos.

However, this topic is probably not relevant in the context of this PR :)

tests/ImageSharp.Tests/IO/ChunkedMemoryStreamTests.cs
JimBobSquarePants: Yeah. Tests have been massivly exapanded. We now test reading larger buffers and test encoding to webp for all test images.

tests/ImageSharp.Tests/IO/ChunkedMemoryStreamTests.cs
JimBobSquarePants: `length` is already determined by the parameter but I've expanded to double the previous maximum length

src/ImageSharp/IO/ChunkedMemoryStream.cs
```diff
@@ -507,48 +402,82 @@ private void EnsureNotDisposed()
...
+        {
+            // Increment chunks sizes with moderate speed, but without using too many buffers from the
+            // same ArrayPool bucket of the default MemoryAllocator.
+            // https://github.com/SixLabors/ImageSharp/pull/2006#issuecomment-1066244720
+            const int b128K = 1 << 17;
+            const int b4M = 1 << 22;
+            return i < 16 ? b128K * (1 << (int)((uint)i / 4)) : b4M;
+        }
+
+        private void Dispose(bool disposing)
```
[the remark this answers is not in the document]
JimBobSquarePants: That's an override of the base `Stream` method though. I can't implement `Dispose` directly. All other implementations have been simplified now though.

JimBobSquarePants:
@antonfirsov Are you happy with me pushing on with this?

tests/ImageSharp.Tests/IO/ChunkedMemoryStreamTests.cs
```diff
@@ -13,6 +13,8 @@ namespace SixLabors.ImageSharp.Tests.IO;
 /// </summary>
 public class ChunkedMemoryStreamTests
 {
+    private readonly Random bufferFiller = new();
```
antonfirsov, line 16:
```suggestion
    private readonly Random bufferFiller = new(123);
```

Setting a seed makes the test run deterministic, which I personally prefer.

antonfirsov (approved):
LGTM. One small suggestion for the test code.

tests/ImageSharp.Tests/IO/ChunkedMemoryStreamTests.cs
JimBobSquarePants: Yeah. Good point.

PR #2855, opened 2024-12-22
Use explicit types

stefannikolei:
Use explicit type instead of var.

Now only this is changed not as in #2803 where some more things were changed and Stylecop was not happy with it

PR #2874, opened 2025-01-30
V3 : Decode Chunky Tile Rows Directly.

### LINKED ISSUE #2873: How do I use ImageSharp with large orthomosaics?
ImageSharp v3.1.6

### Discussed in https://github.com/SixLabors/ImageSharp/discussions/2870

<sup>Originally posted by **aevitas** January 27, 2025</sup>
Hi,

I've been a long time user of ImageSharp for all my imaging needs, so when I started a recent project using satellite imagery and drone footage, I once again came back to this library.

The images I'm working with are large geographical TIFF images, also known as orthomosaics. They can be stitched together from numerous other images, and as a result can become quite large (some of our current images are north of 1.5GB).

I'd like to use ImageSharp to transform, transcode and "tile" these images into smaller images, but when I attempt to load one of these images, I get the following exception:

```
Unhandled exception. SixLabors.ImageSharp.InvalidImageContentException: Cannot decode image. Failed to allocate buffers for possibly degenerate dimensions: 0x0.
 ---> SixLabors.ImageSharp.Memory.InvalidMemoryOperationException: Attempted to allocate a buffer of negative length=-227135488.
   at SixLabors.ImageSharp.Memory.InvalidMemoryOperationException.ThrowNegativeAllocationException(Int64 length)
   at SixLabors.ImageSharp.Memory.UniformUnmanagedMemoryPoolMemoryAllocator.Allocate[T](Int32 length, AllocationOptions options)
```

Debugging this, I can see the following number of bytes being allocated every time:

1. 62424
2. 15606
3. 15606
4. 262144
5. -227135488

Which I assume is four metadata allocations, followed by an allocation for the actual (overflowed size) image data.

Given that the TIFF I'm trying to load is 740MB, I fail to see why it would need to allocate over 2.14GB of memory to process the image. Other libraries such as GDAL are perfectly fine loading these images in an instant, but don't offer the image manipulation I need.

Is this a limitation with ImageSharp specifically? Is there something I'm doing wrong and is there a way to use these images with ImageSharp?

I've explored making my own fork of ImageSharp and replacing the allocation with just flat direct allocations as I'd be running this from a command line so the heap would never be long-lived, but it seemed like a non-trivial modification so I'd like to make sure there aren't any more sensible approaches.

Can provide sample files if needed.</div>

JimBobSquarePants:
### Prerequisites

### Description

 Fixes #2873
Avoids the integer overflow issue by removing the interim allocation completely and decoding directly to the pixel buffer.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -121,10 +121,10 @@ jobs:
           dotnet-version: |
             7.0.x
 
-      - name: Ensure Dotnet Path
+      - name: Ensure DotNet Path
         run: echo "/usr/share/dotnet" >> $GITHUB_PATH
 
-      - name: Confirm .NET Version
+      - name: Confirm DotNet Version
         run: dotnet --version  # This will now find dotnet in /usr/share/dotnet
 
       - name: DotNet Build
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -112,6 +112,8 @@ jobs:
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: |
+            8.0.x
+            7.0.x
             6.0.x
 
       - name: DotNet Setup Preview
@@ -121,12 +123,12 @@ jobs:
           dotnet-version: |
             7.0.x
 
-      - name: Ensure DotNet Path
-        if: ${{ matrix.options.os != 'windows-latest' }}
-        run: echo "/usr/share/dotnet" >> $GITHUB_PATH
+      # - name: Ensure DotNet Path
+      #   if: ${{ matrix.options.os != 'windows-latest' }}
+      #   run: echo "/usr/share/dotnet" >> $GITHUB_PATH
 
-      - name: Confirm DotNet Version
-        run: dotnet --version
+      # - name: Confirm DotNet Version
+      #   run: dotnet --version
 
       - name: DotNet Build
         if: ${{ matrix.options.sdk-preview != true }}
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -67,7 +67,7 @@ jobs:
 
     steps:
       - name: Install libgdi+, which is required for tests running on ubuntu
-        if: ${{ matrix.options.os == 'buildjet-4vcpu-ubuntu-2204-arm' }}
+        if: ${{ contains(matrix.options.os, 'ubuntu') }}
         run: sudo apt-get -y install libgdiplus libgif-dev libglib2.0-dev libcairo2-dev libtiff-dev libexif-dev
 
       - name: Git Config
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -122,10 +122,11 @@ jobs:
             7.0.x
 
       - name: Ensure DotNet Path
+        if: ${{ matrix.options.os != 'windows-latest' }}
         run: echo "/usr/share/dotnet" >> $GITHUB_PATH
 
       - name: Confirm DotNet Version
-        run: dotnet --version  # This will now find dotnet in /usr/share/dotnet
+        run: dotnet --version
 
       - name: DotNet Build
         if: ${{ matrix.options.sdk-preview != true }}
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -43,21 +43,21 @@ jobs:
           #   sdk-preview: true
           #   runtime: -x64
           #   codecov: false
-          # - os: ubuntu-latest
-          #   framework: net6.0
-          #   sdk: 6.0.x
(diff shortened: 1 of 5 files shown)


stefannikolei:
@JimBobSquarePants  Look at https://github.com/SixLabors/ImageSharp/pull/2855 i fixed the build there. And just checked it still works

JimBobSquarePants:
> @JimBobSquarePants Look at #2855 I fixed the build there. And just checked it still works

Ah no, it's something else. I've already done that using the following which covers all cases.

```
if: ${{ contains(matrix.options.os, 'ubuntu') }}
```

What I'm seeing is a difference in behavior with dotnet build between old builds and now. The older SDKs are no longer installed on the image by default and after installing it seems to throw errors for stuff that did not throw before. 

It's a proper head scratcher!

stefannikolei:
> > @JimBobSquarePants Look at #2855 I fixed the build there. And just checked it still works
> 
> Ah no, it's something else. I've already done that using the following which covers all cases.
> 
> ```
> if: ${{ contains(matrix.options.os, 'ubuntu') }}
> ```
> 
> What I'm seeing is a difference in behavior with dotnet build between old builds and now. The older SDKs are no longer installed on the image by default and after installing it seems to throw errors for stuff that did not throw before.
> 
> It's a proper head scratcher!

I saw that. But I just triggered a build in my branch and I got no build errors oO

JimBobSquarePants:
> > > @JimBobSquarePants Look at #2855 I fixed the build there. And just checked it still works
> > 
> > 
> > Ah no, it's something else. I've already done that using the following which covers all cases.
> > ```
> > if: ${{ contains(matrix.options.os, 'ubuntu') }}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What I'm seeing is a difference in behavior with dotnet build between old builds and now. The older SDKs are no longer installed on the image by default and after installing it seems to throw errors for stuff that did not throw before.
> > It's a proper head scratcher!
> 
> I saw that. But I just triggered a build in my branch and I got no build errors oO

I'm running a build against the legacy release/v3.1.x branch. That targets .NET 6 and 7. The main branch targets .NET 8 and 9 both of which are installed on the image.

stefannikolei:
> I'm running a build against the legacy release/v3.1.x branch. That targets .NET 6 and 7. The main branch targets .NET 8 and 9 both of which are installed on the image.


Ah missed that one.. Was wondering why dotnet-setup was bumped. 

Perhaps you can try to not go agains ubuntu-latest instead use ubuntu-22-04 or the version it worked the last time

JimBobSquarePants:
@stefannikolei it works if you install NET 8 and NET 6!! 🤷I'll just update the script to do that.

stefannikolei:
> @stefannikolei it works if you install NET 8 and NET 6!! 🤷I'll just update the script to do that.

just checked which version of ubuntu was used 
image

But when it works it works. Perhaps ubuntu-latest should not be used for the future. Instead use a explicit version, perhaps those issue won't come then.

JimBobSquarePants:
> > @stefannikolei it works if you install NET 8 and NET 6!! 🤷I'll just update the script to do that.
> 
> just checked which version of ubuntu was used image
> 
> But when it works it works. Perhaps ubuntu-latest should not be used for the future. Instead use a explicit version, perhaps those issue won't come then.

I would but they change old images also for some reason I don't understand. 

https://github.com/actions/runner-images/pull/11503

I think whatever issues there are have been fixed for .NET 8+ so going forward it hopefully should not be a problem.

brianpopow (approved):
looks good 👍

aevitas:
Very much looking forward to these changes!

Curious, why are test runners using versions of .NET that are out of support, but not the most recent version? I could see the regressions argument, but arguably no one should be running EOL versions of .NET to begin with.

brianpopow:
@aevitas while the image mentioned in [Discussion 2870](https://github.com/SixLabors/ImageSharp/discussions/2870#discussioncomment-11976852) can be decoded, it still suffers from the same issue mentioned in #2877. I am working on fixing that, too.

JimBobSquarePants:
@brianpopow do you want to use this branch? 

If you can tell me where the issue manifests maybe I can have a look if you’re stuck?

brianpopow:
> @brianpopow do you want to use this branch?

@JimBobSquarePants I would suggest to first merge this branch into main and continue on a new branch, since it is a different issue.

> If you can tell me where the issue manifests maybe I can have a look if you’re stuck?

I think, I know what needs to be done. The horizontal predictor is reversed by applying it to each row of the image, but for tiled images, it needs to be applied to each row of the tile.

brianpopow:
@JimBobSquarePants I think I made a mistake. I thought this was going into the main branch.

JimBobSquarePants:
@brianpopow All good. I wanted to target v3 since we are still a long way from a v4 launch. I'll open a second PR for v4.

PR #2875, opened 2025-01-30
Install libgdiplus on ubuntu

stefannikolei:


stefannikolei:
@JimBobSquarePants  Its working here 

(╯°□°)╯︵ ┻━┻

.github/workflows/build-and-test.yml
```diff
@@ -73,7 +73,7 @@ jobs:
 
     steps:
       - name: Install libgdi+, which is required for tests running on ubuntu
-        if: ${{ matrix.options.os == 'buildjet-4vcpu-ubuntu-2204-arm' }}
+        if: ${{ contains(matrix.options.os, 'ubuntu') }}
```
JimBobSquarePants, line 76: You need to do the same in `code-coverage.yml` also.

JimBobSquarePants (approved):
Thanks for this. 👍

Issue #2877, opened 2025-01-31
Saving TIFF files to a different format produces scan line image


aevitas:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

3.1.6

### Other ImageSharp packages and versions

N/A

### Environment (Operating system, version and so on)

OSX ARM, M1, Rider IDE, .NET 9

### .NET Framework version

.NET 9

### Description

When loading the following TIFF image with ImageSharp, and subsequently saving it to a different format, an image containing exclusively noisy scan lines is produced

Original image:

![Image](https://github.com/user-attachments/assets/be3d54b8-271f-4206-a5a3-d4eafb24de16)

ImageSharp output:

![Image](https://github.com/user-attachments/assets/f39364d3-3973-4b56-8e71-8b902e6731c7)

### Steps to Reproduce

The following code reproduces this issue when used with the attached TIFF file and ImageSharp 3.1.6:

```
var b = Image.Load(tiff.FullName);
b.SaveAsJpeg(Path.Combine(tilesDir.FullName, $"{tileName}-imagesharp.jpg"));
```

[4-7.tif.zip](https://github.com/user-attachments/files/18619086/4-7.tif.zip)

brianpopow:
I think it has something to do with reversing the horizontal predictor and tiled images. Not sure yet howto fix it.

PR #2878, opened 2025-02-02
Undo horizontal prediction for each tile row in case of tiled tiff's

brianpopow:
### Prerequisites

### Description

This PR fixes issue: #2877: The horizontal predictor needs to be reversed for each tile row in case of tiled images.


src/ImageSharp/Formats/Tiff/TiffDecoderCore.cs
```diff
@@ -712,6 +713,13 @@ private void DecodeTilesChunky<TPixel>(
                 {
                     Span<byte> uncompressedPixelRow = uncompressedPixelBufferSpan.Slice(uncompressedPixelBufferOffset, bytesToCopy);
                     tileBufferSpan.Slice(tileBufferOffset, bytesToCopy).CopyTo(uncompressedPixelRow);
+
+                    // Undo the horziontal predictor for each tile row.
+                    if (this.Predictor == TiffPredictor.Horizontal)
```
brianpopow, line 718: @JimBobSquarePants I am not sure, if this is the best approach to undo the predictor here. I tried to do this in HorizontalPredictor.Undo(), but could not figure out a way without replicating the logic to iterate over the tiles in DecodeTilesChunky(). Maybe you have a better idea howto do this?

JimBobSquarePants: So, are you writing the row, then undoing the write?

brianpopow: It currently works like this:
1. For each Tile: Decompress tile data
2. For each row in tile: Undo horizontal predictor
3. When done with all tiles: decode color

The horizontal predictor is used with lzw and deflate compression to achieve better compression. It assumes that neighboring pixel values do not change much and therefore only stores the difference between two consecutive pixels instead the actual pixel values.

Because it is tight to lzw and deflate compression, it feels to me this should be the concern of the decompressor and not be handled outside (which is the case for none tiled images).

I will try to gather more tiled image test cases for all the different color formats and then see if I can figure out a way how I can move the undoing of the predictor inside the uncompressing method.

JimBobSquarePants: Ah, I understand. 

I'm just trying to read through [TiffLibrary](https://github.com/yigolden/TiffLibrary) and [LibTiff.NET](https://github.com/BitMiracle/libtiff.net) to see how it works there as a reference. I agree the decompressor feels like it should be responsible though.

brianpopow: I have moved undoing the predictor for tiles into the HorizontalPredictor class now with `UndoTile()`. Maybe this is better now. I am still open for any better suggestions. 

I think it is ready for review. I have gathered as much different example files of tiled tiff I could.

JimBobSquarePants: Thanks @brianpopow I'll get on it ASAP

aevitas:
@brianpopow is this change available in a nightly/beta build? I'd really like to give this a try.

JimBobSquarePants:
> @brianpopow is this change available in a nightly/beta build? I'd really like to give this a try.

@aevitas you’d need to build from this branch. The PR hasn’t been merged yet.

JimBobSquarePants:
@brianpopow I'll be looking at this over the weekend when I can.

src/ImageSharp/Formats/Tiff/Compression/HorizontalPredictor.cs
```diff
@@ -62,6 +62,126 @@ public static void Undo(Span<byte> pixelBytes, int width, TiffColorType colorTyp
...
+    /// Inverts the horizontal predictor for one row.
+    /// </summary>
+    /// <param name="pixelBytes">Buffer with decompressed pixel data.</param>
+    /// <param name="width">The width in pixels of the row.</param>
+    /// <param name="y">The row index.</param>
+    /// <param name="colorType">The color type of the pixel data.</param>
+    /// <param name="isBigEndian">If set to <c>true</c> decodes the pixel data as big endian, otherwise as little endian.</param>
+    public static void UndoRow(Span<byte> pixelBytes, int width, int y, TiffColorType colorType, bool isBigEndian)
+    {
+        switch (colorType)
```
JimBobSquarePants, line 91: Can this switch not happen outside the loop?

JimBobSquarePants (commented):
This all looks great to me. Thanks for making the changes. 

The only question I have is about moving the condition outside of the loop and making the individual row methods instead work on the whole 2D buffer.

src/ImageSharp/Formats/Tiff/Compression/HorizontalPredictor.cs
JimBobSquarePants: Sorry, I had to pull this down to fully understand it. Makes sense now!

JimBobSquarePants (approved):
Thanks for this @brianpopow Amazing work as ever!

Issue #2859, opened 2025-01-14
AccessViolationException when decoding gif


andreas-eriksson:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

3.1.6

### Other ImageSharp packages and versions

N/A

### Environment (Operating system, version and so on)

Windows 11

### .NET Framework version

.NET 8

### Description

Loading the attached gif causes an AccessViolationException and brings down the process.

 ```
at SixLabors.ImageSharp.Formats.Gif.LzwDecoder.DecodePixelRow(System.Span`1<Byte>)
 at SixLabors.ImageSharp.Formats.Gif.GifDecoderCore.ReadFrameColors[[SixLabors.ImageSharp.PixelFormats.Rgba32, SixLabors.ImageSharp, Version=3.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13]](SixLabors.ImageSharp.IO.BufferedReadStream, SixLabors.ImageSharp.Image`1<SixLabors.ImageSharp.PixelFormats.Rgba32> ByRef, SixLabors.ImageSharp.ImageFrame`1<SixLabors.ImageSharp.PixelFormats.Rgba32> ByRef, System.ReadOnlySpan`1<SixLabors.ImageSharp.PixelFormats.Rgb24>, SixLabors.ImageSharp.Formats.Gif.GifImageDescriptor ByRef)
 at SixLabors.ImageSharp.Formats.Gif.GifDecoderCore.ReadFrame[[SixLabors.ImageSharp.PixelFormats.Rgba32, SixLabors.ImageSharp, Version=3.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13]](SixLabors.ImageSharp.IO.BufferedReadStream, SixLabors.ImageSharp.Image`1<SixLabors.ImageSharp.PixelFormats.Rgba32> ByRef, SixLabors.ImageSharp.ImageFrame`1<SixLabors.ImageSharp.PixelFormats.Rgba32> ByRef)
[...]

PR #2890, opened 2025-02-20
Backport #2859 to release/2.1.x

antonfirsov:
### Prerequisites

### Description
Backport #2859 to release/2.1.x

### DIFF
--- a/tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
+++ b/tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
@@ -4,7 +4,7 @@
 using System;
 using System.IO;
 using Microsoft.DotNet.RemoteExecutor;
-
+using Microsoft.DotNet.XUnitExtensions;
 using SixLabors.ImageSharp.Formats.Bmp;
 using SixLabors.ImageSharp.Memory;
 using SixLabors.ImageSharp.Metadata;
@@ -598,8 +598,7 @@ namespace SixLabors.ImageSharp.Tests.Formats.Bmp
             }
         }
 
-        [Theory]
-        [PlatformSpecific(~TestPlatforms.Linux)] // See discussion on https://github.com/SixLabors/ImageSharp/pull/2890
+        [ConditionalTheory]
         [WithFile(Os2BitmapArray, PixelTypes.Rgba32)]
         [WithFile(Os2BitmapArray9s, PixelTypes.Rgba32)]
         [WithFile(Os2BitmapArrayDiamond, PixelTypes.Rgba32)]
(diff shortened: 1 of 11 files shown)


antonfirsov:
Please squash this mess if it will ever build 😅

I've seen some unrelated WEBP test failures locally, probably they will show up here.

antonfirsov:
Even BMP failures. I don't understand this, neither we changed the product code nor we added tests since 2.1.9. Will continue investigating tomorrow, but if anyone has an idea, let me know!

JimBobSquarePants:
@antonfirsov Github keeps changing what’s installed on images. I discovered with V3 branch that weirdly you had to install NET 8 along with the current target version for anything to build. I would use the yaml file from that branch as the basis and change the target frameworks.

brianpopow:
I cannot reproduce the bmp test failures on windows. When I run:
```
dotnet test -c Release --filter "Format=Bmp"
```

All tests are green.

tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
```diff
@@ -611,6 +611,11 @@ public void BmpDecoder_CanDecode_Os2v2Header<TPixel>(TestImageProvider<TPixel> p
         public void BmpDecoder_CanDecode_Os2BitmapArray<TPixel>(TestImageProvider<TPixel> provider)
             where TPixel : unmanaged, IPixel<TPixel>
         {
+            if (TestEnvironment.IsLinux)
+            {
+                throw new SkipTestException("See discussion on https://github.com/SixLabors/ImageSharp/pull/2890");
+            }
```
antonfirsov, line 617: The Linux failure is a mystery: it only shows when running the test suite. If I run a console app against the compiled `ImageSharp.dll` on Ubuntu, the image loads without an issue. I have spent an hour trying to hook a debugger into VS Code, but it keeps ignoring my breakpoints.

I ended up skipping the test on Linux. (1) It's likely a false alert that appears only with `dotnet test` (2) likely not a regression against the previous version, (3) an edge-case anyways.

brianpopow: I think it is ok to skip those bmp tests on linux, they are extremely niche bitmap variants. It's also a mystery to me why this is different on linux than on windows.

antonfirsov: It was an LFS issue (╯°□°)╯︵ ┻━┻

brianpopow: Wow, good find

antonfirsov:
@JimBobSquarePants it should be good to merge this now and start publishing stuff.

JimBobSquarePants (approved):
LGTM Thanks! 👍

PR #2899, opened 2025-03-11
Add ANI decoder support

Poker-sang:
### Prerequisites

### Description

implements AniDecoder for ImageSharp, and encoder if the pr approved

### Comment

Since the ANI file may contain a two-dimensional array of ImageFrames, I flattened the original ImageFrames as follows

I have kept all `Metadata` as much as possible, but this may result in a structure that is not intuitive, and may require further discussion with you) reserved the possibility of future modifications

drawio

ANI is also a RIFF file, so I referenced RiffHelper, perhaps we should move it out of the WEBP namespace

[[1]](https://www.informit.com/articles/article.aspx?p=1189080&seqNum=3)
[[2]](https://www.cnblogs.com/youlin/p/3282263.html)


src/ImageSharp/Formats/Ani/AniDecoderCore.cs
```diff
@@ -0,0 +1,417 @@
...
+using SixLabors.ImageSharp.Memory;
+using SixLabors.ImageSharp.Memory.Internals;
+using SixLabors.ImageSharp.Metadata;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace SixLabors.ImageSharp.Formats.Ani;
+
+internal class AniDecoderCore(DecoderOptions options) : ImageDecoderCore(options)
+{
+    private enum ListIconChunkType
```
gfoidl, line 24:
```suggestion
    private enum ListIconChunkType : byte
```
To make the type smaller (i.e. the list smaller as well).

src/ImageSharp/Formats/Ani/AniHeader.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    public uint Planes { get; }
+
+    public uint DisplayRate { get; }
+
+    public AniHeaderFlags Flags { get; }
+
+    public static ref AniHeader Parse(ReadOnlySpan<byte> data) => ref Unsafe.As<byte, AniHeader>(ref MemoryMarshal.GetReference(data));
+
+    public void WriteTo(in Stream stream) => stream.Write(MemoryMarshal.AsBytes(MemoryMarshal.CreateReadOnlySpan(in this, 1)));
```
gfoidl, line 31:
```suggestion
    public void WriteTo(Stream stream) => stream.Write(MemoryMarshal.AsBytes(MemoryMarshal.CreateReadOnlySpan(in this, 1)));
```
`Stream` is a reference type, so `in` doesn't provide much value here. Actually it's a [de-optimization](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgEkB5AbhvqdYBKAVwB2GAJb4YLAMIR8AB3EAbGFADKagG7iwMXD2q8AzE1IMZDGgG8aDe01PEUDAGIQIACnUZY2fAy4AJQMALwAfAwAqrgwdJ7Bhg6OTC4AQthQ3r4w/oEhEdGx5AlBSQ52DgDaALIwGAAWEAAmbIrKnnWNLW0KyhwKEhAiuCwAchBsIsriIrMA5kEAupX2ClDiWtgYMPxIqUVx2X4BwQzWAL6rVtTJtfVNre2dDz3tA0Mj45PTswvL13Wm22u2Y+2chxKswYPhO+XOV2oFyAA===) as the argument can't be passed in a cpu-register, instead the address needs to be taken,

JimBobSquarePants:
Thanks @Poker-sang I’ll have a deep look at this asap.

src/ImageSharp/Formats/Ani/AniFrameMetadata.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// Initializes a new instance of the <see cref="AniFrameMetadata"/> class.
+    /// </summary>
+    public AniFrameMetadata()
+    {
+    }
+
+    /// <summary>
+    /// Gets or sets the display time for this frame (in 1/60 seconds)
+    /// </summary>
+    public uint Rate { get; set; }
```
JimBobSquarePants, line 24: We should be normalizing the naming to match other formats. `FrameDelay`

src/ImageSharp/Formats/Ani/AniFrameMetadata.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// <summary>
+    /// Gets or sets the encoding height. <br />
+    /// Can be any number between 0 and 255. Value 0 means a frame height of 256 pixels or greater.
+    /// </summary>
+    public byte? EncodingHeight { get; set; }
+
+    /// <summary>
+    /// Gets or sets the <see cref="ImageMetadata"/> of one "icon" chunk.
+    /// </summary>
+    public ImageMetadata? SubImageMetadata { get; set; }
```
JimBobSquarePants, line 46: We shouldn't nest sub data like this. The metadata needs to own all required data.

src/ImageSharp/Formats/Ani/AniFrameMetadata.cs
```diff
@@ -0,0 +1,70 @@
...
+    IDeepCloneable IDeepCloneable.DeepClone() => new AniFrameMetadata { Rate = this.Rate };
+
+    /// <inheritdoc/>
+    public FormatConnectingFrameMetadata ToFormatConnectingFrameMetadata() => new FormatConnectingFrameMetadata() { Duration = TimeSpan.FromSeconds(this.Rate / 60d) };
+
+    /// <inheritdoc/>
+    public void AfterFrameApply<TPixel>(ImageFrame<TPixel> source, ImageFrame<TPixel> destination)
+        where TPixel : unmanaged, IPixel<TPixel>
+    {
+        throw new NotImplementedException();
```
JimBobSquarePants, line 65: You'll need to implement this.

src/ImageSharp/Formats/Ani/AniFrameMetadata.cs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /// <inheritdoc/>
+    public AniFrameMetadata DeepClone() => new() { Rate = this.Rate };
```
JimBobSquarePants, line 69: Lots of properties missing here.

src/ImageSharp/Formats/Ani/AniFrameMetadata.cs
```diff
@@ -0,0 +1,70 @@
...
+
+    /// <summary>
+    /// Gets or sets the display time for this frame (in 1/60 seconds)
+    /// </summary>
+    public uint Rate { get; set; }
+
+    /// <summary>
+    /// Gets or sets the frames count of **one** "icon" chunk.
+    /// </summary>
+    public int FrameCount { get; set; } = 1;
```
JimBobSquarePants, line 29: This should be the sequence number.

src/ImageSharp/Formats/Ani/AniDecoderCore.cs
```diff
@@ -0,0 +1,416 @@
...
+                }
+
+                isRootFrame = false;
+
+                return target;
+            }));
+
+            ImageFrameMetadata rootFrameMetadata = img.Frames.RootFrame.Metadata;
+            AniFrameMetadata aniFrameMetadata = rootFrameMetadata.GetAniMetadata();
+            aniFrameMetadata.Rate = rate == default ? aniMetadata.DisplayRate : rate[i];
```
JimBobSquarePants, line 113: IsEmpty

src/ImageSharp/Formats/Ani/AniDecoderCore.cs
```diff
@@ -0,0 +1,416 @@
...
+
+        ImageMetadata metadata = new();
+        AniMetadata aniMetadata = this.ReadHeader(dataStartPosition, dataSize, metadata);
+
+        List<(ListIconChunkType Type, Image<TPixel> Image)> frames = [];
+        this.HandleRiffChunk(out Span<int> sequence, out Span<uint> rate, dataStartPosition, dataSize, aniMetadata, frames, DecodeFrameChunk);
+
+        List<ImageFrame<TPixel>> list = [];
+
+        foreach (int i in sequence)
```
JimBobSquarePants, line 58: We need to flatten out the frames and metadata here. Each frame should have an `AniFrameMetadata` property containing all the required information including what sequence the frame belongs to. 

No other format metadata should be added to the output.

JimBobSquarePants (changes requested):
Thanks for this. It's a promising start...

I think we can make some changes here to allow flattening out the frame sequences a little better which should help when writing the encoder. 

I've added some additional comments regarding coding style and merging of functionality also.

src/ImageSharp/Formats/Ani/AniFrameMetadata.cs
Poker-sang: It is actually intentionally designed as `FrameCount`, because each frame of `ani` file may contain multiple subframes (similar to different resolution ICOs), so this field is used to indicate how many subframes are contained in this frame

src/ImageSharp/IO/BufferedReadStream.cs
```diff
@@ -194,6 +199,49 @@ public override int Read(Span<byte> buffer)
         return this.ReadToBufferViaCopyFast(buffer);
     }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public bool TryReadUnmanaged<T>(out T result)
```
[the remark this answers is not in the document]
Poker-sang: I wrote `TryReadUnmanaged` to reduce one memory copy (simplifying from _stream->span->struct_ to _stream->struct_) and to make it easier to read fixed-length chunks each time.

If it's using runtime built-in the quickest way is perhaps `ref Unsafe.As<byte, AniHeader>(ref MemoryMarshal.GetReference(data))`, I'm not quite sure which one you're referring to?

src/ImageSharp/Formats/Webp/RiffOrListChunkHeader.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    public readonly uint Size;
+
+    public readonly uint FormType;
+
+    public ReadOnlySpan<byte> FourCcBytes => MemoryMarshal.CreateReadOnlySpan(ref Unsafe.As<uint, byte>(ref Unsafe.AsRef(in this.FourCc)), sizeof(uint));
+
+    public bool IsRiff => this.FourCc is 0x52_49_46_46; // "RIFF"
+
+    public bool IsList => this.FourCc is 0x4C_49_53_54; // "LIST"
```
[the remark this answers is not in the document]
Poker-sang: Not used yet, just in case

src/ImageSharp/Formats/Webp/RiffOrListChunkHeader.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Six Labors.
+// Licensed under the Six Labors Split License.
+
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace SixLabors.ImageSharp.Formats.Webp;
+
+internal readonly struct RiffOrListChunkHeader
```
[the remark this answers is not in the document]
Poker-sang: Should I split it into `RiffChunkHeader` and `ListChunkHeader`?

Poker-sang:
Thank you for the review ❤ I fixed some simple review comments, but there is a lot of work that needs to continue to be discussed before we can move forward.

Based on your suggestions I'm guessing it may need to be improved in this way:

1. The rate block section could probably be included into `AniMetadata`

2. Include all members of `Ico/Cur/BmpMetadata` in `AniFrameMetadata` and indicate which fields to use via flag. bmp is simpler, but for `Ico/Cur` frames could be specified using `SequenceNumber` for encode.

src/ImageSharp/Formats/Ani/AniFrameMetadata.cs
JimBobSquarePants: Yes. The way I'm thinking is to use a `SequenceNumber` property to identify which individual frames of different resolutions belong to a single ANI frame. Imagine the individual frames following this sequence:
1,1,1,2,3,4,4,4. 

When we write the encoder (we'll need one to complete this PR) we can then use those sequence numbers to group the frames for encoding.

JimBobSquarePants:
> Thank you for the review ❤ I fixed some simple review comments, but there is a lot of work that needs to continue to be discussed before we can move forward.
> 
> Based on your suggestions I'm guessing it may need to be improved in this way:
> 
> 1. The rate block section could probably be included into `AniMetadata`
> 2. Include all members of `Ico/Cur/BmpMetadata` in `AniFrameMetadata` and indicate which fields to use via flag. bmp is simpler, but for `Ico/Cur` frames could be specified using `SequenceNumber` for encode.

I don't think we should store the rate block in the `AniMetadata` as we can have much more granular control when stored as `FrameDelay` in individual `AniFrameMetadata` instances. When encoding we can check if there is any variation in those values and use the header for all rates; otherise individual rate chunks. 

Thinking about it some more I believe a good plan would be to have an enum in `AniFrameMetadata` of type `AniFrameFormat` which states whether the frame is Ico/Cur/Bmp. 

We can then use nullable sub properties for `IcoFrameMetadata`, and `CurFrameMetadata` to `AniFrameMetadata`. which will save some effort. We can add `NotNullWhen` attributes to the properties based on the value of our `AniFrameFormat` property.

src/ImageSharp/Formats/Ani/AniDecoderCore.cs
Poker-sang: How can i remove the FrameMetadata from decoded frame? Can I know if a Metadata is included without creating a new Metadata?

src/ImageSharp/Formats/Webp/RiffOrListChunkHeader.cs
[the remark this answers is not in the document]
JimBobSquarePants: Probably yes.

JimBobSquarePants:
@Poker-sang I'll pull this down and have another deep look tonight.

Poker-sang:
New structure diagram for your reference

drawio

JimBobSquarePants:
@Poker-sang Haven't forgotten this, just focusing on closing some issues first.

wheremyfoodat:
Would like to thank you and say that this PR is super useful to me as I'm currently working on some MacOS mouse cursor software. A lot of cursors online are made for Windows and come in .ani format, so having to support it is unavoidable and I'm very excited to try this out in the future.

src/ImageSharp/Formats/Webp/RiffOrListChunkHeader.cs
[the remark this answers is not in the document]
JimBobSquarePants: Let's only add stuff we need.

src/ImageSharp/IO/BufferedReadStream.cs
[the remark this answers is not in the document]
JimBobSquarePants: It's a potentially dangerous method given that we `stackalloc` inside the method. 

If you look elsewhere (`GifDecoderCore`) you will see that we have explicit `Parse` methods for our structs using a fixed length buffer. You could use a custom length InlineArray (See `InlineArray.tt`) to do the same.

JimBobSquarePants:
@Poker-sang Are you still looking to complete this? 

I've just merged the latest main into your branch and added some comments, but the code still needs some work to clean it up. We'll also need a full encoder implementation.

Poker-sang:
> New structure diagram for your reference
> 
> drawio

The ANI format encoder should first determine the structure of ANI files within ImageSharp (as shown in the structure diagram I shared earlier). but now some Metadata are still not included yet (CUR, ICO, BMP `ImageMetadata`). And I'm not sure if the current structure is ok. So perhaps we should first discuss how to design the APIs for these Metadatas?

ANI encoders are simply combinations of existing encoders which are easy to write. Don't worry about that(

src/ImageSharp/Formats/Ico/IcoConfigurationModule.cs
```diff
@@ -15,6 +13,6 @@ public void Configure(Configuration configuration)
     {
         configuration.ImageFormatsManager.SetEncoder(IcoFormat.Instance, new IcoEncoder());
         configuration.ImageFormatsManager.SetDecoder(IcoFormat.Instance, IcoDecoder.Instance);
-        configuration.ImageFormatsManager.AddImageFormatDetector(new IconImageFormatDetector());
+        configuration.ImageFormatsManager.AddImageFormatDetector(new IcoImageFormatDetector());
     }
```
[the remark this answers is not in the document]
JimBobSquarePants: Required change

src/ImageSharp/Formats/Icon/IconDir.cs
```diff
@@ -5,39 +5,65 @@
...
         => MemoryMarshal.Cast<byte, IconDir>(data)[0];
 
-    public readonly unsafe void WriteTo(Stream stream)
+    /// <summary>
+    /// Writes the icon directory header to the destination stream.
+    /// </summary>
+    /// <param name="stream">The destination stream.</param>
+    public readonly void WriteTo(Stream stream)
         => stream.Write(MemoryMarshal.Cast<IconDir, byte>([this]));
 }
```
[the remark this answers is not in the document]
JimBobSquarePants: Incorrect. You're reading collection-expression syntax as array syntax.

src/ImageSharp/Formats/Ani/AniEncoderCore.cs
```diff
@@ -0,0 +1,468 @@
...
+                AniIconFrameMetadataProvider provider = new(format);
+
+                if (format is AniFrameFormat.Ico)
+                {
+                    this.icoEncoder ??= new IcoEncoderCore(new IcoEncoder
+                    {
+                        PixelSamplingStrategy = this.encoder.PixelSamplingStrategy,
+                        Quantizer = this.encoder.Quantizer,
+                        TransparentColorMode = this.encoder.TransparentColorMode
+                    });
```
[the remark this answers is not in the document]
JimBobSquarePants: Fixed

JimBobSquarePants:
@Poker-sang I've finished this off. Thanks for your help!

Poker-sang:
Thank you! 😆

PR #2762, opened 2024-07-03
Mono AOT decoder workaround for slow jpeg decoding.

JimBobSquarePants:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

This PR is an attempt to work around issues found with the Mono AOT compiler which causes slow performance on IOS Android

https://github.com/dotnet/runtime/issues/71210

In the linked Issue, Analysis from the Mono team highlighted [this indirection](https://github.com/dotnet/runtime/issues/71210#issuecomment-1902783869) as a culprit. 

>The AOT compiler does appear to have problems figuring out which instances to generate. In this specific case, the caller is
`ImageDecoderUtilities:Decode<Rgba32>` which calls `IImageDecoderInternals::Decode` on an argument. So in theory, the aot compiler could figure out that the call could possible go to `JpegDecoderCore::Decode<Rgba32>` and generate that instance. Currently, this kind of analysis is not done.

This PR removes that indirection completely by introducing an internal base class, `ImageDecoderCore` for all `XXDecoderCore` instances. In addition, seeding has been introduced for `SpectralConverter<TPixel>` and others. 

This is currently untested but I'm confident that this should improve matters.

Issue #1446, opened 2020-11-29
Aligned allocation for faster SIMD


antonfirsov:
## Problem

Intel's [optimization manual](https://www.intel.com/content/dam/doc/manual/64-ia-32-architectures-optimization-manual.pdf) places strong emphasis on the importance of alignment of memory addresses SIMD registers are stored/loaded from. Significant performance penalties occur when loads/stores are crossing (64 byte) cache line boundaries.

Managed (array pool) arrays give us no guarantees on the alignment of the `&arrray[0]` base address, which means that all ImageSharp SIMD code suffers globally from these penalties today.

dotnet/runtime#27146 initially proposed a mechanism to control the alignment of objects on the pinned heap, but that feature did not get into the final implementation.

## Idea
Considering that the earliest release where runtime support might be available is .NET 6.0, and it won't be available in erlier releases anyways, I'm proposing a library-specific solution:
1. Introduce `AllocationOptions.AlignedPinned`
2. When the flag is set, return pinned or umanaged buffers with a "hacked" base address (idea taken from https://github.com/dotnet/runtime/issues/33244#issuecomment-595848832):

```C#
IntPtr baseAddress = GetBaseAddressOfPinnedArrayOrAllocateUnmanagedBuffer();

// Align the pointer
IntPtr alignedAddress = (IntPtr)((nint)(baseAddress + sizeof(IntPtr) + (alignment - 1)) & ~(alignment - 1));

// Store the pointer to the memory baseAddress to free right before the aligned pointer 
*(((IntPtr*)alignedAddress) - 1) = baseAddress;

return AlignedBuffer(alignedAddress);
```

## The question I can't figure out

I want to make sure there are no unwanted side-effects on the GC. What buffers should we use with this trick, considering that the (mid to large) buffers we want to align will be always temporary? Note that max lifetime of these buffers is the run of an image codec or a processing operation, pinned allocations will be typically bound to these kind of big, expensive operations and won't happen within individual primitive steps.
- Should we just pin the arrays of our pool?
- Or use `Marshal.Alloc`?
- Is it better to use the Pinned Heap (when available)?

@tannergooding @saucecontrol any thoughts/concerns?

Trying to also summon @benaadams if it's not too impolite.

benaadams:
[...]

PR #2917, opened 2025-05-05
Modernize JPEG Color Converters

JimBobSquarePants:
### Prerequisites

### Description

With .NET 8 we can take advantage of the `VectorXXX<T>` types to provide cross-platform hardware intrinsics with great support on all platforms. This allows us to simplify our code by normalizing the approach and additionally take advantage of performance improvements due to [excellent `Vector128` support on Mono](https://github.com/dotnet/runtime/blob/v8.0.15/src/mono/mono/mini/simd-intrinsics.c#L1149-L1219).

This PR replaces the existing CPU-specific implementation with implementations for `Vector128`, `Vector256`, and `Vector512`.

Benchmarks look good overall (_exceeding or meeting the main branch_). `Vector512` is slightly disappointing and only shows marginal speedup on my machine. I believe this is due to two primary factors:

1. **Throttling** - Intel CPUs aggressively downclock when AVX-512 instructions are executed to manage thermals and power draw. This frequency reduction applies across all cores, which can negate the expected throughput gains from using 512-bit wide vectors.

2. **Memory utilization issues** - AVX-512 loads and stores perform best when data is aligned to 64-byte boundaries. Without guaranteed alignment, memory operations may fall back to unaligned access paths that split operations internally, causing increased micro-op pressure and reduced bandwidth efficiency. Additionally, when working with larger vectors, cache line pressure and prefetcher behavior become more critical; misaligned or poorly laid-out memory can further exacerbate performance bottlenecks. This highlights the importance of using properly aligned memory when utilizing `Vector512<T>` in hot paths.

Handling memory alignment is something [we've discussed before](https://github.com/SixLabors/ImageSharp/issues/1446) and the APIs to do so exist now that would make it a lot easier. I do wonder though whether we should simply default to pinning with alignment given almost all our code goes through some sort of SIMD operation?

@beeradmoore I'm wondering if you'd be able to benchmark this branch using your approach from #2762 to see if we've finally reached decent performance on Android? Only if you have time of course, I wouldn't want to put you out. 

@tannergooding I'm tagging you here if that's OK as we've discussed these changes previously and you input is always super valuable.


beeradmoore:
@JimBobSquarePants, sure thing. Would also be nice to update my results to .NET 9 as well.

By chance do any of them GitHub builds put out a nuget package or am I building it myself for the tests?

JimBobSquarePants:
> @JimBobSquarePants, sure thing. Would also be nice to update my results to .NET 9 as well.
> 
> By chance do any of them GitHub builds put out a nuget package or am I building it myself for the tests?

Legend! 

They only produce a package once merged so you would have to build against this particular branch. Building should work out of the box.

src/ImageSharp/Common/Helpers/SimdUtils.cs
```diff
@@ -36,30 +35,24 @@ internal static Vector4 PseudoRound(this Vector4 v)
...
-    /// <see>
-    ///     <cref>https://github.com/g-truc/glm/blob/master/glm/simd/common.h#L110</cref>
-    /// </see>
     /// </summary>
     /// <param name="v">The vector</param>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     internal static Vector<float> FastRound(this Vector<float> v)
     {
-        if (Avx2.IsSupported)
+        if (Avx2.IsSupported && Vector<float>.Count == Vector256<float>.Count)
```
JimBobSquarePants, line 43: I don't know whether this is ever false. But I wanted the check to be a little more strict.

src/ImageSharp/Common/Helpers/SimdUtils.HwIntrinsics.cs
```diff
@@ -616,44 +616,6 @@ public static Vector256<float> MultiplyAdd(
                 return Fma.MultiplyAdd(vm1, vm0, va);
             }
 
-            if (Avx.IsSupported)
```
tannergooding: nit: You can have a similar path for Arm64
```csharp
if (AdvSimd.IsSupported)
{
    return AdvSimd.FusedMultiplyAdd(va, vm1, vm0);
}
```

When you're able to move to .NET 9+, you can switch all this to just `Vector128.MultiplyAddEstimate(x, y, addend)` which handles this optimization for you.

src/ImageSharp/Common/Helpers/SimdUtils.cs
tannergooding: It can be false even when `Avx2.IsSupported` as someone can set `DOTNET_PreferredVectorBitWidth=128`, so having an explicit check is goodness.

Notably you have `Sse41.RoundToNearestInteger` that can be used for `128-bit`, `Avx512F.RoundScale` that can be used for 512-bit, and `AdvSimd.RoundToNearest` that can be used for Arm64.

.NET 9+ defines `Vector.Round` which can be used to handle the xplat consideration without having to manually code it all (and same method exists on `Vector128/256/512`)

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -193,13 +193,60 @@ public static Vector128<int> ConvertToInt32RoundToEven(Vector128<float> vector)
             return AdvSimd.ConvertToInt32RoundToEven(vector);
         }
 
-        Vector128<float> sign = vector & Vector128.Create(-0.0f);
-        Vector128<float> val_2p23_f32 = sign | Vector128.Create(8388608.0f);
+        Vector128<float> sign = vector & Vector128.Create(-0F);
```
tannergooding, line 196: No general helper exists for .NET 9+ here...

It's something we could probably pattern recognize as `ConvertToInt32(Round(vector))` though. If it's a pattern you'd like to see us optimize, logging a suggestion on dotnet/runtime would be beneficial. Alternatively some `ConvertToInt32(vector, MidpointRounding)` API could be provided as well and an API suggestion could be opened.

src/ImageSharp/Common/Helpers/Vector256Utilities.cs
```diff
@@ -103,13 +103,60 @@ public static Vector256<int> ConvertToInt32RoundToEven(Vector256<float> vector)
...
+    {
+        if (Fma.IsSupported)
+        {
+            return Fma.MultiplyAdd(vm0, vm1, va);
+        }
+
+        if (Avx.IsSupported)
+        {
+            return Avx.Add(Avx.Multiply(vm0, vm1), va);
+        }
```
tannergooding, line 155: This is unnecessary, it does the same thing as the `va + (vm0 * vm1)` logic just below

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -193,13 +193,60 @@ public static Vector128<int> ConvertToInt32RoundToEven(Vector128<float> vector)
...
+        Vector128<float> va,
+        Vector128<float> vm0,
+        Vector128<float> vm1)
+    {
+        if (Fma.IsSupported)
+        {
+            return Fma.MultiplyAdd(vm1, vm0, va);
+        }
+
+        return va + (vm0 * vm1);
```
tannergooding, line 247: Similar comment as on the SimdUtils.HwIntrinsics.cs copy of the method, where you could add an `AdvSimd` path for Arm64 and eventually switch to `MultiplyAddEstimate` when you target .NET 9+

src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverter.CmykVector128.cs
```diff
@@ -69,26 +68,25 @@ public static void ConvertFromRgb(in ComponentValues values, float maxValue, Spa
...
+            for (nuint i = 0; i < n; i++)
             {
-                Vector128<float> ctmp = AdvSimd.Subtract(scale, Unsafe.Add(ref srcR, i));
-                Vector128<float> mtmp = AdvSimd.Subtract(scale, Unsafe.Add(ref srcG, i));
-                Vector128<float> ytmp = AdvSimd.Subtract(scale, Unsafe.Add(ref srcB, i));
-                Vector128<float> ktmp = AdvSimd.Min(ctmp, AdvSimd.Min(mtmp, ytmp));
+                Vector128<float> ctmp = scale - Unsafe.Add(ref srcR, i);
+                Vector128<float> mtmp = scale - Unsafe.Add(ref srcG, i);
+                Vector128<float> ytmp = scale - Unsafe.Add(ref srcB, i);
+                Vector128<float> ktmp = Vector128.Min(ctmp, Vector128.Min(mtmp, ytmp));
```
tannergooding, line 79: Just as a note, on .NET 9+ the behavior of `Min` has changed for x86/x64.

On .NET 8 and prior, `Vector128.Min` differed in behavior from `Math.Min` and it basically just emitted whatever single instruction was fastest. This lead to non-determinism across x64, Arm64, and other platforms when dealing with `-0` and `NaN`.

On .NET 9 and later, `Vector128.Min` now behaves identically to `Math.Min` and is deterministic across all platforms. This can show as a performance regression on x64 for some older hardware as additional instructions must be executed to make this guarantee (newer hardware has dedicated instructions that behave correctly).

If (on .NET 9+) you don't have any concerns around how `-0` is sorted compared to `+0` and you don't have any concerns around `NaN`, then you can use `Vector128.MinNative` to continue getting the "fast" but potentially non-deterministic behavior.

src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverterVector128.cs
```diff
@@ -19,14 +18,14 @@ internal abstract partial class JpegColorConverterBase
...
+    internal abstract class JpegColorConverterVector128 : JpegColorConverterBase
     {
-        protected JpegColorConverterArm(JpegColorSpace colorSpace, int precision)
+        protected JpegColorConverterVector128(JpegColorSpace colorSpace, int precision)
             : base(colorSpace, precision)
         {
         }
 
-        public static bool IsSupported => AdvSimd.IsSupported;
+        public static bool IsSupported => Vector128.IsHardwareAccelerated && Vector128<float>.IsSupported;
```
tannergooding, line 28: nit: You don't need to check `Vector128<float>.IsSupported`, this is statically `true`.

The `Vector128<T>.IsSupported` is namely for generic contexts where you don't know concretely what the `T` is.

tannergooding: In general you can presume that `byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `float`, and `double` are always supported.

You can presume that `nint` and `nuint` are supported on .NET 7+

We will likely add support for `Half` at some point in the future, at which point a `Vector128.AsHalf()` method will exist to "document" this guarantee of support.

tannergooding (approved):
LGTM. Just left a couple comments around things you could do for Arm64 support or things to keep in mind for .NET 9+

beeradmoore:
So some odd things here. It seems between 3.1.8 and this PR there is a regression in Android performance. However, going from 3.1.4 (or that other PR) to even the slowest results in this PR performance has improved. This could be the update to .NET 9, this could be changes in ImageSharp. Below are the numbers I currently have.

This is also with these build configs which is I believe what we found to be optimal from that other issue.

```xml
    <PropertyGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android' AND '$(Configuration)' == 'Release'">
        <EnableLLVM>true</EnableLLVM>
        <RunAOTCompilation>true</RunAOTCompilation>
        <AndroidEnableProfiledAot>false</AndroidEnableProfiledAot>
    </PropertyGroup>

    <PropertyGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android' AND '$(Configuration)' == 'Debug'">
        <!-- performance improvements for debug mode, will break hot reload. -->
        <UseInterpreter>false</UseInterpreter>
    </PropertyGroup>
```

## Debug (this PR)
| Device           | JpgLoad |  JpgResize | PngLoad | PngResize |
|------------------|---------|------------|---------|-----------|
| iPhone           |         |            |         |           |
| iOS Simulator    |         |            |         |           |
| Android          | 1375.0  | 1677.0     | 38.1    | 47.7      |
| Android Emulator | 226.8   | 284.9      | 13.6    | 14.9      |
| macOS            |         |            |         |           |
| Windows          |         |            |         |           |


## Debug (3.1.8)
| Device           | JpgLoad |  JpgResize | PngLoad | PngResize |
|------------------|---------|------------|---------|-----------|
| iPhone           |         |            |         |           |
| iOS Simulator    |         |            |         |           |
| Android          | 1080.9  | 1302.9     | 37.6    | 45.4      |
| Android Emulator | 189.4   | 238.6      | 13.5    | 14.3      |
| macOS            |         |            |         |           |
| Windows          |         |            |         |           |

## Release (this pr)
| Device           | JpgLoad |  JpgResize | PngLoad | PngResize |
|------------------|---------|------------|---------|-----------|
| iPhone           |         |            |         |           |
| iOS Simulator    |         |            |         |           |
| Android          | 339.7   | 467.7      | 20.4    | 27.1      |
| Android Emulator | 91.6    | 116.7      | 9.7     | 10.2      |
| macOS            |         |            |         |           |
| Windows          |         |            |         |           |


## Release (3.1.8)
| Device           | JpgLoad |  JpgResize | PngLoad | PngResize |
|------------------|---------|------------|---------|-----------|
| iPhone           |         |            |         |           |
| iOS Simulator    |         |            |         |           |
| Android          | 290.2   | 393.0      | 19.7    | 24.0      |
| Android Emulator | 80.1    | 104.7      | 15.0    | 16.2      |
| macOS            |         |            |         |           |
| Windows          |         |            |         |           |

## Test devices


| Device           | Hardware             | OS              |
|------------------|----------------------|-----------------|
| iPhone           | iPhone 15 Pro Max    | 18.4.1          |
| iOS Simulator    | iPhone 15            | 18.4            |
| Android          | Pixel 2 XL           | Android 15      |
| Android Emulator | Pixel 3a             | Android 14      |
| macOS            | MacBook Pro M3 Max   | Sequoia 15.4.1  |
| Windows          | AMD 5950X + RTX 3060 | Windows 11 24H2 |

beeradmoore:
I did 3.1.4 in .NET 9, release mode on Android device and unfortunately it outperforms everything else here. It seems improvements we are seeing is from .NET 9 (or less likely Android 14 which got upgraded to Android 15) 
 

## Release - Android physical device
| Device           | JpgLoad |  JpgResize | PngLoad | PngResize |
|------------------|---------|------------|---------|-----------|
| 3.1.4 (.NET 8)   | 515.3   | 712.2      | 26.9    | 35.4      |
| 3.1.4 (.NET 9)   | 285.2   | 389.6      | 20.0    | 24.5      |
| 3.1.8 (.NET 9)   | 290.2   | 393.0      | 19.7    | 24.0      |
| This PR (.NET 9) | 339.7   | 467.7      | 20.4    | 27.1      |


As a mobile dev if I saw these numbers I'd be a little turned off ImageSharp (as a desktop dev I love it in [DLSS Swapper](https://github.com/beeradmoore/dlss-swapper)). I'd need to check against SkiaSharp and rolling my own native Android implementation (using Android classes in .NET) to see if this is a library performance issue or a Android chugs along issue.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: I'll have a look at creating an issue. Thanks.

JimBobSquarePants:
@tannergooding Thanks for the review, I've made all the changes based on your feedback. 

@beeradmoore Those numbers are.... disappointing, I'm not sure where the regression is there. I've pushed some changes based on the feedback to `MultiplyAdd` which should provide a speedup on the YCbCr conversion (_and others_) but otherwise the conversion code should be solid. 

I have, however found potential for some speedup before we hit the color conversion in `Block8x8F` where we are falling back to scalar rounding on non Avx platforms. I should be able to add SIMD there easily enough. I'll do that in a follow up PR though. 

If possible, could you run the benchmark again on this branch one last time before I merge? Regression or not the focus with Mono is heavily orientated on `Vector128` so performance can only improve there.

beeradmoore:
@JimBobSquarePants , updated results

## Debug (3.1.8)
| Device           | JpgLoad |  JpgResize | PngLoad | PngResize |
|------------------|---------|------------|---------|-----------|
| Android          | 1084.1  | 1312.4     | 37.1    | 44.4      |
| Android Emulator | 189.5   | 245.1      | 13.8    | 14.2      |

## Debug (this updated PR)
| Device           | JpgLoad |  JpgResize | PngLoad | PngResize |
|------------------|---------|------------|---------|-----------|
| Android          | 1344.77 | 1586.2     | 37.5    | 48.1      |
| Android Emulator | 233.3   | 285.8      | 13.8    | 15.6      |

## Release (3.1.8)
| Device           | JpgLoad |  JpgResize | PngLoad | PngResize |
|------------------|---------|------------|---------|-----------|
| Android          | 285.5   | 392.9      | 19.3    | 26.1      |
| Android Emulator | 83.7    | 96.2       | 9.4     | 9.9       |

## Release (this updated PR)
| Device           | JpgLoad |  JpgResize | PngLoad | PngResize |
|------------------|---------|------------|---------|-----------|
| Android          | 341.0   | 469.4      | 20.1    | 25.8      |
| Android Emulator | 99.3    | 121.1      | 9.2     | 10.4      |

JimBobSquarePants:
Thanks @beeradmoore 

I'm wondering now if there isn't some other regression elsewhere in the V4 branch. There is ample opportunity for more SIMD in `Block8x8F` though so will do that next.

beeradmoore:
Im going to take the work I did and throw it in a new app with benchmark.net so I can also test SkiaSharp as well as native implementations

So numbers here wont be comparable to those

Any other useful tests to add in?

For each image format (jpg, png, webp)
Load
Load + save 
Load + resize + save

JimBobSquarePants:
@beeradmoore That would be incredibly useful. Thank you!

Issue #2454, opened 2023-05-12
Jpeg compressed tiff with photometric interpretation Separated is not supported


shapeh:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

3.0.1

### Other ImageSharp packages and versions

None

### Environment (Operating system, version and so on)

Windows 11 and Windows Server 2022

### .NET Framework version

.NET 7.0.5

### Description

Hi, 

I am getting this error when trying to resize a tiff image:

**Jpeg compressed tiff with photometric interpretation Separated is not supported**

I looked at this: https://github.com/SixLabors/ImageSharp/issues/12 and maybe it is still relevant, despite edited 2021? 
_"Photometric Interpretation Formats:  Separated (TIFF Extension color spaces)" - not ticked._
Update: https://github.com/SixLabors/ImageSharp/blob/main/src/ImageSharp/Formats/Tiff/README.md - Separated seems to be supported?

Is there a plan for support at a later stage?

Thanks for a great library.

> System.Exception: System.NotSupportedException: Jpeg compressed tiff with photometric interpretation Separated is not supported
>    at SixLabors.ImageSharp.Formats.Tiff.TiffThrowHelper.ThrowNotSupported(String message)
>    at SixLabors.ImageSharp.Formats.Tiff.Compression.Decompressors.JpegTiffCompression.DecodeJpegData(BufferedReadStream stream, Span`1 buffer, CancellationToken cancellationToken)
>    at SixLabors.ImageSharp.Formats.Tiff.Compression.Decompressors.JpegTiffCompression.Decompress(BufferedReadStream stream, Int32 byteCount, Int32 stripHeight, Span`1 buffer, CancellationToken cancellationToken)
>    at SixLabors.ImageSharp.Formats.Tiff.Compression.TiffBaseDecompressor.Decompress(BufferedReadStream stream, UInt64 stripOffset, UInt64 stripByteCount, Int32 stripHeight, Span`1 buffer, CancellationToken cancellationToken)
[...]

PR #2922, opened 2025-05-16
Enable colorimetric normalization of JPEG image data during decode

### LINKED ISSUE #129: Color Changed When Resize
I don't know why :(

Upload image in Github, Gmail have the same issue 
Resize on Imagesharp have the same issue

But when i try resize using net452, use system.drawing, image color is good, not changed, it is oke

`ResizeOptions opt = new ResizeOptions();`
`opt.Mode = ResizeMode.Max;`
`opt.Size = size.Size;`
`image.Resize(opt).Save(outStream);`

# Original

image

# GitHub
bug_612

[bug_612.zip](https://github.com/JimBobSquarePants/ImageSharp/files/829796/bug_612.zip)

Sori if my eng is not good

### LINKED ISSUE #2723: File conversion from jpg to webp changes colors of the image
### Prerequisites

### ImageSharp version

3.1.4

none

Windows 11

### .NET Framework version

6 and 8

### Description

In our application we resize and change images from png and jpg to webp for better performance on the website. This works very good most of the times. But for some jpg images there is a distinct color change.
It might be related to #1567 but I am not sure, doesn't happen when just resizing the jpg image.

### Steps to Reproduce

` using var image = await Image.LoadAsync(file.OpenReadStream()); `
` using var outStream = new MemoryStream(); `
` await image.SaveAsync(outStream, new WebpEncoder()); `

Jpg image:
3_VMware Better Together

webp image looks like this (see webp in zip):
VMware Better Together

### Images

[VMware Better Together.zip](https://github.com/SixLabors/ImageSharp/files/15058593/VMware.Better.Together.zip)

JimBobSquarePants:
### Prerequisites

### Description

* Fixes #2723
* Fixes #129 - opened 8 years, 2 months, and 7 days ago
* Likely impacts #2454 (cc @brianpopow)

This PR adds support for normalizing color data in JPEG files (and TIFF files with JPEG compression) to **sRGB during decoding**.

A new `DecoderOptions.ColorProfileHandling` property controls this behavior with the following options:

* **Preserve** - *Leaves any embedded ICC color profiles intact.* (default, preserves existing behavior)
* **Compact**  - *Removes any embedded Standard sRGB ICC color profiles without transforming pixel data.*
* **Convert**  - *Transforms pixels based on embedded ICC profiles into the sRGB v4 color space, then removes the original profile.*

Several internal helper methods have been introduced to allow decoders to determine and apply the correct behavior.

All shared JPEG color converters now use a common scalar interleaving path to manipulate planar data for `ColorProfileConverter`. There's likely room for SIMD-based performance improvements here (AVX-512?).

@waacton - might be of interest to you since it introduces questions around deriving RGB working spaces from ICC profiles which you are likely to better understand than I.
@saucecontrol - tagging you since your compact sRGB v4 profile implementation was invaluable (and shamelessly borrowed 😁).


JimBobSquarePants:
I'm not sure what is going on with these failing tests?? 

Everything works on my machine.

JimBobSquarePants:
Definitely broke something! All my color!

image

waacton:
Happy to take a look

waacton:
(Sorry, misclicking!) Happy to take a look but will be in a couple of weeks after my break

src/ImageSharp/Metadata/Profiles/ICC/IccProfileHeader.cs
```diff
@@ -11,6 +11,17 @@ namespace SixLabors.ImageSharp.Metadata.Profiles.Icc;
 /// </summary>
 public sealed class IccProfileHeader
 {
+    private static readonly Vector3 TruncatedD50 = new(0.9642029F, 1F, 0.8249054F);
+
+    // sRGB v4 Preference
```
waacton: v2 😉

src/ImageSharp/Metadata/Profiles/ICC/IccProfileHeader.cs
```diff
@@ -97,4 +108,31 @@ public sealed class IccProfileHeader
...
+        }
+
+        // Accept known sRGB profile IDs
+        if (header.Id == StandardRgbV2 || header.Id == StandardRgbV4)
+        {
+            return true;
+        }
+
+        // Fallback: best-guess heuristic
+        return
```
waacton, line 129: This reminded me that profiles with absolute intent aren't currently handled correctly, but doesn't look like they throw exceptions either yet? (Not that I can find such a profile.)

Could be worth including intent in this check?

src/ImageSharp/Formats/DecoderOptions.cs
```diff
@@ -62,9 +64,46 @@ public sealed class DecoderOptions
...
     /// <summary>
     /// Gets a value that controls how ICC profiles are handled during decode.
-    /// TODO: Implement this.
     /// </summary>
-    internal ColorProfileHandling ColorProfileHandling { get; init; }
+    public ColorProfileHandling ColorProfileHandling { get; init; }
 
     internal void SetConfiguration(Configuration configuration) => this.configuration = configuration;
+
+    internal bool TryGetIccProfileForColorConversion(IccProfile? profile, [NotNullWhen(true)] out IccProfile? value)
```
waacton, line 72: This return bool isn't currently used, easier to return the `IccProfile?` itself?

waacton:
I've scanned over the ICC profile stuff, and if the goal is to detect an sRGB profile and bypass using it since sRGB calculations are already implemented programmatically as the default behaviour, then it makes sense to me.

I've also double-checked the sRGB profile IDs just to be sure 👍

src/ImageSharp/Metadata/Profiles/ICC/IccProfileHeader.cs
JimBobSquarePants: I think, in this case, it's not wise to introduce a none sRGB relevant check here. If we attempt to convert an example in the wild someone can raise an issue and we can complete the converter.

src/ImageSharp/Formats/DecoderOptions.cs
JimBobSquarePants: Not at the moment but I think it will be used in other formats.

JimBobSquarePants:
Thanks @waacton very much appreciated!

JimBobSquarePants:
Let's merge this!

PR #2933, opened 2025-06-04
Add ARM support to WEBP Utilities

### LINKED ISSUE #2125: [WEBP] SaveImage is taking more than x100 times slower in AWS Instance than my local machine
### Prerequisites

### ImageSharp version

2.1.1

None

ARM64/Ubuntu/k8s

### .NET Framework version

.NET 6

### Description

Hi guys,

I've observed increased processing time of Image in VM environment.
It's similar to this (https://github.com/SixLabors/ImageSharp/issues/2104) issue but slightly different.

Image service is running in k8s with an unlimited CPU budget and is supposed to process a lot of images concurrently.

Pod info
```
# uname -a
Linux image-service-8566cd55f6-f6hkb 5.4.188-104.359.amzn2.aarch64 #1 SMP Thu Apr 14 20:53:17 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```

My local test of an image save takes 16ms, but in VM takes 1,000ms to 40,000ms (depending on size, 40 seconds with ~1MB jpg)

### Steps to Reproduce

Code
```
    private async Task<Stream> ResizeImage(Stream stream)
    {
        using var image = await Image.LoadAsync(stream);
        var imageStream = new MemoryStream();

        await image.SaveAsync(imageStream, new WebpEncoder()
        {
            FileFormat = WebpFileFormatType.Lossy,
        });
        
        return imageStream;
    }
```

### Images

_No response_

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2125

Ports the `Sse` specific code to general `Vector128` equivalents within all WEBP utilities adding ARM support. 

@tannergooding I've tried my best to polyfil all the various methods in `Vector128Utilties` but I would really appreciate any guidance you can provide to improve them (and fix any bugs 😛)

### DIFF
--- a/src/ImageSharp/Metadata/Profiles/Exif/Tags/ExifTagValue.cs
+++ b/src/ImageSharp/Metadata/Profiles/Exif/Tags/ExifTagValue.cs
@@ -24,14 +24,6 @@ internal enum ExifTagValue
     GPSIFDOffset = 0x8825,
 
     /// <summary>
-    /// Indicates the identification of the Interoperability rule.
-    /// See https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/interoperability/interoperabilityindex.html
-    /// </summary>
-    [ExifTagDescription("R98", "Indicates a file conforming to R98 file specification of Recommended Exif Interoperability Rules (ExifR98) or to DCF basic file stipulated by Design Rule for Camera File System.")]
-    [ExifTagDescription("THM", "Indicates a file conforming to DCF thumbnail file stipulated by Design rule for Camera File System.")]
-    InteroperabilityIndex = 0x0001,
-
-    /// <summary>
     /// A general indication of the kind of data contained in this subfile.
     /// See Section 8: Baseline Fields.
     /// </summary>


src/ImageSharp/Common/Helpers/Vector256Utilities.cs
```diff
@@ -21,24 +20,6 @@ namespace SixLabors.ImageSharp.Common.Helpers;
 internal static class Vector256_
```
stefannikolei, line 20: This could profit from the new c# extension constructs.

JimBobSquarePants: Yeah, I'm interested in them. I'm hoping many of these methods will eventually make it into the runtime so I don't actually need them though.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -158,8 +222,43 @@ public static Vector128<byte> ShiftLeftBytesInVector(Vector128<byte> value, [Con
...
+            return Sse2.ShiftLeftLogical(value, count);
+        }
+
+        // Zero lanes where count >= 16 to match SSE2
+        if (count >= 16)
+        {
+            return Vector128<short>.Zero;
+        }
+
+        if (AdvSimd.IsSupported)
```
TechPizzaDev: These fallbacks seem to be unnecessary on NET8+. The `count >= 16` causes `value << count` to emit just a `vpsllw` without `and` masking.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -347,6 +509,944 @@ public static Vector128<short> PackSignedSaturate(Vector128<int> left, Vector128
...
+    /// </param>
+    /// <param name="right">
+    /// The second vector containing packed 16-bit integers to multiply.
+    /// </param>
+    /// <returns>
+    /// A vector containing the low 16 bits of the products of the packed 16-bit integers
+    /// from <paramref name="left"/> and <paramref name="right"/>.
+    /// </returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector128<short> MultiplyLow(Vector128<short> left, Vector128<short> right)
```
TechPizzaDev: `MultiplyLow` should be equivalent to `Vector128.Multiply` i.e. `left * right`.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -347,6 +509,944 @@ public static Vector128<short> PackSignedSaturate(Vector128<int> left, Vector128
...
+        {
+            return AdvSimd.SubtractSaturate(left, right);
+        }
+
+        if (PackedSimd.IsSupported)
+        {
+            return PackedSimd.SubtractSaturate(left, right);
+        }
+
+        // Widen inputs to 32-bit signed
```
TechPizzaDev, line 1047: Maybe overkill but I'd make these widen-saturate-narrow fallbacks into generic helpers (which may not be optimal on Mono) to reduce amount of duplicated code e.g. 
```cs
interface IBinaryOp<T> { 
    static abstract T Apply(T left, T right); 
}
struct SubtractOp : IBinaryOp<Vector128<T>> { 
    static abstract Vector128<T> Apply(Vector128<T> left, Vector128<T> right) => left - right;
}
// With further overloads for ushort@uint, byte@ushort, sbyte@short.
// Could be made generic over integers too if there was a generic Widen helper...
public static Vector128<short> SaturateOp<T>(
    Vector128<short> left, Vector128<short> right, Vector128<int> min, Vector128<int> max) 
    where T : IOp<Vector128<int>> 
{
    (Vector128<int> leftLo, Vector128<int> leftHi) = Vector128.Widen(left);
    (Vector128<int> rightLo, Vector128<int> rightHi) = Vector128.Widen(right);

    Vector128<int> lo = T.Apply(leftLo, rightLo);
    Vector128<int> hi = T.Apply(leftHi, rightHi);

    lo = Clamp(lo, min, max);
    hi = Clamp(hi, min, max);

    return Vector128.Narrow(lo, hi);
}
```

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -347,6 +509,944 @@ public static Vector128<short> PackSignedSaturate(Vector128<int> left, Vector128
...
+    /// The first vector containing packed signed 16-bit integers to subtract from.
+    /// </param>
+    /// <param name="right">
+    /// The second vector containing packed signed 16-bit integers to subtract.
+    /// </param>
+    /// <returns>
+    /// A vector containing the results of subtracting packed signed 16-bit integers
+    /// </returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector128<short> SubtractSaturate(Vector128<short> left, Vector128<short> right)
```
TechPizzaDev, line 1030: Add `#if NET10_0_OR_GREATER` guard in all narrowing and saturating methods to utilize https://github.com/dotnet/runtime/pull/115525.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -347,6 +509,944 @@ public static Vector128<short> PackSignedSaturate(Vector128<int> left, Vector128
...
+    /// </summary>
+    /// <param name="value">
+    /// The vector containing packed 8-bit integers from which to create the mask.
+    /// </param>
+    /// <returns>
+    /// A 16-bit integer mask where each bit corresponds to the most significant bit of each 8-bit element
+    /// in <paramref name="value"/>.
+    /// </returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static int MoveMask(Vector128<byte> value)
```
TechPizzaDev: Pains me to say this, seeing this nice port, but `Vector128.ExtractMostSignificantBits` is equivalent to `MoveMask` (and that JIT intrinsic can benefit from AVX512 masking).

src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
```diff
@@ -207,58 +204,54 @@ public static void ITransformOne(Span<byte> reference, Span<short> input, Span<b
...
-            int output0 = Sse2.ConvertToInt32(ref0.AsInt32());
-            int output1 = Sse2.ConvertToInt32(ref1.AsInt32());
-            int output2 = Sse2.ConvertToInt32(ref2.AsInt32());
-            int output3 = Sse2.ConvertToInt32(ref3.AsInt32());
+            int output0 = ref0.AsInt32().ToScalar();
+            int output1 = ref1.AsInt32().ToScalar();
+            int output2 = ref2.AsInt32().ToScalar();
+            int output3 = ref3.AsInt32().ToScalar();
 
             Unsafe.As<byte, int>(ref outputRef) = output0;
```
TechPizzaDev: These `Unsafe.As` could be replaced with `Unsafe.WriteUnaligned`.

src/ImageSharp/Formats/Webp/Lossy/Vp8Encoding.cs
```diff
@@ -207,58 +204,54 @@ public static void ITransformOne(Span<byte> reference, Span<short> input, Span<b
...
-            Vector128<byte> ref2 = Vector128<byte>.Zero;
-            Vector128<byte> ref3 = Vector128<byte>.Zero;
             ref byte referenceRef = ref MemoryMarshal.GetReference(reference);
 
             // Load four bytes/pixels per line.
-            ref0 = Sse2.ConvertScalarToVector128Int32(Unsafe.As<byte, int>(ref referenceRef)).AsByte();
-            ref1 = Sse2.ConvertScalarToVector128Int32(Unsafe.As<byte, int>(ref Unsafe.Add(ref referenceRef, WebpConstants.Bps))).AsByte();
-            ref2 = Sse2.ConvertScalarToVector128Int32(Unsafe.As<byte, int>(ref Unsafe.Add(ref referenceRef, WebpConstants.Bps * 2))).AsByte();
-            ref3 = Sse2.ConvertScalarToVector128Int32(Unsafe.As<byte, int>(ref Unsafe.Add(ref referenceRef, WebpConstants.Bps * 3))).AsByte();
+            Vector128<byte> ref0 = Vector128.CreateScalar(Unsafe.As<byte, int>(ref referenceRef)).AsByte();
```
TechPizzaDev: These `Unsafe.As` could be replaced with `Unsafe.ReadUnaligned`.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -25,43 +23,56 @@ internal static class Vector128_
...
 
-    /// <summary>
-    /// Gets a value indicating whether right align operations are supported.
-    /// </summary>
-    public static bool SupportsAlignRight
-    {
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        get => Ssse3.IsSupported || AdvSimd.IsSupported;
+        // Portable fallback: (a + b + 1) >> 1
+        return (left + right + Vector128.Create((byte)1)) >> 1;
```
tannergooding: This isn't quite the same.

`Sse2.Average` (pavg) and `AdvSimd.FusedAddRoundedHalving` (urhadd) are instead effectively doing:
```csharp
return Vector128.Narrow(
    (Vector128.WidenLower(left) + Vector128.WidenLower(right) + Vector128<ushort>.One) >> 1,
    (Vector128.WidenUpper(left) + Vector128.WidenUpper(right) + Vector128<ushort>.One) >> 1
);
```

Which is to say that, it accounts for a potential 9th bit to ensure a correct rounded result. This can be observed for `Average(255, 1)` which should produce `128` but where this fallback will produce `0`

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -25,43 +23,56 @@ internal static class Vector128_
...
+    /// <returns>The <see cref="Vector128{Single}"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector128<float> ShuffleNative(Vector128<float> vector, [ConstantExpected] byte control)
     {
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        get => Sse2.IsSupported || AdvSimd.IsSupported;
+        if (Sse.IsSupported)
+        {
+            return Sse.Shuffle(vector, vector, control);
+        }
```
tannergooding: This is probably not needed. The xplat call should do the "right thing" and will allow the JIT to use something like `Avx.Permute(vector, control)` instead if its a more efficient encoding

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -71,11 +82,11 @@ public static bool SupportsShiftByte
...
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public static Vector128<float> ShuffleNative(Vector128<float> vector, [ConstantExpected] byte control)
+    public static Vector128<int> ShuffleNative(Vector128<int> vector, [ConstantExpected] byte control)
     {
-        if (Sse.IsSupported)
+        if (Sse2.IsSupported)
         {
-            return Sse.Shuffle(vector, vector, control);
+            return Sse2.Shuffle(vector, control);
         }
```
tannergooding: Similar thing here. I'd expect just the xplat call is sufficient and the JIT will continue emitting the optimal shuffle instruction for that control byte.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -88,6 +99,60 @@ public static Vector128<float> ShuffleNative(Vector128<float> vector, [ConstantE
...
+        }
+
+        // Don't use InverseMMShuffle here as we want to avoid the cast.
+        Vector64<short> indices = Vector64.Create(
+            (short)(control & 0x3),
+            (short)((control >> 2) & 0x3),
+            (short)((control >> 4) & 0x3),
+            (short)((control >> 6) & 0x3));
+
+        return Vector128.Create(value.GetLower(), Vector64.Shuffle(value.GetUpper(), indices));
```
tannergooding: A better fallback would be to do this:
```suggestion
         Vector128<short> indices = Vector128.Create(
            0,
            1,
            2,
            3,
            (short)((control & 0x3) + 4),
            (short)(((control >> 2) & 0x3) + 4),
            (short)(((control >> 4) & 0x3) + 4),
            (short)(((control >> 6) & 0x3) + 4));

        return Vector128.Shuffle(value, indices));
```

The reason is that `V64` isn't accelerated on all platforms and you already have a `V128`, so this allows you to avoid decomposing the vector and doing more operations.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -88,6 +99,60 @@ public static Vector128<float> ShuffleNative(Vector128<float> vector, [ConstantE
...
+        }
+
+        // Don't use InverseMMShuffle here as we want to avoid the cast.
+        Vector64<short> indices = Vector64.Create(
+            (short)(control & 0x3),
+            (short)((control >> 2) & 0x3),
+            (short)((control >> 4) & 0x3),
+            (short)((control >> 6) & 0x3));
+
+        return Vector128.Create(Vector64.Shuffle(value.GetLower(), indices), value.GetUpper());
```
tannergooding: Similar comment
```suggestion
         Vector128<short> indices = Vector128.Create(
            (short)(control & 0x3),
            (short)((control >> 2) & 0x3),
            (short)((control >> 4) & 0x3),
            (short)((control >> 6) & 0x3),
            4,
            5,
            6,
            7);

        return Vector128.Shuffle(value, indices));
```

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -133,8 +198,7 @@ public static Vector128<byte> ShiftRightBytesInVector(Vector128<byte> value, [Co
             return AdvSimd.ExtractVector128(value, Vector128<byte>.Zero, numBytes);
         }
 
-        ThrowUnreachableException();
-        return default;
+        return Vector128.Shuffle(value, Vector128.Create((byte)0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15) + Vector128.Create(numBytes));
```
tannergooding, line 206: In .NET 9+ the `Vector128.Create((byte)0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15)` can be `Vector128<byte>.Indices`

The total `Vector128.Create((byte)0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15) + Vector128.Create(numBytes)` can itself become `Vector128.CreateSequence<byte>(start: numBytes, step: 1)`

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -158,8 +222,43 @@ public static Vector128<byte> ShiftLeftBytesInVector(Vector128<byte> value, [Con
 #pragma warning restore CA1857 // A constant is expected for the parameter
         }
 
-        ThrowUnreachableException();
-        return default;
+        return Vector128.Shuffle(value, Vector128.Create((byte)0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15) - Vector128.Create(numBytes));
```
tannergooding, line 230: Similar thing, in .NET 9 this can become `Vector128.CreateSequence<byte>(start: numBytes, step: unchecked((byte)(-1)))`

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
tannergooding: Right, I'd expect we can drop the `Sse2`, `AdvSimd`, and `PackedSimd` path and instead just do:
```csharp
if (count >= 16)
{
    return Vector128<short>.Zero;
}
return value << count;
```

Since `count` is expected to be constant, the `if (count >= 16)` path should typically be dropped and we just get `vpsllw` or the relevant platform specific shift emitted by the xplat operator.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -347,6 +509,944 @@ public static Vector128<short> PackSignedSaturate(Vector128<int> left, Vector128
...
+
+        if (AdvSimd.IsSupported)
+        {
+            Vector128<int> prodLo = AdvSimd.MultiplyWideningLower(left.GetLower(), right.GetLower());
+            Vector128<int> prodHi = AdvSimd.MultiplyWideningLower(left.GetUpper(), right.GetUpper());
+
+            Vector128<long> v0 = AdvSimd.AddPairwiseWidening(prodLo);
+            Vector128<long> v1 = AdvSimd.AddPairwiseWidening(prodHi);
+
+            return Vector128.Narrow(v0, v1);
```
tannergooding: Why not just `AdvSimd.Arm64.AddPairwise(prodLo, prodHi)`?

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
tannergooding: Yep, this can just be `left * right` in all cases

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -347,6 +509,944 @@ public static Vector128<short> PackSignedSaturate(Vector128<int> left, Vector128
...
+    /// </returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Vector128<short> MultiplyHigh(Vector128<short> left, Vector128<short> right)
+    {
+        if (Sse2.IsSupported)
+        {
+            return Sse2.MultiplyHigh(left, right);
+        }
+
+        // Widen each half of the short vectors into two int vectors
```
tannergooding: For Arm64, you can use `AdvSimd.MultiplyWideningLower/Upper`, then shift, then narrow

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
tannergooding: Yep, this can just be `value.ExtractMostSignificantBits()` which does all the same handling here, but which can sometimes be optimized by the JIT for special scenarios and on platforms without a native instruction.

src/ImageSharp/Common/Helpers/Vector256Utilities.cs
```diff
@@ -73,8 +46,7 @@ public static Vector256<byte> ShuffleNative(Vector256<byte> vector, Vector256<by
             return Avx2.Shuffle(vector, indices);
```
tannergooding, line 46: `indices` works differently between `Avx2` and `Vector256`

In particular, `Avx2` shuffles within 2x 128-bit lanes and so it expects `[0, 15]` for the lower and then `[0, 15]` for the upper. (it's effectively doing `V256.Create(V128.Shuffle(vector.GetLower(), indices.GetLower()), V128.Shuffle(vector.GetUpper(), indices.GetUpper())`

While `V256` shuffles within a single 256-bit lane and so it expects `[0, 31]` across the whole thing. The perf is then dependent on what the hardware supports and if the indices cross lanes at all. That is, it is fastest on all hardware if `indices.GetLower()` is only `[0, 15]` and `indices.GetUpper()` is only `[16, 31]`, just because AVX2 doesn't actually provide full-width byte shuffle; only AVX512VBMI capable hardware provides proper full width support.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: I'm only targeting .NET 8 just now. By the time I finally get around to targeting .NET 10, I'll drop 8 support and use the runtime directly.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: Ha! Fantastic!

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: How can it use Avx when we're targeting `Vector128` support only?

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: .NET 8 only unfortunately. Good to see a new API though, writing these is cumbersome and not great to read.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: I'll add that also. I want to be able to support as many Arm chipsets as possible.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
tannergooding: AVX isn't just 256-bit support. It includes a number of newer 128-bit instructions and a general new encoding as well. The same goes for AVX512, where it provides a new encoding and newer 128/256-bit instructions in addition to the 512-bit instruction support.

The JIT knows what the actual underlying hardware supports and will opportunistically light up for newer instruction sets where it will benefit perf or size. So it's already using some of these in various places where safe.

For NAOT or other scenarios where you aren't jitting it won't use these newer instructions or encodings, but will still allow optimizations where feasible. For example, certain `control` inputs may allow you to emit `Sse.UnpackLow` instead and save 1 byte of encoding space, just because they are equivalent in functionality for that particular `control`.

src/ImageSharp/Common/Helpers/Vector256Utilities.cs
JimBobSquarePants: Ah... That's interesting. We're relying on the Avx2 behavior currently. I'll have a look at the current calls to see if I can adjust the indices, though maybe I should rename this `ShufflePerLane` or `Shuffle128`

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: Nothing is actually using that... I must have added it in error.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: Given I'll be getting the runtime versions in the future I'll probably not bother.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: Ah but I only use this method where `Vector256.IsHardwareAccelrated` is false.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: I can't see `Sse2.Shuffle(vector, control)` for float?

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
tannergooding: That can still be AVX capable hardware. V256.IsHardwsreAccelerated requires AVX2

But even without that, the xplat api still gives more flexibility for the JIT to optimize so it’s often preferred where feasible

We tend to be very 1-to-1 for the platform specific APIs

JimBobSquarePants: OK, will update

src/ImageSharp/Common/Helpers/Vector256Utilities.cs
JimBobSquarePants: Went with ShufflePerLane for clarity.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -347,6 +504,865 @@ public static Vector128<short> PackSignedSaturate(Vector128<int> left, Vector128
...
+    {
+        if (Sse2.IsSupported)
+        {
+            return Sse2.MultiplyAddAdjacent(left, right);
+        }
+
+        if (AdvSimd.IsSupported)
+        {
+            Vector128<int> prodLo = AdvSimd.MultiplyWideningLower(left.GetLower(), right.GetLower());
+            Vector128<int> prodHi = AdvSimd.MultiplyWideningLower(left.GetUpper(), right.GetUpper());
```
TechPizzaDev:
```suggestion
            Vector128<int> prodHi = AdvSimd.MultiplyWideningUpper(left, right);
```

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
```diff
@@ -347,6 +504,865 @@ public static Vector128<short> PackSignedSaturate(Vector128<int> left, Vector128
...
+
+            if (AdvSimd.Arm64.IsSupported)
+            {
+                return AdvSimd.Arm64.AddPairwise(prodLo, prodHi);
+            }
+
+            Vector128<long> v0 = AdvSimd.AddPairwiseWidening(prodLo);
+            Vector128<long> v1 = AdvSimd.AddPairwiseWidening(prodHi);
+
+            return Vector128.Narrow(v0, v1);
```
TechPizzaDev: Based on sse2neon it is probably cheaper to not widen. This applies to other helpers too.
```suggestion
            Vector64<int> v0 = AdvSimd.AddPairwise(prodLo.GetLower(), prodLo.GetUpper());
            Vector64<int> v1 = AdvSimd.AddPairwise(prodHi.GetLower(), prodHi.GetUpper());

            return Vector128.Create(v0, v1);
```

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
tannergooding: Right. I was rather trying to say that I expect you can just have this be `Vector128.Shuffle` using the same general mechanism for creating the `Vector128` indices out of the `control`. It can have the same general benefits to codegen and makes it more portable.

src/ImageSharp/Common/Helpers/Vector128Utilities.cs
JimBobSquarePants: Ah yes, the docs specifically say widens to a 32 bit, whereas I was going all the way to 64bit.

JimBobSquarePants:
OK. I think I've got this. Thanks @tannergooding and @TechPizzaDev for the review, very much appreciated!!

PR #2935, opened 2025-06-04
Fix Typo and Offset in BufferedReadStream

stefannikolei:
* While checking ChunkedMemoryStream i saw differences in the Seek method of BufferedReadStream

### Prerequisites

### DIFF
--- a/src/ImageSharp/IO/BufferedReadStream.cs
+++ b/src/ImageSharp/IO/BufferedReadStream.cs
@@ -215,9 +215,9 @@ internal sealed class BufferedReadStream : Stream
     {
         this.Position = origin switch
         {
-            SeekOrigin.Begin => (int)offset,
-            SeekOrigin.Current => (int)(this.Position + offset),
-            SeekOrigin.End => (int)(this.Length + offset),
+            SeekOrigin.Begin => offset,
+            SeekOrigin.Current => this.Position + offset,
+            SeekOrigin.End => this.Length + offset,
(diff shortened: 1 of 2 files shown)


src/ImageSharp/IO/BufferedReadStream.cs
```diff
@@ -213,20 +213,13 @@ public override void Flush()
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public override long Seek(long offset, SeekOrigin origin)
     {
-        switch (origin)
+        this.Position = origin switch
```
JimBobSquarePants, line 216: Hmmm.... I think we should have done the opposite and removed any casting from the chunked stream. There's potential for overflow here.

stefannikolei: what do you mean with casting in this context?

Ok on the next line xD.

Gonna change it in both places

src/ImageSharp/IO/BufferedReadStream.cs
```diff
@@ -69,7 +69,7 @@ public BufferedReadStream(Configuration configuration, Stream stream, Cancellati
     }
 
     /// <summary>
-    /// Gets the number indicating the EOF hits occured while reading from this instance.
+    /// Gets the number indicating the EOF hits occurred while reading from this instance.
```
antonfirsov, line 72: Signature Anton grammar (TM) 😅

Issue #55, opened 2016-12-21
Support of 8-bit images


Sergey-Terekhin:
Does Image class support 8-bit grayscale images? I mean read, write and convert from other color depths

JimBobSquarePants:
Yeah, there's no reason why it couldn't Several of the image formats support reading and writing grayscale images. Storing the image in grayscale would be as simple as implementing an `IPackedPixel` format. 

It might be an idea to add a built in one actually.

paolofulgoni:
@JimBobSquarePants does the idea of including an 8-bit grayscale pixel format still make sense?

merijndejonge:
I'm new to ImageSharp. Could someone explain to me how I would implement IPackedPixel to support 8-bit grayscale?
Thanks a lot!

antonfirsov:
@merijndejonge as a start, I suggest to check out the [related piece of docs](https://docs.sixlabors.com/articles/ImageSharp/PixelFormats.html). (Note: `IPackedPixel` has been replaced with `IPixel<TSelf>`.)

After checking it out, feel free to ask further questions on our [gitlab channel](https://gitter.im/ImageSharp/General), you will get answers much faster there.

PR #2940, opened 2025-06-06
Reduce the number of memory allocations in lossless WebP encoder

### LINKED ISSUE #2934: Lossless WebP encoder allocates A LOT
### Prerequisites

### ImageSharp version

3.1.8

-

Windows 11 with latest updates

### .NET Framework version

.NET 9

### Description

Lossless WebP encoder allocates both a lot of small object and also a few big chunks, which are NOT reused.
It is about 500 K objects and 200-300 MB total memory to encode single UWQHD (3440x1440) image.
It leads to very high pressure on GC. Especially, if there are many images to process.

Me encoding settings are probably the lowest for lossless saving:
```csharp
WebpEncoder {
    FileFormat = WebpFileFormatType.Lossless,
    Quality = 0,
    TransparentColorMode = WebpTransparentColorMode.Preserve,
}
```

I found the following hot paths (for one image):
Image
Image

`PixOrCopy.CreateLiteral` produces TONS of small objects.
`Vp8LEncoder`'s constructor may not look that scary, but let's sequentially encode 10 images:
Image
Image

As you can see, those big chunks are not reused in the future.
I tried to provide a custom memory allocator. It didn't help.
I suppose the allocator is totally ignored there.

Fixing these two moments should greatly accelerate the WebP encoder:
1. `PixOrCopy.CreateLiteral` problem is probably the hardest one as it requires changing of data structures.
2. Changing `Vp8LEncoder` should be much easier.

### Steps to Reproduce

Just encode an image with performance profiles and object allocation tracking.

### Images

_No response_

SladeThe:
This PR closes #2934

The results of benchmark `#55 EncodeWebp`:

#### Before
| Method                     | Mean      | Error     | StdDev   | Gen0      | Gen1      | Gen2      | Allocated   |
|--------------------------- |----------:|----------:|---------:|----------:|----------:|----------:|------------:|
| 'ImageSharp Webp Lossless' | 105.31 ms | 24.582 ms | 1.347 ms | 2000.0000 | 1800.0000 | 1400.0000 | 20490.55 KB |

#### After
| Method                     | Mean      | Error     | StdDev   | Gen0      | Gen1      | Gen2      | Allocated   |
|--------------------------- |----------:|----------:|---------:|----------:|----------:|----------:|------------:|
| 'ImageSharp Webp Lossless' | 100.61 ms | 15.196 ms | 0.833 ms | 2000.0000 | 2000.0000 | 2000.0000 |  3904.68 KB |

#### Pin the refs (without the capacity guard)
| Method                     | Mean      | Error     | StdDev   | Gen0      | Gen1      | Gen2      | Allocated   |
|--------------------------- |----------:|----------:|---------:|----------:|----------:|----------:|------------:|
| 'ImageSharp Webp Lossless' |  97.43 ms |  9.375 ms | 0.514 ms |   2000.0000 | 2000.0000 | 2000.0000 |  3904.98 KB |

### DIFF
--- a/src/ImageSharp/Formats/Tiff/Ifd/EntryReader.cs
+++ b/src/ImageSharp/Formats/Tiff/Ifd/EntryReader.cs
@@ -14,7 +14,7 @@ internal class EntryReader : BaseExifReader
         : base(stream, allocator) =>
         this.IsBigEndian = byteOrder == ByteOrder.BigEndian;
 
-    public List<IExifValue> Values { get; } = new();
+    public List<IExifValue> Values { get; } = [];
 
     public ulong NextIfdOffset { get; private set; }
 
@@ -31,8 +31,6 @@ internal class EntryReader : BaseExifReader
         {
             this.ReadValues64(this.Values, ifdOffset);
             this.NextIfdOffset = this.ReadUInt64();
-
-            //// this.ReadSubIfd64(this.Values);
         }
     }
 
@@ -62,9 +60,9 @@ internal class HeaderReader : BaseExifReader
         {
             this.IsBigTiff = true;
 
-            ushort bytesize = this.ReadUInt16();
+            ushort byteSize = this.ReadUInt16();
             ushort reserve = this.ReadUInt16();
-            if (bytesize == TiffConstants.BigTiffByteSize && reserve == 0)
+            if (byteSize == TiffConstants.BigTiffByteSize && reserve == 0)
             {
                 this.FirstIfdOffset = this.ReadUInt64();
                 return;
--- a/src/ImageSharp/Formats/Tiff/TiffDecoderOptionsParser.cs
+++ b/src/ImageSharp/Formats/Tiff/TiffDecoderOptionsParser.cs
@@ -25,20 +25,22 @@ internal static class TiffDecoderOptionsParser
     /// <returns>True, if the image uses tiles. Otherwise the images has strip's.</returns>
     public static bool VerifyAndParse(this TiffDecoderCore options, ExifProfile exifProfile, TiffFrameMetadata frameMetadata)
     {
-        IExifValue extraSamplesExifValue = exifProfile.GetValueInternal(ExifTag.ExtraSamples);
-        if (extraSamplesExifValue is not null)
+        if (exifProfile.TryGetValue(ExifTag.ExtraSamples, out IExifValue<ushort[]> samples))
         {
-            short[] extraSamples = (short[])extraSamplesExifValue.GetValue();
-            if (extraSamples.Length != 1)
+            // We only support a single sample pertaining to alpha data.
+            // Other information is discarded.
+            TiffExtraSampleType sampleType = (TiffExtraSampleType)samples.Value[0];
+            if (sampleType is TiffExtraSampleType.CorelDrawUnassociatedAlphaData)
             {
-                TiffThrowHelper.ThrowNotSupported("ExtraSamples is only supported with one extra sample for alpha data.");
+                // According to libtiff, this CorelDRAW-specific value indicates unassociated alpha.
+                // Patch required for compatibility with malformed CorelDRAW-generated TIFFs.
+                // https://libtiff.gitlab.io/libtiff/releases/v3.9.0beta.html
+                sampleType = TiffExtraSampleType.UnassociatedAlphaData;
             }
 
-            TiffExtraSampleType extraSamplesType = (TiffExtraSampleType)extraSamples[0];
-            options.ExtraSamplesType = extraSamplesType;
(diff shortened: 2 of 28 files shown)


src/ImageSharp/Formats/Webp/Lossless/HistogramEncoder.cs
```diff
@@ -109,10 +109,9 @@ private static void HistogramBuild(
...
         int x = 0, y = 0;
         int histoXSize = LosslessUtils.SubSampleSize(xSize, histoBits);
-        using List<PixOrCopy>.Enumerator backwardRefsEnumerator = backwardRefs.Refs.GetEnumerator();
-        while (backwardRefsEnumerator.MoveNext())
+
+        foreach (PixOrCopy v in backwardRefs)
         {
-            PixOrCopy v = backwardRefsEnumerator.Current;
             int ix = ((y >> histoBits) * histoXSize) + (x >> histoBits);
             histograms[ix].AddSinglePixOrCopy(v, false);
```
JimBobSquarePants: This method could be updated to take the struct via `in` to avoid the copy.

src/ImageSharp/Formats/Webp/Lossless/Vp8LBackwardRefs.cs
```diff
@@ -1,21 +1,28 @@
...
+    private readonly IMemoryOwner<PixOrCopy> refs;
+    private int count;
+
+    public Vp8LBackwardRefs(MemoryAllocator memoryAllocator, int pixels)
+    {
+        this.refs = memoryAllocator.Allocate<PixOrCopy>(pixels);
+        this.count = 0;
+    }
+
+    public void Add(PixOrCopy pixOrCopy) => this.refs.Memory.Span[this.count++] = pixOrCopy;
```
JimBobSquarePants, line 20: I imaging resolving the span here is a strong candidate for any slowdown using pooled memory.

JimBobSquarePants (commented):
Just a quick read so far but looking good. I'll have to pull it down to review properly.

src/ImageSharp/Formats/Webp/Lossless/HistogramEncoder.cs
SladeThe: I've made the change. But if there is any performance difference, it's hard to notice. The structure is not that big.

src/ImageSharp/Formats/Webp/Lossless/Vp8LBackwardRefs.cs
antonfirsov: I'm getting the code right that `Vp8LBackwardRefs` is a temporary object? If yes, we could `Pin` the memory in the ctr, store the handle and the `PixOrCopy*` pointer and `Unpin` the handle in `Dispose`.

SladeThe: You are right. The resolving here is quite slow.
That's why I switched from index-based `refs` access to `foreach` in my [second commit](https://github.com/SixLabors/ImageSharp/pull/2940/commits/ff9d3a6223c72ca1c5931593e404ce0070db9d9c).

However, the whole change still makes the encoder almost 5% faster.
Also, the result is more stable and less affected by GC pauses (see the reduced Error value in my first post).
The `refs` are read-mostly, and I don't think that relatively slow `Add` is such a big problem.

SladeThe: I've pinned the `refs` and got even better results.
Added it to the top post.
Thank you for pointing it out.

antonfirsov: Wow, exposing `Span<PixOrCopy>.Enumerator` is neat, haven't noticed that!

> still makes the encoder almost 5% faster

In that case I agree that spending effort and increasing complexity to optimize `Add()` is not worth the trouble in this PR. We can profile and optimize the encoder later if we want.

SladeThe: I've just realised, the the pointer access became unsafe.

After storing the span capacity in a field and adding the guard to `Add`, the performance switched to about the same value as it was before the last commit.

Now I tend reverting the commit. At least, for code brevity.

antonfirsov (approved):
Looks good after skimming through all changes.

JimBobSquarePants:
I'm thinking of backporting this to V3. @antonfirsov what do you think?

antonfirsov:
The change looks simple and safe enough to backport.

JimBobSquarePants:
Thanks for this @SladeThe it's very much appreciated!!

PR #2943, opened 2025-06-11
V4 Use EXIF byte order for EXIF encoded strings.

### LINKED ISSUE #2906: EXIF UserComment tag writes Unicode text in UTF-16LE instead of UTF-16BE as specified in the standard
### Prerequisites

### ImageSharp version

3.1.7

None (only the main package is installed)

Windows 11

### .NET Framework version

9.0.200

### Description

When writing Unicode text to the EXIF UserComment tag, ImageSharp is using UTF-16LE encoding instead of UTF-16BE encoding as required by the EXIF specification.

According to the EXIF specification (JEITA CP-3451, section 4.6.5 "User Comments"), when using Unicode encoding, the first 8 bytes should be "UNICODE\0" followed by text encoded in UTF-16BE.

However, ImageSharp is storing the text in UTF-16LE, which causes the UserComment to be displayed incorrectly in many image viewers and editors.

The root cause of the issue is in the [ExifEncodedStringHelpers.cs file (lines 53-60)](https://github.com/SixLabors/ImageSharp/blob/e20e47f647ff20d0717785f84f9d73854ec5ef9d/src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs#L53-L60), where `Encoding.Unicode` is used which represents UTF-16LE. It should be using `Encoding.BigEndianUnicode` instead.

Additionally, the same issue exists when reading UserComment values. In [TryGetEncodedStringValue method](https://github.com/SixLabors/ImageSharp/blob/e20e47f647ff20d0717785f84f9d73854ec5ef9d/src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs#L93-L99), when detecting "UNICODE" encoding, it also uses `Encoding.Unicode` (UTF-16LE) instead of `Encoding.BigEndianUnicode` (UTF-16BE) to decode the value, which means ImageSharp cannot correctly read UserComment values that are properly encoded according to the EXIF specification.

This issue requires careful consideration for backward compatibility. Users who have been writing UserComment tags with previous versions of ImageSharp have data encoded in UTF-16LE format. If the fix simply switches to UTF-16BE encoding for both reading and writing, those existing images would have their UserComment values read incorrectly. A potential solution might involve detecting the byte order or providing migration options to ensure both existing and new data can be handled correctly.

### Steps to Reproduce

You can reproduce this issue with the following code, or check the complete reproduction repository at: https://github.com/nirvash/ImageSharpExifUserCommentEncodingBug

```
using SixLabors.ImageSharp;
using SixLabors.ImageSharp.Metadata.Profiles.Exif;

// Load an image
using var image = Image.Load("sample.jpg");

// Create EXIF profile if it doesn't exist
var exif = image.Metadata.ExifProfile ?? new ExifProfile();

// Set Unicode text in UserComment
exif.SetValue(ExifTag.UserComment, "Hello World! こんにちは世界");

// Apply EXIF profile to the image
image.Metadata.ExifProfile = exif;

// Save the image
image.Save("output.jpg");
```

When examining the UserComment value in the saved image file, you can see that the text is encoded in UTF-16LE instead of UTF-16BE.
When checking with a binary editor, after the "UNICODE\0" header, the letter "H" is stored as 48 00 instead of 00 48 (which is the UTF-16LE byte order).
According to the EXIF specification, it should be stored in UTF-16BE encoding with the byte order 00 48 for the letter "H".

### Images

[output.zip](https://github.com/user-attachments/files/19476719/output.zip)

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2906

When decoding Unicode-encoded text from EXIF, we should follow the specification's guidance and determine the correct decoding approach based on the following priority:

1. **Check for a Byte Order Mark (BOM):**
   If a BOM is present at the start of the encoded string, its endianness should override any other assumptions. This is explicitly allowed by the EXIF 2.2 and 2.3 specifications, even though later versions (e.g., CIPA DC-X008-2019) no longer mention it directly.

2. **Fallback to TIFF Byte Order:**
   If no BOM is present, the byte order of the Unicode text must match the byte order declared in the TIFF header. This was the originally intended behavior according to EXIF 2.2/2.3, and is still necessary for correctly decoding existing conformant files.

This change ensures that we're correctly interpreting the `UserComment`, `GPSProcessingMethod`, and `GPSAreaInformation` tags, all of which use the 8-byte encoding prefix scheme and may contain UTF-16 encoded text.

On encoding, we now honor the TIFF byte order for Unicode strings and optionally support emitting a BOM when desired.

This brings ImageSharp into alignment with the EXIF specification and improves compatibility with real-world image metadata.

In addition to this fix, I've also correctly reimplemented the equality operators for `ExifTag` and performed code quality cleanup to match standards.

### DIFF
--- a/src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
+++ b/src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
@@ -101,8 +101,8 @@ internal static class ExifEncodedStringHelpers
         return false;
     }
 
-    public static uint GetDataLength(EncodedString encodedString, ByteOrder order) =>
-        (uint)GetEncoding(encodedString.Code, order).GetByteCount(encodedString.Text) + CharacterCodeBytesLength;
+    public static uint GetDataLength(EncodedString encodedString) =>
+        (uint)GetEncoding(encodedString.Code, ByteOrder.LittleEndian).GetByteCount(encodedString.Text) + CharacterCodeBytesLength;
 
     public static int Write(EncodedString encodedString, Span<byte> destination)
     {
--- a/src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
+++ b/src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
@@ -16,13 +16,13 @@ internal static class ExifEncodedStringHelpers
     private const ulong UnicodeCode = 0x_45_44_4F_43_49_4E_55;
     private const ulong UndefinedCode = 0x_00_00_00_00_00_00_00_00;
 
-    private static ReadOnlySpan<byte> AsciiCodeBytes => new byte[] { 0x41, 0x53, 0x43, 0x49, 0x49, 0, 0, 0 };
+    private static ReadOnlySpan<byte> AsciiCodeBytes => [0x41, 0x53, 0x43, 0x49, 0x49, 0, 0, 0];
 
-    private static ReadOnlySpan<byte> JISCodeBytes => new byte[] { 0x4A, 0x49, 0x53, 0, 0, 0, 0, 0 };
+    private static ReadOnlySpan<byte> JISCodeBytes => [0x4A, 0x49, 0x53, 0, 0, 0, 0, 0];
 
-    private static ReadOnlySpan<byte> UnicodeCodeBytes => new byte[] { 0x55, 0x4E, 0x49, 0x43, 0x4F, 0x44, 0x45, 0 };
+    private static ReadOnlySpan<byte> UnicodeCodeBytes => [0x55, 0x4E, 0x49, 0x43, 0x4F, 0x44, 0x45, 0];
 
-    private static ReadOnlySpan<byte> UndefinedCodeBytes => new byte[] { 0, 0, 0, 0, 0, 0, 0, 0 };
+    private static ReadOnlySpan<byte> UndefinedCodeBytes => [0, 0, 0, 0, 0, 0, 0, 0];
 
     // 20932 EUC-JP Japanese (JIS 0208-1990 and 0212-1990)
     // https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding?view=net-6.0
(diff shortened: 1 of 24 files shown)


src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
```diff
@@ -50,37 +50,66 @@ private static Encoding JIS0208Encoding
...
+                    encodedString = new EncodedString(code, text);
+                    return true;
+                }
+            }
+            else
+            {
+                string text = GetEncoding(code, order).GetString(textBuffer);
+                encodedString = new EncodedString(code, text);
+                return true;
+            }
```
gfoidl: To avoid that code duplication:
```suggestion
            }
            
            string text = GetEncoding(code, order).GetString(textBuffer);
            encodedString = new EncodedString(code, text);
            return true;
```
The branches before will all `return`.

src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
```diff
@@ -50,37 +50,66 @@ private static Encoding JIS0208Encoding
...
         if (TryDetect(buffer, out CharacterCode code))
         {
-            string text = GetEncoding(code).GetString(buffer[CharacterCodeBytesLength..]);
-            encodedString = new EncodedString(code, text);
-            return true;
+            ReadOnlySpan<byte> textBuffer = buffer[CharacterCodeBytesLength..];
+            if (code == CharacterCode.Unicode && textBuffer.Length >= 2)
+            {
+                // Check BOM
+                if (textBuffer[0] == 0xFF && textBuffer[1] == 0xFE)
```
gfoidl: I don't know if it matters, and when not treat it just as informational comment :wink:

```suggestion
                if (textBuffer.StartsWith((ReadOnlySpan<byte>)[0xFF, 0xFE]))
```
This produces only 1 `cmp` as the JIT emits code that compares it as being a `short`.
But there still some room for improvement here in the JIT, as currently there's a memory load to the "right side" of the comparision, whilst that constand could be built and being just used.
Note: for `char` the JIT is already optimized in that way, only a piece is misisng for `byte` (there's an issue in runtime-repo for this which I don't find ATM).

Same below.

src/ImageSharp/Metadata/Profiles/Exif/Values/ExifShortArray.cs
```diff
@@ -41,9 +41,9 @@ public override bool TrySetValue(object? value)
 
         if (value is int signedInt)
         {
-            if (signedInt >= ushort.MinValue && signedInt <= ushort.MaxValue)
+            if (signedInt is >= ushort.MinValue and <= ushort.MaxValue)
```
gfoidl, line 44: I hesitated to comment about these constructs (also in the files above). For `{byte, ushort}.MinValue` is `0` so these could be optimized to 
```suggestion
            if ((uint)signedInt <= ushort.MaxValue)
```
but the JIT (in recent versions) already does this for us, so no need to do this.

When AOT and code size is taken into account a little bit could saved here by doing the `uint`-cast and IMO w/o lossing readability.

src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
JimBobSquarePants: Very cool!

src/ImageSharp/Metadata/Profiles/Exif/Values/ExifShortArray.cs
JimBobSquarePants: I'll leave these kinds of changes to a separate optimization pass. I'm keen to get the functionality in place for the major release and do passes when I have time.

src/ImageSharp/Metadata/Profiles/Exif/ExifEncodedStringHelpers.cs
JimBobSquarePants: Thanks, yes, this can be cleaned up. Have updated

PR #2951, opened 2025-06-21
Enforce Explicit Types and Target Type new

stefannikolei:
### Prerequisites

### Description
Reworkes #2855 after too Manz conflicts


stefannikolei:
Collection Expression is still a problem.. 
image
SA1515 enforces a new line before this comment.

Do we wanna enforce Collection Expressions? 

I think we have this options
a) Disable SA1515 (StyleCop Analyser seems unmaintened should be probably analysed and looked for alternatives..)
b) Insert the Blank Line
c) #pragma disable

JimBobSquarePants:
> Collection Expression is still a problem.. image SA1515 enforces a new line before this comment.
> 
> Do we wanna enforce Collection Expressions?
> 
> I think we have this options a) Disable SA1515 (StyleCop Analyser seems unmaintened should be probably analysed and looked for alternatives..) b) Insert the Blank Line c) #pragma disable

I think, in this case, moving the comment above the declaration would be the correct solution. As to whether we should enforce collection expressions.... I don't know, I like them very much but sometimes they can reduce readabilty.

stefannikolei:
OK Then let's go this way. 

This PR is complete from my side. When this one is merged I will open a draft for the Collection Expressions so we have a clear view how it is gonna be for this repo. 

Sounds good?

JimBobSquarePants (approved):
Thanks for doing this; much better!

src/ImageSharp/ColorProfiles/Hsl.cs
```diff
@@ -24,7 +24,7 @@ namespace SixLabors.ImageSharp.ColorProfiles;
     /// <param name="l">The l value (lightness) component.</param>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public Hsl(float h, float s, float l)
-        : this(new Vector3(h, s, l))
+        : this(new(h, s, l))
```
antonfirsov, line 27: @JimBobSquarePants I'm late to the party again, but I dislike this particular group of the changes. Now a code reader needs to navigate to the other method definition to see what type is being instantiated here, which makes it harder/slower to read the code. Same for `return new(...)` and similar expressions which remove type information completely from a given line.

stefannikolei: Yeah that's a drawback about this. I will create a PR  which tackles this. Rider/Resharper has more settings for this and was probably a bit more aggressive. Could you check that branch out and look what VS is doing? https://github.com/SixLabors/ImageSharp/pull/2958

antonfirsov: Thanks! #2958 looks good based on the diff, but I can't tell if it's missing something without pulling the code down and review offline.

JimBobSquarePants: Apologies, I missed this. According to [IDE0090](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0090) the analyzer should only be flagging when the type is apparent. Is Rider/Resharper buggy?

stefannikolei: Rider generates a warning at that rule stating that this one is unknown/unsupported. 

It has more granular editorconfig settings.

PR #2963, opened 2025-07-03
Add ToHex method to Argb32

lindexi:
### Prerequisites

- [ x I have written a descriptive pull-request title

### Description

Add ToHex method to Argb32.

The Argb32 will use the AARRGGBB format. Which format is what the WPF and WinUI and MAUI and Avalonia and so on used.

See https://github.com/SixLabors/ImageSharp/discussions/2962

### DIFF
--- a/src/ImageSharp/PixelFormats/PixelImplementations/Argb32.cs
+++ b/src/ImageSharp/PixelFormats/PixelImplementations/Argb32.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
 
+using System.Buffers.Binary;
+using System.Globalization;
 using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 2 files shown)


src/ImageSharp/PixelFormats/PixelImplementations/Argb32.cs
```diff
@@ -298,4 +368,46 @@ private static Argb32 Pack(Vector4 vector)
...
+        if (hex.Length is < 3 or > 4)
+        {
+            return null;
+        }
+
+        var (a, r, g, b) = hex.Length == 3
+            ? ('F', hex[0], hex[1], hex[2])
+            : (hex[0], hex[1], hex[2], hex[3]);
+
+        return new string(new[] { a, a, r, r, g, g, b, b });
```
gfoidl, line 411: This allocates more than needed. Instead the `string.Create`-overload could be used, where you can pass in a state and operate on the string-buffer directly, thus avoiding any intermediate allocation.

Although for .NET 10 the JIT will maybe stack-alloc the intermediate array allocation, but I'm not sure if this is actually the case here.

JimBobSquarePants:
@lindexi I've closed this in favor of #2964

Thanks for your help here.

lindexi:
@JimBobSquarePants OK

Issue #2983, opened 2025-09-15
Exception in LzwString.WriteTo when attempting to load a TIFF image


brianberns:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

3.1.11

### Other ImageSharp packages and versions

None

### Environment (Operating system, version and so on)

Windows 11

### .NET Framework version

.NET 8

### Description

While attempting to load a TIFF image found in the wild, an IndexOutOfRangeException occurs on line 87 in LzwString.WriteTo:

```csharp
    public int WriteTo(Span<byte> buffer, int offset)
    {
        if (this.Length == 0)
        {
            return 0;
        }

        if (this.Length == 1)
        {
            buffer[offset] = this.value;
            return 1;
        }

        LzwString e = this;
        int endIdx = this.Length - 1;
        if (endIdx >= buffer.Length)
        {
            TiffThrowHelper.ThrowImageFormatException("Error reading lzw compressed stream. Either pixel buffer to write to is to small or code length is invalid!");
        }

        for (int i = endIdx; i >= 0; i--)
        {
            buffer[offset + i] = e.value;   // <==== EXCEPTION OCCURS HERE ****
            e = e.previous;
        }

        return this.Length;
    }
```

At this point, `buffer` has a length of 7808, `offset` is 7750, and `i` is 58. Since 7750+58 = 7808, the attempt to write to the buffer is out of range. There appears to be an attempt to test for such an overflow a few lines above (`if (endIdx >= buffer.Length)`), but it does not take the `offset` into account.

The image opens successfully in other tools, such as the Windows Photo app and Photoshop Elements.

I noticed a very similar problem [described in a closed issue](https://github.com/SixLabors/ImageSharp/issues/2490), but this is either a different problem, or the original problem was not actually fixed.

### Steps to Reproduce

I'm attempting to get permission to share the TIFF image file. In the meantime, here is the stack trace:

```
 	System.Private.CoreLib.dll!System.ThrowHelper.ThrowIndexOutOfRangeException()	Unknown
[...]

PR #2986, opened 2025-09-17
When decoding a compressed TIFF, truncate a too-long string instead of throwing an exception

brianberns:
### Prerequisites

### Description
This is a proposed fix for https://github.com/SixLabors/ImageSharp/issues/2983. The idea is to truncate a too-long string, rather than throw an exception. This allows ImageSharp to successfully load some problematic TIFF files.

### DIFF
--- a/src/ImageSharp/Formats/Tiff/Compression/Decompressors/LzwString.cs
+++ b/src/ImageSharp/Formats/Tiff/Compression/Decompressors/LzwString.cs
@@ -77,9 +77,9 @@ public class LzwString
 
         LzwString e = this;
         int endIdx = this.Length - 1;
-        if (endIdx >= buffer.Length)
+        if (offset + endIdx >= buffer.Length)
         {
-            TiffThrowHelper.ThrowImageFormatException("Error reading lzw compressed stream. Either pixel buffer to write to is to small or code length is invalid!");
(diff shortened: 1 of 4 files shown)


brianpopow:
@brianberns Thanks for the contribution. Could you add one unit test for this case?

A test case would look like this:

```
[Theory]
[WithFile(Issue2983, PixelTypes.Rgba32)]
public void TiffDecoder_CanDecode_Issue2983<TPixel>(TestImageProvider<TPixel> provider)
    where TPixel : unmanaged, IPixel<TPixel> => TestTiffDecoder(provider);
```

This should go into Formats/Tiff/TiffDecoderTests.cs

Then you need to add the example image with the issue into tests\Images\Input\Tiff\Issues\Issue2983.tiff

The final thing you need to do is add the image to the Tiff section in TestImages.cs, for example in line 1167:
> public const string Issue2983 = "Tiff/Issues/Issue2983.tiff";

brianberns:
Done. I added the unit test and confirmed that it passes. However, I'm not very experienced with pull requests, so let me know if there are any problems.

JimBobSquarePants (approved):
These changes make sense to me. Thanks for your contribution. Much appreciated!

PR #2987, opened 2025-09-18
Enforce ICC D50 PCS illuminant when converting colours between profiles

waacton:
### Prerequisites

### Description
Following on from https://github.com/InternationalColorConsortium/iccDEV/issues/164, colour conversion between ICC profiles can be made more performant by ignoring the PCS illuminant in the header and using the ICC's definition of D50 - avoiding unnecessary chromatic adaptations.

The ICC reference implementation (now called iccDEV, not iccMAX) does not use a profile's PCS illuminant header field during colour conversion, it always uses XYZ `(0.9642, 1.0, 0.8249)`. Some related extracts from the v4 spec:

---

image

---

image

---

image

---

Only in the [latest proposed spec for v5+](https://color.org/specification/ICC.2-2023.pdf) does it suggest that non-D50 illuminants are supported, though the values should correspond to a new `spectralViewingConditions` tag.

So this PR aims to match the behaviour of the reference implementation:
- Assume all profiles are D50 illuminant (a different value is simply not compatible with the spec)
- Use ICC D50 `(0.9642, 1.0, 0.8249)` as the reference white of profiles, which is different than the existing ASTM D50 `(0.96422, 1, 0.82521)` ❗
- Reduce test tolerances now that ImageSharp and Unicolour are using the same D50 illuminant value for profile colour conversion

### DIFF
--- a/src/ImageSharp/Formats/Tiff/Compression/Decompressors/LzwString.cs
+++ b/src/ImageSharp/Formats/Tiff/Compression/Decompressors/LzwString.cs
@@ -69,25 +69,33 @@ public class LzwString
             return 0;
         }
 
-        if (this.Length == 1)
+        int available = buffer.Length - offset;
+        if (available <= 0)
         {
-            buffer[offset] = this.value;
-            return 1;
+            return 0;
+        }
+
+        int numToWrite = this.Length;
+        if (numToWrite > available)
+        {
+            numToWrite = available;
         }
 
         LzwString e = this;
-        int endIdx = this.Length - 1;
-        if (endIdx >= buffer.Length)
+
+        // if string is too long, skip bytes at the end
+        int toSkip = this.Length - numToWrite;
+        for (int i = 0; i < toSkip; i++)
         {
-            TiffThrowHelper.ThrowImageFormatException("Error reading lzw compressed stream. Either pixel buffer to write to is to small or code length is invalid!");
+            e = e.previous;
         }
 
-        for (int i = endIdx; i >= 0; i--)
+        for (int i = numToWrite - 1; i >= 0; i--)
         {
             buffer[offset + i] = e.value;
             e = e.previous;
         }
 
-        return this.Length;
+        return numToWrite;
     }
 }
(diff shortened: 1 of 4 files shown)


src/ImageSharp/ColorProfiles/ColorProfileConverterExtensionsIcc.cs
```diff
@@ -60,8 +60,8 @@ internal static TTo ConvertUsingIccProfile<TFrom, TTo>(this ColorProfileConverte
         ColorProfileConverter pcsConverter = new(new ColorConversionOptions
         {
             MemoryAllocator = converter.Options.MemoryAllocator,
-            SourceWhitePoint = new CieXyz(converter.Options.SourceIccProfile.Header.PcsIlluminant),
-            TargetWhitePoint = new CieXyz(converter.Options.TargetIccProfile.Header.PcsIlluminant),
+            SourceWhitePoint = KnownIlluminants.D50Icc,
```
JimBobSquarePants, line 63: Should we not be testing for a version number < 5 and assigning the value based on the version?

waacton: Potentially, though whether the ICC code in ImageSharp can support v5+ correctly at all is another question. I wouldn't be surprised if v5-level features were only added to the reference implementation after it was ported here (and now that the reference has been refactored, could be harder to match up). Adapting between PCS illuminants seems a sensible step but it might not meet v5 spec, given how wordy section `10.2.22 spectralViewingConditionsType` is for the `svcn` tag.

And their description on https://www.color.org/icclabs.xalter tries to emphasise how different v5 is:

> this new system is not a simple "dot" release to the V4 spec and that a V5 profile is very different than a V2 or V4

> in many industries, v4 (and even v2) meets existing color management needs and in these industries there will be no drive to move to adopt the v5 specification

I've still not got around to implementing full support for v4 profiles in Unicolour so I've largely buried my head in the sand for v5 🙈. I'm hedging my bets that because v4 took painstaking effort to meet, v5 surely can't be a smoother experience.

In summary I think it's a choice between:
- Explicitly no support for v5 until (or if) it becomes more widely adopted
- Best attempt to support v5 while acknowledging it might deviate from spec

JimBobSquarePants: Thanks for confirming. Yeah, definitely best to deal with V4 for now and V5 if and when it's ever needed.

PR #3008, opened 2025-10-29
Fix detection of canonical sRGB profiles

JimBobSquarePants:
### Prerequisites

### Description

See #2999 

This PR refines the detection of canonical sRGB ICC profiles in ImageSharp by introducing a robust method for identifying standard sRGB matrix+TRC profiles, replacing the previous heuristic approach. It also improves test coverage with new images and reference outputs for various ICC profile scenarios.

* Added a new method `IsCanonicalSrgbMatrixTrc()` to `IccProfile` that accurately detects canonical IEC 61966-2-1 (sRGB) matrix+TRC ICC profiles by checking profile IDs, required tags, and colorant/TRC values, including handling D65 white point adaptation. This method is more robust than the previous header-based heuristic.
* Changed logic in `DecoderOptions` to use `profile.IsCanonicalSrgbMatrixTrc()` instead of the old `IccProfileHeader.IsLikelySrgb()` for determining sRGB profiles, ensuring more accurate color management decisions.
* Removed the old heuristic `IsLikelySrgb` and related sRGB constants from `IccProfileHeader`, as detection is now handled in `IccProfile`.
* Made `IccProfile` a partial class to support the new logic extension.
* Added new test images and reference outputs for ICC profiles, including perceptual (matrix + cLUT  & cLUT-only) profiles and sRGB gray, to ensure the new detection logic is thoroughly validated. 
* Made the extension methods to `ColorProfileConverter` public.

### DIFF
--- a/src/ImageSharp/ColorProfiles/ColorProfileConverterExtensionsCieLabCieLab.cs
+++ b/src/ImageSharp/ColorProfiles/ColorProfileConverterExtensionsCieLabCieLab.cs
@@ -6,8 +6,24 @@ using SixLabors.ImageSharp.Memory;
 
 namespace SixLabors.ImageSharp.ColorProfiles;
 
-internal static class ColorProfileConverterExtensionsCieLabCieLab
+/// <summary>
+/// Allows conversion between two color profiles based on the CIE Lab color space.
+/// </summary>
+public static class ColorProfileConverterExtensionsCieLabCieLab
 {
+    /// <summary>
+    /// Converts a color value from one color profile to another using the specified color profile converter.
+    /// </summary>
+    /// <remarks>
+    /// The conversion process may use ICC profiles if available; otherwise, it performs a manual
+    /// conversion through the profile connection space (PCS) with chromatic adaptation as needed. The method requires
+    /// both source and target types to be value types implementing the appropriate color profile interface.
+    /// </remarks>
+    /// <typeparam name="TFrom">The source color profile type. Must implement <see cref="IColorProfile{TFrom, CieLab}"/>.</typeparam>
+    /// <typeparam name="TTo">The target color profile type. Must implement <see cref="IColorProfile{TTo, CieLab}"/>.</typeparam>
+    /// <param name="converter">The color profile converter to use for the conversion.</param>
+    /// <param name="source">The source color value to convert.</param>
+    /// <returns>A value of type <typeparamref name="TTo"/> representing the converted color in the target color profile.</returns>
     public static TTo Convert<TFrom, TTo>(this ColorProfileConverter converter, in TFrom source)
         where TFrom : struct, IColorProfile<TFrom, CieLab>
         where TTo : struct, IColorProfile<TTo, CieLab>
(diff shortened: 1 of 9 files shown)


src/ImageSharp/Metadata/Profiles/ICC/IccProfile.SRGB.cs
```diff
@@ -0,0 +1,346 @@
...
+        if (h.DataColorSpace != IccColorSpaceType.Rgb)
+        {
+            return false;
+        }
+
+        if (h.ProfileConnectionSpace != IccColorSpaceType.CieXyz)
+        {
+            return false;
+        }
+
```
waacton, line 64: The PCS for the standard v4 preference sRGB profile is LAB (not XYZ like v2) so presumably this won't detect v4 correctly 🤔

waacton: Ah, sorry, I misunderstood; this function will return true if the profile ID is for a known sRGB profile or it contains the same v2 XYZ + TRCs data. 

Checking v4 transform data would be far too intensive I assume?

src/ImageSharp/Metadata/Profiles/ICC/IccProfile.SRGB.cs
```diff
@@ -0,0 +1,346 @@
...
+        {
+            return true;
+        }
+
+        // If the profile declares a D65 white, adapt the colorant columns to D50 and compare again.
+        // We never adapt when they already match, to avoid compounding rounding.
+        if (Near(wtpt, KnownIlluminants.D65.AsVector3Unsafe(), 2e-3F))
+        {
+            CieXyz fromWp = new(wtpt);          // Declared white
+            CieXyz toWp = KnownIlluminants.D50; // PCS white
```
waacton, line 140: I would say this should be `KnownIlluminants.D50Icc` which is what the ICC converters assume the white point to be.

This also matches the `wtpt` tags of both v2 and v4 sRGB profiles (0.96420, 1.0000, 0.82491) - but as a reminder are ignored in the ICC reference implementation and are assumed to be (0.9642, 1, 0.8249).

waacton:
@JimBobSquarePants I admire the attempt to figure out what a profile is _supposed_ to do even if it doesn't quite meet the actual ICC spec, it's a nice touch 🙌

src/ImageSharp/Metadata/Profiles/ICC/IccProfile.SRGB.cs
JimBobSquarePants: I don’t know tbh. I’m trying to keep this as lean as I can but accurate enough to avoid expensive color transformation when I don’t need to. I check for the canonical v4 id before this anyway so at least I can capture those.

src/ImageSharp/Metadata/Profiles/ICC/IccProfile.SRGB.cs
JimBobSquarePants: The reason I didn’t use the ICC one was because I then compare the results to known chromaticies (from Bruce Lindbloom) I would expect the ICC D50 to yield different ones.

JimBobSquarePants:
> @JimBobSquarePants I admire the attempt to figure out what a profile is _supposed_ to do even if it doesn't quite meet the actual ICC spec, it's a nice touch 🙌

Thanks for the review @waacton !!

I’m just trying to save cycles on decode. 🙂

src/ImageSharp/Metadata/Profiles/ICC/IccProfile.SRGB.cs
waacton: I didn't realise non-canonical sRGB profiles were a thing until now anyway, but you'd think that V2 has been round longer so most likely to be been misused or misunderstood over time

src/ImageSharp/Metadata/Profiles/ICC/IccProfile.SRGB.cs
waacton: Good point, especially since these values aren't used downstream as part of the transform itself

PR #3013, opened 2025-11-14
Fix errors reported by rider

stefannikolei:
- Used collection expressions were possible
- Added the used type were needed for clarification
- Disabled an Resharper rule With ImageSharp.sln.DotSettings --> It said that the Attribute ConstantExpected is not valid on floats
- Removed unused TestHelper
- Rewrote some tests to use utf8 string instead of raw byte array to make the test clearer


tests/ImageSharp.Tests/Image/ImageTests.WrapMemory.cs
```diff
@@ -570,7 +570,6 @@ public unsafe void WrapMemory_Pointer_Null(int size, int height, int width)
         }
 
         private static bool ShouldSkipBitmapTest =>
-            !TestEnvironment.Is64BitProcess || (TestHelpers.ImageSharpBuiltAgainst != "netcoreapp3.1" &&
-                                                TestHelpers.ImageSharpBuiltAgainst != "netcoreapp2.1");
+            !TestEnvironment.Is64BitProcess;
```
stefannikolei: This now gives a test error. It seems that those tests were now always skipped..

JimBobSquarePants: The Boolean result is not different here for 64bit builds

stefannikolei: it is different for 64bit builds. Having the second part, then it evaluates to true. Without the second part it evaluates to false

JimBobSquarePants: Ha! You’re right! 

I wonder what’s wrong there then. Will investigate locally when I can.

PR #3014, opened 2025-11-14
Add .NET10.0 as TFM

stefannikolei:


tests/ImageSharp.Tests/Metadata/Profiles/ICC/DataWriter/IccDataWriterPrimitivesTests.cs
```diff
@@ -42,7 +42,7 @@ public void WriteAsciiStringWithNullWritesEmpty()
         byte[] output = writer.GetData();
 
         Assert.Equal(0, count);
-        Assert.Equal([], output);
+        Assert.Equal(Array.Empty<byte>(), output);
```
JimBobSquarePants, line 45: We shouldn’t need to do this. Might be something to do with breaking changes re span overloads though.

src/ImageSharp/ImageSharp.csproj
```diff
@@ -30,7 +30,7 @@
   <Choose>
     <When Condition="$(SIXLABORS_TESTING_PREVIEW) == true">
       <PropertyGroup>
-        <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
+        <TargetFrameworks>net8.0;net9.0;net10.0</TargetFrameworks>
```
JimBobSquarePants: I would just drop 9. I don’t want to extend build times and use up LFS budgets

tests/ImageSharp.Tests/Metadata/Profiles/ICC/DataWriter/IccDataWriterPrimitivesTests.cs
stefannikolei: Yeah that came from new overloads

stefannikolei:
I got some problems on my machine.. One with libgdiplus and all the tests that use System.Drawing as a reference. 

The other is a Tiff test. -->  https://github.com/SixLabors/ImageSharp/issues/3017 When I use a tolerant comparer the test is green.

Now I am only struggling with the FeatureTestRunner. Now going to look into it

stefannikolei:
> I got some problems on my machine.. One with libgdiplus and all the tests that use System.Drawing as a reference.
> 
> The other is a Tiff test. --> #3017 When I use a tolerant comparer the test is green.
> 
> Now I am only struggling with the FeatureTestRunner. Now going to look into it

Tried to debug it on my side. On my machine DOTNET_EnableArm64AdvSimd=0 seems to not work. And on CI/CD we have a problem with DisableSSE .... 

strange...

stefannikolei:
I made a small repro and created an issue for this in the runtime repo

https://github.com/dotnet/runtime/issues/121685

stefannikolei:
> I made a small repro and created an issue for this in the runtime repo
> 
> [dotnet/runtime#121685](https://github.com/dotnet/runtime/issues/121685)

@JimBobSquarePants I got an answer there. The grouping changed and you can now only disable groups and not the individual ones..

What is your favorite way to go on with this? Introduce many #if NET10_0_OR_GREATER in the test code. Or disable the groups for net8 and net10?

tests/ImageSharp.Tests/Formats/Jpg/Block8x8FTests.cs
```diff
@@ -267,7 +267,7 @@ static void RunTest(string srcSeedSerialized, string qtSeedSerialized)
             RunTest,
             srcSeed,
             qtSeed,
-            HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX | HwIntrinsics.DisableSSE);
+            HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX | HwIntrinsics.DisableSSE42);
```
JimBobSquarePants: This should be `DisableHWIntrinsic` not `DisableSSE42` since we actually want to disable all intrinsics and we now use `AdvSimd` and `PackedSimd` in our V128 code.

tests/ImageSharp.Tests/Formats/Jpg/DCTTests.cs
```diff
@@ -152,7 +152,7 @@ static void RunTest(string serialized)
             FeatureTestRunner.RunWithHwIntrinsicsFeature(
                 RunTest,
                 seed,
-                HwIntrinsics.AllowAll | HwIntrinsics.DisableFMA | HwIntrinsics.DisableAVX | HwIntrinsics.DisableHWIntrinsic);
+                HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX2 | HwIntrinsics.DisableAVX | HwIntrinsics.DisableHWIntrinsic);
```
JimBobSquarePants: I actually think we can remove `HwIntrinsics.DisableFMA` here. None of the code tested in this method uses it.

tests/ImageSharp.Tests/Formats/Jpg/DCTTests.cs
```diff
@@ -360,7 +360,7 @@ static void RunTest(string serialized)
...
```
JimBobSquarePants: Same here. It's all V256 and operators or AVX.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
```diff
@@ -19,7 +19,7 @@ public class JpegColorConverterTests
 
     private const int TestBufferLength = 40;
 
-    private const HwIntrinsics IntrinsicsConfig = HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX512F | HwIntrinsics.DisableAVX2;
+    private const HwIntrinsics IntrinsicsConfig = HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX512 | HwIntrinsics.DisableAVX2;
```
JimBobSquarePants: This isn't actually used.

tests/ImageSharp.Tests/Formats/Jpg/JpegColorConverterTests.cs
```diff
@@ -73,7 +73,7 @@ public void GetConverterReturnsCorrectConverterWithRgbColorSpace()
     {
         FeatureTestRunner.RunWithHwIntrinsicsFeature(
             RunTest,
-            HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX512F | HwIntrinsics.DisableAVX2 | HwIntrinsics.DisableSSE2 | HwIntrinsics.DisableHWIntrinsic);
+            HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX512 | HwIntrinsics.DisableAVX2 | HwIntrinsics.DisableSSE42 | HwIntrinsics.DisableHWIntrinsic);
```
JimBobSquarePants: Given that all the V256 converters use either V256 or Avx we can probably simplify this and others to:

```c#
HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX512 | HwIntrinsics.DisableAVX | HwIntrinsics.DisableHWIntrinsic
```

tests/ImageSharp.Tests/Formats/WebP/ColorSpaceTransformUtilsTests.cs
```diff
@@ -71,7 +71,7 @@ private static void RunCollectColorRedTransformsTest()
     public void CollectColorBlueTransforms_WithHardwareIntrinsics_Works() => FeatureTestRunner.RunWithHwIntrinsicsFeature(RunCollectColorBlueTransformsTest, HwIntrinsics.AllowAll);
 
     [Fact]
-    public void CollectColorBlueTransforms_WithoutVector128_Works() => FeatureTestRunner.RunWithHwIntrinsicsFeature(RunCollectColorBlueTransformsTest, HwIntrinsics.DisableSSE41);
+    public void CollectColorBlueTransforms_WithoutVector128_Works() => FeatureTestRunner.RunWithHwIntrinsicsFeature(RunCollectColorBlueTransformsTest, HwIntrinsics.DisableSSE42);
```
JimBobSquarePants: The entries in this class for `WithoutVector128_Works` tests should be `HwIntrinsics.DisableHWIntrinsic`

tests/ImageSharp.Tests/Processing/Processors/Convolution/BokehBlurTest.cs
```diff
@@ -148,7 +148,7 @@ public void BokehBlurFilterProcessor_WorksWithAllPixelTypes<TPixel>(TestImagePro
 
     [Theory]
     [WithFileCollection(nameof(TestFiles), nameof(BokehBlurValues), PixelTypes.Rgba32, HwIntrinsics.AllowAll)]
-    [WithFileCollection(nameof(TestFiles), nameof(BokehBlurValues), PixelTypes.Rgba32, HwIntrinsics.DisableSSE41)]
+    [WithFileCollection(nameof(TestFiles), nameof(BokehBlurValues), PixelTypes.Rgba32, HwIntrinsics.DisableSSE42)]
```
JimBobSquarePants: In tests like this the step would be to disable intrinsics entirely. You need to choose a value that works for both .NET8 and .NET 10. Currently your change will change .NET 10 only any Sse41 code will still run on .NET8

stefannikolei: Yeah I just went the way to go the Next higher Group and wait for your Feedback 😅

tests/ImageSharp.Tests/Common/SimdUtilsTests.cs
```diff
@@ -133,7 +133,7 @@ static void RunTest(string serialized) => TestImpl_BulkConvertByteToNormalizedFl
         FeatureTestRunner.RunWithHwIntrinsicsFeature(
             RunTest,
             count,
-            HwIntrinsics.AllowAll | HwIntrinsics.DisableAVX512F | HwIntrinsics.DisableAVX2 | HwIntrinsics.DisableSSE41);
```
stefannikolei, line 136: This change is causing problems

stefannikolei: @JimBobSquarePants It would be cool If you could look over this one. When I change DisableSSE42 to DisableHWIntrinsics I will get test failures

JimBobSquarePants: That suggests the scalar path is broken. I’ll have a look asap

JimBobSquarePants: In this case we want to drop `HwIntrinsics.DisableSSE41` entirely. This is because we are explicitly testing the `HwIntrinsics` version of the method and disabling intrinsics disables the entire `Vector128` path. 

Our previous test flags ensured we were covering both `Sse41` and `Sse2` on Windows but I think our cross-platform test will ensure coverage in this case since we're using `Vector128<T>` directly.

src/ImageSharp/ImageSharp.csproj
```diff
@@ -30,7 +30,7 @@
   <Choose>
     <When Condition="$(SIXLABORS_TESTING_PREVIEW) == true">
       <PropertyGroup>
-        <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
```
stefannikolei, line 33: @JimBobSquarePants Should we add .net10.0 to the "Otherwise" section too? So the next release would be build against both LTS versions of .NET?

JimBobSquarePants: No, just 8. I want to always target a single framework

silkfire:
@stefannikolei @JimBobSquarePants When can we expect this to be released?

stefannikolei:
> @stefannikolei @JimBobSquarePants When can we expect this to be released?

You can use it today. Nothing stops you. This PR just enables testing against net10. You can always use a lib which targets a lower tfm. Just not the other way

silkfire:
I'm just concerned that the current version of ImageSharp targets .NET 6 which is EOL. A bump to 8 or 10 would be appreciated.

tocsoft:
@silkfire that's not how nuget packages work the Target version of a package is the minimum version of the runtime we are compatible with and nothing to do with what version we run on (which is what the MS supported version stuff is about) that's based entirely on the version you target to compile/deploy your application against.

In conclusion targets in nuget pacakges are about api compatibility only. Microsofts supported version of dotnet is about where you target/run your app and unrelated to this.

This is true for all nuget packages btw not just ours.

PR #3015, opened 2025-11-15
Convert to .slnx

stefannikolei:
Converts the Solution File to the new cleaner Format.


JimBobSquarePants:
Slnx requires .NET 10 for stable support does it not?

stefannikolei:
> Slnx requires .NET 10 for stable support does it not?

Yeah... I could install the .net10 sdk for the .net8 build. Or just throw this PR away until .net8 is dropped

gfoidl:
> Slnx requires .NET 10 for stable support does it not?

.NET SDK 9.0.200 introduced it.

In .NET 10 only [`dotnet new sln`](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/10.0/dotnet-new-sln-slnx-default) defaults to the slnx format.

JimBobSquarePants:
> > Slnx requires .NET 10 for stable support does it not?
> 
> .NET SDK 9.0.200 introduced it.
> 
> In .NET 10 only [`dotnet new sln`](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/10.0/dotnet-new-sln-slnx-default) defaults to the slnx format.

Yes but in our CI we explicitly use only the single .NET 8 SDK for non-preview entries. I need to ensure everything still works there. 

Tests are currently failing though because `TestEnvironment` explicitly looks for the `sln` extension as it searches for the solution directory. If everything works after that I’m happy as long as VS 2022 can open it.

stefannikolei:
Closed for now. .net8 target should only be built with installed .net8 sdk and not with net8 and net10 sdk

Issue #3017, opened 2025-11-15
Test failing on ARM64 Mac


stefannikolei:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

main branch

### Other ImageSharp packages and versions

-

### Environment (Operating system, version and so on)

Mac OS Tahoe

### .NET Framework version

.NET8

### Description

     These tests failed#
    TiffDecoder_CanDecode_YccK<Rgba32>(provider: Tiff/Issues/Issue2454_A.tif[Rgba32])
    TiffDecoder_CanDecode_YccK<Rgba32>(provider: Tiff/Issues/Issue2454_B.tif[Rgba32])

They run with the compare setting --> Exact.

But I get:
 Image difference is over threshold!
Test Environment OS : MacOS
Test Environment is CI : False
Test Environment is .NET Core : True
Test Environment is Mono : False
Test Environment OS Architecture : Arm64
Test Environment Process Architecture : Arm64

<img width="737" height="436" alt="Image" src="https://github.com/user-attachments/assets/2a4fa793-9a64-44c0-96ab-948f6f5d092a" />

### Steps to Reproduce

-

JimBobSquarePants:
Can you share the output in a zip so I can do a visual comparison? TIFF Ycck will be using the JPEG decoder underneath which will be running the `Vector128` color conversion code. Could just be a minor difference that we can’t control.

stefannikolei:
[Tiff1.zip](https://github.com/user-attachments/files/23566566/Tiff1.zip)
[Tiff2.zip](https://github.com/user-attachments/files/23566565/Tiff2.zip)

JimBobSquarePants:
Are you able to debug this on your Mac? I won't be able to see what color conversion code is used. As I recall It should be the ICC path on `TiffYccKVector128`.

stefannikolei:
> Are you able to debug this on your Mac? I won't be able to see what color conversion code is used. As I recall It should be the ICC path on `TiffYccKVector128`. 

Will have a look at it.

stefannikolei:
[...]

PR #3018, opened 2025-11-16
Add ARM hosted runners

stefannikolei:
Added macos-26 and ubuntu-2204 github-runners. Those run on ARM. With this this architecture can be tested on every run and not just on demand like before.

### DIFF
--- a/src/ImageSharp/Formats/Webp/WebpAnimationDecoder.cs
+++ b/src/ImageSharp/Formats/Webp/WebpAnimationDecoder.cs
@@ -112,12 +112,12 @@ internal class WebpAnimationDecoder : IDisposable
         this.webpMetadata = this.metadata.GetWebpMetadata();
         this.webpMetadata.RepeatCount = features.AnimationLoopCount;
 
-        Color backgroundColor = this.backgroundColorHandling == BackgroundColorHandling.Ignore
+        this.webpMetadata.BackgroundColor = this.backgroundColorHandling == BackgroundColorHandling.Ignore
             ? Color.Transparent
             : features.AnimationBackgroundColor!.Value;
 
-        this.webpMetadata.BackgroundColor = backgroundColor;
-
+        bool ignoreMetadata = this.skipMetadata;
+        SegmentIntegrityHandling segmentIntegrityHandling = this.segmentIntegrityHandling;
         Span<byte> buffer = stackalloc byte[4];
         uint frameCount = 0;
         int remainingBytes = (int)completeDataSize;
@@ -135,9 +135,16 @@ internal class WebpAnimationDecoder : IDisposable
 
                     remainingBytes -= (int)dataSize;
                     break;
+                case WebpChunkType.Iccp:
                 case WebpChunkType.Xmp:
                 case WebpChunkType.Exif:
-                    WebpChunkParsingUtils.ParseOptionalChunks(stream, chunkType, this.metadata, this.skipMetadata, this.segmentIntegrityHandling, buffer);
+                    WebpChunkParsingUtils.ParseOptionalChunks(
+                        stream,
+                        chunkType,
+                        this.metadata,
+                        ignoreMetadata,
+                        segmentIntegrityHandling,
+                        buffer);
                     break;
                 default:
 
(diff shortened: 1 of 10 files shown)


tests/ImageSharp.Tests/Formats/Bmp/BmpDecoderTests.cs
```diff
@@ -144,6 +149,11 @@ public void BmpDecoder_CanDecode_4Bit<TPixel>(TestImageProvider<TPixel> provider
     public void BmpDecoder_CanDecode_8Bit<TPixel>(TestImageProvider<TPixel> provider)
         where TPixel : unmanaged, IPixel<TPixel>
     {
+        if (TestEnvironment.IsMacOS && TestEnvironment.IsArm64Process)
```
JimBobSquarePants: I don’t think we should be disabling tests. We’ve been running on ARM for a while

stefannikolei: That's for missing libgdiplus

JimBobSquarePants: Isn't that because the build script restricts installation to ubuntu?

https://github.com/SixLabors/ImageSharp/pull/3018/files#diff-bc668a2c9f2299cef15b222055b4b4d5311646caec2e7610e540cee18ae9b948R131

stefannikolei: Mac does not have apt. You must use homebrew. I can try it out. I haven't because I could not get mono-libgdiplus working on my machine 🙈

JimBobSquarePants: Ah yes, seems well supported though.

https://formulae.brew.sh/formula/mono-libgdiplus

stefannikolei: seems supported. but it seems that it is not that easy. Tests are failing.

stefannikolei:
@JimBobSquarePants this is Reader in my eyes. 

Would just be good if you could Check the Tiff Test. When this is merged I would rebase the net10 branch

JimBobSquarePants:
I just downloaded two zip files you shared before, but I've lost the page I downloaded them from?

image

It looks to me like one is not using the ICC profile?

stefannikolei:
> I just downloaded two zip files you shared before, but I've lost the page I downloaded them from?
> 
> 
> 
> image
> 
> 
> 
> It looks to me like one is not using the ICC profile?
> 
> 

https://github.com/SixLabors/ImageSharp/issues/3017

stefannikolei:
> I just downloaded two zip files you shared before, but I've lost the page I downloaded them from?
> 
> image
> It looks to me like one is not using the ICC profile?

@JimBobSquarePants I posted an update on this in the issue.

tests/ImageSharp.Tests/Formats/Tiff/TiffDecoderTests.cs
```diff
@@ -360,7 +360,7 @@ public void TiffDecoder_CanDecode_YccK<TPixel>(TestImageProvider<TPixel> provide
     {
         using Image<TPixel> image = provider.GetImage(TiffDecoder.Instance);
         image.DebugSave(provider);
-        image.CompareToReferenceOutput(ImageComparer.Exact, provider);
+        image.CompareToReferenceOutput(ImageComparer.Tolerant(), provider);
```
JimBobSquarePants: Use the same comparer as `TiffDecoder_CanDecode_YccK_ICC` here. I think the default one might be too relaxed.

.github/workflows/build-and-test.yml
```diff
@@ -124,6 +135,18 @@ jobs:
           sudo apt-get update
           sudo apt-get -y install libgdiplus libgif-dev libglib2.0-dev libcairo2-dev libtiff-dev libexif-dev
 
+      - name: Install libgdi+, which is required for tests running on macos
+        if: ${{ contains(matrix.options.os, 'macos-26') }}
```
JimBobSquarePants, line 139: I am SO glad you worked this out. I wouldn't have had a clue!!

stefannikolei: Keep this Discussion in mind https://github.com/SixLabors/ImageSharp/discussions/3016 when you pulled the trigger on your mini ;) That helped on my local machine

stefannikolei:
> Good job!!

@JimBobSquarePants You could now cancel your Subscription of buildjet and remove that from the build

PR #3023, opened 2025-11-21
Remove Nullable Disable from ICC Code

stefannikolei:
Removed the nullable disable from the icc Code (And one Tiff). Most parts could be solved relative easily. There are also some parts were the bang operator had to be used.

### Prerequisites

### DIFF
--- a/src/ImageSharp/ColorProfiles/Icc/IccConverterBase.Checks.cs
+++ b/src/ImageSharp/ColorProfiles/Icc/IccConverterBase.Checks.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
 
+using System.Diagnostics.CodeAnalysis;
 using SixLabors.ImageSharp.Metadata.Profiles.Icc;
 
 namespace SixLabors.ImageSharp.ColorProfiles.Conversion.Icc;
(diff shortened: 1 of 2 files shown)


src/ImageSharp/ColorProfiles/Icc/Calculators/CurveCalculator.cs
```diff
@@ -41,7 +40,7 @@ public float Calculate(float value)
         {
             CalculationType.Identity => value,
             CalculationType.Gamma => MathF.Pow(value, this.gamma), // TODO: This could be optimized using a LUT. See SrgbCompanding
-            CalculationType.Lut => this.lutCalculator.Calculate(value),
+            CalculationType.Lut => this.lutCalculator!.Calculate(value),
```
JimBobSquarePants: We’ll need to figure out a better pattern than this as it masks out potential bugs.

src/ImageSharp/ColorProfiles/Icc/Calculators/GrayTrcCalculator.cs
```diff
@@ -11,8 +11,8 @@ internal class GrayTrcCalculator : IVector4Calculator
 {
     private readonly TrcCalculator calculator;
 
-    public GrayTrcCalculator(IccTagDataEntry grayTrc, bool toPcs)
-        => this.calculator = new TrcCalculator(new IccTagDataEntry[] { grayTrc }, !toPcs);
+    public GrayTrcCalculator(IccTagDataEntry? grayTrc, bool toPcs)
+        => this.calculator = new TrcCalculator([grayTrc], !toPcs);
```
JimBobSquarePants, line 15: It’s odd that the DataEntry parameter here is nullable. Especially since we then populate the array with it.

stefannikolei: Ah... I thought I can just change it. But still have some errors in rider due to the other pr... 
The reason why this is nullable lies in the IccConverterBase.Checks --> GetTag. This could return null

JimBobSquarePants: The trick there is to turn it into TryGet…

stefannikolei: If you want I can change that to tryget if you like. ATM I am throwing an exception for the case that Array.Find Return null

JimBobSquarePants: The trick with nullability is always changing the mindset. You basically always want the result pattern for safety. TryGet in the sync world is the best approach for our requirements.

stefannikolei: Yeah. I Followed the Easy Road. BTW started to Look at nullability in jpeg. That's gonna be a beast

PR #3024, opened 2025-11-22
Remove Nullable disable from Png Code

stefannikolei:
I added a new Property to PngChunk --> DataSpan. On all places Data was used the same way. So we have the guard on this place and can just use it.

### Prerequisites

### DIFF
--- a/src/ImageSharp/Formats/Png/PngChunk.cs
+++ b/src/ImageSharp/Formats/Png/PngChunk.cs
@@ -1,8 +1,8 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
-#nullable disable
 
 using System.Buffers;
+using SixLabors.ImageSharp.Memory;
 
 namespace SixLabors.ImageSharp.Formats.Png;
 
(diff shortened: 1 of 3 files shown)


src/ImageSharp/Formats/Png/PngChunk.cs
```diff
@@ -36,7 +36,16 @@ public PngChunk(int length, PngChunkType type, IMemoryOwner<byte> data = null)
     /// Gets the data bytes appropriate to the chunk type, if any.
     /// This field can be of zero length or null.
     /// </summary>
-    public IMemoryOwner<byte> Data { get; }
+    public IMemoryOwner<byte>? Data { get; }
+
+    public ReadOnlySpan<byte> DataSpan
```
stefannikolei, line 41: Do you have a better idea to to this? Or should I check for null at all consuming places?

JimBobSquarePants:
@stefannikolei I'm going to revisit nullability post v4 launch. I want to put some thought to each leftover instance.

stefannikolei:
Let me know. I am happy to help. 

If you want you can create issues and assign some to me 😄

Issue #3025, opened 2025-11-24
ColorProfileHandling.Convert is not converting PNG files


Socolin:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp version

4.0.0-alpha.0.55

### Other ImageSharp packages and versions

Drawing: 3.0.0-alpha.0.6

### Environment (Operating system, version and so on)

Ubuntu 24.04

### .NET Framework version

.net 9

### Description

When loading a PNG image, the ICC Profile does not seems to be converted correctly. 

It seems to be something missing in the PNGDecoder since I was able to manually get the right result by applying the profile manually

Also even if the profile is not applied right now, setting `ColorProfileHandling = ColorProfileHandling.Convert` remove the profile from the image for a png

Note: I think having a `.Mutate(x => x.DrawImageAndConvertIccProfile())` could be a nice addition to make it easier to manage collage of multiple images using various ICC profile without losing quality by comparing everything to sRGB first (or the DrawImage could have an option to convert to the destination image color profile) Should I create another issue for this ?

### Steps to Reproduce

For images from https://displaycal.net/icc-color-management-test/
```cs
using var pngTestImage1 = Image.Load(new DecoderOptions {ColorProfileHandling = ColorProfileHandling.Convert}, "icc-tests/ICC Rendering Intent Test.png");
using var pngTestImage2 = Image.Load(new DecoderOptions {ColorProfileHandling = ColorProfileHandling.Convert}, "icc-tests/ICC Rendering Intent Test (cLUT only).png");
using var pngTestImage3 = Image.Load(new DecoderOptions {ColorProfileHandling = ColorProfileHandling.Convert}, "icc-tests/sRGB_Gray.png");
using var testCanvas1 = new Image<Rgba32>(pngTestImage1.Width, pngTestImage1.Height);
testCanvas1.Mutate(x => x.DrawImage(pngTestImage1, 1.0f));
testCanvas1.Save("icc-test-1.png");

using var testCanvas2 = new Image<Rgba32>(pngTestImage2.Width, pngTestImage2.Height);
[...]

PR #3028, opened 2025-12-05
Apply ICC Profile / Color Conversion when decoding PNG

Socolin:
### Prerequisites

### Description

I took a look at the code to see how to implements #3025 and I ends up writting a bit of code, so I'm sharing (There is not test yet, I'm not sure this is the right way to do this), if you are interested in this, let me know and I can continue the PR and add some tests

First I tried to do it the same way it's done in the `JpegDecoder`, when the pixel data is read. But this seems to be impossible with the PngDecoder since the Chunk containing the ICC profile might be after the image data (It was on my tests images)
Since the decoder is processing a stream (and the stream might not be rewindable) I abandon this idea (the only solution I see would be to store the bytes in memory to be able to find the profile first, but I think this would badly impact the performance / memory usage)

So I ends up doing the conversion at the end of the `PngDecoder.Decode` once all the chunk have been read and the image is already process. I'm using `TPixel.ToScaledVector4()` and `TPixel.FromScaledVector4()` to get the value directly as float for the profile conversion, and to avoid loosing precision if the image is 16 bit depth and TPixel is `Rgba64`

Let me know if this is the correct approach or if I missed something (all comments are welcome)

### DIFF
--- a/src/ImageSharp/ColorProfiles/Rgb.cs
+++ b/src/ImageSharp/ColorProfiles/Rgb.cs
@@ -100,17 +100,6 @@ public readonly struct Rgb : IProfileConnectingSpace<Rgb, CieXyz>
     public Vector4 ToScaledVector4()
         => new(this.AsVector3Unsafe(), 1F);
 
-    /// <summary>
-    /// Expands the color into a generic ("scaled") <see cref="Vector4"/> representation
-    /// with values scaled and usually clamped between <value>0</value> and <value>1</value>.
-    /// The vector components are typically expanded in least to greatest significance order.
-    /// </summary>
-    /// <param name="alpha">The alpha component.</param>
-    /// <returns>The <see cref="Vector4"/>.</returns>
-    [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    public Vector4 ToScaledVector4(float alpha)
-        => new(this.AsVector3Unsafe(), 1F);
-
     /// <inheritdoc/>
     public static void ToScaledVector4(ReadOnlySpan<Rgb> source, Span<Vector4> destination)
     {


src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -2153,4 +2161,58 @@ private static bool IsXmpTextData(ReadOnlySpan<byte> keywordBytes)
 
     private void SwapScanlineBuffers()
         => (this.scanline, this.previousScanline) = (this.previousScanline, this.scanline);
+
+    // FIXME: Maybe this could be a .Mutate(x => x.ApplyIccProfile(destinationProfile)) ? Nothing related to png here
+    private static void ApplyIccProfile<TPixel>(Image<TPixel> image, IccProfile sourceProfile, IccProfile destinationProfile)
```
JimBobSquarePants: I wonder if we could move this to an extension method for the converter itself?

We'd have to rename it also to include the name `RgbaCompatible` since it would not work for other color spaces.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -2153,4 +2161,58 @@ private static bool IsXmpTextData(ReadOnlySpan<byte> keywordBytes)
...
+    {
+        ColorConversionOptions options = new()
+        {
+            SourceIccProfile = sourceProfile,
+            TargetIccProfile = destinationProfile,
+        };
+
+        ColorProfileConverter converter = new(options);
+
+        image.ProcessPixelRows(pixelAccessor =>
```
JimBobSquarePants: I would use `ProcessPixelRowsAsVector4` passing `PixelConversionModifiers.Scale` then you get the SIMD optimized conversion.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -2153,4 +2161,58 @@ private static bool IsXmpTextData(ReadOnlySpan<byte> keywordBytes)
...
+            {
+                using IMemoryOwner<float> rgbBuffer = image.Configuration.MemoryAllocator.Allocate<float>(pixelAccessor.Width * 3);
+                using IMemoryOwner<float> alphaBuffer = image.Configuration.MemoryAllocator.Allocate<float>(pixelAccessor.Width);
+                Span<float> rgbPacked = rgbBuffer.Memory.Span;
+                ref float rgbPackedRef = ref MemoryMarshal.GetReference(rgbPacked);
+                Span<float> alphaPacked = alphaBuffer.Memory.Span;
+                ref float alphaPackedRef = ref MemoryMarshal.GetReference(alphaPacked);
+
+                for (int y = 0; y < pixelAccessor.Height; y++)
```
JimBobSquarePants: Once you use `ProcessPixelRowsAsVector4` this becomes a lot easier.

- Allocate a single `Rgb` buffer and populate via the bulk `Rgb.FromScaledVector4` method.
- Once converted loop the `Rgb` and `Vector4` buffers and use `Rgb.AsVector3Unsafe` method to gather the parts. You can then use `Unsafe.As` to copy the `Vector3` to the `Vector` buffer at the correct offset.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
Socolin: Yes I think too, and from what I saw, Webp will need the same method. Should we add `TFrom` and `TTo` as generic parameter ?

JimBobSquarePants: No generic parameters should be used since it must be compatible with Rgb to work since we're converting from/to the rgba representative `Vector4` - ICC profiles are very particular about this. Rgb and Grayscale both use Rgb profiles but others do not. 

WebP _shouldn't_ need the same method since the specification requires ICC to be before Color information and WebP has very strict conformance. 

https://developers.google.com/speed/webp/docs/riff_container#color_profile

However, we could use this as a stopgap to a more optimized solution there. 

I eventually might add a more optimized version for PNG for specification conformant images and use this as a fallback for those that don't.

The `iCCP` chunk is supposed to be before `iDAT`

https://www.w3.org/TR/png/#5ChunkOrdering

Socolin: I kind of like the possibility to apply the ICC profile during decoding if available and use this as a fallback. If I find the time I'll take a look at this.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
Socolin: ProcessPixelRowsAsVector4 seems to be parralellized, I'll need to allocate the buffer in each iteration instead of reusing one for all the row. Is that ok ?

Socolin: Well, I didn't though C# would allow such memory manipulation :) good to know ! It brings memory of C code, I'm writting some test and I'll push the change

Socolin: I'm getting an error with the paralellism in test, I fixed it with a `lock()` but it could be fixed with `Interlocked.Decrement` / `Increment` The method `MemoryDiagnostics_MemoryReleased` and `MemoryDiagnostics_MemoryAllocated` are called from various threads.

But even with the fix, I wonder if the allocators are all thread safe ?

```
Xunit.Sdk.TrueException
Expected 0 undisposed buffers but found -11
   at SixLabors.ImageSharp.Tests.MemoryAllocatorValidator.TestMemoryDiagnostics.Validate(Int32 expectedAllocationCount) in /home/socolin/projects/imagesharp/ImageSharp/tests/ImageSharp.Tests/MemoryAllocatorValidator.cs:line 61
   at SixLabors.ImageSharp.Tests.MemoryAllocatorValidator.ValidateAllocations(Int32 expectedAllocationCount) in /home/socolin/projects/imagesharp/ImageSharp/tests/ImageSharp.Tests/MemoryAllocatorValidator.cs:line 49
   at SixLabors.ImageSharp.Tests.ValidateDisposedMemoryAllocationsAttribute.After(MethodInfo methodUnderTest) in /home/socolin/projects/imagesharp/ImageSharp/tests/ImageSharp.Tests/ValidateDisposedMemoryAllocationsAttribute.cs:line 27
   at Xunit.Sdk.XunitTestInvoker.<>c__DisplayClass6_0.<AfterTestMethodInvokedAsync>b__1() in /_/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestInvoker.cs:line 93
   at Xunit.Sdk.ExecutionTimer.Aggregate(Action action) in /_/src/xunit.execution/Sdk/Frameworks/ExecutionTimer.cs:line 31
   at Xunit.Sdk.XunitTestInvoker.<>c__DisplayClass6_0.<AfterTestMethodInvokedAsync>b__0() in /_/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestInvoker.cs:line 93
   at Xunit.Sdk.ExceptionAggregator.Run(Action code) in /_/src/xunit.core/Sdk/ExceptionAggregator.cs:line 73
```

JimBobSquarePants: Somethings not getting disposed then. I'll have a look, we should never need to lock.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -2153,4 +2162,39 @@ private static bool IsXmpTextData(ReadOnlySpan<byte> keywordBytes)
...
+
+        ColorProfileConverter converter = new(options);
+
+        image.Mutate(o => o.ProcessPixelRowsAsVector4((pixelsRow, _) =>
+                {
+                    using IMemoryOwner<Rgb> rgbBuffer = image.Configuration.MemoryAllocator.Allocate<Rgb>(pixelsRow.Length);
+                    Span<Rgb> rgbPacked = rgbBuffer.Memory.Span;
+                    ref Rgb rgbPackedRef = ref MemoryMarshal.GetReference(rgbPacked);
+
+                    for (int x = 0; x < pixelsRow.Length; x++)
```
JimBobSquarePants: There's a bulk overload for this.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
```diff
@@ -2153,4 +2162,39 @@ private static bool IsXmpTextData(ReadOnlySpan<byte> keywordBytes)
...
+    {
+        ColorConversionOptions options = new()
+        {
+            SourceIccProfile = sourceProfile,
+            TargetIccProfile = destinationProfile,
+        };
+
+        ColorProfileConverter converter = new(options);
+
+        image.Mutate(o => o.ProcessPixelRowsAsVector4((pixelsRow, _) =>
```
JimBobSquarePants: You need to pass `PixelConversionModifiers.Scale` as the last parameter here or pixel formats that do not use [0..1] ranges will produce odd results.

tests/ImageSharp.Tests/MemoryAllocatorValidator.cs
```diff
@@ -22,7 +22,10 @@ private static void MemoryDiagnostics_MemoryReleased()
         TestMemoryDiagnostics backing = LocalInstance.Value;
         if (backing != null)
         {
-            backing.TotalRemainingAllocated--;
+            lock (backing)
```
JimBobSquarePants: I wonder if moving the color converter inside the loop action removes this requirement. 

We process in chunks so not concerned about initialization overhead.

JimBobSquarePants: Actually, I would do that anyway so that `MemoryAllocator` options property can be set using the same one as the image configuration

src/ImageSharp/Formats/Png/PngDecoderCore.cs
Socolin: > Somethings not getting disposed then. I'll have a look, we should never need to lock.

I assumed it was a concurrency bug because the counter was negative. I'll let you check this.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
Socolin: Fixed

src/ImageSharp/Formats/Png/PngDecoderCore.cs
Socolin: Fixed

tests/ImageSharp.Tests/MemoryAllocatorValidator.cs
Socolin: Oh, there is an hiden allocator in the `ColorConversionOptions`, yeah it should probably use the same as the image configuration (I'll change that)

Moving it inside the loop does not fix the problem

src/ImageSharp/Formats/Png/PngDecoderCore.cs
Socolin: I started adding the conversion during the decode process, it's done for now in `ProcessInterlacedRgbaScanline`. 

If you have some time, I'll appreciate some feedback on this before I continue the other one. Basically I'm duplicating the code and add a `if (iccProfile != null) / else` and adjust the code with the iccProfile. Is that correct ?

JimBobSquarePants:
@Socolin I'll have another look at this soon. I've been wiped out with the flu for the last 10 days so have not been able to do anything. I'll likely pull down your fork and make some initial code changes to add the appropriate extension method to `ColorProfileConverter` to allow both processes.

tests/ImageSharp.Tests/MemoryAllocatorValidator.cs
JimBobSquarePants: I had a good dig around and it's almost certainly a data-race. I've updated the validator to use atomic `Interlocked` ops.

JimBobSquarePants:
@Socolin I have re-implemented the per-image version of the operation in `ImageDecoderCore`.

I also took a closer look at the per-scanline variants, and I am not convinced they are a net win in this case. The color conversion must occur after any blending, which already forces a separation between decoding and conversion. That means we do not get to fuse the work into a single pass anyway, and we explicitly want to avoid per-pixel conversion because the ICC path is computationally expensive and not structured for that granularity.

The per-image approach has a practical advantage here: the conversion is applied over the entire pixel buffer using `ProcessPixelRowsAsVector4`, which internally partitions the work into `n = Environment.ProcessorCount` parallel chunks. Each chunk processes multiple rows, amortizing allocator usage, reducing per-row setup overhead, and allowing the ICC conversion to run over contiguous spans with better cache locality. In contrast, a per-scanline API would serialize more of that overhead. Even if scanlines were scheduled in parallel, each invocation would still pay the fixed costs of buffer setup, conversion dispatch, and synchronization at a much finer granularity, which tends to dominate the actual math.

So while a per-scanline API looks attractive on the surface, in practice it would be slower for this workload. The per-image conversion aligns better with how the ICC code wants to operate, and it lets us fully exploit coarse-grained parallelism rather than fragmenting the work into many small units with higher overhead.

src/ImageSharp/Formats/Png/PngDecoderCore.cs
JimBobSquarePants: Fixed via the validator changes

JimBobSquarePants (approved):
I'm happy with this now. Thanks @Socolin for your help!!

PR #3056, opened 2026-02-03
Add accumulative allocation tracking to allocators

### LINKED ISSUE #2850: Apply `AllocationLimitMegabytes` against the total size of multi-frame images before attempting to decode them
### Discussed in https://github.com/SixLabors/ImageSharp/discussions/2849

<sup>Originally posted by **antonfirsov** December 17, 2024</sup>
```C#
Configuration.Default.MemoryAllocator = MemoryAllocator.Create(new MemoryAllocatorOptions()
{
    AllocationLimitMegabytes = 16
});

// Should throw InvalidMemoryOperationException when the total size of all frames to allocate for this image are over 16MB, even if the individual frame sizes are below 16MB.
var img = Image.Load("multiframe.gif");
```

Even though strictly speaking this would be a hack since the allocation limit is originally meant to be per-buffer, I think this is something users want us to do. For built-in decoders, this is possible to implement by adding an internal validation method to`MemoryAllocator`; we can figure out later whether we want to make the feature available to external decoders.</div>

JimBobSquarePants:
### Prerequisites

### Description

Fixes #2850 

This pull request introduces a new accumulative allocation tracking system to the `MemoryAllocator` infrastructure, allowing for a configurable cap on the total memory allocated over the allocator's lifetime. It adds new configuration options, implements tracking logic, and updates all relevant allocators to enforce and release this new accumulative allocation limit. Other minor improvements include enhanced documentation and some code style refinements.

**Accumulative Allocation Tracking and Limits**

* Added an accumulative allocation limit to `MemoryAllocator`, configurable via the new `AccumulativeAllocationLimitMegabytes` property in `MemoryAllocatorOptions`. This limit is enforced across all allocations, with platform-specific defaults (2GB for 32-bit, 8GB for 64-bit processes). [[1]](diffhunk://#diff-e553b6f86fd8badcfc94a9677f26f7024109f3e54f4ee3cb7eede6a1b260fdf7L26-R62) [[2]](diffhunk://#diff-c77aeda122d81a14076b61136286a245dfcb5054f46ec5eae7f8327900734420R50-R74)
* Implemented logic in `MemoryAllocator` to track total allocated bytes, reserve/release allocation, suppress tracking during batch operations, and wrap allocations with a tracking owner that releases the reservation upon disposal. [[1]](diffhunk://#diff-e553b6f86fd8badcfc94a9677f26f7024109f3e54f4ee3cb7eede6a1b260fdf7R15-R16) [[2]](diffhunk://#diff-e553b6f86fd8badcfc94a9677f26f7024109f3e54f4ee3cb7eede6a1b260fdf7L105-R280)
* Updated `SimpleGcMemoryAllocator`, `UnmanagedMemoryAllocator`, and `UniformUnmanagedMemoryPoolMemoryAllocator` to enforce the accumulative allocation limit and to properly track and release allocations. [[1]](diffhunk://#diff-c69fd1def16d5671114b08978382c62e761313b64006fed18e85bc7106bed9b7R15-R40) [[2]](diffhunk://#diff-c69fd1def16d5671114b08978382c62e761313b64006fed18e85bc7106bed9b7L32-R70) [[3]](diffhunk://#diff-11319e2351c49c14f7e4969922b6d104c20a74975a32772f4670b3e7d821f455R23-R43) [[4]](diffhunk://#diff-0a91588ea3b93a3ac505b878cc8ecf5e686055031431613f4891ccb468b21decR98-R155) [[5]](diffhunk://#diff-0a91588ea3b93a3ac505b878cc8ecf5e686055031431613f4891ccb468b21decR188-R192)

**Configuration and API Improvements**

* Extended `MemoryAllocatorOptions` with the new `AccumulativeAllocationLimitMegabytes` property, including validation and documentation. [[1]](diffhunk://#diff-c77aeda122d81a14076b61136286a245dfcb5054f46ec5eae7f8327900734420R13-R21) [[2]](diffhunk://#diff-c77aeda122d81a14076b61136286a245dfcb5054f46ec5eae7f8327900734420R50-R74)
* Improved XML documentation throughout the allocator APIs to clarify the new behaviors and configuration options. [[1]](diffhunk://#diff-e553b6f86fd8badcfc94a9677f26f7024109f3e54f4ee3cb7eede6a1b260fdf7L26-R62) [[2]](diffhunk://#diff-e553b6f86fd8badcfc94a9677f26f7024109f3e54f4ee3cb7eede6a1b260fdf7R114-R117)

**Interface and Code Style Enhancements**

* Made interface members in `IMemoryGroup<T>` explicitly `public` for consistency and clarity. [[1]](diffhunk://#diff-19f70357b71c107ba32df3ea8d1d4cc458913697bb8715b1eb5a43dc39e354c7L18-R23) [[2]](diffhunk://#diff-19f70357b71c107ba32df3ea8d1d4cc458913697bb8715b1eb5a43dc39e354c7L32-R32) [[3]](diffhunk://#diff-19f70357b71c107ba32df3ea8d1d4cc458913697bb8715b1eb5a43dc39e354c7L42-R42)
* Minor code style and documentation improvements in buffer and enumerator implementations. [[1]](diffhunk://#diff-9fda36d054e954c625cd3d0ab9738545bac5bc52fa486fb217be81e5cfaa6276L34-R51) [[2]](diffhunk://#diff-11319e2351c49c14f7e4969922b6d104c20a74975a32772f4670b3e7d821f455R5)

These changes collectively provide stronger safety against excessive memory consumption, offer more configurability for consumers, and improve maintainability and clarity of the memory allocation subsystem.


antonfirsov:
I'm planning to take a look later this week.

JimBobSquarePants:
@antonfirsov Thanks mate!

src/ImageSharp/Memory/Allocators/MemoryAllocator.cs
```diff
@@ -23,9 +25,41 @@ public abstract class MemoryAllocator
...
+    internal long MemoryGroupAllocationLimitBytes { get; private protected set; } = Environment.Is64BitProcess ? 4L * OneGigabyte : OneGigabyte;
+
+    /// <summary>
+    /// Gets the maximum allowed total allocation size, in bytes, for the current process.
+    /// </summary>
+    /// <remarks>
+    /// The allocation limit is determined based on the process architecture. For 64-bit processes,
+    /// the limit is higher than for 32-bit processes.
+    /// </remarks>
+    internal long AccumulativeAllocationLimitBytes { get; private protected set; } = Environment.Is64BitProcess ? 8L * OneGigabyte : 2L * OneGigabyte;
```
antonfirsov: This is an unwanted breaking change. Users operating high load app/service on decent hardware may hit this arbitrary limitation pretty quickly, requiring them to extend it manually.

Even if we choose to limit the total outstanding memory, the default limit should be infinite.

antonfirsov (commented):
I believe the complexity introduced in this PR doesn't justify the value.

Capping the total outstanding memory acts like a duplication of the platform OOM, but without the actual knowledge of the underlying memory constraints -- instead dying at an arbitrary value the user would need to fine-tune. I don't see this improving ease of use or security.

Even if we decide that a total cap is the way, I think a way simpler implementation should be possible. When allocating a group we can do the following:
- Check if the allocator has the required memory throw if it doesn't. This should handle 99% of the cases without introduce any suppression mechanism.
- As we process the individual allocations, escape at the first `InvalidMemoryOperationException` and dispose the already allocated buffers. Since this is the slow path anyways, the added allocation/deallocation ceremony should be ok.
- Pull up the tracking logic to base classes.

I believe however, that it would be better to go with my original proposal from https://github.com/SixLabors/ImageSharp/issues/2850. With that, allocating an outlier oversized GIF would fail with `InvalidImageContentException` would fail the same way (and  for the same conditions) as allocating an outlier oversized JPG.

I can explore a minimalistic & heuristic solution for that. If it's not doable #2850 is probably not worth the trouble.

JimBobSquarePants:
> I believe the complexity introduced in this PR doesn't justify the value.
> 
> Capping the total outstanding memory acts like a duplication of the platform OOM, but without the actual knowledge of the underlying memory constraints -- instead dying at an arbitrary value the user would need to fine-tune. I don't see this improving ease of use or security.
> 
> Even if we decide that a total cap is the way, I think a way simpler implementation should be possible. When allocating a group we can do the following:
> 
> * Check if the allocator has the required memory throw if it doesn't. This should handle 99% of the cases without introduce any suppression mechanism.
> * As we process the individual allocations, escape at the first `InvalidMemoryOperationException` and dispose the already allocated buffers. Since this is the slow path anyways, the added allocation/deallocation ceremony should be ok.
> * Pull up the tracking logic to base classes.
> 
> I believe however, that it would be better to go with my original proposal from #2850. With that, allocating an outlier oversized GIF would fail with `InvalidImageContentException` would fail the same way (and for the same conditions) as allocating an outlier oversized JPG.
> 
> I can explore a minimalistic & heuristic solution for that. If it's not doable #2850 is probably not worth the trouble.

Thanks for the review @antonfirsov it's very much appreciated. 

I understand the concern about duplicating OOM and introducing an arbitrary cap. The intent here isn't to second-guess the runtime, it's to provide an explicit circuit breaker at the only layer that has full visibility of actual allocated bytes. Something that triggers before we allow someone to bring down a server.

Decoder-level heuristics, whether frame count or pixel count, can't reliably model real memory usage because it depends on pixel format, intermediate buffers, processing, encoding, and concurrency. The allocator is the only place that sees the real byte pressure across all of that, which is why I chose to enforce the cap there rather than per decoder.

I agree that defaulting to a finite value can be seen as a breaking change. I'm absolutely fine with making the default unlimited so existing high-load workloads aren't impacted unless users opt in.

If there's a simpler way to structure the tracking within the allocator base types without the suppression mechanics, I'm open to that. The core goal isn't complexity, it's having a byte-accurate outstanding limit that protects long-running services from hitting OOM.

I don't see this as replacing the #2850 proposal. That addresses format-level validation symmetry. This addresses process-level memory safety across decode, encode, and processing.

src/ImageSharp/Memory/Allocators/MemoryAllocator.cs
JimBobSquarePants: Defaults to `long.MaxValue` now.

antonfirsov:
> I don't see this as replacing the https://github.com/SixLabors/ImageSharp/issues/2850 proposal

Good we agree on this :) I think these are two distinct features, see my latest comment on #3063 regarding that proposal.

> This addresses process-level memory safety across decode, encode, and processing

I'm still not 100% convinced. I would argue that it's more important that we provide tools (e.g. #2850) to avoid such application states for common use-cases (including processing of input from untrusted sources) than to provide extra mechanisms to deal with situations where memory is already running out.

Do you have a scenario in mind where recovering from an `InvalidMemoryOperationException` would better than recovering from an OOM?

PS: It really doesn't help with the GIF bomb! A bunch of frames would be allocated and decoded (using CPU in vain!) already by the time the decoder hits the `InvalidMemoryOperationException`, and if I'm reading the code correctly, there is nothing disposing the those frames the process would need to wait for the finalizer to do the cleanup.

JimBobSquarePants:
> > I don't see this as replacing the #2850 proposal
> 
> Good we agree on this :) I think these are two distinct features, see my latest comment on #3063 regarding that proposal.
> 
> > This addresses process-level memory safety across decode, encode, and processing
> 
> I'm still not 100% convinced. I would argue that it's more important that we provide tools (e.g. #2850) to avoid such application states for common use-cases (including processing of input from untrusted sources) than to provide extra mechanisms to deal with situations where memory is already running out.
> 
> Do you have a scenario in mind where recovering from an `InvalidMemoryOperationException` would better than recovering from an OOM?
> 
> PS: It really doesn't help with the GIF bomb! A bunch of frames would be allocated and decoded (using CPU in vain!) already by the time the decoder hits the `InvalidMemoryOperationException`, and if I'm reading the code correctly, there is nothing disposing the those frames the process would need to wait for the finalizer to do the cleanup.

I must have written a reply and lost it in the sea of open tabs I have at least 5 times now! 

The major difference with `InvalidMemoryOperationException` as I see it is you can set limits that fire before OOM even occurs and protect the server. OOMs aren't something you commonly recover from as I understand it. 

Nothing disposes of those frames, but I think waiting for the finalizer is reasonable. 

I'll add the extra method to calculate projected memory usage to the `ImageInfo` separately. I'm also looking to add another event to the ImageSharp.Web middleware so users can inject it.

JimBobSquarePants:
`ImageInfo.GetPixelMemorySize()` Added See https://github.com/SixLabors/ImageSharp/pull/3108

I still believe this to be viable additional defense.

antonfirsov:
I wanted to come back to this after the canvas review :( I started an alternative implementation on a separate branch, invested many hours already. I wanted to get meaningful conclusions out of it, it might make sense to tweak public API-s to solve this properly.

The current form adds complexity and performance costs that are not proportional to the benefits of an opt-in feature. The suppression count tracking in the allocator, the the extra allocations are especially bad. With this PR everyone has to pay for them though I only expect a minority of users to enable this.

JimBobSquarePants:
> I wanted to come back to this after the canvas review :( I started an alternative implementation on a separate branch, invested many hours already. I wanted to get meaningful conclusions out of it, it might make sense to tweak public API-s to solve this properly.
> 
> The current form adds complexity and performance costs that are not proportional to the benefits of an opt-in feature. The suppression count tracking in the allocator, the the extra allocations are especially bad. With this PR everyone has to pay for them though I only expect a minority of users to enable this.

Sorry, I didn't see you had started something at all. I'm just trying to ship the libraries within a reasonable time frame. 

If it's any consolation I'm seeing no slowdown.

| Method                              | Mean      | Error     | StdDev   | Ratio |
|------------------------------------ |----------:|----------:|---------:|------:|
| 'MagicScaler Load, Resize, Save'    |  33.62 ms |  0.831 ms | 0.216 ms |  0.17 |
| 'ImageSharp TD Load, Resize, Save'  |  38.90 ms |  2.639 ms | 0.685 ms |  0.20 |
| 'ImageSharp Load, Resize, Save'     |  61.20 ms |  1.509 ms | 0.392 ms |  0.32 |
| 'NetVips Load, Resize, Save'        |  63.28 ms |  3.007 ms | 0.781 ms |  0.33 |
| 'SkiaSharp Load, Resize, Save'      |  79.12 ms |  1.274 ms | 0.331 ms |  0.41 |
| 'ImageFree Load, Resize, Save'      | 134.67 ms |  2.775 ms | 0.429 ms |  0.70 |
| 'ImageFlow Load, Resize, Save'      | 141.07 ms |  1.968 ms | 0.511 ms |  0.73 |
| 'System.Drawing Load, Resize, Save' | 192.23 ms |  3.628 ms | 0.942 ms |  1.00 |
| 'ImageMagick Load, Resize, Save'    | 210.09 ms | 11.078 ms | 2.877 ms |  1.09 |

antonfirsov:
>  I'm seeing no slowdown.

LoadResizeSave is not heavy on allocator usage. I expect extensive allocator-only before/after benchmarks to show some regression. If I understand the intent of #3102, allocation costs matter. But my main concern is not performance, it's the technical dept the weird implementation adds. Besides the points I mentioned, there is a now another layer of lifetime-management independent from `UnmanagedBufferLifetimeGuard` which can be also consolidated into a single point.

I really had a lot of thoughts stacking up here but didn't get a chance to share them since I was focusing on the canvas review. I'm absolutely bullish about fixing/changing this for V4 since I still feel ownership responsibilities over the memory subsystem.

>  I'm just trying to ship the libraries within a reasonable time frame.

If you want to ship ASAP, reverting the PR is also an option. We can always add a non-breaking change in a patch if we figure out a reasonable way to do it.

PR #3096, opened 2026-03-29
Add support for OpenEXR image format

brianpopow:
### Prerequisites

### Description

This PR adds support for decoding and encoding of images in the OpenEXR format. OpenEXR is a HDR image format, therefore I have added two new pixel formats `Rgb96` and `Rgba128` which store each color channel as `UInt32`.

The specification can be found here: https://openexr.com/en/latest/index.html#openexr

A reference implementation can be found here: https://github.com/AcademySoftwareFoundation/openexr.git


JimBobSquarePants:
@brianpopow I'll review this asap.

src/ImageSharp/Common/Helpers/ColorNumerics.cs
```diff
@@ -132,6 +134,15 @@ public static byte From16BitTo8Bit(ushort component) =>
...
         (byte)(((component * 255) + 32895) >> 16);
 
+    /// <summary>
+    /// Scales a value from an 32 bit <see cref="uint"/> to
+    /// an 8 bit <see cref="byte"/> equivalent.
+    /// </summary>
+    /// <param name="component">The 32 bit component value.</param>
+    /// <returns>The <see cref="byte"/> value.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static byte From32BitTo8Bit(uint component) => (byte)(component * Scale32Bit);
```
antonfirsov: Perf is maybe not a primary concern at this point but I wonder if there is a cheaper way to correctly calculate this that avoids FP conversion? How does the reference implementation calculate this (if they care about decoding into lower res format)?

src/ImageSharp/IO/BufferedReadStream.cs
```diff
@@ -166,6 +167,31 @@ public override int ReadByte()
...
+    public float ReadSingle(byte[] data)
+    {
+        int offset = 0;
+        int bytesToRead = 4;
+        while (bytesToRead > 0)
+        {
+            int bytesRead = this.Read(data, offset, bytesToRead);
+            if (bytesRead == 0)
+            {
+                throw new ImageFormatException("Not enough data to read a float value from the stream");
```
antonfirsov: BufferedReadStream doesn't expose such methods and it is especially odd to throw ImageFormatException here. Wouldn't it be better to keep this in the codec implementation?

src/ImageSharp/Formats/Exr/Compression/Decompressors/NoneExrCompression.cs
```diff
@@ -0,0 +1,24 @@
...
+internal class NoneExrCompression : ExrBaseDecompressor
+{
+    public NoneExrCompression(MemoryAllocator allocator, uint bytesPerBlock, uint bytesPerRow)
+        : base(allocator, bytesPerBlock, bytesPerRow)
+    {
+    }
+
+    /// <inheritdoc/>
+    public override void Decompress(BufferedReadStream stream, uint compressedBytes, Span<byte> buffer)
+        => stream.Read(buffer, 0, Math.Min(buffer.Length, (int)this.BytesPerBlock));
```
antonfirsov: Is it safe to omit checking the return value here?

src/ImageSharp/Formats/Exr/ExrDecoderCore.cs
```diff
@@ -0,0 +1,778 @@
...
+        byte version = (byte)stream.ReadByte();
+        if (version != 2)
+        {
+            ExrThrowHelper.ThrowNotSupportedVersion();
+        }
+
+        // Next three bytes contain info's about the image.
+        byte flagsByte0 = (byte)stream.ReadByte();
+        stream.ReadByte();
+        stream.ReadByte();
```
antonfirsov: Return value checks omitted here

src/ImageSharp/Formats/Exr/ExrDecoderCore.cs
```diff
@@ -0,0 +1,778 @@
...
+        List<ExrChannelInfo> channels = [];
+        while (attributeSize > 1)
+        {
+            ExrChannelInfo channelInfo = this.ReadChannelInfo(stream, out int bytesRead);
+            channels.Add(channelInfo);
+            attributeSize -= bytesRead;
+        }
+
+        // Last byte should be a null byte.
+        stream.ReadByte();
```
antonfirsov: Also here

src/ImageSharp/Formats/Exr/ExrDecoderCore.cs
```diff
@@ -0,0 +1,778 @@
...
+
+        ExrPixelType pixelType = (ExrPixelType)this.ReadSignedInteger(stream);
+        bytesRead += 4;
+
+        byte pLinear = (byte)stream.ReadByte();
+
+        // Next 3 bytes are reserved bytes and not use.
+        stream.ReadByte();
+        stream.ReadByte();
+        stream.ReadByte();
```
antonfirsov: Also here. Would it make sense to read 4 bytes to `this.buffer` in this case?

antonfirsov (commented):
Couple of places where the return value from `stream.Read(...)` is not checked. I may have missed some, it is worth a comprehensive check.

src/ImageSharp/IO/BufferedReadStream.cs
antonfirsov: I see now that there are similar methods in the codec implementation, but they don't seem to iterate. Does it make sense to do it here? Doesn't the BufferedReadStream logic make it a very rare case that we can't read such small buffers immediately? What is our standard way of handling this in other codecs?

JimBobSquarePants: We don't have any extensions on `BufferedReadStream` like this elsewhere. There was one attempted in this [PR](https://github.com/SixLabors/ImageSharp/pull/2899/changes#r2031160668) but I've already said no to that because it is dangerous. 

Personally, I would avoid them because we then start introducing endianess.

brianpopow: I have moved the ReadSingle() method to the ExrDecoderCore now

src/ImageSharp/Formats/Exr/Compression/Decompressors/NoneExrCompression.cs
brianpopow: No it is not, I have added here and in the other places checks, that enough data was read: 18532898

src/ImageSharp/Common/Helpers/ColorNumerics.cs
antonfirsov: @brianpopow actually, I don't understand this. `From32BitTo8Bit(uint.MaxValue) == 1`. Are you sure this logic is correct? If it's incorrect, I wonder why tests didn't catch it? Wouldn't `(byte)(component >> 16)` do the job?

brianpopow: @antonfirsov Yes I think you are right, this does not look right. 

>  I wonder why tests didn't catch it?

There are no pixel conversion tests yet for the new pixel types. I will try to add tests for the new pixlel types `Rgb96` and `Rgba128` next.

> Wouldn't (byte)(component >> 16) do the job?

Yeah that's more like it, but I think it should be `(byte)(component >> 24)`? Lets say we have for example `Int.Max / 2`, e.g. 2147483647. Converting this to 8 bit should be 127.

JimBobSquarePants:
@brianpopow I'll be able to focus on a proper review this week.

src/ImageSharp/Formats/Exr/Compression/Compressors/NoneExrCompressor.cs
```diff
@@ -0,0 +1,30 @@
...
+
+internal class NoneExrCompressor : ExrBaseCompressor
+{
+    public NoneExrCompressor(Stream output, MemoryAllocator allocator, uint bytesPerBlock, uint bytesPerRow)
+        : base(output, allocator, bytesPerBlock, bytesPerRow)
+    {
+    }
+
+    /// <inheritdoc/>
+    public override ExrCompression Method => ExrCompression.Zip;
```
JimBobSquarePants: Is this correct?

src/ImageSharp/PixelFormats/PixelImplementations/Rgba128.cs
```diff
@@ -0,0 +1,167 @@
...
+
+/// <summary>
+/// Pixel type containing four 32-bit unsigned normalized values ranging from 0 to 4294967295.
+/// The color components are stored in red, green, blue and alpha.
+/// <para>
+/// Ranges from [0, 0, 0, 0] to [1, 1, 1, 1] in vector form.
+/// </para>
+/// </summary>
+[StructLayout(LayoutKind.Sequential)]
+public partial struct Rgba128 : IPixel<Rgba128>, IEquatable<Rgba128>
```
JimBobSquarePants, line 18: Doesn't this mean our processes are all actually lossy?

src/ImageSharp/PixelFormats/PixelImplementations/Rgba128.cs
```diff
@@ -0,0 +1,167 @@
...
+    public static Rgba128 FromRgba64(Rgba64 source) => new(source.R, source.G, source.B, source.A);
+
+    /// <inheritdoc/>
+    public static PixelTypeInfo GetPixelTypeInfo() => PixelTypeInfo.Create<Rgba128>(
+            PixelComponentInfo.Create<Rgba128>(4, 32, 32, 32),
+            PixelColorType.RGB,
+            PixelAlphaRepresentation.Unassociated);
+
+    /// <inheritdoc/>
+    public readonly Rgba32 ToRgba32() => throw new NotImplementedException();
```
JimBobSquarePants: We need to implement all these.

tests/Directory.Build.targets
```diff
@@ -24,7 +24,7 @@
     Do not update to 14+ yet. There's differnce in how the BMP decoder handles rounding in 16 bit images.
     See https://github.com/ImageMagick/ImageMagick/commit/27a0a9c37f18af9c8d823a3ea076f600843b553c
     -->
-    <PackageReference Update="Magick.NET-Q16-AnyCPU" Version="13.10.0" />
+    <PackageReference Update="Magick.NET-Q16-AnyCPU" Version="14.11.1" />
```
JimBobSquarePants, line 27: Oh, you got this working!! Fantastic!! 🙌

src/ImageSharp/Formats/Exr/Compression/Decompressors/ZipExrCompression.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                break;
+            }
+
+            totalRead += bytesRead;
+        }
+
+        if (totalRead == 0)
+        {
+            ExrThrowHelper.ThrowInvalidImageContentException("Could not read zip compressed image data!");
```
JimBobSquarePants: I wonder if we should just break here? Normally we try to return as much pixel information as we can.

src/ImageSharp/Formats/Exr/Compression/ExrBaseDecompressor.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Six Labors.
+// Licensed under the Six Labors Split License.
+
+using SixLabors.ImageSharp.IO;
+using SixLabors.ImageSharp.Memory;
+
+namespace SixLabors.ImageSharp.Formats.Exr.Compression;
+
+internal abstract class ExrBaseDecompressor : ExrBaseCompression
```
JimBobSquarePants, line 12: I know these types are internal but I'm finding it very useful currently to document the crap out of everything (AI really helps here.)

tests/Directory.Build.targets
brianpopow: I had to upgrade to the latest version of Magick.NET, because of an issue decoding some exr files, which was fixed [here](https://github.com/ImageMagick/ImageMagick/pull/8285)

Unfortunately this also changed how some bitmaps which are RLE compressed are decoded. I had to change some testcases to use `CompareToReferenceOutput` instead of using the reference decoder: 0c3ca3da

brianpopow:
> @brianpopow I'll be able to focus on a proper review this week.

@JimBobSquarePants: dont worry keep your time, it is not urgent. 

Just as a heads up: I noticed issues with the pixel conversion methods introduced with the new pixel types. I am trying to add tests for them in the next days. Those conversion methods will change, so maybe exclude them from a review until then, because they will change anyway.

tests/Directory.Build.targets
JimBobSquarePants: I'm absolutely fine with `CompareToReferenceOutput `. Our decoder is rock solid now and that should be enough to protect against regression.

tests/Images/External/ReferenceOutput/BmpDecoderTests/BmpDecoder_CanDecode_RunLengthEncoded_4Bit_WithDelta_Rgba32_pal4rlecut.png
JimBobSquarePants, line 1: This output looks bugged to me. Has the Magick comparison been hiding an issue?

src/ImageSharp/Formats/Exr/Compression/Compressors/NoneExrCompressor.cs
brianpopow: no, but is also not needed. I made this similar to how the TiffCompressors are defined, but currently the compression property is not used. I have removed it because of that.

src/ImageSharp/PixelFormats/PixelImplementations/Rgba128.cs
brianpopow: this is now implemented

src/ImageSharp/PixelFormats/PixelImplementations/Rgba128.cs
brianpopow: Can you elaborate on this, I do not understand exactly what you mean here.

tests/Images/External/ReferenceOutput/BmpDecoderTests/BmpDecoder_CanDecode_RunLengthEncoded_4Bit_WithDelta_Rgba32_pal4rlecut.png
brianpopow: The image looks like this:
pal4rlecut

The image is from the [bmp suite](https://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html) where alot of the edge case image in the tests come from.

From the description of the image:
> An RLE-compressed image that uses “delta” codes, and early EOL & EOBMP markers, to skip over some pixels. It’s okay if the viewer’s image doesn’t exactly match any of the reference images.

It looks like the third image in the "correct display" list: the undefined pixels are black.

src/ImageSharp/Formats/Exr/Compression/ExrBaseDecompressor.cs
brianpopow: I have added more docs with 1e69aa45

src/ImageSharp/Formats/Exr/Compression/Decompressors/ZipExrCompression.cs
brianpopow: I think it is better to throw here, when really 0 bytes are read the data is clearly corrupt.

src/ImageSharp/Common/Helpers/ColorNumerics.cs
brianpopow: I have added tests for the pixel conversion methods for the new pixel types now. There was some issues with the conversions, I think they are fixed now for the most part (they need a review though).

The only thing I could not figure out yet is an issue with `FromVector4()` in the new pixel types.
Currently it looks like this (Similar to how ``Rgb48 does it`):
```
public static Rgb48 FromVector4(Vector4 source)
{
    source = Numerics.Clamp(source, Vector4.Zero, Vector4.One) * Max;
    return new Rgb48((ushort)MathF.Round(source.X), (ushort)MathF.Round(source.Y), (ushort)MathF.Round(source.Z));
}
```
Note: Max is here uint.Max.

The issue is here with the edge case of the maximum values. For example when source is (1.0, 1.0, 1.0, 1.0)
Then source.X will become uint.Max as a float value and casting this back to `uint` will result in 0 with .net8.0.
This seems to be changed with .net 9.0: [fp-to-integer](https://learn.microsoft.com/en-us/dotnet/core/compatibility/jit/9.0/fp-to-integer)

Has anyone an idea howto do this correctly?

tests/Images/External/ReferenceOutput/BmpDecoderTests/BmpDecoder_CanDecode_RunLengthEncoded_4Bit_WithDelta_Rgba32_pal4rlecut.png
JimBobSquarePants: Thanks for checking.

src/ImageSharp/PixelFormats/PixelImplementations/Rgba128.cs
JimBobSquarePants: Sorry what I meant was - Is this pixel format storing more data per component than `Vector4` allows but no, I misread, the're both 32-bits per component.

src/ImageSharp/Common/Helpers/ColorNumerics.cs
```diff
@@ -132,6 +132,15 @@ public static byte From16BitTo8Bit(ushort component) =>
...
         (byte)(((component * 255) + 32895) >> 16);
 
+    /// <summary>
+    /// Scales a value from an 32 bit <see cref="uint"/> to
+    /// an 8 bit <see cref="byte"/> equivalent.
+    /// </summary>
+    /// <param name="component">The 32 bit component value.</param>
+    /// <returns>The <see cref="byte"/> value.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static byte From32BitTo8Bit(uint component) => (byte)(component >> 24);
```
JimBobSquarePants: This is actually truncation, not scaling, and introduces bias across the range. Here's a test demonstrating that vs a safer scaling operation like the one we have for `From16BitTo8Bit`.

```c#
[Fact]
public void From32BitTo8Bit_CurrentImplementation_IsTruncation_NotScaling()
{
	// Use 8421505 as the input because it is the first integer that should map to
	// 1 under correct 32->8 bit scaling.
	//
	// The scaled value is:
	//
	//     round(component * 255 / uint.MaxValue)
	//
	// To find the smallest input that rounds to 1, solve for the point where the
	// scaled value reaches 0.5:
	//
	//     component * 255 / uint.MaxValue >= 0.5
	//
	// Rearranging:
	//
	//     component >= uint.MaxValue / (2 * 255)
	//     component >= 4294967295 / 510
	//     component >= 8421504.5
	//
	// Since component is an integer, the smallest valid value is:
	//
	//     8421505
	//
	// At this point correct scaling yields 1, while (component >> 24) still yields
	// 0, making this the smallest counterexample.

	uint component = 8421505u;

	byte truncated = CurrentFrom32BitTo8Bit(component);
	byte scaled = CorrectFrom32BitTo8Bit(component);

    // Current implementation: floor(component / 2^24)
    // Correct scaling: round(component * 255 / uint.MaxValue)
	Assert.Equal(0, truncated);
	Assert.Equal(1, scaled);
	Assert.NotEqual(truncated, scaled);

	static byte CurrentFrom32BitTo8Bit(uint value) => (byte)(value >> 24);

	static byte CorrectFrom32BitTo8Bit(uint value) =>
		(byte)((((ulong)component) + 8421504UL) / 16843009UL)
}
```

The correct method would be.

```c#
/// <summary>
/// Scales a value from a 32 bit <see cref="uint"/> to an
/// 8 bit <see cref="byte"/> equivalent.
/// </summary>
/// <param name="component">The 32 bit component value.</param>
/// <returns>The <see cref="byte"/> value.</returns>
public static byte From32BitTo8Bit(uint component) =>

	// To scale to 8 bits from a 32-bit value V the required value is:
	//
	//    (V * 255) / 4294967295
	//
	// Since:
	//
	//    4294967295 = 255 * 16843009
	//
	// this reduces exactly to:
	//
	//    V / 16843009
	//
	// To round to nearest using integer arithmetic we add half the divisor
	// before dividing:
	//
	//    (V + 16843009 / 2) / 16843009
	//
	// where:
	//
	//    16843009 / 2 = 8421504.5
	//
	// Using 8421504 ensures correct round-to-nearest behaviour:
	//
	//    8421504 -> 0
	//    8421505 -> 1
	//
	// The addition must be performed in 64-bit to avoid overflow for large
	// input values (for example uint.MaxValue).
	//
	// Final exact integer implementation:
	//
	//    ((ulong)V + 8421504) / 16843009
	(byte)((((ulong)component) + 8421504UL) / 16843009UL);
```

src/ImageSharp/Common/Helpers/ColorNumerics.cs
JimBobSquarePants: I've added a comment and an implementation for `From32BitTo8Bit` and the current one truncates.

I'll have a look at `FromVector4` in the new pixel types

src/ImageSharp/Common/Helpers/ColorNumerics.cs
JimBobSquarePants: I've added this.

src/ImageSharp/Common/Helpers/ColorNumerics.cs
JimBobSquarePants: I think I've fixed it. Managed to spot the issue via the debugger. You can't use float at `uint.MaxValue` scale for your constant as there is not enough precision. Switching to double made the tests pass on my machine (and hopefully the CI). I've added an explanatory comment to the code.

JimBobSquarePants (approved):
@brianpopow  I fixed the remaining issues I could find. I can't see anything else unless you have more to add. 

Otherwise, happy to merge.

src/ImageSharp/Common/Helpers/ColorNumerics.cs
brianpopow: Thanks for fixing this, I already had a feeling that there might be an issue with this

src/ImageSharp/Common/Helpers/ColorNumerics.cs
brianpopow: great that you found and fixed the issue with it!

brianpopow:
> @brianpopow I fixed the remaining issues I could find. I can't see anything else unless you have more to add.
> 
> Otherwise, happy to merge.

@JimBobSquarePants : No, I dont have more to add. I think it is now good to go in now. Thanks for helping with the last issues with it 👍

PR #3115, opened 2026-04-17
JPEG : Optimize ScaledCopyTo and Block8x8F for common scale factors

### LINKED ISSUE #1782: Post IDCT rounding needs to go - it induces noise & generational loss and takes extra time to compute
New AAN IDCT implementation which I'm working on as part of jpeg decoder optimization/refactoring provides a faster and **more accurate** IDCT calculation. 

Current conversion pipeline has this "conform to libjpeg" step:
https://github.com/SixLabors/ImageSharp/blob/07d50d0e0cbe12be7425379195968584d7350106/src/ImageSharp/Formats/Jpeg/Components/Decoder/JpegBlockPostProcessor.cs#L76-L79

This does a huge quality hit (well, not really 'huge' if we are not talking about a couple of re-encodings) with noise. This is especially noticeable for so called 'generational loss'. Below there's a comparison of rounding vs no-rounding jpeg re-encoding without any change to the original image - no quality or subsampling changed.

**Original image**
tree

**5 generations**
gen5_comparison

*Bottom row is a difference between original image and re-encoded one 

**25 generations**
gen25_comparison

---------------------------------------

Of course users won't re-encode their images 5 or even 25 times but this example shows the uselessness of IDCT rounding. Right now ImageSharp has a slower but more accurate IDCT which should be a 'feature', not a 'downside' which must be hidden. And on top of that it takes extra time to compute so it's a win-win :P

Issues like https://github.com/SixLabors/ImageSharp/issues/245 can't be solved simply because JFIF is a **lossy** image container - 1/255 pixel miss is an expected trade-off.

**Update**
Same applies to downscaling decoding from https://github.com/SixLabors/ImageSharp/pull/2076.

No-rounding code provides slightly better PSNR results but linked PR would provide IDCT algorithms with rounding before this issue is resolved.

@JimBobSquarePants @antonfirsov

JimBobSquarePants:
### Prerequisites

### Description

Improves the performance of the JPEG block scaling and normalization logic in `Block8x8F`. The main changes are the addition of specialized, loopless scaling routines for common scale factors and the refactoring of the SIMD normalization methods to clarify their responsibilities and remove unnecessary rounding. Fixes #1782 

The rounding removal affected our test comparisons far less than I expected but yielded clear visible improvements in some images. For example the replaced image here had blocking that was visible to my eye.

image

### Benchmarks

Performance is sitting pretty following the changes. The playground benchmark has us sitting firmly in 2nd and 3rd place.

```
BenchmarkDotNet v0.15.8, Windows 11 (10.0.26200.8117/25H2/2025Update/HudsonValley2)
AMD RYZEN AI MAX+ 395 w/ Radeon 8060S 3.00GHz, 1 CPU, 32 logical and 16 physical cores
.NET SDK 10.0.202
  [Host]   : .NET 10.0.6 (10.0.6, 10.0.626.17701), X64 RyuJIT x86-64-v4
  ShortRun : .NET 10.0.6 (10.0.6, 10.0.626.17701), X64 RyuJIT x86-64-v4

WarmupCount=5
```

| Method                              | Mean      | Error    | StdDev   | Ratio |
|------------------------------------ |----------:|---------:|---------:|------:|
| 'MagicScaler Load, Resize, Save'    |  33.30 ms | 1.865 ms | 0.289 ms |  0.17 |
| 'ImageSharp TD Load, Resize, Save'  |  39.29 ms | 2.155 ms | 0.560 ms |  0.20 |
| 'ImageSharp Load, Resize, Save'     |  59.90 ms | 1.506 ms | 0.233 ms |  0.31 |
| 'NetVips Load, Resize, Save'        |  62.45 ms | 3.032 ms | 0.787 ms |  0.32 |
| 'SkiaSharp Load, Resize, Save'      |  77.42 ms | 2.112 ms | 0.548 ms |  0.40 |
| 'ImageFree Load, Resize, Save'      | 133.77 ms | 0.498 ms | 0.077 ms |  0.70 |
| 'ImageFlow Load, Resize, Save'      | 139.14 ms | 4.920 ms | 1.278 ms |  0.72 |
| 'System.Drawing Load, Resize, Save' | 192.39 ms | 2.858 ms | 0.442 ms |  1.00 |
| 'ImageMagick Load, Resize, Save'    | 207.08 ms | 9.209 ms | 1.425 ms |  1.08 |


src/ImageSharp/Formats/Jpeg/Components/Block8x8F.ScaledCopy.cs
```diff
@@ -29,7 +29,50 @@ public void ScaledCopyTo(ref float areaOrigin, int areaStride, int horizontalSca
...
+
+        if (horizontalScale == 4 && verticalScale == 4)
+        {
+            this.CopyTo4x4Scale(ref areaOrigin, (uint)areaStride);
+            return;
+        }
+
+        // The common 1x, 2x, and 4x integral scales are specialized above.
+        // Uncommon legal factor-3 scales use the generic fallback.
         this.CopyArbitraryScale(ref areaOrigin, (uint)areaStride, (uint)horizontalScale, (uint)verticalScale);
```
antonfirsov, line 76: This method is now inlining a huge implementation blocks which is likely hurting instruction cache efficiency.

If I remember it correctly I skipped the implementation of 1:2, 2:1 etc because I found them somewhat less common less common than 1:2, 2:1.

It might make sense to move this entire block into an `InliningOptions.ColdPath` method.

antonfirsov (commented):
I assume you generated the new reference images by the ImageSharp decoder itself, if I'm mistaken, please ignore this.

Note that AFAIK we never compare codec output to (a version of) itself in tests. This was a strategic direction I pushed for from the beginning in order to make sure we understand how our codecs compare to (de-facto) standard or conformant implementations and understand the amount and the nature of deviations if there are any.  I would very strongly recommend to keep following this strategy in a codec as important as JPEG.

The existing inputs were generated using libjpeg-turbo via ImageMagick CLI. **libjpeg-turbo also has a floating-point pipeline** which could be opted into by setting `jpeg:dct-method=float`. The input generation shouldn't bring too much extra effort, LLMs could likely speed up the work.

Note that it would be likely better to use the IJG libjpeg float pipeline as reference, but I haven't researched ways to do it.

JimBobSquarePants:
> The existing inputs were generated using libjpeg-turbo via ImageMagick CLI.

@antonfirsov I currently have trust issues with ImageMagick. I keep seeing major regression issues in minor releases (14.12.0 appears to break JPEG based TIFF decoding). 

> I assume you generated the new reference images by the ImageSharp decoder itself, if I'm mistaken, please ignore this.

Generated JPEG references for new tests have been using our decoder output for years now. For other formats we haven't been able to find a reliable decoder we can use to compare to as they either ship with issues (BMP) or do not support particular format configurations (TIFF)

I'd say almost all the regenerated references are new versions of our own output.

src/ImageSharp/Formats/Jpeg/Components/Block8x8F.ScaledCopy.cs
JimBobSquarePants: Yeah, great call! The previous codegen had grown to 6634 bytes. With the change, the method is down to 2089 bytes.

Benchmarks look solid also.

antonfirsov: Nit: I rather meant moving the "colder" branches to a separate method then inlining than uninlining the whole method. The current state is good enough though.

antonfirsov (approved):
> Generated JPEG references for new tests have been using our decoder output for years now.

The only reason I originally implemented JPEG testing with reference images was because running  a reference decoder at runtime was impossible or unreliable at that time (also it's expensive for the heavy JPEG test suite). I never wanted the reference images to be self-generated.

Given the release pressure I'm letting this go for now despite the disagreement.

PR #3120, opened 2026-04-19
Add AllocationTrackedMemoryManager and refactor allocators

JimBobSquarePants:
### Prerequisites

### Description

An updated implementation of #3056 

@antonfirsov your comment in the PR was the key to a better implementation for me.

>it might make sense to tweak public API-s to solve this properly

I was so determined not to make a breaking change that I massively overcomplicated the implementation and made it inefficient. 

This PR changes things by implementing a new type `AllocationTrackedMemoryManager` which custom allocator memory managers must implement. By making `MemoryAllocator.Allocate` non virtual and adding an abstract `AllocateCore` I can introduce tracking without the allocation and complexity cost.


antonfirsov:
It would be very helpful for my review if we could revert https://github.com/SixLabors/ImageSharp/pull/3056, and rebase this on top of the original state. I want to review the feature implementation as a whole thing and now it's difficult to react to specific changes from that PR.

JimBobSquarePants:
@antonfirsov 

#3056 has been reverted and this PR rebased over the top. Hopefully this is much easier to review now the diff is clean.

antonfirsov:
Thanks a lot! I'll come back to this once the Drawing API is concluded.

src/ImageSharp/Memory/Allocators/SimpleGcMemoryAllocator.cs
```diff
@@ -12,23 +10,24 @@ namespace SixLabors.ImageSharp.Memory;
...
+    public SimpleGcMemoryAllocator()
+        : this(default)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SimpleGcMemoryAllocator"/> class with custom limits.
+    /// </summary>
+    /// <param name="options">The <see cref="MemoryAllocatorOptions"/> to apply.</param>
+    public SimpleGcMemoryAllocator(MemoryAllocatorOptions options) => this.ApplyOptions(options);
```
antonfirsov, line 25: @JimBobSquarePants `MemoryAllocatorOptions` were not meant to be attached to this type (the feature set is mostly orthogonal). Also, I don't think it brings any value to users to use allocation tracking for anything but the default memory allocator. Scoping the tracking feature as an implementation detail of `UniformUnmanagedMemoryPoolMemoryAllocator` could simplify things significantly.

Sorry, I was extremely distracted last week because of starting my new job, but I REALLY wanted to come back to this. I can do a deeper post-review tomorrow afternoon.

antonfirsov: I noticed the release is out now :( Maybe deleting this ctr and `AllocationTrackedMemoryManager<T>` from the public APIS is not much of a breaking change for a minor update that refactors this. Those things shouldn't really be exposed.

JimBobSquarePants: I had to ship I’m afraid. No need to apologize at all. 

I actually think we should treat `SimpleGcMemoryAllocator` the same as others here. I'd rather normalize everything through a single channel and ensure that all implementations have tracking.

PR #3124, opened 2026-04-26
Add support for decoding PXR24 compressed Exr images

brianpopow:
### Prerequisites

### Description

This PR adds support for decoding EXR images with the pxr24 compression.

### DIFF
--- a/src/ImageSharp/Formats/Exr/Compression/ExrBaseDecompressor.cs
+++ b/src/ImageSharp/Formats/Exr/Compression/ExrBaseDecompressor.cs
@@ -19,9 +19,10 @@ internal abstract class ExrBaseDecompressor : ExrBaseCompression
     /// <param name="allocator">The memory allocator.</param>
     /// <param name="bytesPerBlock">The bytes per row block.</param>
     /// <param name="bytesPerRow">The bytes per row.</param>
+    /// <param name="rowsPerBlock">The pixel rows per block.</param>
     /// <param name="width">The number of pixels per row.</param>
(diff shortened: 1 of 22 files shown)


src/ImageSharp/Formats/Exr/Compression/ExrBaseDecompressor.cs
```diff
@@ -30,6 +34,47 @@ protected ExrBaseDecompressor(MemoryAllocator allocator, uint bytesPerBlock, uin
...
+    {
+        long pos = stream.Position;
+        using ZlibInflateStream inflateStream = new(
+                   stream,
+                   () =>
+                   {
+                       int left = (int)(compressedBytes - (stream.Position - pos));
+                       return left > 0 ? left : 0;
+                   });
+        inflateStream.AllocateNewBytes((int)uncompressedBytes, true);
```
JimBobSquarePants: Shouldn't this be

```c#
inflateStream.AllocateNewBytes((int)compressedBytes, true);
```

`AllocateNewBytes` expects the number of compressed bytes available in the current zlib frame; it uses that value to bound reads from the source stream. Passing `uncompressedBytes` can let the inflater read past the compressed EXR block boundary, so this should use `compressedBytes`.

brianpopow: Yes you are right, it should be `compressedBytes`, changed with: c1de5b61

JimBobSquarePants (approved):
LGTM. Nice one! 👍

PR #3136, opened 2026-05-26
Add support for Apple CgBI PNG images

Erik-White:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp/pulls) open
- [x] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description
Adds support for decoding Apple CgBI PNG images, see: https://theapplewiki.com/wiki/PNG_CgBI_Format
I have not given any consideration to encoding since this is a internal Apple format that I don't think is useful outside of that context.

Closes #410

PR #3137, opened 2026-05-27
Improvements to Apple CgBI PNG handling

Erik-White:
### Prerequisites

### Description
Following on from #3136 , this attempts to streamline and improve the related code as well as improve test coverage.

- CgBI images that are not supported are now explicitly rejected (i.e. bit depth != 8 or truecolor)
- Broke out ChunkedReadStream from ZlibInflateStream to separate the concerns instead of mixing modes in ZlibInflateStream
- Broke out CgBI transform from PngDecoderCore into new PngCgbiProcessor static class
- Added tests for PngCgbiProcessor
- Small improvements to the CgBI shuffle methods (e.g. construct the shuffle patterns only once, remove redundant floor(), replace magic numbers where possible)


Erik-White:
Curious to hear your thoughts. The changes should all be relatively self-contained per commit in case there are sections that you don't think should be included.

JimBobSquarePants:
This looks like a lot of changes. Will take time to review

Erik-White:
Understood. As I mentioned in the description, each commit is fairly self contained but I could also break up the PR into a couple of smaller units if that would be helpful?

tests/ImageSharp.Tests/Formats/Png/PngCgbiProcessorTests.cs
```diff
@@ -0,0 +1,174 @@
...
+    [InlineData(7)]
+    [InlineData(8)]
+    [InlineData(15)]
+    [InlineData(16)]
+    [InlineData(17)]
+    [InlineData(31)]
+    [InlineData(32)]
+    [InlineData(33)]
+    [InlineData(64)]
+    public void ApplyTransformVector512_MatchesScalar(int pixelCount) =>
```
JimBobSquarePants: We have `FeatureTestRunner` that we use for this kind of stuff, so we are not running tests unnecessarily

tests/ImageSharp.Tests/Formats/Png/PngCgbiProcessorTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        {
+            byte a = (p % 7) switch
+            {
+                0 => byte.MinValue,
+                1 => byte.MaxValue,
+                _ => (byte)((((p * 37) + 23) & 0xFF) | 1) // never zero
+            };
+
+            // CgBI premultiplied BGRA: c' = c * a / 255
+            byte r = (byte)((p * 13) & 0xFF);
```
JimBobSquarePants: This is screaming AI to me. random numbers with no explanation.

src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
```diff
@@ -131,114 +53,15 @@ public bool AllocateNewBytes(int bytes, bool isCriticalChunk)
...
-        // Note disposing is done.
-        this.isDisposed = true;
+        this.CompressedStream?.Dispose();
+        this.segmentStream?.Dispose();
     }
 
     [MemberNotNullWhen(true, nameof(CompressedStream))]
     private bool InitializeInflateStream(bool isCriticalChunk)
     {
-        // Apple CgBI IDATs omit the zlib CMF/FLG header and the Adler-32 trailer,
```
JimBobSquarePants, line 234: Why do we no longer return early?

src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
```diff
@@ -8,9 +8,11 @@
 namespace SixLabors.ImageSharp.Compression.Zlib;
 
 /// <summary>
-/// Provides methods and properties for deframing streams from PNGs.
+/// Reads chunked input, parses the zlib CMF/FLG header, and exposes a
+/// <see cref="DeflateStream"/> over the remaining DEFLATE payload. The
+/// Adler-32 trailer is not validated.
 /// </summary>
-internal sealed class ZlibInflateStream : Stream
+internal sealed class ZlibInflateStream : IDisposable
```
JimBobSquarePants: I'm not sure I'm comfortable with a type suffixed `Stream` that is no longer inherits `Stream`. This feels like we're violating runtime design guidelines.

JimBobSquarePants:
>replace magic numbers where possible
What magic numbers?

tests/ImageSharp.Tests/Formats/Png/PngCgbiProcessorTests.cs
Erik-White: I did use it to help generate up some test cases, was probably a bit slack in scrutinizing the helper methods

tests/ImageSharp.Tests/Formats/Png/PngCgbiProcessorTests.cs
Erik-White: I'll look in to that

Erik-White:
> > replace magic numbers where possible
> > What magic numbers?

See this commit: https://github.com/SixLabors/ImageSharp/pull/3137/changes/e6f09d66167acae42f180d566e21acac99fe0ffc
Just tried to use existing values where possible

src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
Erik-White: In my first CgBI PR I had added the `noHeader` parameter as a workaround for streams without headers that came from the CgBI images. It worked fine but I didn't like that it mixed concerns in ZLibInflateStream.

This PR now separates out the concerns so ZlibInflateStream always handles streams with headers, and ChunkedReadStream can be used for those without.

So that early exist logic now lives in PngDecoderCore instead of imposing on ZlibInflateStream:
https://github.com/SixLabors/ImageSharp/pull/3137/changes/BASE..4869c344b3e5c7df9e8b696b6cd6433230178a65#diff-e626c30d90afb58e53b15876a4ee894c9581334c8ddbb2fa2be18759a60aea59R780

src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
Erik-White: My thinking was to use composition over inheritance here, especially since most the of the stream methods on ZlibInflateStream just throw NotSupportedException anyway.

But I'm sure ZlibInflateStream could just inherit from ChunkedReadStream (which inherits from Stream) if that would be preferrable? Or would it make sense to just rename this to something like ZlibInflateReader?

tests/ImageSharp.Tests/Formats/Png/PngCgbiProcessorTests.cs
Erik-White: I could use FeatureTestRunner, but it doesn't test quite the same path as these tests do now. This tests the Vector512 path or the fallback regardless of the instruction set available. Using FeatureTestRunner would just end up skipping the 512 path entirely and going to the next available instead.

I don't have strong feelings either way, testing the fallback path is good coverage but maybe not essential

src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
JimBobSquarePants: That means ChunkedReadStream is now a misnomer as it’s still reading zlib compressed stream is it not?

tests/ImageSharp.Tests/Formats/Png/PngCgbiProcessorTests.cs
JimBobSquarePants: It tests the correct path per platform and ensures the filtering and fallbacks are wired up correctly. The CI covers all the SIMD support.

src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
Erik-White: It's just a stream wrapper and is fairly generic so in theory it could be used for other cases. But yes, right now it's only being used for zlib

JimBobSquarePants:
Sorry for the radio silence here @Erik-White 

I've had my head down on some complex drawing stuff. I'll give this one last pass now

src/ImageSharp/Compression/Zlib/ZlibInflateStream.cs
JimBobSquarePants: `ZlibInflateReader` is best IMO.

PR #3167, opened 2026-08-17
Use Collection expressions and Target-typed new

stefannikolei:
1. Collection expressions (`[]` syntax) replacing `new List<T>()`, `new T[] {}`, `.ToArray()`, etc.
2. Target-typed `new()` expressions where appropriate
3. Using spread elements (`[.. collection]`) instead of `.ToArray()` and `.Concat().ToArray()`
4. Adding `#pragma warning disable/restore CS0649` for intentionally unassigned struct fields

@JimBobSquarePants These were all reported from rider and were fixed automatically. Rider/Resharper finds way more cases than Visual Studio


src/ImageSharp/Formats/Exr/ExrMetadata.cs
```diff
@@ -102,7 +102,7 @@ public FormatConnectingMetadata ToFormatConnectingMetadata()
             ? EncodingType.Lossy
             : EncodingType.Lossless;
 
-        return new()
+        return new FormatConnectingMetadata
```
JimBobSquarePants, line 105: I'm not convinced by this. The return type is known.

src/ImageSharp/Formats/Exr/ExrMetadata.cs
```diff
@@ -136,7 +136,7 @@ or PixelColorType.BGR | PixelColorType.Alpha
             }
         };
 
-        return new()
+        return new ExrMetadata
```
JimBobSquarePants, line 139: Same here. I don't think we can trust Rider

src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverterBase.cs
```diff
@@ -119,49 +119,49 @@ private static JpegColorConverterBase[] CreateConverters()
     /// </summary>
     /// <param name="precision">The precision in bits.</param>
     private static JpegColorConverter<YCbCrOperator> GetYCbCrConverter(int precision)
-        => new JpegColorConverter<YCbCrOperator>(precision);
+        => new(precision);
```
JimBobSquarePants, line 122: What's odd about the other instances is this one and below are correct. There's no consistency in Riders findings.

src/ImageSharp/PixelFormats/PixelBlenders/AssociatedAlphaPorterDuffFunctions.Generated.tt
```diff
@@ -60,8 +60,8 @@ internal static partial class AssociatedAlphaPorterDuffFunctions
 }
 <#+
     private static readonly string[] Composers =
-    {
-        "Src",
+        [
+            "Src",
```
JimBobSquarePants, line 64: spacing appears broken in this file

src/ImageSharp/Processing/Processors/Transforms/Resize/ResizeKernel.cs
```diff
@@ -73,10 +73,10 @@ public Span<float> Values
         {
             if (Vector256.IsHardwareAccelerated)
             {
-                return new(this.bufferPtr, this.Length * 4);
+                return new Span<float>(this.bufferPtr, this.Length * 4);
```
JimBobSquarePants, line 76: Return type is known here. Definitely an issue with Rider.

src/ImageSharp/Formats/Jpeg/Components/ColorConverters/JpegColorConverterBase.cs
stefannikolei: It seems to me that when there is more than one line rider seems to say return type unknown.

JimBobSquarePants:
@stefannikolei I guess we just take it as-is for now and let Jetbrains figure out their issues in the future

PR #3175, opened 2026-08-19
Migrate to xUnitV3

stefannikolei:
This upgrades the Tests to xunitV3. Updating to xUnitV3 v4 (What did they smoke creating this versioning) should be done immediately after this. But it has some breaking changes about TestRunners (dotnet test integration).


tests/ImageSharp.Tests/Processing/Processors/Quantization/QuantizerTests.cs
```diff
@@ -78,75 +78,75 @@ public static readonly TheoryData<IQuantizer> Quantizers
...
-        new HexadecatreeQuantizer(Ordered0_75_ScaleDitherOptions),
-        new WebSafePaletteQuantizer(Ordered0_75_ScaleDitherOptions),
-        new WernerPaletteQuantizer(Ordered0_75_ScaleDitherOptions),
-        new WuQuantizer(Ordered0_75_ScaleDitherOptions),
-
-        new HexadecatreeQuantizer(OrderedDitherOptions),
-        new WebSafePaletteQuantizer(OrderedDitherOptions),
-        new WernerPaletteQuantizer(OrderedDitherOptions),
-        new WuQuantizer(OrderedDitherOptions),
+        (IQuantizer)new HexadecatreeQuantizer(Diffuser0_ScaleDitherOptions),
```
JimBobSquarePants, line 94: This is such an awful API regression from XUnit.

stefannikolei: There are 2 apis found. I could not find an better way.

JimBobSquarePants:
I didn't realize xUnitV3 v4 was a thing. We should do this in one jump, not two. It's already painful.

stefannikolei:
Sure we can do that. Can you push it in SharedInfra. 

Then I can do that. 

I did not want to do it in one step. Cause in v4 are changes for mtp and testrunners. That would have been to complicated in one pass

JimBobSquarePants:
Done!

Should be no need to change the test runner. I've configured it to keep the VSTest one because we need that for coverlet

stefannikolei:
> Done!
> 
> Should be no need to change the test runner. I've configured it to keep the VSTest one because we need that for coverlet

It builds with no changes. So I am finished from my side

tests/ImageSharp.Tests/Processing/Processors/Quantization/QuantizerTests.cs
JimBobSquarePants: Yeah, the API is a dog's dinner. You can pass an array to the constructor but that's awful also.

PR #3181, opened 2026-08-24
Precompute and cache RGB/XYZ matrices in RgbWorkingSpace

stefannikolei:
### Prerequisites

### Description
Move RgbToCieXyz and CieXyzToRgb matrix construction from Rgb into RgbWorkingSpace, computing both matrices once at construction. Pass the precomputed inverse adaptation matrix through all chromatic adaptation calls to eliminate repeated Matrix4x4.Invert invocations during color conversions.

| Benchmark | Method | Before | After | Speedup |
|---|---|---:|---:|---:|
| `ColorspaceCieXyzToRgbConvert` | ImageSharp Convert | 79.70 ns | 19.06 ns | **4.18x** |
| `RgbWorkingSpaceAdapt` | ImageSharp Adapt | 110.70 ns | 58.93 ns | **1.88x** |

@JimBobSquarePants  What do you think about this?

### DIFF
--- a/src/ImageSharp/ColorProfiles/VonKriesChromaticAdaptation.cs
+++ b/src/ImageSharp/ColorProfiles/VonKriesChromaticAdaptation.cs
@@ -26,23 +26,8 @@ public static class VonKriesChromaticAdaptation
     /// <returns>The <see cref="CieXyz"/></returns>
     public static CieXyz Transform(in CieXyz source, (CieXyz From, CieXyz To) whitePoints, Matrix4x4 matrix)
     {
-        CieXyz from = whitePoints.From;
-        CieXyz to = whitePoints.To;
-
-        if (from.Equals(to))
-        {
-            return new CieXyz(source.X, source.Y, source.Z);
-        }
-
-        Vector3 sourceColorLms = Vector3.Transform(source.AsVector3Unsafe(), matrix);
-        Vector3 sourceWhitePointLms = Vector3.Transform(from.AsVector3Unsafe(), matrix);
-        Vector3 targetWhitePointLms = Vector3.Transform(to.AsVector3Unsafe(), matrix);
-
-        Vector3 vector = targetWhitePointLms / sourceWhitePointLms;
-        Vector3 targetColorLms = Vector3.Multiply(vector, sourceColorLms);
-
         Matrix4x4.Invert(matrix, out Matrix4x4 inverseMatrix);
-        return new CieXyz(Vector3.Transform(targetColorLms, inverseMatrix));
+        return Transform(in source, whitePoints, matrix, inverseMatrix);
     }
 
     /// <summary>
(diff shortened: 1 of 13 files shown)


src/ImageSharp/ColorProfiles/VonKriesChromaticAdaptation.cs
```diff
@@ -41,23 +82,28 @@ public static CieXyz Transform(in CieXyz source, (CieXyz From, CieXyz To) whiteP
...
+    /// For callers that already hold the inverse, e.g. <see cref="ColorConversionOptions.InverseAdaptationMatrix"/>,
+    /// to avoid repeating <see cref="Matrix4x4.Invert"/> on every call.
+    /// </remarks>
     /// <param name="source">The span to the source colors.</param>
     /// <param name="destination">The span to the destination colors.</param>
     /// <param name="whitePoints">The conversion white points.</param>
     /// <param name="matrix">The chromatic adaptation matrix.</param>
-    public static void Transform(
+    /// <param name="inverseMatrix">The inverse of <paramref name="matrix"/>.</param>
+    internal static void Transform(
```
JimBobSquarePants, line 101: why was this made internal?

src/ImageSharp/ColorProfiles/VonKriesChromaticAdaptation.cs
```diff
@@ -25,6 +25,47 @@ public static class VonKriesChromaticAdaptation
...
+    /// <remarks>
+    /// For callers that already hold the inverse, e.g. <see cref="ColorConversionOptions.InverseAdaptationMatrix"/>,
+    /// to avoid repeating <see cref="Matrix4x4.Invert"/> on every call.
+    /// </remarks>
+    /// <param name="source">The source color.</param>
+    /// <param name="whitePoints">The conversion white points.</param>
+    /// <param name="matrix">The chromatic adaptation matrix.</param>
+    /// <param name="inverseMatrix">The inverse of <paramref name="matrix"/>.</param>
+    /// <returns>The <see cref="CieXyz"/></returns>
+    internal static CieXyz Transform(
```
JimBobSquarePants, line 64: I'm confused now. We have a public and internal version overloads of this method.

JimBobSquarePants:
@stefannikolei A good idea on paper. I'm just confused by the API now.

stefannikolei:
> @stefannikolei A good idea on paper. I'm just confused by the API now. 

I will have a look.

src/ImageSharp/ColorProfiles/VonKriesChromaticAdaptation.cs
stefannikolei: the public ones are the ones without the inverse matrix. I made the methods with the invers matrix internal, to not change the public api. so at the public level we create the inverse matrix by our selfs to not open space for errors. 

We can also go the way to only offer the methods with the inverse matrix as a parameter.

stefannikolei:
> @stefannikolei A good idea on paper. I'm just confused by the API now. 

Should I make all new overloads public? Or only keep the overloads around which accept matrix and inverse?

